Issue #1, opened 2017-08-25
Create the project skeleton


wagerlabs:
Create the rebar configuration, directories, etc. to allow us to start adding blockchain components.

wagerlabs:
@sennui @zp-sd Are we done with this? Please close the ticket if so. You can also close tickets automatically by adding "Closes issue #xxx" to the commit message.

PR #7, opened 2017-08-26
[GH-1] Add project skeleton

zp-sd:
* add rebar
* create first app aecore
* add first dependency - lager
* create Makefile with basic targets


Makefile
```diff
@@ -0,0 +1,69 @@
...
+	@pkill -9 beam || true
+
+
+#
+# Build rules
+#
+
+.SECONDEXPANSION:
+
+build: $$(KIND)
```
sennui: we need to change those, so they look internal: interal_build, internal_start etc.

sennui: please name those with internal prefix (internal_build, internal_start)

zp-sd:
@sennui Done. Please revisit.

Issue #3, opened 2017-08-25
Identify transaction types and define data structures


wagerlabs:
- Collect the different transaction types from the old repo.
- Figure out if we can discard any.
- Define data structures in a single or multiple `.hrl` files, naming them clearly.

PR #30, opened 2017-08-28
GH-3 Identify transaction types and define them in txs.hrl

zp-sd:
@sennui @wagerlabs This is the initial list of transactions (probably incomplete, but this may evolve once we implement). Thoughts?

### DIFF
--- a/apps/aecore/include/txs.hrl
+++ b/apps/aecore/include/txs.hrl
@@ -1,77 +1,88 @@
 -type(pubkey() :: binary()).
 
--record(coinbase_tx, {account = <<>> :: pubkey(),
-                      nonce = 0      :: non_neg_integer()}).
+-record(coinbase_tx, {
+          account = <<>> :: pubkey(),
+          nonce = 0      :: non_neg_integer()}).
 
--record(spend_tx, {from = <<>> :: pubkey(),
-                   to = <<>>   :: pubkey(),
-                   amount = 0  :: non_neg_integer(),
-                   fee = 0     :: non_neg_integer(),
-                   nonce = 0   :: non_neg_integer()}).
+-record(spend_tx, {
+          from = <<>> :: pubkey(),
+          to = <<>>   :: pubkey(),
+          amount = 0  :: non_neg_integer(),
+          fee = 0     :: non_neg_integer(),
+          nonce = 0   :: non_neg_integer()}).
 
 
 %% Channel-related transactions
 
--record(channel_new_tx, {id = 0          :: non_neg_integer(),
-                         account1 = <<>> :: pubkey(),
-                         account2 = <<>> :: pubkey(),
-                         amount1 = 0     :: non_neg_integer(),
-                         amount2 = 0     :: non_neg_integer(),
-                         fee = 0         :: non_neg_integer(),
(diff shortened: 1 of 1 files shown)


sennui:
@wagerlabs I recognise the orignal as emacs-otp indenting. If we can put rules around the first, I dont mind it (I would like just to make it more specific)

apps/aecore/include/txs.hrl
```diff
@@ -0,0 +1,77 @@
...
+                                amount = 0      :: pubkey(),
+                                fee = 0         :: non_neg_integer(),
+                                nonce = 0       :: non_neg_integer()}).
+
+-record(channel_solo_close_tx, {id = 0         :: non_neg_integer(),
+                                account = <<>> :: pubkey(),
+                                fee = 0        :: non_neg_integer(),
+                                nonce = 0      :: non_neg_integer()}).
+
+-record(channel_slash_tx, {id = 0         :: non_neg_integer(),
```
sennui: can we improve naming? slash is not saying anything? is anyone using it and this is why we use slash? Maybe: Overload?

apps/aecore/include/txs.hrl
```diff
@@ -0,0 +1,77 @@
...
+                          delay = 0       :: non_neg_integer()}).
+
+-record(channel_team_close_tx, {id = 0          :: non_neg_integer(),
+                                account1 = <<>> :: pubkey(),
+                                account2 = <<>> :: pubkey(),
+                                amount = 0      :: pubkey(),
+                                fee = 0         :: non_neg_integer(),
+                                nonce = 0       :: non_neg_integer()}).
+
+-record(channel_solo_close_tx, {id = 0         :: non_neg_integer(),
```
sennui: self close?

apps/aecore/include/txs.hrl
wagerlabs: The indent I prefer is how my Emacs is indenting if you leave the opening brace hanging.

apps/aecore/include/txs.hrl
wagerlabs: I didn't comment on the naming because some of the transactions may go eventually. I do think that having both close and solo close is superfluous as the system should be able to recognize that you are closing the channel unilaterally. Besides, it makes for less conflicts to resolve, e.g. when one side did a close and the other one a solo close.

apps/aecore/include/txs.hrl
wagerlabs: Slash is a [cryptocurrency term](https://medium.com/@VitalikButerin/minimal-slashing-conditions-20f0b500fc6c) so I would leave it. We'll evaluate the transaction itself when we get to channels.

zp-sd:
@wagerlabs Yes, agree with indentation comment. @sennui What is proposed is still emacs indenting.

apps/aecore/include/txs.hrl
zp-sd: Well.. do we want to evaluate these transactions now, or leave it as it is, and massage when channels implementation takes place?

zp-sd:
@sennui @wagerlabs  can we merge?

PR #32, opened 2017-08-29
Proof of Work with sha256

akorosmezey:


lucafavatella:
@akorosmezey I would suggest splitting this in two PRs: one only for Travis, and only for PoW.

ci
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+set -ev # Ref https://docs.travis-ci.com/user/customizing-the-build/#Implementing-Complex-Build-Steps
+set -x
+
+case "${1:?}"-"${2:?}" in
+    before_install-tests)
+        BuildDir="${3:?}"
+        ( cd "${BuildDir:?}" && make nose-env; )
```
lucafavatella: E.g. This target appears not to be in makefile.

rebar.config
```diff
@@ -1,7 +1,7 @@
 {erl_opts, [debug_info, {parse_transform, lager_transform}]}.
-{deps, [
-    {lager, ".*", {git, "https://github.com/erlang-lager/lager.git", {tag, "3.5.1"}}}
-]}.
+{deps, [{lager, ".*", {git, "https://github.com/erlang-lager/lager.git", {tag, "3.5.1"}}},
+        {meck, ".*", {git, "git://github.com/eproxus/meck.git", {tag, "0.8.6"}}}
```
lucafavatella: In Makefile I see we are using rebar3. I suggest we use [profiles](http://www.rebar3.org/docs/profiles) so that the test-only dependencies e.g. meck do not end up in the production OTP release.

wagerlabs: +1

lucafavatella:
@akorosmezey I suggest you [use keywords](https://help.github.com/articles/closing-issues-using-keywords/) so that when PR is merged it automatically closes the ticket.

apps/aecore/src/pow_sha256.erl
```diff
@@ -0,0 +1,105 @@
...
+
+
+%%%=============================================================================
+%%% API
+%%%=============================================================================
+
+%%------------------------------------------------------------------------------
+%% Generate a nonce value that, when used in hash calculation of 'Data', results in
+%% a smaller value than the difficulty. Return 'not_found' if the condition is still
+%% not satisdfied after trying 'Count' times with incremented nonce values.
```
lucafavatella: Minor typo.

apps/aecore/src/pow_sha256.erl
```diff
@@ -0,0 +1,105 @@
...
+integer_to_scientific(I) ->
+    Exp = log2(I),
+    256*Exp + (I bsr (Exp - 7)).
+
+%%------------------------------------------------------------------------------
+%% Base-2 integer logarithm
+%%------------------------------------------------------------------------------
+log2(1) ->
+    0;
+log2(N) ->
```
lucafavatella: Guard `when N > 1` for readability?

apps/aecore/src/pow_sha256.erl
```diff
@@ -0,0 +1,105 @@
...
+%%% @end
+%%%=============================================================================
+-module(pow_sha256).
+
+-export([generate/3,
+         recalculate_difficulty/3]).
+
+-ifdef(TEST).
+-compile(export_all).
+-include_lib("eunit/include/eunit.hrl").
```
zp-sd: why `eunit` here?

apps/aecore/src/pow_sha256.erl
```diff
@@ -0,0 +1,105 @@
...
+%%% API
+%%%=============================================================================
+
+%%------------------------------------------------------------------------------
+%% Generate a nonce value that, when used in hash calculation of 'Data', results in
+%% a smaller value than the difficulty. Return 'not_found' if the condition is still
+%% not satisdfied after trying 'Count' times with incremented nonce values.
+%%------------------------------------------------------------------------------
+-spec generate(Data :: sha256:hashable(), Difficulty :: sci_int(), Count :: integer()) ->
+                      integer() | 'not_found'.
```
lucafavatella: Please prefer as return value:
* Either a tuple `{error, Reason::any()}`
* Or a tuple with at least two elements where the first element is `ok`.

lucafavatella: (of course you can put a more specific `Reason` in this case.)

wagerlabs: +1

apps/aecore/src/pow_sha256.erl
```diff
@@ -0,0 +1,104 @@
...
+-endif.
+
+-include("sha256.hrl").
+
+-define(NONCE_RANGE, 1000000000000000000000000).
+
+%% Integers encrypted in scientific notation:
+%%   256 * <base-2 exponent> + the most significant byte (i.e., the significand)
+-type sci_int() :: integer().
+-type pow_result() :: integer() | {'error', 'generation_count_exhausted'}.
```
lucafavatella: I guess you may prefer `{ok, integer()}` to `integer()` here so to ease matching.

apps/aecore/src/pow_sha256.erl
```diff
@@ -0,0 +1,104 @@
...
+-export([generate/3,
+         recalculate_difficulty/3]).
+
+-ifdef(TEST).
+-compile(export_all).
+-endif.
+
+-include("sha256.hrl").
+
+-define(NONCE_RANGE, 1000000000000000000000000).
```
lucafavatella: I see this is 10^24 - a comment would help readability.

akorosmezey:
I accommodated the comments above. Travis stuff moved to separate PR.

apps/aecore/test/pow_sha256_tests.erl
```diff
@@ -0,0 +1,83 @@
...
+      {"Integer to scientific conversion",
+       fun() ->
+               %% 24*256 + 128
+               ?assertEqual(6272, ?TEST_MODULE:integer_to_scientific(16#1000000)),
+               %% 24*256 + 145
+               ?assertEqual(6289, ?TEST_MODULE:integer_to_scientific(16#1230000))
+       end},
+      {"Integer to scientific and back",
+       fun() ->
+               %% can be converted w/o loosing accuracy
```
lucafavatella: minor typo "lose" - also a few lines below.

apps/aecore/test/pow_sha256_tests.erl
```diff
@@ -0,0 +1,83 @@
+%%%=============================================================================
+%%% @copyright (C) 2017, Aeternity Anstalt
+%%% @doc
+%%%   Unit tests for the pow_sha256 module
```
zp-sd: Do we plan to write `@doc`s for all modules? To me these are unnecessary noice, but that's only my personal opinion.

@wagerlabs @sennui ?

rebar.config
```diff
@@ -19,6 +18,10 @@
                              {include_erts, false},
                              {include_src, true}]}]
             },
+            {test, [{relx, [{dev_mode, true},
+                             {include_erts, false},
```
zp-sd: 💅 indentation

apps/aecore/src/pow_sha256.erl
```diff
@@ -0,0 +1,106 @@
...
+
+%% Integers encrypted in scientific notation:
```
lucafavatella: I would suggest linking https://en.wikipedia.org/wiki/Scientific_notation (or similar).

lucafavatella: encrypted --> represented ?

apps/aecore/test/pow_sha256_tests.erl
wagerlabs: Let's drop the docs for now and get to them once the code has settled.

apps/aecore/src/pow_sha256.erl
```diff
@@ -0,0 +1,106 @@
...
+%%   256 * <base-2 exponent> + the most significant byte (i.e., the significand)
```
lucafavatella: I understand this is not meant for representing any arbitrary integers, please explicate constraints that I understand are:
* Non-negative integer (otherwise, is the sign represented consuming one bit of the significand? How?);
* Positive integer (otherwise could zero have multiple representations i.e. with arbitrary exponent?);
* Integer with at most 8 significant bits (or add comment that some get lost).

Also:
* Please add note that the significand can assume value from 1 to 255 (so that it is more clear that the `256*` with the exponent is for freeing one byte for the significand).

apps/aecore/src/pow_sha256.erl
```diff
@@ -0,0 +1,106 @@
...
+-type sci_int() :: integer().
```
lucafavatella: So this is rather `pos_integer()`? (or `non_neg_integer()`?)

apps/aecore/src/pow_sha256.erl
```diff
@@ -0,0 +1,106 @@
...
+%%%=============================================================================
+
+
+-spec generate_with_nonce(Data :: sha256:hashable(), Difficulty :: sci_int(),
+                          Nonce :: integer(), Count :: integer()) -> pow_result().
+generate_with_nonce(_Data, _Difficulty, _Nonce, 0) ->
+    %% Count exhausted: fail
+    {error, generation_count_exhausted};
+generate_with_nonce(Data, Difficulty, Nonce, Count) ->
+    Hash1 = sha256:hash(Data),
```
lucafavatella: This is recomputing hash at various attempts, but for now it can stay as it is.

apps/aecore/src/pow_sha256.erl
```diff
@@ -0,0 +1,106 @@
...
+    Hash2 = sha256:hash(<<Hash1/binary, Difficulty:16, Nonce:?HASH_BITS>>),
+    case binary_to_scientific(Hash2) of
+        HD when HD < Difficulty ->
+            %% Hash satisfies condition: return nonce
+            Nonce;
```
lucafavatella: In principle the nonce returned may not be the same nonce serialized in the hash. May you please add a comment if there is an assumption here?

The nonce is initialized to a number - let us assume the max e.g. `Nonce = ((10^24) - 1)`. At each attempt (see `Count - 1`) the nonce is increased (see `Nonce + 1`). For the sake of hash computation, the nonce used is the 256 least significant bits of it (see `Nonce:?HASH_BITS`; e.g. `<<257:8>>` evaluates to 
`<<1>>`).

apps/aecore/src/pow_sha256.erl
```diff
@@ -0,0 +1,109 @@
...
+generate_with_nonce(_Data, _Difficulty, _Nonce, 0) ->
+    %% Count exhausted: fail
+    {error, generation_count_exhausted};
+generate_with_nonce(Data, Difficulty, Nonce, Count) ->
+    Hash1 = sha256:hash(Data),
+    Hash2 = sha256:hash(<<Hash1/binary, Difficulty:16, Nonce:?HASH_BITS>>),
+    case binary_to_scientific(Hash2) of
+        HD when HD < Difficulty ->
+            %% Hash satisfies condition: return nonce
+            {ok, Nonce};
```
lucafavatella: In principle the nonce returned may not be the same nonce serialized in the hash. May you please add a comment if there is an assumption here?

The nonce is initialized to a number - let us assume the max e.g. `Nonce = ((10^24) - 1)`. At each attempt (see `Count - 1`) the nonce is increased (see `Nonce + 1`). For the sake of hash computation, the nonce used is the 256 least significant bits of it (see `Nonce:?HASH_BITS`; e.g. `<<257:8>>` evaluates to 
`<<1>>`).

akorosmezey:
Resolved the last few suggestions.

sennui:
close/re-open to trigger travis

PR #49, opened 2017-09-01
Solves GH-43 Add coinbase transaction

zp-sd:
* I started to prefix `ae_core` modules with `aec_`, but I think this is also not perfect and doesn't look too good. On the other hand `aecore` or `ae_core` looks a bit longish to me. What do you think guys?
* `aec_tx` behaviour was created, to be used for all transactions, to standarize them for better consumption by other components.
* I have added dummy `aec_accounts` module to store balances, this should be implemented appropriately when we have Merkle trees APIs in place.
* I didn't add a test for `aec_coinbase_tx`, as it is using dummy `aec_accounts` module. I can add a test mocking that `acccounts` module works as a dict or add a test when we have Merkle trees available.


lucafavatella:
This looks good to me.

> I didn't add a test for aec_coinbase_tx, as it is using dummy aec_accounts module. I can add a test mocking that acccounts module works as a dict or add a test when we have Merkle trees available.

@zp-sd I plan to base work on planting Merkle trees on top of this branch - thanks for preparing it.

zp-sd:
As @lucafavatella is already adding Merkle trees support based on this branch, I think we can merge. I will add test later.

apps/aecore/src/aec_accounts.erl
```diff
@@ -15,12 +15,12 @@ get(Pubkey, _AccountsTree) ->
                        balance = 0},
     {ok, Account}.
 
-get_with_proof(Pubkey, _AccountsTree) ->
+get_with_proofs(Pubkey, _AccountsTree) ->
```
lucafavatella: I am not convinced by the plural. The proof is one.

Issue #29, opened 2017-08-28
Implement Merkle trees


wagerlabs:
- Rewrite [this library](https://github.com/exthereum/merkle_patricia_tree) in Erlang.
- Add unit tests.

lucafavatella:
I propose that implementation of Merkle tree in scope for this ticket:
* Aims at ability to store, lookup, compute root hash predictably by performing updates in a *specific* order;
* Does not aim at exhaustive documentation of internal structure;
* Does not aim at persistence.

I propose aiming at integrating https://github.com/KrzysiekJ/gb_merkle_trees

wagerlabs:
Are we not storing Merkle trees in blocks? How are we not to aim at persistence then? 

If this is a stop-gap measure to unblock the rest of development then please suggest the _real_ solution. I understand that you have researched this extensively.

sennui:
I think he meant the persistence inside the the tree. Because we do persist blocks, we can skip internal persistence

wagerlabs:
@sennui @lucafavatella Then this is acceptable as the permanent solution.

wagerlabs:
1. We are going to build the state of the world by processing the blockchain since genesis, just like Ethereum does it.
2. Light nodes will ask full nodes for proof of existence (proof of balance).
3. We will process transactions in the order they appear in the block.

PR #52, opened 2017-09-01
GH-29 Plant Merkle trees

lucafavatella:
Depends on PRs #49 and #50 (both included in this PR).

This PR contains my state on GH-29. It is a minimal integration in order to check for any blockers. All seems good so far - I need only to check one last element (details below - will not be able to do that today).

I open this PR mainly for ensuring no one is blocked ( @zp-sd ). If needed, this PR may be even merged in the current form.

Items left:
* (Must) Verify how proof of existence links to the leaf / serialized value (I put TODO comment in test);
* Move tree-related stuff out of aec_accounts into a module for trees;
* Identify functions to expose in order to verify proof from module API (rather than digging in internals as the current test does).


lucafavatella:
I see tests on CI passed.

apps/aecore/src/aec_accounts.erl
```diff
@@ -0,0 +1,57 @@
...
+         earn/3,
+         root_hash/1]).
+
+-include("trees.hrl").
+
+empty() ->
+    {ok, gb_merkle_trees:empty()}.
+
+get(Pubkey, AccountsTree) ->
+    case gb_merkle_trees:lookup(Pubkey, AccountsTree) of
```
zp-sd: Do we want to hide calls to `gb_merkle_trees` in `trees.erl`, so that we have Erlang APIs to operate on tree library, and we don't have to modify N modules when we'll try to change tree library in the future?
I propose:
* `trees:lookup(Key, Tree)` / `trees:get(Key, Trees)`
* `trees:merkle_proof(Key, Tree)`
(...)
Wdyt?

apps/aecore/test/aec_accounts_tests.erl
```diff
@@ -0,0 +1,46 @@
...
+    ok.
+
+proof_test() ->
+    {ok, T0} = aec_accounts:empty(),
+    A1 = #account{pubkey = <<"k1">>, balance = 10},
+    {ok, T1} = aec_accounts:put(A1, T0),
+    {ok, H1} = aec_accounts:root_hash(T1),
+
+    {ok, {A1, P1}} = aec_accounts:get_with_proof(A1#account.pubkey, T1),
+    ?debugFmt("~p~n", [P1]),
```
zp-sd: don't commit prints

zp-sd:
It would be good to have https://github.com/aeternity/epoch/pull/49 merged first, so that we review only changes from https://github.com/aeternity/epoch/issues/29 here.

apps/aecore/src/aec_accounts.erl
lucafavatella: I agree. I will do this.

apps/aecore/test/aec_accounts_tests.erl
lucafavatella: Ack.

lucafavatella:
I am closing this PR and opening new one - addressing items identified in description and comments.

PR #55, opened 2017-09-05
Gh 10 mine block using sha256

zp-sd:
Depends on https://github.com/aeternity/epoch/pull/49


apps/aecore/include/trees.hrl
```diff
@@ -1,13 +1,12 @@
--include("common.hrl").
-
 -record(trees, {
           accounts}).
 -type(trees() :: #trees{}).
 
 %% Placeholder to define state Merkle trees
 
 -record(account, {
-          pubkey = <<>>  :: pubkey(),
```
lucafavatella: I would suggest keeping type pubkey as it is a type used over several places of the code and clarifies link back to the account.

apps/aecore/src/aec_blocks.erl
```diff
@@ -0,0 +1,51 @@
...
+-spec new(block(), list(signed_tx()), trees()) -> {ok, block()} | {error, term()}.
+new(LastBlock, Txs, Trees0) ->
+    LastBlockHeight = height(LastBlock),
+    LastBlockDifficulty = difficulty(LastBlock),
+    Height = LastBlockHeight + 1,
+    Difficulty = aec_pow_sha256:recalculate_difficulty(LastBlockDifficulty, 1, 1), %% TODO: set properly
+    case aec_tx:apply_signed(Txs, Trees0, Height) of
+        {ok, Trees} ->
+            {ok, #block{height = Height,
+                        prev_hash = aec_sha256:hash(LastBlock),
```
lucafavatella: This is meant to be the hash of the previous serialized block header.

apps/aecore/src/txs/aec_tx.erl
```diff
@@ -0,0 +1,43 @@
...
+%% API
+
+-spec apply_signed(list(signed_tx()), trees(), non_neg_integer()) -> {ok, trees()} | {error, term()}.
+apply_signed([], Trees, _Height) ->
+    {ok, Trees};
+apply_signed([SignedTx | Rest], Trees0, Height) ->
+    case aec_tx_sign:verify(SignedTx) of
+        ok ->
+            Tx = aec_tx_sign:data(SignedTx),
+            case aec_proofs:prove(Tx, Trees0) of
```
lucafavatella: I understand this line can be deleted.

sennui: @lucafavatella this comment is regarding our decision to abandon proofs?

zp-sd: Well, we still need to check whether an account has enough funds, etc? This is what I meant to have done here.

lucafavatella: > @lucafavatella this comment is regarding our decision to abandon proofs?

@sennui  Not really. I wrote the comment as I was not seeing the intent of this line, that @zp clarified in the following comment...

> Well, we still need to check whether an account has enough funds, etc? This is what I meant to have done here.

@zp I see. So if I understand you correctly a more explicit name would be `verify_tx_assumptions` or `verify_tx_preconditions`, instead of `aec_proofs:prove` (the point is verifying the preconditions, and I understand we do that on the state trees - not proofs). So each transaction module, in addition to the APIs `new` and `run`, would have an API `verify_preconditions`.

apps/aecore/test/aec_mining_tests.erl
```diff
@@ -0,0 +1,55 @@
...
+               meck:expect(aec_pow_sha256, recalculate_difficulty, 3, 856),
+               meck:expect(aec_pow_sha256, generate, 3, {error, generation_count_exhausted}),
+               meck:expect(aec_tx, apply_signed, 3, {ok, Trees}),
+
+               ?assertEqual({error, generation_count_exhausted}, ?TEST_MODULE:mine())
+       end},
+      {"Cannot apply signed tx",
+       fun() ->
+               meck:expect(aec_pow_sha256, recalculate_difficulty, 3, 856),
+               meck:expect(aec_pow_sha256, generate, 3, {ok, 721}),
```
sennui: can we use real pow 256?

apps/aecore/src/txs/aec_tx.erl
zp-sd: I understand that this can be done from within a transaction (in `run` function). Let's see later if we need additional tx behaviour function for it.

apps/aecore/test/aec_mining_tests.erl
zp-sd: @sennui now we do.

apps/aecore/include/txs.hrl
```diff
@@ -33,32 +42,37 @@
...
           nonce = 0       :: non_neg_integer(),
           delay = 0       :: non_neg_integer()}).
+-type(channel_grow_tx() :: #channel_grow_tx{}).
 
 -record(channel_team_close_tx, {
           id = 0          :: non_neg_integer(),
           account1 = <<>> :: pubkey(),
           account2 = <<>> :: pubkey(),
-          amount = 0      :: pubkey(),
+          amount = 0      :: binary(),
```
lucafavatella: This type looks incorrect.

PR #73, opened 2017-09-06
Solves Gh-5: Implement the Cuckoo Cycle PoW

akorosmezey:


akorosmezey:
As far as PoW generation is concerned, the call is more or less compatible with the SHA-256. There are two additional parameters to tweak with. (Shall I add a call with the default values of these?)

Verification is, however, much more efficient using the _verify_ function. This differs from its SHA-256 counterpart.

wagerlabs:
It would be nice to have the call with default values.

apps/aecore/c_src/aec_pow_cuckoo.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include "cuckoo_miner.hpp"
+
+
+#define MAXSOLS 1
+// arbitrary length of header hashed into siphash key
+#define HEADERLEN 80
+
+
+node_t* generate(char* header, int nonce, int ntrims, int nthreads) {
+  printf("Looking for %d-cycle on cuckoo%d(\"%s\",%d) with 50%% edges, %d trims, %d threads\n",
```
sennui: to which log file do those prints go?

PR #74, opened 2017-09-06
This fixes #51. rand:uniform instead of crypto:rand_unfiform

### LINKED ISSUE #51: crypto:rand_uniform/2 is deprecated
crypto:rand_uniform/2 is deprecated and produces a warning. I think we should treat warning as errors. Is there some legitimate reason deprecated function has to be used? 

Code warning introduced by @akorosmezey, please comment on that.

Build log:
```
apps/aecore/src/pow_sha256.erl:111: Warning: crypto:rand_uniform/2 is deprecated and will be removed in a future release; use rand:uniform/1
```

cytadela8:
Small PR for #51. We are going to remove `aec_pow_sha256`, but till that this 5MinutesFix(tm) makes our code warning-less.

### DIFF
--- a/apps/aecore/src/aec_pow_sha256.erl
+++ b/apps/aecore/src/aec_pow_sha256.erl
@@ -108,4 +108,4 @@ log2(N) when N > 1 ->
     1 + log2(N div 2).
 
 pick_nonce() ->
-    crypto:rand_uniform(0, ?NONCE_RANGE).
+    rand:uniform(?NONCE_RANGE).


apps/aecore/src/aec_pow_sha256.erl
```diff
@@ -108,4 +108,4 @@ log2(N) when N > 1 ->
     1 + log2(N div 2).
 
 pick_nonce() ->
-    crypto:rand_uniform(0, ?NONCE_RANGE).
+    rand:uniform(?NONCE_RANGE).
```
sennui: thanks for looking at those warnings. Can you set warnings a errors in rebar?

Issue #10, opened 2017-08-28
Mine block using SHA256


wagerlabs:
Depends on #43.

sennui:
make sure if we can extend the size of the nonce (in case we reached max without solving the block). Write stress tests, to examine relation between nonce range and time necessary to mine a block

PR #75, opened 2017-09-06
GH-10 Add time and difficulty calculations to block

zp-sd:
Closes https://github.com/aeternity/epoch/issues/10. Solves https://github.com/aeternity/epoch/issues/10.

### DIFF
--- a/apps/aecore/src/aec_governance.erl
+++ b/apps/aecore/src/aec_governance.erl
@@ -1,0 +1,18 @@
+-module(aec_governance).
+
+%% API
+-export([recalculate_difficulty_frequency/0,
+         expected_block_mine_rate/0]).
+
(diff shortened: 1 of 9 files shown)


apps/aecore/src/aec_governance.erl
```diff
@@ -0,0 +1,18 @@
...
+
+%% API
+-export([recalculate_difficulty_frequency/0,
+         expected_block_mine_rate/0]).
+
+-define(RECALCULATE_DIFFICULTY_FREQUENCY, 10).
+-define(EXPECTED_BLOCK_MINE_RATE, 300). %% 60secs * 5 = 300secs
+
+
+recalculate_difficulty_frequency() ->
```
sennui: hmm, can you add spec to it?

PR #76, opened 2017-09-07
GH-43 Add eunit test for coinbase_tx

zp-sd:
Closes https://github.com/aeternity/epoch/issues/43

### DIFF
--- a/apps/aecore/src/aec_accounts.erl
+++ b/apps/aecore/src/aec_accounts.erl
@@ -41,9 +41,9 @@ put(Account, AccountsTree) ->
     {ok, _NewAccountsTree} =
         aec_trees:put(Account#account.pubkey, serialize(Account), AccountsTree).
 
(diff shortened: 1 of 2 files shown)


apps/aecore/src/aec_accounts.erl
```diff
@@ -41,9 +41,9 @@ put(Account, AccountsTree) ->
     {ok, _NewAccountsTree} =
         aec_trees:put(Account#account.pubkey, serialize(Account), AccountsTree).
 
-earn(#account{balance = Balance0}, Amount, Height) ->
-    {ok, #account{balance = Balance0 + Amount,
-                  height = Height}}.
+earn(#account{balance = Balance0} = Account0, Amount, Height) ->
```
sennui: nice catch

zp-sd: well, this is thanks to eunit test 😉

Issue #77, opened 2017-09-07
Add Cuckoo Cycle solution to the headers


wagerlabs:
Need to add the 42 ints to the header for ease of verification. See #5

PR #82, opened 2017-09-07
GH-5 Generate call with default number of trims and threads added

akorosmezey:
@wagerlabs supported the idea of adding a function with default values for the number of trims and threads, This PR implements it.


akorosmezey:
The two PoW calls are still not similar, I plan to change it further when addressing Issue #66 and #77

sennui:
looks good!

PR #87, opened 2017-09-08
Gh 83 multi node config

velzevur:


rebar3.crashdump
```diff
@@ -0,0 +1,24 @@
+Error: {badmatch,{error,eexist}}
```
sennui: this file should not be here, right?

sennui:
please squash commits

sennui:
@velzevur all right, I tested that on mac os sierra with opt 20 and dirty schedulers enabled  - all works. What was your setup? We should put on wiki stable envs

Issue #69, opened 2017-09-06
Service to hold blocks and headers in memory


wagerlabs:

lucafavatella:
I had a chat with @sennui for clarifying scope of this ticket.

The service is one and is meant to keep:
* Two key-value storages:
  * Block headers - by block header hash;
  * Block - by block header hash. The block does not include the Merkle trees.
* The top - block header identifying the longest chain.

The service is meant to keep one chain.

State needs to be persisted in RocksDB.
Serialization/deserialization can be basic e.g. term_to_binary.

The service is meant to cover intake of new blocks/headers from sync/mining.

The service can be kept simple by enforcing assumption on inputs e.g.:
* Request to store block header returns error if previous block header unknown;
* Request to store block returns error if block header unknown.

The service is meant to cover swap of subset of top blocks for the ones from longer chain.

Open questions:
* Shall snapshots of state Merkle trees at each height be kept? If yes, where e.g. Merkle state service or chain service?

lucafavatella:
For reference: [where previous documentation on node-local block-related structures used to be located](https://github.com/aeternity/testnet/wiki/Blocks#node-local-block-related-structures).

lucafavatella:
For reference: chains in other blockchains:
https://github.com/bitcoin/bitcoin/blob/v0.14.2/src/chain.h
https://github.com/ethereum/go-ethereum/blob/v1.6.7/core/headerchain.go#L43-L47

lucafavatella:
> ... how are we to find out the balance of an account at a given block height? We'll need this when processing blocks, I think.

@sennui I understand we agree the node keeps (we need to agree where) the state Merkle trees only for what the node considers to be the top/current block, so that the node can (1) mine on the top block and/or (2) produce proof of balance for an existing account at the top block.

@wagerlabs @sennui  I understand we do not identify the need to keep the balance of an account at arbitrary block height.

wagerlabs:
How do we verify then that the transactions in a block we just received are valid?

We'll need to go through the transactions and verify that there was sufficient balance for a payment to happen, for example.

That said, if we receive the next block given our current block height then we can just check our state of the world. If we receive the block that's further out then we'll need to ask our peers to fill in the gap in the chain and update our state of the world using the missing blocks. Right?

Do we include the block height in the transaction structure? I think this is where things may become complicated. If the transaction is at the current height then we can validate now. If it's further out then we'll have to wait. What if it's as a lower block height, though?

[...]

PR #90, opened 2017-09-08
GH-69 Block header hash

lucafavatella:
This PR does not close any issues.

Depends on PR #89 (included in this PR).

I prepared this while working on the chain (#69), that uses block header hash. The hash depends on the network serialization, that is therefore planted in this PR too.

### DIFF
--- a/apps/aecore/src/aec_blocks.erl
+++ b/apps/aecore/src/aec_blocks.erl
@@ -1,7 +1,8 @@
 -module(aec_blocks).
 
 %% API
--export([height/1,
+-export([prev_hash/1,
+         height/1,
          trees/1,
          difficulty/1,
          set_nonce/2,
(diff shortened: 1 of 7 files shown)


apps/aecore/src/aec_blocks.erl
```diff
@@ -70,3 +83,23 @@ to_header(#block{height = Height,
             nonce = Nonce,
             time = Time,
             version = Version}.
+
+-spec serialize_for_network(BlockInternalRepresentation) ->
```
sennui: do we need _for_network suffix? Serialization format is unified for all APIs

lucafavatella: We may have several serialization formats. In the previous testnet we used to persist a serialized format that used to include not only items serialized in the network but also internal representation items i.e.:
* Pointer to persisted state Merkle tree at that block;
* Hashes of ancestor block headers in addition to the previous one only - so that to enable jumping e.g. 100 blocks back in the chain without needing to do 100 lookups.

lucafavatella: And cumulative/total difficulty.

wagerlabs: @lucafavatella Please remember to keep it simple! We don't need an internal serialization format.

PR #92, opened 2017-09-08
[RFC] GH-69 In-memory chain service

lucafavatella:
@sennui This is a milestone towards GH-69 and I would appreciate your review.
* The service lacks support for changing tracked chain / fork, i.e. the service can only track a chain that grows one header at a time on the top. See comment on `force_insert_headers` for this missing implementation;
* The service has no persistence.

Depends on #90 (included in this PR). The only commit distinct for this PR is e67f91f.


sennui:
The chain service is fine! Thanks @lucafavatella 

I am merging event though there are TODOs. Its better for syncing with team.

Issue #109, opened 2017-09-11
Strongest chain


wagerlabs:
We need to calculate the longest chain according to the amount of work done, not just the number of blocks. 

See 

- [What does the term longest chain mean?](https://bitcoin.stackexchange.com/questions/5540/what-does-the-term-longest-chain-mean)
- [Strongest vs longest chain](https://bitcoin.stackexchange.com/questions/29742/strongest-vs-longest-chain-and-orphaned-blocks?noredirect=1&lq=1)
- [Picking the longest chain](https://bitcoin.stackexchange.com/questions/936/how-does-a-client-decide-which-is-the-longest-block-chain-if-there-is-a-fork?rq=1)

PR #115, opened 2017-09-11
GH-109 Hook mining to chain service

lucafavatella:
This PR follows from #92 . This is in scope for GH-109.

@sennui @zp-sd This small PR is mostly for ensuring we are on the same page.
@sennui We need to clarify how state trees are retrieved.


lucafavatella:
@sennui I force-pushed moving the `top` API in `aec_chain`. I consider this ready for re-review. CI still needs to complete.

Issue #120, opened 2017-09-12
.gitignore missing exclusions


cytadela8:
`git status` after build gives:
```
On branch master
Your branch is up-to-date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	apps/aecore/c_src/cuckoo/
	apps/aecore/priv/

nothing added to commit but untracked files present (use "git add" to track)
```
We should exclude those paths

PR #119, opened 2017-09-12
GH-77 Cuckoo Cycle evidence added to headers

akorosmezey:
Refactored PoW generation so that SHA-256 and Cuckoo are
interchangeable.

In order to speed up unit tests, the Cuckoo NIF is compiled with
options that result in a 256-times smaller memory footprint.

Closes #77
Closes #20
Fixes #120
Closes #67

PR #121, opened 2017-09-12
GH-120 Added missing exclusions to .gitignore

cytadela8:
This adds missing exclusions to .gitignore. See #120


cytadela8:
Ok, if generic is good is arguable, but #119 supersedes this.

PR #123, opened 2017-09-12
Stricter CI checks

cytadela8:
This PR makes our checks stricter.
It bases on issues fixed in #116 and #74. Maybe we need even more checks?
Note: this won't pass tests with following dialyzer problems:
```
   0: Unknown type gb_trees:gb_tree/2 
   0: Unknown type http_uri:uri/0
```
First error is fixed in #116. Second waits for a fix.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -34,7 +34,8 @@
                    ]},
             {prod, [{relx, [{dev_mode, false},
                             {include_erts, true},
-                            {include_src, false}]}]
+                            {include_src, false}]},
+                    {erl_opts, [warnings_as_errors, true]}]
             }]
 }.
 
--- a/rebar.config
+++ b/rebar.config
@@ -44,3 +44,5 @@
 
 {post_hooks, [{"(linux|darwin)", clean, "make -C apps/aecore/c_src clean"}
              ]}.
+
+{dialyzer, [{warnings, [unknown]}]}.
--- a/apps/aecore/include/peers.hrl
+++ b/apps/aecore/include/peers.hrl
@@ -1,5 +1,7 @@
+-type uri() :: string().
+
 -record(peer, {
-          uri = ""          :: http_uri:uri(),
+          uri = ""          :: uri(),
           last_seen = 0     :: integer()}). % Erlang system time (POSIX time)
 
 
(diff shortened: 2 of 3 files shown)


rebar.config
```diff
@@ -43,3 +44,5 @@
 
 {post_hooks, [{"(linux|darwin)", clean, "make -C apps/aecore/c_src clean"}
              ]}.
+
+{dialyzer, [{warnings, [unknown]}]}.
```
lucafavatella: I suggest not to include this until a fix for `0: Unknown type http_uri:uri/0` is proposed.

rebar.config
```diff
@@ -34,7 +34,8 @@
                    ]},
             {prod, [{relx, [{dev_mode, false},
                             {include_erts, true},
-                            {include_src, false}]}]
+                            {include_src, false}]},
+                    {erl_opts, [warnings_as_errors, true]}]
```
sennui: all profiles should have warnings as errors. The sooner we spot it, the better.

cytadela8: Test profile can't have, because of `-compile(export_all).`.
For dev1 and local I would say it's questionable. User sees the warning always when compiling so user is informed and we make sure it's not missed by including it in "prod" so Travis will fail.

Sometimes when developing one wants to add a function and not use it at first. Or one wants to add a stub function that ignores some of it's parameters. Such check would prevent one from running dialyzer without fixing warnings first or doing any manual tests.

rebar.config
cytadela8: I fixed that

cytadela8:
I rabase the PR so tests should pass now

rebar.config
lucafavatella: Thanks, looks good.

sennui:
@cytadela8 yes, you are correct, because of export_all macros we cannot use it everywhere.

sennui:
@cytadela8 its good to go, but there is small conflict - please resolve

rebar.config
uwiger: For the test profile, at least in our own code, we should use
```erlang
-compile([export_all, nowarn_export_all]).
```

PR #125, opened 2017-09-13
GH84 swagger external endpoint for receiving a block

velzevur:


config/swagger.yaml
```diff
@@ -0,0 +1,155 @@
...
+          description: Put a new block to node
+          required: true
+          schema:
+            $ref: '#/definitions/Block'
+      responses:
+        '200':
+          description: successful operation
+      security: []
+definitions:
+  Coinbasetx:
```
sennui: where do you use it?

cytadela8:
@velzevur Good enough for my tests and my work. Obviously there are some issues, but probably you are aware some parts aren't finished.

I pointed out what I have seen.

apps/aecore/src/aec_headers.erl
```diff
@@ -41,14 +41,35 @@ get_by_hash(_Hash) ->
...
+    Serialized =
+      #{<<"height">> =>  height(H),
+        <<"prev-hash">> => prev_hash(H),
+        <<"root-hash">> => H#header.root_hash,
+        <<"difficulty">> => H#header.difficulty,
+        <<"nonce">> => H#header.nonce,
+        <<"time">> => H#header.time,
+        <<"version">> => H#header.version
+      },
+    {ok, Serialized}.
```
cytadela8: I did a quick search if we can't do it automatically. It should be possible to do some macro, but I haven't found anything and this isn't too messy.

apps/aehttp/.gitignore
```diff
@@ -0,0 +1,16 @@
...
+*.swp
+*.swo
+.erlang.cookie
+ebin
+log
+erl_crash.dump
+.rebar
+logs
+_build
+.idea
```
cytadela8: Why this .gitignore here? Other apps don't have one. Is it necessary?

apps/aehttp/rebar.config
```diff
@@ -0,0 +1,2 @@
+{erl_opts, [debug_info]}.
+{deps, []}.
```
cytadela8: aehttp shouldn't have a separate rebar.config. I think you should add lines to get this application build to the main rebar.config.

config/swagger.yaml
```diff
@@ -0,0 +1,155 @@
...
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters: []
+      responses:
+        '200':
+          description: successful operation
+#          schema:
+#            $ref: '#/definitions/Header'
```
cytadela8: Should this be commented?

cytadela8:
Also: please don't force push any of those commits. I merge those to my branch so I can use your code. Later we (I can do it) will clean-up the merge commits with some rebase for final PR.

apps/aecore/src/aec_chain.erl
```diff
@@ -143,7 +143,7 @@ init(_Args = [GenesisBlock]) ->
     %% Compute initial state of process.
     {ok, TopHeaderHash} = top_header_db_get(NewTopHeaderDb, ?TOP_HEADER),
     {ok, SerializedTopHeader} = headers_db_get(NewHeadersDb, TopHeaderHash),
-    {ok, TopHeader} = aec_headers:deserialize_from_network(SerializedTopHeader),
+    {ok, TopHeader} = aec_headers:deserialize_from_binary(SerializedTopHeader),
```
wagerlabs: Why rename? The network format is not necessarily binary, e.g. can be JSON. @sennui

apps/aehttp/priv/swagger.json
```diff
@@ -0,0 +1,186 @@
...
+            "description" : "successful operation",
+            "schema" : {
+              "$ref" : "#/definitions/Header"
+            }
+          }
+        },
+        "security" : [ ]
+      }
+    },
+    "/block" : {
```
wagerlabs: What is the difference between GET /top and GET /block?

apps/aecore/src/aec_headers.erl
```diff
@@ -26,24 +24,54 @@ time_in_secs(Header) ->
     Time = Header#header.time,
     aeu_time:msecs_to_secs(Time).
 
--spec serialize_for_network(header()) ->
```
wagerlabs: Kindly consult with others before making these API renaming changes. These should not be a unilateral decision! @sennui

apps/aehttp/priv/swagger.json
velzevur: GET /block will have a different structure once the serialization of blocks is ready

apps/aehttp/src/swagger/swagger_api.erl
```diff
@@ -0,0 +1,325 @@
+-module(swagger_api).
```
uwiger: Shouldn't these modules follow the 'ae' prefix naming conventions?
Presumably, `aeh_` would be a suitable prefix.

velzevur: These are swagger generated files: of course we can rename them, but we will have to rename them for every new swagger code generation. If there is a way to config swagger to follow our naming convention - I'm not aware of it.

apps/aecore/src/aec_chain.erl
cytadela8: Network format is handled differently and inside network code. Request makers use deserializa_from_map, serialize_to_map. Swagger requires a map. My code has one function that invokes jsx.

apps/aecore/src/aec_headers.erl
cytadela8: I had to make this change, because network parts wanted to use a function with different spec.

lucafavatella:
I see that this PR introduces some Erlang auto-generated modules. FYI the auto-generated code may be automatically checked by CI - [ref1](https://github.com/aeternity/testnet/blob/b53f5e1788a3c18651aab7acaa309eebd8929279/.travis.yml#L32) [ref2](https://github.com/aeternity/testnet/blob/b53f5e1788a3c18651aab7acaa309eebd8929279/ci#L32-L49).
This CI check can be deferred of course if it makes things slower now.

apps/aehttp/src/swagger/swagger_api.erl
uwiger: It seems as if you can do it with the option `PACKAGE_NAME`
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ErlangServerCodegen.java#L38

Presumably, this can be set in a config file, using the `--config` option, but looking around, this wasn't very well documented, and the swagger_codegen help text (at least my version) didn't even know what erlang was. This may well have been due to some mistake on my part.

sennui:
superseded by https://github.com/aeternity/epoch/pull/139

PR #127, opened 2017-09-15
GH-109 Clarifications in chain service in preparation of support for strongest chain

lucafavatella:
This PR is in scope for GH-109. It is a preparatory PR, meant to be quick to review and safe to merge, so to reduce the noise in the main PR for GH-109 I plan to open later today.

PR #128, opened 2017-09-15
GH-109 Strongest chain

lucafavatella:
Closes GH-109.
Depends on PR #127 (included in this PR). The only commits distinctive of this PR are:
* 10b707b - that keeps track of the amount of work of the chain;
* 48b8b7e - that adds calculation of strongest chain, and function for forcing it in the chain service.

The unit tests show the intended usage of the chain service.
A test with special relevance is `The alternative header chain has more work, but results in sub-optimal choice because of concurrent insertion` that shows how the chain service behaves with concurrent users. The forcing of a sub-optimal choice may be forbidden by putting extra check in the service.

In order to use an API for getting a linear representation of the work in a block header without conflicting too much with other PRs, I added temporary functions `aec_headers:linear_difficulty/1` and `aec_blocks:linear_difficulty/1`. This relates to PR #119.

### DIFF
--- a/apps/aecore/include/blocks.hrl
+++ b/apps/aecore/include/blocks.hrl
@@ -1,5 +1,7 @@
 -include("trees.hrl").
 
+-define(GENESIS_HEIGHT, 0).
+
 -define(BLOCK_HEADER_HASH_BYTES, 32).
 -type(block_header_hash() :: <<_:(?BLOCK_HEADER_HASH_BYTES*8)>>).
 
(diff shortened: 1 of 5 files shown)


lucafavatella:
@sennui I would welcome your feedback, especially on the tests.
@akorosmezey Functions `aec_headers:linear_difficulty/1` and `aec_blocks:linear_difficulty/1` show how I am dealing with difficulty in this PR.

lucafavatella:
I am closing this PR and opening new one based on current master.

PR #130, opened 2017-09-18
fix missing sys.configs

velzevur:


config/dev1/sys.config
```diff
@@ -0,0 +1,16 @@
+[
+  { epoch, []},
+
+  {aecore, [
+      {keys_dir, "/tmp/keys"},
```
sennui: please remove, keys config should go only to default sys.config. local and dev1-3 can have it in their own root dirs of releases

velzevur: local is using the default sys.config; should I make it use one of the dev configs (dev1 for example) in order to use the foldbacking default dir in the release root?

PR #138, opened 2017-09-19
GH-77 eunit fixes

akorosmezey:
As the significand is stored as a signed value, one byte is lost in precision from the highest target value.
Includes a fix in the Cuckoo NIF makefile, in order to avoid unnecessary recompilation of object files


apps/aecore/include/pow.hrl
```diff
@@ -1,2 +1,2 @@
 -define(HIGHEST_TARGET_SCI, 16#2100ffff).
--define(HIGHEST_TARGET_INT, 16#FFFFFF0000000000000000000000000000000000000000000000000000000000).
+-define(HIGHEST_TARGET_INT, 16#FFFF000000000000000000000000000000000000000000000000000000000000).
```
sennui: can you please elaborate on the consequences of changing this difficulty?

akorosmezey: the reason for this change is that we cannot represent the previous value in scientific notation. The significand must be below 0x7fffff because it is a signed value so we lose accuracy. 
The consequence of it is minor (though it is stricter, unlike I said at standup): we drop about 1 - 0xffff00/0xffffff = 1.5x10^-5 of the hashes even with the maximum threshold.

lucafavatella: I see that this adapts the macro to the comment. Thanks for the link @akorosmezey

PR #137, opened 2017-09-19
GH-114 Plant genesis block

lucafavatella:
In scope for GH-114.

Please see https://github.com/aeternity/epoch/issues/114#issuecomment-330635100 for the remaining work on GH-114.

Issue #114, opened 2017-09-11
Genesis block


wagerlabs:

lucafavatella:
I opened #137 that plants the genesis block and unit tests that the chain service can start from that block. Items left for this ticket:

* [ ] Delete proof_hash from wiki https://github.com/aeternity/epoch/wiki/Block-Header
* [ ] Delete balance_proofs from wiki https://github.com/aeternity/epoch/wiki/Block
* [ ] Change “difficulty” to “target” in wiki https://github.com/aeternity/epoch/wiki/Block-Header

* [ ] Open ticket in backlog for fine tuning the initial difficulty/target.

* [ ] Unit test aec_mining that it can mine first block after genesis block. I prepared patch in separate branch, it depends on:
  * PR 136 (signed transaction fix);
  * GH-111 portion re processing of coinbase transaction when account absent from state tree. I need to interact with @akorosmezey 

* [ ] Plant aec_chain service child in supervision tree using this genesis block. BTW it will conflict with https://github.com/aeternity/epoch/pull/125/files#diff-96a78843df857f370a253ffdee5f50a5R34

* [ ] Transactions as Merkle proof. I need to interact with @akorosmezey as this overlaps with GH-111.

lucafavatella:
Only work left here is ensuring PRs #142 #143 #144 are merged (and ensuring CI passes).

PR #141, opened 2017-09-20
[RFC][DO NOT MERGE] GH-114 Unit test mining first block after genesis

lucafavatella:
Depends on PR #137 (included in this PR).

Depends on coinbase transaction processing able to handle trees without account, that I understand is meant to be addressed in PR #140 (not included in this PR - hence unit tests fail).

The only commit distinctive of this PR is 4e78baf .


lucafavatella:
I spoke with @akorosmezey , I will include this test in a following PR when dependencies are merged.

Issue #111, opened 2017-09-11
Update state of the world from the coinbase transaction


wagerlabs:
- Create account if needed.
- Add block reward.
- Update the transaction Merkle tree.

akorosmezey:
@zp-sd implemented it in PRs #10 and #49, the only task is to create account if not yet present.

PR #143, opened 2017-09-20
GH-114 Plant transactions hash for genesis block

lucafavatella:
These are the minimal "transactions hash"-related changes in scope for GH-114. @akorosmezey This does not code the computation of the root hash when there are transactions, in scope for GH-111.

For context, please see wiki page https://github.com/aeternity/epoch/wiki/Block-Header and comment at the top of `aec_block_genesis` module.


lucafavatella:
I am rebasing on master, that includes timeout fixes now.

lucafavatella:
Assigned to me as I investigate CI failures.

Issue #70, opened 2017-09-06
Build the state of the world


wagerlabs:
- Build the state merkle trees and account balances by processing the blockchain from genesis block.

sennui:
there is also yet another case: merkle tree "checkpoints". If we are to revert a block, because we get longer chain block (say we are +1 to prev-world and we get two new blocks or a block with higher difficulty). Now, we would have to recompute a state tree from the genesis. I wonder, if actually the state tree could be only an "engine" to compute proofs - but a state itself of the world itself would be an array of proofs. That way, it would be easy to revert to "state" to previous block state.

wagerlabs:
+1

sennui:
Do not close this ticket until handling state of the world if fixed after accepting longer fork

We need to force state from the common ancestor and then apply new transactions in regular way

Consult with @lucafavatella where is the longer fork handling in chain service

sennui:
Further work that is not in scope of the Sprint 1
 - handle forcing trees from chain service
 - improve validation from height to comparing hashes
 - adjust caching layer (future pesistance layer) in chain service
   to use trees from blocks as checkpoints 

CC: @wagerlabs @lucafavatella

PR #147, opened 2017-09-20
GH-70 extend chain with state of the world service

sennui:
solves GH-70 with exception of handling incoming longer fork. The API to force state from the first common ancestor is ready, tho -- need to correctly call it from aec_chain

### DIFF
--- a/apps/aecore/src/aec_accounts.erl
+++ b/apps/aecore/src/aec_accounts.erl
@@ -13,6 +13,7 @@
 -include("common.hrl").
 -include("trees.hrl").
 
+%% TODO: remove? We should not need such API (init via trees)
 empty() ->
     {ok, _AccountsTree} = aec_trees:new().
 
--- a/apps/aecore/src/aec_state.erl
+++ b/apps/aecore/src/aec_state.erl
@@ -1,0 +1,139 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2017, Aeternity
+%%% @doc
+%%%     Service to hold merkle tree of state
+%%% @end
+%%% Created : 20 Sep 2017
+%%%-------------------------------------------------------------------
+-module(aec_state).
+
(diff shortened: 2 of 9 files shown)


apps/aecore/src/aec_accounts.erl
```diff
@@ -13,6 +13,7 @@
 -include("common.hrl").
 -include("trees.hrl").
 
+%% TODO: remove? We should not need such API (init via trees)
```
lucafavatella: Well spotted - `aec_trees:all_trees_new/1` (called by `aec_block_genesis:genesis_block/0`) shall call this function `aec_accounts:empty/0` rather than `new/0`. (Not crucial of course.)

apps/aecore/src/aec_state.erl
```diff
@@ -0,0 +1,139 @@
...
+    gen_server:call(?SERVER, get_trees, ?DEFAULT_CALL_TIMEOUT).
+
+-spec apply_txs(list(), height()) -> {ok, {height(), trees()}} | {{error, atom()}, trees()}.
+apply_txs(Txs, AtHeight) ->
+    gen_server:call(?SERVER, {apply_txs, {Txs, AtHeight}}, ?DEFAULT_CALL_TIMEOUT).
+
+%% API needed when external fork has more POW and we need to restart tree from common ancestor
+-spec force_trees(trees(), height()) -> {ok, {height(), trees()}}.
+force_trees(Trees, AtHeight) ->
+    gen_server:call(?SERVER, {force_tree, {Trees, AtHeight}}, ?DEFAULT_CALL_TIMEOUT).
```
lucafavatella: The handle_call close below has `force_trees` (with s).

lucafavatella:
@sennui I proposed initial amendments to this PR in auxiliary PR #152 .
I need to complete the review on interactions/dependencies between chain and state services.

PR #152, opened 2017-09-21
[IGNORE][Part of review of PR 147] GH-70 State service amendments 1

lucafavatella:
I started reviewing PR #147, and in the meantime I coded a few fixes for understanding it better. Feel free to merge this - in `gh-70-state-trees-service` branch - and then squash if you want.

Now I will focus more on reviewing interactions between state and chain services.

### DIFF
--- a/apps/aecore/src/aec_state.erl
+++ b/apps/aecore/src/aec_state.erl
@@ -120,10 +120,11 @@ code_change(_OldVsn, State, _Extra) ->
 setup_trees(0, Trees) ->
     Trees;
 setup_trees(N, Trees) ->
+    TreesAtEndOfPreviousBlock = setup_trees(N-1, Trees),
     {ok, Block} = aec_chain:get_block_by_height(N),
     Txs = aec_blocks:txs(Block),
-    {ok, TreesUpdated} = apply_txs_internal(Txs, N, Trees),
-    setup_trees(N-1, TreesUpdated).
+    {ok, TreesUpdated} = apply_txs_internal(Txs, N, TreesAtEndOfPreviousBlock),
+    TreesUpdated.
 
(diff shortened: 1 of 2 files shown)


apps/aecore/src/aec_state.erl
```diff
@@ -123,10 +120,11 @@ code_change(_OldVsn, State, _Extra) ->
 setup_trees(0, Trees) ->
     Trees;
 setup_trees(N, Trees) ->
+    TreesAtEndOfPreviousBlock = setup_trees(N-1, Trees),
     {ok, Block} = aec_chain:get_block_by_height(N),
     Txs = aec_blocks:txs(Block),
-    {ok, TreesUpdated} = apply_txs_internal(Txs, N, Trees),
-    setup_trees(N-1, TreesUpdated).
```
sennui: OK, something here is not correct, but we need recursion to initialize state from block at given height.

This should go from genesis to N, not from N to genesis.

lucafavatella: Recursion is in line 123.

lucafavatella: It is using stack.

sennui: ok, sorry I missed that. I will refactor to stack-less solution. It may be very heavy with long chaing

lucafavatella: ok

lucafavatella: @sennui Feel free to just close this PR and include/squash some/all these changes in PR 147 (even force-pushing is ok).

sennui: merging this, fixing and will squash

PR #153, opened 2017-09-21
[wip - to be squashed] Gh 70 state trees service

sennui:


apps/aecore/src/aec_chain.erl
```diff
@@ -862,9 +868,22 @@ do_write_block(Block, TopHeader, TopBlock, HeadersDb, BlocksDb) ->
...
+                    %% TODO: can we back propagete state when we check for successor in aec_state?
+                    %%       currently updating db may cause data inconsistency or deadlock
+                    BlockMaybeWithTrees = case StateHeight == BlockHeight of
+                                              true -> aec_blocks:set_trees(Block, Trees);
+                                              false -> Block
+                                          end,
                     %% Store block.
                     {ok, SerializedBlock} =
-                        aec_blocks:serialize_for_network(Block),
+                        aec_blocks:serialize_for_network(BlockMaybeWithTrees),
```
lucafavatella: @sennui Just ensuring we are on the same page:
* The chain service stores block without state trees (see type `blocks_db_value()`). Function `aec_blocks:serialize_for_network` removes state trees - if any - from block;
* This `do_write_block` function returns `NewTopBlock`, then cached in state `#top_state.top_block`, that is without state trees.

So this `BlockMaybeWithTrees` may be deleted. (Not needed - just ensuring I understand correctly the intention.)

apps/aecore/src/aec_chain.erl
```diff
@@ -862,9 +868,22 @@ do_write_block(Block, TopHeader, TopBlock, HeadersDb, BlocksDb) ->
                 {ok, SerializedBlock} ->
                     {error, {block_already_stored, SerializedBlock}};
                 {error, not_found} ->
+                    BlockHeight = aec_headers:height(Header),
+                    TopBlockHeight = aec_blocks:height(TopBlock),
+                    {_ok_OnlyForOrderedBlocks, {StateHeight, Trees}} =
```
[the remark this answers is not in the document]
sennui: yes, maybe simply`_` I want to avoid unnecessary complexity

apps/aecore/src/aec_chain.erl
sennui: @lucafavatella 

* serialize for network surly removes trees thats understandable and required

* a block is kept in block record. it has state trees in it (in a record). Don't we cache it?

We can compute trees for a block only if its Prev+1 Block. So this is why I called it Maybe.
It is fine to have block record out of order and without trees.

apps/aecore/src/aec_state.erl
```diff
@@ -0,0 +1,143 @@
...
+handle_call({force_trees, {Trees, AtHeight}}, _From, State) ->
+    {reply, {ok, {AtHeight, Trees}}, State#state{trees = Trees, height = AtHeight}};
+
+handle_call({apply_txs, {Txs, AtHeight}}, _From,
+            #state{trees = Trees, height = CurrentHeigth} = State) ->
+    {Reply, TreesUpdated, HeightUpdated}
+        = case validate_height(CurrentHeigth, AtHeight) of
+              true ->
+                  {ok, TreesUpdated0} = apply_txs_internal(Txs, Trees, AtHeight),
+                  spawn(?MODULE, check_chain_for_successor, [TreesUpdated0, AtHeight]),
```
lucafavatella: May I suggest using a linked spawn, preferably `proc_lib:spawn_link/3`? I see the state service does not trap exits.

----

(I see this is an asynchronous mechanism for making such that, when a new block is inserted in the chain and the state service had the predecessor (by height), the state service attempts to consider any successor blocks that had been previously inserted in the chain without predecessor.)

apps/aecore/src/aec_chain.erl
lucafavatella: > a block is kept in block record. it has state trees in it (in a record). Don't we cache it?

No, currently we do not cache the state trees in the chain service. And, in order to keep things simpler, I suggest not to for the moment.

More in detail, in the chain service's `#state`, the cached top block `#state.top#top_state.top_block` has type `aec_blocks:block_deserialized_from_network()` i.e. does not have state trees.

apps/aecore/src/aec_chain.erl
```diff
@@ -112,15 +113,20 @@ start_link(GenesisBlock) ->
 stop() ->
     gen_server:stop(?SERVER).
 
-%% Returns the highest known block in the chain with its state trees.
+%% Returns the highest known block in the chain
 %%
 %% The highest known block may be lower than the highest block header
 %% in the chain as returned by `top_header/0`.
+%%
+%% It may happen that we skipped a block and top doesn't hold state trees
```
lucafavatella: @sennui as `top_block/0` does not return state trees, this version of `top/0` never returns state trees. I suggest to delete `top_with_state/0`, and replace `top/0` with the implementation of `top_with_state/0`.
(And refine doc of `top_block` and `top` if needed for clarification.)

sennui:
@lucafavatella , I don't think that its necessary to sync restarts of those services.
Indeed, we need to ensure that state is available before chain service

apps/aecore/src/aec_chain.erl
sennui: @lucafavatella OK, let me remove that for now. I added note in a ticket that should evolve to its own ticket - we may consider keeping trees in cached #block

apps/aecore/src/aec_chain.erl
```diff
@@ -123,11 +122,6 @@ stop() ->
 top() ->
     {ok, _Block} = top_block().
```
lucafavatella: This would need to be the old top_with_state, right?

sennui:
replaced by #158

Issue #84, opened 2017-09-07
Swagger endpoint to receive block


wagerlabs:
We need just the sync endpoints. 

Check the old repo for how it was done before.

Ask @cytadela8 and @akorosmezey for the sync requirements. 

Depends on #45.

velzevur:
#210 is merged

Issue #40, opened 2017-08-29
Peers utility - basic code


cytadela8:
Utility to talk with other peers.
I propose there should be a function for every request. This way we will limit typos in request names (to those functions) and we will be able to check the types of data provided to requests.
This depends on having any external API methods.

This issue contains only the basic code. No complex functions. Just functions common to all requests type and probably one request as an example of implementation.

wagerlabs:
This task will be distributed among a bunch of others. We'll add functions to the _peer library_ as the need arises.

cytadela8:
@wagerlabs As you think, but I consider implementing all non repetitive parts of the library a one day job.

cytadela8:
Let's bring this issue back as this is my area of work. With this number I will implement a simple basic utility. Then for all requests types we can create separate issues.

PR #154, opened 2017-09-21
GH-40 peers utility and GH-84 swagger endpoints for receiving blocks

velzevur:


apps/aecore/src/aec_headers.erl
```diff
@@ -38,24 +38,65 @@ time_in_secs(Header) ->
...
+    {ok, jsx:encode(Map)}.
+    %<<(H#header.height):64,
+    %  (H#header.prev_hash):(?BLOCK_HEADER_HASH_BYTES*8),
+    %  (H#header.root_hash):(?BLOCK_HEADER_HASH_BYTES*8),
+    %  (H#header.target):64,
+    %  (H#header.nonce):64,
+    %  (H#header.time):64,
+    %  (H#header.version):16>>.
+
+-spec deserialize_from_binary(header_binary()) -> {ok, header()}.
```
uwiger: Isn't this identical to serialize_to_network/1 above?

apps/aecore/src/aec_mining.erl
```diff
@@ -45,7 +44,7 @@ mine(Attempts) ->
 -spec mine(block(), non_neg_integer()) -> {ok, block()} | {error, term()}.
 mine(Block, Attempts) ->
     Target = aec_blocks:target(Block),
-    {ok, BlockBin} = aec_headers:serialize_for_network(aec_blocks:to_header(Block)),
+    {ok, BlockBin} = aec_headers:serialize_to_binary(aec_blocks:to_header(Block)),
```
uwiger: Should this not use `aec_headers:serialize_to_network/1` instead?

apps/aecore/src/aec_headers.erl
lucafavatella: The implementation is the same. If the intention is different - e.g. network serialization vs. [canonical](https://github.com/aeternity/epoch/wiki/Block-Header) serialization for the sake of reproducible hashing - please add comment making that explicit.

apps/aecore/src/aec_mining.erl
lucafavatella: I expect `serialize_to_binary` is meant to be canonical serialization. Comment expressing this would help.

config/swagger.yaml
```diff
@@ -0,0 +1,144 @@
...
+      nonce:
+        type: integer
+        format: int64
+      time:
+        type: integer
+        format: int64
+      version:
+        type: integer
+        format: int64
+  Header:
```
lucafavatella: Mismatch to https://github.com/aeternity/epoch/wiki/Block-Header

config/swagger.yaml
```diff
@@ -0,0 +1,144 @@
...
+            $ref: '#/definitions/Block'
+      responses:
+        '200':
+          description: successful operation
+      security: []
+definitions:
+  Tree:
+    type: object
+    properties: {}
+  Block:
```
lucafavatella: Mismatch to https://github.com/aeternity/epoch/wiki/Block

PR #155, opened 2017-09-21
Gh 150 mine continuously

### LINKED ISSUE #150: Mine continuously

akorosmezey:
Built on top of PR #146. 

The miner state machine is started by the supervisor. Currently crashes when actual mining is started by the user until state trees are properly added to the block. Its unit test contains a workaround for this, needs to be removed when problem fixed.
Includes minor unit test fixes.
Closes #150

### DIFF
--- a/apps/aecore/src/aec_miner.erl
+++ b/apps/aecore/src/aec_miner.erl
@@ -1,0 +1,199 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2017, Aeternity Anstalt
+%%% @doc
+%%%    A continuous miner
+%%% @end
+%%%-------------------------------------------------------------------
+-module(aec_miner).
+
+-behaviour(gen_statem).
+
+%%------------------------------------------------------------------------------
+%% API
+%%------------------------------------------------------------------------------
+-export([start_link/0,
+         stop/0,
+         start_mining/0,
+         stop_mining/0,
+         get_balance/0]).
+
+%%------------------------------------------------------------------------------
+%% gen_statem callbacks
+%%------------------------------------------------------------------------------
+-export([init/1,
+         idle/3,
+         running/3,
+         terminate/3,
+         code_change/4,
+         callback_mode/0]).
+
+-include("common.hrl").
+-include("blocks.hrl").
+
+
+-define(SERVER, ?MODULE).
+
+-record(state, {}).
+
+%%%===================================================================
+%%% API
+%%%===================================================================
+
+%%------------------------------------------------------------------------------
+%% Start the state machine
+%%------------------------------------------------------------------------------
+-spec start_link() -> {'ok', pid()} | {error | term()}.
+start_link() ->
+    gen_statem:start_link({local, ?SERVER}, ?MODULE, [], []).
+
+%%------------------------------------------------------------------------------
+%% Stop the state machine
+%%------------------------------------------------------------------------------
+-spec  stop() -> 'ok'.
+stop() ->
+    gen_statem:stop(?SERVER).
+
+%%------------------------------------------------------------------------------
+%% Start mining
+%%------------------------------------------------------------------------------
+-spec start_mining() -> 'ok' | {'error' | term()}.
+start_mining() ->
+    gen_statem:call(?SERVER, start).
+
+%%------------------------------------------------------------------------------
+%% Stop mining
+%%------------------------------------------------------------------------------
+-spec stop_mining() -> 'ok' | {'error' | term()}.
+stop_mining() ->
+    gen_statem:call(?SERVER, stop).
+
+%%------------------------------------------------------------------------------
+%% Get the account balance
+%%------------------------------------------------------------------------------
+-spec get_balance() -> integer() | {'error', term()}.
+get_balance() ->
+    {ok, Pubkey} = aec_keys:pubkey(),
+    {ok, LastBlock} = aec_chain:top(),
+    Trees = aec_blocks:trees(LastBlock),
+    AccountsTree = aec_trees:accounts(Trees),
+    case aec_accounts:get(Pubkey, AccountsTree) of
+        {ok, #account{balance = B}} ->
+            B;
+        _ ->
+            {error, account_not_found}
+    end.
+
+%%%===================================================================
+%%% gen_statem callbacks
+%%%===================================================================
+
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% Whenever a gen_statem is started using gen_statem:start/[3,4] or
+%% gen_statem:start_link/[3,4], this function is called by the new
+%% process to initialize.
+%%
+%% @spec init(Args) -> {ok, StateName, State} |
+%%                     {ok, StateName, State, Timeout} |
+%%                     ignore |
+%%                     {stop, StopReason}
+%% @end
+%%--------------------------------------------------------------------
+init(_) ->
+    {ok, _} = aec_chain:start_link(aec_block_genesis:genesis_block()),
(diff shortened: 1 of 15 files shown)


lucafavatella:
@akorosmezey PR 146 was merged to master. Would you mind rebasing on master?

lucafavatella:
@akorosmezey  Please remove from `aec_miner:init`the call to `aec_chain:start_link`. Please express that in the supervision tree. You may find rebasing on PR #156 (that is based on recent master) useful.

lucafavatella:
> @akorosmezey Please remove from `aec_miner:init` the call to `aec_chain:start_link`. Please express that in the supervision tree. You may find rebasing on PR #156 (that is based on recent master) useful.

@akorosmezey Similarly, please remove the call to `aec_chain:stop` from `aec_miner:terminate`. Reply on supervision tree.

apps/aecore/src/aec_miner.erl
```diff
@@ -0,0 +1,199 @@
...
+idle(_Type, _Msg, State) ->
+    {next_state, idle, State}.
+
+running({call, From}, start, State) ->
+    {next_state, running, State, [{reply, From, {error, already_started}}]};
+running({call, From}, stop, State) ->
+    {ok, TopBlock} = aec_chain:top(),
+    Height = aec_blocks:height(TopBlock),
+    Bal = get_balance(),
+    lager:info("Mining finished at height ~p, balance ~p", [Height, Bal]),
```
lucafavatella: Please remove these 4 lines (getting height of top block, getting balance). A client interested in that can do it itself - without requiring the context of the miner.

(If you want to (I would prefer not), you can put those 4 lines in the `aec_miner:stop_mining` function i.e. in the context of the caller process - not the miner.)

lucafavatella:
@akorosmezey Wrt https://github.com/aeternity/epoch/pull/155/files#diff-18f53debbff20a1bf9efdeb33d180e93R149
the interaction of the miner with the chain is correct, but you would need to check the outcome of some interactions with the chain:
```erlang
case aec_chain:insert_header(...) of
  {error, _} ->
    %% TODO Log error.
    %% TODO Do not let miner die, trigger next mining attempt.
    ...;
  ok ->
    case aec_chain:write_block(...) of
      {error, _} ->
        %% TODO Log error.
        %% TODO Do not let miner die, trigger next mining attempt.
        ...;
      ok ->
        %% TODO Trigger next mining attempt (as you had coded already correctly).
        ...
    end
end
```

----

Just to be extremely clear - the core miner algorithm that I see in the code, and that is correct, is:
* Retrieve highest block with state trees in chain (call to `aec_chain:top/0` inside `aec_mining:mine/1`);
* If successful:
  * Insert mined header in chain if successor of highest block header in chain (`aec_chain:insert_header/1`);
  * Store in chain the block of the just inserted header (`aec_chain:write_block/1`).

apps/aecore/src/aec_miner.erl
```diff
@@ -0,0 +1,199 @@
...
+    {next_state, idle, State, [{reply, From, ok}]};
+running(cast, mine, State) ->
+    case aec_mining:mine(10) of
```
lucafavatella: Please add todo comment to make the mining attempts configurable.

apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -96,7 +96,8 @@ verify(Data, Nonce, Evd, Target) when is_list(Evd) ->
 generate_hashed(_Hash, _Nonce, _Target, _Trims, _Threads, 0) ->
     {error, generation_count_exhausted};
 generate_hashed(Hash, Nonce, Target, Trims, Threads, Retries) when Retries > 0 ->
-    case generate_single(Hash, Nonce, Trims, Threads) of
+    Nonce32 = Nonce band 16#7fffffff,
```
lucafavatella: I see this limits the nonce to non-negative signed 32 bits integer.

@akorosmezey questions:
1. When a solution is found, `Nonce` is returned (line 106). That means the nonce returned is not the nonce used for computation. Is this not wrong?
2. I see the overflow to more than 32 bits is caused by the `Nonce + 1` in the recursion at line 108. Why not addressing this overflow in that line? (I assume that the initial nonce returned by `aec_pow:pick_nonce()` already complies to that - you could add guard in generate_hashed function for ensuring that is the case.)
3. Why non-negative *signed* 32 bits integer (`16#7fffffff`) rather than non-negative *unsigned* 32 bits integer (`16#ffffffff`)?

lucafavatella:
@akorosmezey I see that:
* At startup `aec_miner` attempts to call `aec_keys:pubkey/0` once, and if it returns error `aec_miner` sits idle without mining until somebody calls `aec_miner:start_mining`;
* I see that `aec_keys` (not specific to this PR) implements deferred initialization i.e. at end of init it does not guarantee that pubkey is set even if node started with all parameters (cc @sennui );
* I see that `aec_keys` (not specific to this PR) uses the [gen_server timeout `0` in the return value](https://github.com/aeternity/epoch/blob/d8909c4e4c90b1371507ea56afcfb5e6bcea101d/apps/aecore/src/aec_keys.erl#L133) in order to trigger the completion of the deferred initialization (cc @sennui ). As now (this PR) we have `aec_miner` calling `aec_keys` very soon in supervision tree startup, that deferred initialization may never be called (from OTP doc: ["If an integer time-out value is provided, a time-out occurs unless a request or a message is received within Timeout milliseconds."](http://erlang.org/doc/man/gen_server.html#Module:init-1)).

Please perform the following amendments:
* Please ensure that dependency of `aec_miner` on `aec_keys` is expressed in supervision tree order;
* In `aec_miner`, please attempt not only once but a bunch of times `to call aec_keys:pubkey/0` with a sensible delay among retries, before giving up and sitting in `idle`; 
* In `aec_keys`, please amend the deferred initialization not to rely on the OTP timeout parameter `0` in the return value, but rather send raw message `timeout` (or if you want change atom to something more describing e.g. `deferred_init`) (cc @sennui ).

apps/aecore/test/aec_miner_tests.erl
```diff
@@ -0,0 +1,119 @@
...
+               end},
+              {timeout, 60,
+               {"Run miner for a while",
+                fun() ->
+                        ?assertEqual(ok, ?TEST_MODULE:start_mining()),
+                        timer:sleep(20000),
+                        ?assertEqual(ok, ?TEST_MODULE:stop_mining()),
+                        {ok, TopBlock} = aec_chain:top(),
+                        ?debugFmt("reached height ~p~n", [aec_blocks:height(TopBlock)]),
+                        ?assert(aec_blocks:height(TopBlock) > 0),
```
lucafavatella: Please assert at beginning of test that top has height 0.

apps/aecore/test/aec_miner_tests.erl
```diff
@@ -0,0 +1,119 @@
...
+               end},
+              {"Stop in idle",
+               fun() ->
+                       ?assertEqual({error,not_started}, ?TEST_MODULE:stop_mining())
+               end},
+              {timeout, 60,
+               {"Run miner for a while",
+                fun() ->
+                        ?assertEqual(ok, ?TEST_MODULE:start_mining()),
+                        timer:sleep(20000),
```
lucafavatella: I understand this is 20 seconds (!!!). Could you please add TODO comment to speed up this test e.g. by making configurable number of mining attempts and making it event driven rather than time driven?

lucafavatella:
@akorosmezey Review completed, changes requested. I suggest closing this PR and opening new one.

PR #158, opened 2017-09-22
solves GH-70 extend chain with state of the world service

sennui:


apps/aecore/src/aec_chain.erl
```diff
@@ -112,15 +112,17 @@ start_link(GenesisBlock) ->
 stop() ->
     gen_server:stop(?SERVER).
 
-%% Returns the highest known block in the chain with its state trees.
+%% Returns the highest known block in the chain
```
lucafavatella: Reading the function body, I understand this "with its state trees." can be reinstated.

apps/aecore/src/aec_chain.erl
```diff
@@ -112,15 +112,17 @@ start_link(GenesisBlock) ->
...
 %%
 %% The highest known block may be lower than the highest block header
 %% in the chain as returned by `top_header/0`.
+%%
+%% It may happen that we skipped a block and top doesn't hold state trees
```
lucafavatella: Reading the function body, I understand this line can be removed.

apps/aecore/src/aecore_sup.erl
```diff
@@ -31,7 +31,8 @@ init([]) ->
     {ok, {{one_for_one, 5, 10}, [?CHILD(aec_peers, 5000, worker),
                                   %% XXX empty block instead ot genesis block,
                                   %% ticket #84
+                                 ?CHILD(aec_state, 5000, worker),
```
lucafavatella: @sennui I understand `aec_state` must be started after `aec_chain`, because `aec_state:init` calls `aec_chain`. Am I missing something? (See also [previous PR review](https://github.com/aeternity/epoch/pull/153#pullrequestreview-64347073).)

From OTP doc: ["When the supervisor is started, the child processes are started in order from left to right according to this list."](http://erlang.org/doc/man/supervisor.html#supervision_princ)

lucafavatella:
@sennui Review completed, I requested clarification on supervision tree.

apps/aecore/src/aecore_sup.erl
sennui: you are correct, bug.

sennui:
Applied all requests from @lucafavatella. The sup tree is corrected.

PR #159, opened 2017-09-22
GH-148 GH-151 - Swagger api for GET /block (by height or hash)

velzevur:


apps/aecore/src/aec_headers.erl
```diff
@@ -100,3 +107,9 @@ deserialize_from_binary(B) ->
 hash_header(H) ->
     {ok, BinaryH} = serialize_to_binary(H),
     {ok, aec_sha256:hash(BinaryH)}.
+
+replace_empty_pow(#header{pow_evidence = no_value} = Header) ->
+    Header#header{pow_evidence = 0};
```
lucafavatella: I suggest to embed this conditional in `serialize_to_map` if possible, or at least do not export `replace_empty_pow` as it generates a `#header` not complying to the `#header.pow_evidence` type / default value.

lucafavatella: Same for block.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -28,14 +28,45 @@ handle_request('Ping', #{source := Source,
...
+            Resp = aec_blocks:serialize_to_map(
+                     aec_blocks:replace_empty_pow(Block)),
+            {200, [], Resp};
+        {error, {chain_too_short, _}} ->
+            {404, [], #{reason => <<"chain too short">>}}
+    end;
+
+handle_request('GetBlockByHash', Req, _Context) ->
+    Hash = base64:decode(maps:get('hash', Req)),
+    {ok, Header} = aec_chain:get_header_by_hash(Hash),
```
lucafavatella: Of course this can fail too. Specifying a non-existing hash would cause the client to see e.g. 500.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -28,14 +28,45 @@ handle_request('Ping', #{source := Source,
...
+    case aec_chain:get_block_by_hash(HH) of
+        {ok, Block} ->
+            %% swagger generated code expects the Resp to be proplist or map
+            %% and always runs jsx:encode/1 on it - even if it is already
+            %% encoded to a binary; that's why we use
+            %% aec_blocks:serialize_to_map/1 instead of
+            %% aec_blocks:serialize_for_network/1 
+            Resp = aec_blocks:serialize_to_map(
+                     aec_blocks:replace_empty_pow(Block)),
+            lager:info("RESP ~p", [Resp]),
```
velzevur: not to forget to remove this log

velzevur:
Resolved in #164

PR #161, opened 2017-09-22
untested version of 'PostBlock'

uwiger:
The previous version (mistyped 'PutBlock') was obviously wrong.


velzevur:
Resolved in #164

PR #163, opened 2017-09-22
untested version of 'PostBlock'

uwiger:
This implements a 'PostBlock' handler. It has not yet been tested (but dialyzer approves). The previous handler was obviously incorrect.


velzevur:
Resolved in #164

PR #164, opened 2017-09-22
GH-148 GH-151 and untested block post

velzevur:
GH-148 - get block by height
GH-151 - get block by hash

Untested function for posting a block

### DIFF
--- a/apps/aecore/src/aec_blocks.erl
+++ b/apps/aecore/src/aec_blocks.erl
@@ -13,6 +13,7 @@
          to_header/1,
          serialize_for_network/1,
          deserialize_from_network/1,
+         serialize_to_map/1,
          hash_internal_representation/1]).
 
 -ifdef(TEST).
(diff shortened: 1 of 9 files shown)


apps/aecore/src/aec_blocks.erl
```diff
@@ -116,15 +117,22 @@ serialize_for_network(B = #block{}) ->
     {ok, jsx:encode(serialize_to_map(B))}.
 
 serialize_to_map(B = #block{}) ->
+    Pow =
+        case is_integer(B#block.pow_evidence) of
```
lucafavatella: I expect this to fail always as that field does not seem integer. I expect this to be list of 42 integers.

Refs:
* https://github.com/aeternity/epoch/wiki/Block-Header
* https://github.com/aeternity/epoch/blob/1fa7a978ddaa897e2bb8cfdd948f2cc1e0e8aac9/apps/aecore/c_src/aec_pow_cuckoo_nif.cpp#L55
* https://github.com/aeternity/epoch/blob/1fa7a978ddaa897e2bb8cfdd948f2cc1e0e8aac9/apps/aecore/test/aec_pow_cuckoo_tests.erl#L60

config/swagger.yaml
```diff
@@ -123,15 +147,23 @@ definitions:
       version:
         type: integer
         format: int64
-  Header:
+      pow:
+        type: string
+      txs:
+        type: string
```
lucafavatella: This type looks wrong. Reference https://github.com/aeternity/epoch/wiki/Block
I expect list of transactions.

apps/aehttp/priv/swagger.json
```diff
@@ -59,31 +59,64 @@
...
             "schema" : {
               "$ref" : "#/definitions/Block"
             }
           }
         },
         "security" : [ ]
-      },
+      }
+    },
+    "/block-by-hash" : {
```
sennui: This is probably is subjective, but it doesn't look right to have block-by-hash and block-by-height. block/height/XX and block/hash/YY - also not perfect. block/ZZ and header for a type probably the cleanest.

Not a blocker but keep in mind when desigining the whole API

velzevur:
moved to #168

apps/aehttp/priv/swagger.json
velzevur: I didn't like it either; got inspiration from testnet

PR #165, opened 2017-09-22
fix dialyzer issues

sennui:


apps/aecore/src/aec_state.erl
```diff
@@ -124,25 +125,34 @@ code_change(_OldVsn, State, _Extra) ->
...
+                    Error
+            end;
+        Error ->
+            Error
+    end.
+
 
 
--spec apply_txs_internal(list(), height(), trees()) -> {ok, trees()}.
+-spec apply_txs_internal(list(), trees(), non_neg_integer()) -> {ok, trees()} | {error, term()}.
```
lucafavatella: Wrong order of arguments.

apps/aecore/src/aec_state.erl
```diff
@@ -124,25 +125,34 @@ code_change(_OldVsn, State, _Extra) ->
...
-    Txs = aec_blocks:txs(Block),
-    {ok, TreesUpdated} = apply_txs_internal(Txs, CurrentHeight, TreesAtEndOfPreviousBlock),
-    do_setup_trees(CurrentHeight+1, TopHeight, TreesUpdated).
+    case aec_chain:get_block_by_height(CurrentHeight) of
+        {ok, Block} ->
+            Txs = aec_blocks:txs(Block),
+            case apply_txs_internal(Txs, CurrentHeight, TreesAtEndOfPreviousBlock) of
+                {ok, TreesUpdated} ->
+                    do_setup_trees(CurrentHeight+1, TopHeight, TreesUpdated);
+                Error ->
```
lucafavatella: May I suggest avoiding this open ending case? It help maintainability. E.g. `{error, _} = Error ->` is better.

apps/aecore/src/aec_state.erl
```diff
@@ -124,25 +125,34 @@ code_change(_OldVsn, State, _Extra) ->
...
+                    Error
+            end;
+        Error ->
```
velzevur: I'd agree with Luca about explicit pattern matching of errors

PR #167, opened 2017-09-22
GH-150 Mine continuously

akorosmezey:
Includes fixes for state trees


apps/aecore/src/aec_chain.erl
```diff
@@ -120,8 +120,9 @@ stop() ->
 %% TODO: handle out-of-order blocks
 -spec top() -> {ok, block()}.
 top() ->
-    {ok, {Height, _Trees}} = aec_state:get_trees(),
-    {ok, _BlockWithState} = get_block_by_height(Height).
+    {ok, {Height, Trees}} = aec_state:get_trees(),
+    {ok, BlockWithoutState} = get_block_by_height(Height),
```
lucafavatella: Well spotted!

sennui: @lucafavatella I keep forgetting that we cache blocks in network format. We need to prioritize work on that

apps/aecore/src/aec_miner.erl
```diff
@@ -0,0 +1,230 @@
...
+            lager:error("Mining attempt failed with error: ~p.", [Reason]),
+            gen_statem:cast(?SERVER, mine),
+            {next_state, running, State}
+    end;
+running({call, From}, _Msg, State) ->
+    {next_state, running, State, [{reply, From, {error, not_supported}}]};
+running(_Type, _Msg, State) ->
+    {next_state, idle, State}.
+
+waiting_for_keys(cast, check_keys, State) ->
```
sennui: do we want to wait for keys or simply shut down the whole app?

lucafavatella: @sennui  Because of deferred initialization in aec_keys, we cannot just try once and shut down all.

akorosmezey: @Luca pointed out that keys may be loaded later than the app starts. So I prefer waiting in order to avoid sync problems.

PR #168, opened 2017-09-22
Gh 148 gh 151 and untested block post 2

velzevur:
GH-148 - get block by height
GH-151 - get block by hash

Untested function for posting a block

### DIFF
--- a/apps/aecore/src/aec_headers.erl
+++ b/apps/aecore/src/aec_headers.erl
@@ -7,17 +7,20 @@
          target/1,
          difficulty/1,
          time_in_secs/1,
-	 serialize_to_network/1,
-	 deserialize_from_network/1,
+         serialize_to_network/1,
+         deserialize_from_network/1,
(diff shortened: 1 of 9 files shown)


apps/aecore/src/aec_headers.erl
```diff
@@ -100,3 +109,24 @@ deserialize_from_binary(B) ->
 hash_header(H) ->
     {ok, BinaryH} = serialize_to_binary(H),
     {ok, aec_sha256:hash(BinaryH)}.
+
+serialize_pow_evidence(Ev) ->
+    EvList =
+     case is_list(Ev) andalso length(Ev) =:= ?POW_EV_SIZE of
+          true ->
+              Ev;
+          false ->
```
lucafavatella: This is like open ended. It is ok for now, we may need to make it more specific in future.

sennui:
superseded by https://github.com/aeternity/epoch/pull/170

PR #170, opened 2017-09-22
Gh 149 swagger get account balance rebased on 168

sennui:


apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -3,6 +3,8 @@
 -export([handle_request/3]).
 
 -compile({parse_transform, lager_transform}).
+-include("../../aecore/include/common.hrl").
```
sennui: this should work without jumping between dirs. Are deps set up correctly?

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -90,6 +92,18 @@ handle_request('PostBlock', Req, _Context) ->
...
 
+handle_request('GetAccountBalance', Req, _Context) ->
+    Pubkey = maps:get('pub_key', Req),
+    {ok, LastBlock} = aec_chain:top(),
+    Trees = aec_blocks:trees(LastBlock),
+    AccountsTree = aec_trees:accounts(Trees),
+    case aec_accounts:get(Pubkey, AccountsTree) of
+        {ok, #account{balance = B}} ->
+            {200, [], #{balance => B}};
+        _ ->
```
lucafavatella: Open ended case. In future we can make more specific.

apps/aehttp/priv/swagger.json
```diff
@@ -151,6 +151,37 @@
...
+        "tags" : [ "external" ],
+        "description" : "Get account's balance",
+        "operationId" : "GetAccountBalance",
+        "consumes" : [ "application/json" ],
+        "produces" : [ "application/json" ],
+        "parameters" : [ {
+          "name" : "pub_key",
+          "in" : "query",
+          "description" : "Public key to extract balance for",
+          "required" : true,
```
lucafavatella: This means that client needs to know pubkey before enquiring. How does client know key?
Alternative would be to make this parameter optional, and if not specified simply call `aec_keys:pubkey()`

sennui: @lucafavatella I am merging this, @velzevur please add /getbalance going to current account

PR #194, opened 2017-09-25
GH-77 Use millisecs in target threshold update

akorosmezey:


apps/aecore/src/aec_mining.erl
```diff
@@ -107,7 +107,8 @@ should_recalculate_difficulty(Height) ->
 calculate_difficulty(NewBlock, BlocksToCheckCount) ->
     CurrentTarget = NewBlock#block.target,
     CurrentRate = get_current_rate(NewBlock, BlocksToCheckCount),
```
lucafavatella: @akorosmezey  Please add TODO comment on current "rate" being positive.

This line identifies relevant two headers for computation, extracts time, computes time interval between oldest block and newest block. Depending on constraints on fields of header (depending also on other ticket re clock skew) and validation of such field, oldest block may have timestamp equal or newer (!!!) than newest block.

lucafavatella: @akorosmezey  Thanks for addressing this in `mining_rate_between_blocks`

lucafavatella:
I see 2 reviewers approved.
Of the two CI builds:
* One build ran all *including the Ubuntu packaging and one OSX packaging*. Only *one* OSX packaging left (the OSX packaging jobs are two - one for recent OSX and one for old OSX);
* The other build ran all except the two OSX packaging jobs.

Merging then.

Issue #188, opened 2017-09-25
Optimize Travis CI jobs


wagerlabs:

lucafavatella:
Optimizing OTP installation on Ubuntu CI jobs is not worth the effort, as it usually takes less than 10 seconds.

PR #195, opened 2017-09-25
GH-188 Make aec_mining unit tests more stable

lucafavatella:
As part of GH-188, I am first identifying and fixing unit tests failing non-deterministically quite often, as they are cause of loss of time.


lucafavatella:
I see CI passing. I see 2 reviewers approved. Merging.

Issue #45, opened 2017-08-31
Serialize Merkle trees, blocks and headers for sending over the network


wagerlabs:
See #29 for the Merkle tree implementation.

- The serialization format should be easily parsed by JavaScript clients while being reasonably compact.

wagerlabs:
Let's use JSON for now.

cytadela8:
I think we can add a very nice JSON implementation with swagger. With help of swagger-codegen we might even be able to just pass our record to auto generated code as far as I remember swagger-codegen generated code mechanics.

cytadela8:
Then in future we can create special endpoints for erlang clients with space efficient serialization.

wagerlabs:
+1

velzevur:
Transactions:

- [ ] Coinbase
- [ ]  Send

PR #216, opened 2017-09-27
GH-45 Coinbase transaction serialization

velzevur:


config/swagger.yaml
```diff
@@ -236,6 +236,27 @@ definitions:
...
+        type: string
+      data:
+        type: object
+        schema:
+          oneOf:
+            - $ref: '#/definitions/CoinbaseTx'
+      signatures:
+        type: array
+        items:
+          type: string
```
lucafavatella: I believe this type will need to be detailed further in order to be useful. I understand it shall contain both the signing pubkey and signature, in order for the client to use those to verify the signature on the transaction.

I see in backlog card "Verify signature of the payer on the transaction".

@velzevur If you understand enabling the HTTP client to verify the signature of the transaction is outside scope of GH-45, could you please add card in backlog?

py/tests/test_api_ext.py
```diff
@@ -23,12 +23,21 @@
...
     d = datetime.datetime.utcnow()
     epoch = datetime.datetime(1970,1,1)
     return int((d - epoch).total_seconds())
 
+def signed_coinbase_tx(height):
+    account = "BAAggMEhrC3ODBqlYeQ6dk00F87AKMkV6kkyhgfJ/luOzGUC+4APxFkVgAYPai3TjSyLRObv0GeDACg1ZxwnfHY="
+    coinbase = CoinbaseTx(pubkey = account, nonce = height - 1)
+    return SignedTx(data = coinbase, type = "coinbase",
+            signatures =
+            ["MEUCIQDRusihjYzeL1VgYhruK3v9M9ktwmr5TGmaor6kTPpA5gIgehFypmZyXn7VZNxHhiq/FN4Us7vsVM2frKZeQpLsxPE="])
```
lucafavatella: How did you compute this? If it comes from computation (i.e. it is not random string), you may replace this hardcoded string with the related python computation.

config/swagger.yaml
velzevur: This type follows the current implementation of signatures:
https://github.com/aeternity/epoch/blob/master/apps/aecore/include/txs.hrl#L3
I agree with you that in order to validate a signature - a public key is needed so these should go as a pair. When the structure of #signed_tx.signatures changes - the corresponding section of .yaml should be changed, too.

py/tests/test_api_ext.py
velzevur: Actually this is the product of
https://github.com/aeternity/epoch/blob/master/apps/aecore/src/aec_keys.erl#L152
using a private key, corresponding to the pub_key in the same function, signing the coinbase transaction. It should be a valid signature.

lucafavatella:
I see a [Travis CI job failed](https://travis-ci.org/aeternity/epoch/jobs/280437683#L1254):
```
  1) aec_miner_tests:miner_test_/0: Remove keys while miner runs

     Failure/Error: ?assertEqual({waiting_for_keys,{state}}, sys : get_state ( aec_miner ))

       expected: {waiting_for_keys,{state}}

            got: {running,{state}}

     %% /home/travis/build/aeternity/epoch/_build/test/lib/aecore/test/aec_miner_tests.erl:106:in `aec_miner_tests:-miner_test_/0-fun-3-/0`
```

This is [known fragile test whose fix is deferred](https://github.com/aeternity/epoch/commit/b58f8689d9219e9b7ef3dfd414f071be2f1a1a85), hence I restart job.

apps/aecore/src/txs/aec_tx.erl
```diff
@@ -68,3 +74,26 @@ process_single(#coinbase_tx{} = Tx, Trees, Height) ->
...
 process_single(_Other, _Trees_, _Height) ->
     {error, not_implemented}.
+
+serialize_single(#signed_tx{data = Tx, signatures = Sigs}) ->
+    {Data, Type} =
+      case Tx of
+          %% So far we support only coinbase txs
+          #coinbase_tx{} ->
+              {aec_coinbase_tx:serialize(Tx), <<"coinbase">>}
+      end,
```
sennui: please use this module to hold serialization scheme aec_tx_sign

apps/aecore/src/aec_tx_sign.erl
```diff
@@ -18,15 +20,35 @@ verify(failed_tx) ->
...
-        #signed_tx{signatures = [_]} = SignedTx ->
-            SignedTx
-    end.
+-spec serialize(#signed_tx{}) -> map().
+serialize(#signed_tx{data = Tx, signatures = Sigs})  when is_tuple(Tx) ->
+    {Data, Type} =
+      case Tx of
+          #coinbase_tx{} ->
+              {aec_coinbase_tx:serialize(Tx), <<"coinbase">>}
+      end,
```
sennui: @velzevur please refactor that so we don't populate the code with cases of transaction

Lets let only tx behaviour have this case

PR #219, opened 2017-09-28
GH-218 Clarify name of difficulty-related aec_chain functions

lucafavatella:
Closes GH-218.

I did not add function to get total difficulty at height as I think that clients needing the total difficulty want to a refer to a specific chain, and requesting by height does not guarantee that (after a stronger chain is forced, header at same height as before may be a distinct header).


lucafavatella:
I see CI passed. I see one reviewer approved in current form.

Issue #39, opened 2017-08-29
Implement block syncing


cytadela8:
Depends on 

- [ ] #11 
- [ ] #12 
- [ ] #13

wagerlabs:
Closing this ticket and replacing it by #223.

PR #239, opened 2017-09-29
Initial ping handshake and continuous block sync (issue #39)

uwiger:
- ping must be manually initiated (`aec_sync:connect_peer(Uri)`)
- peers share other peers during ping, leading to autom. discovery
- initial handshake protocol inspired by Ethereum's
- gossip protocol used to broadcast new blocks to all peers

### DIFF
--- a/apps/aecore/src/aec_chain.erl
+++ b/apps/aecore/src/aec_chain.erl
@@ -38,6 +38,9 @@
          has_more_work/1,
          force_insert_headers/1]).
 
+-export([export_chain/1,
+         import_chain/1]).
+
 %% gen_server callbacks
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
@@ -281,6 +284,36 @@ force_insert_headers(HeaderChain) ->
     gen_server:call(?SERVER, {force_insert_headers, HeaderChain},
                     ?DEFAULT_CALL_TIMEOUT).
 
+export_chain(OutFile) ->
+    Headers = all_headers(),
+    Blocks = all_blocks(Headers),
+    file:write_file(
+      OutFile, term_to_binary({{0,1}, Headers, Blocks}, [compressed])).
+
+import_chain(File) ->
+    case file:read_file(File) of
+        {ok, Bin} ->
+            try binary_to_term(Bin) of
+                {{0,1}, Headers, Blocks} ->
+                    io:fwrite("Headers = ~p~n"
+                              "Blocks = ~p~n",
+                              [Headers, Blocks]),
+                    lists:foreach(
+                      fun(H) ->
+                              ok = insert_header(H)
+                      end, Headers),
+                    lists:foreach(
+                      fun(B) ->
+                              ok = write_block(B)
+                      end, Blocks)
+            catch
+                error:Reason ->
+                    {error, Reason}
+            end;
+        Error ->
+            Error
+    end.
+
 %%%===================================================================
 %%% gen_server callbacks
 %%%===================================================================
@@ -426,8 +459,8 @@ handle_call({force_insert_headers, HeaderChain = [_|_]}, _From, State) ->
             {reply, Reply, NewState}
     end;
 handle_call(Request, From, State) ->
-    lager:warning("Ignoring unknown call request from ~p: ~p", [From, Request]),
-    {noreply, State}.
+    lager:warning("Unknown call request from ~p: ~p", [From, Request]),
+    {reply, {error, unknown_request}, State}.
 
 handle_cast(Msg, State) ->
     lager:warning("Ignoring unknown cast message: ~p", [Msg]),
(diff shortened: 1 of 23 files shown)


rebar.config
```diff
@@ -5,7 +5,9 @@
         {rocksdb, "0.12.0"},
         {cowboy, "1.0.4", {git, "https://github.com/ninenines/cowboy.git", {tag, "1.0.4"}}},
         {jsx, {git, "https://github.com/talentdeficit/jsx.git", {tag, "2.8.0"}}},
-        {jesse, {git, "https://github.com/for-GET/jesse.git", {tag, "1.5.0-rc2"}}}
+        {jesse, {git, "https://github.com/for-GET/jesse.git", {tag, "1.5.0-rc2"}}},
+        {gproc, "0.6.1"},
+        {jobs, "0.7.0"}
```
lucafavatella: I am not sure why jobs is not in the rebar.lock.

apps/aecore/src/aec_chain.erl
```diff
@@ -281,6 +284,36 @@ force_insert_headers(HeaderChain) ->
...
+    case file:read_file(File) of
+        {ok, Bin} ->
+            try binary_to_term(Bin) of
+                {{0,1}, Headers, Blocks} ->
+                    io:fwrite("Headers = ~p~n"
+                              "Blocks = ~p~n",
+                              [Headers, Blocks]),
+                    lists:foreach(
+                      fun(H) ->
+                              insert_header(H)
```
lucafavatella: May you please elaborate on why no match to `ok`? I see the the export make such that `Headers` is from genesis to top.

apps/aecore/src/aec_chain.erl
```diff
@@ -281,6 +284,36 @@ force_insert_headers(HeaderChain) ->
...
+                      end, Headers),
+                    lists:foreach(
+                      fun(B) ->
+                              write_block(B)
```
lucafavatella: May you please elaborate on why no match to `ok`?

rebar.config
uwiger: Not sure either

apps/aecore/src/aec_chain.erl
uwiger: The import caused a crash trying to deserialize transactions, so pending a fix of that, it doesn't work.

apps/aecore/src/aec_chain.erl
uwiger: No reason. It should match on ok.

uwiger:
Yes, there need to be tests, but for them to be practical, a fast way of creating blockchains is needed. Running two nodes on the same machine, I once waited 25 minutes for a single block to be mined.

uwiger:
The failure in the test run (job 659.1) seems unrelated to this PR:

```
  1) aec_mining_tests:mine_block_from_genesis_test_/0: Find first block after genesis (PoW module aec_pow_cuckoo)
     Failure/Error: {error,
                        {badmatch,{error,key_not_found}},
                        [{aec_mining_tests,
                             '-mine_block_from_genesis_test_/0-fun-4-',0,
                             [{file,
                                  "/home/travis/build/aeternity/epoch/_build/test/lib/aecore/test/aec_mining_tests.erl"},
                              {line,218}]}]}
     Output: 
```

config/swagger.yaml
```diff
@@ -15,31 +15,26 @@ schemes:
...
+          name: body
+          description: Ping a peer
+          required: true
+          schema:
+            $ref: '#/definitions/Ping'
       responses:
         '200':
           description: Sample response
           schema:
             $ref: '#/definitions/Ping'
```
velzevur: Shouldn't we have a 404 error response, too?

uwiger: Indeed

apps/aecore/src/aec_sync.erl
```diff
@@ -29,25 +45,143 @@
...
 
 start_link() ->
-    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+    ensure_tab(?RECV_POOL, [set, public, named_table]),
+    case gen_server:start_link({local, ?MODULE}, ?MODULE, [], []) of
+        {ok, Pid} ->
+            ets:give_away(?RECV_POOL, Pid, []),
+            {ok, Pid};
+        Other ->
+            ets:delete(?RECV_POOL),
```
velzevur: As discussed - this could pose a danger

apps/aecore/src/aec_sync.erl
```diff
@@ -29,25 +45,142 @@
...
+    #{<<"genesis_hash">> => base64:encode(GHash),
+      <<"best_hash">>    => base64:encode(TopHash),
+      <<"difficulty">>   => Difficulty,
+      <<"source">>       => Source,
+      <<"share">>        => 32,  % TODO: make this configurable
+      <<"peers">>        => []}.
+
+%% Compare local and remote 'Ping' objects.
+%% 1. Compare genesis blocks. If they differ, return error
+%% 2. Compare best_hash. If they are the same, we're in sync
```
sennui: best hash - is it ethereum thing to name like that top hash?

uwiger: That was borrowed from ethereum, yes.

Issue #16, opened 2017-08-28
Implement the mempool


wagerlabs:

cytadela8:
Problems that emerged when implementing mempool with sorted transactions:
- No way to check if transaction is fully valid exposes us to getting invalid transactions and not realizing that.
  - How to penalize peer for giving bad transactions?
  - We need a process that will remove invalid transactions. Maybe we should add block creation to tx_pool and include such code in block_creation?
- Need know fee for transaction solved by adding another wrapper like signed_tx

wagerlabs:
Artur, please make sure that comments on a ticket can be understood without digging around
 for information. Include links to the code that you are referring to, line numbers included. 

Please elaborate on the following

- Why is there no way to check if a transaction is fully valid? Please explain.
- Why do we need to penalize the peer? 
- What does _add block creation to tx_pool_ mean? 
- Why can't we ensure that invalid transactions do not make it into the mempool?
- What do you mean by _Need know fee for transaction solved by adding another wrapper like signed_tx_.

@sennui

cytadela8:
> Why is there no way to check if a transaction is fully valid? Please explain.

> Why can't we ensure that invalid transactions do not make it into the mempool?

Transactions in one block may depend on each other. We sort transactions based on fees, previously we added transactions to the end of list. Previously we could just try to apply new transaction on the end state of trees after all the others. Now It's not possible, because new transaction should be applied between other transactions and other transaction may conflict with it.

> Why do we need to penalize the peer?

It's a good idea to stop listening to someone that provides us with invalid transactions and makes us waste resources.

> What does add block creation to tx_pool mean?

Maybe code for generating a block to mine should be included in tx_pool module. I'm thinking about that, because that process includes full transaction validation and we need to add some process to remove invalid transactions from the pool.

> What do you mean by Need know fee for transaction solved by adding another wrapper like signed_tx.

That's an implementation detail. I added a wrapper like signed_tx to solve problem with accessing fee from mempool. I don't like my solution.

sennui:
Lets walk it step by step:

**Invalid transactions caused by fee-based re-ordering:**
We should not worry about it. Fee based ordering is in favour of our customers and this is also powerful tool to prioritize transactions. They will figure out a way to order them so they are valid

**penalizing the peer:**
sure we can keep some counter - Doesn't look like MVP!

general:
Mind that mempool is kind of mining "manager". It interfaces with external world. It has several duties:
* take new transaction and dispatch them to mining
[...]

Issue #236, opened 2017-09-29
Do not remove transactions from the mempool while mining


wagerlabs:
Transactions will be removed from the mempool when a new block is added to the top of the chain by #232. This may happen when a block is mined and sealed or when a new block becomes the top of the chain. Regardless, transactions should not be removed from the mempool while the block is being mined!

PR #240, opened 2017-09-29
[WIP] GH-16 Implement mempool

lucafavatella:


apps/aecore/test/aec_tx_pool_tests.erl
```diff
@@ -0,0 +1,155 @@
...
+               ?assertEqual(ok, aec_tx_pool:push(STx4)),
+               ?assertMatch({ok, [_,_]}, aec_tx_pool:peek(10)),
+
+               %% A block inserted in chain.
+               ?assertEqual(ok, aec_tx_pool:delete(STx1)),
+               ?assertEqual(ok, aec_tx_pool:delete(STx2)),
+               ?assertEqual({ok, [STx4]}, aec_tx_pool:peek(10)),
+
+               %% A tx already in chain now received from peers.
+               ?assertEqual(ok, aec_tx_pool:push(STx2)),
```
zp-sd: If `STx2` is already in a chain and we re-add it to the pull, such tx will be removed from a pool in mining attempt? Mining shall recognize it as duplicate?

lucafavatella: > If STx2 is already in a chain and we re-add it to the pool, such tx will be removed from the pool in mining attempt?

Txs are removed from pool when added in chain - see GH-236. Mining itself is not meant to remove txs from pool. After successful mining, if mined block added to chain then txs in block shall be then removed.

> Mining shall recognize it as duplicate then?

Yes, thanks to nonce. Mining attempts to apply txs. Each tx is meant to  contain initiating account and its nonce, and each tx is meant to ensure nonce in account is strictly less than nonce in txs. So duplicated tx is detected by mining, and mining shall delete such tx from pool.

Issue #248, opened 2017-10-02
create a multi-node sync test suite


uwiger:
Node syncing currently has no automated tests. I suggest handling this as a separate ticket rather than amending PR #239

PR #252, opened 2017-10-04
Common Test suite for 2-node and 3-node sync

uwiger:
The test suite patches a few modules in order to speed up mining,
then runs two groups of tests: 2-node and 3-node, testing node sync
at startup, and continuous sync during mining.

Re. issue #248

### DIFF
--- a/apps/aecore/test/aecore_sync_SUITE.erl
+++ b/apps/aecore/test/aecore_sync_SUITE.erl
@@ -1,0 +1,460 @@
+-module(aecore_sync_SUITE).
+
+%% common_test exports
+-export(
+   [
+    all/0, groups/0, suite/0,
+    init_per_suite/1, end_per_suite/1,
+    init_per_group/2, end_per_group/2,
+    init_per_testcase/2, end_per_testcase/2
+   ]).
+
+%% test case exports
+-export(
+   [
+    start_first_node/1,
+    start_second_node/1,
+    start_third_node/1,
+    mine_on_first/1,
+    mine_on_second/1,
+    mine_on_third/1,
+    mine_again_on_first/1,
+    restart_first/1,
+    restart_second/1,
+    restart_third/1
+   ]).
+
+-export([proxy/0]).
+
+-include_lib("common_test/include/ct.hrl").
+
+all() ->
+    [
+     {group, two_nodes},
+     {group, three_nodes}
+    ].
+
+groups() ->
+    [
+     {two_nodes, [],
+      [start_first_node,
+       mine_on_first,
+       start_second_node,
+       mine_again_on_first,
+       mine_on_second,
+       restart_second,
+       restart_first]},
+     {three_nodes, [],
+      [start_first_node,
+       mine_on_first,
+       start_second_node,
+       start_third_node,
+       mine_again_on_first,
+       mine_on_second,
+       mine_on_third,
+       restart_third]}
+    ].
+
+suite() ->
+    [].
+
+init_per_suite(Config) ->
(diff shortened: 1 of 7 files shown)


rebar.config
```diff
@@ -41,7 +41,10 @@
             {test, [{relx, [{dev_mode, true},
                             {include_erts, false},
                             {include_src, true}]},
-                    {deps, [{meck, ".*", {git, "git://github.com/eproxus/meck.git", {tag, "0.8.6"}}}
+                    {dist_node, [{setcookie, 'epoch_cookie'},
+                                 {sname, epoch_ct}]},
+                    {deps, [{meck, ".*", {git, "git://github.com/eproxus/meck.git", {tag, "0.8.6"}}},
+                            {erlexec, ".*", {git, "git://github.com/saleyn/erlexec.git", "HEAD"}}
```
lucafavatella: In order to prevent unexpected breakages, may I suggest `{ref, "97a5188"}`.

apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -0,0 +1,460 @@
...
+miner_xform_1({tuple,L,[{atom,_,next_state},{atom,_,running},{var,_,S}]}) ->
+    {tuple,L,[{atom,L,next_state},{atom,L,idle},{var,L,S}]};
+miner_xform_1(_) ->
+    continue.
+
+pow_cuckoo_xform(Forms, _Options) ->
+    parse_trans:plain_transform(fun pow_cuckoo_xform/1, Forms).
+
+pow_cuckoo_xform({string,L,"aec_pow_cuckoo28_nif"}) ->
+    {string,L,"aec_pow_cuckoo20_nif"};
```
lucafavatella: @uwiger  @akorosmezey  @sennui *If* package is meant to include both nif versions, shall this be controlled by configuration rather than parse transform?

uwiger: @lucafavatella currently, both versions _do_ seem to be included, but I agree.
In many cases, it would also be very convenient to be able to select the faster algorithm for interactive testing.

rebar.config
uwiger: Agreed.

uwiger: I've pushed a change for this

apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -0,0 +1,460 @@
...
+
+all() ->
+    [
+     {group, two_nodes},
+     {group, three_nodes}
+    ].
+
+groups() ->
+    [
+     {two_nodes, [],
```
lucafavatella: The tests in the group appear to require running in order. In order to reduce log content in case of a failing test case, it looks like in a future iteration we may specify [`sequence`](http://erlang.org/doc/apps/common_test/dependencies_chapter.html#sequences) property.

apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -0,0 +1,460 @@
...
+     {three_nodes, [],
+      [start_first_node,
+       mine_on_first,
+       start_second_node,
+       start_third_node,
+       mine_again_on_first,
+       mine_on_second,
+       mine_on_third,
+       restart_third]}
+    ].
```
lucafavatella: An interesting additional test for future iteration would be:
* Start node 1, mine one block, stop node 1;
* Start node 2, mine two blocks;
* Start node 1;
* Check node 1 syncs to node 2.

sennui:
test jobs green - merging

Issue #192, opened 2017-09-25
Create a comprehensive release test setup


wagerlabs:
We need to 

- catch issues like `badarith` which we missed twice in a row!
- not slow down the day-to-day regular Travis job stream, i.e. hook this up to the release build only.

lucafavatella:
I understand the manual test approach is:
* Start release `bin/epoch start`;
* Ensure release is started i.e. ensure `bin/epoch ping` succeeds;
* Query top `curl http://127.0.0.1:3013/v1/top` (the port number is the default in the `sys.config`). Observe increasing height;
* Query balance `curl http://127.0.0.1:3013/v1/account/balance`. Observe increasing balance.

lucafavatella:
> not slow down the day-to-day regular Travis job stream, i.e. hook this up to the release build only.

The job that is responsible for building the package and uploading the package to GitHub is called [`package`](https://github.com/aeternity/epoch/blob/a8f6eb2ad30a6adfd632c4a8f67a1d3e982c1469/.travis.yml), and runs (as of master) on 3 types on machine: Ubuntu, recent OSX/Xcode and old OSX/Xcode.

The 3 jobs (i.e. `package` on the 3 types of machine - hence 3 jobs) run alongside all other jobs (i.e. branch push, tag push, pull request open, pull request merge) by building the package. The jobs upload the package [only when the trigger is a tag](https://github.com/aeternity/epoch/blob/a8f6eb2ad30a6adfd632c4a8f67a1d3e982c1469/.travis.yml#L28).

velzevur:
I think we can easily add to the manual approach a download of the whole chain. This would give us even more certainty that everything is fine.
We could easily do it with swagger generated python client

lucafavatella:
> catch issues like badarith which we missed twice in a row!

I did not manage to reproduce `badarith` on a8f6eb2:
* For Ubuntu, I used package built by Travis and uploaded on GitHub, I tested it on VirtualBox (`vagrant init ubuntu/trusty64; vagrant up; vagrant ssh` then see content of `/vagrant`). No `badarith` error. Mining first block took a few minutes, mining second block is taking tens of minutes;
* For OSX, I did not use package. I cleaned my working copy, then I issued `make prod-build` and `make prod-start`. No `badarith` error. Mining first block took a few minutes, mining second block took tens of minutes, mining third block took a few minutes.

wagerlabs:
https://github.com/aeternity/epoch/issues/192

PR #256, opened 2017-10-05
GH-192 Python release integration test

velzevur:
All builds with a tag set to anything will trigger a Tavis JOB (one per OS - linux and osx) to validate the release is correct. The script does the following

- generate a release like tarball (only change being fast mining is enabled)
- extract the tarball
- copy the extracted dir to 2 more directories (so we have node1, node2, node3)
- setup the nodes
- start the nodes
- get each node's top every second untill there are 50 blocks mined
- stop the nodes and cleanup

If a node crashes - all node logs are tailed so we can easily debug it.

It is a huge commit and it is mostly due to that I had to add some Swagger generated python code (so we don't have to install swager-codegen in the JOB). The generated code is located in ` py/tests/swagger_client/ `
You can focus on changes in `.travis.yml`, `Makefile`, `ci`, `py/tests/release.py` and `rebar.config`


lucafavatella:
> TODO: check what's the real value of TRAVIS_TAG when building a release

@velzevur Wrt checking this for the case the build is a tag: I recommend you try this by forking this repo in your github account, enabling Travis CI there, pushing this branch there too, taking there (in your fork) a tag, observing result.

py/tests/release.py
```diff
@@ -0,0 +1,78 @@
...
+from swagger_client.api_client import ApiClient
+
+def build_dir():
+    return "build_dir"
+
+def executable():
+    return os.path.join(build_dir(), "bin", "epoch")
+
+def extract_tarball(tarball_name):
+    extract_dir = build_dir()
```
lucafavatella: Please do not change the build dir. Travis uploads a file from the build dir (see `file_glob: true` and `file: _build/prod/rel/epoch/*-epoch-0.1.0.tar.gz` in .travis.yml) and extracting stuff in the build dir may confuse the upload by Travis. Please use a temporary folder for extracting tarball.

lucafavatella: BTW that file path in travis.yml shall be made more dynamic e.g. `_build/prod/rel/epoch/*-epoch-*.tar.gz` (untested).

velzevur: Yes, it should be

ci
```diff
@@ -58,7 +58,17 @@ case "${1:?}"-"${2:?}" in
         BuildDir="${3:?}"
         ( cd "${BuildDir:?}" && ./rebar3 as prod tar; )
         Prefix=$(package_prefix "${TRAVIS_OS_NAME:?}")
-        cp -p "${BuildDir:?}"/_build/prod/rel/epoch/epoch-0.2.0.tar.gz "${BuildDir:?}"/_build/prod/rel/epoch/"${Prefix:?}"-epoch-0.2.0.tar.gz
+        cp -p "${BuildDir:?}"/_build/prod/rel/epoch/epoch-0.1.0.tar.gz "${BuildDir:?}"/_build/prod/rel/epoch/"${Prefix:?}"-epoch-0.1.0.tar.gz
```
lucafavatella: 0.1.0 or 0.2.0?

rebar.config
```diff
@@ -10,7 +10,7 @@
         {jobs, "0.7.0"}
        ]}.
 
-{relx, [{release, { epoch, "0.2.0" },
+{relx, [{release, { epoch, "0.1.0" },
```
lucafavatella: 0.1.0 or 0.2.0?

ci
```diff
@@ -58,7 +58,17 @@ case "${1:?}"-"${2:?}" in
...
+        ;;
+    script-acceptance-test)
```
lucafavatella: I see the block of code below is not hooked to the CI. Is this meant to be a new job (see .travis.yml) or is this meant to be executed in the `package` job (see .travis.yml)?

ci
velzevur: ah! Nice catch, thanks

ci
```diff
@@ -60,6 +60,26 @@ case "${1:?}"-"${2:?}" in
...
+        if [ "$TRAVIS_TAG" != "" ]; then
+          BuildDir="${3:?}"
+          ( cd "${BuildDir:?}" && ./rebar3 as fastmining tar; )
+          make nose-env
+          make release-integration-test TARBALL=_build/fastmining/rel/epoch/epoch-0.2.0.tar.gz
+        else
+          true
+        fi
+        ;;
```
lucafavatella: This section looks repeated twice.

rebar.config
```diff
@@ -38,6 +38,11 @@
                              {vm_args, "./config/dev1/vm.args"}]}
                     ]
             },
+            {fastmining, [{relx, [{dev_mode, false},
```
lucafavatella: I understand this profile will be deleted once the production release has mining speed configurable, right?
(no need to amend PR for this - just checking we have same understanding)

velzevur: Yes, exactly

ci
```diff
@@ -60,6 +60,12 @@ case "${1:?}"-"${2:?}" in
         Prefix=$(package_prefix "${TRAVIS_OS_NAME:?}")
         cp -p "${BuildDir:?}"/_build/prod/rel/epoch/epoch-0.2.0.tar.gz "${BuildDir:?}"/_build/prod/rel/epoch/"${Prefix:?}"-epoch-0.2.0.tar.gz
         ;;
+    script-acceptance-test)
+        BuildDir="${3:?}"
+        ( cd "${BuildDir:?}" && ./rebar3 as fastmining tar; )
+        make nose-env
+        make release-integration-test TARBALL=_build/fastmining/rel/epoch/epoch-0.2.0.tar.gz
```
lucafavatella: I see master may have moved on from the hardcoded 0.2.0.

py/tests/release.py
```diff
@@ -0,0 +1,170 @@
...
+import swagger_client
+from swagger_client.rest import ApiException
+from swagger_client.apis.external_api import ExternalApi
+from swagger_client.api_client import ApiClient
+
+SETUP = {
+        "node1": {
+            "host": "localhost:9900/v1",
+            "commands": [
+                "sed -ibkp 's/-sname epoch/-sname epoch1/g' ./releases/0.2.0/vm.args"
```
lucafavatella: Again - master may have moved on from hardcoded 0.2.0

py/tests/release.py
```diff
@@ -0,0 +1,170 @@
...
+ [
+  {swagger_port_external, 9900}
+ ]}
+].
+'''
+                },
+        "node2": {
+            "host": "localhost:9902/v1",
+            "commands": [
+                "sed -ibkp 's/-sname epoch/-sname epoch2/g' ./releases/0.2.0/vm.args"
```
lucafavatella: same

py/tests/release.py
```diff
@@ -0,0 +1,170 @@
...
+ [
+  {swagger_port_external, 9902}
+ ]}
+].
+'''
+                },
+        "node3": {
+            "host": "localhost:9903/v1",
+            "commands": [
+                "sed -ibkp 's/-sname epoch/-sname epoch3/g' ./releases/0.2.0/vm.args"
```
lucafavatella: same

Issue #226, opened 2017-09-29
Rebuild state of the world when a stronger chain is found


wagerlabs:
We update the state of the world when the block is on the same chain. We need to update/rebuild when a _stronger_ chain is identified.

We are not going to identify common ancestors for the [Good Peers sprint](https://github.com/aeternity/epoch/milestone/3) but accept the stronger chain and rebuild from genesis.

sennui:
copy-pasting comment from slack


Let's not go in this sprint into resolution where is the common ancestor. In case of stronger fork, lets accept the stronger chain and rebuild from genesis. To make it nicer, we need new api in state tree, that will wrap two existing apis: zero-out internal state  and after that build state from chain

zero out can be do with `force_trees`, and `apply_txs` will apply transactions from the new stronger fork

PR #258, opened 2017-10-05
GH-226 Rebuild state when stronger chain found - case only genesis in common

lucafavatella:
In scope of GH-226.
Supersedes PR #255.

### DIFF
--- a/apps/aecore/src/aec_chain.erl
+++ b/apps/aecore/src/aec_chain.erl
@@ -1159,6 +1159,9 @@ do_force_insert_headers_internal_1(
                   end,
                   HeadersDb,
                   HeaderChainWithWork),
+            {ok, _} = aec_state:force_trees(
+                        aec_blocks:trees(aec_block_genesis:genesis_block()),
+                        ?GENESIS_HEIGHT),
             {ok, NewTopHeaderDb} =
                 top_header_db_put(TopHeaderDb, ?TOP_HEADER, NewTopHeaderHash),
             {ok, {NewHeadersDb, NewBlocksDb}} =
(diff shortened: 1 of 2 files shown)


apps/aecore/src/aec_chain.erl
```diff
@@ -1169,6 +1172,7 @@ do_force_insert_headers_internal_1(
             {ok, NewTopBlock} =
                 do_find_highest_block_from_header_hash(NewTopHeaderHash,
                                                        NewHeadersDb, NewBlocksDb),
+            %% TODO Trigger check chain for successor.
```
lucafavatella: I need to code unit test that fails without this line implemented, then I need to code (or reuse) function to put here.

lucafavatella: I plan to open new PR (depending on this PR) for addressing this point.

apps/aecore/src/aec_chain.erl
```diff
@@ -1159,6 +1159,9 @@ do_force_insert_headers_internal_1(
                   end,
                   HeadersDb,
                   HeaderChainWithWork),
+            {ok, _} = aec_state:force_trees(
```
sennui: This is resetting trees, only

apps/aecore/src/aec_chain.erl
sennui: OK, and this is the second half of work

apps/aecore/src/aec_chain.erl
lucafavatella: Correct.

lucafavatella:
Closing this as #259 now open - more complete as it caters for case missing in this PR.

PR #291, opened 2017-10-17
Added exometer_core and defined an initial set of metrics

uwiger:
The metrics added are in aecore, and are mainly there to illustrate
the concepts. Note that try .. catch clauses around metric updates
are primarily there to not crash EUnit tests.

### DIFF
--- a/apps/aecore/src/aecore.app.src
+++ b/apps/aecore/src/aecore.app.src
@@ -12,7 +12,9 @@
     rocksdb,
     aeutils
    ]},
-  {env, []},
+  {env, [
+        {exometer_predefined, {script, "$PRIV_DIR/exometer_predefined.script"}}
+        ]},
   {modules, []},
 
   {maintainers, []},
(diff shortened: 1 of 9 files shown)


apps/aecore/src/aecore.app.src
```diff
@@ -12,7 +12,9 @@
     rocksdb,
     aeutils
    ]},
-  {env, []},
+  {env, [
+        {exometer_predefined, {script, "$PRIV_DIR/exometer_predefined.script"}}
```
lucafavatella: I see `$PRIV_DIR` is achieved by app `setup` as commented [here](https://github.com/Feuerlabs/exometer_core/issues/58#issuecomment-169685186).

uwiger: Yes, `setup` has quite a few tricks regarding configuration. Exometer uses it among other things to find exometer-related configs in other application environments.

Issue #60, opened 2017-09-06
Check that each block has no more than one coinbase transaction


wagerlabs:

cytadela8:
I think this will just be a part of validation. There is no way to do the check inside coinbase tx code, becouse it doesn't have access to other transactions, just to the trees. We can force coinbase_tx easly to appear only as the first transaction in the block in block validation code. (#17)

(Unimportant note: Analysis made, because I was considering if I can maybe solve this issue now)

Issue #17, opened 2017-08-28
Validate received block


wagerlabs:
Header checks...

- Protocol version should match.
- Fields should not be empty.
- PoW evidence should be correct.
- Difficulty should be no lower than target.
- Time should not be too far into the future. Should be checked against the machine UTC time, not previous block time.

Block checks...
	
- Transaction digest.
- State digest.

wagerlabs:
https://www.pivotaltracker.com/story/show/152481000

PR #307, opened 2017-10-23
GH-17 Add received header and block validation

zp-sd:
* This does not include "state digest" listed out in https://github.com/aeternity/epoch/issues/17, as state is transformed only whenever a block is appended to the chain, hence we cannot do that when receiving a block (e.g. block may land in a pool to wait for its predecessors in the chain)
* `aec_headers:validate(Header)` and `aec_blocks:validate(Block)` are hooked up into `aehttp_dispatch_ext.erl`, but are planned to be moved from it whenever logic is moved out of the http dispatcher
* Closes https://github.com/aeternity/epoch/issues/60

### DIFF
--- a/apps/aecore/src/aec_blocks.erl
+++ b/apps/aecore/src/aec_blocks.erl
@@ -15,7 +15,8 @@
          deserialize_from_network/1,
          serialize_to_map/1,
          deserialize_from_map/1,
-         hash_internal_representation/1]).
+         hash_internal_representation/1,
+         validate/1]).
 
 -ifdef(TEST).
 -compile([export_all, nowarn_export_all]).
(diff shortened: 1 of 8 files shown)


apps/aecore/src/aec_blocks.erl
```diff
@@ -162,3 +163,40 @@ deserialize_from_map(#{<<"height">> := Height,
...
 
+-spec validate(block()) -> ok | {error, term()}.
+validate(Block) ->
+    Validators = [fun validate_coinbase_txs_count/1,
+                  fun validate_txs_hash/1],
+    aeu_validation:run(Validators, Block).
+
+-spec validate_coinbase_txs_count(block()) -> ok | {error, multiple_coinbase_txs}.
+validate_coinbase_txs_count(#block{txs = Txs}) ->
+    CoinbaseTxsCount = lists:filter(
```
happi: Why not do a fold over integers if only is the count you want?

zp-sd: Good point.

zp-sd:
I have rebased on master. When tests pass, it's good to merge.

happi:
Approved.

Issue #64, opened 2017-09-06
Block with no parent


wagerlabs:
We may receive a block that has no parent, e.g. where the full ancestry of the block cannot be determined and the block cannot be validated. 

We need to

- Ask peers to fill in the history of this block.
- Keep the block around until history is filled.
- Validate the block.

kravets:
Not discarding such a block and not disconnecting the peer it came from ( at least for a while ) opens up a DoS vector ...

Issue #59, opened 2017-09-06
Handle orphaned blocks


wagerlabs:
We may find a higher block in a chain that does not include our current block. We'll need to go with the longest chain and orphan our current block. Its transactions need to be returned to the mempool if not part of the longest chain. See #16.

happi:
Closed with PR #308

PR #308, opened 2017-10-24
Introduce functional adt for chain handling.

happi:
Major rewrite of chain handling.

The chain is now kept in a functional ADT handled by aec_chain_state.erl
This chain is kept by a server (aec_chain_server.erl) and the public interface is still aec_chain.erl.
The interface is mostly backwards compatible with the old implementation but some internal functions that were exposed (for testing and for the chain_state) are deprecated and removed.
The new chain can handle alternative chains and blocks without parents.
The state is check pointed at each block on the main chain. (This will be cleaned up by a checkpoint GC in the future.)
The chain automatically switches to the chain with most work.

Closing #64 and #59


gorillainduction:
Sorry for rewriting so much code, guys. This is not an easy PR to review.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -0,0 +1,458 @@
...
+                true  -> State;
+                false -> internal_error({different_header_existing_block, Old, Node})
+            end
+    end.
+
+check_update_after_insert(Node, State) ->
+    case determine_chain_relation(Node, State) of
+        off_chain -> State;
+        in_chain -> update_state_tree(Node, State);
+        {fork, ForkNode} ->
```
sennui: please confirm: here we switch over to fork always when we get it.

We need to add difficulty check for incoming fork (just to make sure if someone is not cheating with lower target). Actually, we need to compute total difficulty in each branch from fork point

happi: Yes, this is done. Look at determine_new_header_top.

apps/aecore/src/aec_chain_server.erl
```diff
@@ -0,0 +1,201 @@
...
+    process_flag(trap_exit, true),
+    GenesisHeader = aec_blocks:to_header(GenesisBlock),
+    {ok, S2} = insert_header(GenesisHeader, new_state()),
+    {ok, S3} = insert_block(GenesisBlock, S2),
+    {ok, S3}.
+
+
+%% State preserving functions
+handle_call(top, _From, State) ->
+    {reply, top(State), State};
```
uwiger: Recognizing that this implementation is a 'shadow' version of `aec_chain`, is the intention to rename this module as `aec_chain`? It's not common practice to have API functions in one module do direct gen_server calls to a server implemented in another. If API functions are added to this module later, it seems as if the intuitive names for these functions are already reserved for internal functions (though perhaps with different arity). I would suggest using an 'internal naming' scheme for these – either `*_int()` or just `*_()` (which is a convention e.g. Tony Rogvall likes to use.)

apps/aecore/src/aec_chain_state.erl
```diff
@@ -0,0 +1,458 @@
...
+set_top_header_hash(H, State) when is_binary(H) -> State#{top_header_hash => H}.
+
+get_top_block_hash(#{top_block_hash := H}) -> H.
+set_top_block_hash(H, State) when is_binary(H) -> State#{top_block_hash => H}.
+
+%%%-------------------------------------------------------------------
+%%% Internal ADT for differing between blocks and headers
+%%%-------------------------------------------------------------------
+
+-record(node, { type    :: 'block' | 'header'
```
uwiger: For persistence reasons, would it perhaps be preferable to use maps here?

apps/aecore/src/aec_chain_state.erl
```diff
@@ -0,0 +1,458 @@
...
+    Height   = node_height(Node),
+    Hash     = Node#node.hash,
+    PrevHash = prev_hash(Node),
+    assert_previous_height(Height, PrevHash, State),
+    case get_top_header_hash(State) of
+        PrevHash -> new_top;
+        undefined when Height > 0 ->
+            off_chain; %% No proper chain yet.
+        undefined when Height =:= 0 ->
+            new_top; %% A genesis block. TODO: Should this be checked?
```
uwiger: I think it should be checked, yes.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -0,0 +1,458 @@
...
+        Hash ->
+            in_chain; %% This is the top header
+        TopHash when is_binary(TopHash), Height =:= 0 ->
+            %% A new genesis block. TODO: This
+            error(new_genesis_block_nyi);
```
uwiger: Why not `internal_error()`?

apps/aecore/src/aec_chain_state.erl
```diff
@@ -0,0 +1,458 @@
...
+    end.
+
+find_new_header_top_from_node(Node, State) ->
+    Height = node_height(Node),
+    Hash = hash(Node),
+    case children(blocks_db_find_at_height(Height + 1, State), Hash) of
+        [] -> [Node];
+        [NextNode] -> find_new_header_top_from_node(NextNode, State);
+        [_|_]  = List ->
+            lists:flatten([find_new_header_top_from_node(N, State)
```
uwiger: why not `lists:flatmap(fun(N) -> find_new_header_top_from_node(N, State) end, List)`?

apps/aecore/src/aec_chain_state.erl
```diff
@@ -0,0 +1,458 @@
...
+children([N|Ns], ParentHash, Acc) ->
+    case prev_hash(N) =:= ParentHash of
+	true  -> children(Ns, ParentHash, [N|Acc]);
+	false -> children(Ns, ParentHash, Acc)
+    end.
+
+determine_new_header_top(Nodes, State) ->
+    Difficulties = [{total_difficulty_at_hash(N#node.hash, State), N}
+                    || N <- Nodes],
+    {_D, N}  = lists:last(lists:sort(Difficulties)),
```
uwiger: Why not `lists:max(Difficulties)`?

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -269,15 +255,14 @@ get_work_test_() ->
                BH0 = aec_blocks:to_header(B0),
                ?assertEqual({ok, BH0}, aec_chain:top_header()),
 
-               %% Check difficulty of genesis - for readability of the test.
-               1.0 = aec_headers:difficulty(BH0),
+               %% Check difficulty of genesis
+	       %% NOTE: floating point matching... not safe.
```
uwiger: My guess is it's *pretty* safe in this case, but still, if it's regarded as unsafe, wouldn't it be better to change it? E.g. into something like `trunc(1000 * ?GENESIS_DIFFICULTY) = trunc(1000, ...)`

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -287,7 +272,8 @@ get_work_test_() ->
                ?assertEqual({ok, BH0}, aec_chain:top_header()),
 
                %% Check difficulty of genesis - for readability of the test.
-               1.0 = aec_headers:difficulty(BH0),
+	       %% NOTE: floating point matching... not safe.
```
uwiger: Same comment as above.

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -671,7 +597,7 @@ longest_header_chain_test_() ->
                %% Generate the two header chains.
                B0 = fake_genesis_block_with_difficulty(),
                0 = aec_blocks:height(B0), %% For readability of the test.
-               1.0 = aec_blocks:difficulty(B0), %% For readability of the test.
+               ?GENESIS_DIFFICULTY = aec_blocks:difficulty(B0),
```
uwiger: Same comment as above ('not safe')

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -688,46 +614,18 @@ longest_header_chain_test_() ->
...
-                 lists:nthtail(1, MainHC))
+               ?assertEqual({ok, HA3}, aec_chain:top_header())
        end},
       {"The alternative header chain has more work - case alternative chain is higher. Force chain including genesis.",
        fun() ->
                %% Generate the two header chains.
                B0 = fake_genesis_block_with_difficulty(),
                0 = aec_blocks:height(B0), %% For readability of the test.
-               1.0 = aec_blocks:difficulty(B0), %% For readability of the test.
+               ?GENESIS_DIFFICULTY = aec_blocks:difficulty(B0),
```
uwiger: Same comment as above ('not safe')

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -744,45 +642,18 @@ longest_header_chain_test_() ->
...
-                 lists:nthtail(1, MainHC))
+               ?assertEqual({ok, HA3}, aec_chain:top_header())
        end},
       {"The alternative header chain has more work - case alternative chain is less high",
        fun() ->
                %% Generate the two header chains.
                B0 = fake_genesis_block_with_difficulty(),
                0 = aec_blocks:height(B0), %% For readability of the test.
-               1.0 = aec_blocks:difficulty(B0), %% For readability of the test.
+               ?GENESIS_DIFFICULTY = aec_blocks:difficulty(B0),
```
uwiger: Same comment as above ('not safe').

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -841,12 +686,12 @@ longest_header_chain_test_() ->
...
```
uwiger: Same comment as above ('not safe').

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -913,21 +739,18 @@ longest_block_chain_test_() ->
                %% Generate the two block chains.
                B0 = fake_genesis_block_with_difficulty(),
                0 = aec_blocks:height(B0), %% For readability of the test.
-               1.0 = aec_blocks:difficulty(B0), %% For readability of the test.
+               ?GENESIS_DIFFICULTY = aec_blocks:difficulty(B0),
```
uwiger: Same comment as above ('not safe').

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -985,47 +786,28 @@ longest_block_chain_test_() ->
...
+                                                      aec_blocks:trees(BA3)))
+
        end},
       {"The alternative block chain has more work - case both main and alternative chains with all blocks. Two blocks in common.",
        fun() ->
                %% Generate the two block chains.
                B0 = fake_genesis_block_with_difficulty(),
                0 = aec_blocks:height(B0), %% For readability of the test.
-               1.0 = aec_blocks:difficulty(B0), %% For readability of the test.
+               ?GENESIS_DIFFICULTY = aec_blocks:difficulty(B0),
```
uwiger: Same comment as above ('not safe').

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -1092,45 +852,28 @@ longest_block_chain_test_() ->
...
-                 end,
-                 lists:nthtail(2, MainHC))
        end},
       {"The alternative block chain has more work - case main chain with only block corresponding to top header (not contiguous to genesis), and alternative chain with only block corresponding to top header (not contiguous to genesis). Only genesis block in common.",
        fun() ->
                %% Generate the two block chains.
                B0 = fake_genesis_block_with_difficulty(),
                0 = aec_blocks:height(B0), %% For readability of the test.
-               1.0 = aec_blocks:difficulty(B0), %% For readability of the test.
+               ?GENESIS_DIFFICULTY = aec_blocks:difficulty(B0),
```
uwiger: Same comment as above ('not safe').

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -1183,41 +910,24 @@ longest_block_chain_test_() ->
...
-                 end,
-                 lists:nthtail(1, MainHC))
        end},
       {"The alternative block chain has more work - case main chain with only one block contiguous to genesis, and alternative chain with two blocks contiguous to genesis. Only genesis block in common.",
        fun() ->
                %% Generate the two block chains.
                B0 = fake_genesis_block_with_difficulty(),
                0 = aec_blocks:height(B0), %% For readability of the test.
-               1.0 = aec_blocks:difficulty(B0), %% For readability of the test.
+               ?GENESIS_DIFFICULTY = aec_blocks:difficulty(B0),
```
uwiger: Same comment as above ('not safe').

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -1274,30 +966,17 @@ longest_block_chain_test_() ->
 
                %% Check top block changed.
                ?assertEqual({ok, HA3}, aec_chain:top_header()),
-               ?assertEqual({ok, BA2}, aec_chain:top_block()),
-
+	       
+	       %% !!!! Race condition here !!!!
+	       timer:sleep(2000),
```
uwiger: Isn't there any way to achieve a sync point here without resorting to timers? A TODO perhaps?

apps/aecore/src/aec_chain_server.erl
```diff
@@ -0,0 +1,201 @@
...
+    gen_server:stop(?SERVER).
+
+%%%===================================================================
+%%% gen_server callbacks
+%%%===================================================================
+ 
+init(_Args = [GenesisBlock]) ->
+    process_flag(trap_exit, true),
+    GenesisHeader = aec_blocks:to_header(GenesisBlock),
+    {ok, S2} = insert_header(GenesisHeader, new_state()),
```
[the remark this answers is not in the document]
happi: Good catch, refactoring artifact.

apps/aecore/src/aec_chain_server.erl
happi: I think it is a better and cleaner practice to separate the outside API and the server implementation. We can discuss offline, and then come back here for a conclusion.

apps/aecore/src/aec_chain_state.erl
happi: This is only when there was no genesis block before.

apps/aecore/test/aec_chain_tests.erl
happi: This timer is not needed with the new implementation which only uses one server.
will remove.

apps/aecore/src/aec_chain_state.erl
gorillainduction: Good one. I will update.

apps/aecore/test/aec_chain_tests.erl
gorillainduction: The race is not there anymore. It was an artifact in the old version. I remove this.

gorillainduction:
Please note that we have aimed for correctness and not speed/space optimizations. The plan is to introduce such optimizations in the next step.

apps/aecore/src/aec_chain_state.erl
gorillainduction: Sure.

apps/aecore/src/aec_chain_state.erl
gorillainduction: I will make this more clear in the code.

apps/aecore/src/aec_chain_state.erl
happi: Maybe. It depends on how we will do persistence and how we will handle code change. If we use mnesia or dets a record is probably easiest. For RocksDB or something else we would need to do some serialization and deserialization anyway. Records might be slightly more efficient here than records. Versioning of the data will have to be handled anyway.

I will look at persistence next and perhaps come back and change this if needed but for now I suggest we go with records.

PR #309, opened 2017-10-25
Use canonical serialization for hash.

### LINKED ISSUE #297: Canonical header hashing
Need to base the header hash on the binary representation of the header.

`aec_header:hash_header/1` is basing its hash on the JSON (as a binary) representation of the header. This is dangerous, as the order of the fields in JSON objects is undefined. 

We should base it on the [canonical representation of the block header](https://github.com/aeternity/epoch/wiki/Block-Header) instead.

happi:
Also provide serialization for persistent storage.

Closes #297

### DIFF
--- a/apps/aecore/src/aec_blocks.erl
+++ b/apps/aecore/src/aec_blocks.erl
@@ -13,6 +13,8 @@
          to_header/1,
          serialize_for_network/1,
          deserialize_from_network/1,
+         serialize_for_store/1,
+         deserialize_from_store/1,
          serialize_to_map/1,
          deserialize_from_map/1,
          hash_internal_representation/1,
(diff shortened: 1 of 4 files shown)


happi:
Some white space errors (tabs) made it in to my commit, I'll fix and update the PR.

akorosmezey:
Is term_to_binary/1 enough here? We may enforce sizes defined in 
https://github.com/aeternity/epoch/wiki/Block-Header
in order to be compatible with other implementations.

happi:
@akorosmezey It is only serialize for store that uses term_to_binary. This is an internal implementation.

The hash is calculated on the binary based on https://github.com/aeternity/epoch/wiki/Block-Header
(Assuming that the hashes in the header has the right sizes, this should perhaps be tested.

uwiger:
It would perhaps have been better to squash the commits first.

happi:
squashed

PR #310, opened 2017-10-30
Transaction broadcast and changes in verification/serialization

uwiger:
- using msgpack to serialize array of maps for signing
- rewrote mempool to use ets tables
- unit tests currently don't pass due to tx verification errors


uwiger:
Closing. Replaced by #323

PR #314, opened 2017-11-02
Gh 185 spend transaction

zp-sd:
What is left:
* sign, put into the mempool and propagate to other peers transaction on `POST /spend-tx`
* change serialization to use msgpack, as it is done in https://github.com/aeternity/epoch/pull/310
* add sum of transaction fees to the miner's account when updating accounts state tree when creating/validating block candidate

### DIFF
--- a/apps/aecore/src/txs/aec_spend_tx.erl
+++ b/apps/aecore/src/txs/aec_spend_tx.erl
@@ -1,0 +1,158 @@
+-module(aec_spend_tx).
+
+%% API
+-export([new/2,
+         check/3,
+         process/3,
+         serialize/1,
+         deserialize/1,
+         type/0]).
+
+-behavior(aec_tx).
+
+-include("common.hrl").
+-include("trees.hrl").
+-include("txs.hrl").
+
+-spec new(map(), trees()) -> {ok, spend_tx()} | {error, sender_account_not_found}.
+new(#{sender := SenderPubkey,
+      recipient := RecipientPubkey,
+      amount := Amount,
+      fee := Fee}, Trees) ->
+    AccountsTrees = aec_trees:accounts(Trees),
+    case aec_accounts:get(SenderPubkey, AccountsTrees) of
+        {ok, #account{nonce = CurrentSenderNonce}} ->
+            {ok, #spend_tx{sender = SenderPubkey,
+                           recipient = RecipientPubkey,
+                           amount = Amount,
+                           fee = Fee,
+                           nonce = CurrentSenderNonce + 1}};
+        {error, notfound} ->
+            {error, sender_account_not_found}
+    end.
+
+-spec check(spend_tx(), trees(), height()) -> {ok, trees()} | {error, term()}.
+check(#spend_tx{recipient = RecipientPubkey} = SpendTx, Trees0, Height) ->
+    Checks = [fun check_tx_fee/3,
+              fun check_sender_account/3],
+    case aeu_validation:run(Checks, [SpendTx, Trees0, Height]) of
+        ok ->
+            case aec_tx_common:ensure_account_at_height(RecipientPubkey, Trees0, Height) of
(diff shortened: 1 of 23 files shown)


apps/aeutils/src/aeu_validation.erl
```diff
@@ -5,7 +5,7 @@
 run([], _Arg) ->
     ok;
 run([F | Rest], Arg) ->
-    case F(Arg) of
+    case apply(F, Arg) of
```
happi: To be a bit more explicit here Arg should be ArgList or something as I guess there are more than one argument since you need apply. (And the validation functions seem to take three arguments. Perhaps even be even more explicit and expect Arg to be a three element list and use F(Arg1, Arg2, Arg3) or even named arguments.

apps/aecore/src/txs/aec_spend_tx.erl
```diff
@@ -0,0 +1,152 @@
...
+    Trees = aec_trees:set_accounts(Trees0, AccountsTrees2),
+    {ok, Trees}.
+
+-spec serialize(spend_tx()) -> map().
+serialize(#spend_tx{sender = Sender,
+                    recipient = Recipient,
+                    amount = Amount,
+                    fee = Fee,
+                    nonce = Nonce}) ->
+    #{<<"sender">> => base64:encode(Sender),
```
uwiger: Please make this an array of maps instead, and without encoding:
```erlang
[#{<<"sender">> => Sender},
 #{<<"recipient">> => Recipient},
 ...]
```
C.f. https://github.com/aeternity/epoch/blob/gh-279-process-transactions/apps/aecore/src/txs/aec_coinbase_tx.erl#L54 (PR #310). Note also the inclusion of type and vsn.

apps/aehttp/priv/swagger.json
```diff
@@ -148,6 +148,12 @@
         "responses" : {
           "200" : {
             "description" : "successful operation"
+          },
+          "400" : {
```
uwiger: Elswhere in this spec, the code `404` has been used for error returns.

apps/aehttp/priv/swagger.json
```diff
@@ -183,6 +189,36 @@
...
+          "required" : true,
+          "schema" : {
+            "$ref" : "#/definitions/SpendTx"
+          }
+        } ],
+        "responses" : {
+          "200" : {
+            "description" : "successful operation"
+          },
+          "400" : {
```
uwiger: Elswhere in this spec, the code 404 has been used for error returns.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -134,6 +134,25 @@ handle_request('GetAccountBalance', Req, _Context) ->
...
+    Trees = aec_blocks:trees(LastBlock),
+    case aec_spend_tx:new(#{sender => SenderPubkey,
+                            recipient => RecipientPubkey,
+                            amount => Amount,
+                            fee => Fee}, Trees) of
+        {ok, _SpendTx} ->
+            %% TODO: sign and push to the mempool
+            {200, [], #{}};
+        {error, sender_account_not_found} ->
+            {400, [], #{reason => <<"sender account not found">>}}
```
uwiger: Should be `404` for consistency.

rebar.config
```diff
@@ -24,7 +24,7 @@
 }.
 
 {plugins,
-  [{tx_xform, {git, "https://github.com/aeternity/tx_xform.git", {branch, "master"}}}]}.
+  [{tx_xform, {git, "https://github.com/zp-sd/tx_xform", {branch, "accept-error-response-from-aec_tx-new/2"}}}]}.
```
uwiger: The PR to `tx_xform` has been merged now, so pls revert this change.

PR #315, opened 2017-11-03
GH-192 Python release integration test [Finishes #152484373]

velzevur:
All builds with a tag set to anything will trigger a Tavis JOB (one per OS - linux and osx) to validate the release is correct. The script does the following

- generate a release like tarball (only change being fast mining is enabled)
- extract the tarball
- copy the extracted dir to 2 more directories (so we have node1, node2, node3)
- setup the nodes
- start the nodes
- get each node's top every second untill there are 50 blocks mined
- stop the nodes and cleanup

If a node crashes - all node logs are tailed so we can easily debug it.

It is a huge commit and it is mostly due to that I had to add some Swagger generated python code (so we don't have to install swager-codegen in the JOB). The generated code is located in ` py/tests/swagger_client/ `
You can focus on changes in `.travis.yml`, `Makefile`, `ci`, `py/tests/release.py` and `rebar.config` 

Example for the result of the PR in action could be seen [here ](https://travis-ci.org/velzevur/epoch/builds/296469256), which results in [this release](https://github.com/velzevur/epoch/releases/tag/v026). The only differences are some Travis settings (to upload the releases to my fork) and changed `VERSION` file to be `0.2.1`


.travis.yml
```diff
@@ -22,7 +22,7 @@ deploy:
     # See https://docs.travis-ci.com/user/deployment/releases/#Authenticating-with-an-OAuth-token
     secure: B4Yl81oyddyE2TQVmVggOgZFN3aZTIt5XjV4mQsy91ctqRxQskIa4nl+vLpNRDHMS+6P9cbEfWdYRXamrOm3PsPOm6KQcoekKbmHH45PdKhHkMbAiUxjXW/QTlufeEPs8ugplAv8anj/d2cb7D2mfzjP6PY1QIbCPbYJGGC0O/zeYHOUxwODyYZQzO6xsBQfOQdHwoT4D/2c1Oo6Mz31j8hxXXJZgXGtEoYjJnxZVHdU1Y7irs8CWlwW7pAjTjEs4cpBVCaM/3W2VQ1I/bW0ZGmZYv8g4Av3uBA6SypAXrqo1yzB2teOUPtxP/WvDzOl/wfZnLQH0u/6uaCt3/XInIYINY8F/Tck31rErPeCrV520WbGebG/ou2uzCHei9cl138doO8qemRf544nrCoSxz39gC2EMcl6PVH3mKpEnj8yLtXuqrfT36mKEzAILb0PQXNFD9Ccx48dGCVa86AZ2zxPnWO9s+veZN4/i9IjW+8sdTIekX96KkAadC/FmMMltbbKHnudEKh2wUujO1ra+OlF6GO7y5E0fEQ8/2C5uwnyMMFwrPAw6snf3pM7z8cEcC/KaqnqjGrOlQ1ze8Ra4NXuMNLwIqquQYxcBx6pJcGU+FavWTFflowzWlTmr5qHj5j3SoRmHqrjEezhHqDr/pvPZp4MNRuJMUOHtcNP4Y4=
   file_glob: true
-  file: _build/prod/rel/epoch/*-epoch-"$(cat VERSION)".tar.gz
+  file: _build/prod/rel/epoch/epoch-*-*.tar.gz
```
lucafavatella: I see this depends on f0e005d too. Ok.

.travis.yml
```diff
@@ -36,6 +36,13 @@ matrix:
       osx_image: xcode8.3 # [`xcode8.3` is Xcode 8.3.3 on OS X 10.12](https://docs.travis-ci.com/user/reference/osx#OS-X-Version)
       language: generic
       env: JOB=package
+    - if: tag IS present # additional JOB to run the release acceptance test for releases only
```
lucafavatella: Please comment on why these additional tests are run on separate job rather than existing job that builds package (and that uploads that on tag), and whether this is meant to be affected once mining speed becomes configurable and the fastmining profile gets removed. This will help when restructuring tests.

py/tests/release.py
```diff
@@ -0,0 +1,192 @@
...
+        print("node died")
+    [stop_node(d) for d in node_dirs]
+    if test_failed:
+        for name, node_dir in zip(node_names, node_dirs):
+            print(name + " logs:")
+            print(tail_logs(node_dir, "epoch.log"))
+            print("\n")
+    shutil.rmtree(root_dir)
+    if test_failed:
+        sys.exit("FAILED")
```
lucafavatella: I see [this results in exit code 1](https://docs.python.org/3/library/sys.html#sys.exit). Ok.

PR #319, opened 2017-11-03
Locate and load config file (currently only supporting peers option)

uwiger:
Fixes #152483487

Config file can be identified by:
- OS environment variable EPOCH_CONFIG
- a file, "epoch.{json,yaml}" in either of:
  + ~/.epoch/<node>/
  + epoch top directory (e.g. _build/dev1/rel/epoch)
- Syntax determined from suffix (.yaml or .json)


lucafavatella:
> ~/.epoch//

@uwiger I expect the user would prefer not having items directly in `~/` but in more confined location e.g. [XDG location `~/.config/epoch/`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables).
Other option is [OTP filename:basedir/2](http://erlang.org/documentation/doc-9.0/lib/stdlib-3.4/doc/html/filename.html#basedir-2).

uwiger:
@lucafavatella, I guess this boils down to a namespace issue: who 'owns' the `~/.config` directory?

uwiger:
I note that the test suite passes now, after a number of job restarts. The failures had nothing to do with this PR. Also note that this PR has no tests specific to the branch. I've tested it manually.

PR #320, opened 2017-11-03
Add persistence to chain, snapshots of state trees, transactions to pool

gorillainduction:
Once again a huge PR. All these issues were so closely related so it made no sense splitting them up.

The aec_miner has been rewritten to make sure no messages are lost.

The block chain is persisted between node restarts if the sys config 'persist' is set to 'true'. This goes for both the blocks, headers and state trees.

When the top block changes through a 'post_block', all transactions are synced to the pool.

The test suites are somewhat more stable.

[Finishes #152471554 #152471640 #152471623 #152471670 #152471762 #152471598 ]

### DIFF
--- a/apps/aecore/src/aec_keys.erl
+++ b/apps/aecore/src/aec_keys.erl
@@ -26,7 +26,8 @@
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
 
--export([new/1, open/1, set/3, sign/1, pubkey/0, delete/0]).
+-export([new/1, open/1, set/3, sign/1, pubkey/0, delete/0,
+         wait_for_pubkey/0]).
 
 -define(SERVER, ?MODULE).
 -define(PUB_SIZE, 65).
@@ -77,6 +78,18 @@ sign(Term) ->
 -spec pubkey() -> {ok, binary()} | {error, key_not_found}.
 pubkey() ->
     gen_server:call(?MODULE, pubkey).
+
+-spec wait_for_pubkey() -> {ok, binary()}.
+wait_for_pubkey() ->
+    wait_for_pubkey(1).
+
+wait_for_pubkey(Sleep) ->
+    case pubkey() of
+        {error, key_not_found} ->
+            timer:sleep(Sleep),
+            wait_for_pubkey(Sleep+10);
+        R -> R
+    end.
 
 -spec open(binary()) -> {ok, Password :: binary()} | {error, keys_not_loaded}.
 open(Password) ->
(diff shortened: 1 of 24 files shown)


apps/aecore/src/aec_keys.erl
```diff
@@ -78,6 +79,18 @@ sign(Term) ->
 pubkey() ->
     gen_server:call(?MODULE, pubkey).
 
+-spec wait_for_pubkey() -> {ok, binary()}.
+wait_for_pubkey() ->
+    wait_for_pubkey(1).
+
+wait_for_pubkey(Sleep) ->
+    case pubkey() of
```
lucafavatella: This wait_for_pubkey may wait forever e.g. if node was started without password.
The code could be simplified by [making `aec_keys` initialization simpler](https://www.pivotaltracker.com/story/show/152483887) hence avoiding need for this method. Though I understand this is not the focus of this PR so no need to change it yet (if I find strong reason to amend this I will add new comment).

apps/aecore/src/aec_miner.erl
```diff
@@ -133,128 +196,263 @@ init(_) ->
...
+    {keep_state, State, [postpone]};
+running(cast, check_keys, State) ->
+    %% The check_keys message should come when waiting.
+    epoch_mining:error("Got check_keys event in running state~n", []),
+    %% Keep the event and wait for a state transition.
+    {keep_state, State, [postpone]};
+%% Catchall, log & drop unwanted messages.
+running(_Type, _Msg, State) ->
+    epoch_mining:error("Unknown event in running: ~p:~p~n", [_Type,_Msg]),
+    {keep_state, running, State}.
```
lucafavatella: @gorillainduction Could you please clarify this return value? It appears to me you want `{keep_state, State}`.

gorillainduction: Indeed. This seems to be a bug. @happi can you comment?

gorillainduction: Although it is a bug, it is mostly left-over from cleaning up the messages in the gen_statem. This should never trigger.

I suggest we merge this, and I create a follow up ticket to fix this buglet to get the merge done today.

apps/aecore/src/aec_miner.erl
```diff
@@ -109,15 +127,60 @@ get_balance() ->
...
+%% If an option in not provided in the application environment nor
+%% in the argument to the init function the default from the state
+%% record definition is used.
+%% If the application environment provides an option it overrides
+%% the default.
+%% If the call argument provides an option it overrides both the
+%% application environment and the default.
+%%
+%% The options are:
+%%
```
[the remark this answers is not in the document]
happi: ;) Indeed. I think I had a plan to do that...

PR #321, opened 2017-11-06
PT-152484093 Validate decrypted keys [Finishes #152484093]

velzevur:
Validating the keys after decrypting them with a password.
Some binary is signed with the private key and then validated with the public one. This is the very definition of async crypto pair and is independent of the algorithms we use.

### DIFF
--- a/apps/aecore/src/aec_keys.erl
+++ b/apps/aecore/src/aec_keys.erl
@@ -28,6 +28,10 @@
 
 -export([new/1, open/1, set/3, sign/1, pubkey/0, delete/0]).
 
+-ifdef(TEST).
+-export([check_keys_pair/5]).
+-endif.
+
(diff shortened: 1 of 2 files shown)


apps/aecore/src/aec_keys.erl
```diff
@@ -296,3 +306,11 @@ to_local_dir(NewFile, Bin) ->
 
 from_local_dir(NewFile) ->
     file:read_file(NewFile).
+
+
+-spec check_keys_pair(binary(), binary(), atom(), atom(), atom()) -> boolean().
+check_keys_pair(PubKey, PrivKey, Algo, Digest, Curve) ->
+    SampleMsg = <<"random message">>,
```
sennui: wasn't there standard verify string for keys? "Sanity message"? This is good, but if you run into it in the future, please adjust.

PR #322, opened 2017-11-06
PT-152575233 Add all block transactions' fees to miner's account

zp-sd:
* Remove trees() argument from aec_tx:new/2
* [Finishes #152575233]

### DIFF
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -527,6 +527,9 @@ update_top_block_hash(Hash, State) ->
 apply_node_transactions([Node|Left], Trees, State) ->
     Txs = aec_blocks:txs(Node#node.content),
     Height = node_height(Node),
+    %% TODO: verify that root hash of state tree after applying transactions
+    %% is equal to #block.root_hash.
+    %% To be done in scope of "PT-152481000 Validate received block".
     {ok, NewTrees} = aec_tx:apply_signed(Txs, Trees, Height),
     apply_node_transactions(Left, NewTrees, State);
 apply_node_transactions([], Trees,_State) ->
--- a/apps/aecore/src/aec_mining.erl
+++ b/apps/aecore/src/aec_mining.erl
@@ -17,17 +17,18 @@
 create_block_candidate() ->
     {ok, LastBlock} = aec_chain:top(),
     Trees = aec_blocks:trees(LastBlock),
-    case get_txs_to_mine(Trees) of
+    case get_txs_to_mine() of
         {error, _} = Error ->
             Error;
         {ok, Txs} ->
-            case aec_blocks:new(LastBlock, Txs, Trees) of
-                {ok, Block0} ->
+            Block0 = aec_blocks:new(LastBlock, Txs, Trees),
+            case aec_blocks:cointains_coinbase_tx(Block0) of
+                true ->
                     Block = maybe_recalculate_difficulty(Block0),
                     {InitialNonce, MaxNonce} = aec_pow:pick_nonces(),
                     {ok, Block, InitialNonce, MaxNonce};
-                {error, _Reason} = Error ->
-                    Error
+                false ->
+                    {error, coinbase_tx_rejected}
             end
     end.
 
@@ -67,10 +68,10 @@ mine(Block, Attempts, InitialNonce, MaxNonce) ->
 
 %% Internal functions
 
--spec get_txs_to_mine(trees()) -> {ok, list(signed_tx()), non_neg_integer()} | {error, term()}.
-get_txs_to_mine(Trees) ->
+-spec get_txs_to_mine() -> {ok, list(signed_tx()), non_neg_integer()} | {error, term()}.
+get_txs_to_mine() ->
     {ok, Txs0} = aec_tx_pool:peek(aec_governance:max_txs_in_block() - 1),
-    case create_coinbase_tx(Trees) of
+    case create_coinbase_tx() of
         {ok, CoinbaseTx} ->
             case aec_keys:sign(CoinbaseTx) of
                 {ok, SignedCoinbaseTx} ->
@@ -82,11 +83,11 @@ get_txs_to_mine(Trees) ->
             Error
     end.
 
--spec create_coinbase_tx(trees()) -> {ok, coinbase_tx()} | {error, term()}.
-create_coinbase_tx(Trees) ->
+-spec create_coinbase_tx() -> {ok, coinbase_tx()} | {error, term()}.
+create_coinbase_tx() ->
     case aec_keys:pubkey() of
         {ok, Pubkey} ->
-            aec_coinbase_tx:new(#{account => Pubkey}, Trees);
+            aec_coinbase_tx:new(#{account => Pubkey});
         {error, _} = Error ->
             Error
     end.
--- a/apps/aecore/src/txs/aec_coinbase_tx.erl
+++ b/apps/aecore/src/txs/aec_coinbase_tx.erl
@@ -1,7 +1,7 @@
 -module(aec_coinbase_tx).
 
 %% API
--export([new/2,
+-export([new/1,
          check/3,
          process/3,
          serialize/1,
@@ -15,8 +15,8 @@
 -include("txs.hrl").
 
 
--spec new(map(), trees()) -> {ok, coinbase_tx()}.
-new(#{account := AccountPubkey}, _Trees) ->
+-spec new(map()) -> {ok, coinbase_tx()}.
+new(#{account := AccountPubkey}) ->
     {ok, #coinbase_tx{account = AccountPubkey}}.
 
 -spec check(coinbase_tx(), trees(), height()) -> {ok, trees()} | {error, term()}.
@@ -28,6 +28,9 @@ check(#coinbase_tx{account = AccountPubkey}, Trees0, Height) ->
             Error
     end.
 
+%% Only aec_governance:block_mine_reward() is granted to miner's account here.
+%% Amount from all the fees of transactions included in the block
+%% is added to miner's account in aec_tx:apply_signed/3.
 -spec process(coinbase_tx(), trees(), height()) -> {ok, trees()}.
 process(#coinbase_tx{account = AccountPubkey}, Trees0, Height) ->
     AccountsTrees0 = aec_trees:accounts(Trees0),
--- a/apps/aecore/src/txs/aec_tx.erl
+++ b/apps/aecore/src/txs/aec_tx.erl
@@ -1,7 +1,8 @@
 -module(aec_tx).
 
 -export([fee/1,
-         apply_signed/3]).
+         apply_signed/3,
+         is_coinbase/1]).
 
 -include("common.hrl").
 -include("trees.hrl").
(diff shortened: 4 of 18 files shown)


rebar.config
```diff
@@ -25,7 +25,7 @@
 }.
 
 {plugins,
-  [{tx_xform, {git, "https://github.com/aeternity/tx_xform.git", {branch, "master"}}}]}.
+  [{tx_xform, {git, "https://github.com/zp-sd/tx_xform.git", {branch, "change-arity-of-aec_tx_new"}}}]}.
```
zp-sd: See https://github.com/aeternity/tx_xform/pull/3.

apps/aecore/src/txs/aec_tx.erl
```diff
@@ -78,3 +106,26 @@ check_single(Tx, Trees, Height) ->
...
+    %% (valuable also during txs processing)
+    MinerPubkey = coinbase_tx_account_pubkey(Txs),
+    AccountsTrees0 = aec_trees:accounts(Trees0),
+
+    {ok, Account0} = aec_accounts:get(MinerPubkey, AccountsTrees0),
+    {ok, Account} = aec_accounts:earn(Account0, TotalFee, Height),
+
+    {ok, AccountsTrees} = aec_accounts:put(Account, AccountsTrees0),
+    Trees = aec_trees:set_accounts(Trees0, AccountsTrees),
+    Trees.
```
sennui: @uwiger @gorillainduction mind that there is no transaction representing update to miners account by sum of all fees

apps/aecore/src/txs/aec_tx.erl
```diff
@@ -40,29 +40,57 @@ fee(#spend_tx{fee = F}) ->
...
 
+-spec verify_and_extract_txs(list(signed_tx())) -> list(tx()).
+verify_and_extract_txs(SignedTxs) ->
+    verify_and_extract_txs(SignedTxs, []).
+
+-spec verify_and_extract_txs(list(signed_tx()), list(tx())) -> list(tx()).
+verify_and_extract_txs([], Txs) ->
+    lists:reverse(Txs);
+verify_and_extract_txs([SignedTx | Rest], Txs) ->
+    case aec_tx_sign:verify(SignedTx) of
```
sennui: this is mock'ed. @uwiger implements the verify API

rebar.config
lucafavatella: @zp-sd I would suggest you:
* Push branch change-arity-of-aec_tx_new in https://github.com/aeternity/tx_xform repo
* Make this rebar.config line refer to https://github.com/aeternity/tx_xform.git specific commit (rather than branch) i.e. `{ref, "fa96672"}` (referring to specific commit rather than master helps dealing with concurrent PRs as in this case)

apps/aecore/test/txs/aec_tx_tests.erl
```diff
@@ -0,0 +1,96 @@
...
+               %% Initial balance - spend_tx amount - spend_tx fee + spend_tx fee + coinbase_t[x reward
+               ?assertEqual(100 - 40 - 9 + 9 + 10, aec_accounts:balance(ResultMinerAccount)),
+               ?assertEqual(80 + 40, aec_accounts:balance(ResultRecipientAccount))
+       end
+      }]}.
+
+
+%% Internals
+
+init_keys() ->
```
lucafavatella: I see these test functions are now common and placed in common module https://github.com/aeternity/epoch/blob/3ddf1b87250adc5e64289d7d5cd7b4b25fa94335/apps/aecore/test/aec_test_utils.erl#L17-L18

(No need to change this PR only for this though.)

apps/aecore/src/aec_mining.erl
```diff
@@ -67,10 +63,10 @@ mine(Block, Attempts, InitialNonce, MaxNonce) ->
 
 %% Internal functions
 
--spec get_txs_to_mine(trees()) -> {ok, list(signed_tx()), non_neg_integer()} | {error, term()}.
-get_txs_to_mine(Trees) ->
+-spec get_txs_to_mine() -> {ok, list(signed_tx()), non_neg_integer()} | {error, term()}.
+get_txs_to_mine() ->
     {ok, Txs0} = aec_tx_pool:peek(aec_governance:max_txs_in_block() - 1),
-    case create_coinbase_tx(Trees) of
+    case create_coinbase_tx() of
```
lucafavatella: I see that other transactions in the block are not passed to creation of the signed coinbase tx, as the amount is not part of the coinbase tx. Ok.

apps/aecore/src/txs/aec_tx.erl
lucafavatella: I see that coinbase tx has the responsibility of identifying the account of the miner, and contains no amont.
@zp-sd Could you please add comment in [`aec_coinbase_tx:process`](https://github.com/aeternity/epoch/pull/322/files#diff-933a22fcb74313128547fe0436956dd9R37) mentioning that the earning of the total of the non-coinbase tx fees is not computed inside the coinbase transaction but rather at the end of the application of all transactions in the block?

apps/aecore/src/txs/aec_tx.erl
```diff
@@ -40,29 +40,57 @@ fee(#spend_tx{fee = F}) ->
     F.
 
 -spec apply_signed(list(signed_tx()), trees(), non_neg_integer()) ->
-                          {ok, trees()} | {error, term()}.
```
lucafavatella: This function `aec_tx:apply_signed` changed its API to always succeed by silently discarding invalid txs. I see this same function is used by the module [building the state trees in the chain](https://github.com/aeternity/epoch/blob/b0950fb9b0ddef9653a720486627c09ffff17c8c/apps/aecore/src/aec_chain_state.erl#L530) that shall see application of all txs succeed.

Please separate function in two:
* Once expecting all txs to be valid, and returning error otherwise (I envisage this to be used by (1) module processing block received from network and (2) module building state trees in chain);
* One silently discarding transactions (for mining - as proposed in this PR).

apps/aecore/src/txs/aec_tx.erl
```diff
@@ -40,29 +40,57 @@ fee(#spend_tx{fee = F}) ->
...
-                    apply_signed(Rest, Trees2, Height);
-                {error, _Reason} = Error ->
-                    Error
-            end;
-        {error, _Reason} = Error ->
-            Error
-    end.
+                          {ok, trees()}.
+apply_signed(SignedTxs, Trees0, Height) ->
+    Txs = verify_and_extract_txs(SignedTxs),
```
lucafavatella: As function `verify_and_extract_txs` silently discards txs, the coinbase tx may be silently discarded too!
Please make a proposal where in the code there is the hardcoded expectation that the first tx among the txs being included in the block to be mined is the coinbase tx (so that the reader does not need to assume so with the risk that block w/o coinbase transaction is mined and then rejected by other nodes in network).

apps/aecore/src/txs/aec_tx.erl
zp-sd: @lucafavatella Coinbase tx still adds `block_mine_reward` to miner's account, which is driven by "governance".

apps/aecore/src/aec_mining.erl
zp-sd: Amount is part of coinbase tx. We simply don't need state trees to create coinbase tx. Amount is taken from simplified governance atm.

apps/aecore/src/txs/aec_tx.erl
zp-sd: I believe that the fact that txs got applied on state tree without failures should be verified by checking hash of state tree (`#block.root_hash` field) when absorbing a block. If a tx cannot be applied, hash simply will not match. And checking `root_hash` is the last thing to be implemented for `PT-#152481000 Validate received block`. I'm not sure if separating this function brings much value then, let me think about it.

apps/aecore/src/txs/aec_tx.erl
zp-sd: There is a check on coinbase tx coint in https://github.com/aeternity/epoch/blob/master/apps/aecore/src/aec_blocks.erl#L246 (it is executed when accepting a block from the network), so we may explicitly expect 1 coinbase tx there, wdyt?

lucafavatella: > There is a check on coinbase tx coint in https://github.com/aeternity/epoch/blob/master/apps/aecore/src/aec_blocks.erl#L246 (it is executed when accepting a block from the network), so we may explicitly expect 1 coinbase tx there, wdyt?

If that function is not called on genesis (I did not check, seems out of scope for this PR), check of equal to 1 is ok.

Though wrt mining, I did not see validation of coinbase tx on block candidate (please let me know if I missed it).

rebar.config
zp-sd: This depends on https://github.com/aeternity/epoch/pull/325 and https://github.com/aeternity/tx_xform/pull/3.

zp-sd:
@lucafavatella Regarding `Review impact of silently discarding txs in apply_signed`: I think this is what needs to be done. Imagine the situation that we have a signed tx in our mempool with either invalid signature or failing `sample_tx:check/3` checks. That way, with current implementation, `apply_signed` returns an error and miner is never able to start mining, as creation of a block candidate fails!

I have added a check that newly created block for mining contains coinbase tx, and also added a comment in `aec_coinbase_tx` regarding adding transaction fees to miner's account.

apps/aecore/src/txs/aec_coinbase_tx.erl
```diff
@@ -28,6 +28,9 @@ check(#coinbase_tx{account = AccountPubkey}, Trees0, Height) ->
             Error
     end.
 
+%% Only aec_governance:block_mine_reward() is granted to miner's account here.
+%% Amount from all the fees of transactions included in the block
+%% is added to miner's account in aec_tx:apply_signed/3.
```
lucafavatella: Thanks.

apps/aecore/src/aec_mining.erl
```diff
@@ -17,17 +17,18 @@
...
     Trees = aec_blocks:trees(LastBlock),
-    case get_txs_to_mine(Trees) of
+    case get_txs_to_mine() of
         {error, _} = Error ->
             Error;
         {ok, Txs} ->
-            case aec_blocks:new(LastBlock, Txs, Trees) of
-                {ok, Block0} ->
+            Block0 = aec_blocks:new(LastBlock, Txs, Trees),
+            case aec_blocks:cointains_coinbase_tx(Block0) of
```
lucafavatella: Thanks.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -527,6 +527,9 @@ update_top_block_hash(Hash, State) ->
 apply_node_transactions([Node|Left], Trees, State) ->
     Txs = aec_blocks:txs(Node#node.content),
     Height = node_height(Node),
+    %% TODO: verify that root hash of state tree after applying transactions
```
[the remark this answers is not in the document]
zp-sd: Well, actually, based on our `blocks.hrl`:
```
root_hash = <<0:?STATE_HASH_BYTES/unit:8>> :: state_hash(), % Hash of all state Merkle trees included in #block.trees
```
so we should check `root_hash` as a hash of all state trees.

Transactions' hash is stored in 
```
txs_hash = <<0:?TXS_HASH_BYTES/unit:8>> :: txs_hash(),
```
and this is already being verified as a part of block verification feature, when a block is received from the network.

PR #323, opened 2017-11-06
Transaction broadcast and changes in verification/serialization

uwiger:
- using msgpack to serialize array of maps for signing
- rewrote mempool to use ets tables
- unit tests currently don't pass due to tx verification errors

fix signature verification

(replaces PR #310)


lucafavatella:
@uwiger This is meant to finish both https://www.pivotaltracker.com/n/projects/2124891/stories/152471683 and https://www.pivotaltracker.com/n/projects/2124891/stories/152481178 , right?

PR #324, opened 2017-11-06
PT-152482389 Push created spend_tx into the mempool

zp-sd:
* [Finishes #152482389]

### DIFF
--- a/apps/aecore/src/aec_next_nonce.erl
+++ b/apps/aecore/src/aec_next_nonce.erl
@@ -1,0 +1,17 @@
+-module(aec_next_nonce).
+
+-export([for_account/0]).
+
+for_account() ->
+    %% TODO: Nonce shall not be determined on accounts state tree only
+    %% To be implemented in PT-152543561
+    case aec_keys:pubkey() of
+        {ok, SenderPubkey} ->
+            {ok, LastBlock} = aec_chain:top(),
+            Trees = aec_blocks:trees(LastBlock),
+            AccountsTrees = aec_trees:accounts(Trees),
+            {ok, Account} = aec_accounts:get(SenderPubkey, AccountsTrees),
+            {ok, aec_accounts:nonce(Account) + 1};
+        {error, key_not_found} = Error ->
+            Error
+    end.
--- a/apps/aehttp/src/aehttp_dispatch_int.erl
+++ b/apps/aehttp/src/aehttp_dispatch_int.erl
@@ -3,10 +3,10 @@
 -export([handle_request/3]).
 
 -spec handle_request(
-    OperationID :: swagger_api:operation_id(),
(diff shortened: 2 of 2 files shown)


apps/aecore/src/aec_next_nonce.erl
```diff
@@ -0,0 +1,17 @@
+-module(aec_next_nonce).
+
+-export([for_account/0]).
+
+for_account() ->
```
sennui: this will need to talk to code in mempool PR

zp-sd: Agreed.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -113,28 +113,26 @@ handle_request('GetAccountBalance', Req, _Context) ->
...
                                #{sender => SenderPubkey,
                                  recipient => RecipientPubkey,
                                  amount => Amount,
                                  fee => Fee,
                                  nonce => Nonce}),
-            %% TODO: sign and push to the mempool
+            {ok, SignedTx} = aec_keys:sign(SpendTx),
+            ok = aec_tx_pool:push(SignedTx),
+            %% TODO: send to peers via external API
+            %% To be done when POST /tx is implemented
```
zp-sd: /cc @uwiger

zp-sd:
@lucafavatella I believe this should be covered in `PT-152481063 Build block proposal`, which is in works right now.

lucafavatella:
> @lucafavatella I believe this should be covered in PT-152481063 Build block proposal, which is in works right now.

Ok. From the code I see [txs picked from mempool](https://github.com/aeternity/epoch/blob/c98342e717fb076dcd8a72cb61d67ac1580860ad/apps/aecore/src/aec_mining.erl#L73) and [then](https://github.com/aeternity/epoch/blob/c98342e717fb076dcd8a72cb61d67ac1580860ad/apps/aecore/src/aec_blocks.erl#L86) checked and [processed](https://github.com/aeternity/epoch/blob/c98342e717fb076dcd8a72cb61d67ac1580860ad/apps/aecore/src/txs/aec_tx.erl#L93) - by calling the [`aec_spend_tx` module](https://github.com/aeternity/epoch/blob/c98342e717fb076dcd8a72cb61d67ac1580860ad/apps/aecore/src/txs/aec_spend_tx.erl).

The only thing I am not sure about is height field of account.

zp-sd:
@lucafavatella Ticket filed `PT-152678722 Investigate whether storing block height is needed in state strees`.

lucafavatella:
> Ticket filed PT-152678722 Investigate whether storing block height is needed in state strees.

Thanks.

PR #331, opened 2017-11-08
Transaction broadcast and changes in verification/serialization (new)

uwiger:
Replaces #323 

* using msgpack to serialize array of maps for signing
* rewrote mempool to use ets tables
* improved unit tests somewhat
* made multi-node test skip following tests on early failure
* fix signature verification


apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -74,15 +75,20 @@ handle_request('GetBlockByHash' = _Method, Req, _Context) ->
...
                       cleanup_genesis(aec_blocks:serialize_to_map(Block)),
+                    lager:debug("Resp = ~p", [Resp]),
                     {200, [], Resp};
                 {error, {block_not_found, _}} ->
                     {404, [], #{reason => <<"Block not found">>}}
             end
     end;
 
+handle_request('GetTxs', _Req, _Context) ->
+    error(nyi);
```
lucafavatella: Is there a ticket tracking implementation of this API?
(I am concerned that this API would allow a small request to receive a huge response.)

apps/aecore/src/aec_tx_sign.erl
```diff
@@ -11,40 +12,45 @@
...
+-spec deserialize(list()) -> #signed_tx{}.
+deserialize([?SIG_TX_TYPE, ?SIG_TX_VSN, TxSer, Sigs]) ->
+    Tx = aec_tx:deserialize(TxSer),
+    #signed_tx{data = Tx, signatures = Sigs}.
 
 %% deterministic canonical serialization.
 -spec serialize_to_binary(#signed_tx{}) -> binary().
 serialize_to_binary(#signed_tx{} = SignedTx) ->
-    jsx:encode(serialize(SignedTx)).
+    msgpack:pack(serialize(SignedTx)).
```
lucafavatella: `aec_tx_sign:serialize_to_binary/1` is used by `aec_blocks:new/3` (via `aec_tx_trees`) for computing [block header `txs_hash` i.e. the "Root hash of the Merkle tree of transactions in the block"](https://github.com/aeternity/epoch/wiki/Block-Header) hence needs to be deterministic and documented (as other nodes in the network need to be able to compute state txs Merkle tree resulting in same root hash).

Could you please:
* Whether this line results in deterministic binary (I understand `jsx:encode(serialize(SignedTx))` was not as JSON can contain an unspecified number of newlines/whitespaces). If out of scope for this PR (as I think it is out of scope) please ensure a ticket is present;
* Open ticket tracking integration of wiki with this root hash computation (i.e. msgpack).

apps/aehttp/src/aehttp_dispatch_ext.erl
uwiger: That's an issue I brought up earlier, put then decided to punt on. A problem with a chunking request (which could be added via an optional parameter and a continuation method) is that we'd need to define a key hash value for transactions in the pool. This, in its turn, depends on the sorting algorithm, which @sennui was working on in parallel.

But I agree that there should be a ticket. It's just been almost impossible to get this merged since the task was quite broad to begin with. I suggest we create that ticket and deal with it separately.

apps/aecore/src/aec_tx_sign.erl
uwiger: As far as I've been able to determine, `msgpack` does produce deterministic binaries. It's a byte-coded protocol, so it's not affected by whitespace or formatting issues.

apps/aecore/test/aec_miner_tests.erl
```diff
@@ -32,30 +32,29 @@ miner_test_() ->
...
-             lists:foreach(
-               fun(F) ->
-                       AbsF = filename:absname_join(TmpKeysDir, F),
-                       {ok, _} = {file:delete(AbsF), {F, AbsF}}
-               end,
-               KeyFiles),
-             ok = file:del_dir(TmpKeysDir),
+             ok = application:stop(gproc),
+             aec_test_utils:aec_keys_cleanup(TmpKeysDir),
+             %% {ok, KeyFiles} = file:list_dir(TmpKeysDir),
```
lucafavatella: This can be deleted (no need to update PR though).

uwiger: Agreed.

uwiger: There's an attempt to delet `TmpKeysDir` on line 61 as well (which likely never worked to begin with since it uses `file:delete/1` on a directory) ...

apps/aehttp/src/aehttp_dispatch_ext.erl
uwiger: Sorry, it's not just a question of the chunking (or not), but the fact that it isn't implemented at all. :)
My plan was to address that as soon as I get this merged.

lucafavatella: Yes, no need to implement in this branch. Thanks.

apps/aecore/src/aec_tx_sign.erl
lucafavatella: Thanks

apps/aecore/src/aec_keys.erl
```diff
@@ -165,10 +170,28 @@ init([Password, KeysDir]) ->
...
+            Signature =
+                crypto:sign(
+                  Algo, Digest, Bin, [PrivKey, crypto:ec_curve(Curve)]),
+            {reply, {ok, #signed_tx{data = Term,
+                                    signatures = [Signature]}}, State}
+    end;
+handle_call({verify, Sigs, Term}, _From, State) ->
+    Signers = aec_tx:signers(Term),
+    Bin = aec_tx:serialize_to_binary(Term),
+    Res = lists:any(fun(Sig) ->
```
lucafavatella: I understand constraints shall be:
* A. Number of signatures (Sigs) shall be the same as number or signers (Signers);
* B. All signatures shall verify with any one of the signers;
* C. Each signature shall verify with a distinct signer.

Because signatures do not include the pubkey, and because of constraint C, it seems a sequential function would be most readable. (I say this as I do not fully understand this implementation respects these constraints, and I have not reviewed tests yet - that may clarify).

uwiger: Yes, I was pondering whether to make it more strict. This should perhaps be another ticket.

apps/aecore/src/aec_tx_sign.erl
```diff
@@ -11,40 +12,45 @@
...
+
+signatures(#signed_tx{signatures = Sigs}) ->
+    Sigs.
 
 verify(failed_tx) ->
-    %% TODO: Verify signed txs
     {error, verification_failed};
-verify(_Tx) ->
-    ok.
+verify(#signed_tx{data = Data, signatures = Sigs}) ->
```
lucafavatella: I see `aec_tx_sign:verify` is called:
* When inserting tx in mempool;
* When building block candidate.

I believe it shall be called also when receiving block - but that is meant to be separate ticket I understand. We shall ensure this is tracked as ticket.

----

I understand all current transactions (coinbase and spend) need to be signed only by one party. When a tx requiring signatures from more than a party (if any such tx) will be implemented, this may require change (e.g. block shall contain only tx with 2 signature, while tx in mempool may have signature of only the originating party for enabling broadcast).
* Considering tx signed with only one party in this PR is ok - no changes needed;
* I suggest that a new ticket (or PR following up this) tracks hardcoding the expectation that tx has exactly one signature, so that it is clear that only txs with one signature is supported.

apps/aecore/src/aec_keys.erl
uwiger: Created PT ticket #152714446

apps/aehttp/src/aehttp_dispatch_ext.erl
uwiger: Created PT ticket #152714484 regarding the chunking.

apps/aecore/src/aec_keys.erl
uwiger: Created PT ticket #152714484 regarding tracking the implementation of the Tx sync method

PR #336, opened 2017-11-09
Added test to check for missing applications + updated aecore.app.src

ThomasArts:
The aecore.app.src had missing applications.
I created a general test for applications to make sure this does not happen.

Updated the list and checked both that the test was strong enough to find problems as well as that it actually passes now.


apps/aecore/test/aecore_tests.erl
```diff
@@ -0,0 +1,22 @@
...
+%% We must be able to start an application when all applications it
+%% depends upon are started.
+application_test() ->
+  App = aecore,
+  application:load(App),
+  application:stop(App),
+  application:set_env(aecore, password, "Thisisweird"),
+  {ok, Deps} = application:get_key(App, applications), 
+  AlreadyRunning = [ Name || {Name, _,_} <- application:which_applications() ],
+  [ ?assertEqual(ok, application:ensure_started(Dep)) || Dep <- Deps ],
```
uwiger: should not `application:ensure_all_started(Dep)` work here?

ThomasArts: Yes, the assert is not needed for the test to work, but for the output to be shown when there is no match. You want to know which error message you got back from ensure_started...

```erlang
**error:{assertEqual,[{module,aecore_tests},
              {line,16},
              {expression,"application : ensure_started ( Dep )"},
              {expected,ok},
              {value,{error,{not_started,goldrush}}}]}
```

ThomasArts: I now see the difference. With ensure_all_started you recurse and find what is needed.
That means that in the .app file you only need those applications that are on top level, if you normally use this all_started primitive.

But how do relx and other tools find the dependencies, also recursively?

PR #339, opened 2017-11-10
use select instead of first/next iteration

uwiger:
[Finishes #152745231]

### DIFF
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -117,10 +117,10 @@ handle_call({delete, Txs}, _From, #state{db = Mempool} = State) ->
     [pool_db_delete(Mempool, pool_db_key(Tx)) || Tx <- Txs],
     {reply, ok, State};
 handle_call({peek, MaxNumberOfTxs}, _From, #state{db = Mempool} = State)
-  when is_integer(MaxNumberOfTxs), MaxNumberOfTxs > 0 ->
-    {ok, Txs} = pool_db_peek(Mempool, MaxNumberOfTxs),
-    Reply = {ok, Txs},
-    {reply, Reply, State};
+  when is_integer(MaxNumberOfTxs), MaxNumberOfTxs >= 0;
+       MaxNumberOfTxs =:= infinity ->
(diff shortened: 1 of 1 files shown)


apps/aecore/src/aec_tx_pool.erl
```diff
@@ -176,21 +176,14 @@ exclude_coinbase({Fee, Origin, Nonce}) ->
 pool_db_open() ->
     {ok, ets:new(?MEMPOOL, [ordered_set, public, named_table])}.
 
--spec pool_db_peek(pool_db(), MaxNumber::pos_integer()) ->
+-spec pool_db_peek(pool_db(), MaxNumber::pos_integer() | infinity) ->
                           {ok, [pool_db_value()]}.
```
gorillainduction: Spec is now wrong. Should return ` [pool_db_value()]`

ThomasArts: You add an extra {ok, ...} in the return type is that intentional?

```erlang
when is_integer(MaxNumberOfTxs), MaxNumberOfTxs >= 0;
       MaxNumberOfTxs =:= infinity ->
    Txs = pool_db_peek(Mempool, MaxNumberOfTxs),
    {reply, {ok, Txs}, State};
```

Instead of what I would expect:
```erlang
when is_integer(MaxNumberOfTxs), MaxNumberOfTxs >= 0;
       MaxNumberOfTxs =:= infinity ->
    Txs = pool_db_peek(Mempool, MaxNumberOfTxs),
    {reply, Txs, State};
```

Txs already contains {ok, [pool_db_value()]}

Ah, no, same remark as above, the spec is wrong, made me believe the return type was wrong.

uwiger: true

uwiger: arguably, the API should be changed to avoid the wrapper, as the call should always return a list (possibly empty)

gorillainduction: There are a lot of these unnecessary wrappers in the code base. Do remove it from this place ;-)

PR #340, opened 2017-11-10
make aeu_env & user config handl more modular

uwiger:
[Finishes PT #152745563]

Adds `aeu_env:get_env/[2,3]` support for hierarchical keys, where for example in an env config:

```erlang
  { aehttp, [
      {swagger_port_external, 3013},
      {internal, [
          {swagger_port, 3113},
          {websocket, [ {port, 3114},
                        {handlers, 100},
                        {tasks, 200}
                      ]}
          ]}
```

one can call `aeu_env:get_env(aehttp, [internal, websocket, port], ?DEFAULT_PORT)`

Also, the contents of any `epoch.{yaml,json}` file is normalized and stored as an `aeutils` env, `'$user_config'`, which can be queried similarly using `aeu_env:user_config(Key)`, or `aeu_env:user_config(Key, Default)`.

Two ways to check the user config are illustrated:
* Add a setup hook (mode: 'normal', phaseno > 100) which checks user_config and if necessary updates the application env.
* Dynamically check the user config while checking the normal env (used in aehttp_app)

### DIFF
--- a/apps/aecore/src/aec_peers.erl
+++ b/apps/aecore/src/aec_peers.erl
@@ -26,6 +26,8 @@
          get_local_peer_uri/0,
          update_last_seen/1]).
 
+-export([check_env/0]).
+
 %% gen_server callbacks
 -export([start_link/0, init/1, handle_call/3, handle_cast/2,
          handle_info/2, terminate/2, code_change/3]).
@@ -193,6 +195,18 @@ get_local_peer_uri() ->
 -spec update_last_seen(uri()) -> ok.
 update_last_seen(Uri) ->
     gen_server:cast(?MODULE, {update_last_seen, uri(Uri), timestamp()}).
+
+%%------------------------------------------------------------------------------
+%% Check user-provided environment
+%%------------------------------------------------------------------------------
+check_env() ->
+    case aeu_env:user_config(<<"peers">>) of
+        {ok, Peers0} when is_list(Peers0) ->
+            Peers = [binary_to_list(P) || P <- Peers0],
+            application:set_env(aecore, peers, Peers);
+        undefined ->
+            ok
+    end.
 
 %%%=============================================================================
 %%% gen_server functions
--- a/apps/aeutils/src/aeu_env.erl
+++ b/apps/aeutils/src/aeu_env.erl
@@ -7,13 +7,48 @@
 %%%-------------------------------------------------------------------
 -module(aeu_env).
 
--export([get_env/3]).
+-export([get_env/2, get_env/3]).
 
+-export([user_config/0, user_config/1]).
 -export([read_config/0]).
 
+-spec get_env(atom(), atom() | list()) -> undefined | {ok, any()}.
+get_env(App, [H|T]) ->
+    case setup:get_env(App, H) of
+        {ok, V} ->
+            get_env_l(T, V);
+        undefined ->
+            undefined
+    end;
+get_env(App, K) when is_atom(K) ->
+    setup:get_env(App, K).
 
 get_env(App, K, Default) ->
-    setup:get_env(App, K, Default).
+    case get_env(App, K) of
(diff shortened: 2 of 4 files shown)


apps/aecore/src/aec_peers.erl
```diff
@@ -194,6 +196,18 @@ get_local_peer_uri() ->
...
     gen_server:cast(?MODULE, {update_last_seen, uri(Uri), timestamp()}).
 
+%%------------------------------------------------------------------------------
+%% Check user-provided environment
+%%------------------------------------------------------------------------------
+check_env() ->
+    case aeu_env:user_config(<<"peers">>) of
+        {ok, Peers0} when is_list(Peers0) ->
+            Peers = [binary_to_list(P) || P <- Peers0],
+            application:set_env(aecore, peers, Peers);
```
lucafavatella: I see this overwrites values in sys.config. Needs mention in release notes.

uwiger: Yes, that is basically the point. I agree that it needs to be documented.

apps/aeutils/src/aeu_env.erl
```diff
@@ -7,13 +7,48 @@
 %%%-------------------------------------------------------------------
 -module(aeu_env).
 
--export([get_env/3]).
+-export([get_env/2, get_env/3]).
 
+-export([user_config/0, user_config/1]).
```
lucafavatella: I see that `get_env/{2,3}` and `user_config/1` are two separate sets of configs. Ok. Comment (not in this PR) would help.

apps/aecore/src/aec_peers.erl
lucafavatella: So this code is reading from user config (e.g. `~/.epoch/NODENAME`) and overwriting with it (in memory) whatever was in `application:get_env` for that key (e.g. from `sys.config`) so that rest of the code in the app can use only plain `application:get_env` for querying.

We shall standardize on how to use these capabilities consistently across the apps - and this is a good start! No need to change this PR.

apps/aeutils/src/aeu_env.erl
```diff
@@ -7,13 +7,48 @@
...
+-spec get_env(atom(), atom() | list()) -> undefined | {ok, any()}.
+get_env(App, [H|T]) ->
+    case setup:get_env(App, H) of
+        {ok, V} ->
+            get_env_l(T, V);
+        undefined ->
+            undefined
+    end;
+get_env(App, K) when is_atom(K) ->
+    setup:get_env(App, K).
```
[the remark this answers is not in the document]
uwiger: `setup:get_env/2` is actually considerably more powerful than `application:get_env/2`, but on 'normal' environment data, it behaves exactly the same. Perhaps the most trivially useful addition in `setup:get_env/2` is that it supports variable expansion in path names, of predefined variables such as `"$APP"` and `"$PRIV_DIR"`, but there is more, as documented in the [setup docs](https://github.com/uwiger/setup/blob/master/doc/setup.md).

apps/aeutils/src/aeu_env.erl
uwiger: yes, the config data from the user config files is stored as a "private" environment variable, partly to simplify debugging (you can easily inspect the structure to see that it conforms to expectation), and partly to also support efficient dynamic checking (such as e.g. in `aehttp_app`. The `user_config/[0,1]` functions are meant to keep the name of the "private namespace" from leaking out.

apps/aecore/src/aec_peers.erl
sennui: 1. we need to definitely let devops know
2. that's usual approach for consumer software, so I totally support it
3. with version upgrades we will need to introduce some sort of validation, to discard unsupported config from local config

PR #344, opened 2017-11-10
PT-152642152 Test persistence [Finishes #152642152]

velzevur:
Test for use case
```
Bob should not have to download the whole blockchain every time

Bob's downloaded blockchain should persist between restarts.
He should only download updates to his blockchain when his node starts.
```
It is dumbed down to _Bob's blocks persist durring restarts_
For this I needed a premined chain (that would be `py/tests/integration/data/bchain.txt`). The module `chain_downloader.py` is responsible for downloading it to a file, loading it from a file and searching in the data structure.

I've also modified a little bit the setup.yaml that every test has its own configurable space.
Another small change is that tests don't have the node hardcoded in the test but instead use configurable alisases ("bob" or "alice")


py/tests/integration/common.py
```diff
@@ -46,57 +47,36 @@ def node_online(api):
...
+        else:
+            config_prefix =  'ERL_FLAGS="-config `pwd`/' + config_filename + '" ' 
+    os.system(config_prefix + "make " + name + "-start")
     api = external_api(name)
-    wait(lambda: node_online(api), timeout_seconds=3, sleep_seconds=0.5)
+    wait(lambda: node_online(api), timeout_seconds=10, sleep_seconds=0.5)
 
 def stop_node(name):
     os.system("make " + name + "-stop")
+    time.sleep(3)
```
gorillainduction: Use a wait here instead

velzevur: well yes, but `wait` expects a predicate - what should we be waiting for? I don't think `ps -fea| grep epoch | wc -l ` is good way of validation that the node is dead

gorillainduction: I guess you could check that it doesn't respond to ping, but that might succeed too early.

Nevermind. Leave it as is for now.

py/tests/integration/test_use_cases.py
```diff
@@ -1,39 +1,97 @@
...
+    """
+    Test persistence:
+    Bob's downloaded blockchain should persist between restarts. He should
+    only download updates to his blockchain when his node starts.
+    """
+    test_settings = settings["test_persistence"]
+    chain_data_file = test_settings["chain_data"]
+
+    # prepare a dir to hold the config and DB files 
+    root_dir = tempfile.mkdtemp()
```
sennui: the erlang node expects those directories? (it is not creating them?)

velzevur: We need a place to put the `.chain` - containing the DB files; I didn't want to interfere with any data currently in the build, so I've created a temporary one

sennui: ok, yes, nev

PR #346, opened 2017-11-10
Bump PoW nonce only in aec miner - rather than in both aec_miner and PoW modules

lucafavatella:
Preparatory PR for PT-152518745.
Cc @akorosmezey 

Context of each commit (in addition to log message in each commit):
* f38d9be - Wrt `mine_block_from_genesis_test_` deletion, that test was from the time we had no acceptance test testing mining from genesis. While performing the changes for nonce bump only by aec_miner, I reconsidered the test. Hence the deletion;
* a3f3ae9 - This is the main point of this PR;
* bd9e594 - This increases code coverage for PoW modules. This commit clarifies responsibilities of PoW modules on single nonce, and also increases testing coverage for successive PR for PT-152518745.

Known issues:
* Removal of nonce increase in PoW module for the sake of simplification introduced attempting only one nonce per block candidate in case of `fetch_new_txs_from_pool_during_mining = true`. This is proposed to be addressed separately from this PR, and is tracked as https://www.pivotaltracker.com/story/show/152796665


lucafavatella:
After this PR is merged, main PR addressing PT-152518745 can be opened.

lucafavatella:
Of the two builds Travis CI starts (one for push and one for PR), one got tests [passing](https://travis-ci.org/aeternity/epoch/jobs/300142120) and one [failing](https://travis-ci.org/aeternity/epoch/jobs/300139045).

The [failing one reports](https://travis-ci.org/aeternity/epoch/jobs/300139045#L1455):
```
%%% aec_eunit_SUITE ==> aec_miner_tests (group eunit): FAILED
%%% aec_eunit_SUITE ==> errors_in_lager_log
```

This may be related to recent change to fail if test logs lager:error, but such changes are in this branch already (as only PR 340 is missing here) so the failure shall happen in both cases. Needs more investigation in order to understand whether to whitelist (as test may need to test error code path). Trying to reproduce locally. 
Cc @akorosmezey

lucafavatella:
> Trying to reproduce locally.

I see locally in branch (bd9e594).

lucafavatella:
I opened:
* PR #348 for removing aec_keys -related intermittent test failures - locally;
  * PR 348 is good to review and merge, though if not merged the CI for this PR 346 can be restarted.
* PR #349 for relaxing constraints on no error logs in test - locally.
  * PR 349 is not ready to merge because debatable. I am investigating lager:error calls in this PR 346.

lucafavatella:
I re-cleaned working dir and ran tests and it did not occur. So I am re-running the [CI job `test` that failed on one of the two builds](https://travis-ci.org/aeternity/epoch/jobs/300139045).

lucafavatella:
... and CI passed. Ready to review.

lucafavatella:
I see two reviewers approved. I locally merged this with recent master, and `make test` and `make dialyzer` passed. Merging.

PR #347, opened 2017-11-10
restore SpendTx [finishes #152755356]

uwiger:
During a merge, and while changing the external representation of transactions in swagger, the SpendTx definition was mistakenly removed. However, it's used in an internal method to create a spend transaction, and was perfectly fine for that purpose.

### DIFF
--- a/apps/aehttp/src/aehttp_dispatch_ext.erl
+++ b/apps/aehttp/src/aehttp_dispatch_ext.erl
@@ -87,7 +87,9 @@ handle_request('GetBlockByHash' = _Method, Req, _Context) ->
     end;
 
 handle_request('GetTxs', _Req, _Context) ->
-    error(nyi);
+    {ok, Txs0} = aec_tx_pool:peek(infinity),
+    Txs = [base64:encode(aec_tx_sign:serialize_to_binary(T)) || T <- Txs0],
+    {200, [], Txs};
 
 handle_request('PostBlock', Req, _Context) ->
     SerializedBlock = add_missing_to_genesis_block(maps:get('Block', Req)),
@@ -102,7 +104,8 @@ handle_request('PostBlock', Req, _Context) ->
 
 handle_request('PostTx', Req, _Context) ->
     SerializedTx = maps:get('SignedTx', Req),
-    SignedTx = aec_tx_sign:deserialize_from_binary(SerializedTx),
+    SignedTx = aec_tx_sign:deserialize_from_binary(
+                 base64:decode(SerializedTx)),
     aec_tx_pool:push(SignedTx, tx_received),
     {200, [], #{}};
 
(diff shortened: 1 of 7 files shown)


apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -102,7 +104,8 @@ handle_request('PostBlock', Req, _Context) ->
 
 handle_request('PostTx', Req, _Context) ->
     SerializedTx = maps:get('SignedTx', Req),
-    SignedTx = aec_tx_sign:deserialize_from_binary(SerializedTx),
+    SignedTx = aec_tx_sign:deserialize_from_binary(
```
sennui: is base64 encoding specific to HTTP protocol? Maybe we should lighten the controller and move it one of core modules?

apps/aeutils/src/aeu_requests.erl
```diff
@@ -85,7 +87,7 @@ send_block(Peer, Block) ->
 -spec send_tx(aec_peers:peer(), aec_tx:signed_tx()) -> response(ok).
 send_tx(Peer, SignedTx) ->
     Uri = aec_peers:uri(Peer),
-    TxSerialized = aec_tx_sign:serialize_to_binary(SignedTx),
+    TxSerialized = base64:encode(aec_tx_sign:serialize_to_binary(SignedTx)),
```
sennui: similar here re base64. Should we introduce e.g. aec_tx_sign:serialize_for_network and make it transparent for Send-Receive handlers?

apps/aeutils/src/aeu_requests.erl
```diff
@@ -94,6 +96,22 @@ send_tx(Peer, SignedTx) ->
             Error
     end.
 
+%% NOTE that this is part of the internal API, thus the standard peer
+%% uris will not work
+new_spend_tx(IntPeer, #{sender_pubkey := Ks,
```
sennui: who is calling this function? Is aeu_requests.erl expected to gather more transaction specific code? Maybe we should contain it in transaction specific module?

apps/aehttp/src/aehttp_dispatch_ext.erl
uwiger: It's possible that the base64 encoding can be left out entirely. The JSON encoders/decoders should be able to handle a raw binary. But encoded JSON with escaped binaries is pretty hard on the eyes.

Strictly speaking it's not the network that justifies using base64, but rather more readable JSON.

apps/aeutils/src/aeu_requests.erl
uwiger: It could well be placed elsewhere, but does reuse the support code in this module.
I assume it will primarily be used for testing purposes.

PR #348, opened 2017-11-10
Create key pair and account when node starts (for test stability)

lucafavatella:
... by guaranteeing that if `aec_keys` service started, key pair has
been created.

This PR is the implementation of the minimal portion of PT-152483887 for test stability.

----

As the `aec_keys`-related test issues were intermittent, you may consider including this in your local branch for an hour or so before approving.

### DIFF
--- a/apps/aecore/src/aec_keys.erl
+++ b/apps/aecore/src/aec_keys.erl
@@ -43,7 +43,7 @@
 -record(state, {
           pub       :: undefined | binary(),
           priv      :: undefined | binary(),
-          pass      :: undefined | binary(),
+          pass      :: password(),
           keys_dir  :: undefined | binary(),
           pub_file  :: undefined | binary(),
           priv_file :: undefined | binary(),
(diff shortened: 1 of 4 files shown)


ThomasArts:
This looks good. If I spotted correctly, there is a subtle change that if password is not in the environment we will get a crashing init (which is fine), whereas before the default was taken and things were silently not working.

lucafavatella:
@sennui @wagerlabs I see two reviewers approved ( @akorosmezey and @ThomasArts - I count Akos too as this PR is not Cuckoo-related) but I am unable to merge because of Travis CI job `package` on OSX not starting. As such job is simply running [`./rebar3 as prod tar`](https://github.com/aeternity/epoch/blob/8924a015698feac3d0e9ec392579c3a95ba14736/ci#L61) and the same `package` job passed on Ubuntu, I am considering merging without waiting for CI.

ThomasArts:
You can restart CI. It’s better to have some tests run publicly to not break it for colleagues.

(I think of dependencies on your machine that travis spots)

Regards
Thomas

PR #353, opened 2017-11-13
Pad private keys before encryption

UlfNorell:
crypto:generate_keys/2 doesn't always give you a 32 byte binary,
it only gives you as many bytes as are required to fit the key. If we
get a shorter key we need to pad it to not crash the block encryption.

### DIFF
--- a/apps/aecore/src/aec_keys.erl
+++ b/apps/aecore/src/aec_keys.erl
@@ -302,7 +302,7 @@ hash(Bin) ->
 %% INFO: keep separate APIs and encrypt both priv & pub to protect external HDs
 %%       (there is known atack vector using master pub)
 encrypt_privkey(Password, Bin) ->
-    crypto:block_encrypt(aes_ecb, hash(Password),  Bin).
+    crypto:block_encrypt(aes_ecb, hash(Password),  pad_privkey(Bin)).
 
 encrypt_pubkey(Password, Bin) ->
     %% TODO: is it safe to use 0s as padding? Consider moving to stream encryption API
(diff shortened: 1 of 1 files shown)


apps/aecore/src/aec_keys.erl
```diff
@@ -319,6 +319,12 @@ padding128(Bin) ->
     Pad = 128 - size(Bin),
     <<Bin/binary, 0:(Pad*8)>>.
 
+%% crypto:generate_keys/2 gives you a binary with as many bytes as are needed to fit the
+%% private key. It does not pad with zeros.
+pad_privkey(Bin) ->
+    Pad = 32 - size(Bin),
```
lucafavatella: May I suggest using macro `PRIV_SIZE` ?

UlfNorell: Ah yes. That makes sense.

apps/aecore/src/aec_keys.erl
```diff
@@ -302,7 +302,7 @@ hash(Bin) ->
 %% INFO: keep separate APIs and encrypt both priv & pub to protect external HDs
 %%       (there is known atack vector using master pub)
 encrypt_privkey(Password, Bin) ->
```
lucafavatella: Hardcoding expectation on return value of `decrypt_privkey` would help readability.

UlfNorell: I'm not sure what you mean (did you mean `encrypt_privkey`?). The problem is that `crypto:block_encrypt/3` crashes if the key is too short and otherwise it returns a binary.

apps/aecore/src/aec_keys.erl
```diff
@@ -319,6 +319,12 @@ padding128(Bin) ->
     Pad = 128 - size(Bin),
     <<Bin/binary, 0:(Pad*8)>>.
```
sennui: please adjust padding here and standarize on leading zeros

UlfNorell:
> Which case is this change addressing? Fetching keys from disk? (I would need this for approving this PR.)

No, the problem is when generating a fresh key-pair (either by calling `aec_keys:new/1` or starting the key server with no keys stored on disk).

apps/aecore/src/aec_keys.erl
UlfNorell: That would complicate `decrypt_pubkey` slightly.

The padding of the public key is a bit different. In the case of the private key we're padding the actual key (since `crypto:generate_key/2` dropped leading zero bytes), but in the public key case we're padding the block to encrypt. I think it makes sense to do the padding on different sides.

sennui: ok, thanks

PR #358, opened 2017-11-14
[Fixes #152817786] Add validation of root hash while traversing the fork

gorillainduction:


apps/aecore/src/aec_chain_state.erl
```diff
@@ -531,6 +534,7 @@ apply_node_transactions([Node|Left], Trees, State) ->
     %% is equal to #block.root_hash.
     %% To be done in scope of "PT-152481000 Validate received block".
     {ok, NewTrees} = aec_tx:apply_signed(Txs, Trees, Height),
```
zp-sd: We may remove the above comment:
```
     %% TODO: verify that root hash of state tree after applying transactions
      %% is equal to #block.root_hash.
      %% To be done in scope of "PT-152481000 Validate received block".
```

PR #359, opened 2017-11-14
Add tx test cases to sync_SUITE + various fixes

uwiger:
[Finishes #152820357]
Errors found and fixed:
- aec_keys must verify key pair read from disk
- aec_keys slightly refactored; check_keypair() fn added
- aec_keys can be configured using epoch.config
- aex_tx_pool:peek() slightly broken
- aeu_requests:transactions() broken
- mempool not updated when mined block written to chain
- add naive (note!) mempool sync to node sync
- aec_sync should always try to write all fetched blocks
- add test cases in sync_SUITE
- tx serialization fixed for swagger compliance
- bug in aeu_env JSON config parsing
- aec_mining pattern-match error when checking mempool for txs

### DIFF
--- a/apps/aeutils/src/aeu_requests.erl
+++ b/apps/aeutils/src/aeu_requests.erl
@@ -61,21 +61,36 @@ block(Peer, Hash) ->
 -spec transactions(aec_peers:peer()) -> response([aec_tx:signed_tx()]).
 transactions(Peer) ->
     Uri = aec_peers:uri(Peer),
-    Response = process_request(Uri, post, "transactions", []),
-    case Response of
-        {ok, Map} ->
-            Txs = maps:get('Transactions', Map, []),
-            {ok, lists:map(
-                   fun(T) ->
-                           aec_tx_sign:deserialize_from_binary(
-                             base64:decode(T))
-                   end, Txs)}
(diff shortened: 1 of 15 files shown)


apps/aeutils/src/aeu_requests.erl
```diff
@@ -61,21 +61,36 @@ block(Peer, Hash) ->
...
+                 lager:error("Error decoding transactions: ~p", [Reason]),
+                 {error, Reason}
+         end
+    catch
+        error:_Reason1 ->
+            lager:error("Wrong response type: ~p", [Response]),
+            {error, wrong_response_type}
     end.
 
+tx_response({ok, #{'Transactions' := Txs}}) -> Txs;
```
sennui: why do we have two different keys in the map envelope

uwiger: It follows from the Swagger specification: `'Transactions'` is an array of transaction objects; `<<"tx">>` is an envelope around the serialized tx binary.

Unfortunately, swagger seems to have problems with more complex array definitions (e.g. 'an array of objects that are themselves arrays of objects'). If someone can achieve a working solution for that, it would be great.

PR #360, opened 2017-11-14
Make it possible to run a single test module from the Makefile

UlfNorell:
For instance,
```
  make test SUITE=apps/aecore/test/aec_eunit TEST=aec_pow_cuckoo_tests
  make eunit TEST=aec_mining_tests
```

`make test` requires a `SUITE` and and optional `TEST` module, `make eunit` just takes
a `TEST` module.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,20 @@
 CORE = rel/epoch/bin/epoch
 VER := $(shell cat VERSION)
 
+CT_TEST_FLAGS =
+EUNIT_TEST_FLAGS =
+ifdef SUITE
+CT_TEST_FLAGS += --suite=$(SUITE)_SUITE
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,16 @@
 CORE = rel/epoch/bin/epoch
 VER := $(shell cat VERSION)
 
+CT_TEST_FLAGS =
+EUNIT_TEST_FLAGS =
+ifdef SUITE
+CT_TEST_FLAGS += --suite=$(SUITE)_SUITE
```
uwiger: It would be sweet to also support `--group` (important for `aecore_sync`)

UlfNorell: Fixed.

PR #361, opened 2017-11-14
Work around intermittent cuckoo cycle failures

UlfNorell:
With 5 parallel threads, the cuckoo cycle miner fails to
find an existing solution around 10% of the time (on my
machine), causing the tests in `aec_pow_cuckoo_tests`
to fail.

The ideal solution to the problem would be to fix the race
conditions in the C++ code, but until that happens the
work around is to call the miner repeatedly (up to 10 times)
until it finds a solution.

### DIFF
--- a/apps/aecore/test/aec_mining_tests.erl
+++ b/apps/aecore/test/aec_mining_tests.erl
@@ -20,6 +20,23 @@
                     145,96,71,82,28,176,23,5,128,17,245,174,170,199,54,248,167,43,185,12,108,91,
                     107,188,126,242,98,36,211,79,105,50,16,124,227,93,228,142,83,163,126,167,206>>).
 
+%% How many times to try to generate solution before giving up.
+%% aec_pow_cuckoo:generate/3 has around a 10% failure rate for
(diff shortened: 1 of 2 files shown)


apps/aecore/test/aec_mining_tests.erl
```diff
@@ -20,6 +20,23 @@
                     145,96,71,82,28,176,23,5,128,17,245,174,170,199,54,248,167,43,185,12,108,91,
                     107,188,126,242,98,36,211,79,105,50,16,124,227,93,228,142,83,163,126,167,206>>).
 
+%% How many times to try to generate solution before giving up.
+%% aec_pow_cuckoo:generate/3 has around a 10% failure rate for
```
zp-sd: how do we know about 10% rate?

UlfNorell: I ran the same problem 10,000 times and counted :).

UlfNorell:
Another option is to wait for #363 and then change the tests to run single threadedly. I filed PT-152877656.

UlfNorell:
These changes are getting integrated with #363 so closing this PR.

PR #362, opened 2017-11-14
Ansible deployment with OpenStack dynamic inventory

dincho:
see README.md for more info

- [x] Re-encrypt travis private SSH key
- [x] OpenStack ENV configuration


.travis.yml
```diff
@@ -1,16 +1,9 @@
-sudo: false
+sudo: required
```
lucafavatella: I need to understand why sudo. It makes CI slower.

ci
```diff
@@ -84,4 +95,21 @@ case "${1:?}"-"${2:?}" in
...
+        # https://docs.travis-ci.com/user/encrypting-files/
+        openssl aes-256-cbc -K $encrypted_a2e547bcd39e_key -iv $encrypted_a2e547bcd39e_iv -in deployment/ansible/files/travis/deploy_key.enc -out /tmp/deploy_rsa -d
+        # ssh agent does not accept world readable keys
+        chmod 600 /tmp/deploy_rsa
+        eval "$(ssh-agent -s)"
+        ssh-add /tmp/deploy_rsa
+
+        BuildDir="${3:?}"
+        DeployEnv="${4:?}"
+        TARBALL="${BuildDir}/_build/prod/rel/epoch/epoch-${Version}.tar.gz"
```
lucafavatella: Let me double check this path.

lucafavatella: (Not necessary for this PR.)

Please do not use this package but rather `"${BuildDir:?}"/_build/prod/rel/epoch/epoch-"${Version:?}"-"${Prefix:?}".tar.gz` as it is what we upload.

If possible also please consider changing https://github.com/aeternity/epoch/pull/362/files#diff-35ea51baf1fe7f0142ad5f950855dde0R74 from `cp -p` to `mv` so to ensure only one copy of the package is around in all parts of the CI scripts.

.travis.yml
dincho: Good question, I spent some time to figure it out too.
Current configuration is also using VM instead of containers, but it looks like travis set it implicitly when it finds "sudo" in the commands, [this line for example](https://github.com/aeternity/epoch/pull/362/files#diff-354f30a63fb0907d4ad57269548329e3L11).

You can check if travis is using VM or container by checking the "Worker Information" in [some of the builds](https://travis-ci.org/aeternity/epoch/jobs/302534493).

```
Worker information
hostname: 80a7a052-5d60-45d0-81f3-5bcc0761362a@1.production-5-worker-org-b-2-gce
version: v3.3.1 https://github.com/travis-ci/worker/tree/1153e5010ef9863bf96264b75fba506109633460
instance: travis-job-629e5d66-b179-435d-a315-78ff29a9db87 travis-ci-amethyst-trusty-1503972832 (via amqp)
startup: 21.383891764s
```

Looking the hostname (GCE) and startup time, it's clear it is using VM not container.
[This documents the expected startup time](https://docs.travis-ci.com/user/reference/overview/#Virtualization-environments).

So I just make it explicit because I moved the scripting in the `ci` script, thus travis no more find this implicit requirement.

While I agree that boot time is much faster for containers, it's not always true that the build summary time would be faster. We can discuss and test this in future, and switch to best option (if possible).

deployment/ansible/deploy.yml
```diff
@@ -0,0 +1,70 @@
...
+          state: directory
+          mode: 0700
+
+    - name: Download epoch package
+      get_url:
+        url: "{{ remote_package }}"
+        dest: "{{ project_root }}/{{ remote_package | basename }}"
+      when: remote_package is defined
+
+    - name: Check if epoch binary exists
```
lucafavatella: I am sure I am missing something - how can this step work before "Extract ..."?

lucafavatella: Extract steps are at lines 43 and 51.

ci
dincho: sure, here is it ccc3ca9

deployment/ansible/deploy.yml
dincho: This actually checks if the package was deployed already and there is bin/epoch, then it execute ping and if "pong" it stops the daemon.

If the binary does not exists, it means "first deploy" and no package and daemon yet, so we don't need the ping and stop commands.

.travis.yml
lucafavatella: Good point. Something to consider in the future. Thanks for the explanation.

ci
lucafavatella: Thanks. Interesting to check CI passes.

uwiger:
FFR, there is now support for a user-provided config file that can be used to set certain parameters.
The config file is read as `epoch` starts, and validated against a JSON-Schema – first iteration [here](https://github.com/aeternity/epoch/blob/master/apps/aeutils/priv/epoch_config_schema.json).

A description (w.i.p.) of how to specify the file, or where to place it for automatic detection, can be found on [this wiki page](https://github.com/aeternity/epoch/wiki/User-provided-configuration). We will add some descriptions of the currently supported parameters.

ci
```diff
@@ -66,4 +77,22 @@ case "${1:?}"-"${2:?}" in
...
+        openssl aes-256-cbc -K $encrypted_a2e547bcd39e_key -iv $encrypted_a2e547bcd39e_iv -in deployment/ansible/files/travis/deploy_key.enc -out /tmp/deploy_rsa -d
+        # ssh agent does not accept world readable keys
+        chmod 600 /tmp/deploy_rsa
+        eval "$(ssh-agent -s)"
+        ssh-add /tmp/deploy_rsa
+
+        BuildDir="${3:?}"
+        DeployEnv="${4:?}"
+        OsLabel=$(package_os_label "${TRAVIS_OS_NAME:?}")
+        TARBALL="${BuildDir:?}"/_build/prod/rel/epoch/epoch-"${Version:?}"-"${OsLabel:?}".tar.gz
```
lucafavatella: At line 62 `cp -p` -> `mv` lost in rebase. No change needed in this PR though.

dincho: Ah, right. It will be overkill to wait travis again

PR #364, opened 2017-11-15
User acceptance test: successfully spending tokens

velzevur:
[Pivotal task](https://www.pivotaltracker.com/story/show/152648914)
```
Alice should be able to send tokens to Bob

Alice should be able to create a spend transaction to send tokens to Bob.
In a controlled environment, Alice should see her transaction appear in the
next block or two that are added to the blockchain. 

Once that happens, Alice should see her account debited and Bob's account
credited with the same number of tokens.

The debit/credit should not happen until the transaction is confirmed, e.g.
included in at least one block in the chain.
```
Test sets up Alice's node with a set of keys (stored in a temp dir). Later on we'd be pushing some premined blocks to her chain that have her account rewarded with all the coinbase rewards. This is how Alice gets some tokens to spend.
Alice's node is started, some blocks are pushed and her balance is validated to have enough tokens to spend. Then a send tokens to Bob transaction is executed (since the node has the keys pair - it can sign it).
We wait for a new block to be mined (this could take a while) and then validate both Alice and Bob's balances.

`py/tests/integration/data/bchain.txt` contains the new premined block chain (it is to be combined with the keys pair)

Example log
```
py.tests.integration.test_spend_tx.test_successful ... 
Node dev1 starting
Alice initial balance is 10
Alice's spend_tx is {'amount': 3, 'fee': 2, 'recipient_pubkey': 'bob_address'}
Transaction sent
Waiting for a next block to be mined
Coinbase reward is 10
Alice's balance (with a coinbase reward) is now 17
Bob's balance is now 3
Node dev1 stopping
```


py/tests/integration/test_spend_tx.py
```diff
@@ -0,0 +1,139 @@
...
+
+    bob_balance0 = {"balance":0} 
+    try:
+        bob_balance0 = alice_api.get_account_balance(pub_key=test_settings["spend_tx"]["bob_pubkey"])
+        assert_equals(bob_balance1.balance, bob_balance0.balance)
+
+    except ApiException as e:
+        assert_equals(e.status, 404) # Bob has no account yet
+    # Alice sends some tokens to Bob 
+    alice_inertnal_api = common.internal_api(alice_node)
```
uwiger: Typo, though consistent with below typo (line 51)

PR #369, opened 2017-11-15
Use 64-bit nonce

akorosmezey:
Needed for 
[PT-152672154](https://www.pivotaltracker.com/story/show/152672154) - 64-bits PoW nonce
[PT-152915267](https://www.pivotaltracker.com/story/show/152915267) - PoW nonce validation


apps/aecore/src/aec_pow.erl
```diff
@@ -131,11 +131,11 @@ target_to_difficulty(Th) ->
 
 -spec pick_nonce() -> integer().
 pick_nonce() ->
-    rand:uniform(?NONCE_RANGE) band 16#7fffffff.
+    rand:uniform(?NONCE_RANGE) band ?MAX_NONCE.
 
 -spec next_nonce(integer()) -> integer().
 next_nonce(N) ->
```
hanssv: If we know that N =< ?MAX_NONCE it is more efficient to write this function as
```erlang
next_nonce(?MAX) -> 0;
next_nonce(N) -> N + 1. 
```
but that might not always be the case...

apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -20,8 +20,13 @@
 -compile([export_all, nowarn_export_all]).
 -include_lib("eunit/include/eunit.hrl").
 -endif.
+-include("pow.hrl").
 
--define(DEFAULT_CUCKOO_ENV, {"mean30", "-t 5", 30}). %% -m 7 caused crash
+-define(DEFAULT_CUCKOO_ENV, {"mean30", "-t 1", 30}). %% -m 7 caused crash
```
hanssv: Is this change intentional? There are issues with running cuckoo miner multi-threaded but I thought we concluded that for the actual miner (running in a separate thread) it will be better to use (5) threads... In some of our tests we'd like to run it single-threaded though!

apps/aecore/test/aec_mining_tests.erl
```diff
@@ -174,4 +175,16 @@ cleanup(_, PoWMod) ->
             ok
     end.
 
+mine(_, _, 0) ->
```
hanssv: Unused code?

lucafavatella: Yes.
Also line 118. I understand this was draft code for actually finding nonce that has a solution for updating test.

apps/aecore/src/aec_pow_sha256.erl
```diff
@@ -46,7 +47,9 @@ generate(Data, Target, Nonce) ->
 -spec verify(Data :: aec_sha256:hashable(), Nonce :: integer(),
              Evd :: aec_pow:pow_evidence(), Target :: aec_pow:sci_int()) ->
                     boolean().
-verify(Data, Nonce, Evd, Target) when Evd == no_value ->
+verify(Data, Nonce, Evd, Target) when Evd == no_value,
+                                      Nonce >= 0,
```
lucafavatella: I see this hardcodes expectation that nonce is in valid range. Thanks.

lucafavatella: This is part of PT-152915267.

lucafavatella:
I see that one CI build passed and one [CI build failed](https://travis-ci.org/aeternity/epoch/jobs/302686519) but I understand it is not related to this PR. I have seen this in other rare occasions.

apps/aecore/src/aec_pow_cuckoo.erl
lucafavatella: @akorosmezey If this is for making tests stable, I believe it is not needed thanks to https://github.com/aeternity/epoch/commit/aba0fda1f7dec68bdd74a8776c9286dbf1662e44 (included in master hence in this PR). Please elaborate.

config/swagger.yaml
```diff
@@ -240,7 +240,7 @@ definitions:
         format: int64
       nonce:
         type: integer
-        format: int64
+        format: uint64
```
lucafavatella: I see this definition brings the HTTP API spec (swagger.yaml) in line with the documentation of the nonce in the header as `u64` in https://github.com/aeternity/epoch/wiki/Block-Header - thanks.

lucafavatella:
I see the [serialization for hash](https://github.com/aeternity/epoch/blob/0e508233e143d94fd361aa116b4584153c69451b/apps/aecore/src/aec_headers.erl#L171) (`<< ... (H#header.nonce):64 ... >>`) is 64 bits [unsigned integer](http://erlang.org/doc/reference_manual/expressions.html#id83099) - ok.

apps/aecore/src/aec_pow_cuckoo.erl
akorosmezey: Thanks for spotting, Fixed in next commit 6a49d96

akorosmezey:
A new PR: #380 is created which is rebased on current master

PR #371, opened 2017-11-16
Fix missing case heads in aec_miner

velzevur:


apps/aecore/src/aec_miner.erl
```diff
@@ -595,6 +601,7 @@ waiting_post_block(Block, State) ->
     int_post_block(Block, State),
     {next_state, waiting_for_keys, State}.
 
+-spec int_post_block(#block{}, #state{}) -> ok | new_state | error.
```
lucafavatella: new_state -> new_top ? Let me double check.

apps/aecore/src/aec_miner.erl
```diff
@@ -577,6 +579,8 @@ configure_post_block(Block, State) ->
         new_top ->
             start_if_auto(State),
             {next_state, configure, State};
+        error ->
+            {next_state, configure, State};
```
lucafavatella: I need to double check if this needs logging here (same for other clauses).

Also I see that int_post_block may return ok for a somehow error case like https://github.com/aeternity/epoch/pull/371/files#diff-18f53debbff20a1bf9efdeb33d180e93R627 . I believe a review of the error cases returned by that function is needed.

Also it may be a good opportunity for refactoring the return value of int_post_block to `{ok, same_top | new_top} | {error, Reason::term()}`

velzevur: I've added a comment for this improvement https://www.pivotaltracker.com/story/show/152807361

PR #374, opened 2017-11-16
[WIP] Reject out-of-range PoW nonces in header validation

akorosmezey:
finishes [PT-152915267](https://www.pivotaltracker.com/n/projects/2124891/stories/152915267) - Validate received block nonce

Checks if nonce is within uint64 limits.
WIP: As a consistent solution, requires PR #369 as uint64 nonces are implemented there.


apps/aecore/src/aec_headers.erl
```diff
@@ -225,14 +225,22 @@ validate_version(_Header) ->
 validate_pow(#header{nonce = Nonce,
                      pow_evidence = Evd,
                      target = Target} = Header) ->
-    Mod = aec_pow:pow_module(),
-    %% Zero nonce and pow_evidence before hashing, as this is how the mined block got hashed.
-    Header1 = Header#header{nonce = 0, pow_evidence = no_value},
-    {ok, HeaderBinary} = serialize_to_binary(Header1),
-    case Mod:verify(HeaderBinary, Nonce, Evd, Target) of
+    case Nonce >= 0 andalso Nonce =< ?MAX_NONCE of
```
lucafavatella: It looks to me the nonce being out of range is a syntactic error in the nonce received from network (or disk). So:
* This function shall enforce nonce range as part of its function clause head guard;
* Nonce out-of-range shall be handled as deserialization error in all points deserializing header or block.

akorosmezey:
Closed and continued in PR #391

PR #375, opened 2017-11-16
Update python integration tests for using the fast miner

velzevur:
Tests used to post premined blocks for populating empty nodes; now they use fast mining instead

### DIFF
--- a/py/tests/integration/setup.yaml
+++ b/py/tests/integration/setup.yaml
@@ -41,16 +41,14 @@ tests: # test specific settings
       nodes:
         bob: dev1
         alice: dev2
-      blocks_to_post: 7
-      chain_data: py/tests/integration/data/bchain.txt
+      blocks_to_mine: 20
 
     test_persistence:
       # Bob's downloaded blockchain should persist between restarts. He should
       # only download updates to his blockchain when his node starts.
       nodes:
         bob: dev1
-      blocks_to_post: 7
-      chain_data: py/tests/integration/data/bchain.txt
+      blocks_to_mine: 7
     test_node_discovery:
       # Assuming Carol's node only knows about Bob upon startup and that Bob's
       # node knows Alice, Carol's node should be able to discover Alice and
@@ -59,8 +57,7 @@ tests: # test specific settings
         alice: dev1
         bob: dev2
         carol: dev3
-      blocks_to_post: 7
-      chain_data: py/tests/integration/data/bchain.txt
+      blocks_to_mine: 7
   test_spend_tx:
     test_successful:
       # Alice should be able to create a spend transaction to send tokens to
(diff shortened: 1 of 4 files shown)


py/tests/integration/setup.yaml
```diff
@@ -41,16 +41,14 @@ tests: # test specific settings
       nodes:
         bob: dev1
         alice: dev2
-      blocks_to_post: 7
-      chain_data: py/tests/integration/data/bchain.txt
```
lucafavatella: @velzevur I am double checking the deletion of the pre-mined chain files.

lucafavatella: It looks still there.
```
$ find . -name bchain.txt
./py/tests/integration/data/bchain.txt
$ ls -lh ./py/tests/integration/data/bchain.txt
-rw-r--r--  1 luca  staff   7.9K 16 Nov 09:34 ./py/tests/integration/data/bchain.txt
```

I see it is used by [`test_ext_api`](https://github.com/aeternity/epoch/pull/375/files#diff-76bf7aa6bd9ae9d90af0e0d0f5156921R36).

lucafavatella:
I see the [package job on OSX failed](https://travis-ci.org/aeternity/epoch/jobs/303101183):

```
Will mine till block 50
[node1] height=2
[node2] height=2
[node3] height=2

[node1] height=2
[node2] height=2
[node3] height=2

...
```

py/tests/integration/setup.yaml
velzevur: yes, for this we will need some premined chain; this functionality is covered in the user acceptance tests; I'll fix the API calls in a subsequent PR

lucafavatella:
> I see the package job on OSX failed:

I am attempting to reproduce locally (by building package and then running release script - as CI) but I am not managing to reproduce test failure.

py/tests/integration/test_use_cases.py
```diff
@@ -176,3 +217,22 @@ def node_peer_url(node_name):
     host = common.node_config(node_name)["host"]
     port = common.node_config(node_name)["ports"]["external_api"]
     return "http://" + host + ":" + str(port) + "/"
+
+def make_fast_mining_config(root_dir, file_name):
+    sys_config = os.path.join(root_dir, file_name)
+    f = open(sys_config, "w")
+    # if autostart is not true - there will be no miner
+    conf ='[{aecore, [{autostart, true},' + \
+                    ' {aec_pow_cuckoo, {"lean16", "-t 5", 16}}]}].'
```
lucafavatella: Does this mean that https://github.com/aeternity/epoch/blob/ccbc61fa6e938661bc0c7d238624f54d4fe41fb0/py/tests/release.py#L20 can be deleted?

PR #377, opened 2017-11-16
User config chain:db_path was not processed [Fixes #152930624]

uwiger:


lucafavatella:
@uwiger Both CI builds are waiting on CI jobs. The rest of jobs passed. OSX jobs may take a while...

PR #380, opened 2017-11-17
Use 64-bit nonce

akorosmezey:
This PR is a successor of PR #369 
Needed for: 
[PT-152672154](https://www.pivotaltracker.com/story/show/152672154) - 64-bits PoW nonce
[PT-152915267](https://www.pivotaltracker.com/story/show/152915267) - PoW nonce validation


lucafavatella:
@akorosmezey Please re-generate pre-mined chain - for the Python tests.
See https://github.com/aeternity/epoch/pull/375#discussion_r151491161 for context. I understand there is make target for regenerating that file.

apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -184,6 +193,43 @@ verify(Hash, Nonce, Soln) ->
...
+    %% We insert our base64-encoded 64-bit Nonce before that (i.e., from byte 65)
+    %% and pass 0 as nonce  for cuckoo. (We use the -x option to write the full buffer.)
+    %% The SHA256 hash is 32 bytes, 44 bytes base64-encoded. There is still room for
+    %% zero characters between header and nonce.
+    %%
+    %% (Base64 encoding: see RFC 3548, Section 3:
+    %% https://tools.ietf.org/html/rfc3548#page-4
+    %% converts every triplet of bytes to 4 characters: from N bytes to 4*ceil(N/3)
+    %% bytes.)
+    NonceStr = base64:encode_to_string(<<Nonce:64/little-unsigned-integer>>),
```
lucafavatella: I understand that choice of endianness in this point is not necessarily related to choice of endianness in serialization of block header. A comment on endianness would help but is not required for this PR.

apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -184,6 +193,43 @@ verify(Hash, Nonce, Soln) ->
...
+%%   hash and a uint64 nonce.
+%% @end
+%%------------------------------------------------------------------------------
+-spec pack_header_and_nonce(binary(), integer()) -> string().
+pack_header_and_nonce(Hash, Nonce) ->
+    %% Cuckoo originally uses 32-bit nonces inserted at the end of its 80-byte buffer.
+    %% This buffer is hashed into the keys used by the main algorithm.
+    %%
+    %% We insert our base64-encoded 64-bit Nonce before that (i.e., from byte 65)
+    %% and pass 0 as nonce  for cuckoo. (We use the -x option to write the full buffer.)
```
lucafavatella: Do we use `-x`?

apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -184,6 +193,43 @@ verify(Hash, Nonce, Soln) ->
...
+%%   Creates the 80-byte Cuckoo buffer (hex encoded) from a base64-encoded
+%%   hash and a uint64 nonce.
+%% @end
+%%------------------------------------------------------------------------------
+-spec pack_header_and_nonce(binary(), integer()) -> string().
+pack_header_and_nonce(Hash, Nonce) ->
+    %% Cuckoo originally uses 32-bit nonces inserted at the end of its 80-byte buffer.
+    %% This buffer is hashed into the keys used by the main algorithm.
+    %%
+    %% We insert our base64-encoded 64-bit Nonce before that (i.e., from byte 65)
```
lucafavatella: Could you please add (deterministic) computation for this 65?

lucafavatella: I mean in comment.

apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -184,6 +193,43 @@ verify(Hash, Nonce, Soln) ->
...
+    %% and pass 0 as nonce  for cuckoo. (We use the -x option to write the full buffer.)
+    %% The SHA256 hash is 32 bytes, 44 bytes base64-encoded. There is still room for
+    %% zero characters between header and nonce.
+    %%
+    %% (Base64 encoding: see RFC 3548, Section 3:
+    %% https://tools.ietf.org/html/rfc3548#page-4
+    %% converts every triplet of bytes to 4 characters: from N bytes to 4*ceil(N/3)
+    %% bytes.)
+    NonceStr = base64:encode_to_string(<<Nonce:64/little-unsigned-integer>>),
+    NonceStart = ?CUCKOO_BUFSIZE - length(NonceStr) - 4,
```
lucafavatella: If `length(NonceStr)` is meant to be always same value, could you please match (i.e. hardcode expectation)? this would help readability a lot.

apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -184,6 +193,43 @@ verify(Hash, Nonce, Soln) ->
...
+    HashStr = base64:encode_to_string(Hash),
+    case length(HashStr) of
```
lucafavatella: Please remove hardcode expectation at this point. I.e. no case as there is actually no conditional I understand - only one code path.
(Feel free to perform some refactoring for clarifying code if you feel that appropriate.)

lucafavatella:
@akorosmezey It would be nice also to include stricter type for PoW nonce as part of PT-152672154.  At the moment nonce is typed in internal representation as [`non_neg_integer()`](https://github.com/aeternity/epoch/blob/af7da1004c956466d280247b932ff0da6cdc9098/apps/aecore/include/blocks.hrl#L43) or `interger()` while actually it shall be `0..?MAX_NONCE`.

lucafavatella:
@akorosmezey I understand that as part of PT-152915267 the following comment on previous PR shall be addressed: https://github.com/aeternity/epoch/pull/374/files#r151498040
i.e. in all the points where the code generates an internal representation of block or header by deserializing (either from network or from disk) that syntactic check on the nonce being in the range 0-MAXNONCE shall be performed, in order not to generate invalid internal representation.

Failure in deserialization from network shall result in either death of the worker of that HTTP/websockt connection or - better - proper HTTP status code  / response returned.

Dailure in deserialization from disk shall - I believe - at least hardcode expectation. Check the rest of deserialization code from disk for understanding whether invalid data on disk is meant to kill the node or log error and discard data.

PR #383, opened 2017-11-17
Test use case: no overspending

velzevur:
```
Bob should not be able to send more tokens than he has

Let's say Bob has 100 tokens. He should not be able to send more than
100 tokens to Alice.

If there's an incoming but unconfirmed deposit into Bob's account then Bob
should not be able to use the incoming tokens until the spend transaction
 they are in is confirmed.
```

Bob's node is started, he mines a couple of blocks and gets some coinbase rewards. Then he tries to send to Alice more token that he has.
We wait for another block to be mined to verify that neither Bob's, nor Alice's balances had changed.

### DIFF
--- a/py/tests/integration/setup.yaml
+++ b/py/tests/integration/setup.yaml
@@ -73,6 +73,22 @@ tests: # test specific settings
         alice: dev1
       blocks_to_mine: 7
       spend_tx:
-        bob_pubkey: bob_address
+        bob_pubkey: Ym9iX2FkZHJlc3M= 
         amount: 42
         fee: 2
+    test_not_enough_tokens:
+    # Bob should not be able to send more tokens than he has
+    #
+    # Let's say Bob has 100 tokens. He should not be able to send more than
+    # 100 tokens to Alice.
+    #
+    # If there's an incoming but unconfirmed deposit into Bob's account then Bob
+    # should not be able to use the incoming tokens until the spend transaction
+    # they are in is confirmed.
+      nodes:
+        bob: dev1
+      blocks_to_mine: 3
+      spend_tx:
+        alice_pubkey: YWxpY2VfYWRkcmVzcw==
+        amount: 1000
+        fee: 20
--- a/py/tests/integration/test_spend_tx.py
+++ b/py/tests/integration/test_spend_tx.py
@@ -29,18 +29,12 @@ def test_successful():
 
     spend_tx_amt = test_settings["spend_tx"]["amount"]
     spend_tx_fee = test_settings["spend_tx"]["fee"]
-    alice_balance = alice_api.get_account_balance()
+    alice_balance = common.get_account_balance(alice_api)
     alice_has_enough_tokens = alice_balance.balance >= spend_tx_amt + spend_tx_fee
     assert_equals(alice_has_enough_tokens, True)
     print("Alice initial balance is " + str(alice_balance.balance))
 
-    bob_balance0 = {"balance":0} 
-    try:
-        bob_balance0 = alice_api.get_account_balance(pub_key=test_settings["spend_tx"]["bob_pubkey"])
-        assert_equals(bob_balance1.balance, bob_balance0.balance)
-
-    except ApiException as e:
-        assert_equals(e.status, 404) # Bob has no account yet
+    bob_balance0 = common.get_account_balance(alice_api, pub_key=test_settings["spend_tx"]["bob_pubkey"])
     # Alice sends some tokens to Bob 
     alice_internal_api = common.internal_api(alice_node)
     spend_tx_obj = SpendTx(
@@ -52,22 +46,19 @@ def test_successful():
     print("Transaction sent")
 
     # ensure Alice balance had not changed
-    alice_balance1 = alice_api.get_account_balance()
+    alice_balance1 = common.get_account_balance(alice_api)
     assert_equals(alice_balance.balance, alice_balance1.balance)
 
     # ensure Bob balance had not changed
-    try:
-        bob_balance1 = alice_api.get_account_balance(pub_key=test_settings["spend_tx"]["bob_pubkey"])
-        assert_equals(bob_balance1.balance, bob_balance0.balance)
-    except ApiException as e:
-        assert_equals(e.status, 404) # Bob has no account yet
+    bob_balance1 = common.get_account_balance(alice_api, pub_key=test_settings["spend_tx"]["bob_pubkey"])
+    assert_equals(bob_balance1.balance, bob_balance0.balance)
 
     # wait for a block to be mined
     print("Waiting for a next block to be mined")
     common.wait_until_height(alice_api, alice_top.height + 1)
 
     alice_new_top = alice_api.get_top()
-    alice_new_balance = alice_api.get_account_balance()
+    alice_new_balance = common.get_account_balance(alice_api)
     
     blocks_mined = alice_new_top.height - alice_top.height
 
@@ -78,7 +69,7 @@ def test_successful():
     expected_balance = alice_balance.balance - spend_tx_amt  + coinbase_reward * blocks_mined
     assert_equals(alice_new_balance.balance, expected_balance)
 
-    bob_balance = alice_api.get_account_balance(pub_key=test_settings["spend_tx"]["bob_pubkey"])
+    bob_balance = common.get_account_balance(alice_api, pub_key=test_settings["spend_tx"]["bob_pubkey"])
     print("Coinbase reward is " + str(coinbase_reward) + ", had mined " +
             str(blocks_mined) + " blocks")
     print("Alice's balance (with a coinbase reward) is now " + str(alice_new_balance.balance))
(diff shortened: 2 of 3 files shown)


py/tests/integration/test_spend_tx.py
```diff
@@ -88,6 +88,83 @@ def test_successful():
...
+    # should not be able to use the incoming tokens until the spend transaction
+    # they are in is confirmed.
+    test_settings = settings["test_not_enough_tokens"]
+    coinbase_reward = common.coinbase_reward() 
+    (root_dir, bob_node, bob_api, bob_top) = setup_node_with_tokens(test_settings, "bob") 
+
+    spend_tx_amt = test_settings["spend_tx"]["amount"]
+    spend_tx_fee = test_settings["spend_tx"]["fee"]
+    bob_balance = bob_api.get_account_balance()
+    bob_has_not_enough_tokens = bob_balance.balance <= spend_tx_amt + spend_tx_fee
```
gorillainduction: Should be < not <=

gorillainduction:
Aside for the assert, this looks good to me.

py/tests/integration/test_spend_tx.py
```diff
@@ -37,7 +38,6 @@ def test_successful():
     bob_balance0 = {"balance":0} 
     try:
         bob_balance0 = alice_api.get_account_balance(pub_key=test_settings["spend_tx"]["bob_pubkey"])
-        assert_equals(bob_balance1.balance, bob_balance0.balance)
 
     except ApiException as e:
         assert_equals(e.status, 404) # Bob has no account yet
```
lucafavatella: This assertion may not be called in case the line `alice_api.get_account_balance(pub_key=test_settings["spend_tx"]["bob_pubkey"])` does not throw exception.

It looks like `nose.tools.assert_raises` with context manager shall be used. See [ref1](https://stackoverflow.com/questions/10716506/where-is-noses-assert-raises-function) and [ref2](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises).

Not needed for this PR though.

py/tests/integration/test_spend_tx.py
```diff
@@ -88,6 +88,83 @@ def test_successful():
...
+    bob_has_not_enough_tokens = bob_balance.balance < spend_tx_amt + spend_tx_fee
+    assert_equals(bob_has_not_enough_tokens, True)
+    print("Bob initial balance is " + str(bob_balance.balance) +
+            " and he will try to spend " + str(spend_tx_amt + spend_tx_fee))
+
+    alice_balance0 = {"balance":0} 
+    try:
+        alice_balance0 = bob_api.get_account_balance(pub_key=test_settings["spend_tx"]["alice_pubkey"])
+    except ApiException as e:
+        assert_equals(e.status, 404) # Alice has no account yet
```
lucafavatella: Same as https://github.com/aeternity/epoch/pull/383/files#r152543959

Not needed for this PR though.

py/tests/integration/test_spend_tx.py
lucafavatella: At lines 59-63 (check that Bob's balance has not changed, under the assumption that no block mined in the meantime) I see similar pattern, and there it is relevant as it is part of the user story ("The debit/credit should not happen until the transaction is confirmed, e.g. included in at least one block in the chain."). The assumption that no block mined in the meantime seems fragile (I assume mining autostart is true).

If correct, please track this in a ticket so that we do not lose track of it.

py/tests/integration/test_spend_tx.py
```diff
@@ -88,6 +88,83 @@ def test_successful():
...
+    print("Bob's spend_tx is " + str(spend_tx_obj))
+    bob_internal_api.post_spend_tx(spend_tx_obj)
+    print("Transaction sent")
+
+    # ensure Bob balance had not changed
+    bob_balance1 = bob_api.get_account_balance()
+    assert_equals(bob_balance.balance, bob_balance1.balance)
+
+    # ensure Alice balance had not changed
+    try:
```
lucafavatella: Fragile. Same as https://github.com/aeternity/epoch/pull/383/files#r152546842

If correct, please track this in a ticket so that we do not lose track of it.

py/tests/integration/test_spend_tx.py
```diff
@@ -88,6 +88,83 @@ def test_successful():
...
+    common.wait_until_height(bob_api, bob_top.height + 1)
+
+    bob_new_top = bob_api.get_top()
+    bob_new_balance = bob_api.get_account_balance()
+    
+    blocks_mined = bob_new_top.height - bob_top.height
+
+    # Since Bob had mined the block he is receiving the coinbase_reward
+    expected_balance = bob_balance.balance + coinbase_reward * blocks_mined
+    assert_equals(bob_new_balance.balance, expected_balance)
```
lucafavatella: This assertion is checking that, even if further blocks have been mined by Bob after the spend tx has been generated, the balance of Bob was not charged with the spend tx. Though what is not clear to me is how does the test ensures that the spend tx was considered for inclusion in block in the first place:
* The [line generating the spend tx](https://github.com/aeternity/epoch/pull/383/files#diff-4bd03bf9816115cc882755103452746eR125) does not assert success - or does it by simply no exception raised? Maybe an assert on the status code would clarify?
* Also, even asserting that spend tx was generated does not clarify whether tx was considered for inclusion in block. What is the expected result of a spend tx issued with not enough amount in the account? Is the spend tx meant to be included in a block and the fee taken? (Maybe a distinct mining node from Bob is needed for that?)

Please comment.

py/tests/integration/test_spend_tx.py
```diff
@@ -88,6 +88,83 @@ def test_successful():
...
+
+    alice_balance = Balance(balance=0) 
+    # ensure Alice balance had not changed
+    try:
+        alice_balance = bob_api.get_account_balance(pub_key=test_settings["spend_tx"]["alice_pubkey"])
```
lucafavatella: This seems fragile. If check on Bob is made more robust (see above), this check on Alice can be removed.

py/tests/integration/test_spend_tx.py
velzevur: There is only one node involved there: Alice's: her node mines. Bob has no node at all so if even if a block is mined - Bob should not receive a reward for it (because it would be mined by Alice). 
WRT exception case - I'll ammend the commit

py/tests/integration/test_spend_tx.py
velzevur: same comment as above

py/tests/integration/test_spend_tx.py
velzevur: Same comment as above: in this test case Bob's node mines and Alice's balance shouldn't change

py/tests/integration/test_spend_tx.py
velzevur: What if the fee is more than account's balance?
Currently if account does not have enough tokens for the transaction - the transaction is [ignored](https://github.com/aeternity/epoch/blob/master/apps/aecore/src/txs/aec_tx.erl#L125-L128) (not being included in the block) and no fee is applied for it.
Also [there is not much to be checked in the return code](https://github.com/aeternity/epoch/blob/master/apps/aehttp/src/aehttp_dispatch_int.erl#L11) since it doesn't validate that there are enough tokens to execute the tx

velzevur:
I've refactored a little bit getting the balance so it always returns a Balance object (even when there is no such account created, balance being 0)
As for checking that the spend tx was considered for inclusion: I don't think there is a good way of doing this while considering the node itself as a black box.

py/tests/integration/setup.yaml
```diff
@@ -73,6 +73,22 @@ tests: # test specific settings
         alice: dev1
       blocks_to_mine: 7
       spend_tx:
-        bob_pubkey: bob_address
+        bob_pubkey: Ym9iX2FkZHJlc3M=
```
lucafavatella: I see this is bob_address in base64. This may be fragile in the future if length of pubkey checked though not in scope for this PR.

```
1> base64:decode("Ym9iX2FkZHJlc3M=").
<<"bob_address">>
```

lucafavatella:
> Command make multi-build && make multi-start && make python-env && make python-single-integration-test TEST_NAME=spend_tx failed locally for me so I plan to test this a bit more locally before approving.

I see the test assumes node built but not started: `make multi-build && make python-env && make python-single-integration-test TEST_NAME=spend_tx` made more progress.

py/tests/integration/test_spend_tx.py
```diff
@@ -88,6 +79,71 @@ def test_successful():
...
+    bob_new_top = bob_api.get_top()
+    bob_new_balance = common.get_account_balance(bob_api)
+    
+    blocks_mined = bob_new_top.height - bob_top.height
+
+    # Since Bob had mined the block he is receiving the coinbase_reward
+    expected_balance = bob_balance.balance + coinbase_reward * blocks_mined
+    assert_equals(bob_new_balance.balance, expected_balance)
+
+    alice_balance = Balance(balance=0)
```
lucafavatella: This is not needed as variable re-assigned in next line (136).  And symbol `Balance` is not imported hence causing failure.
I am preparing patch locally.

lucafavatella: @velzevur  https://github.com/velzevur/epoch/pull/1

lucafavatella:
The only CI job missing for this PR is OSX package job. All other jobs passed.

I believe this PR does not need to wait for OSX package job:
* This PR does not affect creating the package, as it is only test files;
* This PR does not affect script `py/tests/release.py`, that is run in package job.

PR #384, opened 2017-11-17
Do not consume unexpected messages during OS process execution

akorosmezey:
While running PoW generation or validation in a separate OS process,
the code read unexpected messages in addition to the ones expected
from the process. This in not a problem for PoW generation as it runs
from a separate erlang process but the miner performs PoW validation
in its main process. It happened sometimes that the cast messages
the miner is sending to itself were consumed by the receive block
parsing the validation output, confusing the miner state machine.

Finishes PT-153087564

### DIFF
--- a/apps/aecore/src/aec_pow_cuckoo.erl
+++ b/apps/aecore/src/aec_pow_cuckoo.erl
@@ -219,9 +219,7 @@ wait_for_result(#state{os_pid = OsPid,
                           _                     -> Reason
                       end,
             ?error("OS process died: ~p~n", [Reason2]),
-            {error, {execution_failed, Reason2}};
-        _Other ->
-            wait_for_result(State)
+            {error, {execution_failed, Reason2}}
     end.
 
 %%------------------------------------------------------------------------------


apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -219,9 +219,7 @@ wait_for_result(#state{os_pid = OsPid,
                           _                     -> Reason
                       end,
             ?error("OS process died: ~p~n", [Reason2]),
-            {error, {execution_failed, Reason2}};
-        _Other ->
```
sennui: how we deal with those messages later? Removing this receive clause will make them enqueue in process queue.

akorosmezey: Such processes will be enqueued only for the duration of the OS process for PoW generation which is a relatively short interval. In practice, do not expect messages in this category. 
If this is a concern, I can
1. Use a different handler for spawn processes that read these messages 
2. Spawn separate erlang process for verification as @lucafavatella proposes below. In this case, however, I have to introduce a new state in the state machine. As I heard refactoring of the miner if planned, I am not sure if this is not an overkill.

ThomasArts:
The minor state machine is hairy and under rewrite. Your observation that messages can come in at the wrong state is correct. Leaving them in the mailbox makes them be read in thw "wrong" state, which then shows up as an error.

I like your approach to show them as an error, since just throwing them away seems also dangerous. However, I believe that the rewrite to a gen_server will have to address this issue and actually decide what to do with the message.

In short, it doesn't really matter whether we add this or not, we will have to deal with it in the rewrite!

apps/aecore/src/aec_pow_cuckoo.erl
lucafavatella: I see that if any such message is received, it is queued in the aec_miner statem - e.g. see https://github.com/aeternity/epoch/blob/367a0915e1132a80032963a79b26ee7ce23b35c3/apps/aecore/src/aec_miner.erl#L422

PR #388, opened 2017-11-20
[Fixes #152878235] Make the functional adt tests for chain more complete

gorillainduction:
As a step of moving away from the aec_chain_server, I re-implement the tests that goes through the chain server api to testing of the functional adt.

All chain server tests are kept for now, but (at least) the same test coverage should now be on the functional adt so that we know that we do not miss any functionality when moving forward.

The new tests also covers more of the functionality since the chain server tests where adapted from the  original implementation.

Note that the API return values has changed slightly for get_{block,header}_by_height, in the error cases, so the corresponding call sites had to be updated as well. Since there was so few I include the changes in this PR. The corresponding error returns for get_{block,header}_by_hash is left for later, in order to change the functionality as little as possible at this stage.

### DIFF
--- a/apps/aecore/src/aec_chain.erl
+++ b/apps/aecore/src/aec_chain.erl
@@ -51,15 +51,9 @@
         {ok, block()} |
         {error, Reason::{block_not_found, {top_header, header()}}}.
 -type get_header_by_height_reply() ::
-        {ok, header()} |
-        {error, Reason::{chain_too_short, {{chain_height, height()},
-                                           {top_header, header()}}}}.
+        {ok, header()} | {error, atom()}.
 -type get_block_by_height_reply() ::
-        {ok, block()} |
-        {error, Reason::{chain_too_short, {{chain_height, height()},
-                                           {top_header, header()}}} |
-                        {block_not_found, {top_header, header()}}
-        }.
+        {ok, block()} | {error, atom()}.
 -type insert_header_reply_ok() :: ok.
 -type insert_header_reply_error() ::
         {error, Reason::{previous_hash_is_not_top, {top_header, header()}} |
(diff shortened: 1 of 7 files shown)


ThomasArts:
After looking  at the changes, which looked ok, I ran QuickCheck over it:

```erlang
aec_chain_eqc:start({block, 0,
       <<0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0>>,
       <<0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0>>,
       {trees, {0, empty}},
       <<0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0>>,
       [], 553713663, 0, 0, 1, no_value}) ->
  ok.
aec_chain_eqc:get_block_by_hash(<<>>) ->
  {error,
     {block_not_found,
        {top_header,
           {header, 0,
              <<0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0>>,
              <<0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0>>,
              <<0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0>>,
              553713663, 0, 0, 1, no_value}}}}.
```

I guess that also here we would expect an atom as error and not all this junk?

gorillainduction:
@ThomasArts I agree, but I wanted this PR to modify as little as possible of the semantics of the code. I will fix this in the refactoring ticket.

PR #385, opened 2017-11-20
[WIP] Fix race in aec_miner process between mining and block posting

lucafavatella:
Change for testnet. Part of [PT-152968563](https://www.pivotaltracker.com/story/show/152968563).

PR #389, opened 2017-11-20
Fix race in aec_miner process between mining and block posting

lucafavatella:
Supersedes PR #385.

Change for testnet. Part of [PT-152968563](https://www.pivotaltracker.com/story/show/152968563).

### DIFF
--- a/apps/aecore/src/aec_miner.erl
+++ b/apps/aecore/src/aec_miner.erl
@@ -383,7 +383,8 @@ running({call, From}, start, State) ->
     {keep_state, State, [{reply, From, ok}]};
 running({call, From}, suspend, State) ->
     epoch_mining:info("Mining suspended by user.", []),
-    {next_state, idle, State, [{reply, From, ok}]};
+    kill_miner(State#state.miner),
+    {next_state, idle, State#state{miner = none}, [{reply, From, ok}]};
 
 %% Casts: post_block | miner_done | create_block_candidate |
 %%        bump_initial_cycle_nonce | mine | check_keys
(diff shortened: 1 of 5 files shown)


lucafavatella:
The CI build on the `push` is in my repo, and is https://travis-ci.org/lucafavatella/epoch/builds/304782347?utm_source=github_status&utm_medium=notification

The CI build for the pr (as opposed to the `push` one) is showed in this PR.

lucafavatella:
> In the testing: why is it insufficient to just add application:stop(erlexec) in the aec_eunit_SUITE:end_per_test_case??

This would be sufficient, though less clean and probably slower.
Calling application:stop(erlexec) while native miners (e.g. lean16 binary) run would cause the erlexec app to attempt to stop the children, that would cause the OS (depending on OS configuration) to dump the core.

> Or simply ignore there that it is created?

This is sufficient, though a workaround (it involves whitelisting erlexec). After further consideration, I realized that waiting for the binaries to finish execution is faster and cleaner.

Having signal handling in the (upstream) miner binaries in order to avoid core dump would ease this.

ThomasArts:
SO what about the whitelisting as I propose in the code snippet?
-- [{erlexec,"OS Process Manager","1.7.1"}]

Would that do the job? This is far cleaner than adding all the code for each test.

lucafavatella:
> SO what about the whitelisting as I propose in the code snippet?
> -- [{erlexec,"OS Process Manager","1.7.1"}]
> 
> Would that do the job? This is far cleaner than adding all the code for each test.

Alongside changing all `application:start(erlexec)`  to `application:ensure_started(erlexec)`, and removing all `application:stop(erlexec)`, it would do the job. I am preparing alternative PR with that solution.

lucafavatella:
Closing this now that PR #392 is merged.

PR #387, opened 2017-11-20
Use more generic x86-64 Cuckoo PoW binaries for shipping package

lucafavatella:
[PT-152965197](https://www.pivotaltracker.com/story/show/152965197)

~It can be merged, thanks to commit d1b8b77 being tracked in the aeternity fork of the cuckoo repo in branch https://github.com/aeternity/cuckoo/tree/user-specified-compiler-flags~
~If PR https://github.com/tromp/cuckoo/pull/30 gets merged upstream, this commit will get into the upstream master, and removing the need for the user-specified-compiler-flags branch.~
It can be merged, thanks to commit d1b8b77 having been merged upstream and present in the master of the https://github.com/aeternity/cuckoo repo.

PR #390, opened 2017-11-20
Fix harmless excessive verbosity of Cuckoo executables build

lucafavatella:
Depends on PR #387 (included in this PR).

### DIFF
--- a/apps/aecuckoo/Makefile
+++ b/apps/aecuckoo/Makefile
@@ -22,6 +22,7 @@ COMMIT = d1b8b77
 
 .PHONY: all
 all: $(LIBBLAKE) $(EXECUTABLES)
+	@: # Silence the `Nothing to be done for 'all'.` message when running `make all`.
 
 .PHONY: clean
 clean:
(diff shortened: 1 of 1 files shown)


lucafavatella:
Force-pushed rebase on master for CI stability.

hanssv:
This is still chatty since it copies all binaries even when nothing has changed... (This amounts to ~40% of all the output when you do make in an up-to-date `epoch`-repo...
I suggest something like this:
```make
$(LIBBLAKE): $(SRC_LIBBLAKE)
	@mkdir -p priv/lib
	cp -p $< $@

$(EXECUTABLES): priv/bin/%: c_src/src/%
	@mkdir -p priv/bin
	cp -p $< $@
```

and removing the the targets `priv/bin` and `priv/lib`... This will make completely silent when nothing is changed, at the sacrifice of doing some unnecessary `mkdir -p`

hanssv:
Last bit of noise (`make[1]: Nothing to be done for 'all'.` can be removed by adding to the `all:` rule:
```
all: $(LIBBLAKE) $(EXECUTABLES)
	@:
```

lucafavatella:
> Let's do it completely silent when nothing has changed!

@hanssv Done in d68992cf (amended, rebased on master, force-pushed).

lucafavatella:
I see CI passed. I see two reviewers approved. Merging.

PR #391, opened 2017-11-20
Use 64 bit nonce with validation

akorosmezey:
finishes:
* [PT-152915267](https://www.pivotaltracker.com/n/projects/2124891/stories/152915267) - Validate received block nonce
* [PT-152672154](https://www.pivotaltracker.com/n/projects/2124891/stories/152672154) - Investigate 64 bits PoW nonce for increased source of randomness


py/tests/integration/data/bchain.txt
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+[{"nonce": 0, "target": 553713663, "transactions": null, "pow": null, "time": 0, "height": 0, "version": 1, "txs_hash": null, "prev_hash": null, "state_hash": null}]
```
lucafavatella: Not sure this is valid. How did you generate it? Cc @velzevur

hanssv: Akos' work I'm afraid...

velzevur: Please ignore `py/tests/integration/data/bchain.txt`
I'll have it refactored in a subsequent PR

lucafavatella: Os this this bchain.txt change shall be removed from this PR.

lucafavatella:
I am checking https://github.com/aeternity/epoch/wiki/Proof-of-Work - that I understand goes alongside this PT-152672154 PR.

hanssv:
The wiki-page should be updated with the change 30 -> 28 as well

lucafavatella:
As part of PT-152672154, I understand that:
* [PoW wiki page](https://github.com/aeternity/epoch/wiki/Proof-of-Work) shall be refreshed/revisited/polished;
* [Block header wiki page](https://github.com/aeternity/epoch/wiki/Block-Header) shall link to the PoW description in fields `nonce` and `pow`.

Not needed for merging this PR, but mentioning for ensuring we do not lose track of this.

apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -184,6 +191,47 @@ verify(Hash, Nonce, Soln) ->
...
+    %% converts every triplet of bytes to 4 characters: from N bytes to 4*ceil(N/3)
+    %% bytes.)
+    %%
+    %% Like Cuckoo, (and unlike we hash the header) we use little-endian here.
+    NonceStr = base64:encode_to_string(<<Nonce:64/little-unsigned-integer>>),
+    HashStr = base64:encode_to_string(Hash),
+    12 = length(NonceStr),
+    44 = length(HashStr),
+    %% Fill in space between the hash and the nonce so that is spans the
+    %% 76-bytes before the 4 bytes Cuckoo overwrites. 76 - 44 - 12 = 20.
```
lucafavatella: This assumes that Cuckoo miner binary overwrites last 4 bytes of the 80 bytes. Links to assumption (mean and lean) would help reader here. What if actually it overwrites 8 bytes hence eating up some of the nonce?

Also - why attempting to use the end of the buffer? I.e. why not avoiding this 20 bytes altogether?

hanssv:
```c
((u32 *)headernonce)[len/sizeof(u32)-1] = htole32(nonce); // place nonce at end
```

Again this is Akos design... I would probably have put it immediately after the Hash but I didn't have a strong enough opinion to make a change. It is quickly fixed if you think it is much better?

lucafavatella: I had a word with Akos - it was for placing the nonce as close as possible to where the upstream Cuckoo miner binary places it. So this 20-bytes interval can be dropped, I suggest to do so for this PR (so it makes describing the PoW in the wiki easier).

hanssv: Good, I'll move it to the `obvious` place then! I will also make a comment that lean/mean (and verify) will put a nonce = 0 at the end of its buffer...

hanssv:
Moved the Nonce immediately after the Hash, and then got a conflict with #403 that is resolved.

apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -181,6 +187,43 @@ verify(Hash, Nonce, Soln) ->
...
+    %% bytes) for cuckoo to put its nonce (which will be 0 in our case) in.
+    %%
+    %% (Base64 encoding: see RFC 3548, Section 3:
+    %% https://tools.ietf.org/html/rfc3548#page-4
+    %% converts every triplet of bytes to 4 characters: from N bytes to 4*ceil(N/3)
+    %% bytes.)
+    %%
+    %% Like Cuckoo, we use little-endian for the nonce here.
+    NonceStr = base64:encode_to_string(<<Nonce:64/little-unsigned-integer>>),
+    HashStr = base64:encode_to_string(Hash),
```
lucafavatella: May I suggest hardcoding expectations on lengths of base64-encoded NonceStr and HashStr here please? It would make the code more readable.

lucafavatella: Guard on pack_header_and_nonce clause head would help too.

PR #393, opened 2017-11-20
[WIP] Use correct conversion to binary in hash calc for PoW

akorosmezey:
Finishes [PT-152917447](https://www.pivotaltracker.com/n/projects/2124891/stories/152917447) -- Review determinism of serialization of header for computation of PoW


lucafavatella:
I see conflicts. Let me review and re-propose this.

PR #399, opened 2017-11-21
Genesis block token distribution [Finishes #152479338]

velzevur:
`apps/aecore/priv/genesis/accounts.json` stores the public addresses that will have tokens in the genesis block. Currently it has only one address - the one of the spender account in [testnet-keys](https://github.com/aeternity/testnet-keys)

NB: changing anything in this file will result in a different genesis block (with a different hash) that will not be compatible with any chain so far.

This PR also includes some cleanup in `apps/aecore/src/aec_block_genesis.erl` and in `apps/aecore/src/aec_chain_state.erl`

### DIFF
--- a/apps/aecore/src/aec_block_genesis.erl
+++ b/apps/aecore/src/aec_block_genesis.erl
@@ -15,7 +15,7 @@
 %%%   * This means that validation function attempting to check that
 %%%     there is at least a coinbase transaction in a block needs to
 %%%     have a special case for genesis.
-%%% * The hash values in it have special value i.e. all zeros.
+%%% * The hash values in it may have special values i.e. all zeros.
 %%%   * This means that validation function attempting to consider the
 %%%     hashes in a block needs to have a special case for genesis.
 %%% @end
@@ -24,51 +24,53 @@
 
 %% API
 -export([genesis_header/0,
-         genesis_block_as_deserialized_from_network/0,
+         height/0,
          genesis_block/0,
-         height/0]).
+         populated_trees/0]).
 
 -include("common.hrl").
 -include("blocks.hrl").
 
+%% Since preset accounts are being loaded from a file - please use with caution
 genesis_header() ->
-    #header{
-       version = ?GENESIS_VERSION,
-       height = ?GENESIS_HEIGHT,
-       prev_hash = <<0:?BLOCK_HEADER_HASH_BYTES/unit:8>>,
-       txs_hash = <<0:?TXS_HASH_BYTES/unit:8>>,
-       root_hash = <<0:?STATE_HASH_BYTES/unit:8>>,
-       target = ?HIGHEST_TARGET_SCI,
-       pow_evidence = no_value,
-       nonce = 0,
-       time = 0 %% Epoch.
-      }.
-
-genesis_block_as_deserialized_from_network() ->
-    H = genesis_header(),
-    #block{
-       version = H#header.version,
-       height = H#header.height,
-       prev_hash = H#header.prev_hash,
-       txs_hash = H#header.txs_hash,
-       root_hash = H#header.root_hash,
-       target = H#header.target,
-       pow_evidence = H#header.pow_evidence,
-       nonce = H#header.nonce,
-       time = H#header.time,
-       txs = [],
-       trees = _DummyTrees = #trees{}
-      }.
+    aec_blocks:to_header(genesis_block()).
 
 %% Returns the genesis block including the state trees.
 %%
 %% The current implementation of state trees causes a new Erlang term,
-%% representing the initial empty state trees, to be allocated in the
+%% representing the initial state trees, to be allocated in the
 %% heap memory of the calling process.
+%%
+%% Since preset accounts are being loaded from a file - please use with caution
 genesis_block() ->
-    B = genesis_block_as_deserialized_from_network(),
-    {ok, T} = aec_trees:all_trees_new(),
-    B#block{trees = T}.
+    Trees = populated_trees(),
+    #block{
+      version = ?GENESIS_VERSION,
+      height = ?GENESIS_HEIGHT,
+      prev_hash = <<0:?BLOCK_HEADER_HASH_BYTES/unit:8>>,
+      txs_hash = <<0:?TXS_HASH_BYTES/unit:8>>,
+      root_hash = aec_trees:all_trees_hash(Trees),
+      target = ?HIGHEST_TARGET_SCI,
+      pow_evidence = no_value,
+      nonce = 0,
+      time = 0, %% Epoch.
+      trees = Trees
+      }.
+
+populated_trees() ->
+    {ok, T0} = aec_trees:all_trees_new(),
+    PresetAccounts = aec_genesis_block_settings:preset_accounts(),
+    Trees = add_preset_accounts(PresetAccounts, T0),
+    Trees.
+
+add_preset_accounts([], Trees) ->
+    Trees;
+add_preset_accounts([{PubKey, Amt} | T], Trees0) ->
+    Account = aec_accounts:new(PubKey, Amt, 0),
+    AccountsTrees0 = aec_trees:accounts(Trees0),
+    {ok, AccountsTrees} = aec_accounts:put(Account, AccountsTrees0),
+    Trees = aec_trees:set_accounts(Trees0, AccountsTrees),
+    add_preset_accounts(T, Trees).
 
 height() ->
     ?GENESIS_HEIGHT.
--- a/apps/aecore/src/aec_chain.erl
+++ b/apps/aecore/src/aec_chain.erl
@@ -31,7 +31,9 @@
          insert_header/1,
          top/0,
          top_header/0,
-         write_block/1 %% TODO: rename
+         write_block/1, %% TODO: rename
+         genesis_header/0,
+         genesis_block/0
 	]).
 
 -include("common.hrl").
@@ -138,6 +140,12 @@ insert_header(Header) ->
 write_block(Block) ->
     gen_server:call(?CHAIN_SERVER, {write_block, Block}, ?DEFAULT_CALL_TIMEOUT).
 
+genesis_header() ->
+    gen_server:call(?CHAIN_SERVER, get_genesis_header, ?DEFAULT_CALL_TIMEOUT).
+
+genesis_block() ->
+    gen_server:call(?CHAIN_SERVER, get_genesis_block, ?DEFAULT_CALL_TIMEOUT).
+
 %% Returns the amount of work done on the chain i.e. the total
 %% difficulty of the top header.
 -spec get_total_difficulty() -> get_total_difficulty_reply().
@@ -152,26 +160,26 @@ common_ancestor(Hash1, Hash2) ->
                     {common_ancestor, Hash1, Hash2}).
 
 get_transactions_between(Hash1, Root) ->
-    try get_transactions_between(Hash1, Root, []) of
+    {ok, GenesisHeader} = genesis_header(),
+    {ok, GenesisHash} = aec_headers:hash_header(GenesisHeader),
+    try get_transactions_between(Hash1, Root, GenesisHash, []) of
         Transactions ->
             {ok, Transactions}
     catch throw:Error -> Error
     end.
 
-get_transactions_between(Hash, Hash, Transactions) ->
+get_transactions_between(Hash, Hash, _GenesisHash, Transactions) ->
     Transactions;
-get_transactions_between(Hash, Root, Transactions) ->
-    case aec_headers:hash_header(aec_block_genesis:genesis_header()) of
-        %% Current block is genesis
-        Hash -> Transactions;
-        %% Block is not genesis
-        _ ->
-            case get_block_by_hash(Hash) of
-                {ok, #block{prev_hash = Parent,
-                            txs = BlockTransactions}} ->
-                    get_transactions_between(Parent, Root,
-                                             BlockTransactions ++ Transactions);
-                {error,_} -> throw({error, {block_off_chain, Hash}})
-            end
+get_transactions_between(Hash, _Root, GenesisHash, Transactions)
+    when Hash =:= GenesisHash ->
+    %% Current block is genesis
+    Transactions;
+get_transactions_between(Hash, Root, GenesisHash, Transactions) ->
+    case get_block_by_hash(Hash) of
+        {ok, #block{prev_hash = Parent,
+                    txs = BlockTransactions}} ->
+            get_transactions_between(Parent, Root, GenesisHash,
+                                      BlockTransactions ++ Transactions);
+        {error,_} -> throw({error, {block_off_chain, Hash}})
     end.
 
--- a/apps/aecore/src/aec_chain_server.erl
+++ b/apps/aecore/src/aec_chain_server.erl
@@ -65,12 +65,14 @@ stop() ->
 
 init(_Args = [GenesisBlock]) ->
     process_flag(trap_exit, true),
-    {_, State} = insert_block(GenesisBlock, new_state()),
+    {ok, State} = insert_block(GenesisBlock, new_state()),
+    assert_genesis_set(State),
     {ok, State};
 init({Chain, TopBlockHash, TopBlockStateTrees}) ->
     process_flag(trap_exit, true),
     StateTrees = [{TopBlockHash, TopBlockStateTrees}],
-    {_, State} = init_chain(Chain, new_state(Chain, StateTrees)),
+    {ok, State} = init_chain(Chain, new_state(Chain, StateTrees)),
+    assert_genesis_set(State),
     {ok, State}.
 
 init_chain([#block{} = Block | Rest], State) ->
@@ -107,6 +109,14 @@ handle_call(difficulty, _From, State) ->
 handle_call({common_ancestor, Hash1, Hash2}, _From, State) ->
     {reply,
      aec_chain_state:find_common_ancestor(Hash1, Hash2,State), State};
+handle_call(get_genesis_header, _From, State) ->
+    GenesisHash = aec_chain_state:get_genesis_hash(State),
+    Resp = {ok, _} = aec_chain_state:get_header(GenesisHash, State), 
+    {reply, Resp, State};
+handle_call(get_genesis_block, _From, State) ->
+    GenesisHash = aec_chain_state:get_genesis_hash(State),
+    Resp = {ok, _} = aec_chain_state:get_block(GenesisHash, State),
+    {reply, Resp, State};
 
 %% Update functions
 handle_call({insert_header, H}, _From, State) ->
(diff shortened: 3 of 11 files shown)


lucafavatella:
> NB: changing anything in this file will result in a different genesis block (with a different hash) that will not be compatible with any chain so far.

@velzevur  I understand this is the first PR exercising https://www.pivotaltracker.com/n/projects/2124891/stories/152956790
* Wrt deployment in environments, I understand this PR needs to change https://github.com/aeternity/epoch/blob/19e60155719581dda3196912ae4490452381bc11/deployment/ansible/deploy.yml#L11
* Wrt users downloading software, please comment.

Cc @dincho

apps/aecore/src/aec_block_genesis.erl
```diff
@@ -24,47 +24,46 @@
...
+      }.
+
+populated_trees() ->
+    {ok, T0} = aec_trees:all_trees_new(),
+    PresetAccounts = aec_genesis_block_settings:preset_accounts(),
+    Trees = add_preset_accounts(PresetAccounts, T0),
+    Trees.
+
+add_preset_accounts([], Trees) ->
+  Trees;
```
lucafavatella: Inconsistent indentation (no need to update PR for this).

apps/aecore/src/aec_block_genesis.erl
```diff
@@ -24,47 +24,46 @@
...
-    B = genesis_block_as_deserialized_from_network(),
-    {ok, T} = aec_trees:all_trees_new(),
-    B#block{trees = T}.
+    Trees = populated_trees(),
+    #block{
+      version = ?GENESIS_VERSION,
+      height = ?GENESIS_HEIGHT,
+      prev_hash = <<0:?BLOCK_HEADER_HASH_BYTES/unit:8>>,
+      txs_hash = <<0:?TXS_HASH_BYTES/unit:8>>,
+      root_hash = aec_trees:all_trees_hash(Trees),
```
lucafavatella: Please update header of the file. From:
```
 %%% * The hash values in it have special value i.e. all zeros.
```
to something link:
```
 %%% * The hash values in it may have special value i.e. all zeros.
```
(or similar).

apps/aecore/src/aec_block_genesis.erl
```diff
@@ -24,47 +24,46 @@
...
+      target = ?HIGHEST_TARGET_SCI,
+      pow_evidence = no_value,
+      nonce = 0,
+      time = 0, %% Epoch.
+      trees = Trees
+      }.
+
+populated_trees() ->
+    {ok, T0} = aec_trees:all_trees_new(),
+    PresetAccounts = aec_genesis_block_settings:preset_accounts(),
```
lucafavatella: The Merkle tree library depends on the order of operations applied on it. I.e. is same accounts are put in distinct order, a distinct root hash may be obtained. Please comment on this function returning a predictable order of the accounts.

apps/aecore/src/aec_genesis_block_settings.erl
```diff
@@ -0,0 +1,25 @@
+-module(aec_genesis_block_settings).
+
+-export([dir/0,
+         preset_accounts/0]).
+
+dir() ->
+    case application:get_env(aecore, genesis_settings_dir) of
```
lucafavatella: The genesis block shall not be configurable. Please correct me if I am wrong.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -556,7 +544,7 @@ calculate_state_trees(Node, Acc, State) ->
         {ok, Trees} when Acc =:= [] -> {stored, Trees};
         {ok, Trees} -> {calculated, apply_node_transactions(Acc, Trees, State)};
         error when IsGenesis  ->
-            Trees = genesis_state_tree(Node),
+            Trees = aec_blocks:trees(Node#node.content),
```
lucafavatella: I see this removed hardcoded expectation of genesis having no txs. I need to understand this better.

lucafavatella: The hardcoded expectation on no txs in genesis justified the lack of call to [`apply_node_transactions`](https://github.com/velzevur/epoch/blob/19e60155719581dda3196912ae4490452381bc11/apps/aecore/src/aec_chain_state.erl#L564) as [called in the case 2 lines above](https://github.com/aeternity/epoch/pull/399/files#diff-4ac652308bacbca71e5f388759d362c1R545). Please re-instate hardcode expectation on no txs.

apps/aecore/src/aec_genesis_block_settings.erl
velzevur: The idea is having a couple of incompatible distinct networks (for example one for uat and one for the public nodes). This is how we'd guarantee that someone will not connect our distinct networks and make them peers.

lucafavatella: I see. So keep it.

apps/aecore/src/aec_chain_state.erl
gorillainduction: I second adding back the assertion of an empty transactions list. It works for now, but it is better to be safe for the future.

apps/aecore/src/aec_genesis_block_settings.erl
```diff
@@ -0,0 +1,25 @@
...
+            filename:join(code:priv_dir(aecore), "genesis");
+      {ok, GenSettingsDir} ->
+            GenSettingsDir
+    end.
+
+preset_accounts() ->
+    PresetAccountsFile = filename:join(dir(), "accounts.json"),
+    case file:read_file(PresetAccountsFile) of
+        {error, _Err} ->
+            [];
```
gorillainduction: Do we really want to silently fail here?

sennui:
the changes (like file read) should kick in only if there is no persisted block

apps/aecore/src/aec_block_genesis.erl
lucafavatella: thanks

velzevur:
So to your comments:

- This will change the genesis block's hash and will invalidate all chains (unless they're using persistence, please see bellow)
- Accounts being read are sorted by their keys to ensure deterministic applinf of their accounts
- The hardcoded expectation on no txs in genesis in aec_chain_state is back
- The genesis blocks preset accounts directory is not configurable one anymore, it is just `.genesis/accounts.json`
- The persisted chain has a higher priority: if the persistency is on - the genesis file setup is ignored (this is how `aec_chain_server` works)

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -170,8 +169,7 @@ cleanup_genesis(Val) ->
     Val.
 
 add_missing_to_genesis_block(#{<<"height">> := 0} = Block) ->
-    GB = aec_blocks:serialize_to_map(
-           aec_block_genesis:genesis_block_as_deserialized_from_network()),
+    GB = aec_blocks:serialize_to_map(aec_block_genesis:genesis_block()),
```
lucafavatella: So attempting to post block with height 0 attempts to read from disk?

apps/aecore/src/aec_genesis_block_settings.erl
```diff
@@ -0,0 +1,25 @@
...
+        {ok, JSONData} ->
+            Accounts =
+                lists:map(
+                    fun({EncodedPubKey, Amt}) ->
+                        {base64:decode(EncodedPubKey), Amt}
+                    end,
+                    jsx:decode(JSONData)),
+            % ensure deterministic ordering of accounts
+            lists:keysort(1, Accounts)
```
lucafavatella: This function is reading from file and sorting a potentially large list.
Maybe the result of this function - or the whole genesis block and block header - shall be cached in an ETS table lazily initialized?

apps/aehttp/src/aehttp_dispatch_ext.erl
velzevur: Yes, this could be a vector for an attack, I'll modify it to use the genesis block from the chain instead

apps/aecore/src/aec_genesis_block_settings.erl
lucafavatella: This may avoid the need for the `genesis_block_hash` change?

velzevur:
I've added a second commit - replacing all `aec_block_genesis` calls with calls to `aec_chain` ones, so we have an easy way to get the genesis block/header from heap.
The only places left using `aec_block_geneis:genesis_block/0` is in the spawning of aec_chain and the tests.

This way the `aec_chain` is the only source of truth for the genesis block.

If in the standup we decide going lazy-ETS solution - we'd discard the second commit, otherwise I'd squash them before merging.

lucafavatella:
> If no .genesis/accounts.json is found - we consider that the genesis block was not setuped to have any preset accounts

This seems risky, as path misconfiguration / misdetection causes an unexpected genesis. Anyway I will review changes first.

lucafavatella:
> The persisted chain has a higher priority: if the persistency is on - the genesis file setup is ignored (this is how aec_chain_server works)

This means that this PR needs to bump the db_path in the deploy.yml
(I was trying to do the same in https://github.com/aeternity/epoch/pull/410 but I have not managed to test is successfully using the release notes though).

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -170,8 +169,8 @@ cleanup_genesis(Val) ->
     Val.
 
 add_missing_to_genesis_block(#{<<"height">> := 0} = Block) ->
-    GB = aec_blocks:serialize_to_map(
-           aec_block_genesis:genesis_block_as_deserialized_from_network()),
+    {ok, GenesisBlock} = aec_chain:genesis_block(),
```
lucafavatella: It looks like this change is now moving across process heaps a potentially large term - it being the in-memory trees.

apps/aecore/src/aec_chain.erl
```diff
@@ -152,26 +160,26 @@ common_ancestor(Hash1, Hash2) ->
                     {common_ancestor, Hash1, Hash2}).
 
 get_transactions_between(Hash1, Root) ->
-    try get_transactions_between(Hash1, Root, []) of
+    {ok, GenesisHeader} = genesis_header(),
+    GenesisHash = aec_headers:hash_header(GenesisHeader),
```
gorillainduction: aec_headers:hash_header returns {ok, Hash}, right? If you fix this in the code you will solve https://www.pivotaltracker.com/n/projects/2124891/stories/152954773 as well.

apps/aehttp/src/aehttp_dispatch_ext.erl
lucafavatella: And actually this line - by inspection of `empty_fields_in_genesis()` - only needs the genesis header plus the transactions, and the transactions in genesis can be assumed to be empty in this point.

No need to change so far - I need to check aec_chain implementation.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -106,14 +109,25 @@ top_block(?match_state(top_block_hash := X) = State) ->
...
+                    State1
+            end,
+        {ok, State}
     catch throw:?internal_error(What) -> {error, What}
     end.
 
 -spec insert_header(#header{}, state()) -> {'ok', state()} | {'error', any()}.
-insert_header(Header, ?assert_state() = State) ->
-    try {ok, internal_insert(wrap_header(Header), State)}
+insert_header(Header, ?assert_state() = State0) ->
```
gorillainduction: Either there is some genesis stuff missing, or you should revert to the old code.

apps/aecore/src/aec_chain_server.erl
```diff
@@ -107,6 +107,26 @@ handle_call(difficulty, _From, State) ->
 handle_call({common_ancestor, Hash1, Hash2}, _From, State) ->
     {reply,
      aec_chain_state:find_common_ancestor(Hash1, Hash2,State), State};
+handle_call(get_genesis_header, _From, State) ->
+    GenesisHash = aec_chain_state:get_genesis_hash(State),
+    Resp =
+        case aec_chain_state:get_header(GenesisHash, State) of
+            error ->
```
lucafavatella: Is not the presence of the genesis guaranteed by the `init` and by the persisted chain not being corrupted?
If so, I expect this case to be replaced with hardcoded expectation (match) that the genesis_hash is in state.

apps/aecore/src/aec_chain_server.erl
```diff
@@ -107,6 +107,26 @@ handle_call(difficulty, _From, State) ->
...
+                genesis_missing;
+            {ok, Header} ->
+                {ok, Header}
+        end,
+    {reply, Resp, State};
+handle_call(get_genesis_block, _From, State) ->
+    GenesisHash = aec_chain_state:get_genesis_hash(State),
+    Resp =
+        case aec_chain_state:get_block(GenesisHash, State) of
```
lucafavatella: As [above](https://github.com/aeternity/epoch/pull/399/files#r152946605) - hardcoded expectation?

apps/aecore/src/aec_chain_state.erl
```diff
@@ -38,6 +39,7 @@
                     , 'max_snapshot_height' => pos_integer()
                     , 'sparse_snapshots_interval' => pos_integer()
                     , 'keep_all_snapshots_height' => pos_integer()
+                    , 'genesis_block_hash' => 'undefined' | binary()
```
lucafavatella: Can this ever be undefined?

apps/aecore/src/aec_chain_server.erl
```diff
@@ -107,6 +107,26 @@ handle_call(difficulty, _From, State) ->
 handle_call({common_ancestor, Hash1, Hash2}, _From, State) ->
     {reply,
      aec_chain_state:find_common_ancestor(Hash1, Hash2,State), State};
+handle_call(get_genesis_header, _From, State) ->
+    GenesisHash = aec_chain_state:get_genesis_hash(State),
```
lucafavatella: I see the genesis hash is kept in the state so to optimize the retrieval of genesis. As a "get-by-height-0" would otherwise traverse the chain from top all the way down. Ok.

apps/aecore/src/aec_chain_state.erl
lucafavatella: I see this is undefined because of how a new state is created:
* When [state is created knowing genesis block](https://github.com/aeternity/epoch/pull/399/files#diff-675152b2278724570b642ab179e889e3R68), genesis block could be passed to function initializing this state that would compute this field - hence no need of undefined for this case;
* When [state is created from persisted chain](https://github.com/aeternity/epoch/pull/399/files#diff-675152b2278724570b642ab179e889e3R73), it looks like that the persisted state that is argument of init (`Chain, TopBlockHash, TopBlockStateTrees}`) shall contain also `GenesisBlockHash`.

I need to understand whether when starting from persisted state:
* This field ever get set;
* This causes start time to be longer.

lucafavatella: Could you please hardcode expectation at the end of init that this field is a binary (hence undefined). This shall be done in two points:
* [Init from genesis block](https://github.com/aeternity/epoch/pull/399/files#diff-675152b2278724570b642ab179e889e3R69) - this shall be straightforward;
* [Init from persisted chain](https://github.com/aeternity/epoch/pull/399/files#diff-675152b2278724570b642ab179e889e3R74) - this would hardcode expectation that persisted chain must contain genesis header (that seems reasonable).

Optionally, in init that start from persisted chain pass genesis block (https://github.com/aeternity/epoch/pull/399/files#diff-b98da3b40a4c433bc375929f51069cb3R87) and check that the genesis header hash from the persisted chain is the same as the one from the expected genesis.

apps/aecore/src/aec_chain_server.erl
```diff
@@ -107,6 +107,14 @@ handle_call(difficulty, _From, State) ->
...
```
lucafavatella: Who guarantees this is not undefined?

apps/aecore/src/aec_chain_server.erl
```diff
@@ -107,6 +107,14 @@ handle_call(difficulty, _From, State) ->
...
+    Resp = {ok, _} = aec_chain_state:get_header(GenesisHash, State), 
+    {reply, Resp, State};
+handle_call(get_genesis_block, _From, State) ->
+    GenesisHash = aec_chain_state:get_genesis_hash(State),
```
lucafavatella: Who guarantees this is not undefined?

apps/aehttp/src/aehttp_dispatch_ext.erl
lucafavatella: Opened https://www.pivotaltracker.com/story/show/153140910

apps/aecore/src/aec_chain_state.erl
```diff
@@ -106,8 +109,19 @@ top_block(?match_state(top_block_hash := X) = State) ->
...
-    try {ok, internal_insert(wrap_block(Block), State)}
+insert_block(Block, ?assert_state() = State0) ->
+    try 
+        State1 = internal_insert(wrap_block(Block), State0),
+        State =
+            case Block#block.height of
+                0 ->
+                    {ok, GenesisHash} =
+                        aec_blocks:hash_internal_representation(Block),
+                    State1#{genesis_block_hash => GenesisHash};
```
lucafavatella: Is this line changing the genesis???

lucafavatella: I see handle_call(write_block) then ends here.

velzevur: if a new genesis block is added - it gets rejected in the checks in aec_chain_state:internal_insert/2 the check itself is done [here](https://github.com/velzevur/epoch/blob/7cf6f7345cdd0fe95269a68407f79fa1af986a73/apps/aecore/src/aec_chain_state.erl#L423)
anyway - I've added an assertion

velzevur:
All comments are addressed now

apps/aecore/src/aec_chain_server.erl
lucafavatella: I see assertion below - thanks.

apps/aecore/src/aec_chain_server.erl
lucafavatella: I see assertion below - thanks.

apps/aecore/src/aec_chain_state.erl
lucafavatella: > Optionally, in init that start from persisted chain pass genesis block (https://github.com/aeternity/epoch/pull/399/files#diff-b98da3b40a4c433bc375929f51069cb3R87) and check that the genesis header hash from the persisted chain is the same as the one from the expected genesis.

Opened https://www.pivotaltracker.com/story/show/153184647 for this optional item.

lucafavatella: > Can this ever be undefined?

Opened https://www.pivotaltracker.com/story/show/153184755 for tracking this.

PR #402, opened 2017-11-21
PT-153072576 Fix target recalculation for mining

zp-sd:
* Increase target when mining is too slow
* Decrease target when mining is too fast
* [Finishes #153072576]


lucafavatella:
git clone failed on all jobs of https://travis-ci.org/aeternity/epoch/builds/305712393?utm_source=github_status&utm_medium=notification , I suspect network glitch on Travis side: I restarted that CI build.

zp-sd:
@lucafavatella From what I see Push Travis job is failing because it tries to fetch branch from epoch repo, and I'm PRing from my fork.

lucafavatella:
> From what I see Push Travis job is failing because it tries to fetch branch from epoch repo, and I'm PRing from my fork.

@zp-sd I see that this PR is based on recent master (i.e. without recently-merged 384) so the push CI job would have tested the same thing as the pr one (that passed).

PR #403, opened 2017-11-23
Pt 152484212 fill in missing blocks 2 [Fixes #152484212]

uwiger:
* Sync protocol now asks chain (via `aec_miner`) for missing blocks to fetch
* `aec_sync` calls on `aec_miner` to store the fetched blocks, rather than going directly to the chain
* `aecore_sync_SUITE` is refactored, now creating the test target nodes in the CT `log_private` directory instead of using `make multi-build`. This speeds up test prep significantly
* the sync_SUITE proxy now starts much earlier (by injecting itself as a `setup` hook)
* the sync_SUITE code patching is removed, now using the normal test-compiled code
* `aec_events` now packs events in a map, adding sender and time info
*  the sync_SUITE proxy collects all events, allowing the test case to catch up. This reduces tricky timing issues, where the test case previously could get in too late to catch a published event.


dincho:
@uwiger Could you please remove `git rev-parse` from:
- https://github.com/aeternity/epoch/pull/403/files#diff-b67911656ef5d18c4ae36cb6741b7965R188
- https://github.com/aeternity/epoch/pull/403/files#diff-b67911656ef5d18c4ae36cb6741b7965R192

as it should not be needed anymore because of https://github.com/aeternity/epoch/pull/403/files#diff-31d7a50c99c265ca2793c20961b60979R68

apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -121,10 +121,10 @@ generate_single(Hash, Nonce, Target) ->
     %% -s makes mean miner print out the solution. We do not make it
     %% Algo-dependent in order to avoid wiring-in executable names. Lean miner
     %% will complain but ignores it.
-    ?info("Executing cmd: ~p~n", [lists:concat(export_ld_lib_path() ++ ["./", Exe, " -h ", Hash, " -n ", Nonce, " -s ", Extra])]),
```
hanssv: I think this change makes the comment above stale...

uwiger: True...

apps/aecore/src/aec_chain_state.erl
```diff
@@ -279,6 +282,14 @@ node_root_hash(#node{type = block , content = X}) -> aec_blocks:root_hash(X).
 node_target(#node{type = header, content = X}) -> aec_headers:target(X);
 node_target(#node{type = block , content = X}) -> aec_blocks:target(X).
 
+maybe_add_genesis_hash(#{genesis_block_hash := undefined} = State, Node) ->
+    case node_height(Node) =:= 0 of
+        true  -> State#{genesis_block_hash => hash(Node)};
```
lucafavatella: I admit I still need to get a full understanding of this PR, but it looks like this change (that is in commit ["Update genesis hash on header insert as well"](https://github.com/aeternity/epoch/pull/403/commits/6847de2852f162d342e8fe2a850915fae044914e)) makes it possible for a peer to induce this node to amend its understanding of what *the* genesis block is. That looks not correct.

(Related tickets are [PT-153184647](https://www.pivotaltracker.com/story/show/153184647) and [PT-153184755](https://www.pivotaltracker.com/story/show/153184755)).

apps/aehttp/src/aehttp_app.erl
```diff
@@ -11,7 +11,7 @@
 -define(DEFAULT_SWAGGER_EXTERNAL_PORT, 8043).
 -define(DEFAULT_SWAGGER_INTERNAL_PORT, 8143).
 -define(DEFAULT_WEBSOCKET_INTERNAL_PORT, 8144).
--define(INT_ACCEPTORS_POOLSIZE, 100).
+-define(INT_ACCEPTORS_POOLSIZE, 10).
```
lucafavatella: I see that this change gets lost after regenerating Swagger, though this seems inevitable at the moment as it is [hardcoded in upstream codegen](https://github.com/swagger-api/swagger-codegen/blob/28467a5c3f86eac33aa5391a024f4fbf62364646/modules/swagger-codegen/src/main/resources/erlang-server/server.mustache#L4). Ok.

apps/aehttp/src/swagger/swagger_server.erl
```diff
@@ -1,7 +1,7 @@
 -module(swagger_server).
 
 
--define(DEFAULT_ACCEPTORS_POOLSIZE, 100).
+-define(DEFAULT_ACCEPTORS_POOLSIZE, 10).
```
lucafavatella: I see that this change gets lost after regenerating Swagger, though this seems inevitable at the moment as it is [hardcoded in upstream codegen](https://github.com/swagger-api/swagger-codegen/blob/28467a5c3f86eac33aa5391a024f4fbf62364646/modules/swagger-codegen/src/main/resources/erlang-server/server.mustache#L4). Ok.

apps/aecore/src/aec_chain_state.erl
velzevur: @lucafavatella posting a new genesis block [rejection is handled](https://github.com/aeternity/epoch/blob/3539d7b89b0eb37b3e7cf4cb754cdcc8d6cb7e2b/apps/aecore/src/aec_chain_state.erl#L458) before that

apps/aecore/src/aec_chain_state.erl
```diff
@@ -323,6 +334,30 @@ export_block(#node{type = block, hash = H, content = B}, State) ->
         error -> B
     end.
 
+%%%-------------------------------------------------------------------
+%%% Find missing blocks
+%%%-------------------------------------------------------------------
+
+get_missing_block_hashes(undefined, _, _State) ->
```
lucafavatella: It looks to me this clause shall be deleted (in this PR or later), and rather assertions shall be added [at the end of the init of the aec_chain server state](https://github.com/aeternity/epoch/blob/3539d7b89b0eb37b3e7cf4cb754cdcc8d6cb7e2b/apps/aecore/src/aec_chain_server.erl#L75) that:
* Field top_header_hash is a binary (and not undefined) - on the basis that chain contains at least genesis block
* Field top_block_hash is a binary (and not undefined) - on the basis that chain contains at least genesis block

apps/aecore/src/aec_chain_state.erl
```diff
@@ -323,6 +334,30 @@ export_block(#node{type = block, hash = H, content = B}, State) ->
...
+    [];
+get_missing_block_hashes(TopHash, undefined, State) ->
```
lucafavatella: Similarly, I do not see how top_block_hash may be undefined after correct init.

config/dev1/sys.config
```diff
@@ -7,7 +7,6 @@
       {internal, [
           {swagger_port, 3113},
           {websocket, [ {port, 3114},
-                        {handlers, 100},
```
velzevur: Removing this leads to either using the YAML setup or [using the hardcoded default value](https://github.com/aeternity/epoch/blob/master/apps/aehttp/src/aehttp_app.erl#L14). I like it better this way.

apps/aecore/src/aec_conductor.erl
```diff
@@ -628,13 +658,14 @@ post_block_worker(From, Block) ->
             end
     end.
 
+handle_synced_block_reply({block_added, _Block, From}, State) ->
+    common_block_added_reply(From, State);
+handle_synced_block_reply(Result, State) ->
+    handle_post_block_reply(Result, State).
+
 handle_post_block_reply({block_added, Block, From}, State) ->
     aec_events:publish(block_received, Block),
```
lucafavatella: I see the difference between post_block and synced_block is this event publishing. Ok.

apps/aehttp/src/swagger/swagger_server.erl
uwiger: Yes. I personally think that 100 is a ridiculously large acceptor pool, since it's always replenished, and therefore only needs to be large enough not to run out before the pool manager can start new acceptors.

uwiger:
After a comment from @velzevur, I decided to check for missing blocks also when top headers are the same (previously, nothing was done then). While working on that, I noted that some parameters were mixed up in a few function calls. This was likely also the cause of the latest test run failure on Travis (at least the error appeared in the log.)

I also added a 'test_node' element in the `Info` maps of the published events (only in the test proxy). This is just to make it easier to interpret the ct log output.

apps/aecore/src/aec_sync.erl
```diff
@@ -257,125 +277,106 @@ fetch_headers(PeerUri, GHash, Acc) ->
...
-                    fetch_next(aec_headers:prev_hash(Hdr),
-                               PeerUri, GHash, Acc1)
-            end;
+            %% we have a continuous header chain
+            headers_fetched(PeerUri, Acc);
         {error, _} ->
             lager:debug("we don't have the top block header,"
-                        " fetching block (~p)", [HdrHash]),
+                        " fetching block (~p)", [pp(HdrHash)]),
             Acc1 = fetch_block(HdrHash, PeerUri, Acc),
```
lucafavatella: Comment out of scope for this PR, as related to sync rather than filling in of missing blocks:
It looks like that, when syncing, the node does not download and validate the header chain first and *then* download the blocks, but rather go straight for the each block.
@gorillainduction Is this what you meant in your comment https://www.pivotaltracker.com/story/show/152484212/comments/182108262 ?

apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -214,14 +243,45 @@ mine_and_compare(N1, Config) ->
...
+    collect_sync_events(Nodes1),
+    expect_same_top(Nodes, 5),
     expect_same_tx(Nodes).
 
-expect_same(Nodes, Tries) when Tries > 0 ->
+collect_sync_events([]) ->
+    done;
+collect_sync_events(Nodes) ->
+    receive
+        {gproc_ps, event, chain_sync, Msg} ->
```
lucafavatella: Type `gproc_ps, event` --> `gproc_ps_event` ? Ref https://github.com/uwiger/gproc/blob/0.6.1/doc/gproc_ps.md#publish3

This is compatible with local failure I got.

lucafavatella: mmm but how can this pass on CI?

uwiger: I've just pushed a fix for this. There are several timing issues in the sync suite, not least whether the test case has time to start checking for a sync event before the event has already been published. This is why the test proxy saves events and the test case collects and checks past events first. So tests may well pass in the case where events are published before the check happens, even though the following receive pattern is wrong.

PR #404, opened 2017-11-23
Use correct conversion to binary in hash calc for PoW

lucafavatella:
Supersedes PR #393 .

### DIFF
--- a/apps/aecore/include/blocks.hrl
+++ b/apps/aecore/include/blocks.hrl
@@ -46,6 +46,5 @@
           pow_evidence = no_value :: aec_pow:pow_evidence()}).
 -type(header() :: #header{}).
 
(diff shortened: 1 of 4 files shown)


lucafavatella:
I see there is premined chain in repo used for some tests, and this PR would break that. I am investigating if I can amend tests not to need that premined chain, and get that PR in before this one.

Cc @velzevur

lucafavatella:
> I see there is premined chain in repo used for some tests, and this PR would break that. I am investigating if I can amend tests not to need that premined chain, and get that PR in before this one.

I am trying to understand how to run Python tests - PR #406.

velzevur:
There seems to be an issue with python integration codes running in CI, please discard the premined chain issue.

PR #408, opened 2017-11-23
PT-152718992 Validate received block target/difficulty

zp-sd:
* [Finishes #152718992]

### DIFF
--- a/apps/aecore/src/aec_block_genesis.erl
+++ b/apps/aecore/src/aec_block_genesis.erl
@@ -25,7 +25,8 @@
 %% API
 -export([genesis_header/0,
          genesis_block_as_deserialized_from_network/0,
-         genesis_block/0]).
+         genesis_block/0,
+         height/0]).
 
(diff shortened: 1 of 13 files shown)


lucafavatella:
@zp-sd Does this need to move the persisted chain to a distinct folder in the environments i.e. similarly to https://github.com/aeternity/epoch/pull/410 ?

zp-sd:
@lucafavatella The block structure will not change, but we will start to recalculate difficulty for every block/header (whereas we are doing this only every `RECALCULATE_DIFFICULTY_FREQUENCY` headers/blocks now). So validation of previously mined blocks may fail.

lucafavatella:
> So validation of previously mined blocks may fail.

Does this affect loading persisted chain?

Anyway - this seems change large enough for starting mining new chain.

zp-sd:
@lucafavatella Agree to start fresh.

@sennui What do you think?

PR #409, opened 2017-11-24
Pt 152807361 refactor miner (Work In Progress, will not merge)

gorillainduction:
I think this is getting ready for merging, but I want your opinions on it.

This is a refactoring, changing the general principle from the aec_miner gen_statem to the new aec_conductor gen_server. The overall server topology is not changed, this should be done in other tickets in order to be able to split work and minimize conflict.


apps/aecore/src/aec_conductor.erl
```diff
@@ -0,0 +1,601 @@
...
+        , handle_info/2
+        , terminate/2
+        , code_change/3]).
+
+-include("common.hrl").
+-include("blocks.hrl").
+
+-define(SERVER, ?MODULE).
+
+-define(FETCH_NEW_TXS_FROM_POOL, true).
```
happi: This define does not seem to be used. Should it be used in the default in the state record?

apps/aecore/src/aec_conductor.erl
```diff
@@ -0,0 +1,601 @@
...
+
+maybe_block_tag(Tag, #state{blocked_tags = B} = State) ->
+    case tag_concurrency(Tag) of
+        concurrent -> State;
+        singleton  -> State#state{blocked_tags = ordsets:add_element(Tag, B)}
+    end.
+
+tag_concurrency(create_block_candidate) -> singleton;
+tag_concurrency(mining)                 -> singleton;
+tag_concurrency(post_block)             -> concurrent;
```
happi: How will concurrent post block be handled when the chain is turned into a functional ADT instead of a server?

apps/aecore/src/aec_conductor.erl
gorillainduction: yes

apps/aecore/src/aec_conductor.erl
gorillainduction: Good point. Now it works, but it will probably break down then. The validation could be done in a worker, but the insert should probably be atomic.

apps/aecore/src/aec_conductor.erl
```diff
@@ -0,0 +1,601 @@
...
+               nonce = Nonce} = State#state.block_candidate,
+    Info = [{top_block_hash, State#state.seen_top_block_hash}],
+    aec_events:publish(start_mining, Info),
+    Fun = fun() -> {aec_mining:mine(Block, Nonce),
+                    State#state.block_candidate}
+          end,
+    dispatch_worker(mining, Fun, State).
+
+handle_mining_reply({{ok, Block},_Candidate}, State) ->
+    %% TODO: This should listen on some event instead
```
uwiger: Agreed, this should be handled with pub-sub

apps/aecore/src/aec_conductor.erl
```diff
@@ -0,0 +1,601 @@
...
+                         [{height, aec_blocks:height(Block)}]),
+    epoch_mining:info("Miner ~p finished with ~p ~n", [self(), {ok, Block}]),
+    State1 = save_mined_block(Block, State),
+    State2 = State1#state{block_candidate = undefined},
+    case preempt_if_new_top(State2) of
+        no_change         -> start_mining(State2);
+        {changed, State3} -> start_mining(State3)
+    end;
+handle_mining_reply({{error, no_solution}, Candidate}, State) ->
+    try exometer:update([ae,epoch,aecore,mining,retries], 1)
```
uwiger: I have in mind to add an `aec_metrics:try_update/2` function, since this is a very common pattern.

apps/aecore/src/aec_conductor.erl
```diff
@@ -0,0 +1,601 @@
...
+    retry_mining_with_new_nonce(Candidate, State).
+
+retry_mining_with_new_nonce(#candidate{nonce = N, max_nonce = N}, State) ->
+    epoch_mining:info("Failed to mine block, "
+                      "nonce range exhausted (was ~p)", [N]),
+    start_mining(State#state{block_candidate = undefined});
+retry_mining_with_new_nonce(Candidate, State) ->
+    Nonce = Candidate#candidate.nonce,
+    NewNonce = aec_pow:next_nonce(Nonce),
+    epoch_mining:info("Not fetching new txs; "
```
uwiger: Should be a debug message, or perhaps a metric update

gorillainduction: I'll change it

gorillainduction:
Dropping this PR and will create a new one with the squashed commit.

PR #410, opened 2017-11-24
Reset chain in integration environment following PoW change

lucafavatella:
... in 4332c07 as part of PT-152917447.

See also https://www.pivotaltracker.com/story/show/152956790

### DIFF
--- a/deployment/ansible/deploy.yml
+++ b/deployment/ansible/deploy.yml
@@ -8,7 +8,7 @@
     config:
       chain:
         persist: true
-        db_path: .
+        db_path: "."
 
   tasks:
     - name: "Required variables"
--- a/deployment/ansible/deploy.yml
+++ b/deployment/ansible/deploy.yml
@@ -65,6 +65,13 @@
       when: local_package is defined
       tags: [package]
 
+    - name: "Make sure chain path ( {{ config.chain.db_path }} ) exists"
+      file:
+        path: "{{ config.chain.db_path }}"
+        state: directory
+        mode: 0600
+      tags: [config]
+
     - name: Update configuration
       template:
         src: templates/epoch.yaml.j2
--- a/deployment/ansible/deploy.yml
+++ b/deployment/ansible/deploy.yml
@@ -8,7 +8,7 @@
     config:
       chain:
         persist: true
-        db_path: "."
+        db_path: "./db2"
 
   tasks:
     - name: "Required variables"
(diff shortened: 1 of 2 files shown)


dincho:
Adding:
```

    - name: "Make sure chain path exists"
      file:
          path: "{{ project_root }}/{{ config.chain.db_path }}"
          state: directory
      tags: [config]

**Comment by lucafavatella** [comment]
A CI job failed while installing python dep: restarting.

**Comment by lucafavatella** [comment]
> Adding this before config should work:

@dincho  Thanks - I had created proposal locally so I will compare and integrate this PR.

File: deployment/ansible/deploy.yml [hunk 1]
```diff
@@ -8,7 +8,7 @@
     config:
       chain:
         persist: true
-        db_path: .
+        db_path: "./2"
```
dincho: I would prefer some more descriptive name like "chaindb2" or "db/2" or "chaindb/2" or similar

lucafavatella:
> @dincho Thanks - I had created proposal locally so I will compare and integrate this PR.

Done.

deployment/ansible/deploy.yml
```diff
@@ -65,6 +65,13 @@
       when: local_package is defined
       tags: [package]
 
+    - name: "Make sure {{ db_path }} exists"
```
dincho: There is no such variable.
Also if it's set to "." it will read: `"Make sure . exists"` which looks ugly to me.
That's why I suggested hardcoded term instead of var.

dincho: I just tested it and it reads "Make sure {{ db_path }} exists" so it's working anyway.
Change only if you want.

lucafavatella:
I see two approvals.
This PR does not affect CI in PR - as deploy script runs only on master and tags. Anyway all Linux jobs completed successfully (OSX jobs not started). Merging.

deployment/ansible/deploy.yml
```diff
@@ -65,6 +65,13 @@
       when: local_package is defined
       tags: [package]
 
+    - name: "Make sure chain path ( {{ config.chain.db_path }} ) exists"
+      file:
+        path: "{{ config.chain.db_path }}"
```
dincho: @lucafavatella We missed the {{ project_root }} prefix here causing to create the directory in wrong parent

dincho: https://github.com/aeternity/epoch/pull/415

PR #411, opened 2017-11-24
Tarball revision

dincho:
- [x] Add ENV file to a deployed node

### DIFF
--- a/ci
+++ b/ci
@@ -19,12 +19,12 @@ cd $BuildDir
 
 case "${1:?}"-"${2:?}" in
     before_install-*)
-        if [ "${TRAVIS_OS_NAME}" = "osx" ]; then
+        if [ "${TRAVIS_OS_NAME:?}" = "osx" ]; then
             export HOMEBREW_CACHE="$HOME/Library/Caches/Homebrew"
             brew update
             brew install erlang && export PATH="$(brew --prefix erlang)"/bin:"$PATH"
             brew install lynx && export PATH="$(brew --prefix lynx)"/bin:"$PATH"
-        elif [ "${TRAVIS_OS_NAME}" = "linux" ]; then
+        elif [ "${TRAVIS_OS_NAME:?}" = "linux" ]; then
             sudo apt-get -y install lynx
             sudo -H pip install -r deployment/ansible/pip-requirements.txt
             ansible --version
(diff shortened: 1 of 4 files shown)


ci
```diff
@@ -3,29 +3,30 @@
...
+    OsLabel="osx-$(sw_vers -productVersion)"
+fi
 
-Version="$(cat VERSION)"
+Step=${1:?}
+Job=${2:?}
+BuildDir=${3:?}
+DeployEnv=${4:-test}
+Version=$(cat VERSION)
+BuiltPackagePath="${BuildDir}"/_build/prod/rel/epoch/epoch-"${Version}".tar.gz
```
lucafavatella: Please use `:?` for avoiding silent errors.

lucafavatella: i.e. `${Version:?}`

lucafavatella: And I would suggest referring to all vars with `${Var:?}` unless there is a good reason not to - otherwise updating the file becomes error prone.

dincho:
> Please confirm that your comment Add ENV file to a deployed node is not in scope of this PR and that it is tracked in some other ticket.

It's in scope of this one, I'll do it in a while.

ci
dincho: I'll make sure $Version is always valid. 

However in regards to other variables they all have guards during declaration e.g. BuildDir=${3:?} this guarantee that BuildDir is set and "non null" below in the script.

I intentionally removed ${Var:?} all over the script because it adds additional (useless IMO) noise in the script and decrease it's readability.

If you see some downsides or use cases about the above, please let me know so we can change it accordingly.

dincho:
> otherwise updating the file becomes error prone.

Do you mean that someone can change the "declaration" of `BuildDir=${3:?}` and make it possible to brake below during execution?

ci
lucafavatella: If the next developer writes $BuildDIR (notice case change) will get silent error.

lucafavatella: Not a too strong opinion though - if not seen as common best practice no problem in doing as you suggest.

lucafavatella:
> Do you mean that someone can change the "declaration" of BuildDir=${3:?} and make it possible to brake below during execution?

No - that someone could mistype the *use* of BuildDir

dincho:
> Please confirm that your comment Add ENV file to a deployed node is not in scope of this PR and that it is tracked in some other ticket.
> It's in scope of this one, I'll do it in a while.

see 1753ae9

ci
dincho: done in 3682f50, let know if you see something missing

lucafavatella:
I cleaned up pending useless CI OSX jobs, CI shall be a bit faster now.

PR #412, opened 2017-11-24
[Finishes #152807361] Replace aec_miner with aec_conductor

gorillainduction:
Redesign of the central hub of the mining process.

- Use a gen_server rather than a gen_statem for a better overview of
the flow,
- Use monitored worker processes to keep the server responsive.
- Make starting/stopping mining preemptive,
- Add some events for external visibility of the inner workings
- Enable breaking up the code in separate modules for the workers
- Enable joining in other services (e.g., aec_chain_server,
  aec_tx_pool) in the same server.


apps/aecore/src/aec_conductor.erl
```diff
@@ -0,0 +1,644 @@
...
+        , handle_info/2
+        , terminate/2
+        , code_change/3]).
+
+-include("common.hrl").
+-include("blocks.hrl").
+
+-define(SERVER, ?MODULE).
+
+-define(FETCH_NEW_TXS_FROM_POOL, true).
```
happi: Still unused...

gorillainduction: I'll fix

PR #413, opened 2017-11-24
Make crashes in persistence initialization self-descriptive

lucafavatella:
During preparation of PR #410 I noticed that, when db_path is not present, node crashes with enoent but without having the affected path in the crash. Hence this PR.

This should help users.


happi:
We should have a ticket for creating all subdirs on th path.

lucafavatella:
> We should have a ticket for creating all subdirs on th path.

@happi For tracking https://www.pivotaltracker.com/story/show/153171500

lucafavatella:
Ran `epmd -daemon && make test` on clean working copy with merged recent master and PR 399 (7cf6f734).
I see two reviewers approved. CI passed. Merging.

PR #418, opened 2017-11-27
PT#152471813 - Implement Cuckoo PoW verification in Erlang

hanssv:
The verification is a simple path traversal in a graph, the graph is
defined by the output of SipHash, and the SipHash keys are defined by
the BLAKE2b hash of the Header(hash) and the Nonce.

BLAKE2b and SipHash implementations have been verified by running a
QuickCheck property against reference implementations in C. A couple of
EUnit tests are included here for completeness.


apps/aecore/src/aec_headers.erl
```diff
@@ -15,8 +15,6 @@
          deserialize_from_store/1,
          serialize_to_map/1,
          deserialize_from_map/1,
-         serialize_to_binary/1,
```
lucafavatella: Thanks.

apps/aecore/src/aec_siphash.erl
```diff
@@ -0,0 +1,113 @@
+%%%=============================================================================
```
lucafavatella: This seems quite a self-contained library: have you considered splitting this in a separate aeternity/something repo?

apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -141,49 +142,105 @@ generate_single(Hash, Nonce, Target) ->
...
+    {K0, K1} = aec_siphash:create_keypair(Hash, Nonce),
+    EdgeMask = (1 bsl (Size - 1)) - 1,
+    try
+        %% Generate Uv pairs representing endpoints by hashing the proof
+        %% XOR points together: for a closed cycle they must match somewhere
+        %% making one of the XORs zero.
+        {Xor0, Xor1, _, Uvs} =
+            lists:foldl(
+              fun(N, _) when N > EdgeMask ->
+                      throw(?POW_TOO_BIG);
```
lucafavatella: Is throw/catch used only for quitting early this foldl? Removing the catch would ease reasoning about the code for developers (and for tools like dialyzer).

apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -141,49 +142,105 @@ generate_single(Hash, Nonce, Target) ->
...
+                 (N, {_Xor0, _Xor1, PrevN, _Uvs}) when N =< PrevN ->
+                      throw(?POW_TOO_SMALL);
```
lucafavatella: `?POW_TOO_BIG` and `?POW_TOO_SMALL` look specific to an N but that N is not reported in the error. Shall it be included in the tuples?

apps/aecore/src/aec_blake2b.erl
```diff
@@ -0,0 +1,143 @@
+%%%=============================================================================
+%%% @copyright (C) 2017, Aeternity Anstalt
+%%% @doc
+%%%  BLAKE2b implementation in Erlang - for details see: https://blake2.net
```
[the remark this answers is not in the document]
hanssv: I thought of making a separate entity, but it is so small I didn't know if it was worth the effort...

I picked a few of those test vectors for the unit tests.

I've also written the QuickCheck property that checks for all messages and keys and hash lengths that the C reference implementation gives the same hash as the Erlang implementation.

apps/aecore/src/aec_pow_cuckoo.erl
hanssv: Good point, that could be an improvement.

apps/aecore/src/aec_siphash.erl
hanssv: Here I don't think a separate repo is merited, this is just an implementation of a special case of SipHash, not the general thing!

apps/aecore/src/aec_pow_cuckoo.erl
hanssv: I inherited this part from Akos, didn't really think of the style but I think you have a point that the code could be easier to follow without the throw/catch.

sennui:
Consider moving to new utility app or to aecutils app that is also general lib app in our project. No need for separate repository.

hanssv:
Moved the hash functions to aeutils...

hanssv:
Will wait with this commit until after #391 - this will have to be slightly reworked to fit on top of that...

apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -144,57 +144,119 @@ generate_int(Header, Target) ->
...
+                end;
+            _ ->
+                %% matching endpoints imply zero xors
+                throw(?POW_NON_MATCHING)
+        end
     catch
-        C:E ->
-            {error, {unknown, {C, E}}}
+        throw:{error, Reason} ->
+            epoch_pow_cuckoo:error("Proof verification failed for ~p: ~p", [Solution, Reason]),
```
lucafavatella: Logging hash and nonce too would help.

PR #419, opened 2017-11-27
Log only to file - not console [Finishes #153173585]

lucafavatella:
... following experience of miner hung on lager.

Also console logging is not needed.


lucafavatella:
I plan to use it locally for a while before proposing it for inclusion.

lucafavatella:
I see two reviewers approved. CI passed.
I locally merged recent master and: ran `make test` (passed), and build production release and checked logs produced.
Merging.

PR #420, opened 2017-11-28
Add 2 dev environments

dincho:


.travis.yml
```diff
@@ -42,6 +42,22 @@ deploy:
       tags: true
       all_branches: true # Workaround for Travis - see https://github.com/travis-ci/travis-ci/issues/1675#issuecomment-37851765
       condition: "${JOB:?} = package && ${TRAVIS_OS_NAME:?} = linux"
+  - provider: script
+    script: ./ci deploy "${JOB:?}" "${TRAVIS_BUILD_DIR:?}" dev1
+    skip_cleanup: true
+    on:
+      # See https://docs.travis-ci.com/user/deployment#Conditional-Releases-with-on%3A
+      repo: aeternity/epoch
+      branch: env/dev1
```
michalzajda: should we protect the branch in gh?

dincho:
> should we protect the branch in GH?

I don't think so, it will get messy at some point so force push would be good option, otherwise the branch must be recreated.

PR #421, opened 2017-11-28
[WIP] Review mining logs in aec_conductor

lucafavatella:
I prepared this while working on PR #419.

This PR conflicts with PR #403 hence is best merged after that (after rebasing).

----

I plan to run this locally for a while before considering it ready for merge -(hence the WIP tag in the title and the assignment to me).

### DIFF
--- a/config/dev1/sys.config
+++ b/config/dev1/sys.config
@@ -50,7 +50,7 @@
              {handlers, [
                {lager_file_backend, [
                   {file, "log/epoch_mining.log"},
-                  {level, info}
+                  {level, debug}
                ]}
              ]}
            ]},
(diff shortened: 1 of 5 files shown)


config/dev1/sys.config
```diff
@@ -50,7 +50,7 @@
              {handlers, [
                {lager_file_backend, [
                   {file, "log/epoch_mining.log"},
-                  {level, info}
+                  {level, debug}
```
uwiger: Isn't `info` a better hard-coded default? If we want `debug` for the testnet, we should really have a user config to set the loglevel for the mining log. But for now, I guess this is ok.

lucafavatella:
I would need to reconsider and distill what to log at level info. Thanks for your feedback.

PR #422, opened 2017-11-28
PT #153206821 - Change the PoW graph size to 2^28

hanssv:
For the testnet lean30 is a bit over the top and the difficulty/target
adjustment never kicks in. Graph size 2^28 should be more suitable until
we reach thousands of nodes...


hanssv:
This should also be reflected on the Wiki

deployment/ansible/deploy.yml
```diff
@@ -9,7 +9,7 @@
     config:
       chain:
         persist: true
-        db_path: "./db4"
+        db_path: "./db6"
```
lucafavatella: Depending on which PR is merged first, please double check this value in the other PR.

PR #423, opened 2017-11-28
HTTP API account/balance handling invalid addresses

velzevur:
[Pivotal task](https://www.pivotaltracker.com/story/show/153214980)

### DIFF
--- a/apps/aehttp/src/aehttp_dispatch_ext.erl
+++ b/apps/aehttp/src/aehttp_dispatch_ext.erl
@@ -128,17 +128,24 @@ handle_request('GetAccountBalance', Req, _Context) ->
           undefined ->
               {ok, PK} = aec_keys:pubkey(),
               PK;
-          PK ->
-              base64:decode(PK)
+          PK when is_binary(PK) ->
+              try base64:decode(PK)
+              catch _:_ -> not_base64_encoded
+              end
       end,
-    {ok, LastBlock} = aec_chain:top(),
-    Trees = aec_blocks:trees(LastBlock),
-    AccountsTree = aec_trees:accounts(Trees),
-    case aec_accounts:get(Pubkey, AccountsTree) of
-        {ok, #account{balance = B}} ->
-            {200, [], #{balance => B}};
-        _ ->
-            {404, [], #{reason => <<"Account not found">>}}
+    case Pubkey of
+        not_base64_encoded ->
+            {400, [], #{reason => <<"Invalid address">>}};
+        _ when is_binary(Pubkey) ->
+            {ok, LastBlock} = aec_chain:top(),
+            Trees = aec_blocks:trees(LastBlock),
+            AccountsTree = aec_trees:accounts(Trees),
+            case aec_accounts:get(Pubkey, AccountsTree) of
+                {ok, #account{balance = B}} ->
+                    {200, [], #{balance => B}};
+                _ ->
+                    {404, [], #{reason => <<"Account not found">>}}
+            end
     end;
 
 handle_request(OperationID, Req, Context) ->
--- a/apps/aehttp/priv/swagger.json
+++ b/apps/aehttp/priv/swagger.json
@@ -219,6 +219,12 @@
               "$ref" : "#/definitions/Balance"
             }
           },
+          "400" : {
+            "description" : "Invalid address",
+            "schema" : {
+              "$ref" : "#/definitions/Error"
+            }
+          },
           "404" : {
             "description" : "Account not found",
             "schema" : {
--- a/apps/aehttp/src/swagger/swagger_api.erl
+++ b/apps/aehttp/src/swagger/swagger_api.erl
@@ -202,6 +202,8 @@ populate_request_param(OperationID, Name, Req0, ValidatorState) ->
 
 validate_response('GetAccountBalance', 200, Body, ValidatorState) ->
     validate_response_body('Balance', 'Balance', Body, ValidatorState);
+validate_response('GetAccountBalance', 400, Body, ValidatorState) ->
+    validate_response_body('Error', 'Error', Body, ValidatorState);
 validate_response('GetAccountBalance', 404, Body, ValidatorState) ->
     validate_response_body('Error', 'Error', Body, ValidatorState);
 
--- a/config/swagger.yaml
+++ b/config/swagger.yaml
@@ -192,6 +192,10 @@ paths:
           description: Account's balance
           schema:
             $ref: '#/definitions/Balance'
+        '400':
+          description: Invalid address
+          schema:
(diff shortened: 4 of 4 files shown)


velzevur:
I couldn't chose between them. Suggestions are welcome

hanssv:
I would go for "Invalid address" in this case

velzevur:
Swagger description updated to have a consistent message `Invalid address`

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -129,16 +129,23 @@ handle_request('GetAccountBalance', Req, _Context) ->
               {ok, PK} = aec_keys:pubkey(),
               PK;
           PK ->
-              base64:decode(PK)
+              try base64:decode(PK)
+              catch error:{badarg, _} -> not_base64_encoded
```
lucafavatella: Ok.

----

I tested locally:
```
4> f(), try base64:decode(<<"foo">>) catch E:R -> {E, R} end.
{error,{badmatch,false}}
```

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -129,16 +129,23 @@ handle_request('GetAccountBalance', Req, _Context) ->
...
-    {ok, LastBlock} = aec_chain:top(),
-    Trees = aec_blocks:trees(LastBlock),
-    AccountsTree = aec_trees:accounts(Trees),
-    case aec_accounts:get(Pubkey, AccountsTree) of
-        {ok, #account{balance = B}} ->
-            {200, [], #{balance => B}};
+    case Pubkey of
+        not_base64_encoded ->
+            {400, [], #{reason => <<"Invalid address">>}};
         _ ->
```
lucafavatella: This open case is fragile as the atom of the clause above may be mistyped by mistake. I suggest adding guard (e.g. `when is_binary(...)`).

velzevur: One can never be too assertive, I've pushed a change for this

apps/aehttp/src/aehttp_dispatch_ext.erl
lucafavatella: (Just to be clear: no action on this comment - this comment states that this line looks correct.)

lucafavatella: I correct myself: there is issue. base64:decode returns two kinds of errors: badmatch and badarg.

lucafavatella:
```
6> f(), try base64:decode(<<"bob_address">>) catch E:R -> {E,R} end.
{error,{badarg,95}}
7> f(), try base64:decode(<<"bobaddress">>) catch E:R -> {E,R} end.
{error,{badmatch,false}}
```

lucafavatella: So in one case 400 is returned, in other case 500 is returned.
```
$ curl -v http://localhost:3013/v1/account/balance?pub_key=bobaddress
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 3013 (#0)
> GET /v1/account/balance?pub_key=bobaddress HTTP/1.1
> Host: localhost:3013
> User-Agent: curl/7.43.0
> Accept: */*
>
< HTTP/1.1 500 Internal Server Error
< server: Cowboy
< date: Wed, 29 Nov 2017 12:09:18 GMT
< content-length: 0
< content-type: application/json
<
* Connection #0 to host localhost left intact

$ curl -v http://localhost:3013/v1/account/balance?pub_key=bob_address
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 3013 (#0)
> GET /v1/account/balance?pub_key=bob_address HTTP/1.1
> Host: localhost:3013
> User-Agent: curl/7.43.0
> Accept: */*
>
< HTTP/1.1 400 Bad Request
< server: Cowboy
< date: Wed, 29 Nov 2017 12:09:21 GMT
< content-length: 28
< content-type: application/json
<
* Connection #0 to host localhost left intact
```

lucafavatella: So this PR catches the case when the pubkey specified contains chars not in the base64 range. Though does not catch the case that the chars are in the base64 range though it cannot be decoded.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -128,17 +128,24 @@ handle_request('GetAccountBalance', Req, _Context) ->
           undefined ->
               {ok, PK} = aec_keys:pubkey(),
               PK;
-          PK ->
-              base64:decode(PK)
+          PK when is_binary(PK) ->
+              try base64:decode(PK)
+              catch _: _ -> not_base64_encoded
```
hanssv: ugly space after ':'

velzevur: ugly space is no more

velzevur:
@lucafavatella does this account has a balance? We create accounts when they actually receive a balance so while there is no coinbase reward - this account is not created.

lucafavatella:
> does this account has a balance? We create accounts when they actually receive a balance so while there is no coinbase reward - this account is not created.

Yeah - I just figured out locally.

lucafavatella:
I see positive case works.

I set mining autostart, then:
```
$ curl http://localhost:3113/v1/account/pub-key
{"pub_key":"BMx8Lfj42Q7vkESMdF8a0Z01JSc/jwyUGzfPcwZYN9zO/X+8fV6CAmm1Od0OqxWq7SFYv/astoPFGWUzNyPvL8I="}

$ curl -v -G http://localhost:3013/v1/account/balance --data-urlencode "pub_key=BMx8Lfj42Q7vkESMdF8a0Z01JSc/jwyUGzfPcwZYN9zO/X+8fV6CAmm1Od0OqxWq7SFYv/astoPFGWUzNyPvL8I="
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 3013 (#0)
> GET /v1/account/balance?pub_key=BMx8Lfj42Q7vkESMdF8a0Z01JSc%2FjwyUGzfPcwZYN9zO%2FX%2B8fV6CAmm1Od0OqxWq7SFYv%2FastoPFGWUzNyPvL8I%3D HTTP/1.1
> Host: localhost:3013
> User-Agent: curl/7.43.0
> Accept: */*
>
< HTTP/1.1 200 OK
< server: Cowboy
< date: Wed, 29 Nov 2017 13:48:52 GMT
< content-length: 15
< content-type: application/json
<
* Connection #0 to host localhost left intact
{"balance":100}
```

PR #424, opened 2017-11-28
Introduce strict validation of user configs, terminate node if it fails

uwiger:
Validation is run as a `setup` phase hook. These can be wired to
abort processing through `-setup abort_on_error true`. However, due
to a bug in `setup`, this didn't propagate all the way to an aborted
startup. This bug was fixed, and the `setup` dependency upgraded.

The schema has been upgraded to have "additionalProperties": false
at each level.


uwiger:
[Finishes #153138074]

PR #425, opened 2017-11-28
Log location where default config is being looked for

lucafavatella:
Tested alongside PR 424 and 403. (I had placed invalid config in wrong location, so I logged which locations where looked for, hence this PR.)


lucafavatella:
Two reviewers approved. CI passed. I had tested merged with relevant PRs so not expecting conflicts. Merging.

PR #428, opened 2017-11-29
PT-153251639 Use target from previous block in target validation

hanssv:
https://www.pivotaltracker.com/story/show/153251639

### DIFF
--- a/apps/aecore/src/aec_target.erl
+++ b/apps/aecore/src/aec_target.erl
@@ -3,7 +3,7 @@
 %% API
 -export([recalculate/2,
          determine_delta_header_height/1,
-         verify/2]).
+         verify/3]).
 
 -include("common.hrl").
 -include("blocks.hrl").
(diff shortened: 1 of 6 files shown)


apps/aecore/src/aec_target.erl
```diff
@@ -35,10 +35,15 @@ recalculate(Header, InitialHeader) ->
...
-verify(Header, InitialHeader) ->
+-spec verify(header(), header(), header()) -> ok | {error, target_too_high}.
+verify(Header, PrevHeader, InitialHeader) ->
     HeaderTarget = aec_headers:target(Header),
-    ExpectedTarget = recalculate(Header, InitialHeader),
+    PrevTarget   = aec_headers:target(PrevHeader),
+
+    %% Block candidate target was computed using the target of the previous block
+    %% we have to do the same thing here.
+    ExpectedTarget = recalculate(Header#header{ target = PrevTarget }, InitialHeader),
```
happi: Ì am slightly uneasy about the use of the header record directly here.
Perhaps create a set_target function in aec_header.erl

lucafavatella:
> This is not correct. The delta header is calculated from the header when it should be calculated from the prev header if I understand the situation right.

In mining, [the delta header is computed from the block with the height already incremented](https://github.com/aeternity/epoch/blob/420973b7386c77aee29e9334bd7e00f193b87bed/apps/aecore/src/aec_mining.erl#L104), so this looks correct to me.

lucafavatella:
@hanssv Does this PR need to bump the chain DB folder? (reviewing other PR right now, did not double check)

apps/aecore/src/aec_target.erl
```diff
@@ -35,10 +35,16 @@ recalculate(Header, InitialHeader) ->
...
-verify(Header, InitialHeader) ->
+-spec verify(header(), header(), header()) -> ok | {error, target_too_high}.
+verify(Header, PrevHeader, InitialHeader) ->
     HeaderTarget = aec_headers:target(Header),
-    ExpectedTarget = recalculate(Header, InitialHeader),
+    PrevTarget   = aec_headers:target(PrevHeader),
+
+    %% Block candidate target was computed using the target of the previous block
+    %% we have to do the same thing here.
+    ExpectedTarget = recalculate(aec_headers:set_header(Header, PrevTarget),
```
lucafavatella: set_target?

hanssv:
Fixed typo and bumped the chain

lucafavatella:
I killed the 4 OSX jobs of this PR as not needed - PR needs to be updated with test fix anyway.

PR #431, opened 2017-11-29
Update JSON schema

velzevur:
[Pivotal task](https://www.pivotaltracker.com/story/show/152959514)
We expect a `peer_address` to be filled in the YAML file, this address is the one peers are going to use to ping and post blocks to.
If the address is `SCHEMA://HOST:PORT/` according to what is provided the node behaved differently
```
N    |  SCHEMA          |  HOST         |  PORT 
1    |   yes            |  yes          |  yes
2    |   no             |  yes          |  yes
3    |   no             |  yes          |  no
4    |   yes            |  yes          |  no
```
All cases except `1`  and `3` were resulting in a node not starting.

This PR addresses case `4` - provided a `SCHEMA` and `HOST` we use the port of the extenal API to setup the node as a default `PORT` value.
This PR does not address case `2` (when no `SCHEMA` is provided and there is a `PORT`) - node will still crash in this case.

Note: `aec_peers:uri_from_scheme_ip_port/2,3` works with both IPs and host names and should probably be renamed. I haven't investigated it and it is not part of this PR.

### DIFF
--- a/apps/aeutils/priv/epoch_config_schema.json
+++ b/apps/aeutils/priv/epoch_config_schema.json
@@ -9,6 +9,7 @@
             "type"  : "array",
             "items" : {
                 "type"    : "string",
+                "pattern": "^http://[^:\\.\"!#$%^&*()',/]+(\\.[^:\\.\"!#$%^&*()',/]+)*:[0-9]+/$", 
                 "example" : "http://a.b.c:123/"
             }
         },
@@ -24,6 +25,7 @@
                             "description" :
                             "The peer address that the node will present itself with. If port part given, it must match 'port' user config.",
                             "type" : "string",
+                            "pattern": "^http://[^:\\.\"!#$%^&*()',/]+(\\.[^:\\.\"!#$%^&*()',/]+)*:[0-9]+/$", 
                             "example" : "http://a.b.c:123/"
                         },
                         "port" : {
@@ -39,7 +41,7 @@
                     "properties" : {
                         "port" : {
                             "description" :
-                            "Listen port for external HTTP interface.",
+                            "Listen port for internal HTTP interface.",
                             "type" : "integer"
                         }
                     }
--- a/apps/aecore/src/aec_peers.erl
+++ b/apps/aecore/src/aec_peers.erl
@@ -21,6 +21,7 @@
          get_random/1,
          get_random/2,
          uri_from_ip_port/2,
+         uri_from_scheme_ip_port/3,
          uri/1,
          set_local_peer_uri/1,
          get_local_peer_uri/0,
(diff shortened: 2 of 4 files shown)


apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -39,7 +39,7 @@
                     "properties" : {
                         "port" : {
                             "description" :
-                            "Listen port for external HTTP interface.",
+                            "Listen port for internal HTTP interface.",
```
lucafavatella: Thanks.

velzevur:
Yes, just putting host or IP works, it was my mistake; I've double checked this and updated the description above

velzevur:
Since being advised by @uwiger I've added a regex check for _some_ sanity of the the both `peers` and `local_peer`

If an address fails the regex check - the node will not start with an similar error
```
{"Kernel pid terminated",application_controller,"{application_start_failure,setup,{bad_return_value,{{setup_app,start_phase,[run_setup,normal,[]]},{'EXIT',{{{validation_failed,[{error,[{data_invalid,#{<<\"description\">> => <<\"The peer address that the node will present itself with. If port part given, it must match 'port' user config.\">>,<<\"example\">> => <<\"http://a.b.c:123/\">>,<<\"pattern\">> => <<\"^http://[^:\\.\\"!#$%^&*()',/]+(\\.[^:\\.\\"!#$%^&*()',/]+)*:[0-9]+/$\">>,<<\"type\">> => <<\"string\">>},no_match,<<\"http://this$is$invalid:3013/\">>,[<<\"http\">>,<<\"external\">>,<<\"peer_address\">>]}]}]},[{setup,run_setup,0,[{file,\"/home/vlz/projects/aeternity/FORKS/epoch/_build/default/lib/setup/src/setup.erl\"},{line,871}]},{setup_srv,handle_call,3,[{file,\"/home/vlz/projects/aeternity/FORKS/epoch/_build/default/lib/setup/src/setup_srv.erl\"},{line,42}]},{gen_server,try_handle_call,4,[{file,\"gen_server.erl\"},{line,636}]},{gen_server,handle_msg,6,[{file,\"gen_server.erl\"},{line,665}]},{proc_lib,init_p_do_apply,3,[{file,\"proc_lib.erl\"},{line,247}]}]},{gen_server,call,[setup_srv,run_setup,infinity]}}}}}}"}
```

apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -9,6 +9,7 @@
             "type"  : "array",
             "items" : {
                 "type"    : "string",
+                "pattern": "^http://[^:\\.\"!#$%^&*()',/]+(\\.[^:\\.\"!#$%^&*()',/]+)*:[0-9]+/$",
```
lucafavatella: Does this `\\.` (inside `[^` and `]`) mean no `\` and no `.`?

lucafavatella: So https is not allowed. So I understand https://github.com/aeternity/epoch/pull/431/files#diff-0769afb81ddf2fb23a397550262da25cR138 could be deleted.

velzevur: >Does this \\. (inside [^ and ]) mean no \ and no .?

yes, the host can not start with either `\` nor with a `.`

PR #432, opened 2017-11-29
implement retry in sync_SUITE call_proxy()

uwiger:
Fixes the occasional problem that the sync_SUITE calls the test_proxy before it has had a chance to start.

### DIFF
--- a/apps/aecore/test/aecore_sync_SUITE.erl
+++ b/apps/aecore/test/aecore_sync_SUITE.erl
@@ -461,13 +461,24 @@ subscribe(N, Event) ->
 unsubscribe(N, Event) ->
     call_proxy(N, {unsubscribe, Event}).
 
+-define(PROXY_CALL_RETRIES, 5).
+
 call_proxy(N, Req) ->
-    call_proxy(N, Req, 3000).
+    call_proxy(N, Req, ?PROXY_CALL_RETRIES, 3000).
 
 call_proxy(N, Req, Timeout) ->
+    call_proxy(N, Req, ?PROXY_CALL_RETRIES, Timeout).
+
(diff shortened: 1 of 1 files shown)


apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -461,21 +461,35 @@ subscribe(N, Event) ->
...
 call_proxy(N, Req) ->
-    call_proxy(N, Req, 3000).
+    call_proxy(N, Req, ?PROXY_CALL_RETRIES, 3000).
 
 call_proxy(N, Req, Timeout) ->
+    call_proxy(N, Req, ?PROXY_CALL_RETRIES, Timeout).
+
+call_proxy(N, Req, Tries, Timeout) when Tries > 0 ->
     Ref = erlang:monitor(process, {?PROXY, N}),
     {?PROXY, N} ! {self(), Ref, Req},
```
lucafavatella: It appears to me that if:
* At the time the `erlang:monitor` executes the proxy process is not there (hence DOWN with noproc will be received ); and
* At the time this line (`{?PROXY, N} ! {self(), Ref, Req}`) executes the proxy process is there

the reattempt `{?PROXY, N} ! {self(), Ref, Req}` (after consuming the DOWN noproc message) will leave a reply in the mailbox of this main test process (I did not notice attempts to clean mailbox). (The reply has the ref of the monitor hence does not match receive clause.)

These replies may accumulate, though only up to the number of retries to each call in the test (and this will only happen at the beginning of each monitored node anyway, so shall not be many).

uwiger: Yes, but since the test case process only lives for the duration of the test, I've figured that it's not worth the effort (or at least not a priority) to address this.

PR #433, opened 2017-11-30
Pt 153211733 remove unwanted peers

uwiger:
[Finishes #153211733] https://www.pivotaltracker.com/story/show/153211733

* New API: `aec_peers:block_peer(Uri)`, `aec_peers:unblock_peer(Uri)`
* Pings are not scheduled for blocked peers
* Blocked peers are excluded from broadcast lists
* A static blocklist can be specified in the user config (`"blocked_peers"`)
* A `chain_sync` event, `{sync_aborted, Reason}` is published when a ping results in an error.
* A peer is auto-blocked if it passes the wrong genesis block in a ping
* A pinging peer also blocks the other peer if it receives the ping response `"Different genesis blocks"` or `"Not allowed"`
* The sync_SUITE extended with one testcase where `dev2` is pre-blocked by `dev1`, verifying that the sync attempt is aborted.

### DIFF
--- a/apps/aecore/test/aecore_sync_SUITE.erl
+++ b/apps/aecore/test/aecore_sync_SUITE.erl
@@ -16,6 +16,7 @@
     start_first_node/1,
     start_second_node/1,
     start_third_node/1,
+    start_blocked_second/1,
     mine_on_first/1,
     mine_on_second/1,
     mine_on_third/1,
@@ -39,7 +40,8 @@ all() ->
 groups() ->
     [
      {all_nodes, [sequence], [{group, two_nodes},
-                              {group, three_nodes}]},
+                              {group, three_nodes},
+                              {group, one_blocked}]},
      {two_nodes, [sequence],
       [start_first_node,
        test_subscription,
@@ -61,7 +63,11 @@ groups() ->
        mine_again_on_first,
        mine_on_second,
        mine_on_third,
-       restart_third]}
+       restart_third]},
+      {one_blocked, [sequence],
+       [start_first_node,
+        mine_on_first,
+        start_blocked_second]}
     ].
 
 suite() ->
@@ -87,13 +93,15 @@ end_per_suite(Config) ->
     ok.
 
 init_per_group(two_nodes, Config) ->
-    [{nodes, [n(dev1),
-              n(dev2)]} | Config];
+    config(two_nodes, Config);
 init_per_group(three_nodes, Config) ->
-    Config1 = [{nodes, [n(dev1),
-                        n(dev2),
-                        n(dev3)]} | Config],
+    Config1 = config(three_nodes, Config),
     stop_nodes(Config1),
+    Config1;
+init_per_group(one_blocked, Config) ->
+    Config1 = config(two_nodes, Config),
+    stop_nodes(config(three_nodes, Config)),
+    preblock_second(Config1),
     Config1;
 init_per_group(_Group, Config) ->
     Config.
@@ -154,6 +162,18 @@ start_second_node(Config) ->
     ct:log("Peers on dev2: ~p", [rpc_call(N2, aec_peers, all, [])]),
     B1 = rpc_call(N1, aec_conductor, top, []),
     true = expect_block(N2, B1).
+
+start_blocked_second(Config) ->
+    N1 = node_(dev1),
+    N2 = node_(dev2),
+    T0 = os:timestamp(),
+    start_node_(dev2, Config),
+    connect(N2),
+    timer:sleep(2000),
+    ct:log("Peers on dev1: ~p", [rpc_call(N1, aec_peers, all, [])]),
+    ct:log("Unblocked peers on dev1: ~p",
+           [rpc_call(N1, aec_peers, get_random, [10])]),
+    await_sync_abort(T0, [N1, N2]).
 
 tx_first_pays_second(_Config) ->
     N1 = node_(dev1),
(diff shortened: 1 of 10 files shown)


apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -25,8 +35,8 @@
                             "description" :
                             "The peer address that the node will present itself with. If port part given, it must match 'port' user config.",
```
lucafavatella: @velzevur This "If port part given" is the portion that I was mentioning may need update (and potentially delete the port separate config at line 41 below...).

uwiger: The quickest way to fix the problem was to make the code in `aehttp_app` match specifically on the external port, but it does seem silly to make the port part mandatory, and then have an optional "port" property besides it. Better then to specify only the host part, if needed, then optionally a port separately. Especially since the host part is used also for the internal_peer_address.

lucafavatella:
BTW this PR will have to be rebased on the new PR @velzevur is putting together.

uwiger:
... or @velzevur can push modifications to this branch. The advantage would be that there are some test cases that actually rely on the functionality.

lucafavatella:
@uwiger I see Dialyzer jobs failed.
I am stopping the OS jobs as PR will need rebase/amendments anyway.

apps/aeutils/src/aeu_requests.erl
```diff
@@ -27,6 +27,19 @@ ping(Peer) ->
...
+                    aec_peers:block_peer(Uri);
+                _ -> ok
+            end,
+            aec_events:publish(chain_sync,
+                               {sync_aborted, #{uri => Uri,
+                                                reason => Reason}}),
+            {error, Reason};
         {ok, Map} ->
             lager:debug("ping response (~p): ~p", [Uri, pp(Map)]),
             aec_sync:compare_ping_objects(PingObj, Map),
```
lucafavatella: I am a bit surprised that `{error, different_genesis_blocks}` is not checked here. Maybe a comment here would clarify for reader?

----

> A peer is auto-blocked if it passes the wrong genesis block in a ping

I see this is from the point of view of the ping server. Ok.

> A pinging peer also blocks the other peer if it receives the ping response "Different genesis blocks" or "Not allowed"

I see this is from the point of view of the ping client. Ok.

----

Also it is not clear to me whether when node connects to peer, the peer establishes another connection back.

apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -9,8 +9,18 @@
...
+                "example" : "http://somehost.somedomain:123/"
+            }
+        },
+        "blocked_peers" : {
+            "description" :
+            "Pre-configured addresses of epoch nodes NOT to contact",
+            "type"  : "array",
+            "items" : {
+                "type"    : "string",
+                "pattern": "^http://[^:\\.\"!#$%^&*()',/]+(\\.[^:\\.\"!#$%^&*()',/]+)*:[0-9]+/$",
```
lucafavatella: This type seems meant to be the same as peers key above, and complex enough to warrant a type. Though this is only refactoring so not needed for this PR.

apps/aeutils/src/aeu_requests.erl
uwiger: The different genesis blocks outcome is checked on [line 33](https://github.com/aeternity/epoch/blob/822ad0f3be7709cc63f0986772d399310d69510b/apps/aeutils/src/aeu_requests.erl#L33) in aeu_requests. The return values from `process_request()` are a bit misleading.

When a node pings a peer, it's via a HTTP request, so the peer responds with a `PingObject`. Depending of the outcome of the PingObject comparisons, one node may designate itself as client and proceed to fetch blocks and headers from the other node – also via HTTP requests. Both nodes will request the mempool from the other node, and both will request missing blocks, so there will be HTTP requests in both directions.

uwiger:
rebased against master

apps/aehttp/src/aehttp_app.erl
```diff
@@ -36,7 +36,7 @@ start(_StartType, _StartArgs) ->
 
 local_peer_uri() ->
     Port = get_external_port(),
-    local_peer(Port).
+    local_peer(Port, Port).
```
velzevur: maybe a rebase artefact?

apps/aehttp/src/aehttp_app.erl
```diff
@@ -85,7 +85,7 @@ start_websocket_internal() ->
                                  [{env, [{dispatch, Dispatch}]}]),
     ok.
 
-local_peer(Port) ->
+local_peer(Port, ExtPort) ->
```
velzevur: same as above

apps/aehttp/src/aehttp_app.erl
uwiger: Yes, corrected in latest push

apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -466,8 +518,8 @@ unsubscribe(N, Event) ->
 call_proxy(N, Req) ->
     call_proxy(N, Req, ?PROXY_CALL_RETRIES, 3000).
 
-call_proxy(N, Req, Timeout) ->
-    call_proxy(N, Req, ?PROXY_CALL_RETRIES, Timeout).
+%% call_proxy(N, Req, Timeout) ->
+%%     call_proxy(N, Req, ?PROXY_CALL_RETRIES, Timeout).
```
zp-sd: Just remove?

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -188,3 +166,48 @@ add_missing_to_genesis_block(#{<<"height">> := 0} = Block) ->
...
+        false -> handle_ping_(PingObj);
+        true  ->
+            abort_sync(Src, <<"Not allowed">>)
+    end;
+handle_ping(_) ->
+    Reason = <<"Missing source attribute">>,
+    abort_sync(undefined, Reason).
+
+handle_ping_(PingObj) ->
+    LocalPingObj = aec_sync:local_ping_object(),
```
zp-sd: Just a thought: I'm not sure if some of this logic should be moved away from HTTP handler to core APIs, e.g. `aec_sync:abort(Uri, Reason)`, but (if we agree on that) this can be done later.

lucafavatella:
Created rebased branch as https://github.com/aeternity/epoch/tree/PT-153211733-remove-unwanted-peers-rebased (a4ef95ac).

lucafavatella:
@zp-sd @velzevur Could you please re-review and approve?

There was a conflict was in the websocket config in the sync test suite.
I rebased on master and force-pushed this PR in order to solve conflicts (after checking with @uwiger ).

apps/aecore/test/aecore_sync_SUITE.erl
uwiger: I tend to want to comment out some functions if I think it likely that they may be used in future test cases.

apps/aehttp/src/aehttp_dispatch_ext.erl
uwiger: I think the whole aehttp-aecore interaction around ping and sync needs to be re-structured, esp. so that aecore doesn't have any hard-coded dependencies against aehttp. Right now, it's a bit of a mess.

PR #429, opened 2017-11-29
Genesis accounts & keypair config

dincho:
[PT 153171793](https://www.pivotaltracker.com/n/projects/2124891/stories/153171793)

This PR also introduce new ["tester" environment](https://github.com/aeternity/infrastructure/pull/10) where we'll deploy the UAT genesis account keypairs.

https://github.com/aeternity/infrastructure/pull/10 needs to be merged first.

PR #435, opened 2017-11-30
Fixing local_peer internal address and ansible yaml

velzevur:
Changes:

-  Fix `aehttp_app:local_internal_http_uri/0` - it used to return an address based on the `peer_address`; now it returns the correct internal address and port

- Remove the check for  `peer_address`'s port to be the same as node's external HTTP port; this would have prevent users to choose their own ports for example while port forwarding;

- Hence the change above, the port is required part of  `peer_address`; the _default option_ for port is removed

- Fix `anisble` yaml file to include both schema and port for  `peer_address`

- Update of `epoch_config_schema.json` description

### DIFF
--- a/apps/aeutils/priv/epoch_config_schema.json
+++ b/apps/aeutils/priv/epoch_config_schema.json
@@ -23,7 +23,7 @@
                     "properties" : {
                         "peer_address" : {
                             "description" :
-                            "The peer address that the node will present itself with. If port part given, it must match 'port' user config.",
+                            "The peer address that the node will present itself with. It must include scheme, host/ip and a port followed by a trailing slash",
                             "type" : "string",
                             "pattern": "^http://[^:\\.\"!#$%^&*()',/]+(\\.[^:\\.\"!#$%^&*()',/]+)*:[0-9]+/$", 
                             "example" : "http://a.b.c:123/"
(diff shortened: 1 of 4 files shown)


deployment/ansible/templates/epoch.yaml.j2
```diff
@@ -8,7 +8,7 @@ peers:
 
 http:
     external:
-        peer_address: {{ ansible_default_ipv4.address }}
+        peer_address: http://{{ ansible_default_ipv4.address }}:3013/
```
lucafavatella: I see this PR does not include the merged genesis PR though I do not expect conflict https://github.com/aeternity/epoch/pull/429/files#diff-d7503bfba48dd964db179e7dc45da770R15

apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -23,7 +23,7 @@
                     "properties" : {
                         "peer_address" : {
                             "description" :
-                            "The peer address that the node will present itself with. If port part given, it must match 'port' user config.",
+                            "The peer address that the node will present itself with. It must include schema, host/ip and a port followed by a trailing slash",
```
uwiger: The word is actually "scheme", not "schema"

PR #407, opened 2017-11-23
[WIP] Integrate user stories testing in CI

lucafavatella:
Supersedes PR #406.

----

This is as far as I got, failing at the moment.

PR #439, opened 2017-11-30
[WIP] Integrate user stories testing in CI

lucafavatella:
Supersedes PR #407.

----

This is as far as I got, failing at the moment.


lucafavatella:
The reference to this work-in-process is in the Pivotal ticket.

PR #440, opened 2017-11-30
Use travis containers (disable sudo)

dincho:
We no longer have requirements on sudo, so switch back to (faster) containers.
[Travis reference.](https://docs.travis-ci.com/user/reference/overview/)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: required
+sudo: false
 dist: trusty
 language: erlang
 otp_release:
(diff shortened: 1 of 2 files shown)


dincho:
Damn, while the boot time decreased from 20s to 2s. The total build time increased with about a minute 😭.

See new build times: https://travis-ci.org/aeternity/epoch/builds/309962495
And compare to old: https://travis-ci.org/aeternity/epoch/builds/309603345

@happi @lucafavatella I think we should revert this?

PR #441, opened 2017-11-30
Fix websocket port config [Finishes #153297153]

lucafavatella:


lucafavatella:
I tested this locally running two nodes locally (and checking that both nodes started, and checking `$ netstat -ltna | grep '3[01][12][0-9]'` for seeing the port used).

lucafavatella:
sync test suite fixed in force-pushed commit.

lucafavatella:
All Linux CI jobs passed.

PR #448, opened 2017-12-01
Pt 153319614 make forking suite timing safe [Finishes #153319614]

uwiger:
Instead of using `timer:sleep()`, set up subscriptions and wait for `chain_sync` events to arrive from both nodes.

### DIFF
--- a/apps/aecore/test/aecore_forking_SUITE.erl
+++ b/apps/aecore/test/aecore_forking_SUITE.erl
@@ -120,9 +120,10 @@ sync_fork_in_wrong_order(Config) ->
     %% unexepctedly last block of dev1 arrives before rest of the chain
     rpc_call(N2, aec_conductor, post_block, [N1Top]), 
 
+    T0 = os:timestamp(),
     start_node_(dev1, Config),
     connect(N1),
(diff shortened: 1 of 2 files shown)


apps/aecore/test/aecore_forking_SUITE.erl
```diff
@@ -120,18 +120,50 @@ sync_fork_in_wrong_order(Config) ->
     %% unexepctedly last block of dev1 arrives before rest of the chain
     rpc_call(N2, aec_conductor, post_block, [N1Top]), 
 
+    T0 = os:timestamp(),
     start_node_(dev1, Config),
     connect(N1),
-    timer:sleep(200),
+    await_sync_complete(T0, [N1, N2]),
```
gorillainduction: There is still a timing hole here, but it is hard to plug it. The chain_sync event might come before the subscription.

I think this is good enough.

uwiger: @ThomasArts yes, but this code should be refactored into a support library, since it's basically copy-pasted from the sync_SUITE.

PR #450, opened 2017-12-04
Release testing mines 100 blocks 

velzevur:
[Pivotal task](https://www.pivotaltracker.com/story/show/153086291)
While fixing target difficulty recalculation we cut a corner with lowering the test's blocks to mine to 10 (and thus skipping target recalculation)
This PR makes expected_mine_rate configurable and sets it for release.py to mine 10 blocks a minute

### DIFF
--- a/apps/aecore/src/aec_governance.erl
+++ b/apps/aecore/src/aec_governance.erl
@@ -16,7 +16,8 @@ blocks_to_check_difficulty_count() ->
     ?BLOCKS_TO_CHECK_DIFFICULTY_COUNT.
 
 expected_block_mine_rate() ->
-    ?EXPECTED_BLOCK_MINE_RATE.
+    application:get_env(aecore, expected_mine_rate,
+                        ?EXPECTED_BLOCK_MINE_RATE).
 
 block_mine_reward() ->
     ?BLOCK_MINE_REWARD.
--- a/py/tests/release.py
+++ b/py/tests/release.py
@@ -8,6 +8,8 @@ import argparse
 import urllib3
 import shutil
 import pystache
+import logging
+from waiting import wait
 
 import swagger_client
 from swagger_client.rest import ApiException
@@ -28,7 +30,9 @@ SETUP = {
 {aecore,
  [
   {peers, ["http://localhost:9823/",
-           "http://localhost:9833/"]}
+           "http://localhost:9833/"]},
+  {aec_pow_cuckoo, {"mean16s-generic", "-t 5", 16}},
+  {expected_mine_rate, 100}
  ]},
 {aehttp,
  [
@@ -50,7 +54,9 @@ SETUP = {
             "config": '''[
 {aecore,
  [
-  {peers, []}
+  {peers, []},
+  {aec_pow_cuckoo, {"mean16s-generic", "-t 5", 16}},
+  {expected_mine_rate, 100}
  ]},
 {aehttp,
  [
@@ -72,7 +78,9 @@ SETUP = {
             "config": '''[
 {aecore,
  [
-  {peers, []}
+  {peers, []},
+  {aec_pow_cuckoo, {"mean16s-generic", "-t 5", 16}},
+  {expected_mine_rate, 100}
  ]},
 {aehttp,
  [
@@ -89,6 +97,16 @@ SETUP = {
 '''
                 }
         }
+
+def node_is_online(api):
+    try:
+        top = api.get_top()
+        return top.height > -1
+    except Exception as e:
+        return False
+
+def wait_all_nodes_are_online(apis):
+    wait(lambda: all([node_is_online(api) for api in apis]), timeout_seconds=30, sleep_seconds=0.5)
 
 def executable(temp_dir):
     return os.path.join(temp_dir, "bin", "epoch")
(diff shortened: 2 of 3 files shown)


uwiger:
It also sets the default mine rate for the dev nodes to 100, which doesn't seem necessary for `release.py`, since it loads its own config. Is this deliberate?

lucafavatella:
> It also sets the default mine rate for the dev nodes to 100, which doesn't seem necessary for release.py, since it loads its own config. Is this deliberate?

I agree with Ulf, comment welcome.

apps/aecore/src/aec_governance.erl
```diff
@@ -16,7 +16,8 @@ blocks_to_check_difficulty_count() ->
     ?BLOCKS_TO_CHECK_DIFFICULTY_COUNT.
 
 expected_block_mine_rate() ->
-    ?EXPECTED_BLOCK_MINE_RATE.
+    application:get_env(aecore, expected_mine_rate,
```
lucafavatella: I am a bit concerned in using application config for testing-motivated change. Have you considered using a fully black-box approach i.e.:
* Disable the ntpd daemon (or whatever sets continuously the OS time - if there is any such thing); and
* Before starting test, start concurrent OS process that loops full speed setting, at each loop, time to 10 seconds (or whatever needed) forward. This shall cause the timestamp in each new block to be large enough to keep difficulty low. (This of course implies reviewing how the time is retrieved for building block candidate, that shall be - I believe - OS time.)

py/tests/release.py
```diff
@@ -72,7 +78,9 @@
             "config": '''[
 {aecore,
  [
-  {peers, []}
+  {peers, []},
+  {aec_pow_cuckoo, {"lean16", "-t 5", 16}},
```
lucafavatella: These PoW-related settings will conflict with PR #454 - or PR #454 will conflict with this (whatever is ready for merge last).

apps/aecore/src/aec_governance.erl
velzevur: I found it useful to be able to configure it, esp. dev1-3 while manually testing.
As for modifying OS time - how would that play with Travis?

lucafavatella: > As for modifying OS time - how would that play with Travis?

I would expect modifying OS time meaningful only on disposable host e.g. Travis (it would cause some problems on development laptop...). It would need to be tested. I just wanted to make sure we thought of all options.

velzevur:
> It also sets the default mine rate for the dev nodes to 100, which doesn't seem necessary for release.py, since it loads its own config. Is this deliberate?
I found it nice to have not having to wait for 5 minutes for dev nodes to mine blocks. If it causes concern to both of you - I should probably revert the default behaviour.

config/dev1/sys.config
```diff
@@ -19,7 +19,8 @@
       {db_path, "."},
       {persist, false},
       {autostart, false},
-      {aec_pow_cuckoo, {"lean16", "-t 5", 16}}
+      {aec_pow_cuckoo, {"lean16", "-t 5", 16}},
+      {expected_mine_rate, 100}
```
lucafavatella: I see [default is 5 minutes per block](https://github.com/aeternity/epoch/pull/450/files#diff-312e5e881622101a1720cef694345ee9R11), this is 100 milliseconds per block. Ok.

lucafavatella:
> I found it nice to have not having to wait for 5 minutes for dev nodes to mine blocks. If it causes concern to both of you - I should probably revert the default behaviour.

Yes - I suggest not to use this testing capability unless strictly needed.

lucafavatella:
> CI failed :( https://travis-ci.org/aeternity/epoch/jobs/312428854#L1359

Best symptom of error seems to be at https://travis-ci.org/aeternity/epoch/jobs/312428854#L1542 :
```
2017-12-06 13:54:25.724 [debug] <0.2626.0>@aec_sync:do_fetch_block:315 failed to fetch block from <<"http://travis-job-3bf403d1-34a1-4c90-a6b0-858027aba44d:9833/">>; Hash = "<<112,10,16,5,...,224,192,212,123>>"; Error = {error,{parse_error,[badarg,[{jsx_decoder,incomplete,5,[{file,"/home/travis/build/aeternity/epoch/_build/default/lib/jsx/src/jsx_decoder.erl"},{line,135}]},{aeu_requests,process_http_return,1,[{file,"/home/travis/build/aeternity/epoch/_build/prod/lib/aeutils/src/aeu_requests.erl"},{line,197}]},{aeu_requests,block,2,[{file,"/home/travis/build/aeternity/epoch/_build/prod/lib/aeutils/src/aeu_requests.erl"},{line,66}]},{aec_sync,do_fetch_block,2,[{file,"/home/travis/build/aeternity/epoch/_build/prod/lib/aecore/src/aec_sync.erl"},{line,304}]},{aec_sync,fetch_next_hdr,4,[{file,"/home/travis/build/aeternity/epoch/_build/prod/lib/aecore/src/aec_sync.erl"},{line,324}]}]]}}
```

velzevur:
This is a independent problem and will be addressed in its on [PT](https://www.pivotaltracker.com/n/projects/2124891/stories/153442614)
I restarted the JOB

PR #451, opened 2017-12-04
PT-153203716 Create GET /balances external API for accounts dump

zp-sd:
@velzevur HTTP API Python test is to be added when Python tests are fixed and hooked into PR Travis jobs.


zp-sd:
Superseded by https://github.com/aeternity/epoch/pull/461.

PR #452, opened 2017-12-04
Rename aec_trees:new/0

velzevur:
[Pivotal task](https://www.pivotaltracker.com/story/show/153028708)
`aec_trees:new/0` returns a Merkle tree which is not  `#trees{}` (the main record used in aec_trees)
Renamed it to  `aec_trees:new_merkle_tree/0`

### DIFF
--- a/apps/aecore/src/aec_accounts.erl
+++ b/apps/aecore/src/aec_accounts.erl
@@ -19,7 +19,7 @@
 
 %% TODO: remove? We should not need such API (init via trees)
 empty() ->
-    {ok, _AccountsTree} = aec_trees:new().
+    {ok, _AccountsTree} = aec_trees:new_merkle_tree().
 
 new(Pubkey, Balance, Height) ->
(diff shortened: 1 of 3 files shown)


velzevur:
@ThomasArts we have 2 different types of tree structures at play here:

- aec_trees data structure - this is the #trees{} record. [This is NOT a merkle tree](https://github.com/aeternity/epoch/blob/master/apps/aecore/include/trees.hrl#L1)
- a Merkle tree - [this is the #tree{} recrord](https://github.com/aeternity/epoch/blob/master/apps/aecore/src/aec_trees.erl#L20). Currently we use it only for account's balances.

[The current aec_trees:new/0 is creating a Merkle tree and not an aec_trees structure](https://github.com/aeternity/epoch/blob/master/apps/aecore/src/aec_trees.erl#L42-L44). As you can see - this old name is really misleading. This has also resulted in a bug in the chain_state.

PR #454, opened 2017-12-04
Mean miner

lucafavatella:
Finishes [PT-153051503](https://www.pivotaltracker.com/story/show/153051503) and [PT-152915859](https://www.pivotaltracker.com/story/show/152915859).

* [x] Depends on https://github.com/tromp/cuckoo/pull/33 to be reviewed upstream first. (UPDATE: merged)
* [x] Depends on smoke test on Linux environment. (UPDATE: tested on dev1, mines stably and faster than lean.)

### DIFF
--- a/apps/aecuckoo/Makefile
+++ b/apps/aecuckoo/Makefile
@@ -1,24 +1,42 @@
 EXECUTABLES = \
-	priv/bin/mean16 \
-	priv/bin/lean16 \
-	priv/bin/verify16 \
-	priv/bin/mean28 \
-	priv/bin/lean28 \
-	priv/bin/verify28 \
-	priv/bin/mean30 \
+	priv/bin/mean30s-generic \
+	priv/bin/mean28s-generic \
+	priv/bin/mean16s-generic \
+	priv/bin/mean30s-avx2 \
+	priv/bin/mean28s-avx2 \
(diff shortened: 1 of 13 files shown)


lucafavatella:
I force-pushed considering upstream cuckoo merge and binary renames. (And I rebased on master.)

apps/aecuckoo/Makefile
```diff
@@ -1,24 +1,42 @@
...
+	priv/bin/mean30s-generic \
+	priv/bin/mean28s-generic \
+	priv/bin/mean16s-generic \
+	priv/bin/mean30s-avx2 \
+	priv/bin/mean28s-avx2 \
+	priv/bin/mean16s-avx2 \
+	priv/bin/lean30 \
 	priv/bin/lean28 \
+	priv/bin/lean16 \
+	priv/bin/verify30 \
```
hanssv: Except for the "smoke test" is there a point in including the verifiers? We don't use them anymore I realise...

lucafavatella: I think it is useful from a user or operator perspective to have a consistent set of upstream binaries so command-line checks can be performed. I can remove them if that is the agreement though...

lucafavatella:
I see two reviewers approved. I see CI passed.

PR #456, opened 2017-12-05
Remove extra newlines from lager logs

dincho:
This is to unify the log lines, otherwise it chokes the logs collector.


dincho:
Thanks

PR #457, opened 2017-12-05
Only broadcast blocks that are the new top block

gorillainduction:
Rely on pings to fill in the gaps if there are any.

Broadcasted blocks are now guaranteed to be correct with respect to
target and state hash as well.
[Finishes #153363096]


gorillainduction:
Rebased on master

PR #458, opened 2017-12-05
Move data files out of AppName/priv/

velzevur:
[Pivotal task](https://www.pivotaltracker.com/story/show/153362740)

Maybe the persistence `.chain` directory should default in the `data/`, too. Comments are welcome.


dincho:
Indeed it's better to have all mutable data under same parent dir, so I'm 👍 to move .chain and also keys, however this would mean BC break for current environment keys and .chain data.

dincho:
I'm approving this as is (from deployment point of view) there shouldn't be BC breaks.

apps/aecore/src/aec_keys.erl
```diff
@@ -401,7 +401,7 @@ check_env() ->
             undefined ->
                 application:get_env(
                   aecore, keys_dir,
-                  filename:join(code:priv_dir(aecore), "keys"));
+                  filename:join(aeu_env:application_setup_dir(aecore), "keys"));
```
dincho: I asume this is the default keys path?

dincho: I see undefined pattern match. OK.

apps/aecore/test/aecore_forking_SUITE.erl
```diff
@@ -278,7 +278,20 @@ setup_node(N, Top, Epoch, Config) ->
     delete_file(filename:join(RelD, "sys.config.orig")),
     TestD = filename:join(filename:dirname(code:which(?MODULE)), "data"),
     cp_file(filename:join(TestD, "sync_SUITE.config"),
-            filename:join(DDir , "sync_SUITE.config")).
+            filename:join(DDir , "sync_SUITE.config")),
+    cp_genesis_dir(DDir, Epoch).
+
+cp_genesis_dir(DDir, Epoch) ->
```
lucafavatella: Shall this be rather added in aec_test_utils?

deployment/ansible/deploy.yml
```diff
@@ -5,7 +5,7 @@
 
   vars:
     project_root: "{{ ansible_env.HOME }}/node"
-    genesis_accounts_path: "{{ project_root }}/lib/aecore-0.1.0/priv/.genesis/accounts.json"
+    genesis_accounts_path: "{{ project_root }}/data/aecore/.genesis/accounts.json"
```
lucafavatella: Once this is deployed in our environments, devops may delete the old file for avoiding confusion. Cc @dincho

apps/aecore/test/aecore_forking_SUITE.erl
velzevur: Yes, it would be better, I'm on it

deployment/ansible/deploy.yml
dincho: Following PT task would be nice to not loose the track

lucafavatella: @dincho https://www.pivotaltracker.com/story/show/153441285

velzevur:
Force pushed to fix a typo

apps/aeutils/src/aeu_env.erl
```diff
@@ -149,6 +150,9 @@ config_file() ->
             F
     end.
 
+application_setup_dir(AppName) when is_atom(AppName) ->
```
uwiger: Why not simply `data_dir(AppName)`?

config/dev1/sys.config
```diff
@@ -65,6 +65,7 @@
   ]},
 
   {setup, [
-           {abort_on_error, true}
+           {abort_on_error, true},
+           {data_dir, "data"}
```
uwiger: For completeness, perhaps also `{log_dir, "log"}`?

config/dev2/sys.config
```diff
@@ -61,6 +61,7 @@
   ]},
 
   {setup, [
-           {abort_on_error, true}
+           {abort_on_error, true},
+           {data_dir, "data"}
```
uwiger: For completeness, perhaps also {log_dir, "log"}?

config/dev3/sys.config
```diff
@@ -62,6 +62,7 @@
   ]},
 
   {setup, [
-           {abort_on_error, true}
+           {abort_on_error, true},
+           {data_dir, "data"}
```
uwiger: For completeness, perhaps also {log_dir, "log"}?

config/dev1/sys.config
```diff
@@ -65,6 +65,8 @@
   ]},
 
   {setup, [
-           {abort_on_error, true}
+           {abort_on_error, true},
+           {data_dir, "data"},
+           {log_dir, "log"}
```
lucafavatella: How does this hook with the code?

uwiger: @lucafavatella The `setup` API, `setup:data_dir()` and `setup:log_dir()` will read those values, relative to current working dir, then derive the abspath for each and update the application environment. Afterwards, the above mentioned functions will return fully qualified paths to the `epoch/data` and `epoch/log` directories.

Currently, we don't make use of the `log_dir` function, but it should still return the correct path to our actual log directory, if at all possible.

PR #461, opened 2017-12-06
Pt 153203821 + PT 153203716 Create debug/info/maintenance endpoints

zp-sd:
* GET /info
* GET /balances (active only when `enable_debug_endpoints=true` in `aehttp` config)


rebar.config
```diff
@@ -24,7 +24,8 @@
         {dev_mode, true},
         {include_erts, false},
 
-        {overlay, [{copy, "REVISION", "REVISION"}]},
+        {overlay, [{copy, "REVISION", "REVISION"},
+                   {copy, "VERSION" , "VERSION"}]},
```
lucafavatella: Not clear how creates VERSION in the first place?

zp-sd: VERSION is in https://github.com/aeternity/epoch/blob/master/VERSION

apps/aecore/src/aec_chain_state.erl
```diff
@@ -267,6 +268,15 @@ get_missing_block_hashes(?assert_state() = State) ->
     TopBlockHash  = get_top_block_hash(State),
     get_missing_block_hashes(TopHeaderHash, TopBlockHash, State).
 
+get_top_N_blocks_time_summary(?assert_state() = State, N) ->
```
gorillainduction: I suggest: is_integer(N), N > 0
to avoid looping infinitely below.

zp-sd: @gorillainduction Done. Please revisit.

rebar.config
lucafavatella: 😓 thanks

config/swagger.yaml
```diff
@@ -247,6 +257,43 @@ paths:
           schema:
             $ref: '#/definitions/Error'
       security: []
+
+## CAUTION: debug endpoints, implementation may be inefficient
+  /balances:
+    get:
+      tags:
+        - external
```
lucafavatella: Comment suggests internal, this point mentions external. It looks to me it shall be internal.

config/swagger.yaml
```diff
@@ -247,6 +257,43 @@ paths:
...
+      responses:
+        '200':
+          description: successful operation
+          schema:
+            $ref: '#/definitions/AccountsBalances'
+      security: []
+  /info:
+    get:
+      tags:
+        - external
```
lucafavatella: internal or external?

zp-sd: I'd say external. It might be worth dividing into `GET /version` and GET`/debug-info` one day.

config/swagger.yaml
zp-sd: The thing is that it prints the whole state tree, which may be an easy point for DDoS attack, hence the warning.

External will make debugging of UAT/testnet envs easier (you may e.g. check genesis/state trees on all UAT/testnet nodes). Internal makes it available only from your own running node (or from UAT/testnet when you log into the node).

apps/aecore/src/aec_conductor.erl
```diff
@@ -51,12 +51,16 @@
...
-        , get_miner_account_balance/0
-        , next_nonce_for_account/1
         , start_mining/0
         , stop_mining/0
         ]).
 
+%% State trees API
+-export([ get_miner_account_balance/0
+        , next_nonce_for_account/1
+        , get_all_accounts_balances/0
```
lucafavatella: A test on aec_conductor would be helpful - for the assumption that top always have state trees.

zp-sd:
@lucafavatella So we effectively want a ticket to track whether these 2 endpoints shall be internal or external? Is this what you suggested?

lucafavatella:
> @lucafavatella So we effectively want a ticket to track whether these 2 endpoints shall be internal or external? Is this what you suggested?

Yes.

lucafavatella:
> @lucafavatella So we effectively want a ticket to track whether these 2 endpoints shall be internal or external? Is this what you suggested?

With potential split you suggest. Idea being - needs to be done before mainnet.

PR #462, opened 2017-12-06
Recommend user a reasonably high number of max files

lucafavatella:
In scope of PT-153402022.

This can be (reviewed and merged) to master straight away, as these changes are applicable to tag 0.3.3 as well to any new ones.


RELEASE-NOTES.md
```diff
@@ -15,6 +15,11 @@ The instructions below also describe the simplest way to test sending tokens bet
 
 Download the [release binary](https://github.com/aeternity/epoch/releases/tag/v0.3.3-big-spenders) corresponding to your platform, e.g. `epoch-0.3.3-osx-10.12.6.tar.gz`; you would normally find the downloaded package in `~/Downloads` on macOS.
 
+The binaries are tested on the following platforms:
+* Ubuntu 16.04.3 LTS;
```
lucafavatella: Our environments.

RELEASE-NOTES.md
```diff
@@ -105,6 +110,15 @@ mkdir /tmp/node/db
...
 #### Start node
 
+It is recommended that at least 1024 files can be opened in the session where you start the node: this is needed not only for files on disk (e.g. persisted chain) but also for network connections with peers.
+You can check your max number of open files:
+* On Ubuntu, by running `ulimit -n`;
+* On macOS, by running `ulimit -n`.
+
+If your max number of open files is lower than 1024, you are recommended to raise such configuration:
+* On Ubuntu, `ulimit -n 1024` shall be sufficient;
+* On macOS, `ulimit -n 1024` shall be sufficient; if not sufficient on your OS, please report also the output of `sysctl -a | grep files`.
```
lucafavatella: This worked for me. Please test locally too though.

RELEASE-NOTES.md
```diff
@@ -105,6 +110,15 @@ mkdir /tmp/node/db
 
 #### Start node
 
+It is recommended that at least 1024 files can be opened in the session where you start the node: this is needed not only for files on disk (e.g. persisted chain) but also for network connections with peers.
```
lucafavatella: A higher number of file descriptor may be harder to set because of other configs to change.

RELEASE-NOTES.md
dincho: Example:
```
$ sysctl -a | grep files
kern.maxfiles: 49152
kern.maxfilesperproc: 24576
kern.num_files: 4884
```

RELEASE-NOTES.md
dincho: Indeed:
```
$ ulimit -n 1025
-bash: ulimit: open files: cannot modify limit: Operation not permitted
```

dincho: But this is once you change it to 1024, if you don't have limit set yet, you can set it up to kern.maxfilesperproc

dincho: The limits looks even higher on linux.
So, I think it's pretty safe to have higher than 1024 example limit if needed.

PR #464, opened 2017-12-06
Add target that developer is meant to run when pulling repo

lucafavatella:


lucafavatella:
@ThomasArts In my experience git sub-repos tend to be not developer-friendly. Though I am open to proposals! (I would have to postpone such more complex solutions though.)

velzevur:
Thank you

PR #468, opened 2017-12-06
Pt 153442614 aec peers crashes

uwiger:
Remove optimistic use of gb_trees:get from aec_peers [Finishes #153442614]
    - add_and_ping could crash on get() if uri coincided with existing alias
    - also add an 'errored' set, which is excluded from get_random()
      this means we don't try to broadcast to peers we haven't yet been
      able to ping successfully

### DIFF
--- a/apps/aecore/src/aec_peers.erl
+++ b/apps/aecore/src/aec_peers.erl
@@ -242,6 +242,7 @@ check_env() ->
     [check_env_(UKey, AKey) ||
         {UKey, AKey} <- [{<<"peers">>, peers},
                          {<<"blocked_peers">>, blocked_peers}]],
+    check_ping_interval_env(),
     ok.
 
 check_env_(UKey, AKey) ->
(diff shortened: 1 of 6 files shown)


apps/aecore/src/aec_peers.erl
```diff
@@ -836,7 +881,8 @@ maybe_ping_peer(#peer{uri = Uri} = Peer, #state{} = State) ->
     case is_local_uri(Uri, State) of
         false ->
             lager:debug("will ping peer ~p", [Uri]),
-            spawn(fun() -> ping_peer(Peer) end);
+            aec_sync:schedule_ping(Peer, fun ping_peer/1);
+            %% spawn(fun() -> ping_peer(Peer) end);
```
hanssv: This comment could be removed

uwiger: true

PR #469, opened 2017-12-06
Stop slow or suspended native miner

lucafavatella:

----

TODO:
* [x] Timeout only for mining tag;
* [x] Configurable mining attempt timeout, default 1 hour;
* [x] Update doc in aec_conductor.erl
* [x] Record for worker info kept in aec_conductor state
* [ ] Rebase on recent master

### DIFF
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -51,6 +51,7 @@
 
 %% Mining API
 -export([ get_mining_state/0
+        , get_mining_workers/0
         , start_mining/0
         , stop_mining/0
         ]).
@@ -105,6 +106,8 @@
 
 -define(SERVER, ?MODULE).
 
+-define(DEFAULT_MINING_ATTEMPT_TIMEOUT, 60 * 60 * 1000). %% milliseconds
+
 %%%===================================================================
 %%% API
 %%%===================================================================
@@ -135,6 +138,10 @@ stop_mining() ->
 -spec get_mining_state() -> mining_state().
 get_mining_state() ->
     gen_server:call(?SERVER, get_mining_state).
+
+-spec get_mining_workers() -> [pid()].
+get_mining_workers() ->
+    gen_server:call(?SERVER, get_mining_workers).
 
 %%%===================================================================
 %%% State trees API
@@ -313,6 +320,8 @@ handle_call(start_mining,_From, State) ->
     {reply, ok, State1};
 handle_call(get_mining_state,_From, State) ->
     {reply, State#state.mining_state, State};
+handle_call(get_mining_workers, _From, State) ->
+    {reply, worker_pids_by_tag(mining, State), State};
 handle_call(Request, _From, State) ->
     epoch_mining:error("Received unknown request: ~p", [Request]),
     Reply = ok,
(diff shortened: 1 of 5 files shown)


apps/aecore/src/aec_conductor.erl
```diff
@@ -116,6 +121,17 @@ start_link(Options) ->
 stop() ->
     gen_server:stop(?SERVER).
 
+%%%===================================================================
+%%% Test API
+
+-ifdef(TEST).
```
hanssv: Organisational question, except for the use of `?SERVER` this code could just as well go into the test suite? Not sure if it is better, but worth considering.

apps/aecore/src/aec_conductor.erl
```diff
@@ -409,8 +425,9 @@ dispatch_worker(Tag, Fun, State) ->
         false ->
             Wrapper = wrap_worker_fun(Fun),
             {Pid, Ref} = spawn_monitor(Wrapper),
+            {ok, TRef} = timer:exit_after(_TODO=200, Pid, shutdown),
             State1 = block_tag(Tag, State),
-            Workers = orddict:store(Pid, {Tag, Ref}, State1#state.workers),
+            Workers = orddict:store(Pid, {Tag, Ref, TRef}, State1#state.workers),
```
hanssv: I think it would make sense to introduce a record for a worker?! Three parameters in a tuple is stretching my mind :-)

hanssv: Looks good so far 👍

lucafavatella: record makes sense

apps/aecore/src/aec_conductor.erl
lucafavatella: the problem is that would move in the test suite assumption on internal state of the process - especially if records involved. We may actually expose this function also when compiled in non-test mode - let me give further thought to this.

apps/aecore/src/aec_conductor.erl
```diff
@@ -409,8 +425,9 @@ dispatch_worker(Tag, Fun, State) ->
         false ->
             Wrapper = wrap_worker_fun(Fun),
             {Pid, Ref} = spawn_monitor(Wrapper),
+            {ok, TRef} = timer:exit_after(_TODO=200, Pid, shutdown),
```
gorillainduction: I was not aware of this function. Nice!

However, I've had some problems with using timer.erl and prefer using erlang:set_timer/cancel_timer which is using the timer wheel in the vm rather than another gen_server. 

On the other hand, we are not using that many timers, so it is unlikely that we would see problems with this.

Just some ramblings. Do as you please.

apps/aecore/src/aec_conductor.erl
gorillainduction: You could do a gen_server:call from the test suite instead if you don't want to expose the api.

apps/aecore/src/aec_conductor.erl
lucafavatella: record introduced

apps/aecore/src/aec_conductor.erl
lucafavatella: I agree with you on timer functions in erlang module to be preferred over those in timer module.
Though, as you pointed out, we are note using many timers. And the convenience of timer:exit_after made me opt for this.

apps/aecore/src/aec_conductor.erl
lucafavatella: I am addressing this now.

apps/aecore/test/aec_conductor_tests.erl
```diff
@@ -99,6 +113,63 @@ test_run_miner() ->
...
 
+miner_timeout_test_() ->
+    {foreach,
+     fun() ->
+             ok = meck:new(aec_pow_cuckoo, [passthrough]),
+             ok = meck:new(application, [unstick, passthrough]),
+             ok = meck:expect(application, get_env, 3,
+                              fun
+                                  (aecore, mining_attempt_timeout, _) ->
+                                      500;
```
lucafavatella: I am briefly checking if this timeout can be reduced easily (see TODO comment below).

lucafavatella:
@hanssv @gorillainduction This is ready for your review.

apps/aecore/test/aec_conductor_tests.erl
lucafavatella: The only solution that came to me is:
* Set here a small timeout like 10;
* In the test, mock [`aec_conductor:handle_info(DOWN, ...)`](https://github.com/aeternity/epoch/blob/b6116c807ff746ffd5c1419463800e5ae071b76c/apps/aecore/src/aec_conductor.erl#L340) to send a message to the test pid.

This would show that worker pid died (because killed by timer:exit_after).

PR #473, opened 2017-12-07
Pt 152480909 decide what metrics

uwiger:
Initial metrics reporting to DataDog & log [Finishes #152480909]
    
    - reporting is done via an exometer_report instance, publishing
      through aec_events to an exometer_report_logger
    - The logger has two output plugins: one for statsd, one for lager
    - A lager log, epoch_metrics.log with custom formatting logs each
      metric
    - Currently all metrics under [ae|'_'] are reported
    - Setting the metrics_port to 0 tells the reporter not to try
      to connect to the statsd daemon


dincho:
@uwiger I don't see top block height reported, I really find this metric very important to monitor network de-syncs. Could you add it please?

PR #474, opened 2017-12-07
Add a user-config checking tool [Finishes #153312606]

uwiger:
- The tool ends up as epoch/bin/check_config
- Usage: check_config epoch.{json|yaml} [schema]
- Pretty-prints validation errors to make the more user-friendly
- Also some added support for reducing error output for invalid
  user configs

### DIFF
--- a/apps/aeutils/src/aeu_env.erl
+++ b/apps/aeutils/src/aeu_env.erl
@@ -15,8 +15,8 @@
 -export([user_map/0, user_map/1]).
 -export([user_config_or_env/4]).
 -export([read_config/0]).
--export([check_config/1]).
 -export([data_dir/1]).
+-export([check_config/1, check_config/2]).
 
 -type basic_type() :: number() | binary() | boolean().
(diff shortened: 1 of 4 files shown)


apps/aeutils/src/aeu_env.erl
```diff
@@ -135,20 +135,26 @@ lists_map_key_find(_, []) ->
     error.
 
 
-
 read_config() ->
+    read_config(report).
+
+read_config(Mode) ->
```
hanssv: What modes are there?

apps/aeutils/src/aeu_env.erl
```diff
@@ -232,11 +298,14 @@ set_env(App, K, V) ->
     error_logger:info_msg("Set config (~p): ~p = ~p~n", [App, K, V]),
     application:set_env(App, K, V).
 
-read_json(F) ->
+%% read_json(F) ->
```
hanssv: To be removed?

PR #478, opened 2017-12-08
Retry ping if aehttp is slow to start

hanssv:
If for example the node is flooded with transactions, aehttp can be slow
to start. Just retry the ping in this case.

[Finishes #153507135]

### DIFF
--- a/apps/aecore/src/aec_peers.erl
+++ b/apps/aecore/src/aec_peers.erl
@@ -66,7 +66,7 @@ add(Peer, Connect) when is_boolean(Connect) ->
     ok.
 
 %%------------------------------------------------------------------------------
-%% Register an reported source address of known peer P. Ignored if P is unknown
+%% Register a reported source address of known peer P. Ignored if P is unknown
 %%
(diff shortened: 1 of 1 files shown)


apps/aecore/src/aec_peers.erl
```diff
@@ -672,27 +672,34 @@ ping_peer(Peer) ->
...
+                {ok, Map} ->
+                    log_good_ping(Peer#peer.uri),
+                    Peers = maps:get(<<"peers">>, Map, []),
+                    add_and_ping_peers(Peers);
+                _ ->
+                    log_ping_error(Peer#peer.uri)
+            end;
+        {error, timeout} ->
+            lager:debug("timeout waiting for aehttp - no ping (~p) will retry", [Peer#peer.uri]),
+            ping_peer(Peer)
```
uwiger: no retry counter?

hanssv: Not sure what would be the right action when we reach that counter, better to keep trying?

PR #480, opened 2017-12-08
Enable user to check user config offline

lucafavatella:
TODO:
* [ ] Fix `bin/check_config epoch.yaml` returning `env: escript: No such file or directory` (`find . -name escript` outputs `./erts-9.1.4/bin/escript`)
* [x] Fix `tar: Removing leading '/' from member names` when running `tar xf` (resulting in `find . -name check_config` returning 2 results: `./bin/check_config` `./Users/luca/dev/aeternity/epoch/_build/prod/rel/epoch/bin/check_config`)

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -71,7 +71,14 @@
                    ]},
             {prod, [{relx, [{dev_mode, false},
                             {include_erts, true},
-                            {include_src, false}]}
+                            {include_src, false},
+                            {overlay, [{copy,
+                                        "_build/prod/bin/check_config",
+                                        "bin/check_config"},
+                                       {copy,
+                                        "_build/prod/lib/aeutils/priv/epoch_config_schema.json",
+                                        "data/epoch_config_schema.json"}]}
+                           ]}
                    ]}
            ]
 }.


lucafavatella:
> Fix bin/check_config epoch.yaml returning env: escript: No such file or directory (find . -name escript outputs ./erts-9.1.4/bin/escript)

Short term solution is adding in release notes `( export PATH="$(pwd)/$(ls -dr erts-* | head -n 1)/bin:$PATH" && bin/check_config epoch.yaml; )`.

A slightly more complex solution is writing shell wrapper doing the above assuming that it is called from root folder of the OTP release.

A more complex solution is putting in the escript shebang the path of the escript bin, that would rely on:
* [`erts_dir` in relx](https://github.com/erlware/relx/commit/88598b2f166849cd87ab102dcbab4497458b70cd)
* Doing some trick with `escript_shebang` in rebar3, templating using overlay var;
* Either rely on user to call escript from specific folder (in order to start shebang with `./`), or infer absolute path in some way e.g. with a shell script wrapper (and use BASENAME[0] of bash or similar).

lucafavatella:
> Short term solution is adding in release notes ( export PATH="$(pwd)/$(ls -dr erts-* | head -n 1)/bin:$PATH" && bin/check_config epoch.yaml; ).

If going with this solution, this PR is ready to be reviewed and merged.

> A slightly more complex solution is writing shell wrapper doing the above assuming that it is called from root folder of the OTP release.

This would be a bit cleaner though may not fit today.

lucafavatella:
I see two approvers, CI passed.

PR #483, opened 2017-12-08
PT-153511631 Do not include invalid transactions in a block

zp-sd:


apps/aecore/src/txs/aec_tx.erl
```diff
@@ -171,27 +161,25 @@ process_single(Tx, Trees, Height) ->
...
+            Trees0;
+        [SignedCoinbaseTx] ->
+            #coinbase_tx{account = MinerPubkey} = aec_tx_sign:data(SignedCoinbaseTx),
+            AccountsTrees0 = aec_trees:accounts(Trees0),
+
+            {ok, Account0} = aec_accounts:get(MinerPubkey, AccountsTrees0),
+            {ok, Account} = aec_accounts:earn(Account0, TotalFee, Height),
+
+            {ok, AccountsTrees} = aec_accounts:put(Account, AccountsTrees0),
+            Trees = aec_trees:set_accounts(Trees0, AccountsTrees),
```
hanssv: The assignment and next line is unnecessary, but ok for now...

zp-sd:
@ThomasArts Code should look more readable now.

Bump to `db10` is to start with a new chain in integration env. This change may not be backward compatible wrt chain structure - we may have some blocks in the current chain that contain invalid transactions. In this case validation in `aec_chain_state` will fail with `invalid_transactions_in_block`.

We need to discuss on standup what to do with changes that are not backward compatible before next deployment to testnet env.

apps/aecore/src/txs/aec_tx.erl
zp-sd: Agree, but this is just emacs reformat here.. I may revert it if you want.

hanssv: I mean the `Trees = ..., Trees` part?!

hanssv: Never change indentation of badly written code ;-)

gorillainduction:
I guess we should hold on to this until the decision is made for fork/no-fork.

The db  dir change might not belong in this PR.

zp-sd:
@gorillainduction Db dir change probably is just fine for integration env, so I'll leave it for now. But agree on holding on with merging.

apps/aecore/test/aec_mining_tests.erl
```diff
@@ -156,13 +156,15 @@ setup(PoWMod) ->
     meck:new(aeu_time, [passthrough]),
     meck:expect(aeu_time, now_in_msecs, 0, 1510253222889),
     {ok, _} = aec_tx_pool:start_link(),
+    SignedTx = #signed_tx{data = #coinbase_tx{account = <<"pubkey">>},
+                          signatures = [<<"sig1">>]},
     Trees = #trees{accounts = [#account{pubkey = <<"pubkey">>}]},
     meck:expect(aec_trees, all_trees_hash, 1, <<>>),
-    meck:expect(aec_tx, apply_signed, 3, {ok, Trees}),
+    meck:expect(aec_tx, filter_out_invalid_signatures, fun(X) -> X end),
```
zp-sd: Too bad we cannot remove these mecks here and use `aec_test_util:aec_keys_setup/0`... /cc @hanssv

PR #484, opened 2017-12-08
Don't drop the top header in get_adjustment_headers

hanssv:
Stupid refactoring mistake


hanssv:
Removed commented out code

PR #486, opened 2017-12-09
Refactoring of test SUITEs

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/153401537)
Common CT suite functions are moved to `aecore_suite_utils` and there is some small refactoring in it.

Added an initial version for integration tests for the HTTP endpoings (`aehttp_integration_SUITE`). So far there are just 2 endpoints tested: post ping and get top

Also fixed a minor bug in `aec_sync.erl` - when a ping was sent, if some crucial fields were missing, the request was crashing.


apps/aecore/test/aecore_forking_SUITE.erl
```diff
@@ -42,24 +42,27 @@ suite() ->
...
     ok = application:ensure_started(erlexec),
     DataDir = ?config(data_dir, Config),
-    TopDir = top_dir(DataDir),
-    Config1 = [{top_dir, TopDir}|Config],
-    make_shortcut(Config1),
+    TopDir = aecore_suite_utils:top_dir(DataDir),
+    Config1 = [{symlink_name, "latest.fork"},
+               {top_dir, TopDir},
+               {test_dir, "forking_SUITE"},
+               {config_name, "sync_SUITE"}] ++ Config,
```
uwiger: Why this config_name?

apps/aecore/test/aecore_forking_SUITE.erl
```diff
@@ -167,300 +169,21 @@ check_sync_event(#{sender := From, info := Info} = Msg, Nodes) ->
 
 
 mine_one_block(N) ->
-    subscribe(N, block_created),
-    rpc_call(N, application, set_env, [aecore, aec_pow_cuckoo, {"mean16s-generic", "-t 5", 16}]), 
-    StartRes = rpc_call(N, aec_conductor, start_mining, []),
+    aecore_suite_utils:subscribe(N, block_created),
+    aecore_suite_utils:rpc_call(N, application, set_env, [aecore, aec_pow_cuckoo, {"mean16s-generic", "-t 5", 16}]),
```
uwiger: Why is this not statically configured?

apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -899,137 +765,19 @@ maps_get(K, #{} = M, Def) when is_function(Def, 0) ->
...
+    aecore_suite_utils:create_config(Node, Config, EpochCfg,
+                                            [{block_peers, BlockedPeers},
+                                             {add_peers, true}
+                                            ]).
 
 config(two_nodes, Config) ->
-    [{nodes, [n(dev1),
-              n(dev2)]} | Config];
+    [{nodes, [aecore_suite_utils:node_tuple(dev1),
+              aecore_suite_utils:node_tuple(dev2)]} | Config];
```
uwiger: A list comprehension would look nice here.

apps/aecore/test/aecore_suite_utils.erl
```diff
@@ -0,0 +1,359 @@
...
+            erlang:demonitor(Ref),
+            Result
+    after Timeout ->
+            error(proxy_call_timeout)
+    end;
+call_proxy(N, _, _, _) ->
+    erlang:error({proxy_not_running, N}).
+
+
+hostname() ->
```
uwiger: Perhaps the this, and the functions below, should be moved up, so that they don't appear to be part of the proxy_process implementation?

apps/aecore/test/aecore_forking_SUITE.erl
velzevur: so we could easily change it

ThomasArts: Why does a test_dir get the name of a SUITE as well as that a config has the nam of a SUITE? Hard to grasp what's going on here.

The test_dir is a temporary directory created for this SUITE.  atom_to_list(?MODULE) would do. Or rather the name of the Suite could be provided and later used to make a temp dir?? The key is that this is unique for each module, though. This can be enforced, rather then letting it be determined in the suite.

The config name clearly indicates that somehow that part of sync_SUITE should be isolated, because they base upon the same ground.

apps/aecore/test/aecore_forking_SUITE.erl
lucafavatella: Because the PoW can be configured by sys.config but not user config (yet - see https://www.pivotaltracker.com/story/show/153317000). And the current test suite code copies the sys.config without editing it.

apps/aecore/test/aecore_forking_SUITE.erl
velzevur: The [config_name](https://github.com/aeternity/epoch/blob/master/apps/aecore/test/data/sync_SUITE.config) is the name of the config we start the virtual machine with. Currently both forking and sync suites use one file but this could possibly change.

apps/aecore/src/aec_sync.erl
```diff
@@ -91,12 +91,12 @@ compare_ping_objects(Local, Remote) ->
     lager:debug("Compare, Local: ~p; Remote: ~p", [pp(Local), pp(Remote)]),
     Src = maps:get(<<"source">>, Remote),
     Res = case {maps:get(<<"genesis_hash">>, Local),
-                maps:get(<<"genesis_hash">>, Remote)} of
+                maps:get(<<"genesis_hash">>, Remote, missing)} of
```
lucafavatella: This check shall be removed. Rather, the [`swagger.yaml` definition for `Ping`](https://github.com/aeternity/epoch/blob/bb5786751abdff67cc667092e7dd5e7ae45d6724/config/swagger.yaml#L360) shall be enriched with [`required`](https://github.com/OAI/OpenAPI-Specification/blob/1adcb8830aab48506b1ffd876330ab8c22820ad6/versions/2.0.md#schema-object).

apps/aecore/src/aec_sync.erl
```diff
@@ -91,12 +91,12 @@ compare_ping_objects(Local, Remote) ->
...
               {G, G} ->
                   lager:debug("genesis blocks match", []),
                   %% same genesis block - continue
                   case {maps:get(<<"best_hash">>, Local),
-                        maps:get(<<"best_hash">>, Remote)} of
+                        maps:get(<<"best_hash">>, Remote, missing)} of
```
lucafavatella: This check shall be removed. Rather, the [`swagger.yaml` definition for `Ping`](https://github.com/aeternity/epoch/blob/bb5786751abdff67cc667092e7dd5e7ae45d6724/config/swagger.yaml#L360) shall be enriched with [`required`](https://github.com/OAI/OpenAPI-Specification/blob/1adcb8830aab48506b1ffd876330ab8c22820ad6/versions/2.0.md#schema-object).

apps/aecore/src/aec_sync.erl
```diff
@@ -106,7 +106,7 @@ compare_ping_objects(Local, Remote) ->
                           server_get_missing_blocks(Src);
                       _ ->
                           Dl = maps:get(<<"difficulty">>, Local),
-                          Dr = maps:get(<<"difficulty">>, Remote),
+                          Dr = maps:get(<<"difficulty">>, Remote, missing),
```
lucafavatella: This check shall be removed. Rather, the [`swagger.yaml` definition for `Ping`](https://github.com/aeternity/epoch/blob/bb5786751abdff67cc667092e7dd5e7ae45d6724/config/swagger.yaml#L360) shall be enriched with [`required`](https://github.com/OAI/OpenAPI-Specification/blob/1adcb8830aab48506b1ffd876330ab8c22820ad6/versions/2.0.md#schema-object).

lucafavatella:
For the rest it looks good.

apps/aecore/src/aec_sync.erl
velzevur: Will be addressed in [Pivotal story](https://www.pivotaltracker.com/story/show/153629617)

apps/aecore/src/aec_sync.erl
velzevur: Will be addressed in [Pivotal story](https://www.pivotaltracker.com/story/show/153629617)

apps/aecore/src/aec_sync.erl
velzevur: Will be addressed in [Pivotal story](https://www.pivotaltracker.com/story/show/153629617)

lucafavatella: The code below relies on inequality `>` of number vs. atom (`missing`) that may return unexpected result so I suggest changing at least this one in this PR.

apps/aecore/test/aecore_forking_SUITE.erl
ThomasArts: I like the approach to not have to copy files. This would require a bit of thought, but it would be less error prone if we could just start the code on 3 nodes with different settings without a config file... but I fear this is hard to do.

apps/aecore/src/aec_sync.erl
uwiger: Still, a good way to address this would be:
```erlang
compare_ping_objects(#{<<"genesis_hash">> := LGen,
                       <<"best_hash">>    := LBest} = Local,
                     #{<<"genesis_hash">> := RGen,
                       <<"best_hash">>    := RBest,
                       <<"source">>       := Src,
                       <<"difficulty">>   := RDiff} = Remote) ->
 ...;
compare_ping_objects(_, _) ->
    {error, illegal_ping_object}.
```

lucafavatella: This approach would remove the issue described at https://github.com/aeternity/epoch/pull/486/files#r156638162

velzevur:
All tests for HTTP Endpoints are written. Left to cleanup the Python stuff and rebase to cleanup conflicts

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -0,0 +1,657 @@
...
+    lists:foreach(
+        fun(Height) ->
+            {ok, ExpectedBlock} = rpc(aec_conductor, get_block_by_height, [Height]),
+            {ok, H} = aec_blocks:hash_internal_representation(ExpectedBlock),
+            Hash = base64:encode(H),
+            ExpectedBlockMap = block_to_endpoint_map(ExpectedBlock), 
+            {ok, 200, BlockMap} = get_block_by_hash(Hash),
+            ct:log("ExpectedBlockMap ~p, BlockMap: ~p", [ExpectedBlockMap,
+                                                         BlockMap]),
+            % uncomment when fix aec_blocks:serialize_tx to have
```
velzevur: I will create a specific pivotal story to address this.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -0,0 +1,657 @@
...
+    true = length(ExpectedTxs) =:= length(ReturnedTxs),
+    true = lists:all(fun(Tx) -> lists:member(Tx, ExpectedTxs) end, ReturnedTxs),
+
+    {ok, 200, #{<<"balance">> := Bal0}} = get_balance(),  
+    {ok, 404, #{<<"reason">> := <<"Account not found">>}} =
+                  get_balance(base64:encode(ReceiverPubKey)),
+
+
+    % Currently aec_conductor:start fails to mine a new block;
+    % investigate this issue and uncomment the followin lines
```
velzevur: I will create a specific pivotal story to address this.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -0,0 +1,657 @@
...
+            recipient => random_hash(),
+            amount => Amount,
+            fee => Fee,
+            nonce => Nonce}),
+    {ok, SignedTx} = rpc(aec_keys, sign, [SpendTx]),
+    post_tx(SignedTx),
+    {ok, [SignedTx]} = rpc(aec_tx_pool, peek, [infinity]), % same tx 
+    ok.
+
+%% TODO: POST broken tx (currently not handled in API Endpoint)
```
velzevur: I will create a specific pivotal story to address this.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -0,0 +1,657 @@
...
+
+    true = length(Balances) =:= length(ExpectedBalances),
+    true =
+        lists:all(
+            fun(#{<<"pub_key">> := PKEncoded, <<"balance">> := Bal}) ->
+                Account = {base64:decode(PKEncoded), Bal},
+                lists:member(Account, ExpectedBalances) end,
+            Balances),
+    %% TODO: after fixing the aec_conductor mining issue, spend some txs
+    %% and check all accounts
```
velzevur: I will create a specific pivotal story to address this.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -0,0 +1,657 @@
...
+                lists:member(Account, GenesisAccounts) end,
+            Balances),
+    ok.
+
+all_accounts_balances_disabled(_Config) ->
+    rpc(application, set_env, [aehttp, enable_debug_endpoints, false]),
+    {ok, 404, _} = get_all_accounts_balances(),
+    ok.
+
+%% TODO: GetInfo: should it be disabled?
```
velzevur: I will create a specific pivotal story to address this.

apps/aecore/src/aec_blocks.erl
```diff
@@ -146,7 +146,7 @@ serialize_to_map(B = #block{}) ->
      }.
 
 serialize_tx(Tx) ->
-    #{tx => base64:encode(aec_tx_sign:serialize_to_binary(Tx))}.
+    #{<<"tx">> => base64:encode(aec_tx_sign:serialize_to_binary(Tx))}.
```
velzevur: I know addressing of problems should be done in a subsequent PRs but this one was blocking a couple of different tests

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -0,0 +1,657 @@
...
+         {<<"version">>, 42},
+         {<<"pow">>, lists:seq(1, 42)},
+         {<<"transactions">>, []}
+        ]),
+
+    lists:foreach(
+        fun({BrokenField, BlockMap}) ->
+            ct:log("Testing with a broken ~p", [BrokenField]),
+            {ok, Block} = aec_blocks:deserialize_from_map(BlockMap),
+            {ok, 200, _} = post_block(Block),
```
velzevur: The return code of a erronous `post_block` shouldn't be `200`, this will be addressed in [this PT](https://www.pivotaltracker.com/story/show/153629617) as well as other error codes.

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: As discussed with @zp-sd 
* create `/version` for returning the version, revision and genesis
* protect /info and add more data to it if needed (block difficulties maybe?)
* change 404 with {} to smth more appropriate when debug flag is disabled

velzevur:
Removed the unused .py tests
Refactored a little bit the `Makefile` so the python sources are more independent and self contained.
Renamed make target for UAT tests from `python-integration-test` (which made no sense) to `python-uats`
Some python UATs were brought up to date (they were missing the target difficulty)

apps/aecore/src/aec_blocks.erl
ThomasArts: well done, this is obvious incompatible with deserialize.

apps/aecore/src/aec_sync.erl
```diff
@@ -106,11 +106,12 @@ compare_ping_objects(Local, Remote) ->
                       _ ->
                           Dl = maps:get(<<"difficulty">>, Local),
                           Dr = maps:get(<<"difficulty">>, Remote),
-                          if Dl > Dr ->
+                          case Dl > Dr of
```
ThomasArts: This is effectively rewriting an if as a case. I don't see why this fixes a bug in the code. It creates a dependency on aec_sync and if it is just cosmetic, I rather not have it in this PR.

apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -802,7 +500,7 @@ expect_block(N, B) ->
           {?LINE, expect_block, N, B}).
 
 expect_block_(N, B) ->
-    Bn = rpc_call(N, aec_conductor, top, []),
+    Bn = aecore_suite_utils:rpc_call(N, aec_conductor, top, []),
```
ThomasArts: Why does the rpc has to go through aecore_suite_utils ??

apps/aecore/test/aecore_suite_utils.erl
```diff
@@ -0,0 +1,517 @@
...
+    connect(N, 5),
+    report_node_config(N).
+
+subscribe(N, Event) ->
+    call_proxy(N, {subscribe, Event}).
+
+unsubscribe(N, Event) ->
+    call_proxy(N, {unsubscribe, Event}).
+
+rpc_call(Node, Module, Function, Arguments) ->
```
ThomasArts: Should we not leave it to the test designer which timeout to use? I prefer not to have a util function that basically calls a well known library function

apps/aehttp/test/aehttp_integration_SUITE.erl
ThomasArts: In other tests we start conductro with automatic mining set to false.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -0,0 +1,661 @@
...
+    {ok, 404, _} = post_ping(WrongGenHashPing),
+    % node is blocked now 
+    {ok, 404, #{<<"reason">> := <<"Not allowed">>}} =
+        post_ping(maps:put(<<"source">>, Peer, PingObj)),
+    ok.
+
+get_top_empty_chain(_Config) ->
+    {ok, 200, HeaderMap} = get_top(), 
+    ct:log("~p returned header = ~p", [?NODE, HeaderMap]),
+    ExpectedH = rpc(aec_conductor, top_header, []),
```
ThomasArts: Using rpc to aecore is a bit of using the wrong abstraction level. It uses a back door to get expected result. But what are we really testing then... well that the webserver calls a function similar to aec_conductor:top_header(). We do not test that we get back the header of the genesis block.

This is fine, we only want to test the aehttp leayer. But if we want to know that it is calling the right functions, then we could just mock the interface to aecore and check that this is called.

Alternatively, we copy a bit of the test logic and actually check that we get the genesis header back.

apps/aecore/test/aecore_forking_SUITE.erl
uwiger: For debugging, I think it's actually quite helpful to have config files to inspect in separate directories for each node.

apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -76,31 +74,43 @@ suite() ->
...
 init_per_group(two_nodes, Config) ->
     config(two_nodes, Config);
 init_per_group(three_nodes, Config) ->
     Config1 = config(three_nodes, Config),
-    stop_nodes(Config1),
+    lists:foreach(
+        fun(Node) ->
+            aecore_suite_utils:stop_node(Node, Config)
+        end,
+        [dev1, dev2]),
```
lucafavatella: Why not ensuring that `dev3` node is stopped?

config/swagger.yaml
```diff
@@ -373,6 +373,11 @@ definitions:
         type: array
         items:
           type: string
+    required:
```
lucafavatella: I see this starts https://www.pivotaltracker.com/story/show/153629617 
Is this PR meant to finish it as well?

apps/aecore/test/aecore_suite_utils.erl
```diff
@@ -0,0 +1,517 @@
...
+    after 30000 ->
+            StopRes = aecore_suite_utils:rpc_call(Node, aec_conductor, stop_mining, []),
+            ct:log("aec_conductor:stop_mining() (~p) -> ~p", [Node, StopRes]),
+            ct:log("timeout waiting for block event~n"
+                   "~p", [process_info(self(), messages)]),
+            error(timeout_waiting_for_block)
+    end.
+
+mine_blocks(Node, NumBlocksToMine) ->
+    aecore_suite_utils:subscribe(Node, block_created),
```
gorillainduction: We might also want to unsubscribe?

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -0,0 +1,661 @@
...
+    PingObj = rpc(aec_sync, local_ping_object, []),
+    WrongGenHashPing = maps:merge(PingObj, #{<<"source">> => Peer,
+                                             <<"genesis_hash">> => <<"foo">>}),
+    {ok, 404, _} = post_ping(WrongGenHashPing),
+    % node is blocked now 
+    {ok, 404, #{<<"reason">> := <<"Not allowed">>}} =
+        post_ping(maps:put(<<"source">>, Peer, PingObj)),
+    ok.
+
+get_top_empty_chain(_Config) ->
```
lucafavatella: It feels renaming would help, because this test appears to assume only that the top of the chain returned by the HTTP layer is the same as the one returned by Erlang. So:
* It assumes no concurrent modification of the chain (is automining stopped? does not have peers that could push block?) that may be symptom of fragility of test;
* It does not check that chain is "empty" i.e. with only genesis (or if it does, that is not clear to me).

apps/aecore/test/aecore_sync_SUITE.erl
velzevur: This should be refactored. This stops the nodes from the previous run.
I'll fix it

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -0,0 +1,661 @@
...
+    {ok, 200, HeaderMap} = get_top(),
```
lucafavatella: It looks like this test shall:
* Get top by http API
* Get block at height 0 (hence genesis...) by http API
* Compare that the two blocks are the same

config/swagger.yaml
velzevur: No, it doesn't - the Pivotal story has yet more work to be done. Here I only fix the crucial part of the API

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -0,0 +1,661 @@
...
+    BrokenBlocks =
+        lists:map(
+            fun({Key, Value}) ->
+                BrokenBlock = maps:put(Key, Value, CorrectBlockMap),
+                {Key, BrokenBlock}
+            end,
+        [{<<"prev_hash">>, base64:encode(<<"foo">>)},
+         {<<"state_hash">>, base64:encode(<<"foo">>)},
+         {<<"txs_hash">>, base64:encode(<<"foo">>)},
+         {<<"txs_hash">>, base64:encode(<<"foo">>)},
```
lucafavatella: twice txs_hash?

lucafavatella:
Is https://github.com/aeternity/epoch/blob/28193905056cb0451492d4ddb8d14a811e256db0/py/tests/integration/data/bchain.txt still needed?

lucafavatella:
Is https://github.com/aeternity/epoch/blob/28193905056cb0451492d4ddb8d14a811e256db0/py/tests/integration/setup.yaml#L36 still needed?

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: Mining is disabled by default here, too

velzevur: Will be fixed in [Pivotal story](https://www.pivotaltracker.com/story/show/153780170)

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: [Pivotal story](https://www.pivotaltracker.com/story/show/153780309)

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: [Pivotal story](https://www.pivotaltracker.com/story/show/153780170)

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: [Pivotal story](https://www.pivotaltracker.com/story/show/153780478)

velzevur:
Addressing comments:
- [x] aec_sync unneeded changes are removed @ThomasArts 
- [x] `aecore_suite_utils:rpc_call/3` is removed, plain `rpc:call/4` is used instead @ThomasArts  and @lucafavatella 
- [x] missing call to `aecore_suite_utils:unsubscribe/2` is added @gorillainduction 
- [x] duplication ot `txs_hash` is removed @lucafavatella 
- [x] deleted the old premined chain with its corresponding key pair @lucafavatella 
- [x] revisited when aecore_sync_SUITE stops nodes @lucafavatella and @uwiger 
- [x] created missing Pivotal stories for followup fixes of the problems being found
- [x] refactored `get_top_empty_chain` test @ThomasArts and @lucafavatella
- [x] for fixing the issue above, had to [Optimise PostBlock of block with height 0](https://www.pivotaltracker.com/story/show/153180259) @lucafavatella

apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -76,37 +74,59 @@ suite() ->
...
+    ok;
+end_per_group(three_nodes, Config) ->
+    lists:foreach(
+        fun(Node) ->
+            aecore_suite_utils:stop_node(Node, Config)
+        end,
+        [dev1, dev2]),
```
lucafavatella: Third node not stopped?

velzevur:
Found a problem with my fix with the last 2 points, will fix it

velzevur:
- [x] refactored `get_top_empty_chain` test @ThomasArts and @lucafavatella
- [x] for fixing the issue above, had to [Optimise PostBlock of block with height 0](https://www.pivotaltracker.com/story/show/153180259) @lucafavatella 

I've added a [proper comment](https://github.com/velzevur/epoch/blob/04ef9110a7dfcdf573b855d8f12fa84b58984f58/apps/aehttp/src/aehttp_dispatch_ext.erl#L186) for these and [asserts for the assumptions being made](https://github.com/velzevur/epoch/blob/04ef9110a7dfcdf573b855d8f12fa84b58984f58/apps/aehttp/src/aehttp_dispatch_ext.erl#L195-L197)

apps/aecore/test/aecore_sync_SUITE.erl
velzevur: Fixed

PR #488, opened 2017-12-10
[Finishes PT-153539819] Refactor aec_tx_sign to being a real datatype

ThomasArts:
aec_tx_sign lends itself to being a pure datatype, which makes maintenance of our code base easier and helps testing by mocking only one part and no other modules.

### DIFF
--- a/apps/aecore/include/txs.hrl
+++ b/apps/aecore/include/txs.hrl
@@ -1,9 +1,3 @@
--record(signed_tx, {
-          data                       :: term(),
-          signatures = ordsets:new() :: ordsets:ordset(binary())}).
--type(signed_tx() :: #signed_tx{}).
-
-
 %% Basic transactions
 
 -record(coinbase_tx, {
--- a/apps/aecore/src/aec_tx_sign.erl
+++ b/apps/aecore/src/aec_tx_sign.erl
@@ -1,23 +1,104 @@
 -module(aec_tx_sign).
 
+%% @doc Implements a data structure for cryptographically signed transactions.
+%% This is the envelope around transactions to make them cryptographically safe. 
+%% The transactions normally also have keys of the "signers" in the transaction,
+%% which are extracted using the signers/1 function in the respective transaction
+%% handler.
+%%
+%% The purpose of this module is to provide an API for cryptograpically signed 
+%% transactions and hide all implementation details. Therefore, the record 
+%% #signed_tx{} should be kept private and considered an abstract type.
+%%
+%% A transaction can be signed by one or several signers. Each transaction can
+%% determine its own signers by the transaction callback 'signers'. Since we do not
+%% want to depend upon transaction types in this module, the user of
+%% {@module} should first obtain the signers of the transaction and then call this
+%% {@link sign/2} with these signers. There is a {@link sign/3} function that can sign
+%% with respect to a certain block height. This is handy whenever the governance
+%% variables on what crypto to use would change.
+
 %% API
--export([data/1,
+-export([sign/2,
+         sign/3, 
+         data/1,
          signatures/1,
-         verify/1]).
--export([serialize/1,
+         verify/2]).
+
+%% API that should be avoided to be used
+-export([verify/1,
+         serialize/1,
          serialize_to_binary/1,
          deserialize/1,
          deserialize_from_binary/1]).
 
 -include("common.hrl").
--include("txs.hrl").
 
+-record(signed_tx, {
+          data                       :: term(),
+          signatures = ordsets:new() :: ordsets:ordset(binary())}).
+
+-opaque signed_tx() :: #signed_tx{}.
+-export_type([signed_tx/0]).
+
+%% @doc Given a transaction Tx, a private key or list of keys, 
+%% return the cryptographically signed transaction using the default crypto
+%% parameters.
+-spec sign(term(), list(binary()) | binary()) -> signed_tx().
+sign(Tx, PrivKeys) ->
+  sign(Tx, PrivKeys, #{}).
+
+%% @doc Given a transaction Tx, a private key and a crypto map, 
+%% return the cryptographically signed transaction.
+%% @equiv sign(Tx, [PrivKey], CryptoMap).
+sign(Tx, PrivKey, CryptoMap) when is_binary(PrivKey) ->
+  sign(Tx, [PrivKey], CryptoMap);
+sign(Tx, PrivKeys, CryptoMap) when is_list(PrivKeys) ->
+    Bin = aec_tx:serialize_to_binary(Tx),
+    Algo = maps:get(algo, CryptoMap, ecdsa),
+    Digest = maps:get(digest, CryptoMap, sha256),
+    Curve = maps:get(curve, CryptoMap, secp256k1),
+    Signatures = 
+       [ crypto:sign(Algo, Digest, Bin, [PrivKey, crypto:ec_curve(Curve)]) ||
+         PrivKey <- PrivKeys ],
+    #signed_tx{data = Tx,
+               signatures = Signatures}.
+
+%% @doc Get the original transaction from a signed transaction.
+%% Note that no implicit verification is performed, it just returns the data.
+
+-spec data(signed_tx()) -> any().
+%% We have no type yest for any transaction, and coinbase_tx() | spend_tx()
+%% seems restricted as type.
 data(#signed_tx{data = Data}) ->
     Data.
 
+%% @doc Get the signatures of a signed transaction.
+-spec signatures(signed_tx()) -> list(binary()).
 signatures(#signed_tx{signatures = Sigs}) ->
     Sigs.
 
+%% @doc Verify a signed transaction by checking that the provided keys indeed all
+%% have signed this transaction.
+-spec verify(signed_tx(), list(binary())) -> ok | {error, signature_check_failed}.
+verify(#signed_tx{data = Tx, signatures = Sigs}, Signers) ->
+    %% This works even for Signers being one public key!
+    #signed_tx{signatures = NewSigs} = sign(Tx, Signers),
+    case {NewSigs -- Sigs, Sigs -- NewSigs} of
+        {[], []} ->
+          ok;
+        {DSigs1, []} ->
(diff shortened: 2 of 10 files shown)


apps/aecore/src/aec_tx_sign.erl
```diff
@@ -1,23 +1,100 @@
...
+%% determine its own signers by the transaction callback 'signers'. Since we do not
+%% want to depend upon transaction types in this module, the user of
+%% {@module} should first obtain the signers of the transaction and then call this
+%% {@link sign/2} with these signers. There is a {@link sign/3} function that can sign
+%% with respect to a certain block height. This is handy whenever the governance
+%% variables on what crypto to use would change.
+
 %% API
--export([data/1,
+-export([sign/2,
```
gorillainduction: While you're at it, can you sort the export entries, and either make them one entry or explain what the difference is.

apps/aecore/src/aec_tx_sign.erl
```diff
@@ -1,23 +1,100 @@
...
+         serialize/1,
          serialize_to_binary/1,
          deserialize/1,
          deserialize_from_binary/1]).
 
 -include("common.hrl").
--include("txs.hrl").
 
+-record(signed_tx, {
+          data                       :: term(),
```
gorillainduction: I think we can find out what this is. Probably a couple of transaction types.

apps/aecore/src/aec_tx_sign.erl
```diff
@@ -1,23 +1,100 @@
...
+%% parameters.
+sign(Tx, PrivKeys) ->
+  sign(Tx, PrivKeys, #{}).
+
+%% @doc Given a transaction Tx, a private key and a crypto map, 
+%% return the cryptographically signed transaction.
+%% @equiv sign(Tx, [PrivKey], CryptoMap).
+sign(Tx, PrivKey, CryptoMap) when is_binary(PrivKey) ->
+  sign(Tx, [PrivKey], CryptoMap);
+sign(Tx, PrivKeys, CryptoMap) when is_list(PrivKeys) ->
```
gorillainduction: Wouldn't it make sense to make sure that the CryptoMap contains the obligatory fields?

In sign/2 we could seed the argument with the default map.

apps/aecore/src/aec_tx_sign.erl
```diff
@@ -1,23 +1,100 @@
...
+%% @doc Given a transaction Tx, a private key or list of keys, 
+%% return the cryptographically signed transaction using the default crypto
+%% parameters.
+sign(Tx, PrivKeys) ->
+  sign(Tx, PrivKeys, #{}).
+
+%% @doc Given a transaction Tx, a private key and a crypto map, 
+%% return the cryptographically signed transaction.
+%% @equiv sign(Tx, [PrivKey], CryptoMap).
+sign(Tx, PrivKey, CryptoMap) when is_binary(PrivKey) ->
```
hanssv: Missing type spec?

apps/aecore/test/aec_mining_tests.erl
```diff
@@ -31,10 +31,12 @@ mine_block_test_() ->
          fun() ->
                  TopBlock = #block{height = 0,
                                    target = ?HIGHEST_TARGET_SCI},
-                 meck:expect(aec_pow, pick_nonce, 0, 9),
```
gorillainduction: I don't see how this is related to this PR.

If the behavior for this test case changed due to this PR, I think we should find the reason for this.

If this is a general test case improvement, make another PR.

apps/aecore/test/aec_blocks_tests.erl
```diff
@@ -71,23 +71,27 @@ network_serialization_test_() ->
...
+             MalformedTxs = [SignedCoinbase, aec_tx_sign:sign(#coinbase_tx{account = <<"malformed_account">>}, <<"key">>)],
              {ok, MalformedTree} = aec_txs_trees:new(MalformedTxs),
              {ok, MalformedRootHash} = aec_txs_trees:root_hash(MalformedTree),
              Block = #block{txs = CorrectTxs, txs_hash = MalformedRootHash},
              ?assertEqual({error, malformed_txs_hash}, ?TEST_MODULE:validate(Block))
      end,
      fun() ->
-             SignedCoinbase = #signed_tx{data = #coinbase_tx{}},
+             SignedCoinbase = aec_tx_sign:sign(#coinbase_tx{}, <<"key">>),
+             %% This is weird, we test that a coinbase that is signed with a key that is not the account key does validate!!
```
hanssv: Aren't we testing that nothing bad happens if the coinbase tx is bad? The tree will not contain the coinbase tx in this case I think!?

apps/aecore/test/aec_mining_tests.erl
```diff
@@ -161,8 +178,9 @@ setup(PoWMod) ->
     meck:expect(aec_tx, apply_signed, 3, {ok, Trees}),
     meck:expect(aec_keys, pubkey, 0, {ok, ?TEST_PUB}),
     meck:expect(aec_keys, sign, 1,
-                {ok, #signed_tx{data = #coinbase_tx{account = <<"pubkey">>},
-                                signatures = [<<"sig1">>]}}).
+                {ok, aec_tx_sign:sign(#coinbase_tx{account = <<"pubkey">>}, <<"sig1">>)}).
+%% #signed_tx{data = #coinbase_tx{account = <<"pubkey">>},
```
gorillainduction: Remove commented out code.

apps/aecore/test/aec_mining_tests.erl
```diff
@@ -31,10 +31,12 @@ mine_block_test_() ->
          fun() ->
                  TopBlock = #block{height = 0,
                                    target = ?HIGHEST_TARGET_SCI},
-                 meck:expect(aec_pow, pick_nonce, 0, 9),
-
-                 {ok, BlockCandidate, Nonce} = ?TEST_MODULE:create_block_candidate(TopBlock, []),
-                 {ok, Block} = ?TEST_MODULE:mine(BlockCandidate, Nonce),
+                 %% Try enough nonces to find a solution.
```
hanssv: It should be deterministic, or we don't want it as a test... We don't want to loop and try many nonces here.
We are using aec_test_utils:mock_fast_and_deterministic_cuckoo_pow() in the cuckoo case.

apps/aecore/test/aec_mining_tests.erl
```diff
@@ -79,8 +94,9 @@ difficulty_recalculation_test_() ->
                  meck:expect(aec_blocks, new, 3,
                              #block{height = 30,
                                     target = ?HIGHEST_TARGET_SCI,
-                                    txs = [#signed_tx{data = #coinbase_tx{account = <<"pubkey">>},
-                                                      signatures = [<<"sig1">>]}],
+                                    txs = [aec_tx_sign:sign(#coinbase_tx{account = <<"pubkey">>}, <<"sig1">>)],
+                                    %% [#signed_tx{data = #coinbase_tx{account = <<"pubkey">>},
```
hanssv: Commented out code..

apps/aecore/test/aec_tx_pool_tests.erl
```diff
@@ -175,24 +177,25 @@ a_signed_tx(Sender, Recipient, Nonce, Fee) ->
...
     try
         [{_, PrivKey}] = ets:lookup(?TAB, PubKey),
         Signers = aec_tx:signers(Tx),
         true = lists:member(PubKey, Signers),
-        Bin = aec_tx:serialize_to_binary(Tx),
-        Sig = crypto:sign(algo(), digest(), Bin, [PrivKey, curve()]),
-        {ok, #signed_tx{data = Tx,
-                        signatures = [Sig]}}
+        {ok, aec_tx_sign:sign(Tx, PrivKey)}
+        %% Bin = aec_tx:serialize_to_binary(Tx),
```
hanssv: Commented out code

apps/aecore/include/txs.hrl
```diff
@@ -1,8 +1,4 @@
--record(signed_tx, {
-          data                       :: term(),
-          signatures = ordsets:new() :: ordsets:ordset(binary())}).
--type(signed_tx() :: #signed_tx{}).
-
+ %% opaque type for data structure signed_tx
```
hanssv: This comment is odd, types should not be in .hrl files so moving it is right but leaving this comment behind doesn't really help anyone, or?!

apps/aecore/test/aec_tx_pool_tests.erl
gorillainduction: Remove commented out code.

apps/aecore/test/aec_tx_pool_tests.erl
```diff
@@ -175,24 +177,25 @@ a_signed_tx(Sender, Recipient, Nonce, Fee) ->
...
+        %%                 signatures = [Sig]}}
     catch
         error:Err ->
             erlang:error({Err, erlang:stacktrace()})
     end.
 
-algo()   -> ecdsa.
-digest() -> sha256.
-curve()  -> crypto:ec_curve(secp256k1).
+%% algo()   -> ecdsa.
```
gorillainduction: Remove commented out code.

apps/aecore/src/aec_tx_sign.erl
ThomasArts: I dislike the serialize and unserialize to be exported, but need to investigate who is using it.

apps/aecore/test/aec_mining_tests.erl
ThomasArts: The reason was that crypto adds salt and therefore has a non-deterministic result on computing the signed tx. I did not want to hard-code the signed_tx, but realize that I have to.

apps/aecore/src/aec_tx_sign.erl
ThomasArts: Actually, signed_tx should be a parameterized type signed_tx(tx_coinbase()) for example. But a bit too much to fix that in this PT

apps/aecore/src/aec_tx_sign.erl
ThomasArts: The crypto map should come from governance settings. Plan to do that in a separate refactoring

PR #489, opened 2017-12-12
[Fixed PT-153540305] Genesis account file incompatible with unit tests.

ThomasArts:
This patch makes it possible to run "make eunit", which had been broken by the fact that the accounts.json file could not be found in the environment the unit tests were run.

It solves it in a nice way by mocking the accounts reading in the unit tests, but take the file in the integration tests.

Enable creation of genesis block with preset accounts as argument
Update test to use genesis blocks defined in aec_test_utils
Make application test an integration test
Resolve conflict


apps/aecore/src/aec_block_genesis.erl
```diff
@@ -58,19 +62,15 @@ genesis_block() ->
...
-add_preset_accounts([], Trees) ->
-    Trees;
-add_preset_accounts([{PubKey, Amt} | T], Trees0) ->
-    Account = aec_accounts:new(PubKey, Amt, 0),
-    AccountsTrees0 = aec_trees:accounts(Trees0),
-    {ok, AccountsTrees} = aec_accounts:put(Account, AccountsTrees0),
-    Trees = aec_trees:set_accounts(Trees0, AccountsTrees),
-    add_preset_accounts(T, Trees).
+populated_trees(PresetAccounts) ->
+     {ok, Trees0} = aec_trees:all_trees_new(),    %% renaming required in aec_trees!
```
lucafavatella: Indeed - I am planning of doing this renaming (ADT for state trees) in PT-153588467.

apps/aecore/test/aec_eunit_SUITE.erl
```diff
@@ -63,3 +64,28 @@ end_per_testcase(_TC, Config) ->
...
+
+    {ok, Deps} = application:get_key(App, applications), 
+    AlreadyRunning = [ Name || {Name, _,_} <- proplists:get_value(running_apps, Config) ],
+    [ ok = application:ensure_started(Dep) || Dep <- Deps ],
+    
+    meck:new(aec_genesis_block_settings, []),
+    meck:expect(aec_genesis_block_settings, preset_accounts, 0, []),
+    ok = application:start(App),
+    timer:sleep(100),
+    ok = application:stop(App),
```
lucafavatella: application:stop leaves application loaded. I am considering whether application:unload(App) would fit here.

ThomasArts: No, not really, since we have tested that we can load it and we don't have to test that Erlang works. Unloading would just cause the next test using it to load it again and that's guaranteed to work.

Only small extra is that we could have saved application environment data that we don't pass on to next test. But I am not sure it survives stopped apps.

apps/aecore/src/aec_block_genesis.erl
ThomasArts: Great. If you're on it, also make aec_accounts and aec_trees less like spaghetti, they call each other recursively.

PR #490, opened 2017-12-12
Add CI job running eunit target

lucafavatella:


uwiger:
OTOH, we have been running the EUnit tests, at least in aecore, using a CT wrapper ...

lucafavatella:
> OTOH, we have been running the EUnit tests, at least in aecore, using a CT wrapper ...

@uwiger Correct, though at least at one point `make eunit` became broken (see PR #489).

I plan to check this works with master, then merge (two reviewers approved).

PR #494, opened 2017-12-13
[WIP] State separated from block

lucafavatella:
For your very early review as this work may conflict with yours.
This splits Merkle tree ADT in separate module - please refer to commit message for details.
The actual split of state from block is not done yet.

Dialyzer fails at the moment because blocks.hrl initialized `#block.trees` as `#trees{}` that has an invalid accounts tree.

### DIFF
--- a/apps/aeutils/src/aeu_mtrees.erl
+++ b/apps/aeutils/src/aeu_mtrees.erl
@@ -1,0 +1,93 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2017, Aeternity Anstalt
+%%% @doc In-memory Merkle trees.
+%%%
+%%% The root hash depends on the order of the operations performed on
+%%% the tree.
+%%%
+%%% This implementation is a wrapper for `gb_merkle_trees` with
+%%% stricter checks on arguments, less ambiguous return values and
+%%% enabling better type specifications.
+%%%
+%%% @see gb_merkle_trees
+%%% @end
+%%%-------------------------------------------------------------------
+-module(aeu_mtrees).
+
+%% API
+-export([new/0,
+         get/2,
+         get_with_proof/2,
+         put/3,
+         root_hash/1,
+         verify_proof/4,
+         to_orddict/1]).
+
+-export_type([mtree/0,
+              mtree/2,
+              root_hash/0]).
+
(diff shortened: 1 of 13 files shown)


apps/aeutils/src/aeu_mtrees.erl
```diff
@@ -0,0 +1,93 @@
...
+
+-type root_hash() :: <<_:(?HASH_BYTES*8)>>.
+
+%%%===================================================================
+%%% API
+%%%===================================================================
+
+-spec new() -> mtree().
+new() ->
+    {ok, gb_merkle_trees:empty()}.
```
gorillainduction: Why the {ok, _} wrapper?

Why don't you clean these out while you are doing such extensive refactoring anyway.

apps/aeutils/src/aeu_mtrees.erl
```diff
@@ -0,0 +1,93 @@
...
+        {ok, Value} when ?IS_VALUE(Value) ->
+            Proof = gb_merkle_trees:merkle_proof(Key, Tree),
+            {ok, {Value, Proof}};
+        {error, notfound} = E ->
+            E
+    end.
+
+put(Key, Value, Tree) when ?IS_KEY(Key), ?IS_VALUE(Value) ->
+    NewTree = gb_merkle_trees:enter(Key, Value, Tree),
+    {ok, NewTree}.
```
gorillainduction: Again the wrapper is unnecessary.

apps/aeutils/src/aeu_mtrees.erl
lucafavatella: Ok.
The only function that performs meaningful return value transformation is `root_hash`. For the other functions, the only value is stricter validation of args.

lucafavatella: ... and I will follow more closely dict API.

lucafavatella: Actually I realized following gb_trees OTP module (as opposed to dict) makes more sense (this is what the underlying gb_merkle_trees module does), will do that. gb_trees too does not have unnecessary wrappers.
cc @gorillainduction

lucafavatella:
Will be in new PR.

PR #495, opened 2017-12-13
PT-153629569 Move aec_tx behaviour to aetx application

zp-sd:
Depends on https://github.com/aeternity/tx_xform/pull/4.


rebar.config
```diff
@@ -38,7 +38,7 @@
   [{tx_xform, {git, "https://github.com/aeternity/tx_xform.git", {ref, "3e74c07"}}}]}.
```
lucafavatella: This will have to change (it seems it does not affect rebar.lock though).

zp-sd: This will change when aeternity/tx_xform#4 is merged.

gorillainduction:
You might want to add the file headers while you are at it.

PR #496, opened 2017-12-13
Swagger HTTP endpoint documentation

velzevur:
With a small change in the `index.html` this commit is the [default swagger ui](https://github.com/swagger-api/swagger-ui/tree/master/dist)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,8 +154,12 @@ swagger: config/swagger.yaml
 	@swagger-codegen generate -i $< -l erlang-server -o $(SWTEMP)
 	@echo "Swagger tempdir: $(SWTEMP)"
 	@cp $(SWTEMP)/priv/swagger.json $(HTTP_APP)/priv/
+	( cd $(HTTP_APP) && $(MAKE) updateswagger; )
 	@cp $(SWTEMP)/src/*.erl $(HTTP_APP)/src/swagger
 	@rm -fr $(SWTEMP)
+
+swagger-docs:
+	(cd ./apps/aehttp && $(MAKE) swagger-docs);
 
 swagger-python: config/swagger.yaml
 	@swagger-codegen generate -i $< -l python -o $(SWTEMP)
@@ -176,6 +180,7 @@ clean:
 
 distclean: clean
 	( cd apps/aecuckoo && $(MAKE) distclean; )
+	( cd $(HTTP_APP) && $(MAKE) distclean; )
 
 multi-build: dev1-build
 	@rm -rf _build/dev2 _build/dev3
--- a/apps/aehttp/Makefile
+++ b/apps/aehttp/Makefile
@@ -1,0 +1,27 @@
+SWAGGER_UI_GIT =https://github.com/swagger-api/swagger-ui.git
+SWAGGER_UI_GIT_DIR =./priv/swagger-ui
+SWAGGER_DOCS_DIR =./priv/swagger-docs
+
+.PHONY: swagger-docs
+swagger-docs: |$(SWAGGER_UI_GIT_DIR)/.git
+	@mkdir -p $(SWAGGER_DOCS_DIR)
+	@cp -p $(SWAGGER_UI_GIT_DIR)/dist/* $(SWAGGER_DOCS_DIR)
+	@cp -p ./priv/swagger.json $(SWAGGER_DOCS_DIR)
+	@sed -ibkp 's/http:\/\/petstore.swagger.io\/v2\/swagger.json/swagger.json/g' $(SWAGGER_DOCS_DIR)/index.html
+
+$(SWAGGER_UI_GIT_DIR)/.git:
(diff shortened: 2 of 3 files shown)


apps/aehttp/priv/swagger-docs/index.html
```diff
@@ -0,0 +1,95 @@
...
+<div id="swagger-ui"></div>
+
+<script src="./swagger-ui-bundle.js"> </script>
+<script src="./swagger-ui-standalone-preset.js"> </script>
+<script>
+window.onload = function() {
+  
+  // Build a system
+  const ui = SwaggerUIBundle({
+    url: "swagger.json",
```
velzevur: This is the change that specifies our `swagger.json` file

lucafavatella: In order to ease future updates, may I suggest not to include the swagger-ui files in this repo but rather fetch them with a rebar compile pre-hook and script the editing of this file (`sed ...`)? Anyway if it easier to just include the files, this is fine.

(The `aecuckoo` application uses this approach in order not to copy upstream sources - see `apps/aecuckoo/rebar.config`, `apps/aecuckoo/Makefile` and `distclean` target in top-level Makefile.)

Makefile
```diff
@@ -153,7 +153,7 @@ python-download-chain:
 swagger: config/swagger.yaml
 	@swagger-codegen generate -i $< -l erlang-server -o $(SWTEMP)
 	@echo "Swagger tempdir: $(SWTEMP)"
-	@cp $(SWTEMP)/priv/swagger.json $(HTTP_APP)/priv/
+	@cp $(SWTEMP)/priv/swagger.json $(HTTP_APP)/priv/swagger-docs
```
lucafavatella: swagger_router.erl appears to assume this location, so changing it may make starting node not possible (testing locally now).

lucafavatella: Confirmed - `make prod-build` succeeds then `make prod-start` fails.

lucafavatella:
@velzevur Is the Swagger UI meant to be accessed via TCP exposed by running node or via filesystem directly?
I see I can access Swagger UI by `open ./_build/prod/rel/epoch/lib/aehttp-0.1.0/priv/swagger-docs/index.html` i.e. filesystem.

velzevur:
Since this is documentation only - I don't think we need to serve the UI files and filesystem is good enough.

Makefile
lucafavatella: Copying swagger.json in both priv/ and priv/swagger-docs shall make the node working again (at the cost of some redundancy).

velzevur:
After discussing this with @lucafavatella - there had been a major refactoring of the PR. The docs are to be deployed on a dedicated server.
To build the docs locally:
```
make swagger-docs
```

apps/aehttp/Makefile
```diff
@@ -0,0 +1,14 @@
+SWAGGER_UI_GIT =https://github.com/swagger-api/swagger-ui.git
+SWAGGER_UI_GIT_DIR =./priv/swagger-ui
+SWAGGER_DOCS_DIR =./priv/swagger-docs
+
+.PHONY: swagger-docs
+swagger-docs: |$(SWAGGER_UI_GIT_DIR)/.git
+	@mkdir -p $(SWAGGER_DOCS_DIR)
+	@cp -p $(SWAGGER_UI_GIT_DIR)/dist/* $(SWAGGER_DOCS_DIR)
+	@cp -p ./priv/swagger.json $(SWAGGER_DOCS_DIR)
+	@sed -i 's/http:\/\/petstore.swagger.io\/v2\/swagger.json/swagger.json/g' $(SWAGGER_DOCS_DIR)/index.html
```
lucafavatella: On OSX I had to change `-i` to `-ibkp`.

lucafavatella:
I build prod-package from clean repo and I saw that these files do not get included.
I ran `make swagger-docs` then `open apps/aehttp/priv/swagger-docs/index.html` using Firefox and I see the Swagger UI with the actual content of the schema.

apps/aehttp/Makefile
```diff
@@ -0,0 +1,14 @@
+SWAGGER_UI_GIT =https://github.com/swagger-api/swagger-ui.git
+SWAGGER_UI_GIT_DIR =./priv/swagger-ui
+SWAGGER_DOCS_DIR =./priv/swagger-docs
+
+.PHONY: swagger-docs
+swagger-docs: |$(SWAGGER_UI_GIT_DIR)/.git
```
lucafavatella: * Target `clean` nice to have.
* Targets `distclean` nice to have.
* Integration of target `distclean` in main Makefile nice to have.

velzevur:
Added _nice to have_-s make targets and an optional copying of the swagger.json to the docs's dir (if one is present). This way running a top level 
```make swagger```
would update the local copy in the swagger-docs's dir

PR #502, opened 2017-12-14
Initial refactoring on block state

lucafavatella:
Preparatory PR for PT-153588467 - in order to avoid huge PR.

Mostly extracted from WIP PR #494.

### DIFF
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -57,8 +57,7 @@
         ]).
 
 %% State trees API
--export([ get_miner_account_balance/0
-        , next_nonce_for_account/1
+-export([ get_account/1
         , get_all_accounts_balances/0
         ]).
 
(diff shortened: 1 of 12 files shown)


lucafavatella:
After this PR is reviewed and merged, I plan to open second refactoring PR with [this content](https://github.com/aeternity/epoch/compare/cec50bf5...lucafavatella:PT-153588467-block-state-refactoring-2) that [passed Travis CI](https://travis-ci.org/lucafavatella/epoch/builds/316595882) on my fork.
Then the third - and last - PR shall be the one actually splitting the state trees from the block, that I need to code yet.

apps/aecore/src/aec_conductor.erl
```diff
@@ -146,26 +145,12 @@ get_mining_workers() ->
 %%%===================================================================
 %%% State trees API
 
-%% TODO: Added for backwards compability.
--spec get_miner_account_balance() ->   {'ok', integer()}
-                                     | {'error', 'account_not_found'}.
-get_miner_account_balance() ->
-    {ok, Pubkey} = aec_keys:pubkey(),
+-spec get_account(pubkey()) -> {ok, account()} | {error, notfound}.
+get_account(Pubkey) ->
```
gorillainduction: This copies the full state to the calling process. Please make it a specific call instead.

I'm allowing this in this PR, though. You can fix that in the next stage if you please.

lucafavatella: Yes, coding right now, I will include in next stage.

lucafavatella:
> So, why did you remove a test case in aec_sha256_tests? Did it not test the right thing, is it superfluous, did it not pass any more?

@ThomasArts It was testing the feature of hashing an arbitrary Erlang term. Such feature is removed so I removed the associated test.

Alternative would be turning it in a negative test i.e. if attempting to hash a non-binary, function fails.

lucafavatella:
> the feature of hashing an arbitrary Erlang term

@ThomasArts And I removed this feature as it could have allowed silent hashing of `{ok, Header}` return values or so - that could have creeped in in the further state trees-related refactoring that will follow.

PR #503, opened 2017-12-14
Add system metrics [Finishes #153631906]

uwiger:
- Add an exometer probe that subscribes to eper metrics
- Changed the way aecore probes are initialized (start order issues)
- Fixed some timing-related issues in test suites
- Upgraded to a later exometer_core commit (fixing a small bug)

### DIFF
--- a/apps/aecore/priv/exometer_predefined.script
+++ b/apps/aecore/priv/exometer_predefined.script
@@ -10,7 +10,6 @@
  {[ae,epoch,aecore,mining,interval]          , histogram   , []},
  {[ae,epoch,aecore,chain,height]             , gauge       , []},
  {[ae,epoch,aecore,chain,top_change,interval], histogram   , []},
- {[ae,epoch,aecore,chain]                    , chain_probe , []},
  {[ae,epoch,aecore,peers,count]              , gauge       , []},
  {[ae,epoch,aecore,peers,blocked]            , gauge       , []},
  {[ae,epoch,aecore,peers,errored]            , gauge       , []},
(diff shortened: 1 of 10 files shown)


apps/aecore/src/aec_keys.erl
```diff
@@ -404,7 +404,7 @@ check_env() ->
     Pwd =
         case aeu_env:user_config([<<"keys">>, <<"password">>]) of
             undefined ->
-                {ok, P} = application:get_env(aecore, password),
+                P = application:get_env(aecore, password, <<"default_pwd">>),
```
sennui: how do we deal with lack of password? Quick read shows that it will fail in hashining (for decription) call. Maybe we should shut down the node here and require password?

uwiger:
(I had accidentally committed a change to `aec_keys`. It's now been reverted.)

uwiger:
An example of metrics log output:
```
16:19:08.270 ae.epoch.system.io.in.value:55.990|g
16:19:08.270 ae.epoch.system.io.out.value:25.995|g
16:19:08.270 ae.epoch.system.memory.atom.value:842689|g
16:19:08.270 ae.epoch.system.memory.atom.used.value:828654|g
16:19:08.270 ae.epoch.system.memory.binary.value:137152|g
16:19:08.270 ae.epoch.system.memory.ets.value:3048144|g
16:19:08.270 ae.epoch.system.memory.processes.value:32277936|g
16:19:08.270 ae.epoch.system.memory.processes.used.value:32276648|g
16:19:08.270 ae.epoch.system.memory.system.value:38433264|g
16:19:08.270 ae.epoch.system.memory.total.value:70711200|g
16:19:08.270 ae.epoch.system.procs.value:407|g
```

PR #505, opened 2017-12-15
Refactor state trees ADTs

lucafavatella:
This is the [second](https://github.com/aeternity/epoch/pull/502#issuecomment-351860054) refactoring PR re state trees. A distinct PR (on which I am working now) shall finish PT-153588467 (and address https://github.com/aeternity/epoch/pull/502#discussion_r157156760).

This PR also addresses comments https://github.com/aeternity/epoch/pull/489#discussion_r156347276 and https://github.com/aeternity/epoch/pull/494#discussion_r156701434 from previous PRs.


lucafavatella:
@ThomasArts @gorillainduction I would prefer a bird's-eye view review from at least one of you two before requesting other reviewers.

apps/aeutils/src/aeu_mtrees.erl
```diff
@@ -0,0 +1,109 @@
...
+-type mtree(_K, _V) :: gb_merkle_trees:tree().
+
+-type root_hash() :: <<_:(?HASH_BYTES*8)>>.
+
+-type proof() :: gb_merkle_trees:merkle_proof().
+
+%%%===================================================================
+%%% API - subset of OTP `gb_trees` module
+%%%===================================================================
+
```
gorillainduction: I would like the get/2 as well. That crashes if the key is not present.

lucafavatella: Ok, I plan to update this PR with this function.

lucafavatella: Addressed in commit 6ccbeaeb added (no force-push, just added). Hence I would need your re-review @gorillainduction please.

lucafavatella:
> It's a kind of pity that type of the accounts in a tree is extended with undefined. Would have been much nicer if one was unable to create a tree at all without at least have an empty account tree in it. I see no access function that can create a tree with undefined account, thus it should not be needed. Probably some other types need changed, but if this is an ADT, then we can control generating trees that are well behaved.

@ThomasArts You are completely right. I have this `undefined` in the type removed in my local code (still work-in-progress) that shall finish PT-153588467. I should have added comment for clarity.

lucafavatella:
@ThomasArts 

> We dislike the long type type names. Types are not documentation.

I welcome suggestions on this.
While this actual change is merged, changing name if we feel need shall be almost a `sed` command so I would prefer deferring this to future enhancements if possible.

> If you consider aec_txs_trees as an ADT, then the name new is more logical than the name from_txs.

The txs tree has a very narrow use in the code, so I preferred exposing minimal API. In particular, the presence of at least one txs in the only constructor guarantees possibility to compute root hash.

I suggest we expand API in future iteration if needed e.g. by tests.

For name `from_txs`, I loosely followed http://erlang.org/doc/man/gb_trees.html#from_orddict-1 though I can change back if there is consensus.

> If the list of transactions passed in is [] then we could even return the empty tree.

At least one tx enforced - please see comment above.

apps/aeutils/src/aeu_mtrees.erl
gorillainduction: LGTM!

apps/aecore/src/aec_tx_sign.erl
```diff
@@ -124,7 +127,7 @@ deserialize([?SIG_TX_TYPE, ?SIG_TX_VSN, TxSer, Sigs]) ->
     #signed_tx{data = Tx, signatures = Sigs}.
 
 %% deterministic canonical serialization.
--spec serialize_to_binary(signed_tx()) -> binary().
+-spec serialize_to_binary(signed_tx()) -> deterministic_signed_tx_binary().
```
ThomasArts: I propose binary_signed_tx. The type need not include the information that the result of the function is deterministic.

apps/aecore/include/trees.hrl
```diff
@@ -1,5 +1,5 @@
 -record(trees, {
-          accounts :: aec_trees:tree()}).
+          accounts :: undefined | aec_accounts:tree()}).
```
ThomasArts: Do we really need undefined. What happens if we just leave it out? We cannot create undefined accounts!

lucafavatella: Without undefined, dialyzer spots that the block is created in some points without accounts (because of [default value in record](https://github.com/lucafavatella/epoch/blob/6ccbeaeb42d0fe39cf6923182da447d08653da61/apps/aecore/include/blocks.hrl#L27)). (Somehow these changes enabled dialyzer to spot this so I had to add the undefined here.)

lucafavatella: Will remove in following PR, that will remove trees from block.

apps/aecore/src/aec_tx_sign.erl
lucafavatella: ok, adding change now.

lucafavatella: @ThomasArts Addressed in new commit.

PR #506, opened 2017-12-15
Pt 153696599 unwanted dependency

ThomasArts:
aehttp and aec_peers are interdependent. This is a first patch to make them less so.

The API function from aec_peers that is removed, was not used anywhere else than in aehttp.

### DIFF
--- a/apps/aeoracle/src/aeo_state_tree.erl
+++ b/apps/aeoracle/src/aeo_state_tree.erl
@@ -1,0 +1,83 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2017, Aeternity Anstalt
+%%% @doc
+%%% ADT for keeping the state of oracles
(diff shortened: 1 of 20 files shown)


apps/aeoracle/src/aeo_state_tree.erl
```diff
@@ -0,0 +1,83 @@
...
+-type interaction() :: aeo_interaction:interaction().
+-type oracle() :: aeo_oracles:oracle().
+
+%%%===================================================================
+%%% API
+%%%===================================================================
+
+-spec new() -> tree().
+new() ->
+    {ok, Tree} = aec_trees:new_merkle_tree(),
```
lucafavatella: I see this PR contains portions of oracles. Are these intended to be included in the same PR?
I see the oracles tree conflicts with PR #505.

PR #507, opened 2017-12-15
Remove unwanted dependencies

ThomasArts:
The module aec_peers calls aehttp, which calls back to aec_peers again.
This is a first small commit to unraffle the dependencies between these modules aec_sync, aec_peers and aehttp application.

[finishes PT-153696599]

### DIFF
--- a/apps/aecore/test/aec_peers_tests.erl
+++ b/apps/aecore/test/aec_peers_tests.erl
@@ -44,10 +44,6 @@ all_test_() ->
                ?assertEqual(ok, aec_peers:remove("http://someone.somewhere:1337/v1")),
                ?assertEqual(1, length(aec_peers:all()))
        end},
-      {"Uri_from_ip_port",
-       fun() ->
-               ?assertEqual("http://123.123.123.123:1337/", aec_peers:uri_from_ip_port("123.123.123.123", 1337))
-       end},
       {"Remove all",
        fun do_remove_all/0},
       {"Add peer",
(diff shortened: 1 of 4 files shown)


apps/aehttp/src/aehttp_app.erl
```diff
@@ -96,10 +96,13 @@ local_peer(Port) ->
...
                                                   Addr}]});
                 nomatch ->
-                    aec_peers:uri_from_ip_port(Addr, Port)
+                    "http://" ++ Addr ++ ":" ++ integer_to_list(Port) ++ "/"
             end
     end.
 
+%%------------------------------------------------------------------------------
+%% Get url from IP and port. IP format: xxx.xxx.xxx.xxx
+%%------------------------------------------------------------------------------
```
UlfNorell: The comment suggests that you intended to define the `uri_from_ip_port` function locally and use it in the cases above. That would be nicer than inlining it.

apps/aecore/test/aec_peers_tests.erl
```diff
@@ -44,10 +44,6 @@ all_test_() ->
                ?assertEqual(ok, aec_peers:remove("http://someone.somewhere:1337/v1")),
                ?assertEqual(1, length(aec_peers:all()))
        end},
-      {"Uri_from_ip_port",
-       fun() ->
-               ?assertEqual("http://123.123.123.123:1337/", aec_peers:uri_from_ip_port("123.123.123.123", 1337))
```
lucafavatella: I am trying to understand whether there is now an equivalent test for this piece of functionality that now moved to other app. (It appears to be called in sync suite maybe.)

ThomasArts: It was not a test other than checking that ++ works
I removed the function completely and hence no test that ++ works any more

UlfNorell:
I still would have preferred a local function instead of inlining the ugly `++` code, but whatever.

PR #509, opened 2017-12-16
State trees separated from block

lucafavatella:
Finishes PT-153588467.

### DIFF
--- a/apps/aecore/src/aec_blocks.erl
+++ b/apps/aecore/src/aec_blocks.erl
@@ -7,13 +7,11 @@
 %% API
 -export([prev_hash/1,
          height/1,
-         trees/1,
          target/1,
          txs/1,
          difficulty/1,
          time_in_msecs/1,
          set_pow/3,
-         set_trees/2,
          set_target/2,
          new/3,
          to_header/1,
@@ -32,8 +30,7 @@
 -compile([export_all, nowarn_export_all]).
 -endif.
 
--export_type([block_serialized_for_network/0,
-              block_deserialized_from_network/0]).
+-export_type([block_serialized_for_network/0]).
 
 -include("common.hrl").
 -include("blocks.hrl").
@@ -43,7 +40,6 @@
 -define(CURRENT_BLOCK_VERSION, ?GENESIS_VERSION).
 
 -type block_serialized_for_network() :: binary().
--type block_deserialized_from_network() :: #block{trees :: DummyTrees::trees()}.
 
 -spec prev_hash(block()) -> block_header_hash().
 prev_hash(Block) ->
@@ -52,10 +48,6 @@ prev_hash(Block) ->
 -spec height(block()) -> height().
 height(Block) ->
     Block#block.height.
-
--spec trees(block()) -> trees().
-trees(Block) ->
-    Block#block.trees.
 
 -spec target(block()) -> integer().
 target(Block) ->
@@ -78,10 +70,6 @@ set_pow(Block, Nonce, Evd) ->
     Block#block{nonce = Nonce,
                 pow_evidence = Evd}.
 
--spec set_trees(block(), aec_trees:trees()) -> block().
-set_trees(Block, Trees) ->
-    Block#block{trees = Trees}.
-
 -spec set_target(block(), non_neg_integer()) -> block().
 set_target(Block, Target) ->
     Block#block{target = Target}.
@@ -91,7 +79,7 @@ set_target(Block, Target) ->
 txs(Block) ->
     Block#block.txs.
 
--spec new(block(), list(aec_tx_sign:signed_tx()), trees()) -> block().
+-spec new(block(), list(aec_tx_sign:signed_tx()), trees()) -> {ok, block(), trees()}.
 new(LastBlock, Txs, Trees0) ->
     LastBlockHeight = height(LastBlock),
     {ok, LastBlockHeaderHash} = hash_internal_representation(LastBlock),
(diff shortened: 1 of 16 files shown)


apps/aecore/src/aec_chain_state.erl
```diff
@@ -112,7 +115,23 @@ top_block_hash(?match_state(top_block_hash := X)) ->
...
+
+-spec top_block_with_state(state()) -> 'undefined' | {'ok', #block{}, {'state', trees()} | 'no_state_trees'}.
+top_block_with_state(?match_state(top_block_hash := undefined)) -> undefined;
+top_block_with_state(?match_state(top_block_hash := X) = State) ->
+    {ok,
+     export_block(blocks_db_get(X, State)),
+     case state_db_find(X, State) of {ok, Trees} -> {state, Trees}; error -> no_state_trees end}.
+
+-spec account(pubkey(), state()) -> 'no_state_trees' | 'none' | {value, account()}.
+account(_, ?match_state(top_block_hash := undefined)) -> undefined;
```
hanssv: 'undefined' is not part of the return type. Should be 'none'?

apps/aecore/src/aec_chain_state.erl
```diff
@@ -142,7 +161,21 @@ get_block(Hash, ?assert_state() = State) ->
...
+
+-spec get_block_with_state(binary(), state()) -> {'ok', #block{}, {'state', trees()} | 'no_state_trees'} | 'error'.
+get_block_with_state(Hash, ?assert_state() = State) ->
+    case blocks_db_find(Hash, State) of
+        {ok, Node} ->
+            case node_is_block(Node) of
+                true  ->
+                    {ok,
+                     export_block(Node),
+                     case state_db_find(Hash, State) of {ok, Trees} -> {state, Trees}; error -> no_state_trees end};
```
hanssv: Might be nicer to make a helper function for getting the state tree?

apps/aecore/src/aec_conductor.erl
```diff
@@ -145,17 +147,13 @@ get_mining_workers() ->
...
-    Trees = aec_blocks:trees(LastBlock),
-    AccountsTree = aec_trees:accounts(Trees),
-    aec_accounts_trees:lookup(Pubkey, AccountsTree).
+    gen_server:call(?SERVER, {get_account, Pubkey}).
 
 -spec get_all_accounts_balances() -> list({pubkey(), non_neg_integer()}).
 get_all_accounts_balances() ->
-    Top = top(),
-    Trees = aec_blocks:trees(Top),
+    {ok, _Block, {state, Trees}} = top_with_state(),
```
gorillainduction: Make a this a separate call to avoid copying the state.

apps/aecore/src/aec_conductor.erl
```diff
@@ -711,12 +723,14 @@ create_block_candidate(#state{keys_ready = false} = State) ->
     %% Keys are needed for creating a candidate
     wait_for_keys(State);
 create_block_candidate(State) ->
```
gorillainduction: I would very much prefer to solve this in a way that we didn't need to copy the full state to the worker process, but this is out of scope for this ticket.

apps/aecore/src/aec_conductor.erl
```diff
@@ -726,13 +740,13 @@ handle_block_candidate_reply({_Result, OldTopHash}, State)
     start_mining(State#state{block_candidate = undefined});
 handle_block_candidate_reply({Result,_OldTopHash}, State) ->
     case Result of
-        {ok, BlockCandidate, RandomNonce} ->
+        {ok, BlockCandidate, BlockCandidateTrees, RandomNonce} ->
```
gorillainduction: The BlockCandidateTrees should just be the root hash to avoid copying.

Again, this is out of scope for this ticket.

apps/aecore/src/aec_blocks.erl
```diff
@@ -104,15 +92,16 @@ new(LastBlock, Txs, Trees0) ->
...
-    #block{height = Height,
-           prev_hash = LastBlockHeaderHash,
-           root_hash = aec_trees:hash(Trees),
-           trees = Trees,
-           txs_hash = TxsRootHash,
-           txs = Txs1,
-           target = target(LastBlock),
-           time = aeu_time:now_in_msecs(),
-           version = ?CURRENT_BLOCK_VERSION}.
+    NewBlock =
```
gorillainduction: I would rename new/3 to new_with_state/3 and remove the {ok, _, _} and only return a 2-tuple.

New is no longer a desciptive name. If we want a function that don't care about the state, we can add a new/3 as well.

apps/aecore/test/aec_chain_state_tests.erl
```diff
@@ -132,35 +139,35 @@ basic_access_test_block_chain() ->
     ?assertEqual(BH2, top_header(State2)),
 
     %% Check highest block
-    ?compareBlockResults({ok, B0}, {ok, top_block(State0)}),
-    ?compareBlockResults({ok, B1}, {ok, top_block(State1)}),
-    ?compareBlockResults({ok, B2}, {ok, top_block(State2)}),
+    ?compareBlockResults({ok, B0}, {ok, top_block(State0)}), %% TODO Check state trees too.
```
gorillainduction: Remove these TODOs as they were never part of this test. The macro removes the state anyway, and the root hash is already compared.

apps/aecore/test/aec_test_utils.erl
```diff
@@ -135,46 +138,58 @@ unmock_block_target_validation() ->
...
+genesis_block_and_state() ->
+    {ok, B, S} = genesis_block_with_state(),
+    {B, S}.
 
 %% Generic blockchain with only coinbase transactions
 gen_block_chain(Length) ->
     gen_block_chain(Length, ?PRESET_ACCOUNTS).
-  
+
+gen_block_chain_without_state(Length) ->
```
gorillainduction: This is a missnomer. The block chain is generated with the state. It just doesn't return it.

It would be better to pick a new name, or to rename the internal functions.

apps/aecore/test/aec_chain_state_tests.erl
```diff
@@ -372,17 +386,17 @@ broken_chain_test_() ->
 
 broken_chain_postponed_validation() ->
     MainBC = gen_block_chain_by_target([?GENESIS_TARGET, 2, 5], 111),
-    AltChain = [B0, B1, B2, B3] = gen_block_chain_by_target([?GENESIS_TARGET, 2, 1], 222),
+    AltChain = [{_B0,S0}, {B1,S1}, {B2,S2}, {B3,S3}] = gen_block_chain_by_target([?GENESIS_TARGET, 2, 1], 222),
```
gorillainduction: The S* are unused, right? Better to have a util function that doesn't return these anyway.

apps/aecore/src/aec_blocks.erl
lucafavatella: Targeted by #513 (WIP).

apps/aecore/src/aec_conductor.erl
lucafavatella: Targeted by #513 (WIP).

PR #515, opened 2017-12-19
switch to lager fork which doesn't flush events

uwiger:
Note: PR is against oracle-master


uwiger:
PR refererence: https://www.pivotaltracker.com/story/show/153772726

PR #516, opened 2017-12-19
Pings from other chains should not propagate

ThomasArts:
If a peer from a different chain (different genesis block) was connecting to us, we would accept this and ping all its peers, who would all strating ping us, .... 

Fix: if genesis block differs, ignore the node (but not block it, it might just be a wrongly configured node and it gets another chance).

In order to test better we refactor http request into aeu_http_client module

[finished PT-153779356]

### DIFF
--- a/apps/aeutils/src/aeu_http_client.erl
+++ b/apps/aeutils/src/aeu_http_client.erl
@@ -1,0 +1,72 @@
+-module(aeu_http_client).
+
+%% API
+-export([request/4]).
+
+
+request(BaseUri, Method, Endpoint, Params) ->
+    Timeout = aeu_env:user_config_or_env(
+                [<<"http">>, <<"external">>,<<"request_timeout">>],
+                aehttp, http_request_timeout, 1000),
+    CTimeout = aeu_env:user_config_or_env(
+                 [<<"http">>, <<"external">>, <<"connect_timeout">>],
+                 aehttp, http_connect_timeout, min(Timeout, 1000)),
+    HTTPOptions = [{timeout, Timeout}, {connect_timeout, CTimeout}],
+    request(BaseUri, Method, Endpoint, Params, [], HTTPOptions, []).
+
+request(BaseUri, get, Endpoint, Params, Header, HTTPOptions, Options) ->
+    URL = binary_to_list(
+            iolist_to_binary(
+              [BaseUri, Endpoint, encode_get_params(Params)])),
+    lager:debug("GET URL = ~p", [URL]),
+    R = httpc:request(get, {URL, Header}, HTTPOptions, Options),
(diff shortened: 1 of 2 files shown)


apps/aeutils/src/aeu_requests.erl
```diff
@@ -42,16 +42,20 @@ ping(Peer) ->
...
         {ok, Map} ->
             lager:debug("ping response (~p): ~p", [Uri, pp(Map)]),
-            aec_sync:compare_ping_objects(PingObj, Map),
-            check_returned_source(Map, list_to_binary(aec_peers:uri(Peer))),
-            {ok, Map};
+            case aec_sync:compare_ping_objects(PingObj, Map) of
+                ok ->  
+                    check_returned_source(Map, list_to_binary(aec_peers:uri(Peer))),
+                    {ok, Map};
+                Error ->
```
lucafavatella: Match against `{error, _}`, nice to have.

apps/aeutils/src/aeu_http_client.erl
```diff
@@ -0,0 +1,72 @@
...
+request(BaseUri, post, Endpoint, Params, Header, HTTPOptions, Options) ->
+    URL = BaseUri ++ Endpoint,
+    {Type, Body} = case Params of
+                       Map when is_map(Map) ->
+                           %% JSON-encoded
+                           lager:debug("JSON-encoding Params: ~p",[Params]),
+                           {"application/json", jsx:encode(Params)};
+                       [] ->
+                           {"application/x-www-form-urlencoded",
+                            http_uri:encode(Endpoint)}  %% is this correct??
```
lucafavatella: Please prefer splitting refactoring and code change in separate commits, especially when performing changes difficult to track in git (e.g. portion of file split in other file). Though not needed for this PR (nice to have).

apps/aeutils/src/aeu_requests.erl
```diff
@@ -42,16 +42,20 @@ ping(Peer) ->
             {error, Reason};
         {ok, Map} ->
             lager:debug("ping response (~p): ~p", [Uri, pp(Map)]),
-            aec_sync:compare_ping_objects(PingObj, Map),
-            check_returned_source(Map, list_to_binary(aec_peers:uri(Peer))),
-            {ok, Map};
```
lucafavatella: I see - the caller in aec_peers always ended up here.

apps/aeutils/src/aeu_requests.erl
ThomasArts: That's really a matter of taste. We have dialyzer to verify that indeed the Error always is {error, _} and therefore don't need to write it explicitely.

PR #517, opened 2017-12-20
Initial drop of the aeoracle application

gorillainduction:
Most basic functionality is present, but the oracles are not ready for public announcement yet.

In particular, the types of query and response are not checked.

We are also lacking in documentation, but the specs are available at https://github.com/aeternity/protocol/blob/master/oracles/oracles.md

This PR finishes most of the oracle PT tickets. We will manually fix this.


rebar.config
```diff
@@ -1,7 +1,7 @@
 {erl_opts, [debug_info, {parse_transform, lager_transform},
             {lager_extra_sinks, [epoch_mining, epoch_pow_cuckoo,
                                  epoch_metrics]}]}.
-{deps, [{lager, ".*", {git, "https://github.com/erlang-lager/lager.git", {tag, "3.5.1"}}},
+{deps, [{lager, ".*", {git, "https://github.com/aeternity/lager.git", {ref, "5abb0c1"}}},
```
lucafavatella: Does this close a Pivotal ticket?

apps/aeoracle/src/aeoracle.app.src
```diff
@@ -0,0 +1,16 @@
+{application, aeoracle,
+ [{description, "AE Oracle"},
+  {vsn, "0.1.0"},
+  {registered, []},
+  {applications,
+   [kernel,
+    stdlib,
+    lager,
+    crypto
+   ]},
```
uwiger: Now, `aeoracles` has code dependencies against both `aecore` and `aeutils`, which is a bit unfortunate, but I assume outside the scope of this PR. The dependency to `aecore` is circular (at least from a static perspective), and `aeutils` is not a runnable app, so I guess this list is correct.

rebar.config
uwiger: Yes, https://www.pivotaltracker.com/story/show/153772726

gorillainduction: https://www.pivotaltracker.com/n/projects/2124891/stories/153772726

apps/aeoracle/src/aeo_interaction.erl
```diff
@@ -0,0 +1,205 @@
...
+    , #{<<"oracle_address">>  := OracleAddress}
+    , #{<<"response">>        := Response}
+    , #{<<"expires">>         := Expires}
+    , #{<<"response_ttl">>    := RespTTLValue}
+    , #{<<"fee">>             := Fee}
+    ] = List,
+    #interaction{ sender_address = SenderAddress
+                , sender_nonce   = SenderNonce
+                , oracle_address = OracleAddress
+                , response       = Response
```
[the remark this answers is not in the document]
gorillainduction: Found and fixed

apps/aeoracle/src/aeoracle.app.src
gorillainduction: The circularity to aecore is unavoidable unless we merge the applications.

The aeutils should probably be accessed from all apps, but preferably not call any other app.

apps/aeoracle/test/aeo_interaction_tests.erl
```diff
@@ -0,0 +1,88 @@
...
+
+basic_test_() ->
+    [ {"Serialization test", fun basic_serialize/0}
+    , {"Access test", fun basic_getters/0}
+    , {"Update test", fun basic_setters/0}
+    ].
+
+basic_serialize() ->
+    O = new(query_tx(), 1),
+    ?assertEqual(O, deserialize(serialize(O))),
```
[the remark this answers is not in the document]
gorillainduction: Why is this not picked up in the eunit_wrapper? Do we need to add another eunit_SUITE?

gorillainduction:
@lucafavatella Bumped the DB path as requested.

apps/aeoracle/src/aeo_state_tree.erl
```diff
@@ -0,0 +1,166 @@
...
+-export([ interaction_list/1
+        , oracle_list/1
+        ]).
+-endif.
+
+%%%===================================================================
+%%% Types
+%%%===================================================================
+
+-type mtree() :: aeu_mtrees:tree(binary(), binary()).
```
lucafavatella: It would be nice putting more descriptive types than `binary()` for key and value, ideally with type but at least with a var name e.g. `OracleEnabledAccountPubKey :: binary()` and similarly for the value.

apps/aeoracle/src/aeo_state_tree.erl
```diff
@@ -0,0 +1,166 @@
...
+
+-spec root_hash(tree()) -> {ok, aeu_mtrees:root_hash()} | {error, empty}.
+root_hash(#oracle_tree{mtree = MTree}) ->
+    aeu_mtrees:root_hash(MTree).
+
+-spec oracle_list(tree()) -> list(oracle()).
+oracle_list(#oracle_tree{mtree = MTree}) ->
+    ToOracle = fun(MaybeO) ->
+                   try [aeo_oracles:deserialize(MaybeO)]
+                   catch _:_ -> [] end end,
```
lucafavatella: I am concerned about this catch. Should not invalid oracles not slip in the trees in the first place?

apps/aeoracle/src/aeo_state_tree.erl
```diff
@@ -0,0 +1,166 @@
...
+    [ O || {_, Val} <- aeu_mtrees:to_list(MTree), O <- ToOracle(Val) ].
+
+-spec interaction_list(tree()) -> list(interaction()).
+interaction_list(#oracle_tree{mtree = MTree}) ->
+    ToInter = fun(MaybeI) ->
+                  try [aeo_interaction:deserialize(MaybeI)]
+                  catch _:_ -> [] end end,
```
lucafavatella: I am concerned about this catch. Should not invalid oracle interactions not slip in the trees in the first place?

lucafavatella: mmm is the same tree holding two types of items?

apps/aecore/src/aec_trees.erl
```diff
@@ -33,3 +39,34 @@ accounts(Trees) ->
 -spec set_accounts(trees(), aec_accounts_trees:tree()) -> trees().
 set_accounts(Trees, Accounts) ->
     Trees#trees{accounts = Accounts}.
+
+-spec oracles(trees()) -> aeo_state_trees:tree().
```
lucafavatella: aeo_state_trees -> aeo_state_tree ?

apps/aecore/src/aec_trees.erl
```diff
@@ -33,3 +39,34 @@ accounts(Trees) ->
...
+oracles(Trees) ->
+    Trees#trees.oracles.
+
+-spec set_oracles(trees(), aeo_state_trees:tree()) -> trees().
```
lucafavatella: aeo_state_trees -> aeo_state_tree ?

apps/aeoracle/src/aeo_state_tree.erl
lucafavatella: I see - this is test only.

apps/aeoracle/src/aeo_state_tree.erl
lucafavatella: I see - this is test only.

hanssv: This is a function only exported for test. (-ifdef(TEST).)

The try catch is there to only get interactions (and not oracles)

apps/aeoracle/src/aeo_register_tx.erl
```diff
@@ -0,0 +1,179 @@
...
+-spec version() -> non_neg_integer().
+version() ->
+    ?ORACLE_REGISTER_TX_VSN.
+
+%% -- Local functions  -------------------------------------------------------
+
+ensure_not_oracle(PubKey, Trees) ->
+    OraclesTree  = aec_trees:oracles(Trees),
+    case aeo_state_tree:lookup_oracle(PubKey, OraclesTree) of
+        {value, _Oracle} -> {error, account_is_already_an_oracle};
```
lucafavatella: Error reason appear misleading - it may be oracle interaction?

apps/aeoracle/src/aeo_state_tree.erl
hanssv: This is a function only exported for test. (-ifdef(TEST).)

The try catch is there to only get oracles (and not interactions)

apps/aeoracle/src/aeo_register_tx.erl
```diff
@@ -0,0 +1,179 @@
...
+origin(#oracle_register_tx{account = AccountPubKey}) ->
+    AccountPubKey.
+
+%% Account should exist, and have enough funds for the fee + the query_fee.
+-spec check(register_tx(), trees(), height()) -> {ok, trees()} | {error, term()}.
+check(#oracle_register_tx{account = AccountPubKey, nonce = Nonce,
+                          ttl = TTL, fee = Fee}, Trees, Height) ->
+    Checks =
+        [fun() -> aetx_utils:check_account(AccountPubKey, Trees, Height, Nonce, Fee) end,
+         fun() -> ensure_not_oracle(AccountPubKey, Trees) end,
```
lucafavatella: I see this is because processing of this tx enters new oracle (`aeo_state_tree:enter_oracle`). As tree hosts oracles and oracle interactions, this name is misleading. (Assuming keeping two value types in tree,) I suggest renaming as `is_not_defined` (i.e. not defined as either oracle or interaction).

apps/aeoracle/src/aeo_query_tx.erl
```diff
@@ -0,0 +1,202 @@
...
+                         query_fee = QueryFee} = Query, Trees0, Height) ->
+    AccountsTree0 = aec_trees:accounts(Trees0),
+    OraclesTree0  = aec_trees:oracles(Trees0),
+
+    Sender0 = aec_accounts_trees:get(SenderPubKey, AccountsTree0),
+    {ok, Sender1} = aec_accounts:spend(Sender0, QueryFee + Fee, Nonce, Height),
+    AccountsTree1 = aec_accounts_trees:enter(Sender1, AccountsTree0),
+
+    Interaction = aeo_interaction:new(Query, Height),
+    OraclesTree1 = aeo_state_tree:enter_interaction(Interaction, OraclesTree0),
```
lucafavatella: I understand assumption is that interaction is not there. I suggest prefering `aeo_state_tree:insert_interaction` to `aeo_state_tree:enter_interaction` following gb_trees API semantics wrt crash (nice to have, for clarity).

apps/aeoracle/src/aeo_register_tx.erl
```diff
@@ -0,0 +1,179 @@
...
+                            fee           = Fee} = RegisterTx, Trees0, Height) ->
+    AccountsTree0 = aec_trees:accounts(Trees0),
+    OraclesTree0  = aec_trees:oracles(Trees0),
+
+    Account0 = aec_accounts_trees:get(AccountPubKey, AccountsTree0),
+    {ok, Account1} = aec_accounts:spend(Account0, Fee, Nonce, Height),
+    AccountsTree1 = aec_accounts_trees:enter(Account1, AccountsTree0),
+
+    Oracle = aeo_oracles:new(RegisterTx, Height),
+    OraclesTree1 = aeo_state_tree:enter_oracle(Oracle, OraclesTree0),
```
lucafavatella: I understand assumption is that oracle is not there. I suggest prefering `aeo_state_tree:insert_oracle` to `aeo_state_tree:insert_interaction` following gb_trees API semantics wrt crash (nice to have, for clarity).

apps/aeoracle/src/aeo_query_tx.erl
```diff
@@ -0,0 +1,202 @@
...
+
+%% SenderAccount should exist, and have enough funds for the fee + the query_fee.
+%% Oracle should exist, and query_fee should be enough
+%% Fee should cover TTL
+%% TODO: Check the query? Check TTL relative to Oracle TTL?
+-spec check(query_tx(), trees(), height()) -> {ok, trees()} | {error, term()}.
+check(#oracle_query_tx{sender = SenderPubKey, nonce = Nonce,
+                       oracle = OraclePubKey, query_fee = QFee,
+                       query_ttl = TTL, fee = Fee}, Trees, Height) ->
+    Checks =
```
lucafavatella: In processing I see insertion of new interaction `aeo_state_tree:enter_interaction` though I do not see check that interaction is not there. Am I misunderstanding model, or maybe validation shall be here?

apps/aecore/src/aec_tx.erl
```diff
@@ -68,20 +72,33 @@ is_coinbase(Signed) ->
 %%% Internal functions
 %%%=============================================================================
 
-apply_on_state_trees(SignedTxs, Trees, Height) ->
-    apply_on_state_trees(SignedTxs, [], Trees, Height).
+apply_signed_common(SignedTxs, Trees0, Height, Strict) ->
+    Trees1 = aec_trees:perform_pre_transformations(Trees0, Height),
```
lucafavatella: I see - this means that at each and every block the oracles state tree must be pruned.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -772,22 +772,16 @@ update_top_block_hash(Hash, State) ->
 apply_node_transactions([Node|Left], Trees, State) ->
     Txs = aec_blocks:txs(Node#node.content),
     Height = node_height(Node),
-    {ok, FilteredTxs, NewTrees} = aec_tx:apply_signed(Txs, Trees, Height),
-    assert_transactions_valid(FilteredTxs, Txs),
-    assert_state_hash_valid(NewTrees, Node),
-    apply_node_transactions(Left, NewTrees, State);
+    case aec_tx:apply_signed_strict(Txs, Trees, Height) of
```
lucafavatella: Can aec_blocks:new_with_state be refactored to use same function?

apps/aeoracle/src/aeo_state_tree.erl
hanssv: Adding some proper types...

The var name would really be OracleEnabledAccountPubKeyOrOracleInteractionId which isn't very practical ;-)

apps/aeoracle/src/aeo_register_tx.erl
hanssv: If we assume that `check/3` is run before `process/3`, then we have indeed verified that the oracle is not there...

apps/aeoracle/src/aeo_register_tx.erl
hanssv: No, an oracle interaction will never be returned by lookup_oracle...

apps/aecore/src/aec_tx.erl
hanssv: Yes, and the cache we keep make that extremely cheap unless pruning actually has to be performed.

gorillainduction: There is a cache to make this efficient. Finding the next node that needs to be pruned is O(1).

apps/aeoracle/src/aeo_query_tx.erl
hanssv: Maybe explicit validation would be adding clarity... Since the sender nonce is part of the Id of the interaction check_account would stop an attempt to use the same nonce twice...

gorillainduction: This should indeed be validated.

apps/aeoracle/src/aeo_query_tx.erl
gorillainduction: Will fix

apps/aeoracle/src/aeo_register_tx.erl
gorillainduction: Will fix

apps/aeoracle/src/aeo_register_tx.erl
gorillainduction: Will fix

apps/aeoracle/src/aeo_register_tx.erl
gorillainduction: No. If there is an Id clash there will be an exception from aec_oracles:deserialize/1

apps/aeoracle/src/aeo_query_tx.erl
gorillainduction: Added validation.

apps/aeoracle/src/aeo_query_tx.erl
gorillainduction: Added insert_interaction as suggested.

apps/aeoracle/src/aeo_register_tx.erl
gorillainduction: Added insert_oracle and removed enter_oracle since there should be no use for that function.

apps/aecore/src/aec_chain_state.erl
gorillainduction: No. That function has slightly different semantics.

In aec_blocks:new_with_state/3 we actually want to filter away the unwanted transactions for the candidate generation. Here, there should be no unwanted transactions.

A subtle but important difference.

gorillainduction:
@lucafavatella I believe we have addressed your comments now. Please review again.

apps/aeoracle/test/aeo_interaction_tests.erl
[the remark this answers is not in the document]
gorillainduction: Answering myself. Yes we would need it. We skip this for now since the eunit target is run on PR anyway.

PR #518, opened 2017-12-20
[WIP] Warn if the ulimit -n is lower than recommended on node start

dincho:
[PT#153503921](https://www.pivotaltracker.com/story/show/153503921)


dincho:
closing in favour of #525

PR #520, opened 2017-12-20
Re-enable Travis OSX builds

dincho:


.travis.yml
```diff
@@ -58,16 +58,16 @@ deploy:
...
-#     - os: osx
-#       osx_image: xcode8.3 # [`xcode8.3` is Xcode 8.3.3 on OS X 10.12](https://docs.travis-ci.com/user/reference/osx#OS-X-Version)
-#       language: generic
-#       env: JOB=package
+matrix:
+  include:
+    - os: osx
+      osx_image: xcode8.3 # [`xcode8.3` is Xcode 8.3.3 on OS X 10.12](https://docs.travis-ci.com/user/reference/osx#OS-X-Version)
+      language: generic
+      env: JOB=test
```
lucafavatella: Please enable only package job unless you are confident OSX workers consume jobs enough.

.travis.yml
```diff
@@ -58,16 +58,16 @@ deploy:
...
+    - os: osx
+      osx_image: xcode8.3 # [`xcode8.3` is Xcode 8.3.3 on OS X 10.12](https://docs.travis-ci.com/user/reference/osx#OS-X-Version)
+      language: generic
+      env: JOB=package
```
lucafavatella: Please mark OSX package job as allowed failure and enable fast finish. This way OSX package job does not impact timely feedback on PR/branch/tag, though allow best effort package building.
Only issue may be unwanted upload of delayed job on tag, though we can reasonably prevent that by killing OSX package job on tag that we got bored to wait for.

lucafavatella:
@dincho Reviewed - though main reviewer is Travis CI :) and it seems slow.

.travis.yml
dincho: @lucafavatella Re-enabling OSX builds is mainly for statistics purposes ATM.
See the reference email from @travis in #devops.

However, it looks like it won't start for 20 minutes already, which is still a big problem.

lucafavatella: yea that is why I suggest marking it in allow_failures with fast_finish. I assume this will still be ok for statistics.

dincho:
@lucafavatella force pushed requested changes

PR #521, opened 2017-12-20
Pt 153809591 make ring library app

happi:
Removed application ans supervisor from the aering app.


happi:
This PR is based on PR 514 and should be rebased on master after PR514 has been merged.

PR #522, opened 2017-12-20
Wrap configs for lager logs [Finishes #153695043]

uwiger:


uwiger:
Isn't that roughly the average for any run-of-the-mill Java app? ;-)
We can reduce it. Any suggestion?

uwiger:
changed to 4MB per file, squashed and force-pushed

PR #523, opened 2017-12-20
Optimize PostBlock of block with height 0 [Finishes #153180259]

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/153180259)
When we send a genesis block via the HTTP interface - we clean up the empty JSON fields; upon receiving them we should populate them back to produce a serializable JSON structure.
Some points taken into account:
* we don't want to call the `aec_block_genesis:genesis_block_with_state/0` on every post (gensis) block we get because it has side effects (reads from disc)
* we don't want to get the genesis from the chain since it results in a unneeded application depenency

### DIFF
--- a/apps/aecore/src/aec_block_genesis.erl
+++ b/apps/aecore/src/aec_block_genesis.erl
@@ -28,7 +28,14 @@
           genesis_block_with_state/0,
           populated_trees/0 ]).
 
+-export([prev_hash/0,
+         pow/0,
+         txs_hash/0,
+         transactions/0]).
+
+-ifdef(TEST).
 -export([genesis_block_with_state/1]).
+-endif.
 
 -include("common.hrl").
 -include("blocks.hrl").
@@ -37,6 +44,18 @@
 genesis_header() ->
     {B, _S} = genesis_block_with_state(),
     aec_blocks:to_header(B).
+
+prev_hash() ->
+    <<0:?BLOCK_HEADER_HASH_BYTES/unit:8>>.
+
+txs_hash() ->
+    <<0:?TXS_HASH_BYTES/unit:8>>.
+
+pow() ->
+    no_value.
+
+transactions() ->
+    [].
 
 %% Returns the genesis block and the state trees.
 %%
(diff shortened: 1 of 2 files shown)


apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -189,17 +189,11 @@ empty_fields_in_genesis() ->
 % if this changes - both functions should be changes:
 % empty_fields_in_genesis/0 and values_for_empty_fields_in_genesis/0
 values_for_empty_fields_in_genesis() ->
```
velzevur: This function mimics the behaviour of `aec_blocks:serialize_to_map/1` for the fields at hand.
I was looking for a _proper_ way of creating an inproper `#block{}` (having only those fields) but I'd rather not polute aec_blocks' interface.

apps/aecore/src/aec_block_genesis.erl
```diff
@@ -54,11 +73,12 @@ genesis_block_with_state(Map) ->
...
            version = ?GENESIS_VERSION,
            height = ?GENESIS_HEIGHT,
-           prev_hash = <<0:?BLOCK_HEADER_HASH_BYTES/unit:8>>,
-           txs_hash = <<0:?TXS_HASH_BYTES/unit:8>>,
+           prev_hash = prev_hash(),
+           txs_hash = txs_hash(),
            root_hash = aec_trees:hash(Trees),
            target = ?HIGHEST_TARGET_SCI,
-           pow_evidence = no_value,
+           txs = [],
```
lucafavatella: call to `transactions()` here please

velzevur:
Now `aec_block_genesis:transactions/0` is called in a gen block creation.

apps/aecore/src/aec_block_genesis.erl
```diff
@@ -38,6 +45,18 @@ genesis_header() ->
     {B, _S} = genesis_block_with_state(),
     aec_blocks:to_header(B).
 
+prev_hash() ->
+    <<0:?BLOCK_HEADER_HASH_BYTES/unit:8>>.
+
+txs_hash() ->
+    <<0:?TXS_HASH_BYTES/unit:8>>.
+
+pow() ->
```
ThomasArts: This probably should be called pow_evidence to be compatible with terminology.

PR #525, opened 2017-12-20
Ulimit start warning

uwiger:
Based on #518. Rebased against master, and re-wrote the `cp_dir()` function in `aec_suite_utils` to be more portable, since the `sys.config` and `vm.ags` files ended up being deleted when running on Ubuntu.


hooks/pre_start.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+OPEN_FILES_RECOMMENDED=1024
```
sennui: I think this is too low. We mine a block/~5min, each kept in separate file (sure, we close FDs, but what if we restart and re-read?). Also each outgoing connection to a peer takes an FD. Can we make it 1024*256?

dincho: You mean 262144?

Default hard limit on linux (ubuntu 16) is 65536. Raising hard limit is more sophisticated for users and they'd have hard time raising it.

```
vagrant@epoch:/app$ ulimit -Hn
65536
```

And this is OSX:
```
$ sysctl -a | grep files
kern.maxfiles: 49152
kern.maxfilesperproc: 24576
kern.num_files: 4633
```

dincho: So, 24576 is the max portable (for our supported operating systems: Ubuntu & OSX).

lucafavatella:
@dincho  Shall #518 be closed than?

dincho:
> @dincho Shall #518 be closed than?

yes, closed.

hooks/pre_start.sh
sennui: Agreed on slack that 24576 is fine for average user

dincho:
I tested this on Ubuntu 16. Travis is testing on Ubuntu 14.

dincho:
Tests are also passing on OSX 10.13.1

hooks/pre_start.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+OPEN_FILES_RECOMMENDED=24576
+OPEN_FILES_LIMIT=$(ulimit -n)
+
+if [ "$OPEN_FILES_RECOMMENDED" -gt "$OPEN_FILES_LIMIT" ]; then
+    echo "WARNING: ulimit -n is $OPEN_FILES_LIMIT; $OPEN_FILES_RECOMMENDED is the recommended minimum."
+    echo "Try 'ulimit -n $OPEN_FILES_RECOMMENDED' before starting the node."
```
lucafavatella: After having run `bin/epoch start` and got this warning (and the node actually had started), I ran `ulimit -n 24576` as recommended then I re-ran `bin/epoch start` as the message would suggest. And this second invocation did not raise warning and returned exit status zero. Though the OS PID is the same! (I checked with `pgrep beam`).
So it looks to me that user experience is not optimal: user thinks that actually started node with correct ulimit though he/she did not .

lucafavatella: I suggest amending the message slightly in order not to induce to user to think that, after getting warning, simply running `ulimit -n` and then run `bin/epoch start` again and not getting warning fixed it. I suggest the following:
```
echo "WARNING: ulimit -n is $OPEN_FILES_LIMIT; $OPEN_FILES_RECOMMENDED is the recommended minimum."
echo "You are recommended to ensure the node is stopped and raise the maximum number of open files (try 'ulimit -n $OPEN_FILES_RECOMMENDED') before starting the node."
```

dincho: Unfortunately this the best we can do for now without providing custom start script (and trashing rebar's one).

dincho:
@lucafavatella I updated the message in 4b24cf5c34383946e62ab47aede47b237571fe4c

PR #527, opened 2017-12-21
Check config start script extension

dincho:
[PT 153509936](https://www.pivotaltracker.com/story/show/153509936).

Purpose: Makes usage of check_config much easier than https://github.com/aeternity/epoch/commit/4de3bbe0ad87c299e89720ef115e82667ea9eeb5#diff-89f3cba2993eebd391f32c27f3772d7bR115

Usage: `bin/epoch check_config epoch.yaml`

### DIFF
--- a/apps/check_config/priv/extensions/check_config.sh
+++ b/apps/check_config/priv/extensions/check_config.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+# Simple wrapper to run check_config
+# Relx extensions cannot be binary files but sourced bash files only
+
+$SCRIPT_DIR/check_config $@
--- a/apps/check_config/priv/extensions/check_config.sh
+++ b/apps/check_config/priv/extensions/check_config.sh
@@ -2,5 +2,7 @@
 
 # Simple wrapper to run check_config
 # Relx extensions cannot be binary files but sourced bash files only
+# Also append ERTS BINDIR to PATH to run the escript
 
+PATH=$BINDIR:$PATH
 $SCRIPT_DIR/check_config $@
--- a/rebar.config
+++ b/rebar.config
@@ -32,11 +32,16 @@
                    {copy, "VERSION" , "VERSION"},
                    {mkdir, "data/aecore/.genesis"},
                    {copy, "data/aecore/.genesis/accounts.json", "data/aecore/.genesis/accounts.json"},
-                   {copy, "hooks/pre_start.sh", "bin/hooks/pre_start.sh"}]},
+                   {copy, "hooks/pre_start.sh", "bin/hooks/pre_start.sh"},
+                   {copy, "apps/check_config/priv/extensions/check_config.sh", "bin/extensions/check_config"}
+                   ]},
 
         {extended_start_script, true},
         {extended_start_script_hooks, [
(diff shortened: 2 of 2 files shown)


lucafavatella:
@dincho Shall 4959643d39a05652b9a6160d09c9e4de75b03076 be included?

rebar.config
```diff
@@ -31,8 +31,22 @@
...
+        {extended_start_script_hooks, [
+          {pre_start, [{custom, "hooks/pre_start.sh"}]}
+        ]},
+        {extended_start_script_extensions, [
+              {check_config, "extensions/check_config"}
+        ]}]
+}.
+
+{plugins,
+  [{tx_xform, {git, "https://github.com/aeternity/tx_xform.git", {ref, "d57aaed"}}}]}.
```
dincho: probably merged mistake, should be removed

dincho:
> @dincho Shall 4959643 be included?

no see https://github.com/aeternity/epoch/pull/527#discussion_r158231437

lucafavatella:
For reference: currently release notes do: `( export PATH="$(pwd)/$(ls -dr erts-* | head -n 1)/bin:$PATH" && bin/check_config epoch.yaml; )`

dincho:
My test (after the fix)
```
===> tarball /app/_build/prod/rel/epoch/epoch-0.3.5.tar.gz successfully created!
vagrant@epoch:/app$ mkdir /tmp/node; cd /tmp/node
vagrant@epoch:/tmp/node$ tar xf /app/_build/prod/rel/epoch/epoch-0.3.5.tar.gz
vagrant@epoch:/tmp/node$ bash
vagrant@epoch:/tmp/node$ PATH="/bin:/usr/bin"
vagrant@epoch:/tmp/node$ ./bin/epoch check_config
Usage: check_config file [schema]
```

apps/check_config/priv/extensions/check_config.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Simple wrapper to run check_config
+# Relx extensions cannot be binary files but sourced bash files only
+# Also append ERTS BINDIR to PATH to run the escript
+
+PATH=$BINDIR:$PATH
```
lucafavatella: @dincho  Does `$SCRIPT_DIR/check_config` below fork? If so, does this line export the PATH for children?

Re-testing now.

lucafavatella: Apparently it worked for me now in one run, so the PATH problem worked for me (macOS 10.13.2)

dincho: It works in my tests, I didn't wanter to pollute the environment if there is no explicit need.

uwiger:
A minor quibble:
```
uwpro-2:epoch uwiger$ _build/dev1/rel/epoch/bin/epoch check_config             
Usage: check_config file [schema]
```

The usage text should perhaps be updated to reflect the new way of using the script.

lucafavatella:
> The usage text should perhaps be updated to reflect the new way of using the script.

Tracked as https://www.pivotaltracker.com/story/show/153848785
Merging.

PR #532, opened 2017-12-21
Fix unreliable start_mining

hanssv:
The initial timeout delivery mechanism is not reliable, this is a workaround.


lucafavatella:
This finished PT-153851716

PR #533, opened 2017-12-21
Pt 153809803 ring icode

happi:
Add intermediate code and an assembler and byte code phase to the ring compiler.

[Finishes 153809803]

### DIFF
--- a/apps/aering/src/aer_compiler.erl
+++ b/apps/aering/src/aer_compiler.erl
@@ -14,6 +14,8 @@
 
 -export([test/0]).
 
+-include("aer_icode.hrl").
+
 file(Filename) ->
     file(Filename, []).
 
@@ -68,9 +70,9 @@ ast_args([{arg, _, Name, Type}|Rest], Acc) ->
 ast_args([], Acc) -> lists:reverse(Acc).
                                  
 
-ast_body({id, _,_Name},_Icode) ->
+ast_body({id, _, Name},_Icode) ->
     %% TODO Look up id in env
-    ['DUP'].
+    #var_ref{name = Name}.
     
 
 ast_fun_to_icode(Name, Args, Body, #{functions := Funs} = Icode) ->
--- a/apps/aering/src/aer_compiler.erl
+++ b/apps/aering/src/aer_compiler.erl
@@ -5,7 +5,8 @@
 %%%     Compiler from Aeterinty Ring language to the Aeternity VM, aevm.
 %%% @end
 %%% Created : 12 Dec 2017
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -18,6 +18,9 @@ endif
 
 PYTHON_DIR = py
 PYTHON_TESTS = $(PYTHON_DIR)/tests
+PIP = $(PYTHON_BIN)/pip
```
velzevur: Do we need pip here? I can not see it being used

apps/aering/src/aer_compiler.erl
```diff
@@ -5,7 +5,8 @@
 %%%     Compiler from Aeterinty Ring language to the Aeternity VM, aevm.
 %%% @end
 %%% Created : 12 Dec 2017
-%%% aec_conductor:stop_mining(),aer_compiler:file(a,c).
+%%% aec_conductor:stop_mining().
+%%% aer_compiler:file( identity, [pp_ast,pp_icode,pp_assembler,pp_bytecode, pp_ring_code]).
```
UlfNorell: Should this be here? If so maybe add a short explanation.

apps/aering/src/aer_icode_to_asm.erl
```diff
@@ -0,0 +1,63 @@
...
+                                     , address := Address} = Env, Code) ->
+    Env0 = add_function(Name, Address, Env),
+    {BodyCode, NewEnv} = assemble_body(Body, add_args(Args, Env0, FS), Code),
+    assemble(More, NewEnv, BodyCode);
+assemble([], Env, Code) -> {Env, Code}.
+
+
+assemble_body({var_ref, Id}, #{sp := SP} = Env, Code) ->
+    SL = lookup_var(Id, Env),
+    Instr = aeb_opcodes:mnemonic(?DUP1 + SP + SL),
```
UlfNorell: Can we have a function to compute a suitable `?DUP` (and others) opcode, so that the knowledge that `?DUP2 = ?DUP1 + 1` is more contained?

apps/aering/src/aer_compiler.erl
happi: Temporary stuff to run from the command shell.. I'll leave it in for now.

apps/aering/src/aer_icode_to_asm.erl
happi: Ill make a ticket.

Makefile
happi: It is used when you pull in the external test library.

PR #534, opened 2017-12-21
Refactor HTTP endpoint for /info

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/153780478)
* created a public /version for returning the `version`, `revision` and `genesis hash`
* protected /info (to be active only when a setting is on) and added more data to it - block difficulties
* when setting is off, returning `401` (instead of `404`)

### DIFF
--- a/apps/aehttp/priv/swagger.json
+++ b/apps/aehttp/priv/swagger.json
@@ -513,6 +513,31 @@
             "schema" : {
               "$ref" : "#/definitions/AccountsBalances"
             }
+          },
+          "403" : {
+            "description" : "Balances not enabled",
+            "schema" : {
+              "$ref" : "#/definitions/Error"
+            }
+          }
+        },
+        "security" : [ ]
+      }
+    },
+    "/version" : {
+      "get" : {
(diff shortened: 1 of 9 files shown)


apps/aehttp/priv/swagger.json
```diff
@@ -532,6 +551,12 @@
             "schema" : {
               "$ref" : "#/definitions/Info"
             }
+          },
+          "401" : {
```
dincho: I don't know if 401 (Unauthorized) is requirement, but I think 403 (Forbidden) if semantically better.

https://httpstatuses.com/401
https://httpstatuses.com/403

velzevur:
Changed return HTTP code for disabled endpoints from `401` to `403`

config/swagger.yaml
```diff
@@ -726,6 +754,8 @@ definitions:
       time_delta_to_parent:
```
lucafavatella: I see - "time_delta_to_parent" is present only for non-genesis blocks.

PR #537, opened 2017-12-22
Resolve test issues with mining difficutly

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/153780170)
The problem used to be recalibration of the mining difficulty durring tests - when a test tried to mine the eleventh block - it took more than 30s and the `aecore_suit_tests:mine_blocks_loop/1` timed out. Curriously   all tests used to mine right up to block 10, otherwise we would have found it earlier.

Also cleaned up a little bit `aecore_suite_utils` interface, removing `mine_one_block/1` function.

### DIFF
--- a/apps/aecore/test/aecore_suite_utils.erl
+++ b/apps/aecore/test/aecore_suite_utils.erl
@@ -12,8 +12,8 @@
 
 -export([start_node/2,
          stop_node/2,
-         mine_one_block/1,
-         mine_blocks/2]).
+         mine_blocks/2,
+         mine_blocks/3]).
 
 -export([node_tuple/1,
          node_name/1,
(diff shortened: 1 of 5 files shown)


apps/aecore/test/aecore_suite_utils.erl
```diff
@@ -93,27 +93,12 @@ stop_node(N, Config) ->
...
-            ct:log("timeout waiting for block event~n"
-                   "~p", [process_info(self(), messages)]),
-            error(timeout_waiting_for_block)
-    end.
-
 mine_blocks(Node, NumBlocksToMine) ->
+    mine_blocks(Node, NumBlocksToMine, 100).
+
+mine_blocks(Node, NumBlocksToMine, MiningRate) ->
+    rpc:call(Node, application, set_env, [aecore, expected_mine_rate, MiningRate],
```
velzevur: This could be tricky because it has side effects on the node doing the mining. If you find a valid reason to set back the original `expected_mine_rate` (before the mining) after the mining is finished, please comment bellow.

velzevur:
@ThomasArts it is 30 seconds _per block_ as it used to be. The difference is actually in setting:
``` rpc:call(Node, application, set_env, [aecore, expected_mine_rate, MiningRate], 100)```
where `100` being the default value. This sets the node aiming at a block in every 100ms. The timeouts were caused by expecting the node to be mining a block once in every 10 minutes.

File: apps/aehttp/test/aehttp_integration_SUITE.erl [hunk 2]
```diff
@@ -437,12 +435,23 @@ all_accounts_balances(_Config) ->
...
     aecore_suite_utils:mine_blocks(aecore_suite_utils:node_name(?NODE),
                                    BlocksToMine),
+    ReceiversAccounts = [{random_hash(), Idx} || Idx <- lists:seq(1, Receivers)],
+    lists:foreach(
+        fun({ReceiverPubKey, AmountToSpent}) ->
+            {ok, 200, _} = post_spend_tx(ReceiverPubKey, AmountToSpent, Fee)
+        end,
+        ReceiversAccounts),
+
+    % mine a block to include the txs
```
lucafavatella: I expect this test to eventually show to be fragile, as the mempool is not checked for inclusion of all txs.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -185,7 +185,7 @@ get_top_empty_chain(_Config) ->
     ok.
 
 get_top_non_empty_chain(_Config) ->
-    ok = aecore_suite_utils:mine_one_block(aecore_suite_utils:node_name(?NODE)),
+    {ok, _} = aecore_suite_utils:mine_blocks(aecore_suite_utils:node_name(?NODE), 1),
```
lucafavatella: I see match of return value (without binding of variables) here though not in most other calls to `aecore_suite_utils:mine_blocks`: maybe slipped in other points? Anyway matching seems not relevant because of exception `erlang:error` raised in case of not-ok.

velzevur: Yes, the patten matching is not neccessary

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: I've refactor the test to _exclude_ the public addresses of the recipients for any transaction that didn't make it into the block and was left in the mempool.

lucafavatella: I see potentially all txs may have been left in mempool.

velzevur:
Asserted that all txs are included in the mined block

PR #542, opened 2017-12-22
Enable user configuration of Cuckoo miner and PoW

lucafavatella:


lucafavatella:
I ran CI on this PR merged on recent master (68697ac8) and it passed.

PR #545, opened 2017-12-28
HTTP endpoint enhancement - POST /tx

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/153780309)
Added some checks for data validity for incoming txs (inter node communication).
Handled similar case in operationId `GetBlockByHash`

### DIFF
--- a/apps/aehttp/priv/swagger.json
+++ b/apps/aehttp/priv/swagger.json
@@ -128,6 +128,12 @@
               "$ref" : "#/definitions/Block"
             }
           },
+          "400" : {
+            "description" : "Invalid hash",
+            "schema" : {
+              "$ref" : "#/definitions/Error"
+            }
+          },
           "404" : {
             "description" : "Block not found",
             "schema" : {
@@ -183,7 +189,19 @@
           "schema" : {
             "$ref" : "#/definitions/Tx"
           }
-        } ]
+        } ],
+        "responses" : {
+          "200" : {
+            "description" : "Successful operation"
+          },
+          "400" : {
+            "description" : "Invalid transaction",
+            "schema" : {
+              "$ref" : "#/definitions/Error"
+            }
+          }
+        },
+        "security" : [ ]
       }
     },
     "/transactions" : {
--- a/apps/aehttp/src/aehttp_dispatch_ext.erl
+++ b/apps/aehttp/src/aehttp_dispatch_ext.erl
@@ -48,26 +48,30 @@ handle_request('GetBlockByHeight', Req, _Context) ->
 
 handle_request('GetBlockByHash' = _Method, Req, _Context) ->
     lager:debug("got ~p; Req = ~p", [_Method, pp(Req)]),
-    Hash = base64:decode(maps:get('hash', Req)),
-    case aec_conductor:get_header_by_hash(Hash) of
-        {error, header_not_found} ->
-            {404, [], #{reason => <<"Block not found">>}};
-        {ok, Header} ->
-            {ok, HH} = aec_headers:hash_header(Header),
-            case aec_conductor:get_block_by_hash(HH) of
-                {ok, Block} ->
-                    %% swagger generated code expects the Resp to be proplist
-                    %% or map and always runs jsx:encode/1 on it - even if it
-                    %% is already encoded to a binary; that's why we use
-                    %% aec_blocks:serialize_to_map/1 instead of
-                    %% aec_blocks:serialize_for_network/1
-                    lager:debug("Block = ~p", [pp(Block)]),
-                    Resp =
-                      cleanup_genesis(aec_blocks:serialize_to_map(Block)),
-                    lager:debug("Resp = ~p", [pp(Resp)]),
-                    {200, [], Resp};
-                {error, block_not_found} ->
-                    {404, [], #{reason => <<"Block not found">>}}
+    case base64_safe_decode(maps:get('hash', Req)) of
+        {error, not_base64_encoded} ->
+            {400, [], #{reason => <<"Invalid hash">>}};
+        {ok, Hash} ->
+            case aec_conductor:get_header_by_hash(Hash) of
+                {error, header_not_found} ->
+                    {404, [], #{reason => <<"Block not found">>}};
+                {ok, Header} ->
+                    {ok, HH} = aec_headers:hash_header(Header),
+                    case aec_conductor:get_block_by_hash(HH) of
+                        {ok, Block} ->
+                            %% swagger generated code expects the Resp to be proplist
+                            %% or map and always runs jsx:encode/1 on it - even if it
+                            %% is already encoded to a binary; that's why we use
+                            %% aec_blocks:serialize_to_map/1 instead of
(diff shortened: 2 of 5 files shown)


apps/aehttp/priv/swagger.json
```diff
@@ -128,6 +128,12 @@
               "$ref" : "#/definitions/Block"
             }
           },
+          "400" : {
```
lucafavatella: Responses appears missing for POST /tx in swagger.json: issue in generation from YAML? inconsistent YAML indentation? something I am missing?

velzevur: Yes, there had been an identation issue in the yaml file. Patched now

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -95,29 +99,38 @@ handle_request('PostBlock', Req, _Context) ->
...
-    SerializedTx = maps:get(<<"tx">>, Tx),
-    SignedTx = aec_tx_sign:deserialize_from_binary(
-                 base64:decode(SerializedTx)),
-    lager:debug("deserialized: ~p", [pp(SignedTx)]),
-    PushRes = aec_tx_pool:push(SignedTx, tx_received),
-    lager:debug("PushRes = ~p", [pp(PushRes)]),
-    {200, [], #{}};
+    case base64_safe_decode(maps:get(<<"tx">>, Tx)) of
+        {error, not_base64_encoded} ->
+            {400, [], #{reason => <<"Invalid hash">>}};
```
lucafavatella: "Invalid hash" sounds incorrect.

lucafavatella: Maybe "Invalid base64-encoding"?

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -431,6 +498,16 @@ miner_balance(_Config) ->
     {ok, 200, #{<<"balance">> := Bal}} = get_balance(base64:encode(PubKey)),
     ok.
 
+balance_broken_hash(_Config) ->
```
lucafavatella: Test case name seems misleading. Maybe "balance_broken_address"?

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -413,11 +415,76 @@ post_correct_tx(_Config) ->
...
+    {ok, PubKey} = rpc(aec_keys, pubkey, []),
+    {ok, Nonce} = rpc(aec_next_nonce, pick_for_account, [PubKey]),
+    {ok, SpendTx} =
+        aec_spend_tx:new(
+          #{sender => PubKey,
+            recipient => random_hash(),
+            amount => Amount,
+            fee => Fee,
+            nonce => Nonce}),
+    {ok, SignedTx} = rpc(aec_keys, sign, [SpendTx]),
```
lucafavatella: I see most of the above is the same as in test case post_correct_tx. Factoring portions (maybe in aecore_suite_utils) would help readability of the test - though not required for this PR.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -413,11 +415,76 @@ post_correct_tx(_Config) ->
...
+    [T, V, SerializedTx, Sigs] = aec_tx_sign:serialize(SignedTx),
+    lists:foreach(
+        fun(Key) ->
+            BrokenTx = lists:filter(
+                fun(#{Key := _}) -> true;
+                (_) -> false
+                end,
+                SerializedTx),
+            EncodedBrokenTx = base64:encode(msgpack:pack([T, V, BrokenTx, Sigs])),
```
lucafavatella: I see this test tests that attempting to post a (base64-encoded) signed tx with the tx stripped of all except one of its fields is detected as broken. There are many ways a signed tx can be broken:
* Removed or added or shuffled fields in the signed tx envelope;
* Amended field values in the signed tx envelope;
* Removed or added or shuffled fields in the spend tx;
* Amended field values in the spend tx.

This approach for making a tx broken is one of the many possible, and is good enough. So no changes needed on this.

----

I also see this test has the implied assumption that the hardcoded list of fields below (type, vsn, sender, recipient, amount, fee, nonce) is the full list of fields of the spend tx, that is correct though makes the test fragile to future tx internal format changes. I recommend hardcoding this expectation in the test so to ease future updates of the test in case of future tx internal format change.

apps/aehttp/test/aehttp_integration_SUITE.erl
lucafavatella: This is only a test case rename so not strictly needed for this PR.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -413,11 +415,76 @@ post_correct_tx(_Config) ->
...
+                aec_spend_tx:new(
+                  #{sender => PubKey,
+                    recipient => random_hash(),
+                    amount => Amount,
+                    fee => Fee,
+                    nonce => Nonce}),
+            {ok, SignedTx} = rpc(aec_keys, sign, [SpendTx]),
+            <<_, BrokenHash/binary>> =
+                base64:encode(aec_tx_sign:serialize_to_binary(SignedTx)),
+            {ok, 400, #{<<"reason">> := <<"Invalid hash">>}} = post_tx(BrokenHash)
```
lucafavatella: See https://github.com/aeternity/epoch/pull/545/files#r159230045

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -413,11 +415,76 @@ post_correct_tx(_Config) ->
...
+        fun(_) ->
+            {ok, SpendTx} =
+                aec_spend_tx:new(
+                  #{sender => PubKey,
+                    recipient => random_hash(),
+                    amount => Amount,
+                    fee => Fee,
+                    nonce => Nonce}),
+            {ok, SignedTx} = rpc(aec_keys, sign, [SpendTx]),
+            <<_, BrokenHash/binary>> =
```
lucafavatella: I see apart from this line (that strips initial base64 char) is same as test case post_correct_tx, with new random fake recipient created each time.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -413,11 +415,76 @@ post_correct_tx(_Config) ->
...
+         <<"vsn">>,
+         <<"sender">>,
+         <<"recipient">>,
+         <<"amount">>,
+         <<"fee">>,
+         <<"nonce">>]),
+
+    ok.
+
+post_broken_hash_tx(_Config) ->
```
lucafavatella: Test case name seems misleading. Maybe "post_broken_base64_tx"?

This is only a test case rename so not strictly needed for this PR.

lucafavatella:
> Required:
> 
>     Error reason "Invalid hash" when posting tx with wrong base64 encoding is misleading

I see this addressed - thanks.

PR #548, opened 2018-01-02
user config of metrics reporting

uwiger:
See PT Issue: https://www.pivotaltracker.com/story/show/153632015

### DIFF
--- a/apps/aecore/src/aec_metrics.erl
+++ b/apps/aecore/src/aec_metrics.erl
@@ -34,6 +34,8 @@
 -record(st, {type_map = [] :: list(),
              prefix = []   :: string()}).
 
+-record(filter, {}).
+
 -record(statsd, {
           socket        :: inet:socket() | undefined,
           tref          :: reference() | undefined,
@@ -43,6 +45,9 @@
           reconnect_interval = ?DEFAULT_RECONNECT_INTERVAL :: pos_integer()
          }).
 
+%% Hard-coded default
+default_dests() ->
+    [log, send].
 
 %%===================================================================
 %% Metrics API
@@ -82,6 +87,9 @@ start_reporters() ->
                                            {state, []}]},
                                   {output, [{mode, plugin},
                                             {module, ?MODULE},
+                                            {state, filter}]},
+                                  {output, [{mode, plugin},
+                                            {module, ?MODULE},
                                             {state, statsd}]},
                                   {output, [{mode, plugin},
                                             {module, ?MODULE},
@@ -113,19 +121,22 @@ subscriber_loop(Logger) ->
             subscriber_loop(Logger)
     end.
 
+logger_init_output(filter) ->
+    {ok, #filter{}};
 logger_init_output(log) ->
     {ok, log};
 logger_init_output(statsd) ->
     {ok, init_statsd()}.
 
-logger_handle_data(#{} = Data, log = St) ->
+logger_handle_data(#{} = Data, #filter{} = St) ->
+    apply_filter(Data, St);  % sets do_log and do_send flags
+logger_handle_data(#{do_log := true} = Data, log = St) ->
     epoch_metrics:info("~s", [line(Data)]),
     {Data, St};
-logger_handle_data(Data, log = St) ->
-    {Data, St};
-logger_handle_data(Data, #statsd{} = St) ->
-    statsd_handle_data(Data, St).
-
+logger_handle_data(#{do_send := true} = Data, #statsd{} = St) ->
+    statsd_handle_data(Data, St);
+logger_handle_data(Data, St) ->
+    {Data, St}.
 
 %%===================================================================
 %% exometer_report callbacks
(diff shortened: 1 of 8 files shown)


apps/aecore/src/aec_metrics_rpt_dest.erl
```diff
@@ -0,0 +1,185 @@
...
+                      Actions = maps:get(<<"actions">>, R, <<"log,send">>),
+                      Cmd = <<"type=",Type/binary,"/",
+                              Name/binary,"/",
+                              DPs/binary>>,
+                      try
+                          Found = aec_metrics_lib:find_entries_cmd([Cmd]),
+                          make_destinations(Found, Actions, Tab)
+                      catch
+                          error:E ->
+                              lager:error("CRASH: ~p~n~p~n",
```
dincho: Can we somehow remove the newlines here?

see https://github.com/aeternity/epoch/pull/456

apps/aecore/src/aec_metrics.erl
```diff
@@ -34,6 +34,8 @@
 -record(st, {type_map = [] :: list(),
              prefix = []   :: string()}).
 
+-record(filter, {}).
```
ThomasArts: Would it not be more flexible and future proof to have a map here instead? Just an idea, not needed at the moment.

apps/aecore/src/aecore_app.erl
```diff
@@ -19,6 +19,7 @@ start(_StartType, _StartArgs) ->
 start_phase(create_metrics_probes, _StartType, _PhaseArgs) ->
     aec_metrics:create_metrics_probes();
 start_phase(start_reporters, _StartType, _PhaseArgs) ->
+    aec_metrics_rpt_dest:check_config(),
```
lucafavatella: I suggest checking the return value here.

apps/aecore/src/aec_metrics_lib.erl
```diff
@@ -0,0 +1,199 @@
...
+    Parts = re:split(S, "/"),
+    lists:foldl(
+      fun(P, Acc) ->
+              parse_part(part_type(P), P, Acc)
+      end, #{name   => [],
+             type   => '_',
+             status => Status0,
+             dps    => default}, Parts).
+
+part_type(<<"type=", _/binary>> ) -> type;
```
ThomasArts: Would be nice to somehere (not at the cost of having jsx) have a function that parses binary strings into maps.

apps/aecore/src/aec_metrics_lib.erl
```diff
@@ -0,0 +1,199 @@
...
+    case split("**", Parts) of
+        {_, []} ->
+            [replace_parts_1(Parts)];
+        {Before, After} ->
+            Head = replace_parts_1(Before),
+            Tail = replace_parts_1(After),
+            [Head ++ Pad ++ Tail || Pad <- pads()]
+    end.
+
+pads() ->
```
ThomasArts: Really? Would a function pad(N) -> [ '_'  || _ lists:seq(1,N)] not be nicer and then pads(K) -> [ pads(N) || N <- lists:seq(1,K)].

apps/aecore/src/aec_metrics_rpt_dest.erl
```diff
@@ -0,0 +1,185 @@
...
+            T = ets:new(?TAB, [set, public, named_table,
+                               {keypos, #metric.key},
+                               {heir, self(), []}]),
+            T;
+        _ ->
+            ?TAB
+    end.
+
+populate_tab(Tab) ->
+    case aeu_env:user_map([<<"metrics">>,<<"rules">>]) of
```
ThomasArts: Here's the binary strings to map function that I refered to!

apps/aecore/src/aec_metrics.erl
uwiger: Actually, it would be more consistent right now to make it an atom; it's only used for pattern-matching in order to determine plugin context, and the 'log' plugin uses an atom instead.

apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -217,6 +217,42 @@
                     "How often (in ms) to try reconnecting to the daemon",
                     "type" : "integer",
                     "minimum" : 0
+                },
+                "rules" : {
+                    "description" :
+                    "Filter rules guiding logging/sending of metrics.",
```
lucafavatella: A short description of the default behaviour would help.

apps/aecore/src/aec_metrics_lib.erl
uwiger: Yes ... that was copied from riak_core, and those guys were pretty focused on speed.
My favorite solution would be to extent ets:select() match specs. :)

uwiger:
@lucafavatella that's correct. The default behavior is exactly as before, so no entry in the 'filter table' is interpreted as "both log and send to DD".

PR #549, opened 2018-01-03
Release notes for 0.4.1 [Finishes #153988152]

lucafavatella:
Depends on PR #547.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -239,7 +239,7 @@ mining:
             extra_args: ""
             node_bits: 28
 ```
-... then stop and start the node.
+... then stop and start the node (`( cd /tmp/node; bin/epoch stop; bin/epoch start; )`).
 
 #### Verify that node connected to the testnet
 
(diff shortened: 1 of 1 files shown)


RELEASE-NOTES.md
```diff
@@ -181,6 +179,68 @@ If the node successfully mines a block, you shall read log entries like the foll
...
+```
+... to ...
+```yaml
+mining:
+    autostart: true
+    cuckoo:
+        miner:
+            executable: lean28
+            extra_args: ""
+            node_bits: 28
```
dincho: Could we just change the executable and rely on other defaults? Showing `node_bits` to users would be confusing.

RELEASE-NOTES.md
```diff
@@ -181,6 +179,68 @@ If the node successfully mines a block, you shall read log entries like the foll
...
+```
+... then stop and start the node.
```
dincho: `/tmp/node/bin/epoch stop; /tmp/node/bin/epoch start` ?

RELEASE-NOTES.md
lucafavatella: No, because that would require a [schema change](https://github.com/aeternity/epoch/blob/133c78c047c54b0f22a6afc0e8514d6e65d6b0ca/apps/aeutils/priv/epoch_config_schema.json#L165).

RELEASE-NOTES.md
lucafavatella: Will add hint.

PR #550, opened 2018-01-03
Pt 154020800 aec sync mempool fetch

ThomasArts:
Small steps: split PT-153857451 into some smaller taks, of which this is the first.

Fixes two errors: 
1) shall not fetch mempool of chain with different genesis block
2) Don't match on found key... key may be different due to aliassing and normalization

Prepare for utf8 encoding of URIs, clearer types and propare for making peer() an opaque type. Basically not using it in API of external modules (such as tests).

### DIFF
--- a/apps/aecore/src/aec_peers.erl
+++ b/apps/aecore/src/aec_peers.erl
@@ -1,7 +1,7 @@
 %%%=============================================================================
 %%% @copyright 2017, Aeternity Anstalt
 %%% @doc
-%%%    Module storing peers list and providing funtionc for peers interaction
+%%%    Module storing peers list and providing funtions for peers interaction.
 %%% @end
 %%%=============================================================================
 -module(aec_peers).
@@ -40,6 +40,9 @@
 -endif.
 
 -include("peers.hrl").
+-define(MIN_PING_INTERVAL,   3000).
+-define(MAX_PING_INTERVAL, 120000).
+
 
 -type uri() :: http_uri:uri().
 -type get_peer_result() :: {ok, peer()} | {error, string()}.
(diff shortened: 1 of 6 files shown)


apps/aecore/src/aec_sync.erl
```diff
@@ -80,37 +80,35 @@ local_ping_object() ->
...
-                                  ok;
-                             true ->
-                                  start_sync(Src)
-                          end
-                  end,
-                  ok;
-              _ ->
-                  {error, different_genesis_blocks}
-          end,
-    fetch_mempool(Src),
```
velzevur: Nice catch

apps/aecore/src/aec_peers.erl
```diff
@@ -40,6 +40,9 @@
 -endif.
 
 -include("peers.hrl").
+-define(MIN_PING_INTERVAL,   3000).
+-define(MAX_PING_INTERVAL, 120000).
+
 
 -type uri() :: http_uri:uri().
```
uwiger: Why keep this type around when it's not being used anymore?

ThomasArts: It's the next refactoring. It's still in quite a few places in this code.

apps/aecore/src/aec_peers.erl
```diff
@@ -51,14 +54,14 @@
 %%------------------------------------------------------------------------------
 %% Add peer by url or supplying full peer() record. Try to connect.
 %%------------------------------------------------------------------------------
--spec add(uri() | peer()) -> ok.
+-spec add(http_uri:uri() | peer()) -> ok.
```
lucafavatella: If touching the spec-ing using this type, I understand this is not the way to go as this type is documented but not actually exported by `http_uri` module.

PR #553, opened 2018-01-03
PT-154027381 Difficulty reporting

lucafavatella:
This is a by-product of difficulty reporting-related investigation in environments.
This change does not introduce functional changes, though it is a small clarification.

Cc @dincho


lucafavatella:
Opening enhanced one.

PR #554, opened 2018-01-04
Wrap tester_keys_password in single quotes

dincho:
Otherwise it cannot contain spaces

https://www.pivotaltracker.com/story/show/154046229


dincho:
> For your information: the right word is "quote". There are single quotes ' and double quotes ".

Oh, It's a rushing mistake :) Thanks.

PR #556, opened 2018-01-04
Remove the notion of aliases from aec_peers

ThomasArts:
In the aec_peers modules there is a notion of aliases of a certain peer. This is a legacy notion to deal with the different names that localhost can have. We deal with that in configurations now and can remove the aliases for that part.

Aliases as implemented at the moment are unclear. We do trust any incoming URL even if it is not a valid URL to replace the URL we have used to approach the peer. This might be good in a trusted environment, it seems risky to just trust any peer's source response and start talking to that peer instead. Moreover, a lot of code deals with administering the hash of the different peer, even if that peer is basically the same, but has an extra path, the hash will change.

[finish PT-154024594]

### DIFF
--- a/apps/aecore/src/aec_peers.erl
+++ b/apps/aecore/src/aec_peers.erl
@@ -11,7 +11,6 @@
 %% API
 -export([add/1,
          add/2,
-         register_source/2,
          add_and_ping_peers/1,
          block_peer/1,
          unblock_peer/1,
@@ -20,7 +19,6 @@
          info/0,
          info/1,
          all/0,
-         aliases/0,
          get_random/0,
          get_random/1,
          get_random/2,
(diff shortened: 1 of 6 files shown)


uwiger:
Shouldn't "source" also be removed from the ping_object? (in aec_sync)

uwiger:
Currently, the "source" element in the ping_object is treated as mandatory ([here](https://github.com/aeternity/epoch/blob/master/apps/aecore/src/aec_sync.erl#L82)). If we're not actually using it for anything, at least there is no need for that.

ThomasArts:
We could remove the course field, because it seems ot to be used... but I only see the aecore parts, probably mobile apps or other browsers use it for logging or so?

dincho:
Removing the usage of ping source address would be great security improvement. I mentioned my concerns about it a couple of times already.

I didn't found time to write a PoC but I think the network could be attacked with "Smurf" like attack ATM.

ThomasArts:
The source is only used for incoming pings, but for outgoing pings, the response source is ignored and the original peer uri is used instead. Meaning, we cannot be fooled to block a different peer then we pinged.

PR #558, opened 2018-01-05
PT-154027381 Non-stale chain total difficulty metric

lucafavatella:
Fixed by disabling exometer cache for total difficulty metric - that is currently affected by a race condition and should have negligible impact on performance.

Finishes [PT-154027381](https://www.pivotaltracker.com/story/show/154027381).
Supersedes PR GH-553.

### DIFF
--- a/apps/aecore/src/aec_metrics.erl
+++ b/apps/aecore/src/aec_metrics.erl
@@ -54,6 +54,7 @@ default_dests() ->
 update(Metric, Value) ->
     exometer:update(Metric, Value).
 
+-spec try_update([atom(), ...], number()) -> ok | {error, not_found}.
 try_update(Metric, Value) ->
     try update(Metric, Value)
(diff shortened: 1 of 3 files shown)


uwiger:
Can you expand on what the race condition is?

uwiger:
Sorry, I see that you've done so: https://github.com/Feuerlabs/exometer_core/issues/105

apps/aecore/src/aec_metrics.erl
```diff
@@ -54,6 +54,7 @@ default_dests() ->
 update(Metric, Value) ->
     exometer:update(Metric, Value).
 
+-spec try_update([atom(), ...], number()) -> ok | {error, not_found}.
```
uwiger: I think using the exported type `exometer:name()` would be more appropriate.

PR #560, opened 2018-01-05
Ansible rolling deploys

dincho:
This introduce [Anisible rolling deploys](http://docs.ansible.com/ansible/latest/playbooks_delegation.html#rolling-update-batch-size) so that hosts are taken down for deploy in batch size of 25%.

This would guarantee much more stable network and as a side effect this makes it more similar to the production network where we don't control and deploy all nodes at once.

https://www.pivotaltracker.com/story/show/153589689

### DIFF
--- a/deployment/ansible/deploy.yml
+++ b/deployment/ansible/deploy.yml
@@ -24,6 +24,8 @@
   remote_user: epoch
   # Facts are already collected in the previous playbook
   gather_facts: no
+  max_fail_percentage: 25
+  serial: "25%"
 
   vars:
     project_root: "{{ ansible_env.HOME }}/node"
--- a/deployment/ansible/deploy.yml
+++ b/deployment/ansible/deploy.yml
@@ -33,6 +33,10 @@
 
     env: unknown
     config:
+      http:
+        external:
+          peer_address: http://{{ ansible_default_ipv4.address }}:3013/
+          port: 3013
       chain:
         persist: true
         db_path: "./db12"
@@ -141,3 +145,18 @@
     - name: Start epoch daemon
       command: "{{ project_root }}/bin/epoch start"
       tags: [daemon]
+
+    - name: Wait epoch node API to boot
+      wait_for:
+        port: "{{ config.http.external.port }}"
+        host: "{{ ansible_ssh_host }}"
+        timeout: 5
+      connection: local
+      tags: [health-check]
+
+    - name: API health check
+      uri:
+        url: http://{{ ansible_ssh_host }}:{{ config.http.external.port }}/v1/top
+        timeout: 5
+      connection: local
+      tags: [health-check]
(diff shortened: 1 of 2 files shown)


deployment/ansible/deploy.yml
```diff
@@ -24,6 +24,8 @@
   remote_user: epoch
   # Facts are already collected in the previous playbook
   gather_facts: no
+  max_fail_percentage: 25
+  serial: "25%"
```
lucafavatella: I see the end of this playbook is `bin/epoch start` that does not guarantee that node is ok. May I suggest ending playbook with more thorough check e.g. on a specific useful HTTP API being up and responding meaningfully (e.s. `/v1/top`)?

dincho: Here you go: 3e0acc1

deployment/ansible/deploy.yml
```diff
@@ -117,3 +145,18 @@
...
+        port: "{{ config.http.external.port }}"
+        host: "{{ ansible_ssh_host }}"
+        timeout: 5
+      connection: local
+      tags: [health-check]
+
+    - name: API health check
+      uri:
+        url: http://{{ ansible_ssh_host }}:{{ config.http.external.port }}/v1/top
+        timeout: 5
```
lucafavatella: I see this is the timeout for one attempt to connect to the server. I see no retries so I assume 1 (i.e. only one attempt). I recommend:
* Specify retries (e.g. 10);
* Specify time to wait among retries (e.g. 10s);
* Test on a box comparable to UAT with a copy of the DB of current testnet (just in case timing does not work on testnet deployment...).

----

On the code side, I see that [aehttp - the app responsible for starting HTTP listeners - waits on aecore](https://github.com/aeternity/epoch/blob/10a3bf23fe0c155da0ab7152b320c8e7244efced/apps/aehttp/src/aehttp.app.src#L15), and that aecore initialization requires the persisted chain to be read.

dincho: https://github.com/aeternity/epoch/pull/560/files#diff-2ad1947fe9b869ffd23a6a86872e65e0R158 guarantee that HTTP server is running (with timeout 5). It's internally polling so we don't need retries.

This just checks if API response code is 200 (no internal errors).

dincho: This should be sufficient for now 602c2c9

dincho:
> @dincho If you could review/amend/integrate https://github.com/aeternity/epoch/wiki/Release-process it
> would help - especially in case deployment fails for one of the "25%" groups.

@lucafavatella Is this good enough https://github.com/aeternity/epoch/wiki/Release-process/_compare/47318480fea9586aff7f9d759e7e32024d18e5b1...12717843090896b6a9b088bd125c8c71cf3727bb ?

lucafavatella:
> @lucafavatella Is this good enough https://github.com/aeternity/epoch/wiki/Release-process/_compare/47318480fea9586aff7f9d759e7e32024d18e5b1...12717843090896b6a9b088bd125c8c71cf3727bb ?
Yest

> @lucafavatella Is this good enough https://github.com/aeternity/epoch/wiki/Release-process/_compare/47318480fea9586aff7f9d759e7e32024d18e5b1...12717843090896b6a9b088bd125c8c71cf3727bb ?

Yes. Thanks.

PR #563, opened 2018-01-08
Bug fixes and improvements to aec_syn and aec_peers

ThomasArts:
This PR fixes a few bugs found by QuickCheck (see PT-153857451). It also simplifies the code to make it easier to maintain and reduces the dependencies.

- Make the peers utf8 compliant by normalizing them as binaries
- remove unused function
- Change the type of get_random
- State of aec_sync is unused.
- Completely isolate the peer() type inside aec_peers
- Remove unused code
- No need to filter blocked peers, they are not there
- Fixed: configure same peer twice only results in one ping

[Finished PT-153857451]


apps/aecore/src/aec_peers.erl
```diff
@@ -237,164 +235,146 @@ check_ping_interval_env() ->
...
--record(state, {peers :: gb_trees:tree(binary(),peer()),
-                blocked = gb_sets:new() :: gb_sets:set(uri()),
-                errored = gb_sets:new() :: gb_sets:set(uri()),
-                local_peer_uri :: uri(),
-                local_peer_host :: string() | undefined,
-                local_peer_port :: integer() | undefined}).
+-record(state, {peers :: gb_trees:tree(binary(), peer()),
+                blocked = gb_sets:new() :: gb_sets:set(http_uri:uri()),
+                errored = gb_sets:new() :: gb_sets:set(http_uri:uri()),
+                local_peer :: peer()}).
```
uwiger: Ok, perhaps at least a comment about which attributes of the #peer{} record that are relevant for `local_peer`? I believe it's just the uri-related parts...

apps/aecore/src/aec_peers.erl
```diff
@@ -237,164 +235,146 @@ check_ping_interval_env() ->
...
-                    {reply, {ok, Peer}, State};
-                false ->
-                    Iterator = gb_trees:iterator_from(list_to_binary(RandomHash), State#state.peers),
-                    {_, Peer, _} = gb_trees:next(Iterator),
-                    {reply, {ok, Peer}, State}
-            end
-    end.
+handle_call({get_random, N0, Exclude}, _From, #state{peers = Tree0,
+                                                     errored = Errored} = State) ->
+    %% first, remove all blocked peers
```
uwiger: "remove all **errored** peers", right? Why not the blocked too?
Ah, because blocked peers are removed from the `peers` set when blocked.

apps/aecore/src/aec_peers.erl
```diff
@@ -434,109 +416,25 @@ metrics(#state{peers = Peers, blocked = Blocked,
...
+-spec log_good_ping(peer()) -> ok.
+log_good_ping(Peer) ->
+    gen_server:cast(?MODULE, {good_ping, Peer, timestamp()}).
 
--spec hash_uri(uri()) -> binary().
-hash_uri(Uri) when is_list(Uri)->
-    hash_uri(list_to_binary(Uri));
+-spec hash_uri(http_uri:uri()) -> binary().
 hash_uri(Uri) ->
     Hash = crypto:hash(md4,Uri),  %TODO add some random for execution but constant salt, so people can't mess with uri-s to hide on our list.
```
uwiger: Is this comment relevant? (I know it's not yours)

apps/aecore/src/aec_peers.erl
ThomasArts: Yes, that's right, at the moment we use scheme, host and port. But in the future possibly also the path. The gain on using this type is that we can re-use the universal encoding and decoding of URIs.

apps/aecore/src/aec_peers.erl
ThomasArts: Well spotted, the comment is wrong. The invariant that we guarantee is that Blocked peers is disjunct to peers. Not so for errored peers, because we need to remember when we latest communicated with them to be able to try again in due time. This is not stored in the URI.

Note, however, that keeping them in peers is expensive. If you have 10,000 errored peers it takes a long time each round you want to find a few none errored once.

apps/aecore/src/aec_peers.erl
ThomasArts: Not my comment. I don't understand how one could potentially do any harm with the hash anyway. I left it because I really did not understand it...

uwiger: Me too. :)

PR #564, opened 2018-01-08
Add configuration example

dincho:
I followed the [directory structure guidelines](http://erlang.org/doc/design_principles/applications.html#id81839) to choose a path

After this PR is merged we should change https://github.com/aeternity/epoch/wiki/User-provided-configuration to point to the example config.

https://www.pivotaltracker.com/story/show/153555276

### DIFF
--- a/apps/aeutils/doc/examples/epoch.yaml
+++ b/apps/aeutils/doc/examples/epoch.yaml
@@ -1,0 +1,72 @@
+---
+# Pre-configured addresses of epoch nodes to contact
+peers:
+    # TestNet Addresses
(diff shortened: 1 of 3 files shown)


lucafavatella:
I also tested locally that if I break the new unit test, `make eunit` fails.

dincho:
@lucafavatella I [updated](https://github.com/aeternity/epoch/wiki/User-provided-configuration/_compare/b0654e0691c3fcd0c9e0b862d1fa2569242f16a8...523d9ea1e9bc94b87a6bc6b43146ad3364367dad) the configuration wiki page. Let me know if you had something different/better in mind.

PR #566, opened 2018-01-09
base58Check support, first commit

uwiger:
See
* https://www.pivotaltracker.com/story/show/153897408
* https://www.pivotaltracker.com/story/show/153767903

First commit, only Erlang code addressed. Outstanding questions:

* exactly which prefixes to use? Specifically, different `block_hash` prefixes? A special `genesis_hash` prefix?
* Whether to add the prefix _before_ base58Check encoding or after? An unencoded prefix is easier to read, but less secure, since it's not included in the validation check.


happi:
I think perhaps the block version should be bumped.
Possibly we could also be backwards compatible (handling base 64 on the network for blocks of previous version).

lucafavatella:
> I think perhaps the block version should be bumped.

@happi Maybe you meant the [HTTP API version](https://github.com/aeternity/epoch/blob/52b1475d1723fefc67b2774d0ef4deca56d62227/config/swagger.yaml#L10)?

I did not find chain format changes.

apps/aecore/src/aec_base58c.erl
```diff
@@ -0,0 +1,73 @@
...
+                                              crypto:hash(sha256, Bin)),
+    C.
+
+
+type2pfx(block_hash)     -> <<"bh">>;
+type2pfx(block_tx_hash)  -> <<"bx">>;
+type2pfx(transaction)    -> <<"tx">>;
+type2pfx(oracle_pubkey)  -> <<"ok">>;
+type2pfx(account_pubkey) -> <<"ak">>;
+type2pfx(Other)          -> Other.
```
lucafavatella: I expect integration tests run all APIs at least once. I would suggest deleting this open-ended clause.

apps/aecore/src/aec_base58c.erl
```diff
@@ -0,0 +1,73 @@
...
+
+pfx2type(<<"bh">>) -> block_hash;
+pfx2type(<<"bx">>) -> block_tx_hash;
+pfx2type(<<"tx">>) -> transaction;
+pfx2type(<<"ok">>) -> oracle_pubkey;
+pfx2type(<<"ak">>) -> account_pubkey;
+pfx2type(Other)    -> Other.
```
lucafavatella: I expect integration tests run all APIs at least once. I would suggest deleting this open-ended clause.

apps/aecore/src/aec_blocks.erl
```diff
@@ -144,9 +144,9 @@ serialize_for_network(B = #block{}) ->
 
 serialize_to_map(B = #block{}) ->
     #{<<"height">> => height(B),
-      <<"prev_hash">> => base64:encode(prev_hash(B)),
-      <<"state_hash">> => base64:encode(B#block.root_hash),
-      <<"txs_hash">> => base64:encode(B#block.txs_hash),
+      <<"prev_hash">> => aec_base58c:encode(block_hash, prev_hash(B)),
+      <<"state_hash">> => aec_base58c:encode(block_hash, B#block.root_hash),
```
lucafavatella: Distinct type e.g. block_state_hash? This is hash of the state (i.e. summary of all state trees in a block), not hash of block header.

lucafavatella: Distinct type e.g. block_state_hash? This is hash of the state (i.e. summary of all state trees in a block), not hash of block header.

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -71,18 +76,23 @@ handle_request('PostOracleQueryTx', #{'OracleQueryTx' := OracleQueryTxObj}, _Con
...
-                    nonce        => Nonce,
-                    oracle       => base64:decode(EncodedOraclePubkey),
-                    query        => Query,
-                    query_fee    => QueryFee,
-                    query_ttl    => {QueryTTLType, QueryTTLValue},
-                    response_ttl => {delta, ResponseTTLValue},
-                    fee          => Fee}),
-            sign_and_push_to_mempool(OracleQueryTx),
-            {200, [], #{}};
+            case safe_decode(account_pubkey, EncodedOraclePubkey) of
```
lucafavatella: oracle_pubkey?

apps/aecore/src/aec_sync.erl
```diff
@@ -84,29 +84,62 @@ compare_ping_objects(RemoteUri, Local, Remote) ->
         {G, G} ->
```
lucafavatella: I understand that the base64/base58 encoding is not part of the business logic, it is rather a type that should (ideally) be handled inside swagger.

May I suggest that the base64/base58 encoding/decoding is moved in `aehttp_dispatch_ext:handle_request('Ping', ...)` (so away from aec_sync), and that HTTP status code 400 (with descriptive error reason) is returned. In this was two genesis objects where (for any reason) one is base64 and the other base58 (or any other encoding/decoding quirk) is detected as bad client request format - rather than `different_genesis_block`.

ThomasArts: That is not exactly the right place to handle it. There are server and client side. What we should do is to handle with pure data in aec_syn and peer and only encode and decode in http request/response. The request and response for the client side are in aeu_http_client. Incomping pings do get to aehttp_dispatch_ext... but can be filtered before.

apps/aecore/src/aec_sync.erl
```diff
@@ -84,29 +84,62 @@ compare_ping_objects(RemoteUri, Local, Remote) ->
...
-             ok;
+                            {ok, start_sync}
+                    end;
+                false ->
+                    {error, invalid_block_hash}
+            end
+    end.
+
+
+validate_remote_hash(Hash) ->
```
thepiwo: wouldn't the validation make more sense in the `aec_base58c` module?

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -267,7 +274,12 @@ abort_sync(Uri, Code, Reason) ->
                        reason => Reason}}),
       {Code, [], #{reason => Reason}}.
 
-base64_safe_decode(K) when is_binary(K) ->
-      try {ok, base64:decode(K)}
-      catch _:_ -> {error, not_base64_encoded}
-      end.
+safe_decode(Type, Hash) ->
```
thepiwo: shouldn't this be in `aec_base58c` as well?

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -169,3 +179,14 @@ sign_and_push_to_mempool(Tx) ->
     ok = aec_tx_pool:push(SignedTx),
     lager:debug("pushed; peek() -> ~p",
                 [pp(aec_tx_pool:peek(10))]).
+
+safe_decode(Type, Enc) ->
```
thepiwo: when this is in `aec_base58c` we could save duplicate code from `apps/aehttp/src/aehttp_dispatch_ext.erl`

thepiwo:
@lucafavatella 

> You may mention that by backing up the key pair the account balance should be preserved.

If the chain is restarted, backing up the keys would not preserve the balance, as it is no longer in the chain. When hard-forking without resetting the chain, new clients would not be able to sync from genesis to the fork-block, as the validation changed.

lucafavatella:
@thepiwo 

> If the chain is restarted, backing up the keys would not preserve the balance, as it is no longer in the chain. When hard-forking without resetting the chain, new clients would not be able to sync from genesis to the fork-block, as the validation changed.

Correct, though I understand no chain fork aimed in this PR. cc @uwiger

uwiger:
Code merged (in modified form) as part of PR #576 - closing

PR #570, opened 2018-01-10
Template for GitHub issues [Finishes#154206112]

evdoks:


.github/issue_template.md
```diff
@@ -0,0 +1,33 @@
...
+## Logs, error output, etc.
+ 
+ (`epoch.yaml` configuration file, `log/epoch.log` log file plus other related logs, e.g., `log/epoch_mining.log`; if it’s long, please paste to https://ghostbin.com/ and insert the link here)
+ 
+ ...
+
+
+## Specifications
+
+  - Version:
```
dincho: It's not pretty clear to me what Version/Platform/Subsystem is.

dincho:
Could you also make issue_template.md upper case and move it to root for consistency.

Eventually I propose to move all the documents (not only GH related) to `docs` (as GH supports that too). However this would also need change for https://github.com/aeternity/epoch/blob/master/rebar.config#L39

.github/issue_template.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Specifications
+
+
+  - Virtualization: [*none / metal / container / VM*]
+  - Hardware specs: 
+    * RAM: 
+    * CPU:
+  - OS: [*distribution, version*]
+  - Node Version:
```
sennui: we need to know how the OTP was installed in the system (brew/source/packages) and which OTP is it

Also we need to know how `epoch` was installed (source/packages)

dincho: @sennui I think we should "support" only the official release deployments, our master is not stable enough to support deploys installed from it, is it?

If we agree to support only package installs from official release, then we ship with OTP already. The only external dependancy AFAIK is libssl.

PR #572, opened 2018-01-10
Restructure Oracle state trees

hanssv:
The idea is to have the oracle interaction objects as a sub-tree under
the oracle itself. Otherwise it will be (almost) impossible to list all
interactions that belong to a particular oracle.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,6 +1,7 @@
 # About this release
 
 [This release](https://github.com/aeternity/epoch/releases/tag/v0.4.1) is focused on stability of the testnet - the public test network of nodes and introduces some API for apps usability build on top of it.
+It also introduces a backward incompatible change in the chain format.
 
 Please follow the instructions below and let us know if you have any problems by [opening a ticket](https://github.com/aeternity/epoch/issues).
 
@@ -114,6 +115,7 @@ Ensure the configured path for storing the blockchain exists:
 ```
 mkdir /tmp/node/my_db
 ```
+As of release "v0.5.0", as the chain format changed from the previous release, please ensure that you do not reuse a persisted blockchain produced by the previous releases "v0.4.x".
 
 You can validate the configuration file before starting the node:
 ```
--- a/apps/aeoracle/src/aeo_query_tx.erl
+++ b/apps/aeoracle/src/aeo_query_tx.erl
@@ -99,15 +99,14 @@ origin(#oracle_query_tx{sender = SenderPubKey}) ->
 %% SenderAccount should exist, and have enough funds for the fee + the query_fee.
 %% Oracle should exist, and query_fee should be enough
 %% Fee should cover TTL
-%% TODO: Check the query? Check TTL relative to Oracle TTL?
 -spec check(query_tx(), trees(), height()) -> {ok, trees()} | {error, term()}.
 check(#oracle_query_tx{sender = SenderPubKey, nonce = Nonce,
                        oracle = OraclePubKey, query_fee = QFee,
-                       query_ttl = TTL, fee = Fee} = Q, Trees, Height) ->
+                       query_ttl = TTL, response_ttl = RTTL, fee = Fee} = Q, Trees, Height) ->
     Checks =
         [fun() -> aetx_utils:check_account(SenderPubKey, Trees, Height, Nonce, Fee + QFee) end,
          fun() -> aeo_utils:check_ttl_fee(Height, TTL, Fee - ?ORACLE_QUERY_TX_FEE) end,
-         fun() -> check_oracle(OraclePubKey, Trees, QFee) end,
+         fun() -> check_oracle(OraclePubKey, Trees, QFee, Height, TTL, RTTL) end,
          fun() -> check_interaction(Q, Trees, Height) end
         ],
 
@@ -215,20 +214,34 @@ for_client(#oracle_query_tx{sender        = SenderPubKey,
 %% -- Local functions  -------------------------------------------------------
 
 check_interaction(Q, Trees, Height) ->
-    Oracles = aec_trees:oracles(Trees),
-    Id      = aeo_interaction:id(aeo_interaction:new(Q, Height)),
-    case aeo_state_tree:lookup_interaction(Id, Oracles) of
+    Oracles  = aec_trees:oracles(Trees),
+    I        = aeo_interaction:new(Q, Height),
+    OracleId = aeo_interaction:oracle_address(I),
+    Id       = aeo_interaction:id(I),
+    case aeo_state_tree:lookup_interaction(OracleId, Id, Oracles) of
         none       -> ok;
         {value, _} -> {error, oracle_interaction_already_present}
     end.
 
-check_oracle(OraclePubKey, Trees, QueryFee) ->
+check_oracle(OraclePubKey, Trees, QueryFee, Height, QTTL, RTTL) ->
     OraclesTree  = aec_trees:oracles(Trees),
     case aeo_state_tree:lookup_oracle(OraclePubKey, OraclesTree) of
         {value, Oracle} ->
             case QueryFee >= aeo_oracles:query_fee(Oracle) of
-                true  -> ok;
+                true  -> check_oracle_ttl(Oracle, Height, QTTL, RTTL);
                 false -> {error, query_fee_too_low}
             end;
         none -> {error, oracle_does_not_exist}
+    end.
+
+check_oracle_ttl(O, Height, QTTL, RTTL) ->
+    try
+        Delta  = aeo_utils:ttl_delta(Height, QTTL),
+        MaxTTL = aeo_utils:ttl_expiry(Height + Delta, RTTL),
+        case aeo_oracles:expires(O) < MaxTTL of
+            false -> ok;
+            true  -> {error, too_long_ttl}
+        end
+    catch _:_ ->
+        {error, invalid_ttl}
     end.
--- a/apps/aeoracle/src/aeo_state_tree.erl
+++ b/apps/aeoracle/src/aeo_state_tree.erl
@@ -8,13 +8,13 @@
 -module(aeo_state_tree).
 
 %% API
--export([ get_interaction/2
+-export([ get_interaction/3
         , get_oracle/2
         , empty/0
         , enter_interaction/2
         , insert_interaction/2
         , insert_oracle/2
-        , lookup_interaction/2
+        , lookup_interaction/3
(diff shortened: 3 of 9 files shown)


hanssv:
This change is chain altering, probably should bump the db...

hanssv:
Will bump version and chain DB and squash the patches before this patch can be merged...

hanssv:
Squashed commits and bumped PROTOCOL_VERSION, GENESIS_VERSION and deployment DB path.

No functional change from previous commits.

RELEASE-NOTES.md
```diff
@@ -114,6 +115,7 @@ Ensure the configured path for storing the blockchain exists:
 ```
 mkdir /tmp/node/my_db
 ```
+As of release "v0.5.0-<placeholder>", as the chain format changed from the previous release, please ensure that you do not reuse a persisted blockchain produced by the previous releases "v0.4.x".
```
lucafavatella: I see https://www.pivotaltracker.com/story/show/153989778 has in title `v0.5.0`, I think you can use that rather than "v0.5.0-<placeholder>".

apps/aeoracle/src/aeo_state_tree.erl
```diff
@@ -30,32 +30,45 @@
...
+-type itrees() :: gb_tree:tree(aeo_oracles:id(), itree()).
+-type cache_item() :: {oracle, aeo_oracles:id()}
+                    | {interaction, aeo_oracles:id(), aeo_interaction:id()}.
+-type cache() :: gb_sets:set({integer(), cache_item()}).
 -type block_height() :: non_neg_integer().
 
--record(oracle_tree, { mtree = gb_merkle_trees:empty() :: mtree()
-                     , cache = cache_new() :: cache()
+-record(oracle_tree, { otree  = aeu_mtrees:empty() :: otree()
+                     , itrees = gb_trees:empty()   :: itrees()
```
lucafavatella: Not a merkle tree? I am trying to understand how it works.

apps/aeoracle/src/aeo_state_tree.erl
```diff
@@ -30,32 +30,45 @@
...
+%% the root hash of the interaction tree.
+%%
+%% We also keep a cache of objects to expire.
+
+-type otree() :: aeu_mtrees:tree(aeo_oracles:id(), aeo_oracles:serialized()).
+-type itree() :: aeu_mtrees:tree(aeo_interaction:id(), aeo_interaction:serialized()).
 -type interaction() :: aeo_interaction:interaction().
 -type oracle() :: aeo_oracles:oracle().
--type cache() :: gb_sets:set({integer(), binary()}).
+-type itrees() :: gb_tree:tree(aeo_oracles:id(), itree()).
```
lucafavatella: btw it is gb_trees (with s)

apps/aeoracle/src/aeo_state_tree.erl
lucafavatella: I see - so each oracle has an oracle interaction tree, though the oracle internal representation does not contain the oracle interaction tree (as I was expecting) instead it contains its root hash. Then the oracle interactions trees - one per oracle - in the internal representation are stored completely separately, and the consistency between the two is kept with additional code.

I admit this is not very clear from the comment https://github.com/aeternity/epoch/pull/572/files/7d61e16e9d72e5dd8e6a9ddeff10b940c748c1c8#diff-e72d14988d10e82668a1a08726718bb8R33

apps/aeoracle/src/aeo_query_tx.erl
```diff
@@ -215,20 +215,34 @@ for_client(#oracle_query_tx{sender        = SenderPubKey,
...
         {value, Oracle} ->
             case QueryFee >= aeo_oracles:query_fee(Oracle) of
-                true  -> ok;
+                true  -> check_oracle_ttl(Oracle, Height, QTTL, RTTL);
                 false -> {error, query_fee_too_low}
             end;
         none -> {error, oracle_does_not_exist}
     end.
+
+check_oracle_ttl(O, Height, QTTL, RTTL) ->
```
lucafavatella: In type aeo_oracles:fixed_ttl(), could you please replace integer() with height() (from common.hrl) or non_neg_integer()? (for readability)

In type aeo_oracles:relative_ttl(), could you please replace integer() with whether it is meant to be pos - neg - nonpos- nonneg? (for readability)

apps/aeoracle/src/aeo_query_tx.erl
```diff
@@ -103,11 +103,11 @@ origin(#oracle_query_tx{sender = SenderPubKey}) ->
 -spec check(query_tx(), trees(), height()) -> {ok, trees()} | {error, term()}.
```
lucafavatella: Can comment at line 102 "%% TODO: Check the query? Check TTL relative to Oracle TTL?" be deleted?

apps/aeoracle/src/aeo_query_tx.erl
```diff
@@ -215,20 +215,34 @@ for_client(#oracle_query_tx{sender        = SenderPubKey,
...
             end;
         none -> {error, oracle_does_not_exist}
     end.
+
+check_oracle_ttl(O, Height, QTTL, RTTL) ->
+    try
+        Delta  = aeo_utils:ttl_delta(Height, QTTL),
+        MaxTTL = aeo_utils:ttl_expiry(Height + Delta, RTTL),
+        case aeo_oracles:expires(O) < MaxTTL of
+            false -> ok;
```
lucafavatella: This checks that oracle interaction expires after (or at same time of) oracle. Ok.

apps/aeoracle/src/aeo_state_tree.erl
```diff
@@ -67,103 +80,131 @@ prune(Height, #oracle_tree{} = Tree) ->
...
+    int_prune(cache_safe_peek(Cache1), Height, Tree1);
+int_prune({Height1,_Id}, Height2, Tree) when Height2 < Height1 ->
+    Tree.
+
+delete({oracle, Id}, Tree) ->
+    OTree  = aeu_mtrees:delete(Id, Tree#oracle_tree.otree),
+    ITrees = gb_trees:delete(Id, Tree#oracle_tree.itrees),
+    Tree#oracle_tree{ otree = OTree, itrees = ITrees };
+delete({interaction, OracleId, Id}, Tree) ->
+    %% It is possible that the oracle expired first.
```
lucafavatella: I understand this means "It is possible that the oracle expiration was processed already, hence having deleted the related interactions." as the oracle interaction is meant to expire after (or at same time of) oracle.

apps/aeoracle/src/aeo_state_tree.erl
```diff
@@ -67,103 +80,131 @@ prune(Height, #oracle_tree{} = Tree) ->
...
+    %% Since {X, {oracle, Hash}} sorts before {X, {interaction, H1, H2}} the
+    %% inverse cannot happen.
+    case gb_trees:lookup(OracleId, Tree#oracle_tree.itrees) of
+        {value, ITree} ->
+            ITree1 = aeu_mtrees:delete(Id, ITree),
```
lucafavatella: Shall any refunding take place? See https://github.com/aeternity/protocol/blob/f578d480f8c8e13c1425f26ebb732db01d099ab3/oracles/oracle_state_tree.md#pruning-of-oracle-interaction-objects

apps/aeoracle/src/aeo_state_tree.erl
```diff
@@ -67,103 +80,131 @@ prune(Height, #oracle_tree{} = Tree) ->
...
-    {Cache, MTree}.
-
+    Tree1 = delete(Id, Tree#oracle_tree{ cache = Cache1 }),
+    int_prune(cache_safe_peek(Cache1), Height, Tree1);
+int_prune({Height1,_Id}, Height2, Tree) when Height2 < Height1 ->
+    Tree.
+
+delete({oracle, Id}, Tree) ->
+    OTree  = aeu_mtrees:delete(Id, Tree#oracle_tree.otree),
+    ITrees = gb_trees:delete(Id, Tree#oracle_tree.itrees),
```
lucafavatella: Shall any refunding take place? See https://github.com/aeternity/protocol/blob/f578d480f8c8e13c1425f26ebb732db01d099ab3/oracles/oracle_state_tree.md#pruning-of-oracle-interaction-objects

apps/aeoracle/src/aeo_state_tree.erl
hanssv: One could imagine that Dialyzer can warn about that... Good catch.

apps/aeoracle/src/aeo_query_tx.erl
hanssv: Yes, you are right, this was something we skipped for the initial release but I fixed it here!

apps/aeoracle/src/aeo_query_tx.erl
hanssv: No, it checks that the oracle does not expire before the the oracle interaction. We do not want to risk orphaned oracle interactions.

apps/aeoracle/src/aeo_state_tree.erl
hanssv: Yes, eventually. This was one of the corners cut to make the first release. I'll give it a quick thought and see if it is easy enough to do here... But I think not, we only have access to the oracle state tree, not the accounts tree.

apps/aeoracle/src/aeo_state_tree.erl
hanssv: Exactly. In the case that the interaction expire at the same block height as the oracle itself we will prune the oracle (and its interactions) first (entries in the cache are sorted like that).

QuickCheck was kind enough to alert me of this corner case...

apps/aeoracle/src/aeo_state_tree.erl
hanssv: I've rephrased the comment - and clarified that the reason for having the interactions in a separate tree is to avoid lots of serialize/deserialize on insert/update.

apps/aeoracle/src/aeo_state_tree.erl
hanssv: See below.

hanssv:
Updated with some minor fixes and improvements according to Luca's comments.

apps/aeoracle/src/aeo_state_tree.erl
lucafavatella: I see https://www.pivotaltracker.com/story/show/154292757 - thanks.

PR #574, opened 2018-01-11
HTTP Endpoint: getTransactionFromBlock

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/154168041)
**getTransactionFromBlock** ([ref](http://web3js.readthedocs.io/en/1.0/web3-eth.html#gettransactionfromblock)) 
**Description**: Returns a transaction based on a block hash or number and the transactions index position.
**Endpoints**:
GET /block/tx/height/{height}/{tx_index}
GET /block/tx/hash/{hash}/{tx_index}
GET /block/tx/latest/{tx_index}
**Options**:
*tx_objects* - (optional, default false) If true, the returned block will contain all transactions as objects, if false it will only contains the transaction message pack encoded hashes

### DIFF
--- a/apps/aehttp/src/aehttp_dispatch_int.erl
+++ b/apps/aehttp/src/aehttp_dispatch_int.erl
@@ -155,14 +155,10 @@ handle_request('GetBlockByHeightInternal', Req, _Context) ->
     get_block(fun() -> aec_conductor:get_block_by_height(Height) end, Req);
 
 handle_request('GetBlockByHashInternal', Req, _Context) ->
-    Hash =
-        try base64:decode(maps:get('hash', Req))
-              catch _:_ -> not_base64_encoded
-              end,
-    case Hash of
-        not_base64_encoded ->
+    case aec_base58c:safe_decode(block_hash, maps:get('hash', Req)) of
+        {error, _} ->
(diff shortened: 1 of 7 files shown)


velzevur:
I didn't implement the following:
GET /block/tx/genesis/{tx_index}
GET /block/tx/pending/{tx_index}
as those make no sense (genesis block doesn't have any transactions and they order isn't fixed in the block being currently mined)

sennui:
Referring to your API work in general: For transactions like get tx number in block - maybe we should cache it somewhere in block structure?

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -278,6 +299,36 @@ get_block_txs_count(Fun) when is_function(Fun, 0) ->
             Err
     end.
 
+get_block_tx_by_index(Fun, Index, Req) when is_function(Fun, 0) ->
+    case get_block_from_chain(Fun) of
+        {ok, Block} ->
+            Txs = aec_blocks:txs(Block),
+            case length(Txs) < Index orelse Index =< 0 of
```
uwiger: A `try lists:nth(Index, Txs) of ... catch` could work here. It would eliminate the O(n) `length/1` check.

PR #575, opened 2018-01-11
[WIP] PT-154211489 Fix timeout on node start on reading chain from disk

lucafavatella:
Finishes [PT-154211489](https://www.pivotaltracker.com/story/show/154211489).

----

Tested this on an environment where I was constantly getting the error. With this patch node started at first time (though of course it is not full proof).


lucafavatella:
Preferred PR #577 with exact same content though based on tag for easing maintenance. Closing.

PR #579, opened 2018-01-12
[0.4.1.1] PT-154279032 Bump release version

lucafavatella:


dincho:
We should also update master release notes

dincho:
> We should also update master release notes

Re-thinking, it might not need to.

PR #583, opened 2018-01-15
Add signature type and use base58 encoding instead of base64

ThomasArts:
After introducing base58 some new code still contained base64 and was merged. This is now fixed.

The base64 encoding in oracle part is left. It is not used and fixed in the upcoming pull request on oracles already.

The base64 encoding in cuckoo is also left there, since this is only used internally.


dincho:
Full path: `deployment/ansible/group_vars/integration/accounts.yml`

lucafavatella:
@ThomasArts @uwiger This PR shall finish PT-153897408 and PT-153767903, right?

PR #584, opened 2018-01-15
Failing test fix block_txs_count_pending

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/154206801)
Since block candidate is a rapidly changing structure, this produces a lot of race conditions for the integration test to handle.
This PR lowers the expectations of the integration test and moves them to a better suited unit test.

### DIFF
--- a/apps/aecore/test/aec_conductor_tests.erl
+++ b/apps/aecore/test/aec_conductor_tests.erl
@@ -381,6 +381,74 @@ test_block_publishing() ->
     ok.
 
 %%%===================================================================
+%%% Pending block tests
+%%%===================================================================
+
+block_candidate_test_() ->
+    {foreach,
+     fun() ->
+          TmpKeysDir = setup_common(),
+          {ok, _} = ?TEST_MODULE:start_link([{autostart, false}]),
+          meck:new(aec_mining, [passthrough]),
+          meck:expect(aec_mining, create_block_candidate,
+              fun(TopBlock, TopBlockState, AdjChain) ->
+                  timer:sleep(100),
+                  meck:passthrough([TopBlock, TopBlockState, AdjChain])
+              end),
+          meck:expect(aec_mining, mine,
+              fun(_, _) ->
+                  timer:sleep(3000),
+                  {error, no_solution}
+              end),
+          TmpKeysDir
+     end,
+     fun(TmpKeysDir) ->
+          meck:unload(aec_mining),
+          teardown_common(TmpKeysDir),
+          ok
+     end,
+     [
+      {"Get block candidate", fun test_get_block_candidate/0}
+     ]}.
+
+test_get_block_candidate() ->
+    assert_stopped_and_genesis_at_top(),
+    ?assertEqual({error, not_mining}, ?TEST_MODULE:get_block_candidate()),
+    {ok, MyAccount} = aec_keys:pubkey(),
+    lists:foreach(
+        fun(_) ->
+            Tx = #spend_tx{sender = MyAccount,
+                          recipient = MyAccount,
+                          nonce = 0, fee = 0},
+            {ok, STx} = aec_keys:sign(Tx),
+            ok = aec_tx_pool:push(STx, tx_received)
+        end,
+        lists:seq(1, 3)),
+    ?TEST_MODULE:start_mining(),
+    ?assertEqual({error, miner_starting}, ?TEST_MODULE:get_block_candidate()),
+    true = aec_test_utils:wait_for_it(
+              fun() -> {error, miner_starting} =/= ?TEST_MODULE:get_block_candidate() end,
+              true),
+    TopBlock = ?TEST_MODULE:top(),
+    {ok, BlockCandidate} = ?TEST_MODULE:get_block_candidate(),
+    {ok, TopBlockHash} = aec_blocks:hash_internal_representation(TopBlock),
+    ?assertEqual(TopBlockHash, aec_blocks:prev_hash(BlockCandidate)),
+    {ok, AllTxsInPool} = aec_tx_pool:peek(infinity),
+    ?assertEqual(true,
+        lists:all(
+            fun(SignedTx) ->
+                Tx = aec_tx_sign:data(SignedTx),
+                case aec_tx_dispatcher:handler(Tx) of
+                    aec_coinbase_tx -> true;
+                    _ ->
+                        lists:member(Tx, AllTxsInPool)
+                end
+            end,
+            aec_blocks:txs(BlockCandidate))),
+    ?TEST_MODULE:stop_mining(),
+    ?assertEqual({error, not_mining}, ?TEST_MODULE:get_block_candidate()),
+    ok.
+%%%===================================================================
 %%% Helpers
 %%%===================================================================
 
--- a/apps/aehttp/test/aehttp_integration_SUITE.erl
+++ b/apps/aehttp/test/aehttp_integration_SUITE.erl
@@ -827,20 +827,40 @@ internal_block_pending(_Config) ->
                 <<"BlockWithTxsHashes">>,
                 block_to_endpoint_map(PendingBlock)),
     ct:log("Expected pending block ~p", [ExpectedPendingTx]),
-    {ok, 200, PendingTxDefault} = get_internal_block_preset("pending",
-                                                             default),
-    {ok, 200, PendingTxHashes} = get_internal_block_preset("pending",
-                                                             false),
-    true = equal_block_maps(ExpectedPendingTx, PendingTxDefault), 
-    true = equal_block_maps(ExpectedPendingTx, PendingTxHashes), 
+    GetPending =
+        fun TryToGet(_Opt, Attempts) when Attempts < 1 ->
+                {error, no_pending_block_returned};
+            TryToGet(Opt, Attempts) ->
+                case get_internal_block_preset("pending", Opt) of
+                    {ok, 200, P} -> {ok, P};
+                    {ok, 404, _} -> TryToGet(Opt, Attempts -1)
+                end
+        end,
+    {ok, PendingTxDefault} = GetPending(default, 10),
+    {ok, PendingTxHashes} = GetPending(false, 10),
+    ValidateKeys =
+        fun(Map1, Map2, Key) ->
+            true = maps:get(Key, Map1, not_found1) =:=
+                   maps:get(Key, Map2, not_found2)
+        end,
+    % no block should have been mined, so the same prev_hash
+    ValidateKeys(ExpectedPendingTx, PendingTxDefault, <<"prev_hash">>),
+    ValidateKeys(ExpectedPendingTx, PendingTxDefault, <<"data_schema">>),
+
+    % no block should have been mined, so the same prev_hash
+    ValidateKeys(ExpectedPendingTx, PendingTxHashes, <<"prev_hash">>),
+    ValidateKeys(ExpectedPendingTx, PendingTxHashes, <<"data_schema">>),
+
     ExpectedPendingTxsObjects = maps:put(<<"data_schema">>,
                 <<"BlockWithTxs">>,
                 block_to_endpoint_map(PendingBlock, #{tx_objects => true})),
     ct:log("Expected pending block with tx objects~p",
            [ExpectedPendingTxsObjects]),
-    {ok, 200, PendingTxsObjects} = get_internal_block_preset("pending",
-                                                             true),
-    true = equal_block_maps(ExpectedPendingTxsObjects, PendingTxsObjects),
+    {ok, PendingTxObjects} = GetPending(true, 10),
+    
+    % no block should have been mined, so the same prev_hash
+    ValidateKeys(ExpectedPendingTxsObjects, PendingTxObjects, <<"prev_hash">>),
+    ValidateKeys(ExpectedPendingTxsObjects, PendingTxObjects, <<"data_schema">>),
     ok.
 
 internal_get_block_generic(GetExpectedBlockFun, CallApiFun) ->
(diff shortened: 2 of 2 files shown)


apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -906,10 +922,19 @@ block_txs_count_pending(_Config) ->
     rpc(aec_conductor, start_mining, []),
```
lucafavatella: Why 2 start_mining?

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -906,10 +922,19 @@ block_txs_count_pending(_Config) ->
...
     timer:sleep(100),% so the miner is started
```
lucafavatella: Can this sleep be removed now that there are attempts?

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -906,10 +922,19 @@ block_txs_count_pending(_Config) ->
...
+                {error, no_pending_block_returned};
+            TryToGet(Attempts) ->
+                case get_block_txs_count_preset("pending") of
+                    {ok, 200, P} -> {ok, P};
+                    {ok, 404, _} -> TryToGet(Attempts -1)
+                end
+        end,
+    {ok, Txs} = rpc(aec_tx_pool, peek, [infinity]),
+    ct:log("Txs count in pool: ~p", [length(Txs)]),
+    {ok, #{<<"count">> := TxsCount}} = GetPending(10),
```
lucafavatella: Reuse aec_test_utils:wait_for_it?

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -906,10 +922,19 @@ block_txs_count_pending(_Config) ->
...
-    TxsCount = length(aec_blocks:txs(PendingBlock)), 
-    {ok, 200, #{<<"count">> := TxsCount}} =
-        get_block_txs_count_preset("pending"),
+    GetPending =
+        fun TryToGet(Attempts) when Attempts < 1 ->
+                {error, no_pending_block_returned};
+            TryToGet(Attempts) ->
+                case get_block_txs_count_preset("pending") of
+                    {ok, 200, P} -> {ok, P};
+                    {ok, 404, _} -> TryToGet(Attempts -1)
```
lucafavatella: no sleep in between attempts?

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: This seems to be a mistake

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -906,10 +922,19 @@ block_txs_count_pending(_Config) ->
...
+                end
+        end,
+    {ok, Txs} = rpc(aec_tx_pool, peek, [infinity]),
+    ct:log("Txs count in pool: ~p", [length(Txs)]),
+    {ok, #{<<"count">> := TxsCount}} = GetPending(10),
+    true = TxsCount =< length(Txs),
```
lucafavatella: It appears to me that by the time "aec_tx_pool:peek" is called, all txs may have been already removed from mempool. Hence this line may be executed as`true = 1 =< 0` that causes badmatch exception.

May I suggest you describe that this test aims at testing so to simplify the logic?

apps/aehttp/test/aehttp_integration_SUITE.erl
lucafavatella: It seems to me that this test shall be reduced to the following algorithm:
```
do
{
  ensure mining stopped
  ensure there is a known number of txs in mempool, such that they fit in a block
  start mining
}
until ( (pending block txs is same as number of txs previously known to be in mempool plus coinbase) OR (10 seconds worth of attempts passed) )
```

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -817,20 +817,36 @@ internal_block_pending(_Config) ->
...
+                end
+        end,
+    {ok, PendingTxDefault} = GetPending(default, 10),
+    {ok, PendingTxHashes} = GetPending(false, 10),
+    ValidateKeys =
+        fun(Map1, Map2, Key) ->
+            true = maps:get(Key, Map1, not_found1) =:=
+                   maps:get(Key, Map2, not_found2)
+        end,
+    ValidateKeys(ExpectedPendingTx, PendingTxDefault, <<"prev_hash">>),
```
lucafavatella: I understand this may have a different prev_hash than ExpectedPendingTx.

lucafavatella: Maybe a text description of the aim of the test would help simplifying it?

apps/aecore/test/aec_conductor_tests.erl
```diff
@@ -381,6 +381,72 @@ test_block_publishing() ->
...
+        fun(_) ->
+            Tx = #spend_tx{sender = MyAccount,
+                          recipient = MyAccount,
+                          nonce = 0, fee = 0},
+            {ok, STx} = aec_keys:sign(Tx),
+            ok = aec_tx_pool:push(STx, tx_received)
+        end,
+        lists:seq(1, 3)),
+    ?TEST_MODULE:start_mining(),
+    ?assertEqual({error, miner_starting}, ?TEST_MODULE:get_block_candidate()),
```
lucafavatella: I see this relies on aec_mining:create_block_candidate sleep 100ms. Time-based though shall always work.

apps/aecore/test/aec_conductor_tests.erl
```diff
@@ -381,6 +381,72 @@ test_block_publishing() ->
...
+    timer:sleep(2000),
```
lucafavatella: Is it possible to replace this with aec_test_utils:wait_for_it ?

apps/aecore/test/aec_conductor_tests.erl
```diff
@@ -381,6 +381,72 @@ test_block_publishing() ->
...
+    ?assertEqual(TopBlockHash, aec_blocks:prev_hash(BlockCandidate)),
+    {ok, AllTxsInPool} = aec_tx_pool:peek(infinity),
+    ?assertEqual(true,
+        lists:all(
+            fun(SignedTx) ->
+                Tx = aec_tx_sign:data(SignedTx),
+                case aec_tx_dispatcher:handler(Tx) of
+                    aec_coinbase_tx -> true;
+                    _ ->
+                        lists:member(Tx, AllTxsInPool)
```
lucafavatella: This means "All txs in the block candidate (not yet mined - and never will as mocked to `no_solutions`) are in the mempool.". Ok.

apps/aehttp/test/aehttp_integration_SUITE.erl
lucafavatella: I see [CI error](https://travis-ci.org/aeternity/epoch/jobs/329058030#L1487):
```
%%% aehttp_integration_SUITE ==> all_endpoints.internal_endpoints.block_txs_count_pending: FAILED
%%% aehttp_integration_SUITE ==> {{badmatch,{error,no_pending_block_returned}},
 [{aehttp_integration_SUITE,block_txs_count_pending,1,
                            [{file,"/home/travis/build/aeternity/epoch/_build/test/lib/aehttp/test/aehttp_integration_SUITE.erl"},
                             {line,936}]},
  {test_server,ts_tc,3,[{file,"test_server.erl"},{line,1529}]},
  {test_server,run_test_case_eval1,6,[{file,"test_server.erl"},{line,1045}]},
  {test_server,run_test_case_eval,9,[{file,"test_server.erl"},{line,977}]}]}
```

I think it is because of lack of sleep (in test I mean).

velzevur: We don't actually know in advance the count of transactions in the block candidate (esp. this is prone to races) and thus we can not use aec_test_utils:wait_for_it/2,3

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: Removing txs from mempool happens when a new block is mined. This test expects that no block is mined durring test's execution. For this we mine the first X blocks (that have no difficulty readjustment) and then set the expected mine rate to be one block per hour. Taken into account the previous X blocks and the high mine rate should make the next block really harder to mine.

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: This should not have a different prev_hash (since it is not expected to mine a block durring that part of the test)

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -817,20 +817,36 @@ internal_block_pending(_Config) ->
...
+    ValidateKeys(ExpectedPendingTx, PendingTxDefault, <<"data_schema">>),
+
+    ValidateKeys(ExpectedPendingTx, PendingTxHashes, <<"prev_hash">>),
```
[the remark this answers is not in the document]
velzevur: This should not have a different prev_hash (since it is not expected to mine a block durring that part of the test)

apps/aecore/test/aec_conductor_tests.erl
velzevur: aec_test_utils:wait_for_it/2,3 needs an exact value to match in order to consider the condition met. Since we have no guarantee for the order of the transactions in the block, we can not use it.
I could implement a new function aec_test_utils:wait_for_it_not/2,3 that waits for the result NOT to be the one passed as a parameter (in our case - that the get_block_candidate/0 returns something different thant {error, miner_starting}

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: The idea behind the test is to veirfy that we have a _pending like_ block being returned by the API. Since the block candidate changes really rapidly, the check if it is actually the block being mined is left to the common tests

velzevur:
Since there will be subsequent PR, I've stopped Travis' checks to free some resources
The build had successfully passed checks in my fork [here](https://travis-ci.org/velzevur/epoch/builds/329408425) and I've restarted the `test` JOB a couple of times to check for a race

apps/aecore/test/aec_conductor_tests.erl
velzevur: I could not think of a better way to do it. I was thinking of blocking the execution of a process until a certain message is received but aec_minig:create_block_candidate is executed in the process of one of the conductor's workers...

apps/aehttp/test/aehttp_integration_SUITE.erl
happi: You can. You just have to be a bit creative.
Set the value to true and write the fun so that it only evaluates to true when your condition is met.

velzevur:
Reimplemented the test for transaction count so it now:
1. Put some spend txs to mempool (no more than block's maximum allowed size)
2. Start mining (on a high difficulty)
3. Try a couple of times to get the pending block's txs count
4. If fetched, it should equal the amount of txs spent (in **step 1**)

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -1345,7 +1369,8 @@ get_pending_block() ->
 add_spend_txs() ->
     MineReward = rpc(aec_governance, block_mine_reward, []),
     MinFee = rpc(aec_governance, minimum_tx_fee, []),
-    MaxTxs = MineReward div (1 + MinFee),
+    MaxTxsInBlock = rpc(aec_governance, max_txs_in_block, []),
+    MaxTxs = min(MineReward div (1 + MinFee), MaxTxsInBlock), % so it can fit in one block
```
lucafavatella: (Not needed for this PR.) This 1 here is the same 1 as the amount (10 lines below), a variable would clarify.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -1345,7 +1369,8 @@ get_pending_block() ->
 add_spend_txs() ->
     MineReward = rpc(aec_governance, block_mine_reward, []),
     MinFee = rpc(aec_governance, minimum_tx_fee, []),
-    MaxTxs = MineReward div (1 + MinFee),
+    MaxTxsInBlock = rpc(aec_governance, max_txs_in_block, []),
```
lucafavatella: -1 for coinbase?

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -912,14 +928,22 @@ block_txs_count_pending(_Config) ->
...
+                case get_block_txs_count_preset("pending") of
+                    {ok, 200, P} -> {ok, P};
+                    {ok, 404, _} ->
+                        timer:sleep(100),
+                        TryToGet(Attempts -1)
+                end
+        end,
+    ct:log("Inserted transactions count ~p", [InsertedTxsCount]),
+    {ok, #{<<"count">> := TxsCount}} = GetPending(10),
+    true = TxsCount =:= InsertedTxsCount + 1,
```
lucafavatella: On failure this line will say "badmatch of value false on line xxx". May I suggest making potential failure more informative?
* Either log TxsCount (the test already logs InsertedTxsCount.); or
* Replace line with `{TxsCount, TxsCount} = {(InsertedTxsCount + 1), TxsCount}` (or similar).

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -912,14 +928,22 @@ block_txs_count_pending(_Config) ->
...
+            TryToGet(Attempts) ->
+                case get_block_txs_count_preset("pending") of
+                    {ok, 200, P} -> {ok, P};
+                    {ok, 404, _} ->
+                        timer:sleep(100),
+                        TryToGet(Attempts -1)
+                end
+        end,
+    ct:log("Inserted transactions count ~p", [InsertedTxsCount]),
+    {ok, #{<<"count">> := TxsCount}} = GetPending(10),
```
lucafavatella: I suggest adding comment on idea behind test design like "`The custom set high expected inter-block interval (`expected_mine_rate`) makes almost certain that the retrieved pending block is the first after having added the spend txs."

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -827,20 +827,36 @@ internal_block_pending(_Config) ->
                 <<"BlockWithTxsHashes">>,
```
lucafavatella: I see that two lines above (at line 825) `get_pending_block()` re-attempts. I would therefore recommend removal of `timer:sleep(100)` at line 824, as that makes it more likely for a block having already been mined (even if high difficulty for high inter-block interval set).

apps/aehttp/test/aehttp_integration_SUITE.erl
lucafavatella: > This should not have a different prev_hash (since it is not expected to mine a block durring that part of the test)

Maybe writing this as a comment here would help reader.

apps/aehttp/test/aehttp_integration_SUITE.erl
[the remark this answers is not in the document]
lucafavatella: > This should not have a different prev_hash (since it is not expected to mine a block durring that part of the test)

Maybe writing this as a comment here would help reader.

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: Actually we do have the TxsCount logged as we log every request being made
```

*** User 2018-01-16 14:46:36.735 ***
GET "http://127.0.0.1:3113/v1/block/txs/count/pending"


*** User 2018-01-16 14:46:36.736 ***
Return code 200, Body "{\"count\":5}"
```
I found it redundant to log it once more

lucafavatella: Ah I see. Ok then.

velzevur: I've added a log for better clarity

hanssv:
All 123 tests passed.

Dialyzer says 👍

PR #586, opened 2018-01-15
PT-154323788 Test that semantically invalid tx is not included in block

lucafavatella:
In scope of [PT-154323788](https://www.pivotaltracker.com/story/show/154323788).

### DIFF
--- a/apps/aecore/test/aecore_sync_SUITE.erl
+++ b/apps/aecore/test/aecore_sync_SUITE.erl
@@ -25,7 +25,9 @@
     restart_second/1,
     restart_third/1,
     tx_first_pays_second/1,
+    tx_first_pays_second_more_it_can_afford/1,
     ensure_tx_pools_empty/1,
+    ensure_tx_pools_one_tx/1,
     no_system_metrics_logged/1
    ]).
 
@@ -38,6 +40,7 @@ groups() ->
     [
      {all_nodes, [sequence], [{group, two_nodes},
                               {group, three_nodes},
+                              {group, semantically_invalid_tx},
                               {group, one_blocked}]},
      {two_nodes, [sequence],
       [start_first_node,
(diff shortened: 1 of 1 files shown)


apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -63,10 +66,17 @@ groups() ->
...
-       [start_first_node,
-        mine_on_first,
-        start_blocked_second]}
+     {semantically_invalid_tx, [sequence],
+      [start_first_node,
+       start_second_node,
+       ensure_tx_pools_empty,
+       tx_first_pays_second_more_it_can_afford,
+       mine_on_second,
+       ensure_tx_pools_one_tx]},
```
lucafavatella: This proves https://www.pivotaltracker.com/story/show/154339293

lucafavatella:
The package CI job failed with a network error, clearly not depending on this PR. Restarting job.

Error was:
> The command "wget -o $HOME/erlang.tar.bz2 ${archive_url}" failed and exited with 1 during .

apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -220,19 +238,33 @@ tx_first_pays_second(Config) ->
...
+                    fun(N) ->
+                            case get_pool(N) of
+                                {ok, Pool} when is_list(Pool) ->
+                                    ct:log("Pool (~p) = ~p", [N, Pool]),
+                                    Pool
+                            end
+                    end, Ns),
+              if
+                  length(APool) =:= TxsCount ->
+                      lists:all(fun(P) when P =:= APool -> true;
```
uwiger: This will work since the pools are consistently ordered, but I'm not sure if this fact is that well documented.

lucafavatella: > This will work since the pools are consistently ordered, but I'm not sure if this fact is that well documented.

Tracked this documentation and test as https://www.pivotaltracker.com/story/show/154362443

lucafavatella:
I see 3 approvals, CI passed. Merging.

PR #587, opened 2018-01-15
PT-154323788 Report mempool size as metric

lucafavatella:
In scope of [PT-154323788](https://www.pivotaltracker.com/story/show/154323788).

### DIFF
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -24,7 +24,8 @@
          delete/1,
          peek/1,
          fork_update/2,
-         get_max_nonce/1]).
+         get_max_nonce/1,
+         size/0]).
 
 %% gen_server callbacks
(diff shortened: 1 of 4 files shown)


uwiger:
An argument could be made that `[ae,epoch,aecore,tx_pool,pending_txs]` could simply be a `gauge` metric, updated each time the pool changes.

Another observation might be that it might be time to write an 'ae_sample_probe' behavior to reduce the code duplication. The different probes we have basically only differ in the sample functions.

lucafavatella:
> Another observation might be that it might be time to write an 'ae_sample_probe' behavior to reduce the code duplication. The different probes we have basically only differ in the sample functions.

I tracked this refactoring as separate ticket https://www.pivotaltracker.com/story/show/154362264

lucafavatella:
I see two approvals, CI passed. Merging.

PR #592, opened 2018-01-16
HTTP API version v1 -> v2

ThomasArts:
Changing API version from v1 to v2
- updated in config/swagger.yaml
- Modified manually in documentation, python tests
- Modified tests that used hard-coded v2 (bascially in aehttp
- Re-generated code (swagger.json, swagger_router and endpoints.hrl).
Generated code probably should not be part of the repo, but for now we do it this way.

[Finishes PtT-154222225]

### DIFF
--- a/apps/aecore/src/aec_peers.erl
+++ b/apps/aecore/src/aec_peers.erl
@@ -505,6 +505,8 @@ ping_peer(Peer) ->
                 {error, protocol_violation} ->
                     block_peer(Uri);
                 _ ->
+                    %% If we ping a peer with wrong API version, time out in aue_request 
+                    %% and Peer is errored until it upgrades/downgrades to the same version.
                     log_ping_error(Peer)
             end;
         {error, timeout} ->
@@ -661,5 +663,3 @@ parse_uri(Uri) ->
 -spec uri_of_peer(peer()) -> http_uri:uri().
 uri_of_peer(#peer{host = Host, scheme = Scheme, port = Port}) ->
     aeu_requests:pp_uri({Scheme, Host, Port}).
-
-
--- a/apps/aehttp/src/aehttp_dispatch_ext.erl
+++ b/apps/aehttp/src/aehttp_dispatch_ext.erl
@@ -237,6 +237,9 @@ add_missing_to_genesis_block(Val) ->
     Val.
 
 handle_ping(#{<<"source">> := Src} = PingObj) ->
+    %% Source only contains host and port
+    %% If the http API version differs, then response ping will go
+    %% to wrong endpoint and eventually timeout.
     IsBlocked = aec_peers:is_blocked(Src),
     case IsBlocked of
         false -> handle_ping_(Src, PingObj);
--- a/apps/aeutils/include/endpoints.hrl
+++ b/apps/aeutils/include/endpoints.hrl
@@ -1,0 +1,138 @@
+%% This code is generated from swagger.yaml
+%% do not manually change it.
+
+
+operation(Id) ->
+    maps:get(Id, operations()).
+
+operations() ->
+    #{ 
+        'GetAccountBalance' => #{
+            path => <<"/v2/account/balance">>,
+            method => get
+        },
+        'GetAccountsBalances' => #{
+            path => <<"/v2/balances">>,
+            method => get
+        },
+        'GetBlockByHash' => #{
+            path => <<"/v2/block-by-hash">>,
+            method => get
+        },
+        'GetBlockByHeight' => #{
+            path => <<"/v2/block-by-height">>,
+            method => get
+        },
+        'GetInfo' => #{
+            path => <<"/v2/info">>,
+            method => get
+        },
+        'GetTop' => #{
+            path => <<"/v2/top">>,
+            method => get
+        },
+        'GetTxs' => #{
+            path => <<"/v2/transactions">>,
+            method => get
+        },
+        'GetVersion' => #{
+            path => <<"/v2/version">>,
+            method => get
+        },
+        'Ping' => #{
+            path => <<"/v2/ping">>,
+            method => post
+        },
+        'PostBlock' => #{
+            path => <<"/v2/block">>,
+            method => post
+        },
+        'PostTx' => #{
+            path => <<"/v2/tx">>,
+            method => post
+        },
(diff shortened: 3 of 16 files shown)


config/swagger.yaml
```diff
@@ -920,18 +920,30 @@ paths:
...
+  Pow:
+    type: array
+    items:
+      type: integer
+      format: int32
+    minItems: 42
+    maxItems: 42
+  Uri:
+    type: string
+    ## format: uri
```
lucafavatella: I understand this can be safely uncommented (documentation [here](https://github.com/OAI/OpenAPI-Specification/blob/ba49b592c13594d06db4ea99fe4cd795b3d83ace/versions/2.0.md#data-types)).

config/swagger.yaml
```diff
@@ -920,18 +920,30 @@ paths:
       security: []
 
 definitions:
+  EncodedHash:
+    type: string
+  Pow:
+    type: array
+    items:
+      type: integer
+      format: int32
```
lucafavatella: [Unsigned integer 32 bits](https://github.com/aeternity/protocol/blob/f578d480f8c8e13c1425f26ebb732db01d099ab3/CONSENSUS.md#solution-encoding-rules) so `int32` is unable to contain real values sometimes. Options are (in my order of preference):
* format uint32 (not documented in swagger doc, but shall be ok to use it as a hint - see [here](https://github.com/OAI/OpenAPI-Specification/blob/ba49b592c13594d06db4ea99fe4cd795b3d83ace/versions/2.0.md#data-types))
* no format
* format int64 (specified by swagger spec), potentially with min value 0

apps/aeutils/src/aeu_requests.erl
```diff
@@ -1,3 +1,38 @@
...
+%%      Erlang maps or other types used internally. 
+%%%     Some business logic may be verified already here. 
+%%      For example,
+%%%     the ping request provides a "share" parameter stating how many
+%%%     peers it maximally want to receive in the response.
+%%%     It is hard to express this relation in the JSON schema(s), but 
+%%%     easy to verify here that the list of returned pings has a length 
+%%%     not exceeding Share. 
+%%%     
+%%%     Not that we perform both the JSON encoding and the actual
```
lucafavatella: Note

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -237,6 +237,9 @@ add_missing_to_genesis_block(Val) ->
     Val.
 
 handle_ping(#{<<"source">> := Src} = PingObj) ->
+    %% Source is only contains host and port
```
hanssv: s/is only/only/

apps/aeutils/include/endpoints.hrl
```diff
@@ -0,0 +1,138 @@
+%% This coude is generated from swagger.yaml
```
hanssv: s/coude/code/

config/swagger.yaml
ThomasArts: The Swagger specification tells us that uri is a valid format, but it is not accepted by the tool chain!

apps/aeutils/src/aeu_http_client.erl
```diff
@@ -54,20 +58,10 @@ process_http_return(R) ->
...
-
-%% str(A) when is_atom(A) ->
-%%     atom_to_binary(A, latin1);
-str(S) when is_list(S); is_binary(S) ->
-    S.
-
-uenc(V) ->
-    http_uri:encode(V).
+encode_get_params(Params) when is_list(Params) ->
+    ["?", string:join( [ [http_uri:encode(K), "=" , http_uri:encode(V)]
```
lucafavatella: string:join seems obsolete

config/swagger.yaml
ThomasArts: But we only use 4 byte integers... Having int64 blows up the random tests that take any arbitrary integer in that range. It's the min and max in that case we would need to set.

apps/aeutils/src/aeu_http_client.erl
ThomasArts: fine, lists:join then.

config/swagger.yaml
ThomasArts: I believe that if we have in the past and still used int64 everywhere, then we can run as well with int32 in a format field. And yes, all int64 mentioned before also were unsigned, but this was and is unspecified. This is an unused over specification of what the API can deal with, completely in line how swagger works.

apps/aecore/src/aec_peers.erl
```diff
@@ -505,6 +505,8 @@ ping_peer(Peer) ->
                 {error, protocol_violation} ->
                     block_peer(Uri);
                 _ ->
```
lucafavatella: I see this is equivalent to `{error, _}`.

ThomasArts: No it is not, because in protocol validation the side effect is blocking the peer, for other errors (like difference in version) the peer is not blocked.

apps/aecore/src/aec_peers.erl
```diff
@@ -505,6 +505,8 @@ ping_peer(Peer) ->
...
+                    %% If we ping a peer with wrong API version, time out in aue_request
```
lucafavatella: I do not expect aeu_requests:ping to time out rather to either return an error value or err.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -237,7 +237,7 @@ add_missing_to_genesis_block(Val) ->
     Val.
 
 handle_ping(#{<<"source">> := Src} = PingObj) ->
-    %% Source is only contains host and port
```
lucafavatella: You may be squashing changes addressing comments, and that are not automatically generated, in commit meant for generated code changes. It may be easier to just squash as one commit.

velzevur:
Please change the [Python expectation](https://github.com/aeternity/epoch/blob/master/py/tests/swagger_client/configuration.py#L48) for the version of the API.
You can do this by hand or running
```
make swagger-python```
Nobody has ran it for quite some time, I think the current state is bug prone. I'll address it in [other PT](https://www.pivotaltracker.com/story/show/154368099)

PR #593, opened 2018-01-16
Docker

dincho:
- add Dockerfile to allow building docker images `docker build -t epoch .`
- add documentation how to use the image
- add section in the release notes

Todo:
- [x] add support to change `peers`
- [ ] @dincho to update the *Configuration Examples* on [User Configuration Wiki](https://github.com/aeternity/epoch/wiki/User-provided-configuration)

Test with:
```bash
docker run -d -p 3013:3013 aetrnty/epoch
```
or follow the docs.

https://www.pivotaltracker.com/story/show/153835478

[PT ticket](https://www.pivotaltracker.com/story/show/154393528) to track setup of automatic builds


docs/docker-testnet.md
```diff
@@ -0,0 +1,35 @@
...
+
+Docker image is automatically build and published on [DockerHub](https://hub.docker.com/r/aetrnty/epoch/). All tagged source code versions have their own docker image tag as well. Latest tagged ("stable") version is tagged with `latest`.
+Master branch of the source code is tagged as `dev`.
+
+To pull the latest "stable" image run:
+```bash
+docker pull aetrnty/epoch
+```
+
+## Start a Node
```
lucafavatella: How to stop the node? (For saving power at end of testing.)

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+ENV SHELL /bin/bash
+ENV PEER_ADDRESS 127.0.0.1
+
+WORKDIR /home/epoch/node
+
+# Erl handle SIGQUIT instead of the default SIGINT
+STOPSIGNAL SIGQUIT
+
+EXPOSE 3013
```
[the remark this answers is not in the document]
dincho: good idea see 1bdebb7

docs/docker-testnet.md
dincho: Well, I tried to abstain from documenting docker itself instead of our image but here you go: 161a61b

RELEASE-NOTES.md
```diff
@@ -40,6 +40,18 @@ Information, e.g. height, of the top block of the longest chain as seen by these
 * http://31.13.248.102:3013/v1/top
 * http://31.13.248.105:3013/v1/top
 
+### Using Docker
```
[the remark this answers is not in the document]
dincho: Well, I think few rows is nice to have for much easier user on-boarding.

We have to refactor the notes anyway and split to: install from package, run with docker, troubleshooting, localnet with docker (and without??) in a separate PR/ticket.

docs/docker-testnet.md
```diff
@@ -19,8 +19,17 @@ To start a docker node and join the testnet run:
...
 ```
 
-- *Your public IP address will be automatically determined and used in `peer_address` configuration option*
-- *Make sure you have a working port forwarding setup on your firewall to be able to fully participate in the testnet*
+Verify the node is running:
+```bash
+curl localhost:3013/v1/top
+```
+
+Your public IP address will be automatically determined and used in `peer_address` configuration option. You can change your `peer_address` by using `PEER_ADDRESS` Docker environment variable e.g.:
```
ThomasArts: Is this related to the peer_address in epoch.yaml ? Because in that peer_address the port can be specified as part of the address and in this example you only specify the IP address

ThomasArts: Only when reading the next commit, it is clear that EXTERNAL_PEER_ADDRESS is something different. Not sure why you need both.

dincho: @ThomasArts Could you please check the latest changes, I changed this multiple times. The last we have is:
- EXTERNAL_PEER_ADDRESS (e.g. http://1.2.3.4:3013/)
- PEERS_ADDRESS_0 (e.g. http://31.13.248.102:3013/)

I really can't come up with better idea to provide and array/list of PEERS, thus this ugly PEERS_ADDRESS_0 and the cons. that we cannot set more than one peer address.

docs/docker-testnet.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+## Stop a Node
+
+To stop a docker node run:
+```bash
+docker stop epoch_node0
+```
+
+## Configuration
```
lucafavatella: How does user consult logs? (epoch*.log)

dincho: They could like on any docker container `docker exec epoch_node0 cat log/epoch.log`

lucafavatella: A note where the logs are stored in the container would help the user I believe (the `docker exec epoch_node0 cat log/epoch.log` would be enough). Other option would be encouraging user to use volume for logs.

dincho: updated in c8dcad2

docs/docker-testnet.md
```diff
@@ -0,0 +1,77 @@
...
+## Execute command on a running node
+
+To execute any command on a running node use:
+```bash
+docker exec epoch_node0 CMD
+```
+
+For example to check what's in the epoch logs run:
+```bash
+docker exec epoch_node0 tail log/epoch.log
```
lucafavatella: thanks

dincho:
Wiki has been updated https://github.com/aeternity/epoch/wiki/User-provided-configuration/_compare/523d9ea1e9bc94b87a6bc6b43146ad3364367dad...d2e7c05124885ba2b345d773389784ed27c9c005

PR #595, opened 2018-01-16
Pt 154281908 hand coded contract

happi:
This PR implements a simple assembler for the AEVM.
You can now write AEVM/EVM byte code as assembler code in a file with labels in stead of hard coded addresses. This code can then be parsed and turned into byte code.

There is also an example of a hand coded implementation of the identity contract which uses the Ring ABI and returns the same integer as you give as an argument to the call.

The example does not contain an init function yet which should be needed for Ring contracts.
A future PT and PR will add integration test cases for the assembler/VM integration.

(NOTE: this is not needed for the 0.5 release but it should have little impact.)

[Finishes #154281908]

### DIFF
--- a/apps/aebytecode/src/aeb_asm.erl
+++ b/apps/aebytecode/src/aeb_asm.erl
@@ -1,7 +1,29 @@
 %%%-------------------------------------------------------------------
 %%% @copyright (C) 2017, Aeternity Anstalt
-%%% @doc
-%%%     Assembler for aevm machine code
+%%% @doc Assembler for aevm machine code.
+%%%
+%%%      Assembler code can be read from a file.
+%%%      The assembler has the following format
+%%%      Comments start with 2 semicolons and runs till end of line
+%%%         ;; This is a comment
+%%%      Opcode mnemonics start with an upper case letter.
+%%%         DUP1
+%%%      Identifiers start with a lower case letter
+%%%         an_identifier
+%%%      All identifiers has to be decleard with a label.
+%%%      A label is an identifier that ends with a colon.
+%%%      The label is the byte code address of the next instruction.
+%%%         a_label:
+%%%      Immidiates can be of four types:
+%%%       1. integers
+%%%          42
+%%%       2. hexadecimal integers starting with 0x
+%%%          0x0deadbeef0
+%%%       3. 256-bit hash strings starting with #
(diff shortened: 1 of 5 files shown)


apps/aebytecode/src/aeb_asm.erl
```diff
@@ -1,14 +1,38 @@
...
+%%%         ;; This is a comment
+%%%      Opcode mnemonics start with an upper case letter.
+%%%         DUP1
+%%%      Identifiers start with a lower case letter
+%%%         an_identifier
+%%%      All identifiers has to be decleard with a label.
+%%%      A label is an identifier that ends with a colon.
+%%%      The label is the byte code address of the next instruction.
+%%%         a_label:
+%%%      Immidiates can be of four types:
```
hanssv: s/Immidiates/Immediates/

apps/aering/test/contracts/identity.aesm
```diff
@@ -0,0 +1,63 @@
+;; CONTRACT: Identity
+
+             PUSH1              0
+             CALLDATALOAD
+
+             DUP1
+             ;; Should be the has of
```
ThomasArts: the hash of

apps/aebytecode/src/aeb_asm.erl
```diff
@@ -37,25 +53,37 @@ to_hexstring(ByteList) ->
...
     OpCode = aeb_opcodes:m_to_op(Op),
     OpSize = aeb_opcodes:op_size(OpCode),
-    to_bytecode(Rest, Address + OpSize, Env, [OpCode|Code]);
-to_bytecode([{int,_line, Int}|Rest], Address, Env, Code) ->
-    to_bytecode(Rest, Address, Env, [Int|Code]);
-to_bytecode([{hash,_line, Hash}|Rest], Address, Env, Code) ->
-    to_bytecode(Rest, Address, Env, [Hash|Code]);
-to_bytecode([{id,_line, ID}|Rest], Address, Env, Code) ->
-    to_bytecode(Rest, Address, Env, [{ref, ID}|Code]);
-to_bytecode([{lable,_line, Lable}|Rest], Address, Env, Code) ->
```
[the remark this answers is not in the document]
happi: One would think that after 20 years of writing compilers I could spell to label... but for some reason I just cant.

PR #596, opened 2018-01-16
Add block hash to internal block APIs

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/154359261)

### DIFF
--- a/config/swagger.yaml
+++ b/config/swagger.yaml
@@ -486,7 +486,7 @@ paths:
           type : integer
         - in: query
           name: tx_objects
-          description: 'Transactions as objects (default is as MessagePack hashes)'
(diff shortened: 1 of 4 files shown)


config/swagger.yaml
```diff
@@ -965,6 +965,8 @@ definitions:
         properties:
           data_schema:
             type: string
+          hash:
+            type: string
```
ThomasArts: When you merge with master later on, please use  $ref: '#/definitions/EncodedHash'

velzevur:
Addressed @ThomasArts 's comment.
A successful run of the Travis' checks can be seen [here](https://travis-ci.org/velzevur/epoch/builds/330419185)

PR #602, opened 2018-01-18
Refactor HTTP integration tests [Finishes #154273965]

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/154273965)
* Refactored all tests to be ran subsequently without the need of starting and stopping of the node being tested
* Lowered the amount of itterations of tests where this didn't provide any stronger guarantees
* Exposed `aec_conductor:reinit_chain/0` to be used by the tests (that do depend on an empty chain so we don't restart a node just for that).

All the inefficiencies stated above being removed - the integration tests should run a couple of times faster.

### DIFF
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -90,6 +90,10 @@
         , top_header_hash/0
         ]).
 
+%% for tests
+-export([reinit_chain/0
+        ]).
+
 %% gen_server API
 -export([ start_link/0
         , start_link/1
@@ -305,6 +309,10 @@ top_header() ->
 top_header_hash() ->
     gen_server:call(?SERVER, get_top_header_hash).
 
+-spec reinit_chain() -> header().
+reinit_chain() ->
+    gen_server:call(?SERVER, reinit_chain).
+
 %%%===================================================================
 %%% gen_server callbacks
 %%%===================================================================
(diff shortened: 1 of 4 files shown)


velzevur:
A successful tests run can be seen [here](https://travis-ci.org/velzevur/epoch/builds/330410179)

velzevur:
In `aec_conductor:reinit_chain/0` I've handled the discarding the currently running mining attempt and restarting it.

apps/aecore/src/aec_conductor.erl
```diff
@@ -90,6 +90,9 @@
         , top_header_hash/0
         ]).
 
+-export([reinit_chain/0
```
lucafavatella: This may benefit from a comment that it is aimed at tests.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -305,19 +335,18 @@ block_by_height(_Config) ->
...
 block_not_found_by_height(_Config) ->
-    TopHeader = rpc(aec_conductor, top_header, []), 
-    0 = aec_headers:height(TopHeader), %chain is empty
-    NumberOfChecks = 10,
+    InitialHeight = aec_blocks:height(rpc(aec_conductor, top, [])),
+    NumberOfChecks = 5,
     lists:foreach(
         fun(_) ->
-            Height = rand:uniform(99) + 1, % random number 1-100
+            Height = rand:uniform(99) + 1 + InitialHeight, % random number 1-100
```
hanssv: Unless InitialHeight is always 0 the comment is wrong...

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -287,10 +315,12 @@ get_top_non_empty_chain(_Config) ->
...
 block_by_height(_Config) ->
     BlocksToCheck = 4,
+    InitialHeight = aec_blocks:height(rpc(aec_conductor, top, [])),
+    BlocksToMine = max(BlocksToCheck - InitialHeight, 0),
     aecore_suite_utils:mine_blocks(aecore_suite_utils:node_name(?NODE),
-                                   BlocksToCheck),
+                                   BlocksToMine),
     TopHeader = rpc(aec_conductor, top_header, []), 
-    BlocksToCheck = aec_headers:height(TopHeader),
+    true = BlocksToCheck =< aec_headers:height(TopHeader),
```
lucafavatella: It would be nice to show bad height in case of error in order to have more clues on the failure.
E.g. `case aec_headers:height(TopHeader) of TopHeaderHeight when TopHeaderHeight >= BlocksToCheck end,`

apps/aehttp/test/aehttp_integration_SUITE.erl
lucafavatella: Comment now misplaced/

lucafavatella: I am concerned that while tests executes from `aec_conductor:top` to here, the node managed to find one ore more blocks (unlikely but possible). Would it make sense to add a padding of e.g. 2 blocks? (out of scope of this PR of course)

velzevur: The node mining a new block is quite unlikely since there is no ongoing mining attempt. Since this is an expectation for the test,  I'll add an appropriate assertion.

velzevur:
Addressed all comments

PR #603, opened 2018-01-19
Fix edoc.

happi:


happi:
Travis failed on execution timeout. @ThomasArts has tested locally.

PR #605, opened 2018-01-19
Implement a functional Merkle Patricia Tree

gorillainduction:
This PR adds an implementation of Merkle Patricia Trees as (very vaguely)  defined in https://github.com/ethereum/wiki/wiki/Patricia-Tree with input from the reference implementation at https://github.com/ethereum/go-ethereum/tree/master/trie

This PR also include an implementation of RLP https://github.com/ethereum/wiki/wiki/RLP

The implementation lacks proofs.

Additional testing has been done using Quickcheck https://github.com/Quviq/epoch-eqc

### DIFF
--- a/apps/aeutils/src/aeu_mp_trees.erl
+++ b/apps/aeutils/src/aeu_mp_trees.erl
@@ -1,0 +1,543 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2017, Aeternity Anstalt
+%%% @doc Merkle Patricia Trees
+%%%
+%%% @end
+%%%-------------------------------------------------------------------
+
+-module(aeu_mp_trees).
+
+-export([ new/0
+        , new/2
+        , delete/2
+        , get/2
+        , pp/1
+        , put/3
+        , root_hash/1
+        ]).
+
+-export_type([tree/0
+             ]).
+
+-record(mpt, { hash = <<>> :: <<>> | hash() %% <<>> for the empty tree
+             , db   = dict:new() :: dict:dict()
+             }).
+
+-opaque tree() :: #mpt{}.
+
+-type tree_node() :: null() | leaf() | extension() | branch().
+
+-type null()         :: <<>>.
+-type leaf()         :: {'leaf', path(), value()}.
+-type extension()    :: {'ext', path(), hash()}.
+-type branch()       :: {'branch', branch_tuple()}.
+-type branch_tuple() :: {enc_node(), enc_node(), enc_node(), enc_node(),
+                         enc_node(), enc_node(), enc_node(), enc_node(),
+                         enc_node(), enc_node(), enc_node(), enc_node(),
+                         enc_node(), enc_node(), enc_node(), enc_node(),
+                         value()}.
+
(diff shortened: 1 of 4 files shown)


apps/aeutils/src/aeu_mp_trees.erl
```diff
@@ -0,0 +1,543 @@
...
+        , get/2
+        , pp/1
+        , put/3
+        , root_hash/1
+        ]).
+
+-export_type([tree/0
+             ]).
+
+-record(mpt, { hash = <<>> :: <<>> | hash() %% <<>> for the empty tree
```
ThomasArts: In the spirit of using one specific hash function, would it make sense to use the type of that hash library, e.g. aec_sha256:hash()?

apps/aeutils/src/aeu_mp_trees.erl
```diff
@@ -0,0 +1,543 @@
...
+-type encoded_path() :: binary().  %% Compact encoding of hex sequence.
+
+-type key()          :: <<_:4,_:_*4>>. %% Non-empty hexstring
+-type value()        :: aeu_rlp:encodable().
+-type hash()         :: <<_:256>>.
+
+-type db()           :: dict:dict().
+
+%%-define(debug_on, true).
+-ifdef(debug_on).
```
ThomasArts: the default macro is DEBUG

apps/aeutils/src/aeu_mp_trees.erl
gorillainduction: Agreed. There is a ticket for doing that, and the current available abstracted hash function is specifying the sha256, so I didn't bother using it.

apps/aeutils/src/aeu_mp_trees.erl
gorillainduction: I can remove the debug stuff altogether.

apps/aeutils/src/aeu_mp_trees.erl
```diff
@@ -0,0 +1,543 @@
...
+
+-record(mpt, { hash = <<>> :: <<>> | hash() %% <<>> for the empty tree
+             , db   = dict:new() :: dict:dict()
+             }).
+
+-opaque tree() :: #mpt{}.
+
+-type tree_node() :: null() | leaf() | extension() | branch().
+
+-type null()         :: <<>>.
```
ThomasArts: The null() type is not used in for example tree() (mpt record). Is that on purpose? Just wondering whether the <<>> in the tree is actually also a null type. In other words, if we were to rename null() to atom null instead, would type checker complain?

gorillainduction: The null node is very special. It is implicitly present in all parts of the tree, it contains the null value etc. It is encoded as <<>>, so setting the type to 'null' would not work unless we used the value 'null' as well, which would not work since 'null' is not encodable in RLP.

gorillainduction: The tree() type is just the top level wrapper of the ADT. It is not a tree in the internal encoding of the tree. It is unwrapped/wrapped at the API functions. The internal tree has the type tree_node() which is including the null() type.

PR #606, opened 2018-01-19
Pt 154363216 integration test asm aevm

happi:
A simple ct framework (file location) for intra application tests.

The change to the aeb_asm_scan is to safely create atoms without list_to_existing_atoms when not all modules and hence not all atoms are loaded.


Makefile
```diff
@@ -118,9 +118,13 @@ dialyzer-install:
 dialyzer:
 	@./rebar3 dialyzer
 
-test:
+test: test-app-integration
 	@./rebar3 as test do release, ct $(CT_TEST_FLAGS) --sys_config config/test.config
 
+test-app-integration:
+	@./rebar3 ct $(CT_TEST_FLAGS) --dir test/*/
```
lucafavatella: It is not clear to me why a separate `rebar3 ct` invocation needed. I shall try moving the test to a new test-only app before approving.

lucafavatella: I see this is actually documented in https://www.pivotaltracker.com/story/show/154363216

lucafavatella: @happi I am trying to understand the need for separate `rebar3 ct` call. May you please have a look at https://github.com/aeternity/epoch/pull/612 ?

lucafavatella: As agreed please add `as test` (even if I expect it to be default profile).

lucafavatella:
On 1b42451a , clean working dir:
```
$ epmd -daemon && make test
...
...
%%% aec_eunit_SUITE ==> eunit.aec_tx_common_tests: OK
...
%%% aer_eunit_SUITE ==> eunit.aer_scan_tests: OK
...
%%% aebytecode_aevm_SUITE ==> execute_identy_fun_from_file: OK
All 127 tests passed.
===> Verifying dependencies...
===> Generating code from "config/swagger.yaml" writing to "apps/aeutils/src/endpoints.erl"
===> Compiling aecuckoo
===> Compiling aebytecode
===> Compiling aering
===> Compiling aetx
===> Compiling aecontract
===> Compiling aevm
===> Compiling aeoracle
===> Compiling aeutils
===> Compiling check_config
===> Compiling aecore
===> Compiling aehttp
===> Building escript...
Directory /Users/luca/dev/aeternity/epoch/_build/test/extras/test/* is invalid
===> Error running tests:
  {make_failed,["/Users/luca/dev/aeternity/epoch/_build/test/extras/test/*/all"]}
make: *** [test] Error 1
```

PR #607, opened 2018-01-19
Add Naming System initial implementation

sennui:
Finishes: #154277099, #154276940, #154277070, #154279185, #154276968, #154277126, #154313308, #154277497, #154277237, #154314897,

### DIFF
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -35,6 +35,8 @@
         , top_header/1
         , top_header_hash/1
         , get_genesis_hash/1
+        , name_entry/2
+        , resolve_name/3
         ]).
 
 %% used by aec_db for bootstrapping from persistence/backup
@@ -308,6 +310,30 @@ get_top_N_blocks_time_summary(?assert_state() = State, N)
             get_N_nodes_time_summary(TopNode, State, N);
         error ->
             []
+    end.
+
+-spec name_entry(binary(), state()) ->
+                        {'ok', map()} |
+                        {'error', 'no_state_trees'} |
+                        {'error', 'name_not_found'}.
+name_entry(Name, ?assert_state() = State) ->
+    TopHash = get_top_header_hash(State),
+    case state_db_find(TopHash, State) of
+        {ok, Trees} ->
+            aens:get_name_entry(Name, aec_trees:ns(Trees));
+        error ->
+            {error, no_state_trees}
+    end.
+
+-spec resolve_name(atom(), binary(), state()) -> {'ok', binary()} |
+                                                 {error, atom()}.
+resolve_name(Type, Name, ?assert_state() = State) ->
+    TopHash = get_top_header_hash(State),
+    case state_db_find(TopHash, State) of
+        {ok, Trees} ->
+            aens:resolve(Type, Name, aec_trees:ns(Trees));
+        error ->
+            {error, no_state_trees}
     end.
 
 %%%===================================================================
--- a/apps/aecore/src/aec_governance.erl
+++ b/apps/aecore/src/aec_governance.erl
@@ -5,7 +5,13 @@
          expected_block_mine_rate/0,
          block_mine_reward/0,
          max_txs_in_block/0,
-         minimum_tx_fee/0]).
+         minimum_tx_fee/0,
+         name_preclaim_expiration/0,
+         name_claim_burned_fee/0,
+         name_claim_max_expiration/0,
+         name_protection_period/0,
+         name_claim_preclaim_delta/0,
+         name_registrars/0]).
 
 -define(BLOCKS_TO_CHECK_DIFFICULTY_COUNT, 10).
 -define(EXPECTED_BLOCK_MINE_RATE, 300000). %% 60secs * 1000ms * 5 = 300000msecs
@@ -29,3 +35,24 @@ max_txs_in_block() ->
 
 minimum_tx_fee() ->
     1.
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%% Naming system variables
+
+name_preclaim_expiration() ->
+    300.
+
+name_claim_burned_fee() ->
+    3.
+
+name_claim_max_expiration() ->
+    50000.
+
+name_protection_period() ->
+    2016.
+
+name_claim_preclaim_delta() ->
+    1.
+
+name_registrars() ->
+    [<<"aet">>, <<"test">>].
--- a/apps/aecore/src/aec_trees.erl
+++ b/apps/aecore/src/aec_trees.erl
@@ -13,12 +13,14 @@
 -export([accounts/1,
          hash/1,
          new/0,
+         ns/1,
          oracles/1,
          contracts/1,
          perform_pre_transformations/2,
          set_accounts/2,
          set_oracles/2,
-         set_contracts/2
+         set_contracts/2,
+         set_ns/2
         ]).
 
 %%%%=============================================================================
@@ -29,7 +31,8 @@
 new() ->
     #trees{accounts  = aec_accounts_trees:empty(),
            oracles   = aeo_state_tree:empty(),
-           contracts = aect_state_tree:empty()
+           contracts = aect_state_tree:empty(),
+           ns        = aens_state_tree:empty()
           }.
 
 hash(Trees) ->
@@ -43,6 +46,14 @@ accounts(Trees) ->
 set_accounts(Trees, Accounts) ->
     Trees#trees{accounts = Accounts}.
 
+-spec ns(trees()) -> aens_state_tree:tree().
+ns(Trees) ->
+    Trees#trees.ns.
+
+-spec set_ns(trees(), aens_state_tree:tree()) -> trees().
+set_ns(Trees, Names) ->
+    Trees#trees{ns = Names}.
+
 -spec oracles(trees()) -> aeo_state_tree:tree().
 oracles(Trees) ->
     Trees#trees.oracles.
@@ -51,8 +62,10 @@ oracles(Trees) ->
 set_oracles(Trees, Oracles) ->
     Trees#trees{oracles = Oracles}.
 
+-spec perform_pre_transformations(trees(), non_neg_integer()) -> trees().
 perform_pre_transformations(Trees, Height) ->
-    set_oracles(Trees, aeo_state_tree:prune(Height, oracles(Trees))).
+    Trees1 = set_oracles(Trees, aeo_state_tree:prune(Height, oracles(Trees))),
+    set_ns(Trees1, aens_state_tree:prune(Height, ns(Trees1))).
 
 -spec contracts(trees()) -> aect_state_tree:tree().
 contracts(Trees) ->
@@ -70,9 +83,11 @@ internal_hash(Trees) ->
     AccountsHash = pad_empty(aec_accounts_trees:root_hash(accounts(Trees))),
     ContractsHash = pad_empty(aect_state_tree:root_hash(contracts(Trees))),
     OraclesHash = pad_empty(aeo_state_tree:root_hash(oracles(Trees))),
+    NamingSystemHash = pad_empty(aens_state_tree:root_hash(ns(Trees))),
     List = lists:sort([ {<<"accounts"/utf8>> , AccountsHash}
                       , {<<"contracts"/utf8>>, ContractsHash}
                       , {<<"oracles"/utf8>>  , OraclesHash}
+                      , {<<"ns"/utf8>>       , NamingSystemHash}
                       ]),
     TopTree = lists:foldl(fun({Key, Val}, Acc) ->
                                   aeu_mtrees:enter(Key, Val, Acc)
--- a/apps/aehttp/src/aehttp_dispatch_ext.erl
+++ b/apps/aehttp/src/aehttp_dispatch_ext.erl
@@ -139,6 +139,25 @@ handle_request('GetAccountBalance', Req, _Context) ->
             end
     end;
 
+handle_request('GetName', Req, _Context) ->
+    Name = case maps:get('name', Req) of
+               undefined ->
+                   {400, [], #{reason => <<"Name not specified">>}};
+               N when is_binary(N) ->
+                   N
+           end,
+    case aec_conductor:get_name_entry(Name) of
+        {error, name_not_found} ->
+            {404, [], #{reason => <<"Name not found">>}};
+        {ok, NameEntry} ->
+            #{<<"name">>     := Name,
+              <<"name_ttl">> := NameTTL,
+              <<"pointers">> := Pointers} = NameEntry,
+            {200, [], #{name     => Name,
+                        name_ttl => NameTTL,
+                        pointers => Pointers}}
+    end;
+
 handle_request('GetAccountsBalances', _Req, _Context) ->
     case application:get_env(aehttp, enable_debug_endpoints, false) of
         true ->
--- a/apps/aehttp/src/aehttp_dispatch_int.erl
+++ b/apps/aehttp/src/aehttp_dispatch_int.erl
@@ -16,8 +16,7 @@ handle_request('PostSpendTx', #{'SpendTx' := SpendTxObj}, _Context) ->
             #{<<"recipient_pubkey">> := EncodedRecipientPubkey,
               <<"amount">>           := Amount,
               <<"fee">>              := Fee} = SpendTxObj,
-            case aec_base58c:safe_decode(
-                   account_pubkey, EncodedRecipientPubkey) of
+            case aec_conductor:resolve_name(account_pubkey, EncodedRecipientPubkey) of
                 {ok, DecodedRecipientPubkey} ->
                     {ok, SpendTx} =
                         aec_spend_tx:new(
@@ -77,7 +76,7 @@ handle_request('PostOracleQueryTx', #{'OracleQueryTx' := OracleQueryTxObj}, _Con
               <<"fee">>           := Fee} = OracleQueryTxObj,
             QueryTTLType = binary_to_existing_atom(maps:get(<<"type">>, QueryTTL), utf8),
             QueryTTLValue= maps:get(<<"value">>, QueryTTL),
-            case aec_base58c:safe_decode(oracle_pubkey, EncodedOraclePubkey) of
+            case aec_conductor:resolve_name(oracle_pubkey, EncodedOraclePubkey) of
                 {ok, DecodedOraclePubkey} ->
                     {ok, OracleQueryTx} =
                         aeo_query_tx:new(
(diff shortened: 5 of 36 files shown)


apps/aecore/src/aec_chain_state.erl
```diff
@@ -271,7 +272,11 @@ find_common_ancestor(Hash1, Hash2, ?assert_state() = State) ->
 get_state_trees_for_persistence(?assert_state() = State) ->
     state_db_to_list(State).
 
--spec new_from_persistence([#block{}|#header{}], [{Hash :: binary(), #trees{}}]) -> state().
```
hanssv: This spec should not be removed I think?

apps/aecore/src/aec_trees.erl
```diff
@@ -70,9 +83,11 @@ internal_hash(Trees) ->
     AccountsHash = pad_empty(aec_accounts_trees:root_hash(accounts(Trees))),
     ContractsHash = pad_empty(aect_state_tree:root_hash(contracts(Trees))),
     OraclesHash = pad_empty(aeo_state_tree:root_hash(oracles(Trees))),
+    NamingSystemHash = pad_empty(aens_state_tree:root_hash(ns(Trees))),
     List = lists:sort([ {<<"accounts"/utf8>> , AccountsHash}
                       , {<<"contracts"/utf8>>, ContractsHash}
                       , {<<"oracles"/utf8>>  , OraclesHash}
+                      , {<<"ns"/utf8>>      , NamingSystemHash}
```
hanssv: badly aligned `,`

apps/aens/src/aens_claim_tx.erl
```diff
@@ -0,0 +1,210 @@
...
+                       nonce      = None,
+                       name       = Name,
+                       name_salt = NameSalt,
+                       fee        = Fee}) ->
+    [#{<<"type">>       => type()},
+     #{<<"vsn">>        => version()},
+     #{<<"account">>    => AccountPubKey},
+     #{<<"nonce">>      => None},
+     #{<<"name">>       => Name},
+     #{<<"name_salt">> => NameSalt},
```
hanssv: name_salt is badly aligned throughout the module

config/swagger.yaml
```diff
@@ -218,6 +218,31 @@ paths:
           schema:
             $ref: '#/definitions/Error'
       security: []
+  /name:
+    get:
+      tags:
+        - external
+      operationId: GetName
+      description: 'Get name entry from Naming System'
+      consumes:
```
velzevur: Since [this commit](https://github.com/aeternity/epoch/commit/7d45685f9e1faab407ec01117749c9824406c2eb) of @ThomasArts , we should not use a `consumes` clause in GETs

config/swagger.yaml
```diff
@@ -218,6 +218,31 @@ paths:
...
+      responses:
+        '200':
+          description: Name data
+          schema:
+            $ref: '#/definitions/NameEntry'
+        '404':
+          description: Name not found
+          schema:
+            $ref: '#/definitions/Error'
+      security: []
```
velzevur: Since [this commit](https://github.com/aeternity/epoch/commit/7d45685f9e1faab407ec01117749c9824406c2eb) of @ThomasArts , we should remove the `security` clause

apps/aens/src/aens.erl
```diff
@@ -0,0 +1,78 @@
...
+        [Binary] ->
+            aec_base58c:safe_decode(Type, Binary);
+        [RegistrarNamespace|_Namespace] ->
+            %% When we have more registrars, pick faster and still gov friendly structure
+            %% If we allow to purchase registrars, this will live in regular tree
+            case [RN || RN <- aec_governance:name_registrars(), RN =:= RegistrarNamespace] of
+                [] ->
+                    {error, registrar_unknown};
+                _ ->
+                    {ok, StateTrees} = aec_conductor:top_state_trees(),
```
lucafavatella: May I suggest not to define `top_state_trees/0` in `aec_conductor` but rather defining `aec_conductor:get_name/1` (similarly to `aec_conductor:get_account/1`)?

apps/aens/src/aens.erl
```diff
@@ -0,0 +1,78 @@
...
+                            end;
+                        {error, _} = Err ->
+                            Err
+                    end
+            end
+    end.
+
+-spec get_name_entry(binary()) -> {ok, map()} | {error, atom()}.
+get_name_entry(Name) ->
+    {ok, StateTrees} = aec_conductor:top_state_trees(),
```
lucafavatella: May I suggest not to define `top_state_trees/0` in `aec_conductor` but rather defining `aec_conductor:get_name/1` (similarly to `aec_conductor:get_account/1`)?

apps/aecore/src/aec_governance.erl
```diff
@@ -29,3 +35,24 @@ max_txs_in_block() ->
...
+    50000.
+
+name_protection_period() ->
+    2016.
+
+name_claim_preclaim_delta() ->
+    1.
+
+name_registrars() ->
+    [<<"aet">>, <<"test">>].
```
lucafavatella: From https://github.com/aeternity/protocol/blob/da3a79f9399f73f78833bb74f51ddd66b49ac647/AENS.md#registrars :
> `.aet` will be exclusively available on the mainnet and `.test` will be exclusive to the testnet.

So this shall rather read some kind of reference data (or hardcode different value according to ifdef).

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -139,6 +139,25 @@ handle_request('GetAccountBalance', Req, _Context) ->
             end
     end;
 
+handle_request('GetName', Req, _Context) ->
+    Name = case maps:get('name', Req) of
+               undefined ->
```
lucafavatella: Would it be possible to enrich the [Swagger spec query parameter](https://github.com/aeternity/epoch/pull/607/files#diff-89b8ec7c2f28f3e3c7d21ca9bf070bb7R223) with property [`required` true](https://github.com/OAI/OpenAPI-Specification/blob/55a4c6d8498b24faa8529b386f04bfb47a3ecfa5/versions/2.0.md#fixed-fields-7), and hardcode expectation here on `maps:get/2` to always return binary - relying on Swagger codegen to return 400? Cc @velzevur @ThomasArts

apps/aens/src/aens.erl
```diff
@@ -0,0 +1,78 @@
...
+    NSTree = aec_trees:ns(StateTrees),
+    case get_name(Name, NSTree) of
+        {ok, #{<<"name">>     := Name,
+               <<"name_ttl">> := TTL,
+               <<"pointers">> := Pointers}} ->
+            {ok, #{<<"name">>     => Name,
+                   <<"name_ttl">> => TTL,
+                   <<"pointers">> => jsx:encode(Pointers)}};
```
lucafavatella: Maybe [Swagger schema](https://github.com/aeternity/epoch/pull/607/files#diff-89b8ec7c2f28f3e3c7d21ca9bf070bb7R223) shall define this field as object/array (rather than string), and JSON encoding delegated to Swagger glue code?

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -143,6 +142,136 @@ handle_request('PostOracleUnsubscribe', _Req, _Context) ->
...
+              <<"fee">>       := Fee} = NameUpdateTxObj,
+            case aec_base58c:safe_decode(name, NameHash) of
+                {ok, DecodedNameHash} ->
+                    {ok, UpdateTx} =
+                        aens_update_tx:new(
+                          #{account   => PubKey,
+                            nonce     => Nonce,
+                            name_hash => DecodedNameHash,
+                            name_ttl  => NameTTL,
+                            pointers  => jsx:decode(Pointers,[{labels, atom}]),
```
lucafavatella: I understand this is an internal API, though did you consider not creating atoms?

apps/aens/src/aens_names.erl
```diff
@@ -0,0 +1,132 @@
...
+     #{<<"expires">>  := Expires},
+     #{<<"status">>   := Status},
+     #{<<"ttl">>      := TTL},
+     #{<<"pointers">> := Pointers}] = List,
+    #name{hash     = Hash,
+          owner    = Owner,
+          expires  = Expires,
+          status   = binary_to_existing_atom(Status, utf8),
+          ttl      = TTL,
+          pointers = jsx:decode(Pointers,[{labels, atom}])}.
```
lucafavatella: This looks like deserializing from state trees, and state trees in long chain may contain a lot of different labels. Could we avoid keeping them as atoms?

I assume labels are not in an enumeration. And if they are in an enumeration then we can use existing atoms only.

apps/aens/src/aens_update_tx.erl
```diff
@@ -0,0 +1,201 @@
...
+             #{<<"hash">>     := NameHash},
+             #{<<"name_ttl">> := NameTTL},
+             #{<<"pointers">> := Pointers},
+             #{<<"ttl">>      := TTL},
+             #{<<"fee">>      := Fee}]) ->
+    #ns_update_tx{account   = AccountPubKey,
+                  nonce     = Nonce,
+                  name_hash = NameHash,
+                  name_ttl  = NameTTL,
+                  pointers  = jsx:decode(Pointers,[{labels, atom}]),
```
lucafavatella: This looks deserialized from network on external API: could we avoid creating atoms?

gorillainduction:
I looked at the state trees handling, and it looks fine. It needs some rewriting for the db backend for MPT, but I can merge that after this is merged and fix that.

This is one big PR ;-)

apps/aens/src/aens.erl
```diff
@@ -0,0 +1,78 @@
...
+
+%% API
+-export([resolve/2,
+         get_name_entry/1]).
+
+%%%===================================================================
+%%% Types
+%%%===================================================================
+
+-define(LABEL_SEPARATOR, <<".">>).
```
lucafavatella: Could you please add comment that this must not be in alphabet of base58?

apps/aens/src/aens.erl
```diff
@@ -0,0 +1,78 @@
...
+
+%%%===================================================================
+%%% API
+%%%===================================================================
+
+-spec resolve(atom(), binary()) -> {ok, binary()} | {error, atom()}.
```
lucafavatella: Could you please specify further the type of the `Type` argument from `atom()` to `account_pubkey|oracle_pubkey`? It clarifies intended usage to the reader.

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -16,8 +16,7 @@ handle_request('PostSpendTx', #{'SpendTx' := SpendTxObj}, _Context) ->
             #{<<"recipient_pubkey">> := EncodedRecipientPubkey,
               <<"amount">>           := Amount,
               <<"fee">>              := Fee} = SpendTxObj,
-            case aec_base58c:safe_decode(
-                   account_pubkey, EncodedRecipientPubkey) of
+            case aens:resolve(account_pubkey, EncodedRecipientPubkey) of
```
lucafavatella: May be worth an integration of the description in the corresponding field in swagger.yaml

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -77,7 +76,7 @@ handle_request('PostOracleQueryTx', #{'OracleQueryTx' := OracleQueryTxObj}, _Con
               <<"fee">>           := Fee} = OracleQueryTxObj,
             QueryTTLType = binary_to_existing_atom(maps:get(<<"type">>, QueryTTL), utf8),
             QueryTTLValue= maps:get(<<"value">>, QueryTTL),
-            case aec_base58c:safe_decode(oracle_pubkey, EncodedOraclePubkey) of
+            case aens:resolve(oracle_pubkey, EncodedOraclePubkey) of
```
lucafavatella: May be worth an integration of the description in the corresponding field in swagger.yaml

apps/aens/src/aens.erl
```diff
@@ -0,0 +1,78 @@
...
+resolve(Type, Binary) ->
+    case lists:reverse(binary:split(Binary, ?LABEL_SEPARATOR, [global, trim])) of
```
lucafavatella: I see the `trim` option makes the last `.` ignored: may you please add a comment?
```
$ erl
Erlang/OTP 20 [erts-9.2] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false]

Eshell V9.2  (abort with ^G)
1> binary:split(<<"a.test">>, <<".">>, [global, trim]).
[<<"a">>,<<"test">>]
2> binary:split(<<"a..test">>, <<".">>, [global, trim]).
[<<"a">>,<<>>,<<"test">>]
3> binary:split(<<"a.test.">>, <<".">>, [global, trim]).
[<<"a">>,<<"test">>]
4> binary:split(<<".a.test">>, <<".">>, [global, trim]).
[<<>>,<<"a">>,<<"test">>]
5> binary:split(<<"a.test.">>, <<".">>, [global]).
[<<"a">>,<<"test">>,<<>>]
6> binary:split(<<"a.test. ">>, <<".">>, [global, trim]).
[<<"a">>,<<"test">>,<<" ">>]
...
8> binary:split(<<"">>, <<".">>, [global, trim]).
[]
9> binary:split(<<".">>, <<".">>, [global, trim]).
[]
10> binary:split(<<".a">>, <<".">>, [global, trim]).
[<<>>,<<"a">>]
11> binary:split(<<"a.">>, <<".">>, [global, trim]).
[<<"a">>]
```

apps/aens/src/aens.erl
```diff
@@ -0,0 +1,78 @@
...
+            %% If we allow to purchase registrars, this will live in regular tree
+            case [RN || RN <- aec_governance:name_registrars(), RN =:= RegistrarNamespace] of
+                [] ->
+                    {error, registrar_unknown};
+                _ ->
+                    {ok, StateTrees} = aec_conductor:top_state_trees(),
+                    NameTree = aec_trees:ns(StateTrees),
+                    case get_name(Binary, NameTree) of
+                        {ok, #{<<"pointers">> := Pointers}} ->
+                            case proplists:get_value(Type, Pointers) of
```
lucafavatella: I see this point relies on `Pointers` having atoms as keys - though this point could be adapted easily.

config/sys.config
```diff
@@ -19,7 +19,7 @@
       {password, <<"secret">>},
       {db_path, "."},
       {persist, false},
-      {aec_pow_cuckoo, {"mean28s-generic", "-t 5", 28}}
+      {aec_pow_cuckoo, {"mean16s-generic", "-t 5", 16}}
```
lucafavatella: Leftover from local testing?

apps/aecore/src/aec_governance.erl
sennui: @lucafavatella this is incorrect spec. We need to correct it.

apps/aehttp/src/aehttp_dispatch_int.erl
sennui: @lucafavatella I went with atoms, to match atoms in internal in the dispatcher. It is an injection verctor, true. We should refactor.

apps/aens/src/aens_names.erl
sennui: The header file shows that currently we support only one atom. So, yes, it looks like we need to migrate do binary

apps/aens/src/aens.erl
sennui: yes, this is to avoid potential user's DNS habits (where the last dot is correct)

config/sys.config
sennui: yes, thanks

apps/aens/src/aens_claim_tx.erl
```diff
@@ -0,0 +1,210 @@
...
+origin(#ns_claim_tx{account = AccountPubKey}) ->
+    AccountPubKey.
+
+-spec check(claim_tx(), trees(), height()) -> {ok, trees()} | {error, term()}.
+check(#ns_claim_tx{account = AccountPubKey, nonce = Nonce,
+                   fee = Fee, name = Name, name_salt = NameSalt}, Trees, Height) ->
+    %% TODO: Maybe include burned fee in tx fee. To do so, mechanism determining
+    %% which part of fee goes to miner and what gets burned is needed.
+    %% One option is to change tx:fee/1 callback to tx:fee_for_miner/1.
+    BurnedFee = aec_governance:name_claim_burned_fee(),
```
lucafavatella: From https://github.com/aeternity/protocol/blob/da3a79f9399f73f78833bb74f51ddd66b49ac647/AENS.md#mechanisms :
> Fees are the main mechanism to discourage spam and squatting. This initial version will burn the governance fee, in order to enable us to change this behaviour in the future. If we start out by giving this fee to miners, they will be very hesitant to accept any changes, which will impact their income negatively. Thus starting out with burning could allow us an easier path for future update to the fee structures.

From https://github.com/aeternity/protocol/blob/da3a79f9399f73f78833bb74f51ddd66b49ac647/AENS.md#claim :
> Note that the fee here is a distinct fee from the normal transaction fee.

From the above I had implied that the claim tx was meant to contain two fees: normal fee for miner, and fee-to-be-burned (that https://github.com/aeternity/protocol/blob/da3a79f9399f73f78833bb74f51ddd66b49ac647/AENS.md#claim calls simply as `fee`).

So the comment in the code above appears to be out of sync with protocol doc - at least in my understanding.

lucafavatella:
Travis CI package (?) job [failed](https://travis-ci.org/aeternity/epoch/jobs/335560167#L1403):
```
make[1]: Entering directory `/home/travis/build/aeternity/epoch/py'
Traceback (most recent call last):
  File "tests/release.py", line 14, in <module>
    import swagger_client
  File "/home/travis/build/aeternity/epoch/py/tests/swagger_client/__init__.py", line 19, in <module>
    from swagger_client.api.external_api import ExternalApi
  File "/home/travis/build/aeternity/epoch/py/tests/swagger_client/api/__init__.py", line 6, in <module>
    from swagger_client.api.external_api import ExternalApi
  File "/home/travis/build/aeternity/epoch/py/tests/swagger_client/api/external_api.py", line 21, in <module>
    from swagger_client.api_client import ApiClient
  File "/home/travis/build/aeternity/epoch/py/tests/swagger_client/api_client.py", line 27, in <module>
    import swagger_client.models
  File "/home/travis/build/aeternity/epoch/py/tests/swagger_client/models/__init__.py", line 51, in <module>
    from swagger_client.models.single_tx_hash import SingleTxHash
  File "/home/travis/build/aeternity/epoch/py/tests/swagger_client/models/single_tx_hash.py", line 20, in <module>
    from swagger_client.models.tx_hash_or_object import TxHashOrObject  # noqa: F401,E501
ImportError: No module named tx_hash_or_object
make[1]: *** [release-test] Error 1
```

hanssv:
Except for the build failure it LGTM!

PR #613, opened 2018-01-22
PT-154171539 boilerplate for create contract transaction

UlfNorell:
This PR adds the appropriate fields to the create contract transaction, and sets up the infrastructure for building and applying such transactions.

Note that this doesn't close the ticket, this part is pure boilerplate and doesn't do anything interesting with the data in the transaction. That will be added once we have contract state trees in place.


UlfNorell:
I guess the swagger stuff goes in [PT-154171508](https://www.pivotaltracker.com/story/show/154171508) (Contract HTTP APIs)?

PR #614, opened 2018-01-22
WS integration test infrastructure

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/154081975)
### Description
Added an utils file (`apps/aehttp/test/aehttp_ws_test_utils.erl`) for handling all the client's websocket logic. This includes some wrappers around the WS client - since by nature WS are async and this does not fit really well in tests - the functions that needed to be synchronous are sync, and the others are left async. Currently it has the support for:

* start a websocket client (sync)

* stop a websocket client (async)

* send message (async)

* subscribe/unsubscribe the test process for an event (described bellow) (sync)

* wait for a certain event (sync)

### Arguments
The main concern had been not letting race conditions inside tests. The work flow is the following:
1. Subscribe the test for certain events
2. Do something
3. Wait for the expected event

Since 1 and 3 are sync calls to the WS client process, this guarantees us that no events are lost due to races.

### Example
from ws_block_mined test
```
    ok = ?WS:subscribe_for_event(ConnPid, miner, mined_block),
    aecore_suite_utils:mine_blocks(aecore_suite_utils:node_name(?NODE), 1),
    {ok, #{<<"height">> := _Height, <<"hash">> := _Hash}} = ?WS:wait_for_event(miner, mined_block),
    ok = ?WS:unsubscribe_for_event(ConnPid, miner, mined_block),
```
We subscribe the test process to event with origin `miner` and action `mined_block`, then we mine a block and wait for the message from the WS process to arrive (default timeout is 5s).

### TODO
~~Move the starting and stopping of a WS client (being boiler plate) to the `init_per_testcase/2` and `init_per_testcase/2` respectively~~

- [x] A WS connection can subscribe to certain events (for example _this_ oracle events). This functionality is not implemented yet in the test client. This would also require some renaming - we need to distinguish test subscribes to expected events with WS connection's subscriptions. Probably test's subscribtions would be renamed to registering or something like that. Comments are welcome.
- [x] Write some WS tests to cover the current functionality

### DIFF
--- a/apps/aehttp/test/aehttp_integration_SUITE.erl
+++ b/apps/aehttp/test/aehttp_integration_SUITE.erl
@@ -1353,24 +1353,74 @@ ws_block_mined(_Config) ->
     ok = aehttp_ws_test_utils:stop(ConnPid),
     ok.
 
+%% Currently the websockets are a queue: they have a maximim amount of
+%% acceptors. Every WS trying to connect after all acceptoprs are used
+%% goes into a queue. When the queue is full - the node starts rejecting
+%% any new incoming WS connections
 ws_refused_on_limit_reached(_Config) ->
-    AllowedWsCount = rpc(aeu_env, user_config_or_env,
+    %% maximum amount of acceptors
+    MaxWsCount = rpc(aeu_env, user_config_or_env,
                           [[<<"websocket">>, <<"internal">>, <<"acceptors">>],
                           aehttp, [internal, websocket, handlers], 10]), 
+    %% Maximum WS connections hanging in the queue
+    WSQueueSize = rpc(aehttp_app, ws_handlers_queue_max_size, []),
+    WSDieTimeout = 1000,
+    ct:log("Websocket acceptors: ~p, websocket acceptor's queue size ~p",
+           [MaxWsCount, WSQueueSize]),
     %% assert no WS running on the node
     0 = open_websockets_count(), 
+    %% start as many WS as needed to consume all acceptors
     WSPids =
         lists:map(
             fun(_) ->
               {ok, ConnPid} = ws_start_link(),
               ConnPid
             end,
-            lists:seq(1, AllowedWsCount)),
-    AllowedWsCount = open_websockets_count(), 
-    {error, timeout} = ws_start_link(),
+            lists:seq(1, MaxWsCount)),
+    %% assert expectation for amount of connected WSs
+    MaxWsCount = open_websockets_count(), 
+    WaitingPids =
+        lists:map(
+            fun(_) ->
+                %% try to connect a WS client; assert it does not connect and
+                %% is waiting
+                {error, still_connecting, WaitingPid} = ws_start_link(),
+                MaxWsCount = open_websockets_count(), 
+                WaitingPid
+            end,
+            lists:seq(1, WSQueueSize)),
+
+    %% Now both the acceptopr pool and queue are full. Try to connect a new WS
+    %% client and assert it fails
+    {error, rejected} = ws_start_link(),
+
+    %% split currently connected WS clients into two groups: first with as
+    %% many as there are WS clients waiting in the queue and all the rest in a
+    %% seperate list. The first group would be used for stopping WS clients
+    %% one by one while validating that one of the waiting WS clients connects
+    %% for each one stopped
+    {FirstWSsPids, OtherWSsPids} = lists:split(WSQueueSize, WSPids),
     lists:foreach(
-        fun(ConnPid) -> ?WS:stop(ConnPid) end,
-        WSPids),
+        fun(Pid) ->
+            %% stop one
+            ?WS:stop(Pid),
+            %% another one connects in its place and it doesn't matter which one
+            {ok, some_pid} = ?WS:wait_for_connect(some_pid),
+            %% total amount of connected WS clients is still the maximum
+            MaxWsCount = open_websockets_count()
+        end,
+        FirstWSsPids),
+    %% cleanup
+    lists:foreach(fun ?WS:stop/1, OtherWSsPids),
+    timer:sleep(100), % wait for all of them to die out
+    {ok, WSQueueSize} =
+        aec_test_utils:wait_for_it_or_timeout(fun open_websockets_count/0,
+                                              WSQueueSize, WSDieTimeout),
+    lists:foreach(fun ?WS:stop/1, WaitingPids),
+    timer:sleep(100), % wait for all of them to die out
+    {ok, 0} =
+        aec_test_utils:wait_for_it_or_timeout(fun open_websockets_count/0,
+                                              0, WSDieTimeout),
     ok.
 
 %% ============================================================
--- a/apps/aehttp/test/aehttp_integration_SUITE.erl
+++ b/apps/aehttp/test/aehttp_integration_SUITE.erl
@@ -94,9 +94,17 @@
     block_txs_list_by_hash_invalid_range/1
    ]).
 
+%% internal endpoints
+-export(
+   [ws_get_genesis/1,
+    ws_block_mined/1,
+    ws_refused_on_limit_reached/1
+   ]).
+
 -include_lib("common_test/include/ct.hrl").
 -define(NODE, dev1).
 -define(DEFAULT_TESTS_COUNT, 5).
+-define(WS, aehttp_ws_test_utils).
 
 all() ->
     [
@@ -106,7 +114,9 @@ all() ->
 groups() ->
     [
      {all_endpoints, [sequence], [{group, external_endpoints},
-                                  {group, internal_endpoints}]},
+                                  {group, internal_endpoints},
+                                  {group, websocket}
+                                  ]},
      {external_endpoints, [sequence], 
       [
         % pings
@@ -182,7 +192,14 @@ groups() ->
         block_txs_list_by_hash,
         block_txs_list_by_height_invalid_range,
         block_txs_list_by_hash_invalid_range
+      ]},
+     {websocket, [sequence],
+      [
+       ws_get_genesis,
+       ws_block_mined,
+       ws_refused_on_limit_reached
       ]}
+
     ].
 
 suite() ->
@@ -209,36 +226,21 @@ end_per_suite(_Config) ->
 
 init_per_group(all_endpoints, Config) ->
     Config;
-init_per_group(Group, Config) when Group =:= external_endpoints
-                            orelse Group =:= internal_endpoints ->
+init_per_group(_Group, Config) ->
     aecore_suite_utils:start_node(?NODE, Config),
     aecore_suite_utils:connect(aecore_suite_utils:node_name(?NODE)),
-    [{start_node_per_tc, false} | Config];
-init_per_group(_, Config) ->
-    [{start_node_per_tc, true} | Config].
+    Config.
 
+end_per_group(all_endpoints, _Config) ->
+    ok;
 end_per_group(_Group, Config) ->
-    case proplists:get_value(start_node_per_tc, Config, false) of
-        true -> pass;
-        false -> aecore_suite_utils:stop_node(?NODE, Config)
-    end,
+    aecore_suite_utils:stop_node(?NODE, Config),
     ok.
 
 init_per_testcase(_Case, Config) ->
-    case proplists:get_value(start_node_per_tc, Config, true) of
-        true ->
-            aecore_suite_utils:start_node(?NODE, Config),
-            aecore_suite_utils:connect(aecore_suite_utils:node_name(?NODE));
-        false -> pass 
-    end,
-    ct:log("testcase pid: ~p", [self()]),
     [{tc_start, os:timestamp()}|Config].
 
 end_per_testcase(_Case, Config) ->
-    case proplists:get_value(start_node_per_tc, Config, true) of
-        false -> pass;
-        true -> aecore_suite_utils:stop_node(?NODE, Config)
-    end,
     Ts0 = ?config(tc_start, Config),
     ct:log("Events during TC: ~p", [[{N, aecore_suite_utils:all_events_since(N, Ts0)}
                                      || {_,N} <- ?config(nodes, Config)]]),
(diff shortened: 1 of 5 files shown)


apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -1309,6 +1323,34 @@ block_txs_list_by_hash_invalid_range(_Config) ->
...
     ok.
 
+%% ============================================================
+%% Websocket tests 
+%% ============================================================
+
+ws_get_genesis(_Config) ->
+    {Host, Port} = ws_host_and_port(),
+    {ok, ConnPid} = ?WS:start(Host, Port),
+    ?WS:send(ConnPid, chain, get, #{height => 0}),
```
lucafavatella: Could you please ensure this is [documented](https://github.com/aeternity/protocol/blob/68b27350f157e05c6930d037a626695e6a4fad3b/epoch/api/README.md#websocket-api-definition) ?
(It may require moving out of the oracle page some more generic portions.)

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -1309,6 +1323,34 @@ block_txs_list_by_hash_invalid_range(_Config) ->
...
+    ExpectedBlockMap = maps:remove(<<"data_schema">>, BlockMap),
+    Block = ExpectedBlockMap,
+    ok = ?WS:unsubscribe_for_event(ConnPid, chain, requested_data),
+    aehttp_ws_test_utils:stop(ConnPid),
+    ok.
+
+ws_block_mined(_Config) ->
+    {Host, Port} = ws_host_and_port(),
+    {ok, ConnPid} = ?WS:start(Host, Port),
+    ok = ?WS:subscribe_for_event(ConnPid, miner, mined_block),
```
lucafavatella: Could you please ensure this is [documented](https://github.com/aeternity/protocol/blob/68b27350f157e05c6930d037a626695e6a4fad3b/epoch/api/README.md#websocket-api-definition) ?
(It may require moving out of the oracle page some more generic portions.)

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -0,0 +1,155 @@
...
+         websocket_info/3,
+         websocket_terminate/3
+        ]).
+
+start(Host, Port) ->
+    crypto:start(),
+    ssl:start(),
+    WsAddress = "ws://" ++ Host ++ ":" ++ integer_to_list(Port) ++ "/websocket",
+    ct:log("connecting to ~p", [WsAddress]),
+    {ok, Pid} = websocket_client:start_link(WsAddress, ?MODULE, self()),
```
lucafavatella: This is actually linking. Could you please consider renaming from start to start_link ?

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -0,0 +1,155 @@
...
+    ok = wait_for_event(websocket, connected),
+    {ok, Pid}.
+    
+
+stop(ConnPid) ->
+    ct:log("Websocket disconnecting"),
+    ConnPid ! stop.
```
uwiger: Perhaps you could monitor `ConnPid` and wait for a `'DOWN'` message?

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -0,0 +1,155 @@
...
+         onconnect/2,
+         ondisconnect/2,
+         websocket_handle/3,
+         websocket_info/3,
+         websocket_terminate/3
+        ]).
+
+start(Host, Port) ->
+    crypto:start(),
+    ssl:start(),
```
lucafavatella: I see these are application:start of crypto and ssl. Shall these two lines be removed from here and moved to aehttp_integration_SUITE:init_per_group websocket? (And maybe corresponding stops in end_per_group.)
Please consider (I am not sure what current design in tests is):
* Matching return value (so to fail at correct point if e.g. crypto cannot start);
* Using start/ensure_started/ensure_all_started and stop.

BTW I see these crypto an ssl are [dependencies of websocket_client](https://github.com/sanmiguel/websocket_client/blob/2b6228211389fbd0b1694e5b71e33da48d98d2b7/src/websocket_client.app.src#L5)

apps/aehttp/src/ws_int_dispatch.erl
```diff
@@ -28,7 +28,7 @@ is_valid(_, _, _) ->
     true.
 
 do_execute(chain, get, QueryPayload) ->
-    #{<<"type">> := Type} = QueryPayload,
+    Type = maps:get(<<"type">>, QueryPayload, <<"block">>),
```
lucafavatella: Why a default? I would rather expect to keep this code simpler not having the default, documenting what is needed (in protocol repo) and requiring client to comply to documented schema.
(I may not understand the problem here so please correct me if wrong.)

apps/aehttp/test/aehttp_ws_test_utils.erl
lucafavatella: I see - originally taken from https://github.com/sanmiguel/websocket_client/tree/2b6228211389fbd0b1694e5b71e33da48d98d2b7#usage

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -0,0 +1,155 @@
...
+unsubscribe_for_event(ConnPid, Origin, Action) ->
+    Event = {Origin, Action},
+    ConnPid ! {unsubscribe, self(), Event},
+    ok = wait_for_event(unsubscribed, Event, 1000),
+    ok.
+    
+
+-spec wait_for_event(atom(), atom()) -> ok | {ok, map()} | timeout.
+wait_for_event(Origin, Action) ->
+    wait_for_event(Origin, Action, 5000).
```
lucafavatella: Macros for timeouts?

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -0,0 +1,155 @@
...
+subscribe_for_event(ConnPid, Origin, Action) ->
+    Event = {Origin, Action},
+    ConnPid ! {subscribe, self(), Event},
+    ok = wait_for_event(subscribed, Event, 1000),
+    ok.
+
+unsubscribe_for_event(ConnPid, Origin, Action) ->
+    Event = {Origin, Action},
+    ConnPid ! {unsubscribe, self(), Event},
+    ok = wait_for_event(unsubscribed, Event, 1000),
```
lucafavatella: Macros for timeouts?

apps/aehttp/test/aehttp_ws_test_utils.erl
lucafavatella: It does not matter much, though websocket_client:start_link types 3rd arg as list.

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -0,0 +1,155 @@
...
+         websocket_terminate/3
+        ]).
+
+start(Host, Port) ->
+    crypto:start(),
+    ssl:start(),
+    WsAddress = "ws://" ++ Host ++ ":" ++ integer_to_list(Port) ++ "/websocket",
+    ct:log("connecting to ~p", [WsAddress]),
+    {ok, Pid} = websocket_client:start_link(WsAddress, ?MODULE, self()),
+    ok = wait_for_event(websocket, connected),
```
lucafavatella: I see - on init argument (self()) is automatically subscribed to waiting_connected, and onconnect subscription is removed and message is sent to caller. This may lead to unexpected message in mailbox of caller of aehttp_ws_test_utils:start (in case of timeout/slowness/race). It would be nice to have a line of comment documenting this potential message in.

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -0,0 +1,155 @@
...
+    {ok, Subs};
+websocket_handle({text, MsgBin}, _ConnState, Subs) ->
+    Msg = jsx:decode(MsgBin, [return_maps]), 
+    ct:log("Received msg ~p~n", [Msg]),
+    Origin = binary_to_existing_atom(maps:get(<<"origin">>, Msg), utf8),
+    Action = binary_to_existing_atom(maps:get(<<"action">>, Msg), utf8),
+    Payload = maps:get(<<"payload">>, Msg, []),
+    Subscribed = get_subscribed({Origin, Action}, Subs),
+    lists:foreach(
+        fun(Pid) -> inform_subscribed(Pid, Origin, Action, Payload) end,
```
lucafavatella: I see this is a real asynchronous notification.

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -0,0 +1,155 @@
...
+    Msg =
+        case Payload == [] orelse Payload == #{} of
+            true -> Msg0;
+            false -> maps:put(payload, Payload, Msg0)
+        end,
+    ct:log("Sending to server ~p", [Msg]),
+    websocket_client:cast(ConnPid, {text, jsx:encode(Msg)}),
+    ok.
+
+subscribe_for_event(ConnPid, Origin, Action) ->
```
lucafavatella: Could you please add a line of comment mentioning that caller will receive in mailbox messages looking like {websocket_event, ...} (or similar - better if with a type).

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -0,0 +1,155 @@
...
+        true -> ct:log("No subscribers for this event");
+        false -> pass
+    end,
+    {ok, Subs}.
+
+websocket_info(stop, _ConnState, _Subs) ->
+    {close, <<>>, "stopped"};
+websocket_info({subscribe, SubscribedPid, Event}, _ConnState, Subs0) ->
+    Subs = put_subscription(SubscribedPid, Event, Subs0),
+    inform_subscribed(SubscribedPid, subscribed, Event),
```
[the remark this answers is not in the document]
lucafavatella: > Would it be possible to use websocket_client:send

It looks no, as it frames the message in `{send, _}`

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -0,0 +1,155 @@
...
+    {ok, Subs};
+websocket_info({unsubscribe, SubscribedPid, Event}, _ConnState, Subs0) ->
+    Subs = delete_subscription(SubscribedPid, Event, Subs0),
+    inform_subscribed(SubscribedPid, unsubscribed, Event),
```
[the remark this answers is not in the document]
lucafavatella: > Would it be possible to use websocket_client:send

It looks no, as it frames the message in `{send, _}`

velzevur:
I think I've addressed all comments. PR for protocol update to current state of WS chain interaction can be found [here](https://github.com/aeternity/protocol/pull/21)
I'm proceeding with renaming of the test's subscription to certain events to `register`-ing

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -147,7 +156,6 @@ groups() ->
       ]},
      {internal_endpoints, [sequence], 
       [
-        broken_spend_tx,
```
lucafavatella: Spurious change?

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -209,36 +223,21 @@ end_per_suite(_Config) ->
 
 init_per_group(all_endpoints, Config) ->
     Config;
-init_per_group(Group, Config) when Group =:= external_endpoints
-                            orelse Group =:= internal_endpoints ->
+init_per_group(_Group, Config) ->
     aecore_suite_utils:start_node(?NODE, Config),
     aecore_suite_utils:connect(aecore_suite_utils:node_name(?NODE)),
-    [{start_node_per_tc, false} | Config];
```
lucafavatella: I see this is a test simplification. Ok.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -209,36 +224,21 @@ end_per_suite(_Config) ->
 
 init_per_group(all_endpoints, Config) ->
     Config;
-init_per_group(Group, Config) when Group =:= external_endpoints
-                            orelse Group =:= internal_endpoints ->
+init_per_group(_Group, Config) ->
```
lucafavatella: I expect that this change, because of the group-nesting group `all_endpoints`, makes this function body to be run twice (without the corresponding `end_per_group` having been run already). So the requirement on this function body is that it is idempotent. It is idempotent at the moment, as the node start can be run again and it leaves the node running, and and the aehttp app_started event is recorded in the proxy. Though for avoiding future breakages I suggest to either:
A. Restore start of node only for specified leaf groups (external_endpoints, internal_endpoints, websockets); or
B. Add comment that body here needs to be idempotent.

I recommend A.

lucafavatella: > A. Restore start of node only for specified leaf groups (external_endpoints, internal_endpoints, websockets); or

I had not noticed that clause before already handles `all_endpoints`, so please ignore me on this. Thanks for pointing that out.

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -0,0 +1,165 @@
...
+    ct:log("Sending to server ~p", [Msg]),
+    websocket_client:cast(ConnPid, {text, jsx:encode(Msg)}),
+    ok.
+
+%% when an WS event occours, the registered process will receive a message in
+%% the mailbox. Message structure will be:
+%% {websocket_event, Origin::atom(), Action::atom()}
+%% {websocket_event, Origin::atom(), Action::atom(), Payload::map()}
+%% where the Origin and Action are the same as the ones for which the process
+%% had registered. This message is consumed by wait_for_event/2,3
```
lucafavatella: This means a test process cannot start two WS clients and discriminate the received messages by source WS connection. Though not needed for current tests so ok.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -1309,6 +1311,68 @@ block_txs_list_by_hash_invalid_range(_Config) ->
...
+    0 = open_websockets_count(), 
+    WSPids =
+        lists:map(
+            fun(_) ->
+              {ok, ConnPid} = ws_start_link(),
+              ConnPid
+            end,
+            lists:seq(1, AllowedWsCount)),
+    AllowedWsCount = open_websockets_count(), 
+    {error, timeout} = ws_start_link(),
```
lucafavatella: I understand the exercised timeout is DEFAULT_SUB_TIMEOUT i.e. 1000 (1s). Ok.

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -0,0 +1,169 @@
...
+
+start_link(Host, Port) ->
+    WsAddress = "ws://" ++ Host ++ ":" ++ integer_to_list(Port) ++ "/websocket",
+    ct:log("connecting to ~p", [WsAddress]),
+    {ok, Pid} = websocket_client:start_link(WsAddress, ?MODULE, self()),
+    case wait_for_event(websocket, connected, ?DEFAULT_SUB_TIMEOUT) of
+        ok ->
+            {ok, Pid};
+        timeout ->
+            {error, timeout}
```
lucafavatella: When this happen, does WS client slow to connect (potentially for server that ran out of acceptors) stay lingering expecting to connect? Hence eventually consuming connection? Hence eventually potentially putting unexpected message in mailbox of test process?

velzevur:
Enchanced the test for verifying the maximum amount of connected WS clients (and that we actually queue the next X incoming ones)

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -1353,24 +1353,65 @@ ws_block_mined(_Config) ->
...
+    %% for each one stopped
+    {FirstWSsPids, OtherWSsPids} = lists:split(WSQueueSize, WSPids),
     lists:foreach(
-        fun(ConnPid) -> ?WS:stop(ConnPid) end,
-        WSPids),
+        fun(Pid) ->
+            %% stop one
+            ?WS:stop(Pid),
+            %% another one connects in its place and it doesn't matter which one
+            {ok, some_pid} = ?WS:wait_for_connect(some_pid),
```
lucafavatella: Can this `some_pid` work? Maybe SomePid?

lucafavatella: oh not used - I see...

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -1353,24 +1353,65 @@ ws_block_mined(_Config) ->
...
+            %% total amount of connected WS clients is still the maximum
+            MaxWsCount = open_websockets_count()
+        end,
+        FirstWSsPids),
+    %% cleanup
+    lists:foreach(fun ?WS:stop/1, OtherWSsPids),
+    WSQueueSize = open_websockets_count(),
```
lucafavatella: Started MaxWsCount+WSQueueSize. Then stopped WSQueueSize. I would expect MaxWsCount left?

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -109,9 +115,12 @@ onconnect(_WSReq, Regs) ->
     inform_registered(WaitingPid, websocket, connected),
     {ok, Regs1}.
 
+
+ondisconnect({error, {400, <<"Bad Request">>}}, Regs) ->
```
lucafavatella: When does this happen? Worth documenting in protocol repo?

apps/aehttp/test/aehttp_integration_SUITE.erl
lucafavatella: [CI failed](https://travis-ci.org/aeternity/epoch/jobs/333152759), there may be some miscalculation here indeed (actually test failed 2 lines below...).

apps/aehttp/test/aehttp_ws_test_utils.erl
velzevur: When all Websocket acceptors are consumed and the queue for incoming new Websockets is full, we start rejecting new connections. This is the standart `cowboy` message being send to the request. It is not documented in the protocol repo, I'm not sure if it would add more clarity or the opposite.

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: Indeed started MaxWsCount+WSQueueSize
Then stopped `WSQueueSize` and `OtherWSsPids` (where OtherWSsPids = MaxWsCount - WSQueueSize, the WS-s that were not stopped while stopping the first WSQueueSize number of WS-s). Since WSQueueSize + OtherWSsPids = MaxWsCount, we have actually stopped MaxWsCount amount of WS-s, leaving us with just WSQueueSize left. I think the test assumption is correct

apps/aehttp/test/aehttp_ws_test_utils.erl
lucafavatella: mmm it could surprise some clients so I would consider it worth documenting it

PR #616, opened 2018-01-23
Add proofs to MPTs

gorillainduction:
https://www.pivotaltracker.com/n/projects/2124891/stories/154515354

Also  tested with quickcheck.


hanssv:
Updated QuickCheck model to also test the proofs: https://github.com/Quviq/epoch-eqc

PR #617, opened 2018-01-23
Pt 154246728 introduce mnesia 2

uwiger:
Introduces a mnesia backend which takes the place of `aec_persistence` for persistent storage.

The `aec_persistence` module is kept practically unchanged, partly to enable migration, but could also possibly evolve into a form of external backup.


config/sys.config
```diff
@@ -73,6 +73,8 @@
       ]}
   ]},
 
+  {mnesia, [{dir, "data/mnesia"}]},
```
lucafavatella: I understand this directory is automatically created, right?

When wanting to restart chain with different block version, shall the directory specified here be fresh? (If yes, this shall be hooked in Ansible scripts preferably via user config.)
Current (0.5.0) approach when bumping block version (hard fork) and restarting chain is required includes amending the deployment script so that environments start from fresh DB folder (see [sample commit](https://github.com/aeternity/epoch/commit/6d0837dc287ffd2ec2d583fb4389d6052e2b38b7#diff-2ad1947fe9b869ffd23a6a86872e65e0R14) - via [wiki page on release process](https://github.com/aeternity/epoch/wiki/Release-process)).

uwiger: In the config file, we have `db_path`, but I decided not to reuse that, since it at least initially is likely to point to a place where `aec_persistence` files reside.

One could of course create `DbPath/mnesia`, but relying on anything from epoch.config is tricky until we address the start order problem (since we can't be sure that the config can be read before mnesia starts).

uwiger: So if the directory is deleted, a new schema and empty database will be created in the same location.

apps/aecore/src/aecore_app.erl
```diff
@@ -16,12 +16,15 @@ start(_StartType, _StartArgs) ->
     ok = lager:info("Starting aecore node"),
     aecore_sup:start_link().
 
+start_phase(load_database, _StartType, _PhaseArgs) ->
```
lucafavatella: How is this called? I do not see it among `start_phases` in aecore.app.src

config/sys.config
lucafavatella: > In the config file, we have db_path, but I decided not to reuse that, since it at least initially is likely to point to a place where aec_persistence files reside.

FYI In the release notes we explicitly mention whether a persisted blockchain shall not be reused.

lucafavatella: > One could of course create DbPath/mnesia, but relying on anything from epoch.config is tricky until we address the start order problem (since we can't be sure that the config can be read before mnesia starts).

Does mnesia support symbolic links? If yes, we could hardcode `data/mnesia` (or similar) in the `sys.config`, and then mention in release notes that if user wants DB somewhere else from the default, he/she needs to ensure that `data/mnesia` is a symlink to the desired location.

lucafavatella: > One could of course create DbPath/mnesia, but relying on anything from epoch.config is tricky until we address the start order problem (since we can't be sure that the config can be read before mnesia starts).

@uwiger My proposal in PR #626, please review.
IMO not getting the DB dir from user config would impact significantly the user experience.

uwiger:
Rebased and added some extras:
* Mnesia is not explicitly started from aecore, eliminating a possible race condition with the setup start hooks checking the database and persistency
* The mnesia database is now located under `{db_path}/mnesia`, if a db_path has been given as a config variable.
* If `persist = false`, and an existing database is found on disk, a new `mnesia dir` path is set, pointing to a non-existent directory (the previous mnesia dir with a millisecond timestamp appended)

apps/aecore/src/aec_db.erl
```diff
@@ -0,0 +1,298 @@
+-module(aec_db).
+
+-export([check_db/0,           % called from setup hook
+         load_database/0,      % called in aecore app start phase
```
lucafavatella: Is this not called?

apps/aecore/src/aecore_app.erl
```diff
@@ -14,14 +14,18 @@
 
 start(_StartType, _StartArgs) ->
     ok = lager:info("Starting aecore node"),
+    ok = application:ensure_started(mnesia),
```
lucafavatella: Why is aec_persistence child still started under aecore_sup ?

lucafavatella:
@uwiger Created PR #630 against this branch wrt a couple of small items.

apps/aecore/src/aec_db.erl
```diff
@@ -0,0 +1,298 @@
...
+                 end).
+
+tables(Mode) ->
+    [?TAB(aec_blocks)
+   , ?TAB(aec_headers)
+   , ?TAB(aec_chain_state)
+   , ?TAB(aec_block_state)
+   , ?TAB(aec_oracle_state)
+   , ?TAB(aec_account_state)
+   , ?TAB(aec_name_service_state)
```
lucafavatella: ... and contracts?

apps/aecore/src/aec_db.erl
lucafavatella: ... and I see `db_path` in deployment/ansible/deploy.yml not bumped. So it looks to me that the intention of this PR is not to restart chain (by not changing `db_path`) though, given that this `load_database` is not hooked anywhere (I understand), it will restart the chain - basically having `.chain` and `mnesia` side by side and completely independent. @uwiger Is that intentional?

lucafavatella:
Started [CI build of this PR merged with recent master with new trees](https://travis-ci.org/lucafavatella/epoch/builds/333298365).

apps/aecore/src/aec_db.erl
uwiger: The `load_database()` function is called from a start_phase hook in `aecore`, which, unfortunately, has not been defined in `aecore.app`. I will fix that.

lucafavatella: I saw the fix and approved. So I understand intention is loading from a .chain old style DB if found.

apps/aecore/src/aecore_app.erl
```diff
@@ -14,14 +14,21 @@
...
 start_phase(create_metrics_probes, _StartType, _PhaseArgs) ->
+    lager:debug("start_phase(create_metrics_probes, _, _)", []),
     aec_metrics:create_metrics_probes();
 start_phase(start_reporters, _StartType, _PhaseArgs) ->
+    lager:debug("start_phase(start_reporters, _, _)", []),
     aec_metrics_rpt_dest:check_config(),
     aec_metrics:start_reporters().
 
+
 stop(_State) ->
```
gorillainduction: Shouldn't we stop mnesia as well?

uwiger: There would be a point to that. In general, I would object that the only reason it's started by aecore is to respect a start dependency; otherwise, mnesia should be regarded as started in its own right. In practice, it shouldn't really matter (except that some unit tests have to explicitly stop mnesia.)

PR #618, opened 2018-01-23
PT-153778514 Initial consensus review

lucafavatella:
These is cleanup I performed while performing initial assessment of work for [PT-153778514](https://www.pivotaltracker.com/story/show/153778514).


lucafavatella:
I see two approvals, I see CI passed. I am running checks with recent master merged then I will merge.

PR #619, opened 2018-01-24
Pt 154474371 initial ring compiler

happi:
Bringing in John's compiler.
Adding preamble to get calldata on the stack.
Adding a test case to run the identity contract.


happi:
I made tickets of @UlfNorell 's comments.

PR #622, opened 2018-01-24
WIP Use mp_trees rather than gb_merkle_trees

gorillainduction:
I label this WIP since I should decide if I should await the database before proceeding.

Please review this as if the WIP is not stated.

Note that this PR is replacing the gb_merkle_trees with the functional version of aeu_mp_trees without using any of the benefits of MPTs. The hashes will be the same when a proper db connection is added, but then we can start rewriting the state handling to make some proper benefit of the MPTs.


deployment/ansible/deploy.yml
```diff
@@ -40,7 +40,7 @@
           port: 3013
       chain:
         persist: true
-        db_path: "./db15"
+        db_path: "./db16"
```
lucafavatella: master has db17, this shall be db18 (just for not forgetting)

apps/aeutils/src/aeutils.app.src
```diff
@@ -9,8 +9,7 @@
     gproc,
     jobs,
     exometer_core,
-    yamerl,
-    gb_merkle_trees
```
lucafavatella: Leftover in rebar.config and rebar.lock

gorillainduction:
Rebased, resolved conflicts, and removed gb_merkle_trees from rebar files.

lucafavatella:
Somehow CI failed https://travis-ci.org/aeternity/epoch/jobs/333260094#L1242

gorillainduction:
I had missed removing a bogus dep stated in aecore.app.src

lucafavatella:
I see both CI builds finished. I see two approvers. Merging soon.

PR #623, opened 2018-01-24
Add contract state tree to top level tree + fork chain

hanssv:
This should fix the issue seen on the integration network.

### DIFF
--- a/apps/aecontract/src/aect_state_tree.erl
+++ b/apps/aecontract/src/aect_state_tree.erl
@@ -8,7 +8,9 @@
 -module(aect_state_tree).
 
 %% API
--export([empty/0, insert_contract/2]).
+-export([ empty/0
+        , insert_contract/2
+        , root_hash/1]).
 
 -export_type([tree/0]).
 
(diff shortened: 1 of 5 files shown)


apps/aecontract/src/aect_state_tree.erl
```diff
@@ -34,3 +36,7 @@ insert_contract(_Contract, Tree) ->
     %% PLACEHOLDER
     Tree.
 
+-spec root_hash(tree()) -> {ok, aeu_mtrees:root_hash()} | {error, empty}.
+root_hash(#contract_tree{}) ->
+    {error, empty}.
```
lucafavatella: Addition of TODO comment here welcome.

lucafavatella: Ah it is empty anyway so there is not way that can be forgotten. (Other functions have PLACEHOLDER for remembering.)

lucafavatella:
I see this PR 6ca45347 is based directly on master ca11409e , so the two CI builds (push and pr) are actually redundant:
* pr CI build still needs to start
* push CI build has not completed static_analysis and package yet

hanssv:
The release notes already say this release is a chain altering release, not sure there is anything else to say?!

There is no API for getting aect_create_tx into the system, but if one would enter one (into attached shell maybe?) it would only affect the account tree since the `inser_contract` is just a placeholder.

PR #630, opened 2018-01-25
[PR on PR 617] Pt 154246728 introduce mnesia 2 - update doc

lucafavatella:


docs/docker-testnet.md
```diff
@@ -66,12 +66,12 @@ docker run -d -p 3013:3013 -e PEERS_ADDRESS_0=http://31.13.248.103:3013/ aetrnty
 
 ## Persisting Data
 
-To persist blockchain data and node keys between container runs, use [Docker volumes](https://docs.docker.com/engine/admin/volumes/volumes/). Replace `~/.aeternity/.chain` with location of your choice.
+To persist blockchain data and node keys between container runs, use [Docker volumes](https://docs.docker.com/engine/admin/volumes/volumes/). Replace `~/.aeternity/data` with location of your choice.
```
dincho: Is the default db_path = `data` now ?
If so, IMO we should also change the example configuration files

lucafavatella: Do you mean https://github.com/aeternity/epoch/blob/999c46392991310c8dfcfda83f8fda152f5e537e/apps/aeutils/test/data/epoch_full.yaml#L43 ?

uwiger: Not sure we have to change anything here. The mnesia database will place itself in whatever directory we want, and will coexist there with the old persistence files.

lucafavatella: > Not sure we have to change anything here.

@uwiger  For sure a constraint we have in `-v ~/.aeternity/data:/home/epoch/node/data` is that *with the config that the Docker image uses* the DB dir must be `/home/epoch/node/data`.

lucafavatella:
> The PR needs to be rebased ...

@uwiger Rebased.

docs/docker-testnet.md
lucafavatella: Linked this from https://www.pivotaltracker.com/story/show/154636528/comments/185105839

dincho:
Last time I checked data dir contained some other file as well. We're using docker bind mounts and this information will be obfuscated see https://docs.docker.com/engine/admin/volumes/

uwiger:
@dincho to be clear, mnesia will create a "mnesia" subdirectory under the DbPath, and write nothing outside of it.

dincho:
@uwiger I shouldn’t post a general comment but to specific line. I’m concerned the Docker example changes in this PR

lucafavatella:
This PR is now invalid as the target branch / PR is now merged. So I shall close this and reopen against master.

lucafavatella:
See #632.

PR #632, opened 2018-01-25
[WIP] PT-154246728 Update docs and Docker following introduction of mnesia DB

lucafavatella:
~Supersedes #630 now that #617 is merged.~

Finishes:
* [PT-154738050](https://www.pivotaltracker.com/story/show/154738050) (leftover of PT-154246728);
* [PT-154636528](https://www.pivotaltracker.com/story/show/154636528).

TODO:
* [ ] Decide final location of config used by Docker (@dincho and @ThomasArts );
* [ ] User experience wrt memory usage (documented memory set for container, mean vs. lean).
  * Docker uses available memory unless quota specified on command line. If Docker host has not enough memory for the container, the miner fails. Post-0.5.0 memory usage increased.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,9 @@ RUN apt-get -qq update && apt-get -qq -y install openssl curl \
 # Deploy application code from builder container
 COPY --from=builder /app/_build/prod/rel/epoch /home/epoch/node
 
+# Deploy TestNet user configuration template
+COPY --from=builder /app/apps/aeutils/test/data/epoch_testnet.yaml /home/epoch/node/epoch_testnet.yaml
+
 # Epoch app won't run as root for security reasons
 RUN useradd --shell /bin/bash epoch \
     && chown -R epoch:epoch /home/epoch
--- a/docs/docker-testnet.md
+++ b/docs/docker-testnet.md
@@ -66,12 +66,12 @@ docker run -d -p 3013:3013 -e PEERS_ADDRESS_0=http://31.13.248.103:3013/ aetrnty
 
 ## Persisting Data
 
-To persist blockchain data and node keys between container runs, use [Docker volumes](https://docs.docker.com/engine/admin/volumes/volumes/). Replace `~/.aeternity/data` with location of your choice.
+To persist blockchain data and node keys between container runs, use [Docker volumes](https://docs.docker.com/engine/admin/volumes/volumes/). Replace `~/.aeternity/db` with location of your choice.
 
 
 ```bash
 docker run -d -p 3013:3013 \
-    -v ~/.aeternity/data:/home/epoch/node/data \
+    -v ~/.aeternity/db:/home/epoch/node/data/mnesia \
     -v ~/.aeternity/keys:/home/epoch/node/keys \
     aetrnty/epoch
 ```
--- a/docs/docker-testnet.md
+++ b/docs/docker-testnet.md
@@ -66,12 +66,12 @@ docker run -d -p 3013:3013 -e PEERS_ADDRESS_0=http://31.13.248.103:3013/ aetrnty
 
 ## Persisting Data
 
-To persist blockchain data and node keys between container runs, use [Docker volumes](https://docs.docker.com/engine/admin/volumes/volumes/). Replace `~/.aeternity/.chain` with location of your choice.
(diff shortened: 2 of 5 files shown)


docs/docker-testnet.md
```diff
@@ -66,12 +66,12 @@ docker run -d -p 3013:3013 -e PEERS_ADDRESS_0=http://31.13.248.103:3013/ aetrnty
...
 ## Persisting Data
 
-To persist blockchain data and node keys between container runs, use [Docker volumes](https://docs.docker.com/engine/admin/volumes/volumes/). Replace `~/.aeternity/.chain` with location of your choice.
+To persist blockchain data and node keys between container runs, use [Docker volumes](https://docs.docker.com/engine/admin/volumes/volumes/). Replace `~/.aeternity/data` with location of your choice.
 
 
 ```bash
 docker run -d -p 3013:3013 \
-    -v ~/.aeternity/.chain:/home/epoch/node/.chain \
+    -v ~/.aeternity/data:/home/epoch/node/data \
```
dincho: Last time I checked data dir contained some other files as well. We're using docker bind mounts and this information will be obfuscated see https://docs.docker.com/engine/admin/volumes/

lucafavatella: @dincho I read:
> Bind mounts ... rely on the host machine’s filesystem having a specific directory structure available.

Are you saying that this will not work unless the client created `~/.aeternity/data` upfront? I expect then that the `docker -v ...` errs if that folder is not present so that user will read the error message, create folder and re-run command? Hence it looks to me no change needed in doc.

dincho: Nope, sorry for not being detailed here.

When bind mount is used as in `-v ~/.aeternity/data:/home/epoch/node/data` if there is any files/directories in `/home/epoch/node/data` in the container, they would become unaccessible.

Example/Test:
```bash
$ docker run -it --rm --entrypoint=bash aetrnty/epoch
epoch@8f8ac1e8959b:~/node$ ls data/
aecore  epoch_config_schema.json

$ mkdir /tmp/.aeternity && touch /tmp/.aeternity/blah
$ docker run -it --rm -v /tmp/.aeternity:/home/epoch/node/data --entrypoint=bash aetrnty/epoch
epoch@61f6fa7c7dae:~/node$ ls data/
blah
```

lucafavatella: > Last time I checked data dir contained some other files as well.

I see 
```
/tmp/node $ ls -d data/*
data/aecore                     data/epoch_config_schema.json   data/mnesia
```

Thanks for spotting this.

lucafavatella: > When bind mount is used as in -v ~/.aeternity/data:/home/epoch/node/data if there are any files/directories in /home/epoch/node/data in the container, they would become unaccessible.
> 
> Example/Test:

And thanks for clear example.

lucafavatella:
> I assume this is tested with the concrete path in the example /home/epoch/node/data/mnesia because I'm not yet fully aware how and when this dir is created.

@dincho On Mon I plan to make a proposal to fix https://www.pivotaltracker.com/story/show/154636528 and test this (hence assigned ticket and this PR to me).

Dockerfile
```diff
@@ -13,6 +13,9 @@ RUN apt-get -qq update && apt-get -qq -y install openssl curl \
 # Deploy application code from builder container
 COPY --from=builder /app/_build/prod/rel/epoch /home/epoch/node
 
+# Deploy TestNet user configuration template
+COPY --from=builder /app/apps/aeutils/test/data/epoch_testnet.yaml /home/epoch/node/epoch_testnet.yaml
```
dincho: I'm afraid this is suboptimal for this kind of change.
This is adding new layer, see [docker docs](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#minimize-the-number-of-layers) on the subject.

The example config should be part of the documentation, and the documentation should be part of the build artifact.

What we're doing here makes no sense to me, we renamed the example file to "test data", now we're using the "test data" as configuration template.

@ThomasArts as the original initiator for this change, could we somehow resolve this issue in a better way ?

lucafavatella:
I am currently considering fixing memory allocation issues without hardcoding lean miner in the config, hence re-assigned to me. Cc @dincho

lucafavatella:
Opening clean rebased one.

PR #635, opened 2018-01-26
Bump db_path to db19

dincho:
It looks like #617 introduces BC break in the persistence layer but db_path version was not bumped.


dincho:
closing of favour of #636

PR #639, opened 2018-01-26
move load_database() call [Finishes #154661732]

uwiger:
The `wait_for_tables()` call to mnesia was made too late, causing a possible race condition, note least with the `aec_conductor` process.


apps/aecore/src/aecore_app.erl
```diff
@@ -15,11 +15,9 @@
 start(_StartType, _StartArgs) ->
     ok = lager:info("Starting aecore node"),
     ok = application:ensure_started(mnesia),
+    aec_db:load_database(),
```
lucafavatella: Comment in export of `load_database/0` shall be updated.
Wait for tables timeout shall be either user configurable or infinity.

PR #640, opened 2018-01-26
Pt 154281622 encode calldata

happi:
Provide a function for turning a function call into calldata.

[finishes #154281622]


apps/aering/test/aer_abi_tests.erl
```diff
@@ -0,0 +1,12 @@
+-module(aer_abi_tests).
+
+-include_lib("eunit/include/eunit.hrl").
+
+encode_call_with_integer_test() -> 
+    <<0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
```
lucafavatella: I understand spec is https://github.com/aeternity/protocol/pull/9/files#diff-a7778cdeae90e4a2ccdc834c003760e7R227

happi: The (offline) spec has been updated as the language was implemented. I have to fix the protocoll repo.

apps/aering/src/aer_abi.erl
```diff
@@ -0,0 +1,41 @@
...
+    %% TODO: check that function exists in cotract.
+    FunctionHandle = encode_function(Contract, Function),
+    Call = {FunctionHandle, Argument},
+    {0, Data} = aer_data:to_binary(Call),
+    _ArgumentType = get_type(Argument),
+    %% TODO: Verify that the type matches the function signature.
+    Data.
+
+encode_function(_Contract, Function) ->
+     << <<X>> || X <- Function>>.
```
lucafavatella: I understand this is equivalent to `list_to_binary(Function)`?

happi: Almost. It will truncate values > 255 to 8 bits without crashing.

test/ring_bytecode_aevm_SUITE.erl
```diff
@@ -28,15 +32,15 @@ execute_identy_fun_from_ring_file(_Cfg) ->
     ok = aeb_disassemble:pp(Code),
 
     %% Create the call data
-    Call = {<<"main">>, 42},
-    {0, Data} = aer_data:to_binary(Call),
-    Res =
+    CallData = aer_abi:create_calldata(Code, "main", 42),
+    io:format("CallData: ~p~n", [CallData]),
```
sennui: do you want to leave it over here?

happi: Ill remove in my next branch/PR when it actually will be used.

PR #642, opened 2018-01-26
Add CircleCI configuration

dincho:
- each commit is build then test dialyzer and eunit in parallel
- master branch is build, test, dialyzer, eunit and deployed to
integration in the end. It also generates package artifacts for both
linux and osx
- env/dev1 and env/dev2 are just packaged and deployed
- tags with format /^v.*$/ are build, tested and deployed to UAT with
manual approval
- test job stores the build logs for further inspection

The new Dockerfile is used to build the builder image [aetrnty/builder](https://hub.docker.com/r/aetrnty/builder/) used in the CI.

Examples:
- [Commit build](https://circleci.com/workflow-run/68ac0637-b849-4054-8ef1-97657491b1c8)
- [Dev environment deploy](https://circleci.com/workflow-run/2466f455-16d5-49e5-83b6-709e3d0aae77)
- [Master branch](https://circleci.com/workflow-run/19dfd816-6265-409f-990f-7921b9a97a7a)
- [Tag build (deploy to UAT)](https://circleci.com/workflow-run/52c64f0c-4f8d-4b8e-8e8d-3db8f0da5c58)

Possible future improvements:
- [Getting more containers](https://circleci.com/pricing/)
- [Docker layer caching](https://circleci.com/docs/2.0/docker-layer-caching/)
- [Better resource class](https://circleci.com/docs/2.0/configuration-reference/#resource_class)
- [JUnit test summary](https://circleci.com/docs/2.0/collect-test-data/)

Todo:
- [x] [Setup CircleCI environment](https://circleci.com/gh/aeternity/epoch/edit#env-vars)
- [x] [Add CircleCI deploy key to epoch servers](https://github.com/aeternity/infrastructure/pull/19)
- [x] Temporary disable deploys until we disable travis deploys

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,12 +121,13 @@ jobs:
             epmd -daemon
             make eunit
 
-  dialyzer:
+  static_analysis:
     <<: *container_config
     steps:
       - checkout
       - *restore_rebar_cache
       - *restore_build_cache
+      - run: ./rebar3 edoc
       - restore_cache:
           keys:
             - dialyzer-cache-v2-{{ .Branch }}-{{ .Revision }}
@@ -198,7 +199,7 @@ jobs:
 # and all its transitively dependent jobs must also have a filters tags section.
 workflows:
   version: 2
-  build_and_test:
+  build_test_deploy:
     jobs:
       - build:
           requires: ~
@@ -232,7 +233,7 @@ workflows:
             tags:
               only: /^v.*$/
 
-      - dialyzer:
+      - static_analysis:
           requires:
             - build
           filters:
@@ -257,7 +258,7 @@ workflows:
           requires:
             - test
             - eunit
-            - dialyzer
+            - static_analysis
           filters:
             branches:
               only: master
@@ -266,7 +267,7 @@ workflows:
           requires:
             - test
             - eunit
-            - dialyzer
+            - static_analysis
             - linux_package
           filters:
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,8 +107,6 @@ jobs:
             make test
       - store_artifacts:
           path: _build/test/logs
-      - store_artifacts:
-          path: _build/test/logs/latest.sync/*/log/*.log
 
   eunit:
     <<: *container_config
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,9 +12,10 @@ references:
     restore_cache:
       key: *rebar_cache_key
 
+  build_cache_key: &build_cache_key build-cache-v1-{{ .Revision }}
   restore_build_cache: &restore_build_cache
     restore_cache:
-      key: build-cache-v1
+      key: *build_cache_key
 
   packages_workspace: &packages_workspace /tmp/packages
   build_package: &build_package
@@ -82,7 +83,7 @@ jobs:
           name: Build
           command: ./rebar3 as test do release
       - save_cache:
-          key: build-cache-v1
+          key: *build_cache_key
           paths:
             - "_build"
             - "apps/aecuckoo/c_src"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -263,6 +263,8 @@ workflows:
           filters:
             branches:
               only: master
+            tags:
+              only: /^v.*$/
 
       - deploy_integration:
           requires:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ references:
         if [ "$(uname -s)" == "Darwin" ]; then
             PKG_SUFFIX="osx-$(sw_vers -productVersion)"
         elif [ "$(uname -s)" == "Linux" ]; then
-            PKG_SUFFIX="linux"
+            PKG_SUFFIX="ubuntu"
         fi
         make prod-package
         mkdir ${PACKAGES_DIR:?}
@@ -41,7 +41,7 @@ references:
         if [ "$(uname -s)" == "Darwin" ]; then
             PKG_SUFFIX="osx-$(sw_vers -productVersion)"
         elif [ "$(uname -s)" == "Linux" ]; then
-            PKG_SUFFIX="linux"
+            PKG_SUFFIX="ubuntu"
         fi
         epmd -daemon
         make python-env
@@ -65,7 +65,7 @@ references:
         environment:
           PACKAGES_DIR: *packages_workspace
         command: |
-          export PACKAGE=${PACKAGES_DIR:?}/epoch-$(cat VERSION)-linux.tar.gz
+          export PACKAGE=${PACKAGES_DIR:?}/epoch-$(cat VERSION)-ubuntu.tar.gz
           cd deployment/ansible
           ansible-playbook -i inventory/openstack.yml --limit="epoch:&${DEPLOY_ENV:?}" \
             --extra-vars "package=${PACKAGE:?} env=${DEPLOY_ENV:?} \
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ references:
         fi
         make prod-package
         mkdir ${PACKAGES_DIR:?}
-        cp _build/prod/rel/epoch/epoch-$(cat VERSION).tar.gz ${PACKAGES_DIR:?}/epoch-$(cat VERSION)-${PKG_SUFFIX}.tar.gz
+        mv _build/prod/rel/epoch/epoch-$(cat VERSION).tar.gz ${PACKAGES_DIR:?}/epoch-$(cat VERSION)-${PKG_SUFFIX}.tar.gz
 
   test_package: &test_package
     run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,305 @@
+version: 2
+
+references:
+  container_config: &container_config
+    docker:
+      - image: aetrnty/builder
+        user: builder
+    working_directory: ~/epoch
+
+  rebar_cache_key: &rebar_cache_key rebar-cache-{{ checksum "rebar.lock" }}-{{ checksum "rebar.config" }}
+  restore_rebar_cache: &restore_rebar_cache
+    restore_cache:
+      key: *rebar_cache_key
+
+  restore_build_cache: &restore_build_cache
+    restore_cache:
+      key: build-cache-v1
+
+  packages_workspace: &packages_workspace /tmp/packages
+  build_package: &build_package
+    run:
+      name: Build Package Tarball
+      environment:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,305 @@
...
+      - *restore_build_cache
+      - restore_cache:
+          keys:
+            - dialyzer-cache-v2-{{ .Branch }}-{{ .Revision }}
+            - dialyzer-cache-v2-{{ .Branch }}-
+            - dialyzer-cache-v2-
+      - run:
+          name: Update dialyzer PLT
+          command: make dialyzer-install
+      - save_cache:
```
ThomasArts: Not sure how this works, but it definitely looks good that one can restore the plt before building on top of it.

.circleci/config.yml
```diff
@@ -0,0 +1,305 @@
...
+                - env/dev1
+                - env/dev2
+            tags:
+              only: /^v.*$/
+
+      - osx_package:
+          requires:
+            - test
+            - eunit
+            - dialyzer
```
ThomasArts: Don't we call it static-analysis these days? It's both dialyzer and edoc. Later on also xref, but later.

ThomasArts: We don't need to run dialyzer on OsX if we already have run it on ubuntu.

.circleci/config.yml
```diff
@@ -0,0 +1,305 @@
...
+      - checkout
+      - *restore_rebar_cache
+      - *restore_build_cache
+      - run:
+          name: Test
+          command: |
+            epmd -daemon
+            make eunit
+
+  dialyzer:
```
ThomasArts: proposal is to call this static_analysis and add edoc to this job as we recently did for travis

.circleci/config.yml
dincho: `dialyzer-cache-v2-{{ .Branch }}-{{ .Revision }}` - this tries to restore dialyzer cache for the same branch and commit SHA/revision as the running job, IF NOT ->
`dialyzer-cache-v2-{{ .Branch }}-` - try to restore the cache for the same branch, but regardless of the revision, IF NOT ->
`dialyzer-cache-v2-` - try to restore the cache from any branch. I'm not 100% sure how to this works, I assume it restores from the most recent build. If this is not too effective we could try to restore from `master` instead ?

I hope dialyzer is clever enough to cope only with the changes between the cached PLT and the updated source code resulting in much faster update, does it?

.circleci/config.yml
dincho: This actually do not run the dialyzer twice, it's just and requirement that dialyzer must success before the osx packaging can start.

You can see this workflow here: https://circleci.com/workflow-run/19dfd816-6265-409f-990f-7921b9a97a7a

If I remove this requirement, the workflow could produce an osx package which is not successfully analyzed by dialyzer. I consider this broken package.

.circleci/config.yml
```diff
@@ -0,0 +1,305 @@
...
+  build_package: &build_package
+    run:
+      name: Build Package Tarball
+      environment:
+        PACKAGES_DIR: *packages_workspace
+      command: |
+        if [ "$(uname -s)" == "Darwin" ]; then
+            PKG_SUFFIX="osx-$(sw_vers -productVersion)"
+        elif [ "$(uname -s)" == "Linux" ]; then
+            PKG_SUFFIX="linux"
```
lucafavatella: I notice this used to be `ubuntu`.

.circleci/config.yml
```diff
@@ -0,0 +1,305 @@
...
+        fi
+        make prod-package
+        mkdir ${PACKAGES_DIR:?}
+        cp _build/prod/rel/epoch/epoch-$(cat VERSION).tar.gz ${PACKAGES_DIR:?}/epoch-$(cat VERSION)-${PKG_SUFFIX}.tar.gz
```
lucafavatella: Could you please prefer mv to cp so avoid unexpected archives around?

.circleci/config.yml
```diff
@@ -0,0 +1,305 @@
...
+      - run:
+          name: Build
+          command: ./rebar3 as test do release
+      - save_cache:
+          key: build-cache-v1
+          paths:
+            - "_build"
+            - "apps/aecuckoo/c_src"
+            - "apps/aecuckoo/priv/bin"
+            - "apps/aecuckoo/priv/lib"
```
lucafavatella: Are these bin and lib saved across different runs (e.g. distinct commits) or across different stages of the same run (e.g. build then test)? My concern that change in build env across commits may make these binaries cause failure.

.circleci/config.yml
```diff
@@ -0,0 +1,305 @@
...
+      - *restore_build_cache
+      - run:
+          name: Test
+          command: |
+            epmd -daemon
+            make test
+      - store_artifacts:
+          path: _build/test/logs
+      - store_artifacts:
+          path: _build/test/logs/latest.sync/*/log/*.log
```
lucafavatella: Why only latest.sync?

.circleci/config.yml
dincho: Yeah, I was a little bit out of sync. Updated in 383eb19

.circleci/config.yml
lucafavatella: or at least - why only this one has special treatment?

.circleci/config.yml
dincho: updated in 383eb19

.circleci/config.yml
dincho: What else?

.circleci/config.yml
```diff
@@ -0,0 +1,305 @@
...
+      - *test_package
+      - *store_package_artifacts
+      - persist_to_workspace:
+          root: *packages_workspace
+          paths:
+            - "*.tar.gz"
+
+  osx_package:
+    macos:
+      xcode: "9.0"
```
lucafavatella: We need to know package name before the release if we keep package name inside release notes.

.circleci/config.yml
dincho: While it looks more natural to me to keep the rebar build immutable from "outside" and just "pick" what's needed, I applied your request because probably you have better reasoning. a7607ac

.circleci/config.yml
```diff
@@ -0,0 +1,306 @@
...
+          type: approval
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v.*$/
+
+      - deploy_uat:
+          requires:
+            - linux_package
```
lucafavatella: How do we make sure that the package uploaded to GitHub is the same deployed to UAT?

.circleci/config.yml
```diff
@@ -0,0 +1,306 @@
...
+        if [ "$(uname -s)" == "Darwin" ]; then
+            PKG_SUFFIX="osx-$(sw_vers -productVersion)"
+        elif [ "$(uname -s)" == "Linux" ]; then
+            PKG_SUFFIX="ubuntu"
+        fi
+        make prod-package
+        mkdir ${PACKAGES_DIR:?}
+        mv _build/prod/rel/epoch/epoch-$(cat VERSION).tar.gz ${PACKAGES_DIR:?}/epoch-$(cat VERSION)-${PKG_SUFFIX}.tar.gz
+
+  test_package: &test_package
```
lucafavatella: Does CircleCI support upload of package to GitHub release?

.circleci/config.yml
dincho: Ah, I tried to use workspaces for this at first, then after switching to cache I forgot to use a local scope key. Fixed in 33e2fba.

.circleci/config.yml
```diff
@@ -0,0 +1,307 @@
...
+                - env/dev1
+                - env/dev2
+            tags:
+              only: /^v.*$/
+
+      - linux_package:
+          filters:
+            branches:
+              only:
+                - master
```
lucafavatella: I see packages (linux and osx) built and smoke tested only on master - not on PRs/pushes. How does developer know that he/she broke package smoke test? I am concerned failures on master may go silent and package smoke test issue noticed only on tag.

.circleci/config.yml
lucafavatella: latest.fork

dincho: This actually as I try to workaround an issue with symlinks that I think CircleCI has with artifacts.
See last task in https://circleci.com/gh/aeternity/epoch/37#artifacts/containers/0

lucafavatella:
> test job stores the build logs for further inspection

I understand this is why failed_logs.escript is not called. Ok.

.circleci/config.yml
```diff
@@ -0,0 +1,306 @@
...
+              only: /^v.*$/
+
+      - osx_package:
+          requires:
+            - test
+            - eunit
+            - static_analysis
+          filters:
+            branches:
+              only: master
```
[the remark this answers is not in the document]
dincho: fixed in 0ea73d7 *This bug was even visible on the example workflow ...*

.circleci/config.yml
dincho: Currently it is not the same package in travis too? Or I'm missing something?

.circleci/config.yml
lucafavatella: seems like requires further fine tuning

.circleci/config.yml
dincho: I'm afraid not. We could do that manually until we find a simple tool that we can use to do that (I believe that couple of `curl` should also do.)

.circleci/config.yml
dincho: I've never seen failed "package" job on travis so far, that's why I was convinced to propose to take out it from the PR workflow and keep it only for master.

Furthermore we also should work to make sure that failed master is much more visible than now (e.g. slack notification *but when we migrate to paid plan, because current slack hooks are exhausted) and fixed with the highest priority.

I believe this would overall improve the PR build times with the cost of very small risk.

.circleci/config.yml
dincho: According to the CCI docs it should be 10.12.6, see
https://circleci.com/docs/1.0/build-image-macos/#selecting-xcode-version-and-operating-system and https://circle-macos-docs.s3.amazonaws.com/image-manifest/build-282/index.html

dincho: v2.0 doc: https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions

.circleci/config.yml
dincho: see 28aa4b7

dincho:
I disabled deploys in 0c26070

.circleci/config.yml
ThomasArts: That would probably work well. Worst case, dialyzer will start all over, but in all good cases, we save time.

PR #644, opened 2018-01-29
HTTP client endpoints: customized tx serialization [Finishes #154658054]

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/154658054)
[Github issue](https://github.com/aeternity/protocol/issues/15)
Provided a distinct serialization of transactions for clients, independend of the gossip protocol.
Added `block_heigh` and `block_hash` to transaction objects returned to client's APIs


apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -1736,16 +1737,20 @@ header_to_endpoint_top(Header) ->
...
 block_to_endpoint_map(Block) ->
     block_to_endpoint_map(Block, #{tx_objects => false}).
 
 block_to_endpoint_map(Block, Options) ->
-    SerializeFun =
+    Encoding =
         case maps:get(tx_objects, Options, false) of
-            true -> fun aec_blocks:serialize_client_readable/1;
-            false -> fun aec_blocks:serialize_to_map/1
+            false -> message_pack;
```
uwiger: This seems unintuitive to me. Why does `tx_objects` imply a certain encoding? Why not make it explicit instead?

velzevur: This is by client's API request. I agree on that making it more explicit would be better. I'll discuss it with @dadaphl if it would be a big change on his side.

velzevur:
Renamed the query parameter from the boolean `tx_objects` to the enum `tx_encoding` having self desriptive values of `message_pack` and `json`.
Will address the corresponding change in swagger's discirminators (`data_schema`) in another PR
Rebased on top of master.

apps/aehttp/priv/swagger.json
```diff
@@ -597,12 +598,13 @@
...
-          "name" : "tx_objects",
+          "name" : "tx_encoding",
           "in" : "query",
           "description" : "Transactions encoding: MessagePack (default) or JSON",
           "required" : false,
-          "type" : "boolean",
-          "default" : false
+          "type" : "string",
+          "default" : "message_pack",
+          "enum" : [ "message_pack", "json" ]
```
lucafavatella: Maybe a Swagger definition may have avoided repetition? Though totally optional of course.

apps/aecore/src/aec_tx_sign.erl
```diff
@@ -138,6 +139,19 @@ deserialize_from_binary(SignedTxBin) when is_binary(SignedTxBin) ->
...
+-spec serialize_for_client(json|message_pack, #header{}, aec_tx_sign:signed_tx()) ->
+                              binary() | map().
+serialize_for_client(Encoding, Header, #signed_tx{}=S) ->
+    {ok, Hash} = aec_headers:hash_header(Header),
+    serialize_for_client(Encoding, S, aec_headers:height(Header), Hash).
+
+serialize_for_client(message_pack, #signed_tx{}=S, BlockHeight, BlockHash) ->
+    GenericData = [#{<<"block_height">> => BlockHeight},
+                   #{<<"block_hash">> => BlockHash}],
+    aec_base58c:encode(transaction, msgpack:pack(serialize(S) ++ GenericData));
```
lucafavatella: The usage of msgpack here seems similar to `serialize_to_binary` though with `++ GenericData`. Are we sure user does not want same msgpack as in `serialize_to_binary` (as in Merkle tree of txs)?

apps/aehttp/priv/swagger.json
velzevur: I've just tried this but for some reason current swagger does not accept query params to be schema refs. For example
```                                                                          
  /block-by-hash:                                                                                    
    get:
      tags:
        - external 
      operationId: GetBlockByHash 
      description: 'Get a block by hash'                                                             
      consumes:
        - application/json                                                                           
      produces:
        - application/json                                                                           
      parameters: 
        - in: query
          name: hash
          description: 'Hash of the block to fetch'                                                  
          required: true                                                                             
          schema: 
````
is invalid, no matter that  `'#/definitions/EncodedHash' ` is a valid definition

File: apps/aecore/src/aec_tx_sign.erl (hunk 3 above)
Comment (velzevur) [reply apps/aecore/src/aec_tx_sign.erl]: It had been a special requirement for the client APIs the transaction objects to include certain fields. Currently those are just the hash and height of the block that included the transaction but there will be more to come (transaction's hash for example).
Having this in mind, we can not have the same msgpack both for the gossip protocol and the client APIs

**Comment by velzevur** [comment]
Added stronger test guarantees. After approval, will squash the commits into one.

File: apps/aecore/src/aec_tx_sign.erl [hunk 4]
```diff
@@ -147,11 +148,30 @@ serialize_for_client(Encoding, Header, #signed_tx{}=S) ->
 
 serialize_for_client(message_pack, #signed_tx{}=S, BlockHeight, BlockHash) ->
     GenericData = [#{<<"block_height">> => BlockHeight},
-                   #{<<"block_hash">> => BlockHash}],
-    aec_base58c:encode(transaction, msgpack:pack(serialize(S) ++ GenericData));
+                   #{<<"block_hash">> => aec_base58c:encode(block_hash, BlockHash)}],
+    aec_base58c:encode(transaction, msgpack:pack(serialize(S) ++ [GenericData]));
```
lucafavatella: /me trying to understand why putting list of two items inside a list - resulting in list of one item that is list of two items.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -1747,7 +1747,21 @@ block_to_endpoint_map(Block) ->
 block_to_endpoint_map(Block, Options) ->
     Encoding = maps:get(tx_encoding, Options, message_pack),
     BMap = aec_blocks:serialize_client_readable(Encoding, Block),
-    aehttp_dispatch_ext:cleanup_genesis(BMap).
+    Expected = aehttp_dispatch_ext:cleanup_genesis(BMap),
+
+    %% Validate that all transactions have the correct block height and hash
+    ExpectedTxs = maps:get(<<"transactions">>, Expected, []),
```
lucafavatella: This default `[]` may cause things to slip though. Shall this rather case/switch on height of block, and if not genesis then hardcode expectation that `<<"transactions">>` is there otherwise hardcode expectation that `<<"transactions">>` is not there hence return `[]`?

apps/aecore/src/aec_tx_sign.erl
lucafavatella: This is used here https://github.com/aeternity/epoch/pull/644/commits/d85817d364f9f94d03a52ac81ae6cd3489c74308#diff-43772af989581a69a6ac311448a78975R162

apps/aecore/src/aec_tx_sign.erl
```diff
@@ -147,11 +148,30 @@ serialize_for_client(Encoding, Header, #signed_tx{}=S) ->
...
 serialize_for_client(json, #signed_tx{data = Tx, signatures = Sigs},
                      BlockHeight, BlockHash) ->
     #{<<"tx">> => aec_tx:serialize_for_client(Tx),
       <<"block_height">> => BlockHeight,
       <<"block_hash">> => aec_base58c:encode(block_hash, BlockHash),
       <<"signatures">> => lists:map(fun(Sig) -> aec_base58c:encode(signature, Sig) end, Sigs)}. 
+
+extract_encoded_generic_data(message_pack, Bin) ->
```
lucafavatella: As a reader I do not easily correlate `extract_encoded_generic_data` with `serialize_for_client` - closely coupling is implicit. I would suggest either enriching name of `extract_encoded_generic_data`, or (better) write type spec to `extract_encoded_generic_data` and `serialize_for_client` so that it is clear they operate on same data.

apps/aecore/src/aec_tx_sign.erl
```diff
@@ -147,11 +148,30 @@ serialize_for_client(Encoding, Header, #signed_tx{}=S) ->
...
+    {transaction, MsgPackBin} = aec_base58c:decode(Bin),
+    {ok, [_Type, _Version, _TxSer, _Sigs, GenericData]} = msgpack:unpack(MsgPackBin),
+    [#{<<"block_height">> := BlockHeight},
+     #{<<"block_hash">> := BlockHashEncoded}] = GenericData,
```
lucafavatella: These items are tagged - that is good.

velzevur:
Added an assertion on empty transactions list (possible only for genesis block)

Renamed generic_data in the context of aec_tx_sign to be meta_data: function aec_tx_sign:extract_encoded_generic_data/2 to aec_tx_sign:meta_data_from_client_serialized/2 to represent its close connection to client serialized objects.

config/swagger.yaml
```diff
@@ -637,10 +637,11 @@ paths:
           required: true
           type : integer
         - in: query
-          name: tx_objects
+          name: tx_encoding
           description: 'Transactions encoding: MessagePack (default) or JSON'
```
uwiger: Should this not say "message_pack" and "json", i.e. the actual values?

config/swagger.yaml
```diff
@@ -666,10 +667,11 @@ paths:
           required: true
           type : string 
         - in: query
-          name: tx_objects
+          name: tx_encoding
           description: 'Transactions encoding: MessagePack (default) or JSON'
```
uwiger: Should this not say "message_pack" and "json", i.e. the actual values?

config/swagger.yaml
```diff
@@ -694,10 +696,11 @@ paths:
         - application/json
       parameters:
         - in: query
-          name: tx_objects
+          name: tx_encoding
           description: 'Transactions encoding: MessagePack (default) or JSON'
```
uwiger: Should this not say "message_pack" and "json", i.e. the actual values?

config/swagger.yaml
```diff
@@ -714,10 +717,11 @@ paths:
...
```
uwiger: Should this not say "message_pack" and "json", i.e. the actual values?

config/swagger.yaml
```diff
@@ -734,10 +738,11 @@ paths:
...
```
uwiger: etc...

config/swagger.yaml
velzevur: It should and it does. [3 lines bellow](https://github.com/aeternity/epoch/pull/644/files#diff-89b8ec7c2f28f3e3c7d21ca9bf070bb7R491):
```
enum: [message_pack, json]```

File: config/swagger.yaml (hunk 9 above)
Comment (velzevur) [reply config/swagger.yaml]: It should and it does. [3 lines bellow](https://github.com/aeternity/epoch/pull/644/files#diff-89b8ec7c2f28f3e3c7d21ca9bf070bb7R491):
```
enum: [message_pack, json]
```

File: config/swagger.yaml (hunk 10 above)
Comment (velzevur) [reply config/swagger.yaml]: It should and it does. [3 lines bellow](https://github.com/aeternity/epoch/pull/644/files#diff-89b8ec7c2f28f3e3c7d21ca9bf070bb7R491):
```
enum: [message_pack, json]
```

File: config/swagger.yaml (hunk 11 above)
Comment (velzevur) [reply config/swagger.yaml]: It should and it does. [3 lines bellow](https://github.com/aeternity/epoch/pull/644/files#diff-89b8ec7c2f28f3e3c7d21ca9bf070bb7R491):
```
enum: [message_pack, json]
```

File: config/swagger.yaml (hunk 12 above)
Comment (velzevur) [reply config/swagger.yaml]: It should and it does. [3 lines bellow](https://github.com/aeternity/epoch/pull/644/files#diff-89b8ec7c2f28f3e3c7d21ca9bf070bb7R600):
```
enum: [message_pack, json]
```

lucafavatella:
I see 2 approvals, CI passed, master CI passed.

PR #648, opened 2018-01-29
Regression test for node start with chain [Finishes #154667814]

lucafavatella:
The regression test works:
* I based this branch on master - before the node start fix - and you can see from [CI build of push on my fork](https://travis-ci.org/lucafavatella/epoch/builds/334727970) that the CI `package` job fails there - hence catching the regression;
* The CI build on this PR merges recent master so it passes as it includes the fix.


lucafavatella:
I see CI passed, 4 approvals. No conflicting change in master.

PR #649, opened 2018-01-29
[WIP] Pt 154171508 contract http api

happi:


apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -119,16 +119,16 @@ call_contract(_Cfg) ->
     %% Now call check that we can call it.
     Fee           = 107,
     GasPrice      = 2,
-    Arg           = 42,
-    {0, CallData} = aer_data:to_binary({<<"main">>, Arg}),
+    Arg           = <<"42">>,
```
ThomasArts: Do we pass the arguments as strings? If so, is 42 then to be interpreted hex or decimal? In the test it's decimal.

apps/aehttp/priv/swagger.json
```diff
@@ -1623,6 +1687,42 @@
...
+        }
+      }
+    },
+    "ByteCode" : {
+      "type" : "object"
+    },
+    "ContractCall" : {
+      "type" : "object",
+      "properties" : {
+        "code" : {
```
ThomasArts: is this the hash to the contract byte code on the chain?

apps/aehttp/priv/swagger.json
```diff
@@ -1623,6 +1687,42 @@
...
+          "type" : "string"
+        },
+        "function" : {
+          "type" : "string"
+        },
+        "arg" : {
```
ThomasArts: Is the idea to extend this later to a list of argument objects, of which the contract defines the type, or is this always a string representation that is parsed lower down? If so, it might be nice with a comment in YAML where the string representation is defined.

apps/aering/src/aer_abi.erl
```diff
@@ -12,15 +12,28 @@
...
     %% TODO: check that function exists in cotract.
     FunctionHandle = encode_function(Contract, Function),
+    case aer_constants:string(Argument) of
+        {ok, ParsedArgument} ->
+            encode_call(FunctionHandle, ParsedArgument);
+        {error, _} ->
+            {error, argument_syntax_error}
+    end.
+
+encode_call(FunctionHandle, ArgumentAst) ->
```
ThomasArts: You need to agree with compiler on the format. You expect ("add", (1, 2)) but compiler expects ("add", 1, 2).

apps/aecontract/test/aecontract_SUITE.erl
happi: From the "outside" http we pass arguemnts as strings of Ring data.

apps/aehttp/priv/swagger.json
happi: No this is for off-chain call where "code" is the code encoded as a hex string.

apps/aehttp/priv/swagger.json
happi: The function is just the name of the called function.

apps/aering/src/aer_abi.erl
happi: Yes that has to be fixed. I asked a question about this in the ring channel at some point.
My impression is that ring functions always takes one argument: a tuple.

py/tests/swagger_client/__init__.py
```diff
@@ -19,6 +19,10 @@
 from .models.balance import Balance
```
ThomasArts: remove this file.... it is auto generated and will give conflicts

py/tests/swagger_client/apis/external_api.py
```diff
@@ -40,6 +40,214 @@ def __init__(self, api_client=None):
                 config.api_client = ApiClient()
```
ThomasArts: remove the python files from this PR, they will give problems and are auto generated anyway,

happi:
@hanssv Working on python test. Not sure they will make the release, but I hope so.

happi:
@ThomasArts I have removed the files you indicated. Seems like they also exist in e.g. PR 665.
Perhaps there should be something in gitignore?

What exactly do you mean by "I would probably not add the generated python tests to this commit."?

apps/aering/src/aer_abi.erl
ThomasArts: It takes a tuple function name and then one item per argument.... but this can be fixed later

apps/aehttp/priv/swagger.json
```diff
@@ -1224,6 +1224,70 @@
           }
         }
       }
+    },
+    "/contract/compile" : {
+      "post" : {
+        "tags" : [ "external" ],
+        "description" : "Compila a ring contract from source and return byte code",
```
uwiger: "Compila" -> "Compile"

config/swagger.yaml
```diff
@@ -1042,6 +1042,66 @@ paths:
           schema:
             $ref: '#/definitions/Error'
       security:
+  /contract/compile:
+    post:
+      tags:
+        - external
+      operationId: 'CompileContract'
+      description: 'Compila a ring contract from source and return byte code'
```
uwiger: "Compila" -> "Compile"

apps/aehttp/priv/swagger.json
happi: Ill fix in my next PR containing tests unless there are other changes.

lucafavatella:
[A CI job](https://travis-ci.org/aeternity/epoch/jobs/335693416#L434) failed in downloading Erlang - I am taking the liberty to restart it.

PR #650, opened 2018-01-30
Auto-unblock peers after 15 minutes and don’t block trusted peers

hanssv:
While contemplating more advanced peer handling strategies, this should solve the current problems with integration network, etc. in a lightweight way.


hanssv:
sha3 fail to compile on Travis, no idea what is going on there...?!

PR #651, opened 2018-01-30
Make clean removes _build/ completely

ThomasArts:


Makefile
```diff
@@ -180,6 +180,7 @@ killall:
 
 clean:
 	@./rebar3 clean
+	@rm -rf _build/
```
lucafavatella: `( cd $(HTTP_APP) && $(MAKE) clean; )` ?

ThomasArts: and that for all directories in which we have a make file, I guess...

Makefile
```diff
@@ -180,6 +180,10 @@ killall:
 
 clean:
 	@./rebar3 clean
+	( cd apps/aecuckoo && $(MAKE) clean; )
```
lucafavatella: Actually this is already meant to be run by rebar pre_hook in apps/aecuckoo/rebar/config

PR #654, opened 2018-01-30
Fixed bug in spend_tx, do spend followed by earn

hanssv:
Previously we fetched both sender and receiver from the state tree and
updated the accounts separately. If sender == receiver we overwrote the
sender update.


hanssv:
Yes, updated with bump of the deployment DB path.

lucafavatella:
I see two reviewers approved. I see CI passed. I pinged Hans and ok to merge. I check master DB path does not conflict. Merging.

PR #656, opened 2018-01-30
[WIP] Add index plugin + accounts/1 tx callback

uwiger:
Introduced some support for emergency-patching OTP modules.

Related issues:
https://www.pivotaltracker.com/story/show/154672455 (add indexes for tx mapping)
https://www.pivotaltracker.com/story/show/154658302 (HTTP API tx endpoints: filter by tx type)
https://www.pivotaltracker.com/story/show/154658088 (transactions related to account)


apps/aeoracle/src/aeo_query_tx.erl
```diff
@@ -115,6 +116,10 @@ check(#oracle_query_tx{sender = SenderPubKey, nonce = Nonce,
         {error, Reason} -> {error, Reason}
     end.
 
+-spec accounts(query_tx()) -> [pubkey()].
+accounts(#oracle_query_tx{sender = SenderPubKey}) ->
```
hanssv: Should the oracle be part of the accounts (i.e. the oracle that gets the query?)

otp_patches/mnesia.erl.patch
```diff
@@ -0,0 +1,136 @@
+diff --git a/lib/mnesia/src/mnesia.erl b/lib/mnesia/src/mnesia.erl
```
hanssv: How stable is this? It seems like it could be a problem if it changes slightly, say from R20.1 to R20.2 or so?

apps/aeoracle/src/aeo_query_tx.erl
uwiger: Good question ...

otp_patches/mnesia.erl.patch
uwiger: I checked with current OTP master, but it would perhaps be better to have subdirectories for different supported OTP versions.

deployment/ansible/deploy.yml
```diff
@@ -40,7 +40,7 @@
           port: 3013
       chain:
         persist: true
-        db_path: "./db20"
+        db_path: "./db21"
```
lucafavatella: Requires rebase on master and further bump.

rebar.config
```diff
@@ -65,7 +65,10 @@
                                         "{{output_dir}}/bin/check_config"},
                                        {copy,
                                        "_build/dev1/lib/aeutils/priv/epoch_config_schema.json",
-                                       "{{output_dir}}/data/epoch_config_schema.json"}]}
+                                       "{{output_dir}}/data/epoch_config_schema.json"},
+                                       {copy,
+                                        "./otp_patches/ebin",
+                                        "{{output_dir}}/patches"}]}
```
lucafavatella: I see that this approach from `otp_patches` top dir builds using pre-hook in top rebar.config then copies files in release `patches` dir then it relies on `-pa` of vm.args for loading in release. This build&ship approach seems custom: would it be possible to add smoke test in the `package` job for the prod release exercising functionality (e.g. maybe at end of release.py - i.e. after chain has at least one coinbase - start nodes again then call specific HTTP API on one node checking response).

I understand that the point of using `-pa` is aiming at inclusion of patched modules with same name of OTP ones in front of code path: does it work in embedded mode? (I admit I did not test it.)

I am not saying to change approach, though for the sake of maintainability may I suggest we *consider* modularizing otp_patches / compile pre_hook / copy step in release build as separate OTP app? The tricky bit would be whether it is possible to make user app load module with same name of module in OTP app, and make it take precedence in code path.

uwiger: A possible way forward is that the patch system is extended as-needed.

Packaging patches as a separate OTP application would be problematic, as the OTP `systools` script builder doesn't accept duplicate module names, and so would refuse to generate the start script. The `-pa` flag does work with embedded code loading.

One caveat with the chosen approach is that it does require the current OTP version to have source included.

gorillainduction:
I can't really say much about the patch handling. And the index/transaction table seems unused?

I guess this is intended follow up work?

Anyway, the approach seems sound. I suggest we merge this after the next release, though. To give us time to test the functionality before it is released.

uwiger:
@gorillainduction yes, the index/transaction table is meant for the endpoint API, which is on @velzevur 's plate.

lucafavatella:
> Anyway, the approach seems sound. I suggest we merge this after the next release, though. To give us time to test the functionality before it is released.

@uwiger Could you please split any independent urgent fix in separate PR - I understand it is "only" https://www.pivotaltracker.com/story/show/154737069

velzevur:
CI failed with
```
%%% aec_eunit_SUITE ==> aec_sync_test: FAILED
%%% aec_eunit_SUITE ==> {errors_in_lager_log,["15:44:15.107 [error] Supervisor exec_app had child exec started with exec:start_link([]) at <0.20958.0> exit with reason killed in context shutdown_error\n\n"]}
```
Since this test failure has nothing to do with this PR, I took the liberty of restarting the Travis JOB

PR #657, opened 2018-01-30
Add docker docs how to change config file

dincho:
Better example would be:

```bash
docker run -d -p 3013:3013 \
    -v ~/.aeternity/epoch.yaml:/home/epoch/node/epoch.yaml \
    aetrnty/epoch
```

However, it does not seems that bind mount works the same way as with directories to hide the container (target) file.

[PT-154764969](https://www.pivotaltracker.com/n/projects/2124891/stories/154764969)

### DIFF
--- a/docs/docker-testnet.md
+++ b/docs/docker-testnet.md
@@ -66,7 +66,7 @@ docker run -d -p 3013:3013 -e PEERS_ADDRESS_0=http://31.13.248.103:3013/ aetrnty
 
 ### Changing the configuration file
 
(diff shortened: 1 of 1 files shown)


dincho:
> Could you please include full comment - plus snippet - in the commit message of one of the commits that is included in master (i.e. no squash merge)? Otherwise next person looking at this piece of doc will think "oh, I can do better!"

Good idea, I'll do that at the time of merge (I do squash and merge and I can change the message)

PR #659, opened 2018-01-30
Pt 154635474 mnesia as mpt backend

gorillainduction:
https://www.pivotaltracker.com/n/projects/2124891/stories/154635474

To enable better storage in the db, the structure of oracle queries, and contract calls has been changed.

The persistence server has also been removed since it is no longer used.


apps/aecontract/src/aect_state_tree.erl
```diff
@@ -75,37 +68,28 @@ lookup_contract(Id, Tree) ->
...
-    %% Update the call tree hash in the contract
-    {ok, Hash} = aeu_mtrees:root_hash(CallTree1),
-    Contract1  = aect_contracts:set_calls_hash(Hash, Contract),
-
-    %% Update the contract tree and the calls tree
-    CtTree1    = aeu_mtrees:enter(CtId, aect_contracts:serialize(Contract1), CtTree),
-    CallTrees1 = gb_trees:enter(CtId, CallTree1, CallTrees),
+    %% Insert the new call into the history
+    Serialized = aect_call:serialize(Call),
+    CtTree1    = aeu_mtrees:insert(CallTreeId, Serialized, CtTree),
```
lucafavatella: This means inserting in tree both contracts and calls.
This also means inserting in same tree with key of different length.

(I am trying to understand rationale behind and impact.)

apps/aecore/src/aec_trees.erl
```diff
@@ -29,12 +30,17 @@
...
-           contracts = aect_state_tree:empty(),
+    #trees{accounts  = aec_accounts_trees:empty_with_backend(),
+           oracles   = aeo_state_tree:empty_with_backend(),
+           contracts = aect_state_tree:empty_with_backend(),
            ns        = aens_state_tree:empty()
           }.
 
+-spec commit_to_db(trees()) -> trees().
+commit_to_db(Trees) ->
+    %% Make this in a transaction to get atomicity.
```
lucafavatella: Do you mean DB transaction between state trees and block? (in aec_chain_state:state_db_put)

apps/aeoracle/src/aeo_state_tree.erl
```diff
@@ -31,26 +34,23 @@
 %% in the oracle field in the state tree, does not work well. Since the state
 %% tree has to be a Merkle tree all nodes are serialized. This would mean
 %% deserialize/serialize of all queries when adding or updating a single
-%% query. Instead we have a forest of Merkle trees (on Merkle tree per
-%% oracle) in a separate structure, and only store the root hash of that
-%% queries tree in the oracle field.
+%% query. Instead we store the queries prefixed with the oracle id in
+%% the same tree as the oracles. This is to enable iteration over a single
+%% oracle's queries.
```
lucafavatella: Requires update of https://github.com/aeternity/protocol/blob/da3a79f9399f73f78833bb74f51ddd66b49ac647/oracles/oracle_state_tree.md#oracle-state-tree

apps/aecontract/src/aect_state_tree.erl
gorillainduction: The rational is to keep all things in the MPTree so that it will get the befefits as other tree nodes (primarily that nodes are shared between block states).

Earlier, the forrest of gb_trees was getting duplicated for all block states.

Prepending the parent id is for finding the calls to one particular contract. This is not used here yet (See oracles)

apps/aecore/src/aec_trees.erl
gorillainduction: I am referring to the line below. The commit of the state trees need to be atomic to not miss any nodes.

lucafavatella:
I see two approvals. I see CI build of branch push passed.
I see that CI build of PR (i.e. branch merged to master) failed though I also notice that:
* This PR (888a5ba7) is also based on f05d6d97 that is current master, so PR build is same as push build;
* [Error](https://travis-ci.org/aeternity/epoch/jobs/335627838#L1447) is [known](https://www.pivotaltracker.com/story/show/154770099/comments/185310636).

PR #664, opened 2018-01-31
JUnit Reports

dincho:
Enable JUnit reports so we can integrate it into [CircleCI Test Data Collecting](https://circleci.com/docs/2.0/collect-test-data/).

This will give us some nice info about failed tests.

I was trying to enable this just in the CI by:
`CT_TEST_FLAGS="-ct_hooks cth_surefire" make test` but I get:
> ===> Invalid option -ct_hooks on task ct

I changed the Makefile first so this flags can be passed forward:
```
diff --git a/Makefile b/Makefile
index 39685b8..811b5bc 100644
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 CORE = rel/epoch/bin/epoch
 VER := $(shell cat VERSION)
 
-CT_TEST_FLAGS =
+CT_TEST_FLAGS ?=
```

It looks like I don't know how to pass this option to `rebar ct` or it does not supports it. `ct_run` with this option works as expected.

However, I don't think this report add significant additional time to the test runs, so we could keep it in the rebar.config.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,6 +105,8 @@ jobs:
           command: |
             epmd -daemon
             make test
+      - store_test_results:
+          path: _build/test/logs
       - store_artifacts:
           path: _build/test/logs
 
--- a/apps/aecore/test/aec_coinbase_tx_tests.erl
+++ b/apps/aecore/test/aec_coinbase_tx_tests.erl
@@ -80,7 +80,7 @@ create_coinbase_tx_no_account_test() ->
               {"Check coinbase trx without existing account in state: shall create account",
                fun() ->
                        {Succ, Trees} = aec_coinbase_tx:check(CoinbaseTx, Trees0, 9),
-                       ?assertEqual(fail, Succ),
+                       ?assertEqual(ok, Succ),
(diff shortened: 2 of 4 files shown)


dincho:
> I guess it is not possible to see the insight for the Circle CI job that ran on this PR/branch?

Oh, I forgot to post the link (during the standup). https://circleci.com/build-insights/gh/aeternity/epoch/junit_reports

However, I see this as as downside because I don't think there is aggregated report, but per branch only. Still the master should gave us some good info.

lucafavatella:
> Oh, I forgot to post the link (during the standup). https://circleci.com/build-insights/gh/aeternity/epoch/junit_reports

@dincho  Gave me 404

dincho:
@lucafavatella @ThomasArts It's weird you're getting 404. Are you logged in in CCI?

dincho:
> Could you squash the commits such that we don't first commit broken test and then repair them?

Sure, I'll do prior merging.

lucafavatella:
> @lucafavatella @ThomasArts It's weird you're getting 404. Are you logged in in CCI?

Indeed as logged in I see the link.

PR #675, opened 2018-02-01
Add hostname option when using mnesia dir bind mount

dincho:
rel: https://www.pivotaltracker.com/story/show/154763964


lucafavatella:
Thanks.

PR #676, opened 2018-02-01
Add docker compose config

dincho:
To start the localnet with epoch source from the repo, you have to build the image first: 
```bash
docker-compose build node1
docker-compose up
```

To test the API see [`docs/docker.md`](https://github.com/aeternity/epoch/blob/d5f8305f7ab6c85b099b2de86549d1cdf69efa9b/docs/docker.md#localnet) for details.

TODO:
- [x] nginx proxy with CORS (allow *) config

[PT 154801173](https://www.pivotaltracker.com/story/show/154801173)


docs/docker.md
```diff
@@ -1,4 +1,4 @@
-# Join the testnet with Docker
```
lucafavatella: Could you please update the link in RELEASE-NOTES.md?

docs/docker.md
```diff
@@ -88,3 +88,53 @@ docker run -d -p 3013:3013 --hostname node0 \
 ```
 
 **Note: make sure `hostname` option is set when reusing the mnesia data directory**
+
+# Localnet
```
lucafavatella: This seems relevant enough to be mentioned in the release notes. Would you mind adding a note in the [release notes first paragraph](https://github.com/aeternity/epoch/blob/bc08fc66a61a6cd70f8b926107d5e5f5890dc78a/RELEASE-NOTES.md#about-this-release) deleting all the other stuff - as this is the first user-visible improvement in this release cycle?

docs/docker.md
```diff
@@ -88,3 +88,53 @@ docker run -d -p 3013:3013 --hostname node0 \
...
+
+To cleanup the associated docker volumes, `-v` option could be used:
+
+```bash
+docker-compose down -v
+```
+
+More details can be found in [`docker-compose` documentation](https://docs.docker.com/compose/reference/).
+
+=======
```
lucafavatella: Bad formatting.

docs/docker.md
```diff
@@ -88,3 +88,53 @@ docker run -d -p 3013:3013 --hostname node0 \
 ```
 
 **Note: make sure `hostname` option is set when reusing the mnesia data directory**
+
+# Localnet
+
+Small local network (not connected to testnet) can be created with `docker-compose`.
+It runs three nodes using the `mean16s-generic` miner (fastest generic miner) and a proxy server to allow CORS.
```
lucafavatella: Is the usage of mean16 intended for short inter-block time? If yes, this works only for initial ten blocks or so, as after that the difficulty regulation kicks in, with expected 5 minutes inter-block time. See https://github.com/aeternity/epoch/blob/bc08fc66a61a6cd70f8b926107d5e5f5890dc78a/apps/aecore/src/aec_governance.erl#L25-L26 for test-only configuration variable (`expected_mine_rate`) that may be used for shortening inter-block time.

docker-entrypoint.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 # If not provided set to public facing IP address of this node
-EXTERNAL_PEER_ADDRESS="${EXTERNAL_PEER_ADDRESS:-http://$(curl -s https://api.ipify.org)}:3013/"
+EXTERNAL_PEER_ADDRESS="${EXTERNAL_PEER_ADDRESS:-http://$(curl -s https://api.ipify.org):3013/}"
```
lucafavatella: Good catch.

docs/docker.md
```diff
@@ -88,3 +88,53 @@ docker run -d -p 3013:3013 --hostname node0 \
...
+
+Both external and internal API are exposed to the docker host, the URL pattern is as follows:
```
lucafavatella: How is user meant to reach WebSocket internal endpoint (3114) of each node? (See [overview of endpoints](https://github.com/aeternity/protocol/blob/0b4b55f3a533770c82758be7b01942d5b74e8344/epoch/api/README.md#overview).)

docker-compose.yml
```diff
@@ -0,0 +1,54 @@
+# Small local network of two nodes using the fastest mean16
```
lucafavatella: 3 nodes?

docker/Dockerfile-epoch
```diff
@@ -21,12 +21,17 @@ RUN useradd --shell /bin/bash epoch \
 USER epoch
 ENV SHELL /bin/bash
 
+# Create data directories in advance so that volumes can be mounted in there
+# see https://github.com/moby/moby/issues/2259 for more about this nasty hack
+RUN mkdir -p /home/epoch/node/data/mnesia \
+    && mkdir -p /home/epoch/node/keys
```
lucafavatella: I noticed that if I run `docker-compose up -d` without `docker-compose build` then `curl http://localhost:3002/internal/v2/account/pub-key` does not work. The reason is that nodes 1 2 3 die straight away not being able to write in DB dir ("Not allowed to write in Mnesia dir"). Though `docker-compose build` includes this change in the Docker image that enables it to work. Ok.

docker/epoch_node1_mean16.yaml
```diff
@@ -0,0 +1,22 @@
+---
+peers:
```
lucafavatella: I see the 3 nodes are in sync. Good.

docker/nginx-default.conf
```diff
@@ -0,0 +1,41 @@
+server {
+    listen 3001;
+
+    location / {
+        include cors.conf;
+        proxy_pass http://node1:3013/;
```
lucafavatella: I noticed that, after having started the localnet (`docker-compose up -d`) and checked they were all reachable via curl via proxy, when I stopped all containers one by one (`docker stop epoch_node1_1`, `docker stop epoch_node2_1`, `docker stop epoch_node3_1`) and then I started the first one (`docker start epoch_node3_1`), this started container was not reachable via curl via proxy ("No route to host"). Then I started the other two nodes too and all 3 were reachable via curl.

I expect this glitch is related to Docker (Docker for Mac "Docker version 17.12.0-ce, build c97c6d6") and not actionable on our side. Something to keep in mind if users report this kind of behaviour.

docker-compose.yml
```diff
@@ -0,0 +1,54 @@
...
+    build:
+      context: .
+      dockerfile: docker/Dockerfile-epoch
+    image: aetrnty/epoch
+    hostname: node1
+    environment:
+      EPOCH_CONFIG: /home/epoch/epoch.yaml
+    volumes:
+      - ${PWD}/docker/epoch_node1_mean16.yaml:/home/epoch/epoch.yaml
+      - node1_db:/home/epoch/node/data/mnesia
```
lucafavatella: I tested that persisted DB in docker-compose works. I have not tested that bind mount in https://github.com/aeternity/epoch/blob/067d8dc15a2f76f3d2b38e087020bfa2bf75a905/docs/docker.md#persisting-data was not broken.

docker/nginx-default.conf
dincho: Well, this setup intended usage is via `docker-compose` only, if you hack around with the containers itself with `docker` command only you're responsible yourself to setup the networking and port mapping.

docs/docker.md
dincho: See https://www.pivotaltracker.com/story/show/154904411

docker/Dockerfile-epoch
dincho: Indeed you need the latests image of build by this Dockerfile.

tillkolter:
Good job @dincho! I was able to build and up everything and the three nodes are mining.   

For æpp development purposes, I would still wish that at least one of these three nodes exposes a websocket port, so that community developers could start testing and developing oracles right away. 

And to test requests against the internal ports at least one of these should be exposed as well imho.

tillkolter:
For my dev server websocket, I set up this Nginx config below. 

Dont know if you were planning to integrate that or if internal security guidelines forbid that, but at least for a public oracle demo, this would help a lot.

```
map $http_upgrade $connection_upgrade {
    default upgrade;
    '' close;
}

upstream websocket {
   server localhost:3124;
}

server {
   # Use the standard internal epoch port
   listen 3104;
   location / {
      more_set_headers 'Access-Control-Allow-Origin: *';

      proxy_pass http://websocket;
      proxy_http_version 1.1;

      proxy_set_header Upgrade "websocket";
      proxy_set_header Connection "upgrade";

      proxy_pass_request_headers      on;

   }
   proxy_connect_timeout 7d;
   proxy_send_timeout 7d;
   proxy_read_timeout 7d;

}
```

dincho:
> For æpp development purposes, I would still wish that at least one of these three nodes exposes a websocket port, so that community developers could start testing and developing oracles right away.

Working on it, currently the websocket listen address cannot be configured and it's 127.0.0.1, the nginx proxy is on separate container and cannot access it ATM.

dincho:
> And to test requests against the internal ports at least one of these should be exposed as well imho.



> And to test requests against the internal ports at least one of these should be exposed as well imho.

Do you mean the `internal API` ? It should be exposed at `/internal` path, could you try `http://localhost:3001/internal/account/balance` ?

docker-compose.yml
dincho: fixed in c928c53

docs/docker.md
dincho: fixed in c928c53

docs/docker.md
dincho: done in c928c53

docs/docker.md
dincho: I added a row in c928c53 and removed old stuff, however I could come in with something good for the very first removed it until the actual release.

docker-compose.yml
dincho: It's hard to tell because the image is suitable for UAT only and hard to join the INT with it. As if you build your own, you'll have the `master` branch source code with it.

dincho: I mean, you have to change the genesis accounts too.

docs/docker.md
dincho: The only problem that I see to ship a docker specific sys.config (and change this) is that it will will often go out of sync with the src sys.config. Any idea how to solve this?

lucafavatella: @dincho 
```
$ ./_build/prod/rel/epoch/bin/epoch start

$ ./_build/prod/rel/epoch/bin/epoch eval 'application:get_env(aecore, expected_mine_rate).'
undefined

$ ./_build/prod/rel/epoch/bin/epoch stop
ok

$ ./_build/prod/rel/epoch/bin/epoch start -aecore expected_mine_rate 100

$ ./_build/prod/rel/epoch/bin/epoch eval 'application:get_env(aecore, expected_mine_rate).'
{ok,100}
```

dincho: That looks great

lucafavatella:
I see CI unrelated error `Supervisor exec_app had child exec started with exec:start_link([]) at <0.20820.0> exit with reason killed in context shutdown_error`. @dincho I will restart job in 2 minutes.

docker/Dockerfile-epoch
amir-hadi: Little suggestion here: Maybe we can mark the directories as volumes? As these directories are important, they should not be deleted by docker without the users consent (docker rm -v), right?

docker-compose.yml
```diff
@@ -0,0 +1,54 @@
...
+      - node1_keys:/home/epoch/node/keys
+
+  node2:
```
amir-hadi: I made some tests about removing the node2 and node3 explicit configurations here, so there would only be a `service` called `node`. The user of this could then scale to any number of nodes he wants. It worked basically, but it would break the setup with NGINX. Also as (I now know) a usual setup always consists of 3 nodes for developers, so my idea may be useful it later deployments in large scale.

docs/docker.md
```diff
@@ -88,3 +88,52 @@ docker run -d -p 3013:3013 --hostname node0 \
...
+
+```bash
+docker-compose down -v
+```
+
+More details can be found in [`docker-compose` documentation](https://docs.docker.com/compose/reference/).
+
+## Image Version
+
+Docker compose uses the `aetrnty/epoch:latest` image, it will be pulled from [docker hub](https://hub.docker.com/r/aetrnty/builder/) if it's not found locally.
```
lucafavatella: Comment says aetrnty/epoch though link is aetrnty/builder ???

docs/docker.md
dincho: @lucafavatella I added this option in cea0a63

dincho: in addition 30b16ce

PR #682, opened 2018-02-02
Add Status code to return value for further validation

ThomasArts:
Since we don't check status code of response, we may mix up bodies with wrong status code with bodies of correct responses. That means, we can interpret {ok, 500, #{}} as {ok, 200, Block}.
Swagger type checking should probably do a better job, but for now, this is a fix to not be crashed by a hostile node that sends us rubish.


lucafavatella:
A job failed for known error, job restarted.
```
%%% aec_eunit_SUITE ==> {errors_in_lager_log,["14:11:41.849 [error] Supervisor exec_app had child exec started with exec:start_link([]) at <0.20764.0> exit with reason killed in context shutdown_error\n\n"]}
```

PR #685, opened 2018-02-05
Add WebSocket subscribeable event on new block

hanssv:
Generalize the WS subscribe mechanism
Update the tests accordingly
Add unsubscribe for WS subscription mechanism

https://www.pivotaltracker.com/story/show/154081927

[Finishes PT#154081927]

### DIFF
--- a/apps/aehttp/src/ws_handler.erl
+++ b/apps/aehttp/src/ws_handler.erl
@@ -17,9 +17,6 @@
 
 -define(GPROC_KEY, {p, l, {?MODULE, broadcast}}).
 
--define(SUBSCRIBE_EVENTS, [ block_created              %% node mined a block
-                          ]).
-
 -type id() :: pid().
 -export_type([id/0]).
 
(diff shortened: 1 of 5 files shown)


apps/aehttp/src/ws_handler.erl
```diff
@@ -17,26 +17,17 @@
...
 -type id() :: pid().
 -export_type([id/0]).
 
 init({tcp, http}, _Req, _Opts) ->
     {upgrade, protocol, cowboy_websocket}.
 
 websocket_init(_TransportName, Req, _Opts) ->
-    {QsVals, Req1} = cowboy_req:qs_vals(Req),
-    SubscribeEvents = subscribe_events(QsVals),
+    {_QsVals, Req1} = cowboy_req:qs_vals(Req),
```
velzevur: Do we still need this?

hanssv: No clue, since the variable was Req1 I thought it was safest to keep it...

velzevur: Cowboy has this thingy about always using the last `cowboy_req` _object_. You can read about it [here](https://ninenines.eu/docs/en/cowboy/1.0/guide/req/).
```
Almost all the calls to the cowboy_req module will return an updated request object.
Just like you would keep the updated State variable in a gen_server, you MUST keep
the updated Req variable in a Cowboy handler. Cowboy will use this object to know
whether a response has been sent when the handler has finished executing.
```
So you can safely remove the line with the `qs_vals` and use `Req` instead of `Req1`.
No update required, though.

PR #689, opened 2018-02-05
Pt 154921144 extend mpt iterator

gorillainduction:
Extend MPT (and aeu_mtrees) with iterator for traversing only to a specified path depth.

This is handy for traversing all oracles or all contracts (and not queries/contract_calls).

Also, fixed some corner cases where we couldn't start from a non-existing key in the traversal (as you can in ets/mnesia).

https://www.pivotaltracker.com/n/projects/2124891/stories/154921144

### DIFF
--- a/apps/aeutils/src/aeu_mp_trees.erl
+++ b/apps/aeutils/src/aeu_mp_trees.erl
@@ -15,7 +15,9 @@
         , delete/2
         , get/2
         , iterator/1
+        , iterator/2
         , iterator_from/2
+        , iterator_from/3
         , iterator_next/1
         , pp/1
         , put/3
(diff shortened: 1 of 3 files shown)


apps/aeutils/src/aeu_mp_trees.erl
```diff
@@ -159,19 +164,27 @@ verify_proof(Key, Val, Hash, ProofDB) ->
 iterator(#mpt{hash = Hash, db = DB}) ->
     #iter{key = <<>>, root = Hash, db = DB}.
```
hanssv: More elegant(?) to say `iterator(Tree, [])`

apps/aeutils/src/aeu_mp_trees.erl
```diff
@@ -159,19 +164,27 @@ verify_proof(Key, Val, Hash, ProofDB) ->
...
 
+-spec iterator(tree(), iterator_opts()) -> iterator().
+iterator(#mpt{hash = Hash, db = DB}, Opts) ->
+    process_iterator_opts(#iter{key = <<>>, root = Hash, db = DB}, Opts).
+
 %%% @doc Iterator from a key. Key doesn't need to exist. Calling
 %%% iterator_next/1 gives the next value after Key.
 -spec iterator_from(key(), tree()) -> iterator().
 iterator_from(Key, #mpt{hash = Hash, db = DB}) ->
     #iter{key = Key, root = Hash, db = DB}.
```
hanssv: More elegant(??) to say `iterator_from(Key, Tree, [])`

PR #690, opened 2018-02-05
HTTP API endpoint: get list of connected nodes/peers

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/154658433)
[Github issue](https://github.com/aeternity/protocol/issues/30)
### Endpoint
Since there might be **a lot of** connected peers, this is a debug API, one should explicitly opt in to use it.
```
GET /debug/peers
```
### Examples
To turn it on, one should set `aehttp`'s app setting `enable_debug_endpoints` to true. It is false by default.
The following examples were taken from CT logs.

#### API disabled
```
GET "http://127.0.0.1:3113/v2/debug/peers"
Return code 403, Body "{\"reason\":\"Call not enabled\"}"
```

#### Peers list
```
GET "http://127.0.0.1:3113/v2/debug/peers"
Return code 200, Body "{\"blocked\":[],\"peers\":
[{\"last_seen\":1517844051766,\"uri\":\"http://someone.somewhere:1337\"},
{\"last_seen\":1517844051770,\"uri\":\"http://someonelse.somewhereelse:1337\"}]}"
```
`peers` is an array of objects, each one containing the peer `uri` and the `last_seen` timestamp. If the timestamp is 0, the peer had never been actually reached.

#### Blocked list
```
GET "http://127.0.0.1:3113/v2/debug/peers"
Return code 200, Body "{\"blocked\":
[\"http://blocked.peer:1337\",
\"http://otherblocked.peer:1337\"],
\"peers\":[]}"
```
`blocked` is an array of blocked peer `uri`-s.


velzevur:
For some reason the embeded definition of the peers was causing Swagger-codegen to produce broken python code.

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -411,6 +410,23 @@ handle_request('GetAccountBalance', Req, _Context) ->
...
+                    fun({URI, LastSeen}) ->
+                        #{uri => URI,
+                          last_seen => LastSeen}
+                    end,
+                    aec_peers:all()),
+            Blocked = aec_peers:blocked(),
+            {200, [], #{peers => Peers,
+                        blocked => Blocked}};
+        false ->
+            {403, [], #{reason => <<"Call not enabled">>}}
```
hanssv: "Call not enabled" sounds a bit strange, but I have no obviously better error message

velzevur: I guess we'd have some other debug calls and they could probably share the same response message.

velzevur:
Rebased on top of mater and resolved conflicts.

velzevur:
Rebased again on top of mater and resolved conflicts.

PR #692, opened 2018-02-06
Pt 154867174 calldata endpoint

happi:
Add an HTTP endpoint for encoding calldata.
Also add missing call contranct handler.

I wrote: "I am not 100% confident that I got the correct files added after swagger-codegen so please check that a bit extra."
Now, I am more confident.

[Finishes #154867161 #154867174 ]


config/swagger.yaml
```diff
@@ -1191,6 +1191,39 @@ paths:
...
+            properties:
+              code:
+                type: string
+              function:
+                type: string
+              arg:
+                type: string
+      responses:
+        '200':
+          description: Binary ecoded calldata
```
hanssv: s/ecode/encode

PR #696, opened 2018-02-06
Pt 154171039 business logic

ThomasArts:
Check that we are not getting more peers than we want to.

[Delivers #154171039]

### DIFF
--- a/apps/aeutils/src/aeu_requests.erl
+++ b/apps/aeutils/src/aeu_requests.erl
@@ -50,7 +50,7 @@
 
 -type response(Type) :: {ok, Type} | {error, string()}.
 
--spec ping(http_uri:uri(), map()) -> response(map()).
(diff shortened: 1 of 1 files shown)


apps/aeutils/src/aeu_requests.erl
```diff
@@ -84,8 +82,13 @@ ping(Uri, LocalPingObj) ->
                   RemoteObj = Map#{<<"genesis_hash">> => RemoteGHash,
                                    <<"best_hash">>  => RemoteTopHash},
                   RemotePeers = maps:get(<<"peers">>, Map, []),
-                  lager:debug("ping response (~p): ~p", [Uri, pp(RemoteObj)]),
-                  {ok, RemoteObj, RemotePeers};
+                  case maps:get(<<"peers">>, Map, []) of
+                      RemotePeers when is_list(RemotePeers), length(RemotePeers) < Share ->
```
hanssv: I had expected =< ?

PR #698, opened 2018-02-07
upgrade setup [Fixes #154990570]

uwiger:
> A crash has been observed in testing, as the setup application tries to patch in the sasl application. This is a minor bug in the setup app, fixed in version 2.0.1.

Example crash:
```
%%% aecore_sync_SUITE ==> {{badmatch,
     {badrpc,
         {'EXIT',
             {function_clause,
                 [{filename,join,[[]],[{file,"filename.erl"},{line,399}]},
                  {setup,'-roots_of/1-fun-0-',2,
                      [{file,
                           "/home/builder/epoch/_build/default/lib/setup/src/setup.erl"},
                       {line,733}]},
```

The bug in setup was triggered by the addition of the patch path `"patches/ebin"`.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -3,7 +3,7 @@
                                  epoch_metrics]}]}.
 
 {deps, [{lager, ".*", {git, "https://github.com/aeternity/lager.git", {tag,"3.6.0.ae1"}}},
-        {setup, "2.0.0"},
+        {setup, "2.0.2"},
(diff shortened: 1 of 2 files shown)


uwiger:
I noticed that the `2.0.1` version tag on setup was misapplied. Fixed, and bumped setup to `2.0.2`, squashed and force-pushed.

PR #699, opened 2018-02-07
Add JRE and libsodium to builder docker image

dincho:


dincho:
> I understand JRE needed for make swagger-check in CI. Why libsodium-dev?

@hanssv Needs libsodium for his blake2 work

> Is this Docker image already published somewhere?

yes at https://hub.docker.com/r/aetrnty/builder/

> Is this Docker image intended to be published somewhere? Where? If it is the published builder image, could you please rename the var used at

Let's track and decide later because we could also use `FROM aetrnty/builder as builder` as an option?

lucafavatella:
> Let's track and decide later because we could also use FROM aetrnty/builder as builder as an option?

Ok. Tracked as https://www.pivotaltracker.com/story/show/154993662

deployment/Dockerfile
```diff
@@ -19,6 +19,14 @@ RUN OTP_DOWNLOAD_URL="https://github.com/erlang/otp/archive/OTP-${OTP_VERSION}.t
     && export ERL_TOP=`pwd` \
     && ./otp_build autoconf && ./configure && make -j$(nproc) && make install
 
+ENV LIBSODIUM_VERSION=1.0.16
+RUN LIBSODIUM_DOWNLOAD_URL="https://github.com/jedisct1/libsodium/releases/download/${LIBSODIUM_VERSION}/libsodium-${LIBSODIUM_VERSION}.tar.gz" \
+    && curl -fsSL -o libsodium-src.tar.gz "$LIBSODIUM_DOWNLOAD_URL" \
+    && mkdir libsodium-src \
+    && tar -zxf libsodium-src.tar.gz -C libsodium-src --strip-components=1 \
+    && cd libsodium-src \
+    && ./configure && make -j$(nproc) && make install && ldconfig
```
lucafavatella: I assume libsodium-src and libsodium-src.tar.gz could be deleted?
(And similarly for OTP above?)

dincho: Yes, probably. I don't focus too much on the container size ATM.

I'll merge it right way (as it's approved), but we could work on container size improvements. @amir-hadi might be interested in this as well.

dincho:
CircleCI is greed. merging.

PR #701, opened 2018-02-07
introduce mnesia_rocksdb backend

uwiger:
The original ticket was aimed at introducing a Leveldb backend, but experiments indicated that it might be harder to get the leveldb backend working on newer OTP versions than porting the existing backend to rocksdb (which is also actively supported).

https://www.pivotaltracker.com/story/show/154872636


rebar.lock
```diff
@@ -79,6 +85,8 @@
  {<<"msgpack">>, <<"128AE0A2227C7E7A2847C0F0F73551C268464F8C1EE96BFFB920BC0A5712B295">>},
  {<<"parse_trans">>, <<"1BAD3B959941CC53FFD6F4769A5D2666F9CDF179B2D62826636497D3FBAD9EC0">>},
  {<<"poolboy">>, <<"6B46163901CFD0A1B43D692657ED9D7E599853B3B21B95AE5AE0A777CF9B6CA8">>},
+ {<<"rocksdb">>, <<"C92B48703D4812C8BC571E0FBB7681F0899F35C4E4330F1CF646D79357A6AFE4">>},
  {<<"setup">>, <<"1203F4CDA11306C2E34434244576DED0A7BBFB0908D9A572356C809BD0CDF085">>},
+ {<<"sext">>, <<"14559B219FAF11F9EBA7A350DFFC69C966C668439C54947A42D67167002798C4">>},
```
hanssv: Do we dare using this? I have seen failing properties :-)

hanssv: We should be fine unless we confuse 0 with -0 and 0.0...

uwiger: hehe, that's what you get when you let QuickCheck people into your project. :)

PR #705, opened 2018-02-07
Update docs from CircleCI

michalbagrowski:


.circleci/config.yml
```diff
@@ -156,6 +156,14 @@ jobs:
           root: *packages_workspace
           paths:
             - "*.tar.gz"
+  docs:
+    <<: *container_config
+    steps:
+      - checkout
+      - run:
+          command: |
+            make swagger-docs
```
dincho: I'd suggest to move this to the build step and always publish it as CircleCI artifacts.

After that in the `docs` step (I'd also suggest renaming it to `deploy_api_docs` sync the generated docs with the remote docs repo.

.circleci/config.yml
```diff
@@ -212,6 +220,14 @@ workflows:
                 - env/dev2
             tags:
               only: /^v.*$/
+      - docs:
+          requires:
+            - build
+          filters:
```
dincho: Filters are ok.

Makefile
```diff
@@ -173,6 +173,26 @@ swagger: config/swagger.yaml $(SWAGGER_CODEGEN_CLI)
 swagger-docs:
 	(cd ./apps/aehttp && $(MAKE) swagger-docs);
 
+SWAGGER_STATUS = $(shell cd /tmp/epoch-api-docs/ && git status --porcelain)
```
dincho: This process is CI only, let's not pollute the main Makefile with it. `deployment/deploy_api_docs.sh` sounds much better place to me.

Makefile
```diff
@@ -173,6 +173,26 @@ swagger: config/swagger.yaml $(SWAGGER_CODEGEN_CLI)
...
+swagger-publish-docs: swagger-publish-docs-init
+	cd /tmp/epoch-api-docs/; \
+	git add swagger.json; \
+	git config user.email "circleci@aeternity.com"; \
+	git config user.name "CircleCi Aeternity"; \
+	if [ -z "$(SWAGGER_STATUS)" ]; then \
+		echo "Nothing to commit"; \
+	else \
+		git commit -a -m 'Update epoch docs CircleCI'; \
+		git push origin master; \
```
dincho: Pushing master to master is OK, but we laso have to tag this repo as well when the main repo is tagged

Makefile
```diff
@@ -173,6 +173,26 @@ swagger: config/swagger.yaml $(SWAGGER_CODEGEN_CLI)
 swagger-docs:
 	(cd ./apps/aehttp && $(MAKE) swagger-docs);
 
+SWAGGER_STATUS = $(shell cd /tmp/epoch-api-docs/ && git status --porcelain)
+
+swagger-publish-docs-init:
+	git clone git@github.com:aeternity/epoch-api-docs.git /tmp/epoch-api-docs
+	cp apps/aehttp/priv/swagger-docs/swagger.json /tmp/epoch-api-docs/
```
dincho: I don't think the swagger.json is enough, about about the generated documentation itself? Am I missing something?

michalbagrowski: based on comment in pivotal task

```
@dimitarivanov
Once the files are installed, we don't have to update them, except for swagger.json. It has all the information required for the docs and we should sync it.```

File: Makefile (hunk 4 above)
Comment (michalbagrowski) [reply Makefile]: duty noted

File: Makefile (hunk 3 above)
Comment (michalbagrowski) [reply Makefile]: duty noted

File: .circleci/config.yml (hunk 1 above)
Comment (michalbagrowski) [reply .circleci/config.yml]: ok will dig around docs to see how setup this as circle ci artifact

File: Makefile (hunk 5 above)
Comment (dincho) [reply Makefile]: Seems legit, I missed that 👍

File: .circleci/config.yml (hunk 1 above)
Comment (michalbagrowski) [reply .circleci/config.yml]: @dincho not sure this can work as CircleCi artifacts of i understand this correct, this will store docs in CircleCi, not push them to github.

Comment (dincho) [reply .circleci/config.yml]: yes, we just want to see the docs for each commit, but publish to gh-pages only master and tags

Comment (michalbagrowski) [reply .circleci/config.yml]: oh ok make sense

File: deployment/deploy_api_docs.sh [hunk 6]
```diff
@@ -0,0 +1,19 @@
...
+
+git add swagger.json;
+git config user.email "circleci@aeternity.com";
+git config user.name "CircleCi Aeternity";
+if [ -z `git status --porcelain` ]; then
+    echo "Nothing to commit";
+else
+    git commit -a -m 'Update epoch docs CircleCI';
+    git push origin master;
+    git tag $TAG_MAIN
```
dincho: As far as I can understand this it retag latest tag with the latest commit ?

I think you should use and check `CIRCLE_TAG` from https://circleci.com/docs/2.0/env-vars/#build-details
Somethign like:
```
if $CIRCLE_TAG
git tag $CIRCLE_TAG
git push origin $TAG_MAIN
else
git commit -a -m 'Update epoch docs CircleCI'
git push origin master
fi

File: .circleci/config.yml (hunk 1 above)
Comment (dincho) [reply .circleci/config.yml]: Never-mind, i don't think it will be usable.

File: deployment/deploy_api_docs.sh (hunk 6 above)
Comment (dincho) [reply deployment/deploy_api_docs.sh]: I'm not still sure what should be pushed to the docs master.
1. epoch master docs
2. epoch lastest release docs

@sennui @lucafavatella What do you think?

Comment (michalbagrowski) [reply deployment/deploy_api_docs.sh]: yes, CIRCLE_TAG will be better option here. Will fix. also yes, when and what should be pushed to docs :D and when ;]

File: .circleci/config.yml (hunk 1 above)
Comment (michalbagrowski) [reply .circleci/config.yml]: @dincho i almoust finished implementation :P

Comment (michalbagrowski) [reply .circleci/config.yml]: cleanup ;]

File: deployment/deploy_api_docs.sh (hunk 6 above)
Comment (sennui) [reply deployment/deploy_api_docs.sh]: Given that we have single docs repo that gets hosted as official doc it has to be the latest official release.

I wonder how we can help out our aepps team that would like to use newer version that is not yet released. API repo per environment & push to API repo during deployment work?

Comment (lucafavatella) [reply deployment/deploy_api_docs.sh]: @dincho I would expect master to contain the docs generated from epoch master.

Though I would expect the docs of each release to be pushed as well - with tag (or branch - whatever easier) name inferred from epoch tag names.

Comment (michalzajda) [reply deployment/deploy_api_docs.sh]: @lucafavatella unfortunately only one version is pretty printed

Comment (michalbagrowski) [reply deployment/deploy_api_docs.sh]: @dincho updatet to user CIRCLE_TAG if exists to make commit tag to api-docs repo

Comment (dincho) [reply deployment/deploy_api_docs.sh]: @michalzajda we can hack on this with some 3td party tools like:
- https://htmlpreview.github.io
- https://rawgit.com

**Comment by sennui** [comment]
please squash before merging

**Comment by michalbagrowski** [comment]
@dincho @michalzajda @lucafavatella updated it will now, push docs from `master epoch` -> `master epoch-api-docs`, and any TAG will go to TAG, i think main issue is that with github pages, it create website from https://aeternity.github.io/epoch-api-docs/ ..

File: .circleci/config.yml [hunk 7]
```diff
@@ -212,6 +220,16 @@ workflows:
                 - env/dev2
             tags:
               only: /^v.*$/
+      - deploy_api_docs:
+          requires: ~
```
lucafavatella: What does this line mean?

lucafavatella:
> ... it create website from https://aeternity.github.io/epoch-api-docs/ ..

I assume that is expected that that link shows a broken page at the moment because of https://github.com/aeternity/epoch-api-docs/blob/6f9b28cb816e0cf00b18431cf5271ed5d71c5d2c/swagger.json#L1 and that it is meant to get fixed once this PR merged on master.

deployment/deploy_api_docs.sh
```diff
@@ -0,0 +1,26 @@
...
+if [ -z "$STATUS" ]; then
+    echo "Nothing to commit, docs did not change";
+else
+    git commit -a -m 'Update epoch docs CircleCI';
+
+    if [ "$CIRCLE_BRANCH" == "master" ]; then
+        git push origin master
+    elif [ -n "$CIRCLE_TAG" ]; then
+        git tag $CIRCLE_TAG
+        git push origin $CIRCLE_TAG
```
lucafavatella: These two `$CIRCLE_TAG` may misbehave for tags with spaces in them... though only theoretical as we do not use spaces.

dincho:
> I wonder how difficult it is to improve that UI upstream project in this point by teaching it to take file from query param.

Sounds like a good idea and then just passing the tagged config like: https://aeternity.github.io/epoch-api-docs?config=https://github.com/aeternity/epoch-api-docs/blob/v0.6.0-56-g87e8290/swagger.json

dincho:
> Not sure I got the full current context right for this PR though blocking it for now as I understand we are aiming to aeternity/epoch-api-docs#1 (comment)

Yes indeed.

@michalbagrowski let's push the epoch/swagger.json to api-docs/master only on epoch repo tag. We don't need master to master anymore.

acidsennui:
@dincho just to confirm you wish to have epoch tag -> api-doc master?

dincho:
> @dincho just to confirm you wish to have epoch tag -> api-doc master?

Yes.

.circleci/config.yml
```diff
@@ -212,6 +220,16 @@ workflows:
...
+          requires:
```
dincho: I don't think you need the build? We probably need `static_analysis` requirement to ensure the swagger docs are in sync before publishing cc @lucafavatella

.circleci/config.yml
```diff
@@ -212,6 +220,16 @@ workflows:
...
+            - build
+          filters:
+            branches:
+              only: master
```
dincho: We don't need master (any branch) builds:

```
            branches:
              ignore: /.*/
```

acidsennui:
@dincho updated

deployment/deploy_api_docs.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -ex
+git clone git@github.com:aeternity/epoch-api-docs.git /tmp/epoch-api-docs
+cp apps/aehttp/priv/swagger-docs/swagger.json /tmp/epoch-api-docs/
+cd /tmp/epoch-api-docs/
+
+git add swagger.json;
+
+git config user.email "circleci@aeternity.com";
```
dincho: I configured this as CircleCI environment variables, let's try removing it

deployment/deploy_api_docs.sh
```diff
@@ -0,0 +1,22 @@
...
+git config user.name "CircleCi Aeternity";
+
+STATUS=`git status --porcelain`
+
+if [ -z "$STATUS" ]; then
+    echo "Nothing to commit, docs did not change";
+else
+    if [ -n "$CIRCLE_TAG" ]; then
```
dincho: if you configure the CircleCI as mentioned in my previous comment, you shouldn't need this branching (check) anymore.

acidsennui:
@dincho updated

acidsennui:
@dincho updated, waiting for info does it require `build` or something else ;]

.circleci/config.yml
lucafavatella: Spec is [here](https://circleci.com/docs/2.0/configuration-reference/#requires), no mention of `~`
Let me see YAML spec

lucafavatella: I see it also [here](https://github.com/aeternity/epoch/blob/66ea4da55bb919eac10985c53ac920b46fba6cc3/.circleci/config.yml#L246) though not for linux_package

.circleci/config.yml
lucafavatella: Indeed depending on swagger-check would help

.circleci/config.yml
```diff
@@ -156,6 +156,14 @@ jobs:
           root: *packages_workspace
           paths:
             - "*.tar.gz"
+  deploy_api_docs:
+    <<: *container_config
+    steps:
+      - checkout
+      - run:
+          command: |
+            make swagger-docs
```
lucafavatella: Why is this `make swagger-docs` needed - if only the swagger.json pushed?

dincho:
> make swagger-docs needed at all?
> waiting for job containing swagger-check to succeed before pushing?

Arent we going to remove the generated swagger stuff ?

lucafavatella:
> Arent we going to remove the generated swagger stuff ?

@dincho  Not the swagger.json (well there are two swagger.json - I expect this script can take the one generated by `make swagger` that is already under version control and checked by `make swagger-check`.)

.circleci/config.yml
dincho: @lucafavatella It means "null" AFAIK, I think the "requires" is required key
 still this line is obsolete and copy-paste leftover probably

.circleci/config.yml
dincho: Yep, but we don't have separate job for it, so requirement should go to `static_analysis` however it implicitly requires `build`

.circleci/config.yml
dincho: Yep @michalbagrowski we don't need this one.

.circleci/config.yml
dincho: @michalbagrowski ^^

dincho:
> @dincho updated, waiting for info does it require build or something else ;]

@acidsennui https://github.com/aeternity/epoch/pull/705#discussion_r168133134

dincho:
@acidsennui What's the reason this PR is not build by CircleCI?

dincho:
@acidsennui Looks unavailable ATM, so I'm merging this myself.

PR #709, opened 2018-02-08
PT-154838110 Enable logging on console

amir-hadi:
This allows docker users to see log output directly using dockers logging features.


dincho:
I believe this is proposition for https://www.pivotaltracker.com/story/show/154838110

amir-hadi:
@lucafavatella Maybe this was to naive, I am not (yet?) an Erlang developer. I do not understand why console logging should be slower or blocking more than logging to a file, but as far as I understood you, this is just about the miner right? We could disable the logging for the miner and let the rest log to the console?

Logging to console is something really important when you want to use containerized environments. I believe we could do another workaround which I just found, but that seems very hacky to me. We can create a symbolic link of /dev/stdout to the appropriate log file, then docker will be able to read the logs, even if they are in a file.

dincho:
> We can create a symbolic link of /dev/stdout to the appropriate log file,

Already tried, didn't make it work, also lager is doing log rotations itself, which should be considered too.

lucafavatella:
> Maybe this was to naive, I am not (yet?) an Erlang developer. I do not understand why console logging should be slower or blocking more than logging to a file, but as far as I understood you, this is just about the miner right? We could disable the logging for the miner and let the rest log to the console?

@amir-hadi It was about the epoch_mining log file (by miner Erlang module - analogous of current conductor).

I have nothing against this change. Though - given experience in https://www.pivotaltracker.com/story/show/153173585/comments/182923773 - I would like to see this configuration change (or refined - e.g. not sure why logging metrics to console/stdout) running continuously in an environment (not integration) for two days before the release (shall be merged before that though). Cc @dincho @sennui  @happi @uwiger

amir-hadi:
Once I have access to the environments I will push this branch and let it run for some days. Then we will have hands on data to check if its running or not.

PR #711, opened 2018-02-08
Teach Circle CI to store node logs for smoke test on package

lucafavatella:

Background is [failure of CI in master on Thu 8th Feb morning](https://www.pivotaltracker.com/story/show/154770099/comments/185698992).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,11 +33,13 @@ references:
         mkdir ${PACKAGES_DIR:?}
         mv _build/prod/rel/epoch/epoch-$(cat VERSION).tar.gz ${PACKAGES_DIR:?}/epoch-$(cat VERSION)-${PKG_SUFFIX}.tar.gz
 
+  package_tests_workspace: &package_tests_workspace /tmp/package_tests
   test_package: &test_package
     run:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -46,7 +48,12 @@ references:
         fi
         epmd -daemon
         make python-env
-        make python-release-test TARBALL=${PACKAGES_DIR:?}/epoch-$(cat VERSION)-${PKG_SUFFIX}.tar.gz
+        mkdir ${PACKAGE_TESTS_DIR:?}
```
lucafavatella: @dincho I suspect a potential failure on this mkdir when running both linux_package and osx_package though based on my tests I am reasonably confident there is no conflict. I aim at definitely proving it with [this build that is the same branch as this plus the osx_package job enabled](https://circleci.com/gh/lucafavatella/epoch/273).

lucafavatella: Proved.

lucafavatella:
> ... requiring empty dir looks weird to me

@dincho The script deploys in that dir 3 nodes so the check on dir being empty is for preventing potential spurious existing content to create silent non-determinism in the test.

PR #712, opened 2018-02-08
Fix intermittent package test failing for node startup issue

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/154770099


sennui:
It looks safe. This was added to cover unit tests which used default OS OTP, and in consequence root dir was in restricted location. After your change keys will go to data dir, which is relatively new and safe development

PR #714, opened 2018-02-08
Test aec_keys startup for non-determinism in key generation

lucafavatella:
Coded while preparing https://github.com/aeternity/epoch/pull/712 - while aiming at test for reproducing node startup issue though issue was somewhere else so this test is not a regression test.


lucafavatella:
I see smoke test of package on Circle CI failed - probably for same reason fixed in https://github.com/aeternity/epoch/pull/712
Restarting job.

lucafavatella:
I see 2 approvals. CI passed. Merging.

PR #715, opened 2018-02-08
Change hash function to BLAKE2 (almost) everywhere

hanssv:
The exceptions are EVM that uses SHA3 in two places, and key-handling that uses SHA256 (inside the Erlang crypto module).

Introduce aec_hash:hash(HashType, Binary) - where HashType is one of
pubkey, header, tx, pow, aens, evm. All types except evm is mapped to a
Blake2 hash, evm still uses sha3.
    
Removes the alternative PoW (sha256) - it hasn't been used in a long time.

Updates RELEASE_NOTES to reflect the new dependency on `libsodium` increase PROTOCOL_VERSION/GENESIS_VERSION and bumps deployment DB-path.


rebar.config
```diff
@@ -26,6 +26,7 @@
         {setup, "2.0.2"},
         {parse_trans, "3.1.0"},
         {mnesia_rocksdb, ".*", {git, "https://github.com/aeternity/mnesia_rocksdb.git", {ref,"ff823c4"}}},
+        {enacl, ".*", {git, "https://github.com/jlouis/enacl.git", {ref, "c8403ab"}}},
```
sennui: Maybe lets fork it to ae organisation in github?

RELEASE-NOTES.md
```diff
@@ -27,8 +27,12 @@ The binaries are tested on the following platforms:
 * macOS Sierra (x86-64);
 * macOS High Sierra (x86-64).
 
-The macOS package has a hard dependency on OpenSSL v1.0.0 installed with [Homebrew](https://brew.sh/) in its default path `/usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib`.
-In case you have installed it in a non-default path, you could use a symlink to work around the issue.
+The macOS package has a hard dependency on OpenSSL v1.0.0 installed with [Homebrew](https://brew.sh/) in its default path `/usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib` and also a hard
+dependency on libsodium v1.0.16 installed with [Homebrew](https://brew.sh/) also in its default path `/usr/local/opt/libsodium/lib/libsodium.23.dylib`.
+In case you have installed either of them in a non-default path, you could use symlink(s) to work around the issue.
+
+To run on Ubuntu 16.04 you need to have a libsodium shared libriary (v1.0.16) in `/usr/local/lib/libsodium.so.23`. (`wget https://download.libsodium.org/libsodium/releases/libsodium-1.0.16.tar.gz`,
```
uwiger: libriary -> library

apps/aecore/src/aec_base58c.erl
```diff
@@ -63,8 +63,8 @@ split(Bin) ->
     binary:split(Bin, [<<"$">>], []).
 
 check_str(Bin) ->
-    <<C:32/bitstring,_/binary>> = crypto:hash(sha256,
-                                              crypto:hash(sha256, Bin)),
+    <<C:32/bitstring,_/binary>> =
+        aec_hash:blake2s_hash(aec_hash:blake2s_hash(Bin)),
```
uwiger: Is this correct though? The base58Check encoding is defined as using SHA256.
https://en.bitcoin.it/wiki/Base58Check_encoding

hanssv:
Let's stick to the specification for the base58 check... (and fixed spelling)

hanssv:
BTW, the Travis CI checks will not pass for this PR... Travis doesn't know about `libsodium`

apps/aecore/test/aec_mining_tests.erl
```diff
@@ -139,27 +129,20 @@ difficulty_recalculation_test_() ->
...
-setup(PoWMod) ->
-    case PoWMod of
-        aec_pow_cuckoo ->
-            ok = meck:new(aeu_env, [passthrough]),
-            aec_test_utils:mock_fast_and_deterministic_cuckoo_pow(),
-            ok = application:ensure_started(erlexec);
-        aec_pow_sha256 ->
-            ok
-    end,
+     }].
```
lucafavatella: Now enclosing list not needed any longer.

apps/aecore/src/aec_txs_trees.erl
```diff
@@ -46,5 +46,5 @@ enter(K, V, T) ->
 
 enter_signed_tx(SignedTx, TxsTree) ->
     V = aec_tx_sign:serialize_to_binary(SignedTx),
-    K = aec_sha256:hash(V),
+    K = aec_hash:hash(tx, V),
```
lucafavatella: May I suggest signed_tx (or similar) rather than tx here? I would prefer keeping distinction of tx from the signed one for readability.

hanssv:
Yes, aec_peers was skipped on purpose, the idea was to replace SHA256 with Blake2b - MD4 should be adequate for this task. (And also the whole peer mechanism will be reworked so lets not change what is already scheduled to be changed...)

hanssv:
Rebased on latest master, squashed.

Fixed the issue with `Dockerfile` - and added enough extra work for Travis to be happy.

apps/aecore/src/aec_txs_trees.erl
```diff
@@ -15,7 +15,7 @@
 -include("common.hrl").
 -include("blocks.hrl").
 
--type key() :: aec_sha256:hash(value()).
+-type key() :: aec_hash:hash(tx, value()).
```
uwiger: Shouldn't this be `aec_hash:hash(signed_tx, value())`?

apps/aeutils/src/aeu_mp_trees.erl
```diff
@@ -658,7 +658,7 @@ hexchar(X) when X > -1, X < 16 -> $A + X - 10.
 
 -spec node_hash(enc_node()) -> hash().
 node_hash(<<>>) -> error(no_hash_for_null);
-node_hash(Bin) when byte_size(Bin) < 32   -> crypto:hash(sha256, Bin);
+node_hash(Bin) when byte_size(Bin) < 32   -> aec_hash:hash(header, Bin);
```
lucafavatella: I see this line has assumption that `aec_hash:hash(header, ...)` is 32 bytes - that is valid. If implementation of aec_hash changes I expect tests of aeu_mp_trees to fail (on decode) so no change needed.

hanssv:
Fixed conflict

PR #716, opened 2018-02-09
Expire peers that we cannot reach

ThomasArts:
Peers that we cannot reach for an hour are removed 
(unless they are pre-configured).

[finishes PT-#154131745]

### DIFF
--- a/apps/aecore/src/aec_peers.erl
+++ b/apps/aecore/src/aec_peers.erl
@@ -55,6 +55,8 @@
 -define(MAX_PING_INTERVAL, 120000).
 
 -define(DEFAULT_UNBLOCK_INTERVAL, 15 * 60 * 1000).
+-define(PEER_ERROR_EXPIRY, 60 * 60 * 1000).
+
 
 %% We parse the uri's with http_uri, therefore we use types from that module.
 -record(peer, {
@@ -64,6 +66,7 @@
           port              :: http_uri:port(),
           path              :: http_uri:path(),
           last_seen = 0     :: integer(), % Erlang system time (POSIX time)
+          expire            :: undefined | integer(), %% Erlang system time: when to drop this peer
           last_pings = []   :: [integer()], % Erlang system time
           ping_tref         :: reference() | undefined,
           trusted = false   :: boolean() % Is it a pre-configured peer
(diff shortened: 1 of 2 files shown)


apps/aecore/src/aec_peers.erl
```diff
@@ -554,11 +556,22 @@ log_ping_and_set_reping(Res, CalcF, Uri, Time, State) ->
...
             update_ping_metrics(Res),
-            Peer1 = save_ping_event(Res, Time, Peer),
-            Peer2 = start_ping_timer(CalcF, Peer1),
-            Peers = enter_peer(Peer2, State#state.peers),
-            Errored = update_errored(Res, uri_of_peer(Peer), State#state.errored),
-            State#state{peers = Peers, errored = Errored}
+            %% drop an errored peer if last time seen is longer than expire interval ago
+            Expire = 
+               application:get_env(aecore, peer_error_expired, ?DEFAULT_EXPIRE_INTERVAL),
+            case Res == error andalso has_been_seen(Peer) andalso Time > Peer#peer.last_seen + Expire
```
hanssv: If it has "never been seen" we will never remove it?

apps/aecore/src/aec_peers.erl
```diff
@@ -554,11 +556,22 @@ log_ping_and_set_reping(Res, CalcF, Uri, Time, State) ->
...
         {value, _Hash, Peer} ->
             update_ping_metrics(Res),
-            Peer1 = save_ping_event(Res, Time, Peer),
-            Peer2 = start_ping_timer(CalcF, Peer1),
-            Peers = enter_peer(Peer2, State#state.peers),
-            Errored = update_errored(Res, uri_of_peer(Peer), State#state.errored),
-            State#state{peers = Peers, errored = Errored}
+            %% drop an errored peer if last time seen is longer than expire interval ago
+            Expire = 
+               application:get_env(aecore, peer_error_expired, ?DEFAULT_EXPIRE_INTERVAL),
```
uwiger: perhaps `peer_error_expiry` would be more correct?

apps/aecore/src/aec_peers.erl
ThomasArts: True. In the kademlia setting we would probably keep an expiry per node and update it with 3600 seconds whenever we successfully ping it.

apps/aecore/src/aec_peers.erl
ThomasArts: Of course, that's a much better name. I rebased against master and changed it accordingly

apps/aecore/src/aec_peers.erl
```diff
@@ -64,6 +66,7 @@
           port              :: http_uri:port(),
           path              :: http_uri:path(),
           last_seen = 0     :: integer(), % Erlang system time (POSIX time)
+          expire            :: undefined | integer(), %% Erlang system tyme: when to drop this peer
```
hanssv: s/tyme/time

PR #718, opened 2018-02-09
Delete non-deterministic mining test from aec_conductor

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/154770099


dincho:
> I can't see that the test is non-deterministic, we should be able to mine a block? How long it takes might vary a little bit, but that can be fixed by making the timeout long enough?

We tried the timeouts while testing Jenkins, still flaky. If you go this direction, I would propose some "delta" that would be added to the timeout that represents the executing system "lag". That way it will fail fast enough on fast systems and wait enough on slow systems, because we could change the "delta" on system basis.

PR #719, opened 2018-02-09
Review running eunit tests via common test

lucafavatella:
Coded while preparing https://github.com/aeternity/epoch/pull/718

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -248,7 +248,7 @@ internal-clean: $$(KIND)
 	dev2-start, dev2-stop, dev2-attach, dev2-clean \
 	dev3-start, dev3-stop, dev3-attach, dev3-clean \
 	dialyzer \
-	test aevm-test-deps\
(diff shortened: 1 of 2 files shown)


lucafavatella:
> When you create a _SUITE file in the test directory and add compile directive
> -compile({parse_transform, ct_eunit_xform}).
> you will automatically include the enuit tests in a suite.

@ThomasArts Apparently [we do not](https://github.com/aeternity/epoch/blob/08b58bb4af575fcf91822b431b1c858b32a27bea/apps/aecore/test/aec_eunit_SUITE.erl#L23-L26). Hence my comment in rebar.config in commit "Clarify how to include eunit tests among common tests".

> But it is slow and unnecessary, so not something you want to do standard.

Ack.
I wanted to run that locally, and it was not trivial to figure out how. Hence this PR - in case other Devs wonder.

lucafavatella:
Not strictly needed so closing without merging.

PR #721, opened 2018-02-09
HTTP Endpoint for getting unsingned contract txs

velzevur:
This is the first of a series of commits implementing endpoints for getting valid unsinged transactions that can be signed outside the node and posted again.

Also implemented some helper functions for chaining request processing functions.

#### Create contract
Get a semantically correct unsigned `contract_create_tx`
```
POST /v2/tx/contract/create
```
mandatorary params: owner, code, vm_version, deposit, amount, gas, gas_price, fee, call_data
optional params: nonce; if not provided - `owner`-s account next nonce is used.
#### Create contract
Get a semantically correct unsigned `contract_call_tx`
```
POST /v2/tx/contract/call
```
mandatorary params: caller, contract, vm_version, amount, gas, gas_price, fee, call_data
optional params: nonce; if not provided - `caller`-s account next nonce is used.


apps/aecontract/src/aect_ring.erl
```diff
@@ -53,6 +56,15 @@ hexstring_encode(Code) ->
            || <<X:4, Y:4>> <= Code >>,
     <<"0x", CodeAsHexString/binary >>.
     
+hexstring_decode(Code) ->
+    <<"0x", CodeAsHexString/binary >> = Code,
+    0 = bit_size(CodeAsHexString) rem 8,
```
hanssv: Would it get past the `/binary` at the previous line if it wasn't full bytes?

velzevur:
```
1> Bin = <<"0x">>.
<<"0x">>
2> <<"0x", Rest/binary>> = Bin.
<<"0x">>
3> Rest.
<<>>
```

velzevur: for completeness
```
4> bit_size(Rest).
0
```

apps/aecontract/src/aect_ring.erl
```diff
@@ -53,6 +56,24 @@ hexstring_encode(Code) ->
            || <<X:4, Y:4>> <= Code >>,
     <<"0x", CodeAsHexString/binary >>.
     
+hexstring_decode(Code) ->
+    <<"0x", CodeAsHexString/binary >> = Code,
```
hanssv: This will ensure `CodeAsHexString` is 8*n bits...

PR #723, opened 2018-02-12
Update epoch API version to 0.6.0

dincho:


dincho:
As a historical note as already mentioned by @sennui, the API could have it's own lifecycle and versioning, however we agreed that it would be more easy and clear to have the same version as the epoch.

PR #724, opened 2018-02-12
Fix epoch image Dockerfile

dincho:
- use aetrnty/builder as builder image
- copy libsodium to epoch image

I think this also addresses [PT 154993662](https://www.pivotaltracker.com/story/show/154993662)

### DIFF
--- a/docker/Dockerfile-epoch
+++ b/docker/Dockerfile-epoch
@@ -1,12 +1,4 @@
-FROM erlang:20.2 as builder
-
-ENV LIBSODIUM_VERSION=1.0.16
-RUN LIBSODIUM_DOWNLOAD_URL="https://github.com/jedisct1/libsodium/releases/download/${LIBSODIUM_VERSION}/libsodium-${LIBSODIUM_VERSION}.tar.gz" \
-    && curl -fsSL -o libsodium-src.tar.gz "$LIBSODIUM_DOWNLOAD_URL" \
-    && mkdir libsodium-src \
-    && tar -zxf libsodium-src.tar.gz -C libsodium-src --strip-components=1 \
-    && cd libsodium-src \
-    && ./configure && make -j$(nproc) && make install && ldconfig
+FROM aetrnty/builder as builder
 
 ADD . /app
 RUN cd /app && make prod-build
@@ -14,12 +6,17 @@ RUN cd /app && make prod-build
 # Put epoch node in second stage container
 FROM ubuntu:16.04
 
-# OpenSSL is shared lib dependency
-RUN apt-get -qq update && apt-get -qq -y install openssl curl \
-    && rm -rf /var/lib/apt/lists/*
-
 # Deploy application code from builder container
 COPY --from=builder /app/_build/prod/rel/epoch /home/epoch/node
+
+# Copy custom built libs
+# Note the `ldconfig` requirement which is implicitly run by openssl install
+COPY --from=builder /usr/local/lib /usr/local/lib
+
+# OpenSSL is shared lib dependency
+RUN apt-get -qq update && apt-get -qq -y install openssl curl \
+    && ldconfig \
+    && rm -rf /var/lib/apt/lists/*
 
 # Epoch app won't run as root for security reasons
 RUN useradd --shell /bin/bash epoch \


docker/Dockerfile-epoch
```diff
@@ -1,26 +1,23 @@
...
 
 # Put epoch node in second stage container
 FROM ubuntu:16.04
 
+# Deploy application code from builder container
+COPY --from=builder /app/_build/prod/rel/epoch /home/epoch/node
+
+# Copy custom built libs
+# Note the `ldconfig` requirement which is implicitly run by openssl install
+COPY --from=builder /usr/local/lib /usr/local/lib
```
lucafavatella: Can this be made more explicit by copying each relevant file one by one? I am concerned of unfulfilled expectations silently slipping through.

docker/Dockerfile-epoch
```diff
@@ -1,26 +1,23 @@
-FROM erlang:20.2 as builder
-
-ENV LIBSODIUM_VERSION=1.0.16
-RUN LIBSODIUM_DOWNLOAD_URL="https://github.com/jedisct1/libsodium/releases/download/${LIBSODIUM_VERSION}/libsodium-${LIBSODIUM_VERSION}.tar.gz" \
-    && curl -fsSL -o libsodium-src.tar.gz "$LIBSODIUM_DOWNLOAD_URL" \
-    && mkdir libsodium-src \
-    && tar -zxf libsodium-src.tar.gz -C libsodium-src --strip-components=1 \
-    && cd libsodium-src \
-    && ./configure && make -j$(nproc) && make install && ldconfig
+FROM aetrnty/builder as builder
```
lucafavatella: Do I understand correctly that this string here does not prevent user to build locally Dockerfile-builder then build this Dockerfile from that local builder? (i.e. this Dockerfile is not dependent on Docker Hub rather locally built builder takes the precedence.)

amir-hadi: If you create (build) the builder image and tag it `aetrnty/builder`, it would be used. Otherwise it will go to the hub and download the image named `aetrnty/builder`.

docker/Dockerfile-epoch
dincho: I already tried and had issues with links - all symlinks sources are copied which broke the linux linker. I don't want to run extra command to create the required links to the .so, besides the verbosity it would need explicit so version e.g. `23.1.0`

This is what's currently copied (this is my VM, but it's the same):
```
ls -la /usr/local/lib/
total 5056
drwxr-xr-x  6 root root     4096 Feb 12 13:04 .
drwxr-xr-x 10 root root     4096 Jan  6  2017 ..
drwxr-xr-x  8 root root     4096 Feb 12 12:41 erlang
-rw-r--r--  1 root root  3162460 Feb 12 13:04 libsodium.a
-rwxr-xr-x  1 root root      943 Feb 12 13:04 libsodium.la
lrwxrwxrwx  1 root root       19 Feb 12 13:04 libsodium.so -> libsodium.so.23.1.0
lrwxrwxrwx  1 root root       19 Feb 12 13:04 libsodium.so.23 -> libsodium.so.23.1.0
-rwxr-xr-x  1 root root  1978592 Feb 12 13:04 libsodium.so.23.1.0
drwxr-xr-x  2 root root     4096 Feb 12 13:04 pkgconfig
drwxrwsr-x  4 root staff    4096 Feb 12 12:16 python2.7
drwxrwsr-x  3 root staff    4096 Jan  6  2017 python3.5
```

@amir-hadi any idea on this?

lucafavatella: Taking the problem from another angle - having a smoke test for the Docker image would prove that the image was built correctly (I am thinking of instantiating container and exercising its functionalities from the HTTP API). And this smoke testing is tracked separately - in [1](https://www.pivotaltracker.com/story/show/155124244) and [2](https://www.pivotaltracker.com/story/show/155124252).

amir-hadi: @dincho What I understood from Lucas feedback is that he doesn't want to copy the whole /usr/local/lib and replace everything, instead make it more explicit by copying only the libsodium files.

So it would look a bit more like this:
`COPY --from builder /usr/local/lib/libsodium* /usr/local/lib`

Did I get you right @lucafavatella?

PR #726, opened 2018-02-12
CircleCI - run nightly UAT tests

dincho:
Tests was ran 00:00 UTC last night: https://circleci.com/workflow-run/140fb9a6-eb3a-4ef2-bc97-7f6c287ede74 (indeterminate ?)

Successful builds can be found here:
- https://circleci.com/gh/aeternity/epoch/1542
- https://circleci.com/gh/aeternity/epoch/1527

However, I had to increase the test timeouts in 3e0e94df1244a1711af67a550c4aec44efab9c82 to make it work in CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -192,14 +192,20 @@ jobs:
             epmd -daemon
             make python-env && make swagger && make multi-build
             make python-uats
+      - run:
+          name: Prepare JUnit Report
+          command: mv py/nosetests.xml py/junit.xml
       - store_test_results:
-          path: py/nosetests.xml
+          path: py
       - store_artifacts:
           path: _build/dev1/rel/epoch/log
+          destination: node1/
       - store_artifacts:
           path: _build/dev2/rel/epoch/log
+          destination: node2/
       - store_artifacts:
           path: _build/dev3/rel/epoch/log
+          destination: node3/
 
   deploy_integration:
     <<: *container_config
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,6 +180,27 @@ jobs:
       - *store_test_package_artifacts
       - *store_package_artifacts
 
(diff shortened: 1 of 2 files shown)


dincho:
I'll rework the commit history before merging.

.circleci/config.yml
```diff
@@ -180,6 +180,33 @@ jobs:
...
+    <<: *container_config
+    steps:
+      - checkout
+      - *restore_rebar_cache
+      - *restore_build_cache
+      - run:
+          name: UAT Tests
+          command: |
+            epmd -daemon
+            make python-env && make swagger && make multi-build
```
lucafavatella: Why `make swagger`?

.circleci/config.yml
```diff
@@ -180,6 +180,33 @@ jobs:
...
+            make python-uats
```
lucafavatella: Who does the start of the nodes? I would expect a `make multi-start` somewhere or equivalent.

.circleci/config.yml
dincho: @lucafavatella I followed the docs here: https://github.com/aeternity/epoch/tree/master/py

If the docs are wrong, let's update them in this PR too.

.circleci/config.yml
dincho: See above comment

dincho: @velzevur can you bring some light here?

velzevur: In each test we prepare a specific sys.config and start nodes using it. Nodes shouldn't be running before starting the tests.

lucafavatella: Are [ws-related python tests](https://github.com/aeternity/epoch/blob/84a9bef744e329f764e1b68966cbc0dfde523512/Makefile#L144) meant to be run too? Deleted?
Cc @velzevur

PR #729, opened 2018-02-12
Fix Swagger check after concurrent merge of PRs

lucafavatella:


lucafavatella:
This PR is change of 2 comments in Python test-only files. It impacts only static_analysis job, it passed, hence merging.

PR #732, opened 2018-02-12
Make aec_chain_state use database rather than a functional state

gorillainduction:
* Blocks and headers are stored unless they are blatantly malformed.
* Blocks are stored in db as headers + transactions
* Previous hash is indexed in db
* The top block hash is driving top header hash rather than the opposite
* Avoid giving error messages for other nodes than the one currently added
* Add stricter checking before adding state to db
* If a node has a stored state in the db, it needs no more checking
* Better restart from persisted states

https://www.pivotaltracker.com/n/projects/2124891/stories/153442321
https://www.pivotaltracker.com/n/projects/2124891/stories/154877787

### DIFF
--- a/apps/aecore/src/aec_conductor_chain.erl
+++ b/apps/aecore/src/aec_conductor_chain.erl
@@ -59,14 +59,14 @@ reinit(State) ->
                        end).
 
 init_state(State) ->
-    case aec_db:get_chain() of
-        [] ->
+    case aec_db:get_genesis_hash() of
+        undefined ->
             {GB, _GBState} = aec_block_genesis:genesis_block_with_state(),
             State1 = State#state{chain_state = aec_chain_state:new()},
             {ok, State2} = insert_block(GB, State1),
             State2;
-        Chain ->
-            ChainState = aec_chain_state:new_from_persistence(Chain),
+        Hash when is_binary(Hash) ->
+            ChainState = aec_chain_state:new_from_persistence(),
             State#state{chain_state = ChainState}
     end.
 
@@ -199,7 +199,10 @@ insert_header(Header, State) ->
     ChainState1 = State#state.chain_state,
     case aec_chain_state:insert_header(Header, ChainState1) of
         {ok, ChainState2} ->
-            persistence_store_header(Header, ChainState2),
+            aec_db:ensure_transaction(
+              fun() ->
+                      persist_chain(ChainState2)
+              end),
             {ok, State#state{chain_state = ChainState2}};
         {error,_Reason} = E -> E
     end.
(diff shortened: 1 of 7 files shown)


apps/aecore/src/aec_db.erl
```diff
@@ -86,7 +92,7 @@
 
 tables(Mode) ->
     [?TAB(aec_blocks)
-   , ?TAB(aec_headers)
+   , ?TAB(aec_headers, [{index, [prev_hash]}])
```
lucafavatella: In am trying to understand better this index and its test coverage.
Considering that for now Mnesia indexes are delicate (otp_patches), is this index exercised during the [package smoke test](https://github.com/aeternity/epoch/blob/ac3622ea50a37a34084ccdcb320584b0239a7192/.circleci/config.yml#L52)?
If not, is it at least exercised in any one of the aehttp_integration_SUITE/aecore_sync_SUITE/aecore_forking_SUITE? If not, could you please at least add a specific test like in this [(temporary) test (now removed because exercised by API test)](https://github.com/aeternity/epoch/commit/4e60719fb1124d2e28ddf470e0a5c24e91399a66#diff-14dc0eab04e011248c7640cb3c1cf0deL18) ?

gorillainduction: I refrained from using the plugin indexes, but use a proper, ancient direct index instead. This does not need to be tested more.

It is tested indirectly by the aec_chain_state_tests, though. The index is instrumental for reasoning about the graph structure of the chain.

lucafavatella: I see that mistyping the index name here (e.g. `prev_hash2`) makes the table creation fail in `make test`. Good.

apps/aecore/src/aec_db.erl
```diff
@@ -158,6 +182,33 @@ get_header(Hash) ->
...
+    end.
+
+find_chain_node(Hash) ->
+    case ?t(mnesia:read(aec_headers, Hash)) of
+        [H] -> header_to_node(H);
+        [] -> none
+    end.
+
+find_chain_node_successors(Hash) ->
+    Objects = ?t(mnesia:index_read(aec_headers, Hash, prev_hash)),
```
lucafavatella: I see that mistyping the index name here (e.g. `prev_hash3`) causes `make test` to fail. Good.

apps/aecore/src/aec_conductor_chain.erl
```diff
@@ -211,50 +214,27 @@ resolve_name(Type, Name, State) ->
...
+              persist_chain(State)
+      end).
 
 persist_chain(State) ->
-    aec_db:ensure_transaction(fun() -> do_persist_chain(State) end).
-
-do_persist_chain(State) ->
-    case aec_chain_state:top_header_hash(State) of
+    case aec_chain_state:get_genesis_hash(State) of
         undefined -> ok;
```
lucafavatella: I see that `init_state` guarantees the genesis hash to be present. I see this `persist_chain`function is called when inserting a block - *including genesis* - or header.

Do I understand correctly that, maybe by making the `insert_block` function call in `init_state` more specific, this function `persist_chain` can be made to not conditionally not-persist chain? I find that a function `persist_chain` that may *not* do any side effect for a case that I understand shall not happen may be improved in readability.

apps/aecore/src/aec_db.erl
```diff
@@ -158,6 +182,33 @@ get_header(Hash) ->
...
+        [] -> none
+    end.
+
+find_chain_node(Hash) ->
+    case ?t(mnesia:read(aec_headers, Hash)) of
+        [H] -> header_to_node(H);
+        [] -> none
+    end.
+
+find_chain_node_successors(Hash) ->
```
lucafavatella: I tested index works after restart. Good.

I built release `make prod-build`, I created config with persisted DB (and lean16 minutes for getting initial 10 blocks faster), I started node, I used this `find_chain_node_successors` that behaved as expected. I stopped and started node, and this function returned same block (linear chain) as before.

gorillainduction:
@lucafavatella 
Currently, the index is not persisted by default, but mnesia rebuilds the index on start. This will eventually increase the startup time, but not not increase the disk usage.

I expect the increased disk/memory for the index to be much smaller than the copy of all blocks/headers that was the previous behavior, so I expect this to be significantly smaller in RAM for disk db, and "a bit" smaller for non-persisted db.

We should probably observe the RAM behavior in UAT nodes, and we can update the recommendations to the next release.

PR #733, opened 2018-02-13
Add ldconfig to libsodium install instructions

dincho:


RELEASE-NOTES.md
```diff
@@ -27,7 +27,7 @@ dependency on libsodium v1.0.16 installed with [Homebrew](https://brew.sh/) also
 In case you have installed either of them in a non-default path, you could use symlink(s) to work around the issue.
 
 To run on Ubuntu 16.04 you need to have a libsodium shared library (v1.0.16) in `/usr/local/lib/libsodium.so.23`. (`wget https://download.libsodium.org/libsodium/releases/libsodium-1.0.16.tar.gz`,
-unpack, then do `./configure && make && sudo make install`)
+unpack, then do `./configure && make && sudo make install && sudo ldconfig`)
```
lucafavatella: I see https://github.com/aeternity/epoch/blob/0685427f4f42c2ce5f7bcfcd4c18d750a2bd276c/docker/Dockerfile-builder#L28

dincho: that is run with root while ubuntu commands are usually run as sudo user

PR #735, opened 2018-02-13
Pt 154838110 logging on console

amir-hadi:
Enable logging on the console, so we can see logs using `docker logs`


amir-hadi:
@lucafavatella Is there an option to use `aev_env:user_config_or_env` in the `sys.config` file? Very new to erlang. I'd like to be able to configure log level etc using the `epoch.yaml`.

lucafavatella:
I understand this superseded https://github.com/aeternity/epoch/pull/709

PR #740, opened 2018-02-13
Pt 154904411 ws binding address

amir-hadi:
In this branch I refactored the server configuration to allow the configuration of listen addresses for the existing interfaces.

What I have also done is refactor the default listen address of the internal interface to be 127.0.0.1, as recommended.

I also changed the localnet docker setup of node1 to listen_address `0.0.0.0`, which will allow nginx and users (later) to access that port.

EDIT: This PR addresses https://www.pivotaltracker.com/n/projects/2124891 only, more related work will be done in https://www.pivotaltracker.com/story/show/155249688


amir-hadi:
Just a quick note: I have added a small function `get_and_parse_ip_address_from_config_or_env`, which makes it easier to get IP addresses from config. I hope thats ok. Otherwise I'd remove it and make it more explicit in other `get_*_listen_address` functions.

apps/aehttp/src/aehttp_app.erl
```diff
@@ -69,31 +74,52 @@ start_swagger_internal() ->
...
     ]),
     {ok, _} = cowboy:start_http(http, PoolSize,
-                                 [{port, Port}, {ip, {127, 0, 0, 1}}],
+                                 [{port, Port}, {ip, ListenAddress}],
                                  [{env, [{dispatch, Dispatch}]}]),
     ok.
 
+get_and_parse_ip_address_from_config_or_env(CfgKey, App, EnvKey, Default) ->
+    Config = aeu_env:user_config_or_env(CfgKey, App, EnvKey, Default),
+    case is_binary(Config) of
```
dincho: Why do you need this check, in what case the config value is not binary?

amir-hadi: I set the `-define` statements to be tuples again, like the defaults really were before.

dincho: I would prefer binary defaults to get rid the of this extra checks

apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -58,6 +58,11 @@
                             "pattern": "^http://[^:\\.\"!#$%^&*()',/]+(\\.[^:\\.\"!#$%^&*()',/]+)*:[0-9]+/$",
                             "example" : "http://somehost.somedomain:123/"
                         },
+                        "listen_address": {
+                          "description": "Listen address for external interface. This should be publicly accessible",
+                          "type": "string",
```
lucafavatella: Would it be possible to add a pattern for restricting IPv4 address? (As I understand it is what is currently tested.)

docker/epoch_node1_mean16.yaml
```diff
@@ -7,6 +7,10 @@ http:
     external:
         peer_address: http://node1:3013/
 
+websocket:
+    internal:
+        listen_address: 0.0.0.0
```
lucafavatella: Shall the internal HTTP port be set too? Did you test that?

amir-hadi: It can be set, I didn't test it, because that functionality was there before my PR already.

lucafavatella: I expect [this](https://github.com/aeternity/epoch/pull/740/files?diff=split#diff-735feba3cdeb8646424e3f5fe388244dR66) changed things, I suggest retest of this point.

amir-hadi:
```yaml
websocket:
    internal:
        listen_address: 0.0.0.0
        port: 4000```

This worked

File: apps/aeutils/priv/epoch_config_schema.json (hunk 2 above)
Comment (amir-hadi) [reply apps/aeutils/priv/epoch_config_schema.json]: Wrote you about that on Slack, a working regex for IPv4, IPv6 and potential hostnames is quite complicated. As this is a optional field I am OK with not specifying the pattern here.

File: docker/epoch_node1_mean16.yaml (hunk 3 above)
Comment (lucafavatella) [reply docker/epoch_node1_mean16.yaml]: > Shall the internal HTTP port be set too? Did you test that?

@amir-hadi In my comment was not precise - I apologise. I think we were not on the same page.

I meant the internal HTTP *listen address* - not port. I.e. this:

> I expect [this](https://github.com/aeternity/epoch/pull/740/files?diff=split#diff-735feba3cdeb8646424e3f5fe388244dR66) changed things, I suggest retest of this point.

Comment (amir-hadi) [reply docker/epoch_node1_mean16.yaml]: I hope that I got you know, I did test, too.

```yaml
http:
    external:
        peer_address: http://node1:3013/
        listen_address: 0.0.0.0
    internal:
        listen_address: 0.0.0.0

websocket:
    internal:
        listen_address: 0.0.0.0
```

lucafavatella: Do I understand correctly that unless you amend docker/epoch_node1_mean16.yaml with the following section the user does not manage to contact internal HTTP endpoint?
```yaml
http:
    internal:
        listen_address: 0.0.0.0
```

(I check better doc and nginx config...)

dincho: @amir-hadi What @lucafavatella is trying to explain that you also must change the ` docker/epoch_node1_mean16.yaml` to add:

```
http:
    internal:
        listen_address: 0.0.0.0
```

Because we changed the default listen address of the internal API from `0.0.0.0` to `127.0.0.1`, thus you cannot reach it anymore outside of the container, therefore the nginx proxy cannot reach it too as well.

lucafavatella:
I see two approvals. I see CI passed. Merging.

PR #741, opened 2018-02-13
add tests on rpt destination filtering

uwiger:
Ref https://www.pivotaltracker.com/story/show/154133896

Added unit and system test cases for destination filtering, and indeed found a couple of bugs.

### DIFF
--- a/apps/aecore/src/aec_metrics_lib.erl
+++ b/apps/aecore/src/aec_metrics_lib.erl
@@ -1,16 +1,35 @@
 -module(aec_metrics_lib).
 
 -export([find_entries_cmd/1,
+         find_entries_cmd/2,
+         filter_entries/2,
          get_values_cmd/1]).
 
 find_entries_cmd(Arg) ->
-    find_entries_cmd(Arg, enabled).
+    find_entries_(Arg, enabled, fun exometer:select/1).
 
 find_entries_cmd(Arg, Status) ->
+    find_entries_(Arg, Status, fun exometer:select/1).
+
+filter_entries(Entries, Arg) ->
+    %% Entries shall have the same format as the product of
+    %% exometer:select/1, i.e. [{Name, Type, Status}].
+    %% Metrics that don't exist are filtered out
+    Filtered = lists:filter(
+                 fun({N, _Type, Status}) ->
+                         Status =:= enabled andalso metric_exists(N)
+                 end, Entries),
+    Select = fun(Patterns) ->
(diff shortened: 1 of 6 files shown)


apps/aecore/src/aec_metrics_rpt_dest.erl
```diff
@@ -170,11 +170,17 @@ make_destinations(Found, Actions0, Tab) ->
     Dests = mk_dests_(Actions0),
     lists:foreach(
       fun({Metrics, DPs}) ->
-              [ets:insert(Tab, #metric{key = {N,D}, destinations = Dests})
+              [ets_insert_new(Tab, #metric{key = {N,D}, destinations = Dests})
                || {N, _, _} <- Metrics,
                   D <- DPs]
       end, Found).
 
+ets_insert(Tab, Obj) ->
```
happi: Why is this function introduced?
Perhaps it should have a less specific name to allow for future rewrite to use another backend?

uwiger: It was introduced to simplify debugging via tracing. I usually add a comment to that effect.

uwiger:
I've realized that while the filter (now) seems to work, it doesn't pick up metrics created after the initial population of the filter table (again with the complex start order ...). The system metrics observed by Dincho seem to have been created afterwards.

I'm working on a change in the code to apply the filter rules to metrics if they can't be found in the filter table; the result will then be cached in the table for future reference.

apps/aecore/src/aec_metrics_lib.erl
```diff
@@ -1,16 +1,30 @@
...
 
 find_entries_cmd(Arg) ->
-    find_entries_cmd(Arg, enabled).
+    find_entries_(Arg, enabled, fun exometer:select/1).
 
 find_entries_cmd(Arg, Status) ->
+    find_entries_(Arg, Status, fun exometer:select/1).
+
+filter_entries(Entries, Arg) ->
+    Entries = [{N,D,S} || {N,D,S} <- Entries,
```
lucafavatella: This line silently discards items in `Entries` that do not match `{_,_,_}`, making it fragile to future code changes. May I suggest replacing this with code that hardcodes the expectation on each item of `Entries`.

uwiger: This gave me reason to double-check the implementation, and there are actually a few more things I'd like to correct. Will push a new version a bit later.

uwiger:
Rebased and squashed.

PR #742, opened 2018-02-14
#155219691 - Avoid fetching already fetched blocks in sync

hanssv:
Before fetching from a peer, check with aec_conductor if we have the
block locally.

Also refactor and most importantly rename the fetch_chain (used to be
named fetch_headers although it fetched blocks...).

### DIFF
--- a/apps/aecore/src/aec_sync.erl
+++ b/apps/aecore/src/aec_sync.erl
@@ -1,7 +1,7 @@
 %%%=============================================================================
 %%% @copyright 2017, Aeternity Anstalt
 %%% @doc
-%%%    Module storing peers list and providing funtionc for peers interaction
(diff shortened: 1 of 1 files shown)


apps/aecore/src/aec_sync.erl
```diff
@@ -1,7 +1,7 @@
 %%%=============================================================================
 %%% @copyright 2017, Aeternity Anstalt
 %%% @doc
-%%%    Module storing peers list and providing funtionc for peers interaction
+%%%    Module storing peers list and providing funtions for peer interaction
```
lucafavatella: functions

hanssv: I like doing improvements in small steps, let's leave some low hanging fruit for the next person editing :-)

PR #743, opened 2018-02-14
PT-154313327 Add name and label validation

zp-sd:
* accept unicode names
* store name in its ASCII representation
* introduce IDNA library

### DIFF
--- a/apps/aehttp/src/aehttp_dispatch_ext.erl
+++ b/apps/aehttp/src/aehttp_dispatch_ext.erl
@@ -236,15 +236,18 @@ handle_request('GetAccountBalance', Req, _Context) ->
 handle_request('GetCommitmentHash', Req, _Context) ->
     Name         = maps:get('name', Req),
     Salt         = maps:get('salt', Req),
-    CHash        = aens:get_commitment_hash(Name, Salt),
-    EncodedCHash = aec_base58c:encode(commitment, CHash),
-    {200, [], #{commitment => EncodedCHash}};
+    case aens:get_commitment_hash(Name, Salt) of
+        {ok, CHash} ->
+            EncodedCHash = aec_base58c:encode(commitment, CHash),
+            {200, [], #{commitment => EncodedCHash}};
+        {error, Reason} ->
+            ReasonBin = atom_to_binary(Reason, utf8),
+            {400, [], #{reason => <<"Name validation failed with a reason: ", ReasonBin/binary>>}}
+    end;
 
 handle_request('GetName', Req, _Context) ->
     Name = maps:get('name', Req),
     case aec_chain:name_entry(Name) of
-        {error, name_not_found} ->
-            {404, [], #{reason => <<"Name not found">>}};
         {ok, NameEntry} ->
             #{<<"name">>     := Name,
               <<"hash">>     := Hash,
@@ -253,7 +256,12 @@ handle_request('GetName', Req, _Context) ->
             {200, [], #{name      => Name,
                         name_hash => aec_base58c:encode(name, Hash),
                         name_ttl  => NameTTL,
-                        pointers  => Pointers}}
+                        pointers  => Pointers}};
+        {error, name_not_found} ->
+            {404, [], #{reason => <<"Name not found">>}};
+        {error, Reason} ->
+            ReasonBin = atom_to_binary(Reason, utf8),
+            {400, [], #{reason => <<"Name validation failed with a reason: ", ReasonBin/binary>>}}
     end;
 
 handle_request('GetAccountsBalances', _Req, _Context) ->
--- a/apps/aehttp/src/aehttp_dispatch_int.erl
+++ b/apps/aehttp/src/aehttp_dispatch_int.erl
@@ -210,15 +210,21 @@ handle_request('PostNameClaimTx', #{'NameClaimTx' := NameClaimTxObj}, _Context) 
             #{<<"name">>      := Name,
               <<"name_salt">> := NameSalt,
               <<"fee">>       := Fee} = NameClaimTxObj,
-            {ok, ClaimTx} =
-                aens_claim_tx:new(
-                  #{account   => PubKey,
-                    nonce     => Nonce,
-                    name      => Name,
-                    name_salt => NameSalt,
-                    fee       => Fee}),
-            sign_and_push_to_mempool(ClaimTx),
-            {200, [], #{name_hash => aec_base58c:encode(name, aens_hash:name_hash(Name))}};
+            case aens:get_name_hash(Name) of
+                {ok, NameHash} ->
+                    {ok, ClaimTx} =
+                        aens_claim_tx:new(
+                          #{account   => PubKey,
+                            nonce     => Nonce,
+                            name      => Name,
+                            name_salt => NameSalt,
+                            fee       => Fee}),
+                    sign_and_push_to_mempool(ClaimTx),
+                    {200, [], #{name_hash => aec_base58c:encode(name, NameHash)}};
+                {error, Reason} ->
+                    ReasonBin = atom_to_binary(Reason, utf8),
+                    {400, [], #{reason => <<"Name validation failed with a reason: ", ReasonBin/binary>>}}
+            end;
         {error, account_not_found} ->
             {404, [], #{reason => <<"No funds in an account">>}};
         {error, key_not_found} ->
--- a/apps/aehttp/test/aehttp_integration_SUITE.erl
+++ b/apps/aehttp/test/aehttp_integration_SUITE.erl
@@ -1779,17 +1779,16 @@ block_txs_list_by_hash_invalid_range(_Config) ->
     ok.
 
 naming_system_manage_name(_Config) ->
-
     {ok, PubKey} = rpc(aec_keys, pubkey, []),
-    PubKeyEnc = aec_base58c:encode(account_pubkey, PubKey),
-    Name       = <<"fooo.bar.test">>,
-    NameSalt   = 12345,
-    NameTTL    = 60000,
-    Pointers   = <<"{\"account_pubkey\":\"", PubKeyEnc/binary, "\"}">>,
-    TTL        = 10,
-    NHash      = aens_hash:name_hash(Name),
-    Fee        = 2,
-    MineReward = rpc(aec_governance, block_mine_reward, []),
+    PubKeyEnc   = aec_base58c:encode(account_pubkey, PubKey),
+    Name        = <<"詹姆斯詹姆斯.test"/utf8>>,
+    NameSalt    = 12345,
+    NameTTL     = 60000,
+    Pointers    = <<"{\"account_pubkey\":\"", PubKeyEnc/binary, "\"}">>,
+    TTL         = 10,
+    {ok, NHash} = aens:get_name_hash(Name),
+    Fee         = 2,
+    MineReward  = rpc(aec_governance, block_mine_reward, []),
 
     %% Mine 10 blocks to get some funds
     aecore_suite_utils:mine_blocks(aecore_suite_utils:node_name(?NODE), 10),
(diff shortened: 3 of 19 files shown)


deployment/ansible/deploy.yml
```diff
@@ -40,7 +40,7 @@
           port: 3013
       chain:
         persist: true
-        db_path: "./db27"
+        db_path: "./db29"
```
lucafavatella: Not sure why 29 rather than 28 (as master is 27) though it is fine :)

rebar.lock
```diff
@@ -15,6 +15,10 @@
   {git,"https://github.com/ninenines/cowlib.git",
        {ref,"d544a494af4dbc810fc9c15eaf5cc050cced1501"}},
   1},
+ {<<"enacl">>,
```
lucafavatella: Thanks.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -244,15 +244,18 @@ handle_request('GetAccountBalance', Req, _Context) ->
...
     Salt         = maps:get('salt', Req),
-    CHash        = aens:get_commitment_hash(Name, Salt),
-    EncodedCHash = aec_base58c:encode(commitment, CHash),
-    {200, [], #{commitment => EncodedCHash}};
+    case aens:get_commitment_hash(Name, Salt) of
+        {ok, CHash} ->
+            EncodedCHash = aec_base58c:encode(commitment, CHash),
+            {200, [], #{commitment => EncodedCHash}};
+        {error, Reason} ->
+            ReasonBin = atom_to_binary(Reason, utf8),
```
lucafavatella: This is relying on undocumented/unspec-ed `aens:get_commitment_hash` return type.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -261,7 +264,12 @@ handle_request('GetName', Req, _Context) ->
             {200, [], #{name      => Name,
                         name_hash => aec_base58c:encode(name, Hash),
                         name_ttl  => NameTTL,
-                        pointers  => Pointers}}
+                        pointers  => Pointers}};
+        {error, name_not_found} ->
+            {404, [], #{reason => <<"Name not found">>}};
+        {error, Reason} ->
+            ReasonBin = atom_to_binary(Reason, utf8),
```
lucafavatella: This is relying on undocumented/unspec-ed `aec_conductor:get_name_entry` return type.

apps/aens/test/aens_utils_test.erl
```diff
@@ -0,0 +1,28 @@
...
+
+to_ascii_test() ->
+    ?assertEqual({error, no_registrar},
+                 aens_utils:to_ascii(<<"abcdefg">>)),
+    ?assertEqual({error, multiple_namespaces},
+                 aens_utils:to_ascii(<<"abcd.efgh.aet">>)),
+    ?assertEqual({error, registrar_unknown},
+                 aens_utils:to_ascii(<<"abcd.aettt">>)),
+    LongName = base64:encode(crypto:strong_rand_bytes(1000)),
+    ?assertEqual({error, name_too_long},
```
lucafavatella: I have not yet understood the significance of `aens_utils:to_ascii/1`, though I see some length checks here. I am concerned of:
* Off-by-one mistakes; and (orthogonally)
* Byte sequence of which to consider length (whole name? single label? if at least a Unicode char: the Unicode representation passed, a normalized Unicode representation of the string passed, the punycode representation of a normalized Unicode representation of the string passed?)

Some more tests may be beneficial (not sure whether here or in other points of the PR - still checking).

zbigniewpekala: Regarding length checks: this should be an option of Erlang-IDNA lib (shall be length of punycode representation), but pinned lib does not support it yet. Hence I added it here, to have very basic validation for now.

zp-sd: Regarding length checks: this should be an option of Erlang-IDNA lib (shall be length of punycode representation), but pinned lib does not support it yet. Hence I added it here, to have very basic validation for now.

apps/aehttp/src/aehttp_dispatch_ext.erl
zp-sd: Yes, how do we want to document that? Swagger? Erlang spec?

lucafavatella: My point is that is this code path is exercised then `Reason` may be not an atom and this line will cause client to see 500 rather than 400.

So the point is: `can aens:get_commitment_hash` be spec-ed so that it returns `{error, atom()}`? If not, then this code path shall not use atom_to_binary but rather some pretty-printing.

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -210,15 +210,21 @@ handle_request('PostNameClaimTx', #{'NameClaimTx' := NameClaimTxObj}, _Context)
...
+                        aens_claim_tx:new(
+                          #{account   => PubKey,
+                            nonce     => Nonce,
+                            name      => Name,
+                            name_salt => NameSalt,
+                            fee       => Fee}),
+                    sign_and_push_to_mempool(ClaimTx),
+                    {200, [], #{name_hash => aec_base58c:encode(name, NameHash)}};
+                {error, Reason} ->
+                    ReasonBin = atom_to_binary(Reason, utf8),
```
lucafavatella: Same here. Not critical as it is error path anyway though user would see 500 rather than 400. And 400 without reason message is better than 500.

apps/aehttp/src/aehttp_dispatch_ext.erl
zp-sd: I see, agreed, I'll change that specs to `atom().`

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -1775,17 +1775,16 @@ block_txs_list_by_hash_invalid_range(_Config) ->
...
-    Name       = <<"fooo.bar.test">>,
-    NameSalt   = 12345,
-    NameTTL    = 60000,
-    Pointers   = <<"{\"account_pubkey\":\"", PubKeyEnc/binary, "\"}">>,
-    TTL        = 10,
-    NHash      = aens_hash:name_hash(Name),
-    Fee        = 2,
-    MineReward = rpc(aec_governance, block_mine_reward, []),
+    PubKeyEnc   = aec_base58c:encode(account_pubkey, PubKey),
+    Name        = <<"詹姆斯詹姆斯.test"/utf8>>,
```
lucafavatella: I see this test is using a name with Unicode non-ASCII chars in it - testing all the way through the HTTP APIs. Good.

apps/aehttp/src/aehttp_dispatch_int.erl
zp-sd: Spec changed to `term()`

apps/aehttp/src/aehttp_dispatch_ext.erl
zp-sd: Spec changed to `atom()`.

apps/aehttp/test/aehttp_integration_SUITE.erl
lucafavatella: It would be nice to have small additions in this test for checking handling of punycode in API.
One small addition I am thinking about is, after [the check that the name is present passing it as Unicode](https://github.com/aeternity/epoch/pull/743/files#diff-14dc0eab04e011248c7640cb3c1cf0deR1833), performing the same get_name call passing the punycode representation of the domain. This may be done in a following PR.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -1775,17 +1775,16 @@ block_txs_list_by_hash_invalid_range(_Config) ->
     ok.
 
 naming_system_manage_name(_Config) ->
```
lucafavatella: * Would it be possible to add equivalent test with client specifying name as punycode rather than Unicode?

* Would it be possible to add test with two concurrent name pre-claim / claim going on, with one client specifying name as punycode and other as Unicode?

(In separate subsequent PR I mean.)

zp-sd: Yes, this is feasible, but IMO such this is not a good idea, as such test will become stale.

`xn--` is a prefix of all punycode encoded domains. Based on our spec https://github.com/aeternity/protocol/blob/master/AENS.md#name, we should be using `CheckHyphens=true` config, which must satisfy the below:
```
If CheckHyphens, the label must not contain a U+002D HYPHEN-MINUS character in both the third and fourth positions.
```
See http://unicode.org/reports/tr46/#Validity_Criteria.

So my understanding is that our IDNA lib will not accept punycode domains (names starting with `xn--`) once it has proper validation implemented.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -244,15 +244,18 @@ handle_request('GetAccountBalance', Req, _Context) ->
 handle_request('GetCommitmentHash', Req, _Context) ->
     Name         = maps:get('name', Req),
     Salt         = maps:get('salt', Req),
-    CHash        = aens:get_commitment_hash(Name, Salt),
-    EncodedCHash = aec_base58c:encode(commitment, CHash),
-    {200, [], #{commitment => EncodedCHash}};
+    case aens:get_commitment_hash(Name, Salt) of
```
lucafavatella: I see the commitment hash is computed on the ASCII. Ok.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -253,7 +256,12 @@ handle_request('GetName', Req, _Context) ->
             {200, [], #{name      => Name,
```
lucafavatella: I am a bit concerned returning the name of the entry.

1. Looking at `aens_claim_tx:check` I understand that ns_claim_tx does not need to have name already in form as returned by `aens_utils:to_ascii`;
2. Looking at `aens_claim_tx:process` I understand the actual name is not stored in the state tree (neither as specified in ns_claim_tx nor as `aens_utils:to_ascii`);
  1a. This seems not in line with [the spec](https://github.com/aeternity/protocol/blob/ce2af0674a5f427ddb2d9d5b547a302eeeeb81bb/AENS.md#aens-entry).
3. I see at least function `aec_conductor:get_name_entry` that queries the state tree passing the name as argument, then adding it in the return value - giving the impression to the reader that the name in that form is part of the state tree, while it is not;
4. I see when HTTP client queries `GetName` it gets back the name as passed in the request - giving the impression that the name is registered in that form - while it is not the case (rather the (hashed) ascii-fied name).

apps/aehttp/test/aehttp_integration_SUITE.erl
lucafavatella: > One small addition I am thinking about is, after the check that the name is present passing it as Unicode, performing the same get_name call passing the punycode representation of the domain.

Just a clarification: here I am not implying it shall succeed, though it would nice to have the assertion on how it shall behave - potentially commented out. (And yet, not needed for this PR.)

lucafavatella: > One small addition I am thinking about is, after the check that the name is present passing it as Unicode, performing the same get_name call passing the punycode representation of the domain.

Another small addition I am thinking about - not related to punycode this time - is, after the check that the name is present passing it as Unicode, performing the same get_name call passing a different Unicode representation of the same name (some Unicode strings have some "equivalent" Unicode representations e.g. composed vs. decomposed, and they do not need necessarily to fit any of the specified normalization forms while still being valid Unicode). (And yes, not needed for this PR.)

apps/aehttp/test/aehttp_integration_SUITE.erl
lucafavatella: > So my understanding is that our IDNA lib will not accept punycode domains (names starting with xn--) once it has proper validation implemented.

I suggest adding this as a test eventually (not this PR).

lucafavatella:
Travis CI is slow, though Linux package job ran on Circle CI ok. I checked on OSX builds.

PR #744, opened 2018-02-15
HTTP API for unsigned nameservice txs

velzevur:
Endpoints:
```
POST /tx/name/preclaim
POST /tx/name/claim
POST /tx/name/update
POST /tx/name/transfer
POST /tx/name/revoke
```


apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -520,9 +522,15 @@ contract_transactions(_Config) ->
 
     %% prepare a contract_create_tx and post it
     {ok, 200, #{<<"tx">> := EncodedUnsignedTx}} = get_contract_create(ValidEncoded),
-    sign_and_post_tx(EncodedUnsignedTx),
+    {ok, SerisalizedUnsignedTx} = aec_base58c:safe_decode(transaction, EncodedUnsignedTx),
```
hanssv: /Serisalized/Serialized

velzevur:
Fixed typo

apps/aehttp/priv/swagger.json
```diff
@@ -380,6 +380,186 @@
...
+          }
+        } ],
+        "responses" : {
+          "200" : {
+            "description" : "Successful operation",
+            "schema" : {
+              "$ref" : "#/definitions/Tx"
+            }
+          },
+          "400" : {
```
zp-sd: After https://github.com/aeternity/epoch/pull/743/files error msg `#{reason => <<"Name validation failed with a reason: ", ReasonBin/binary>>}` will be possible together with 400 response code, but I think this will cover it, right?

apps/aehttp/priv/swagger.json
```diff
@@ -2259,11 +2439,20 @@
...
+          "type" : "integer",
+          "format" : "int64"
         }
       },
       "example" : {
         "fee" : 0,
-        "commitment" : "commitment"
+        "commitment" : "commitment",
+        "nonce" : 6,
+        "account" : { }
```
zp-sd: Should we have real sample hash as example `account`? 🤔

apps/aehttp/priv/swagger.json
```diff
@@ -2280,12 +2469,21 @@
...
+          "format" : "int64"
         }
       },
       "example" : {
         "name_salt" : 0,
         "fee" : 6,
-        "name" : "name"
+        "name" : "name",
+        "nonce" : 1,
+        "account" : { }
```
zp-sd: Should we have real sample hash as example account? 🤔

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -2439,6 +2569,27 @@ get_oracle_response(Data) ->
...
+
+get_name_update(Data) ->
+    Host = external_address(),
+    http_request(Host, post, "tx/name/update", Data).
+
+get_name_transfer(Data) ->
+    Host = external_address(),
+    http_request(Host, post, "tx/name/transfer", Data).
+
+get_name_revoke(Data) ->
```
zp-sd: Why do we call these `get_`? Isn't `post_` a bit more natural?

apps/aehttp/priv/swagger.json
velzevur: We should define first so swagger-codegen could actually generate it...

lucafavatella:
@velzevur Could you please rebase?

velzevur:
Rebased and updated

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: the POST method is used out of neccesity. In fact we're *getting* the unsigned transaction (from a code point of view). Probably it is arguably which is better :(

apps/aehttp/src/aehttp_helpers.erl
```diff
@@ -157,6 +159,29 @@ verify_key_in_state_tree(Key, StateTreeFun, Lookup, Entity) ->
...
+        Name = maps:get(NameKey, State),
+        case aens:get_name_hash(Name) of
+            {ok, _} -> ok;
+            {error, Reason} ->
+                ReasonBin = atom_to_binary(Reason, utf8),
+                {error, {400, [],
+                        #{reason => <<"Name validation failed with a reason: ",
+                                                ReasonBin/binary>>}}}
+        end
+
```
zp-sd: 💅  redundant newline

apps/aehttp/src/aehttp_helpers.erl
```diff
@@ -157,6 +159,29 @@ verify_key_in_state_tree(Key, StateTreeFun, Lookup, Entity) ->
         end
     end.
 
+verify_name(NameKey) ->
+    fun(_Req, State) ->
+        Name = maps:get(NameKey, State),
+        case aens:get_name_hash(Name) of
+            {ok, _} -> ok;
+            {error, Reason} ->
+                ReasonBin = atom_to_binary(Reason, utf8),
```
zp-sd: As @lucafavatella pointed out in another PR it would be good to have specs that ensure we're getting atoms that will be converted to binary, so that we don't crash here.

Would you change spec of `aens:get_name_hash/1` from `{error, term()}` to `{error, atom()}`?

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -212,6 +214,86 @@ handle_request('PostOracleResponse', #{'OracleResponseTx' := Req}, _Context) ->
                                                  aec_tx:serialize_to_binary(Tx))}}
     end;
 
+handle_request('PostNamePreclaim', #{'NamePreclaimTx' := Req}, _Context) ->
+    ParseFuns = [parse_map_to_atom_keys(),
+                 read_required_params([account, commitment, fee]),
+                 base58_decode([{account, account, account_pubkey},
+                                {commitment, commitment, commitment}]),
+                 get_nonce(account)
```
lucafavatella: Currently considering whether external ok.

velzevur: We should either do this or provide an endpoint for getting once's nonce (which would probably be the same...)

lucafavatella:
I see 3 approvals. CI passed. No conflicts. I see this is API code only (looking up internals i.e. state trees and mempool). Merging.

PR #745, opened 2018-02-15
Avoid calling is_connected_to_genesis in vain

hanssv:


apps/aecore/src/aec_sync.erl
```diff
@@ -290,32 +290,41 @@ fetch_chain(Uri, GHash) ->
...
         {error, Reason} ->
             lager:debug("fetching top block (~p) failed: ~p", [Uri, Reason])
     end.
 
-fetch_chain(GHash, Uri, GHash, Acc) ->
+%% Fetch the chain, keep track of if we can be connected to genesis.
+%% Initially we can be connected, later we can only be connected
+%% after we have actually fetched a block from another peer, and then
+%% found a block locally (in the extreme case that local block will
+%% be the genesis).
```
lucafavatella: How I reasoned it is:
> If I attempt to fetch a block, I find it locally, it is not connected to genesis, and I find the previous block locally too, then this such previous block cannot be connected to genesis either.

hanssv: Yes, except for the first block (when we have fetched Top from another peer) that block might already be local but still connected to genesis!

PR #749, opened 2018-02-15
fix runtime modification of lager hwm

uwiger:
Ref. https://www.pivotaltracker.com/story/show/154207930

Only apply on error_logger and the file_backend (and apply correctly).

### DIFF
--- a/apps/aecore/src/aecore_app.erl
+++ b/apps/aecore/src/aecore_app.erl
@@ -74,12 +74,10 @@ set_hwm(HWM) when is_integer(HWM) ->
     if_running(lager, fun() -> live_set_hwm(HWM) end).
 
 live_set_hwm(Hwm) ->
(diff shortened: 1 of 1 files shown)


lucafavatella:
I built locally package, deployed as per release notes, set hwm in epoch.yaml to 51 then at start I get:
```
14:43:44.082 [warning] lager_error_logger_h dropped 0 messages in the last second that exceeded the limit of 51 messages/sec
```
.... so 51 is picked.

lucafavatella:
Starting with `./bin/epoch console -noshell -noinput` (as Docker image does) I managed to see that console sees the 51 limit:
```
14:53:11.102 [warning] lager_error_logger_h dropped 0 messages in the last second that exceeded the limit of 51 messages/sec
```
(I did not manage to hit this number outside startup.)

PR #750, opened 2018-02-15
PT-154994576 pin erlang otp osx

amir-hadi:
This PR should avoid updating erlang without intending to do so. With homebrew we can only install erlang by choosing the major version, like: `brew install erlang@20`. So as a workaround we have copied the formula as part of our repository and install it only using the formula from the repository.

The installed erlang version is 20.2.3.

The `ci` script has *not* been tested, as I don't know how to execute it. But when I do get it right, it is only used on Travis and will be removed anytime soon.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,7 +179,7 @@ jobs:
           command: |
             brew update
             brew install libsodium
-            brew install file://`pwd`/deployment/erlang.rb
+            brew install file://`pwd`/deployment/homebrew/erlang.rb
       - *build_package
       - *test_package
       - store_artifacts:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,7 +179,7 @@ jobs:
           command: |
             brew update
             brew install libsodium
-            brew install erlang
+            brew install file://`pwd`/deployment/erlang.rb
       - *build_package
       - *test_package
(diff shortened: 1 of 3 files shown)


lucafavatella:
> The ci script has not been tested, as I don't know how to execute it. But when I do get it right, it is only used on Travis and will be removed anytime soon.

It ran apparently correctly https://travis-ci.org/aeternity/epoch/jobs/341940306#L2042

amir-hadi:
@lucafavatella @dincho Please see the following:
```yaml
######################################################################## 100.0%
==> Pouring erlang-20.2.2.sierra.bottle.tar.gz
==> Caveats
Man pages can be found in:
  /usr/local/opt/erlang/lib/erlang/man

Access them with `erl -man`, or add this directory to MANPATH.
==> Summary
🍺  /usr/local/Cellar/erlang/20.2.2: 7,038 files, 277.3MB
```

This happened in [this build](https://circleci.com/gh/aeternity/epoch/2171), see step 'Install required tools'. You can check [the branch](https://github.com/aeternity/epoch/tree/PT-154994576-pin-erlang-otp-20.2.2-osx) to see the pinned version in `erlang.rb`.

lucafavatella:
I should have merged as squash-and-merge. Though not a big issue.

PR #756, opened 2018-02-16
Use a MPT as the cache to enable sharing and making the state smaller

gorillainduction:
https://www.pivotaltracker.com/n/projects/2124891/stories/154877960

### DIFF
--- a/deployment/ansible/deploy.yml
+++ b/deployment/ansible/deploy.yml
@@ -40,7 +40,7 @@
           port: 3013
       chain:
         persist: true
-        db_path: "./db31"
+        db_path: "./db32"
       keypair:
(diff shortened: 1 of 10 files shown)


deployment/ansible/deploy.yml
```diff
@@ -40,7 +40,7 @@
           port: 3013
       chain:
         persist: true
-        db_path: "./db30"
+        db_path: "./db31"
```
lucafavatella: Master already has db31 https://github.com/aeternity/epoch/blob/93b8c460d7551e290714a32233458a3bd6d0c62f/deployment/ansible/deploy.yml#L43

PR #757, opened 2018-02-16
use latest exometer_core commit

uwiger:
Ref https://www.pivotaltracker.com/story/show/155262302


lucafavatella:
I see 2 approvals. CI passed. Merging.

PR #758, opened 2018-02-16
Plant release notes for 0.7.0

lucafavatella:
In scope for https://www.pivotaltracker.com/story/show/154519751

TODO:
* [x] List changes since 0.6.0
  * Rough
* [x] Eye-ball

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,6 +1,6 @@
 # About this release
 
-[This release](https://github.com/aeternity/epoch/releases/tag/v0.7.0) is focused on smart contracts.
+[This release](https://github.com/aeternity/epoch/releases/tag/v0.7.0) is focused on smart contracts: it provides HTTP API for compiling and running smart contract and creating contract transactions.
 It also:
 * Refines the naming system by validating names and labels;
 * Reviews hashing in the blockchain, preferring BLAKE2b;
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,7 +1,13 @@
 # About this release
 
-Changes:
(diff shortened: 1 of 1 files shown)


RELEASE-NOTES.md
```diff
@@ -1,7 +1,13 @@
 # About this release
 
-Changes:
-* Add [small localnet](docs/docker.md#localnet) docker-compose configuration for testing and development purposes
+[This release](https://github.com/aeternity/epoch/releases/tag/v0.7.0) is focused on smart contracts: it provides HTTP API for compiling and running smart contract and creating contract transactions.
+It also:
```
evdoks: As I understand, the contracts are only executed _off-chain_, so it makes sense to add that: `Currently only off-chain execution of smart contracts is possible. On-chain execution is planned for the next release.`

RELEASE-NOTES.md
```diff
@@ -1,7 +1,13 @@
...
+* Refines the naming system by validating names and labels;
+* Reviews hashing in the blockchain, preferring BLAKE2b;
+* Expands the user API;
+* Adds a [small localnet](docs/docker.md#localnet) docker-compose configuration for testing and development purposes;
+* Refines the usage of the persisted database for the blockchain, optimizing memory usage;
+* Improves the stability of the testnet.
 
 This release introduces backward incompatible changes in the chain format.
```
evdoks: Is it due to switching to Blake2b? If yes, I would explicitly state that switching to the new hashing function introduces backward incompatible changes.

RELEASE-NOTES.md
lucafavatella: As discussed, on/off chain already mentioned in intended usage of API linked from release notes.

([Exact page](https://github.com/aeternity/protocol/blob/epoch-v0.7.0/epoch/api/contract_api_usage.md).)

PR #762, opened 2018-02-16
Fix Swagger generate test Python client following concurrent PRs

lucafavatella:


lucafavatella:
Change in one comment in one generated test file. Merging with one approval.

PR #769, opened 2018-02-17
PT-154980819 Stop resolving name in revoke period

zp-sd:


apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -2096,14 +2096,25 @@ naming_system_manage_name(_Config) ->
...
 
     %% Submit name transfer tx and check it is in mempool
-    NameRecipient      = random_hash(),
-    {ok, 200, _}       = post_name_transfer_tx(NHash, NameRecipient, Fee),
-    {ok, [TransferTx]} = rpc(aec_tx_pool, peek, [infinity]),
-    NameRecipient      = aens_transfer_tx:recipient_account(aec_tx_sign:data(TransferTx)),
+    {ok, DecodedPubkey} = aec_base58c:safe_decode(account_pubkey, PubKeyEnc),
+    {ok, 200, _}        = post_name_transfer_tx(NHash, DecodedPubkey, Fee),
+    {ok, [TransferTx]}  = rpc(aec_tx_pool, peek, [infinity]),
+    DecodedPubkey       = aens_transfer_tx:recipient_account(aec_tx_sign:data(TransferTx)),
```
hanssv: This will conflict with #770 but easy to resolve

apps/aens/src/aens_utils.erl
```diff
@@ -71,9 +79,9 @@ check_name_owner(Name, AccountPubKey) ->
     end.
 
 check_claimed_status(Name) ->
```
hanssv: The name of this function is slightly strange? `is_claimed`?

zp-sd: I wanted to follow `check_` prefix for all the checks. Or maybe you want to replace `is_revoked` with `is_claimed` (the first was slightly better in the other place, that's why I used it) 🤔 ?

But I'm not devoted to any of these, and can adjust if you prefer ☺️

hanssv: No strong preference - just found it slightly strange...

PR #770, opened 2018-02-19
TXs refactoring - PT-155121931 central adt for txs

hanssv:
This is a first version, it is complete, but I expect some comments on how to further improve this!


hanssv:
Fixed the conflict

hanssv:
Rebased on Luca's intermittent test failure fix to avoid the intermittent test failure :-)

deployment/ansible/deploy.yml
```diff
@@ -40,7 +40,7 @@
           port: 3013
       chain:
         persist: true
-        db_path: "./db30"
+        db_path: "./db31"
```
lucafavatella: Is the consensus affected?

hanssv: Yes, some tx types has changed name so their serialization will be different

lucafavatella: So as things stand now we shall bump block and genesis version. Though it can be done in separate PR.

apps/aecontract/src/aect_call_tx.erl
```diff
@@ -37,15 +36,14 @@
...
 
 
--define(CONTRACT_CALL_TX_TYPE, <<"contract_call">>).
 -define(CONTRACT_CALL_TX_VSN, 1).
 -define(CONTRACT_CALL_TX_FEE, 2).
 
--type call_tx() :: #contract_call_tx{}.
+-opaque tx() :: #contract_call_tx{}.
 
--export_type([call_tx/0]).
```
lucafavatella: This type is used somewhere though our Dialyzer config does not pick this. It looks like there are other cases like this.
I suggest merging this PR as is then type fixups (and Dialyzer improvement maybe) can find place in following PR.

----

Rough local testing result:
```
   0: Unknown type aect_call_tx:call_tx/0
   0: Unknown type aect_create_tx:create_tx/0
   0: Unknown type aens_claim_tx:claim_tx/0
   0: Unknown type aens_preclaim_tx:preclaim_tx/0
   0: Unknown type aens_transfer_tx:transfer_tx/0
   0: Unknown type aens_update_tx:update_tx/0
```

lucafavatella: I started it as https://www.pivotaltracker.com/story/show/155400295

deployment/ansible/deploy.yml
lucafavatella: Tracked as https://www.pivotaltracker.com/story/show/155400352 so we remember.

PR #771, opened 2018-02-19
Include fix for DB persisted backend

lucafavatella:


lucafavatella:
I see two approvals. Waiting for CI completing test run.

PR #777, opened 2018-02-20
make swagger-docs is required, there is no docs commited into repository

acidsennui:


.circleci/config.yml
```diff
@@ -173,6 +173,7 @@ jobs:
       - checkout
       - run:
           command: |
+            make make swagger-docs
```
lucafavatella: It does not make sense to have two make.

michalbagrowski:
OK no i see `make swagger-docs` are creating new swagger.json then one in repository
fixed.

PR #779, opened 2018-02-21
Address most Dialyzer warnings with option `unknown`

lucafavatella:
In scope for https://www.pivotaltracker.com/story/show/155400295

TODO:
* [ ] http_uri - will address in following PR


lucafavatella:
The http_uri changes are a bit more involved than these. I will merge this PR and address http_uri in follow up PR.

PR #781, opened 2018-02-21
Teach Dialyzer to check for unknown functions and types

lucafavatella:
**Do not merge as CI fails!**


apps/aecore/src/aec_peers.erl
```diff
@@ -60,27 +60,32 @@
...
 -record(peer, {
-          uri = ""          :: http_uri:uri(),   %% try not to use!
+          uri = ""          :: http_uri_uri(),   %% try not to use!
           scheme            :: http_uri:scheme(),
-          host              :: http_uri:host(),
-          port              :: http_uri:port(),
-          path              :: http_uri:path(),
+          host              :: http_uri_host(),
+          port              :: http_uri_port(),
+          path              :: http_uri_path(),
```
gorillainduction: You need `http_uri_path() | 'undefined'` here, or adapt init/0 to set this field.

lucafavatella: ... or delete path altogether as it seems unused.

apps/aens/src/aens_state_tree.erl
```diff
@@ -40,7 +40,7 @@
 -type nstree() :: aeu_mtrees:mtree(mkey(), mvalue()).
 -type commitment() :: aens_commitments:commitment().
 -type name() :: aens_names:name().
--type cache() :: aeu_mtrees:tree(cache_key(), cache_value()).
```
lucafavatella: Leftovers from recent merge.

apps/aecore/src/aec_peers.erl
```diff
@@ -64,7 +64,6 @@
           scheme            :: http_uri:scheme(),
           host              :: http_uri:host(),
           port              :: http_uri:port(),
-          path              :: http_uri:path(),
```
lucafavatella: Cc @ThomasArts

apps/aecore/src/aec_peers.erl
```diff
@@ -60,26 +60,30 @@
 
 %% We parse the uri's with http_uri, therefore we use types from that module.
 -record(peer, {
-          uri = ""          :: http_uri:uri(),   %% try not to use!
+          uri = ""          :: http_uri_uri(),   %% try not to use!
```
lucafavatella: These are definitely workarounds and OTP http_uri shall be patched, though we still need these workarounds for current OTP. I suggest merging these workarounds and I plan to see how it easy it is to put together patch from OTP.

lucafavatella:
I see `:0: Unknown type http_uri:uri/0`, I need to understand why. I am re-compiling locally from scratch.

lucafavatella:
> I see :0: Unknown type http_uri:uri/0, I need to understand why. I am re-compiling locally from scratch.

I see it is jesse.erl and jesse_state.erl

hanssv:
Looks good so far...

apps/aecore/src/aec_peers.erl
lucafavatella: I opened OTP PR 1724 after having tested it makes Dialyzer pass. Though not sure when it will be merged.

lucafavatella:
I force pushed:
* I rebased on master;
* I enriched the commit re workaround of http_uri types by depending on [a modified jesse version](https://github.com/aeternity/jesse/commit/4640ad182a4915f45734cdaf96dcf4b84ab2e04e). I did not propose this upstream as the copied type is applicable only to recent OTP versions;
* I reordered the commits so CI is meant to pass at each commit.

The proposed http_uri patch is submitted to OTP (1724) and already went through a round of review. Though it is not yet merged so it is not in any released tag. Hence I propose to merge this PR with the http_uri workarounds so that we have `unknown` Dialyzer option enabled in master.

PR #788, opened 2018-02-23
Preparing for - PT-154292757 oracle refund - move types from header files

hanssv:
While implementing oracle query refunds I struggled with types being placed in header files - this PR aims to make things a bit better...


apps/aecore/src/aec_target.erl
```diff
@@ -14,7 +14,7 @@
 %% Returns {error | chain_too_short_to_recalculate_target} if initial height is a negative value
```
uwiger: Not that it's part of this PR, but why `|` here?

hanssv: That is just a the comment (and the comment is wrong, it returns `{error, chain_too_short...}`)

PR #789, opened 2018-02-23
PT-154292757 Oracle query refund

hanssv:
When an Oracle Query TTL expire and the oracle hasn't answered - the sender should get a refund.

### DIFF
--- a/apps/aeoracle/src/aeo_state_tree.erl
+++ b/apps/aeoracle/src/aeo_state_tree.erl
@@ -244,7 +244,7 @@ oracle_refund(Q, H, ATree) ->
                 none ->
                     lager:error("Account ~p could not be found for refunding oracle query ~p",
                                 [aeo_query:sender_address(Q), aeo_query:id(Q)]),
-                    ATree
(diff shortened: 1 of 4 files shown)


apps/aeoracle/src/aeo_state_tree.erl
```diff
@@ -195,31 +193,62 @@ add_query(How, I, #oracle_tree{otree = OTree} = Tree) ->
...
+    {aeu_mtrees:delete(Id, OTree), ATree1}.
+
+oracle_refund(Q, H, ATree) ->
+    case aeo_query:is_closed(Q) of
+        false ->
+            case aec_accounts_trees:lookup(aeo_query:sender_address(Q), ATree) of
+                {value, Account} ->
+                    {ok, Account1} = aec_accounts:earn(Account, aeo_query:fee(Q), H),
+                    aec_accounts_trees:enter(Account1, ATree);
+                none ->
```
gorillainduction: I'm not entirely convinced that this is the correct behavior. It is a pretty bad situation, that currently cannot happen since there is no way of deleting an account.

Compare with a spend to an non-existing account. That would create the account, right?

PR #790, opened 2018-02-23
Pt 153478436 dont copy block to miner

gorillainduction:
Avoid excessive copying and recalculation of block candidate.

https://www.pivotaltracker.com/n/projects/2124891/stories/153478436

### DIFF
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -544,48 +544,66 @@ start_mining(#state{block_candidate = #candidate{top_hash = OldHash},
     %% Candidate generated with stale top hash.
     %% Regenerate the candidate.
     create_block_candidate(State#state{block_candidate = undefined});
-start_mining(#state{} = State) ->
+start_mining(#state{block_candidate = Candidate} = State) ->
     epoch_mining:info("Starting mining"),
-    #candidate{block = Block,
-               nonce = Nonce} = State#state.block_candidate,
-    Info = [{top_block_hash, State#state.seen_top_block_hash}],
+    BlockBin = Candidate#candidate.bin,
+    Nonce    = Candidate#candidate.nonce,
+    Target   = aec_blocks:target(Candidate#candidate.block),
+    Info     = [{top_block_hash, State#state.seen_top_block_hash}],
     aec_events:publish(start_mining, Info),
-    Fun = fun() -> {aec_mining:mine(Block, Nonce),
-                    State#state.block_candidate}
+    Fun = fun() ->
+                  {aec_mining:mine(BlockBin, Target, Nonce)
+                  , BlockBin}
           end,
     dispatch_worker(mining, Fun, State).
 
-handle_mining_reply({{ok, Block},_Candidate}, State) ->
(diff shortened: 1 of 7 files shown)


gorillainduction:
I forgot to run the eunit before making the PR. Will fix.

apps/aecore/src/aec_mining.erl
```diff
@@ -38,18 +38,8 @@ need_to_regenerate(Block) ->
...
-    BlockBin = aec_headers:serialize_for_hash(aec_blocks:to_header(Block)),
-    case aec_pow_cuckoo:generate(BlockBin, Target, Nonce) of
-        {ok, {Nonce, Evd}} ->
-            {ok, aec_blocks:set_pow(Block, Nonce, Evd)};
-        {error, no_solution} = Error ->
-            Error;
-        {error, {runtime, _}} = Error ->
-            Error
-    end.
+mine(BlockBin, Target, Nonce) ->
```
lucafavatella: Variable name without computation visible may confuse reader, I suggest renaming `BlockBin` to `HeaderBin`. Though totally optional of course.

gorillainduction: Agreed. This is a better name. I should also put the spec back.

apps/aecore/src/aec_conductor.erl
```diff
@@ -544,48 +544,66 @@ start_mining(#state{block_candidate = #candidate{top_hash = OldHash},
...
-    aec_metrics:try_update([ae,epoch,aecore,mining,blocks_mined], 1),
-    State1 = State#state{block_candidate = undefined},
-    case handle_mined_block(Block, State1) of
-        {ok, State2} ->
-            State2;
-        {{error, Reason}, State2} ->
-            epoch_mining:error("Block insertion failed: ~p.", [Reason]),
-            start_mining(State2)
+handle_mining_reply(_Reply, #state{block_candidate = undefined} = State) ->
+    %% Something invalidated the block candidate already.
```
lucafavatella: I see this case could have happened even before this change, though this clause was not there. So good to add. Ok.

apps/aecore/src/aec_conductor.erl
```diff
@@ -544,48 +544,66 @@ start_mining(#state{block_candidate = #candidate{top_hash = OldHash},
...
+            case handle_mined_block(Block, State1) of
+                {ok, State2} ->
+                    State2;
+                {{error, Reason}, State2} ->
+                    epoch_mining:error("Block insertion failed: ~p.", [Reason]),
+                    start_mining(State2)
+            end;
+        _Other ->
+            %% This mining effort was for an earlier block candidate.
+            epoch_mining:error("Found solution for old block", []),
```
lucafavatella: As the pid here is the one of the conductor - not the worker - and this case shall not happen often in my understanding, maybe operator would like the nonce logged here for correlating event that made this happen? (`Candidate#candidate.bin` may be too much info.)

gorillainduction: This is the case when a new candidate has been built because of some trigger (most likely, the top changed), but the old mining had already sent the answer. I'm not even sure that this can happen, since it should show up as an "unsolicited worker answer", but it is a bit of a safe guard.

PR #792, opened 2018-02-26
PT-155025536 Extend oracle TTL

hanssv:
Per request by the aepps team: https://github.com/aeternity/protocol/issues/39

Adding the possibility to extend the TTL of an oracle. (The extension is a delta TTL thus always positive which should make for very few surprises in 3rd party code.)

+ Tests, API endpoints and tests for API endpoints.


hanssv:
For clarity let's bump the DB-path...

hanssv:
Rebased and fixed stupid conflict

PR #793, opened 2018-02-26
Check locked deps on CI

lucafavatella:
This is the simplest proposal for https://www.pivotaltracker.com/story/show/155326038
Proof that it works: please see [this CI run on branch containing the changes in this PR cherry-picked on tag v0.7.0](https://circleci.com/gh/lucafavatella/epoch/706).

I considered investigating whether it is meaningful/feasible/cheap to rather add a `--dry-run` option to `rebar3 upgrade`, though I ran out of time today so I suggest to review&merge this as it is. It solves the problem of rebar.lock potentially out-of-sync.

### DIFF
--- a/scripts/rebar_lock_check
+++ b/scripts/rebar_lock_check
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+## Check rebar.lock is up-to-date.
+##
+## Copies `rebar.lock` file to temporary directory, then attempts to
+## re-generate it, checks it is the same, finally deletes the
+## temporary directory.
+##
+## The re-generation of the `rebar-lock` file is attempted by:
+## * First running `rebar3 unlock` - so to delete the old lock file;
+## * Then running `rebar3 upgrade` - so to generate a lock file
+##   attempting to minimize contamination by already-fetched
+##   dependencies (e.g. transitive ones) - if any.
+##
+## On failure neither the `rebar.lock` file nor the fetched
+## dependencies are restored.
+
+on_exit() {
+    rm -r "${TmpDir:?}"
+}
+
+set -e
+
+R="${1:?}" ## Absolute path of `rebar3` executable.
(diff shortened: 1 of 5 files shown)


scripts/rebar_lock_check
```diff
@@ -0,0 +1,42 @@
...
+## On failure neither the `rebar.lock` file nor the fetched
+## dependencies are restored.
+
+on_exit() {
+    rm -r "${TmpDir:?}"
+}
+
+set -e
+
+R="${1:?}" ## Absolute path of `rebar3` executable.
```
dincho: Can you please use more descriptive variable names? A/B/C/D vars really does not make sense to me, even in bash, even with the great comments you added.

Some suggestions:
- `rebar_bin` or `RebarBin` 
- `working_bin`
- `rebar_lock_path`

(btw we should agree on some style guide and follow it, e.g. https://google.github.io/styleguide/shell.xml)

scripts/rebar_lock_check
```diff
@@ -0,0 +1,42 @@
...
+## directory.
+LP="${D:?}"/rebar.lock ## Lock file path.
+OldLP="${TmpDir:?}"/rebar.lock.backup
+cp -p "${LP:?}" "${OldLP:?}"
+
+## Re-generate `rebar.lock` file.
+( cd "${D:?}" && "${R:?}" unlock && "${R:?}" upgrade; )
+
+## Check that re-generating `rebar.lock` led to same generated file.
+diff -u "${OldLP:?}" "${LP:?}"
```
dincho: Should we add `diffutils` package as explicit dependency in our docker builder image? I guess the risk of missing it is very low.

scripts/rebar_lock_check
lucafavatella: In the past when I used more descriptive variable names I received comments to use short var names (one or two chars max)... I agree with some one style guide.

lucafavatella:
I see two approvals, CI passed, master is stable. Merging.

PR #794, opened 2018-02-26
Pt 155517005 api header by hash

ThomasArts:
Add an API to get the header instead of the block for a specific hash.

This is needed to build the chain backward when we synchronize with a node that is ahead of us.

### DIFF
--- a/apps/aehttp/src/aehttp_dispatch_ext.erl
+++ b/apps/aehttp/src/aehttp_dispatch_ext.erl
@@ -81,6 +81,25 @@ handle_request('GetBlockByHash' = _Method, Req, _Context) ->
             end
     end;
 
+handle_request('GetHeaderByHash', Req, _Context) ->
+    case aec_base58c:safe_decode(block_hash, maps:get('hash', Req)) of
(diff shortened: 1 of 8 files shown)


apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -81,6 +81,27 @@ handle_request('GetBlockByHash' = _Method, Req, _Context) ->
...
+    case aec_base58c:safe_decode(block_hash, maps:get('hash', Req)) of
+        {error, _} ->
+            {400, [], #{reason => <<"Invalid hash">>}};
+        {ok, Hash} ->
+            case aec_chain:get_header(Hash) of
+                {ok, Header} ->
+                    %% swagger generated code expects the Resp to be proplist
+                    %% or map and always runs jsx:encode/1 on it - even if it
+                    %% is already encoded to a binary; that's why we use
+                    %% aec_blocks:serialize_to_map/1
```
hanssv: Ehm? No. We are using aec_headers:serialize_to_map/1

PR #796, opened 2018-02-26
Unique coinbase transactions

velzevur:
* removed core_txs.hrl
* added a block height field to coinbase transactions


apps/aecore/src/aec_coinbase_tx.erl
```diff
@@ -21,15 +21,17 @@
 
 -include("common.hrl").
 
--record(coinbase_tx, {account   = <<>> :: pubkey()}).
+-record(coinbase_tx, {account       = <<>> :: pubkey(),
+                      block_height  = -1 :: integer()}).
```
lucafavatella: May I suggest stronger type (proper height - not negative - so Dialyzer checks stronger)? Default may be omitted too (that is my preference) though I understand it is debatable.

apps/aecore/src/aec_coinbase_tx.erl
```diff
@@ -72,18 +77,21 @@ signers(#coinbase_tx{account = AccountPubkey}) -> [AccountPubkey].
 -define(CB_TX_VSN, 1).
 
 -spec serialize(tx()) -> [map()].
-serialize(#coinbase_tx{account = Account}) ->
+serialize(#coinbase_tx{account = Account, block_height = Height}) ->
     [#{<<"vsn">> => version()},
-     #{<<"acct">> => Account}].
+     #{<<"acct">> => Account},
+     #{<<"h">> => Height}].
```
lucafavatella: This affects the txs Merkle tree root hash in the block header hence consensus. Please bump db_path in deploy script.

apps/aetx/test/aetx_tests.erl
```diff
@@ -33,7 +33,8 @@ apply_signed_txs_test_() ->
                               [MinerAccount, AnotherAccount]),
 
                %% Create 3 signed transactions (2 valid, 1 invalid)
-               {ok, CoinbaseTx} = aec_coinbase_tx:new(#{account => MinerPubkey}),
+               {ok, CoinbaseTx} = aec_coinbase_tx:new(#{account => MinerPubkey,
+                                                        block_height => 30}),
```
lucafavatella: I see this 30 is the same as in https://github.com/aeternity/epoch/pull/796/files#diff-3da268090e9138b3876166f3da939a52R55
May I suggest to use variable in order to ease readability and maintainability?

apps/aecore/src/aec_coinbase_tx.erl
```diff
@@ -22,7 +22,7 @@
 -include("common.hrl").
 
 -record(coinbase_tx, {account       = <<>> :: pubkey(),
-                      block_height  = -1 :: integer()}).
```
ThomasArts: I believed you used the -1 as a placeholder for "undefined". Now your initial record has an undefined value, which is not an integer and also not a non_neg_integer(). This is only an issue if you ever use #coinbase_tx{} without setting the block_height field.

velzevur: Yes. We must not have coinbases without block height included.

PR #800, opened 2018-02-27
HTTP API /tx/contract/create enhancement

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/155553440)
Small enhancements:
* `/tx/contract/create` returns the contract's address as well as the transaction and transaction hash. This contract address will be valid as soon as the transaction is signed, posted an included in block
* contract addresses are not encoded anymore (treated as strings)
* oracle addresses are encoded as `oracle_pubkey`-s


apps/aehttp/src/swagger/swagger_external_handler.erl
```diff
@@ -313,6 +313,7 @@ allowed_methods(Req, State) ->
 
 
 
+
```
dincho: what's wrong with the spacing here?

velzevur: That would be swagger-codegen

config/swagger.yaml
```diff
@@ -2732,6 +2732,16 @@ definitions:
     required:
     - tx
     - tx_hash
+  CreateContractUnsignedTx:
+    allOf:
```
lucafavatella: I see https://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-5.5.3.2
Ok.

PR #804, opened 2018-02-28
PT-155586216: Add pubkey to ping message

eproxus:
Open questions:

- [ ] Should the field be called `pub_key`? What if we later want to base it on something else or hash it?
- [ ] Do we need specific tests for this field? We have `correct_ping` and `broken_pings` tests already in `aehttp_integration_SUITE` already, which cover basic stuff like missing fields...


eproxus:
There is a test testing that an object `#{}` returns `400 Bad Request` which implicitly tests this. It is also a required field in the Swagger specification. Adding manual verification for each field seems like a chore, and could better be done in automated tests.

ThomasArts:
So what happens if an old node (not having a pub_key in the ping) talks to a new node? Is there an error that blocks the peer, are we just not responding? It seems that you claim it is a backward incompatible change and old nodes will not talk to us. If so, we need to bump the protocol version when this comes in a release.

eproxus:
I guess right now that node would get a 400 from it's pings to a newer node, because the request would not be valid (missing the required `pub_key` field). If this needs to be a special error, we need more code of course.

eproxus:
@ThomasArts Should we bump the Swagger version? To `0.8.0`?

lucafavatella:
> @ThomasArts Should we bump the Swagger version? To 0.8.0?

@eproxus Currently we bump the Swagger version at each release https://github.com/aeternity/epoch/wiki/Release-process
I suggest to keep this out of this PR.

eproxus:
Closing because of moving to Noise

dincho:
@eproxus please cleanup your branches (on remote - GH)

PR #807, opened 2018-02-28
Update dialyzer cache plt version

dincho:
Unfortunately it seems CircleCI does not support wildcards in path.

https://www.pivotaltracker.com/story/show/155250917

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,7 +173,7 @@ jobs:
       - save_cache:
           key: dialyzer-cache-v2-{{ .Branch }}-{{ .Revision }}
           paths:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -173,7 +173,7 @@ jobs:
       - save_cache:
           key: dialyzer-cache-v2-{{ .Branch }}-{{ .Revision }}
           paths:
-            - _build/default/rebar3_20.2.2_plt
+            - _build/default/rebar3_20.2.3_plt
```
lucafavatella: I see this follows https://github.com/aeternity/docker-builder/pull/2/files

PR #810, opened 2018-02-28
Teach Circle CI to merge master on PR when PR from fork

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/155429174

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ references:
       name: Checkout merged master
       command: |
         if [ -n "$CIRCLE_PULL_REQUEST" ]; then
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -15,7 +15,7 @@ references:
       name: Checkout merged master
       command: |
         if [ -n "$CIRCLE_PULL_REQUEST" ]; then
-          git fetch origin +refs/pull/${CIRCLE_PULL_REQUEST##*/}/merge
+          git fetch https://github.com/aeternity/epoch.git +refs/pull/${CIRCLE_PULL_REQUEST##*/}/merge
```
dincho: Let's use SSH instead of HTTPS, this is how CCI authenticates, otherwise public checkout limit applies IIRC

PR #821, opened 2018-03-02
Minimize effort required for maintaining release notes

lucafavatella:
Finishes https://www.pivotaltracker.com/story/show/155651305


docs/release-notes/RELEASE-NOTES-0.9.0.md
```diff
@@ -6,45 +6,51 @@ It:
...
 
 ## Retrieve the software for running a node
 
-Download the [release binary](https://github.com/aeternity/epoch/releases/tag/v0.9.0) corresponding to your platform, e.g. `epoch-0.9.0-osx-10.12.6.tar.gz`; you would normally find the downloaded package in `~/Downloads` on macOS.
+You can run a node by using either:
+* The published [release binary](https://github.com/aeternity/epoch/releases/tag/v0.9.0) corresponding to your platform; or
+* The published [Docker image `aetrnty/epoch`](https://github.com/aeternity/epoch/blob/v0.9.0/docs/docker.md).
 
-The binaries are tested on the following platforms:
+The release binaries are tested on the following platforms:
```
dincho: I think this should be moved to installation.md, mentioning the releases are published on GitHub.

docs/release-notes/RELEASE-NOTES-0.9.0.md
```diff
@@ -6,45 +6,51 @@ It:
...
+The instructions for installing a node using a release binary are in [the dedicated separate document](../../docs/installation.md).
 
-HTTP API endpoints are specified in the [swagger.yaml](https://github.com/aeternity/epoch/blob/v0.9.0/config/swagger.yaml); a swagger.json version of the same specification is present in the release binary at path `lib/aehttp-0.1.0/priv/swagger.json`, and its interactive visualization is available [online](https://aeternity.github.io/epoch-api-docs/?config=https://raw.githubusercontent.com/aeternity/epoch/v0.9.0/apps/aehttp/priv/swagger.json).
-WebSocket API endpoints are [specified](https://github.com/aeternity/protocol/blob/epoch-v0.9.0/epoch/api/README.md).
-
-The intended usage of the API (HTTP and WebSocket) of the node is [documented](https://github.com/aeternity/protocol/blob/epoch-v0.9.0/epoch/api/README.md).
+For installation of a node using the Docker image, please refer to [its documentation online](https://github.com/aeternity/epoch/blob/v0.9.0/docs/docker.md).
 
 ## Join the testnet
```
dincho: I think this section ca be safely moved to configuration.md section

docs/release-notes/RELEASE-NOTES-0.9.0.md
```diff
@@ -6,45 +6,51 @@ It:
...
-The binaries are tested on the following platforms:
+The release binaries are tested on the following platforms:
 * Ubuntu 16.04.3 LTS (x86-64);
 * macOS Sierra (x86-64);
 * macOS High Sierra (x86-64).
 
-The macOS package has a hard dependency on OpenSSL v1.0.0 installed with [Homebrew](https://brew.sh/) in its default path `/usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib` and also a hard
-dependency on libsodium v1.0.16 installed with [Homebrew](https://brew.sh/) also in its default path `/usr/local/opt/libsodium/lib/libsodium.23.dylib`.
-In case you have installed either of them in a non-default path, you could use symlink(s) to work around the issue.
+The user configuration is documented in the [wiki](https://github.com/aeternity/epoch/wiki/User-provided-configuration).
```
dincho: IMO this line belongs to configuration.md

docs/release-notes/RELEASE-NOTES-0.9.0.md
```diff
@@ -6,45 +6,51 @@ It:
...
 
 Please join the testnet by following the instructions below, and let us know if you have any problems by [opening a ticket](https://github.com/aeternity/epoch/issues).
 Troubleshooting of common issues is documented [in the wiki](https://github.com/aeternity/epoch/wiki/Troubleshooting).
 
 The instructions below describe:
 * [How to retrieve the released software for running a node](#retrieve-the-software-for-running-a-node);
+* [How to install a node](#install-node);
 * [How to join the testnet](#join-the-testnet).
 
 ## Retrieve the software for running a node
```
dincho: Also assuming my comments below, I also think all the content below could be reduced to simple list of documentation links.

docs/installation.md
```diff
@@ -0,0 +1,35 @@
...
+In the instructions below, the node is deployed in directory `/tmp/node`: you may prefer to deploy the node in an alternative (and less ephemeral) location - e.g. a `node` directory inside your home directory - by amending the instructions accordingly.
+It is recommended that the partition where the node directory is has at least 10 GB free: this is needed for the chain and the log files.
+
+Open a Terminal window or get to the command line.
+
+Create a directory and unpack the downloaded package (you may need to amend the directory and/or file name of the package):
+```bash
+mkdir /tmp/node
+cd /tmp/node
+tar xf ~/Downloads/epoch-0.9.0-osx-10.12.6.tar.gz
```
dincho: we can use wildcard here `tar xf ~/Downloads/epoch-*.tar.gz` ?

docs/release-notes/RELEASE-NOTES-0.9.0.md
```diff
@@ -6,45 +6,51 @@ It:
...
 
-Download the [release binary](https://github.com/aeternity/epoch/releases/tag/v0.9.0) corresponding to your platform, e.g. `epoch-0.9.0-osx-10.12.6.tar.gz`; you would normally find the downloaded package in `~/Downloads` on macOS.
+You can run a node by using either:
+* The published [release binary](https://github.com/aeternity/epoch/releases/tag/v0.9.0) corresponding to your platform; or
```
dincho: you can use https://github.com/aeternity/epoch/releases/latest

docs/release-notes/RELEASE-NOTES-0.9.0.md
lucafavatella: Agreed. Addressed in 06e9974b

docs/installation.md
lucafavatella: What if user has multiple archives there? I do not like hardcoding package version and osx version, though (mitigated with "amend" comment in parentheses) I think it is still a better user experience.

I think potential better alternative would be something like "Let us assume you downloaded release binary and it has absolute path `RELEASE_ARCHIVE` ..."

docs/release-notes/RELEASE-NOTES-0.9.0.md
lucafavatella: You mean mainly the IP addresses of testnet?

docs/release-notes/RELEASE-NOTES-0.9.0.md
lucafavatella: I considered that, though I would prefer not to perform this change in this iteration.

I feel there is distinction between how node *can* be configured, and how node is *meant* to be configured. I kept this line for the former, and I put the latter in configuration.md, whose scope I describe as "how to configure your epoch node installed using a release binary for joining a public network of nodes (e.g. testnet) knowing an initial network peer to join."

docs/release-notes/RELEASE-NOTES-0.9.0.md
lucafavatella: I am open to counter-proposal. Though after considering that I preferred keeping all separate files focused. I.e. docker.md has all Docker related, while the new {configuration,installation,operation}.md files are focused on release binary only.

lucafavatella:
I added a70f468 proposing further (intended) simplification.

docs/release-notes/RELEASE-NOTES-0.9.0.md
```diff
@@ -1,50 +1,63 @@
...
-Download the [release binary](https://github.com/aeternity/epoch/releases/tag/v0.9.0) corresponding to your platform, e.g. `epoch-0.9.0-osx-10.12.6.tar.gz`; you would normally find the downloaded package in `~/Downloads` on macOS.
+You can run a node by using either:
+* The published [release binary][this-release] corresponding to your platform; or
+* The published [Docker image `aetrnty/epoch`][docker].
 
-The binaries are tested on the following platforms:
-* Ubuntu 16.04.3 LTS (x86-64);
-* macOS Sierra (x86-64);
-* macOS High Sierra (x86-64).
+[dockerTODO]: https://github.com/aeternity/epoch/blob/master/v0.9.0/docker.md
```
dincho: todo?

lucafavatella: > Also propose make most links point to current branch during development.

I propose that links link to master during development, and we change them to tag before release. These TODOs are meant to be placeholders of which links shall be amended. If it looks too complex I am happy to revert back to hardcoding tag number (e.g. 0.9.0) even during development.

lucafavatella: BTW your comment made me see a typo, that I fixed in 98cbabc

PR #823, opened 2018-03-06
PT-155506940 update sophia parser

UlfNorell:
This closes [PT-155506940](https://www.pivotaltracker.com/story/show/155506940).

The new syntax gets rid of curly braces and semicolons in favour of Python-style layout rules.


apps/aering/src/aer_constants.erl
```diff
@@ -0,0 +1,10 @@
+-module(aer_constants).
```
happi: This could be moved into aer_parser as a function (constant_string or similar).
There should only be one call to this function as far as I know.
Can be fixed as a seprate ticket later.

PR #825, opened 2018-03-06
PT-154885468:  Use cowboy to handle http api requests

fabiankrol:
General TODOs:
- use configuration for http acceptors pool 
- add tests for requests with wrong http method

Following steps (in order): 
1. use custom validation instead of swagger's one (reuse: https://github.com/aeternity/epoch/commit/f68db7625e9243c1d05456af6254750749279bf4)
2. validate `aehttp_dispatch_{ext,int}` response 
3. validate http client request (ref: https://github.com/aeternity/epoch/commit/f68db7625e9243c1d05456af6254750749279bf4)
4. ...
5. remove swagger generated code

### DIFF
--- a/apps/aehttp/src/aehttp_api_handler.erl
+++ b/apps/aehttp/src/aehttp_api_handler.erl
@@ -1,0 +1,50 @@
+-module(aehttp_api_handler).
+
+-export([allowed_methods/2]).
+-export([init/3]).
+-export([rest_init/2]).
+-export([content_types_accepted/2]).
+-export([content_types_provided/2]).
+-export([delete_resource/2]).
+-export([handle_request_json/2]).
+
+-record(state, {
+    operation_id :: atom(),
+    operation_params :: #{},
+    allowed_methods :: [binary()],
+    logic_handler :: atom()
+}).
+
+init(_Transport, Req, Opts) ->
+    {upgrade, protocol, cowboy_rest, Req, Opts}.
+
+rest_init(Req, {OperationId, AllowedMethods, LogicHandler, Params}) ->
+    State = #state{
+        operation_id = OperationId,
+        operation_params = Params,
+        allowed_methods = AllowedMethods,
+        logic_handler = LogicHandler
+    },
+    {ok, Req, State}.
+
+allowed_methods(Req, State = #state{ allowed_methods = Methods }) ->
+    {Methods, Req, State}.
+
+content_types_accepted(Req, State) ->
+    {[{<<"application/json">>, handle_request_json}], Req, State}.
+
+content_types_provided(Req, State) ->
+    {[{<<"application/json">>, handle_request_json}], Req, State}.
+
+delete_resource(Req, State) ->
+    handle_request_json(Req, State).
+
+handle_request_json(Req0, State = #state{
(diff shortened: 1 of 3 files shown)


apps/aehttp/src/aehttp_api_handler.erl
```diff
@@ -0,0 +1,50 @@
...
+
+delete_resource(Req, State) ->
+    handle_request_json(Req, State).
+
+handle_request_json(Req0, State = #state{
+        operation_id = OperationId,
+        operation_params = Params,
+        logic_handler = LogicHandler
+    }) ->
+    Context = #{},
```
ThomasArts: Could we put the source IP from the Req0 in this Context?

apps/aehttp/src/aehttp_api_handler.erl
```diff
@@ -0,0 +1,50 @@
...
+    {Code, Headers, Body} = LogicHandler:handle_request(OperationId, Params, Context),
+    {ok, Req} = cowboy_req:reply(Code, Headers, jsx:encode(Body), Req0),
```
ThomasArts: Strange variable name Req, should it not be a Reply?

apps/aehttp/src/aehttp_app.erl
```diff
@@ -47,28 +46,27 @@ stop(_State) ->
...
+start_http_api() ->
+    ok = start_http_api(external, aehttp_dispatch_ext, swagger_external_handler),
+    ok = start_http_api(internal, aehttp_dispatch_int, swagger_internal_handler).
+
+start_http_api(Target, LogicHandler, SwaggerHandler) ->
+    PoolSize = get_http_api_acceptors(Target),
+    Port = get_http_api_port(Target),
+    ListenAddress = get_http_api_listen_address(Target),
+
+    [{'_', Paths0}] = swagger_router:get_paths(LogicHandler),
```
ThomasArts: do we need to generate router code ourselfs as well?

apps/aehttp/src/aehttp_api_handler.erl
velzevur: This is cowboy_req and it is fine

apps/aehttp/src/aehttp_api_handler.erl
fabiankrol: Yes, we can put here any low level information needed by business layer. IMHO adding this should be done in separate PR.

apps/aehttp/src/aehttp_app.erl
michalzajda: there is 'trails' library for that, but I am not sure if we should start with it

apps/aehttp/src/aehttp_api_handler.erl
fabiankrol: `cowboy_req:reply` returns modified `Req0`. It's still the same `Req*` object hence naming.

apps/aehttp/src/aehttp_app.erl
fabiankrol: Yes, all the request to `swagger_*` modules should be replaced by our generators

kkszysiu:
Good to merge.

PR #827, opened 2018-03-06
Bugfix: HTTP endpoint /account/txs/{address}

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/155732977)
Adding transactions in the mempool caused crashes in the endpoint `GET /account/txs/{address}`

### DIFF
--- a/apps/aehttp/src/aehttp_dispatch_int.erl
+++ b/apps/aehttp/src/aehttp_dispatch_int.erl
@@ -881,8 +881,8 @@ get_account_transactions(Account, Req) ->
                   fun({mempool, SignedTxA}, {mempool, SignedTxB}) ->
                       TxA = aetx_sign:tx(SignedTxA),
                       TxB = aetx_sign:tx(SignedTxB),
-                      {aetx:origin(TxA), aetx:nonce(TxA)} >
-                      {aetx:origin(TxB), aetx:nonce(TxB)};
+                      {aetx:origin(TxA), aetx:nonce(TxA), TxA} >=
+                      {aetx:origin(TxB), aetx:nonce(TxB), TxB};
                      ({mempool, _}, {_, _}) -> true;
                      ({_, _}, {mempool, _}) -> false;
                      ({HeaderA, SignedTxA}, {HeaderB, SignedTxB}) ->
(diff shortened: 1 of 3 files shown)


lucafavatella:
It may be worth [a note in the release notes](https://github.com/aeternity/epoch/blob/1ecddc43eef7d9cbd6629bd117f0fb2ed6115588/docs/release-notes/RELEASE-NOTES-0.9.0.md#about-this-release) e.g. "Improves the stability of the testnet." (catchall).

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -876,7 +878,14 @@ get_account_transactions(Account, Req) ->
                                               Account),
             Res =
               lists:sort(
-                  fun({HeaderA, SignedTxA}, {HeaderB, SignedTxB}) ->
+                  fun({mempool, SignedTxA}, {mempool, SignedTxB}) ->
+                      TxA = aetx_sign:tx(SignedTxA),
+                      TxB = aetx_sign:tx(SignedTxB),
+                      {aetx:origin(TxA), aetx:nonce(TxA)} >
+                      {aetx:origin(TxB), aetx:nonce(TxB)};
```
lucafavatella: Do I understand correctly this line is meant to sort txs from same origin account in descending order of nonce? (It appears to me what it is doing based on code inspection.)

I understand this is not the first point in the function that introduced this approach, though I am trying to understand. From lists:sort/2 doc: "Fun(A, B) is to return true if A compares less than or equal to B in the ordering, otherwise false."

Other thing I am not sure about is equality. If order is meant to be in descending nonce order, I would expect `>=` not `>`.

(I picked the nonce though the principle may apply to all of the function.)

velzevur: > Do I understand correctly this line is meant to sort txs from same origin account in descending order of nonce? (It appears to me what it is doing based on code inspection.)
The transactions in this list are all with the same account somehow involved (as being one of aetx:accounts/1) but not mandatorary the origin (in the context of a spend_tx - the account could be the reciever as well).

PR #828, opened 2018-03-06
Add UAT chain snapshots

dincho:
EDIT: [Example build.](https://circleci.com/gh/aeternity/epoch/3742#artifacts/containers/0) It's using integration env as the UAT is not updated with Mnesia fixes yet.

[PT 155625323](https://www.pivotaltracker.com/story/show/155625323)

### DIFF
--- a/deployment/ansible/mnesia_backup.yml
+++ b/deployment/ansible/mnesia_backup.yml
@@ -1,0 +1,37 @@
+---
+# Playbook used to download (archived) Mnesia backup from last host in the playbook hosts
+#
+# By default the archive destination is overwritten if already exists,
+# if archive overwrite is not intended the caller is responsible for setting correct (e.g. unique) "backup_suffix"
+#
+# Example steps to restore a mnesia backup:
+# 1. Stop the node
+# 2. Delete existing mnesia database if any
+# 3. Disable mining autostart and empty the peers list in the config
+# 4. Start the node
+# 5. Restore the backup with "mnesia:restore("/tmp/mnesia.bak", [])."
+# 6. Enable mining autostart and restore the peers list in the config
+# 7. Start the node
+
+- name: Backup, archive and download Mnesia database
+  hosts: "{{ ansible_play_hosts | last }}"
+  remote_user: epoch
+
+  vars:
+    epoch_bin: "{{ ansible_env.HOME }}/node/bin/epoch"
+    download_dir: /tmp/mnesia_backups
+    backup_suffix: backup
+
+  tasks:
+    - name: Backup Mnesia database
+      command: "{{ epoch_bin }} eval 'ok = mnesia:backup(\"/tmp/mnesia_{{ ansible_hostname }}_{{ backup_suffix }}\").'"
+
+    - name: Archive Mnesia backup
+      archive:
+        path: /tmp/mnesia_{{ ansible_hostname }}_{{ backup_suffix }}
+        dest: /tmp/mnesia_{{ ansible_hostname }}_{{ backup_suffix }}.gz
+
+    - name: Download the Mnesia backup archive
+      fetch:
+        src: /tmp/mnesia_{{ ansible_hostname }}_{{ backup_suffix }}.gz
+        dest: "{{ download_dir }}"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,6 +46,25 @@ references:
     restore_cache:
       key: *build_cache_key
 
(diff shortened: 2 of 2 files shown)


deployment/ansible/download_chain.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    - name: Parse YAML configuration
+      set_fact:
+        epoch_vars: "{{ cat_output.stdout | from_yaml }}"
+      when: config_file.stat.exists
+
+    - name: Archive blockchain
+      archive:
+        path: "{{ project_root }}/{{ epoch_vars.chain.db_path }}/"
+        dest: "{{ ansible_env.HOME }}/{{ epoch_vars.chain.db_path }}.tgz"
```
lucafavatella: This path is created on the node the Ansible client runs i.e. the CI executor, right?

lucafavatella: I mean - if this creates the dest file on the managed node i.e. the UAT node it is not ok.

dincho: It's on the managed node, what's your concerns ?

lucafavatella: > It's on the managed node, what's your concerns ?

Path may be already present. What is behaviour then?
If script happens to run on Sun and Tue on same node and db_path is still same in deployment, issue in script somehow may lead to old archive copied.

If file created in managed node, as I now understand it is, may I suggest naming archive timestamp+db_path (or better timestamp+nodename+db_path) rather than only db_path? This would prevent any silent copying of pre-existing archive.

dincho:
Let's wait https://github.com/aeternity/mnesia_rocksdb/pull/5

deployment/ansible/download_chain.yml
dincho: > Path may be already present. What is behaviour then?

It overwrites it. Already tested.

dincho: `db_path` is as the name implies filesystem path, not suitable to be used in the filenames as is.
I prefer to keep ansible playbooks as simple as possible, thus just added a `archive_suffix` variable that can be used for uniqueness.

The changes are in 20bd2918

deployment/ansible/mnesia_backup.yml
```diff
@@ -0,0 +1,60 @@
...
+        - ping.stdout == "pong"
+
+    - name: Archive Mnesia backup
+      archive:
+        path: /tmp/mnesia_{{ ansible_hostname }}_{{ backup_suffix }}
+        dest: /tmp/mnesia_{{ ansible_hostname }}_{{ backup_suffix }}.gz
+      when: 
+        - epoch.stat.exists
+        - ping.stdout == "pong"
+        - backup.stdout == "ok"
```
lucafavatella: Does it mean that if `mnesia:backup` returned `{error, Reason}` then playbook succeeds (i.e. only some "skipped" so passed and exit code 0)?

deployment/ansible/mnesia_backup.yml
```diff
@@ -0,0 +1,60 @@
...
+      failed_when: no
+      register: ping
+      when: epoch.stat.exists
+
+    - name: Backup Mnesia database
+      command: "{{ project_root }}/bin/epoch eval 'mnesia:backup(\"/tmp/mnesia_{{ ansible_hostname }}_{{ backup_suffix }}\").'"
+      register: backup
+      when: 
+        - epoch.stat.exists
+        - ping.stdout == "pong"
```
lucafavatella: Does it mean that if node down then playbook succeeds?

deployment/ansible/mnesia_backup.yml
```diff
@@ -0,0 +1,60 @@
...
+    - name: Check if epoch binary exists
+      stat:
+        path: "{{ project_root }}/bin/epoch"
+      register: epoch
+
+    - name: Ping epoch daemon
+      command: "{{ project_root }}/bin/epoch ping"
+      failed_when: no
+      register: ping
+      when: epoch.stat.exists
```
lucafavatella: Does it mean that if epoch not deployed then playbook succeeds?

dincho:
> Unclear error handling, requested clarifications.

@lucafavatella I'm just used to not fail a playbooks. Error handling has been improved in 5e75f0d49a4b3f053fc7ebed71b82cad5bfbcfd8

uwiger:
A future extension could be to make a more complete backup set, e.g.:
* database
* any on-site patches (we don't really have that now)
* the `epoch.yaml` file

dincho:
@uwiger The only idea of this feature right now is that someone can grab the whole blockchain at some state and plant it locally for testing. It's not really a backup.

PR #829, opened 2018-03-06
Update sync to load chain from below

ThomasArts:
This fixes long delays in synchronization (or failing to do so) by starting to synchronize a chain from a joinedly agreed block upward.

We use a binary search to find this block we agree upon and then ask the chain with highest difficulty for the blocks building towards that highest difficulty.

If we get another ping request from the peer, we will not start a second sync process.
If we get another ping from a different peer, we will also start synchronizing against that peer, but by first getting the header and only in case the header is not present, also getting the block.
This is subject to later optimizations.

In addition:
Add client side of GetHeaderByHeight
Add API GetBlockByHeight
Introduce a sync pool

### DIFF
--- a/apps/aecore/src/aec_sync.erl
+++ b/apps/aecore/src/aec_sync.erl
@@ -37,6 +37,8 @@
 -compile([export_all, nowarn_export_all]).
 -endif.
 
+-record(sync_peer, {difficulty, from, to, hash, uri}).
+
 %%%=============================================================================
 %%% API
 %%%=============================================================================
@@ -104,7 +106,7 @@ compare_ping_objects(RemoteUri, Local, Remote) ->
                          lager:debug("Our difficulty is higher", []),
                          server_get_missing_blocks(RemoteUri);
                        true ->
-                         start_sync(RemoteUri)
+                         start_sync(RemoteUri, maps:get(<<"best_hash">>, Remote), Dr)
                     end
             end,
             fetch_mempool(RemoteUri),
@@ -114,8 +116,8 @@ compare_ping_objects(RemoteUri, Local, Remote) ->
     end.
 
 
-start_sync(Uri) ->
-    gen_server:cast(?MODULE, {start_sync, Uri}).
+start_sync(Uri, RemoteHash, RemoteDifficulty) ->
+    gen_server:cast(?MODULE, {start_sync, Uri, RemoteHash, RemoteDifficulty}).
 
 server_get_missing_blocks(Uri) ->
     gen_server:cast(?MODULE, {server_get_missing, Uri}).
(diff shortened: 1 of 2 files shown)


ThomasArts:
Did an experiment: On a 2 node cluster with a fork at block 243, longest fork up to 262, shorter fork up to 246 it takes:
Old solution: 88 seconds
New solution: 6 seconds

If just starting from zero on one node, it takes about 3 blocks per second to sync in both cases (new solution slightly faster). But old solution can break and then needs to restart from scratch, whereas new solution can continue).

apps/aecore/src/aec_sync.erl
```diff
@@ -188,6 +257,26 @@ terminate(_, _) ->
...
+                  to = To,
+                  uri = Uri} = Sync, Pool) ->
+  {NewUri, NewPool} = 
+      case lists:keyfind(Sync#sync_peer.uri, #sync_peer.uri, Pool) of
+          false ->
+             {true, [Sync | Pool]};
+          RSync when RSync#sync_peer.from > From ->
+             {false, lists:keyreplace(Uri, #sync_peer.uri, Pool, 
+                                      RSync#sync_peer{difficulty = 
+                                                        max(Difficulty, RSync#sync_peer.difficulty),
```
lucafavatella: I see that `difficulty` and `to` get assigned separately by two max operations. I would expect that one of the two shall drive, and the other shall follow.

----

I see difficulty is used for keeping sync pool sorted, I did not find other usages of it. Not sure whether the sort is actually exploited, it looks not.
The to is actually relevant for understanding when to stop syncing.

I see `difficulty` and `to` come originally from same header - i.e. header advertized by peer as "best hash".

apps/aecore/src/aec_sync.erl
```diff
@@ -278,59 +367,129 @@ do_forward_tx(Tx, Uri) ->
...
     end.
 
-is_connected_to_genesis(Hash) ->
-    lager:debug("Checking if ~p is connected to genesis", [pp(Hash)]),
-    aec_chain:hash_is_connected_to_genesis(Hash).
+fetch_chain_from_header(Uri, Height, HeaderHash) ->
+  case do_fetch_block(HeaderHash, Uri) of
+    {ok, true, Block} ->
+      lager:debug("Calling post_block(~p) --> hash of header ~p", [pp(Block), HeaderHash]),
+      case aec_conductor:add_synced_block(Block) of
```
lucafavatella: I see approach is similar to before in only posting blocks to conductor (I do not see calls to `aec_conductor:post_header/1`).

apps/aecore/src/aec_sync.erl
lucafavatella: On a wider note probably out of scope for this PR, in the ping object the difficulty is taken from top block (i.e. with state) while the best hash from top header. Then I am not sure difficulty is used, neither verified as being part of a header. Also, I am not sure whether "top header hash" is guaranteed to be successor of top block (i.e. with verified state) - as would be required by following header chain for syncing though requiring syncing block at each height (from below).

PR #830, opened 2018-03-06
Add docker epoch image builds

dincho:
[PT 155067543](https://www.pivotaltracker.com/story/show/155067543)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,6 @@ references:
     working_directory: ~/epoch
     environment:
       DOCKER_CLIENT_VERSION: "17.09.0-ce"
-      DOCKERHUB_REPO: aetrnty/epoch
 
   setup_remote_docker: &setup_remote_docker
     setup_remote_docker:
@@ -30,7 +29,7 @@ references:
     run:
       name: Build Docker image
       command: |
-        ~/bin/docker build -t $DOCKERHUB_REPO:$DOCKER_TAG .
+        ~/bin/docker build .
 
   merge_checkout: &merge_checkout
     run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -315,6 +350,13 @@ workflows:
             tags:
               only: /^v.*$/
 
+      - docker_image_build:
+          filters:
+            branches:
```
lucafavatella: If slow or we run out of containers, we may run this only on master.

.circleci/config.yml
```diff
@@ -8,6 +8,29 @@ references:
...
+        curl -L -o /tmp/docker-${DOCKER_CLIENT_VERSION:?}.tgz \
+          https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_CLIENT_VERSION:?}.tgz
+        tar -xz -C /tmp -f /tmp/docker-${DOCKER_CLIENT_VERSION:?}.tgz
+        mkdir ~/bin && mv /tmp/docker/docker ~/bin
+
+  docker_build: &docker_build
+    run:
+      name: Build Docker image
+      command: |
+        ~/bin/docker build -t $DOCKERHUB_REPO:$DOCKER_TAG .
```
lucafavatella: Why the need for tagging it at all? No uploading is done so it looks like no tagging needed.

Also, I wonder if any state is persisted between runs so there is clash in tag name? (that BTW is branch name)

.circleci/config.yml
dincho: It's about 11minutes.

.circleci/config.yml
dincho: I'll tag and push in the following pull requests, anyway I removed the obsoleted tagging (in scope of this PR) in 0d1f9069 for the sake of atomicity

.circleci/config.yml
lucafavatella: So maybe worth running it only on master?

.circleci/config.yml
lucafavatella: thanks

.circleci/config.yml
dincho: I don't mind if we setup proper monitoring on master failure and development policy to fix it with highest priority. But let's do that in future, once we agree on policy and have proper monitoring.

PR #832, opened 2018-03-07
Compile time check for running node

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/153479824)
Aiming at better user experience: when a user downloads an updated version of the source and tries running the tests - they will fail if there is an already running `epoch` instance.

This PR makes it a little better, issuing a compile time warning if there is an `epoch` instance running

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,7 +121,7 @@ dialyzer:
 test:
 	$(eval EPOCH_PROCESSES := $(shell ps -fea | grep "bin/epoch" | grep -v grep | wc -l))
 	@if [ $(EPOCH_PROCESSES) -gt 0 ] ; then \
-		echo "An Epoch node is already running";\
+		(echo "An Epoch node is already running, tests are skipped" >&2; exit 1);\
 	else \
 		./rebar3 as test do release, ct $(CT_TEST_FLAGS) --sys_config config/test.config; \
 	fi
--- a/Makefile
+++ b/Makefile
@@ -121,7 +121,7 @@ dialyzer:
 test:
 	$(eval EPOCH_PROCESSES := $(shell ps -fea | grep "bin/epoch" | grep -v grep | wc -l))
 	@if [ $(EPOCH_PROCESSES) -gt 0 ] ; then \
-		(echo "An Epoch node is already running, tests are skipped" >&2; exit 1);\
+		(echo "Failed testing: another Epoch node is already running" >&2; exit 1);\
 	else \
 		./rebar3 as test do release, ct $(CT_TEST_FLAGS) --sys_config config/test.config; \
 	fi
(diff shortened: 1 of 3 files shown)


dincho:
This should be runtime check, not compile time. Generally it's bad practice to mix both.

velzevur:
@dincho it is completly fine to run a couple of nodes on the same machine. You shouldn't see a warning for every next node started.
Since some tests fail because of running nodes - this warning is there to resolve confusion. Since all tests compile the code before the test runs - compile time is fine. You also get the warning if you compile the node while have a node running. Please note that this is a reminder and does not stop the compile process.

I think the compile time check is fine, comments are welcome.

dincho:
@velzevur I got the initial idea. I'm just saying it's conceptually wrong to me. Sounds more like a nasty workaround than anything helpful.

 I don't see anything technically broken in this change (but conceptually), thus I'm not posting a review, but just a 5c comment :)

dincho:
> I understand the main use case is that some test suites do not like if node already running: may I suggest to rather enhance the init of the specific test(s) and make the test fail hard if such precondition not satisfied?

👍 sounds much better to me

velzevur:
As suggested, PR is modified to check for a running node only for the tests that could actually be affected by it.

That would be just `make test` target

Makefile
```diff
@@ -119,7 +119,12 @@ dialyzer:
 	@./rebar3 dialyzer
 
 test:
-	@./rebar3 as test do release, ct $(CT_TEST_FLAGS) --sys_config config/test.config
+	$(eval EPOCH_PROCESSES := $(shell ps -fea | grep "bin/epoch" | grep -v grep | wc -l))
+	@if [ $(EPOCH_PROCESSES) -gt 0 ] ; then \
+		echo "An Epoch node is already running";\
```
lucafavatella: exit code non-zero if this code path reached?
stderr rather than stdout?

velzevur:
[Comment](https://www.pivotaltracker.com/story/show/153479824/comments/186928137) added in the pivotal story

Makefile
```diff
@@ -119,7 +119,12 @@ dialyzer:
 	@./rebar3 dialyzer
 
 test:
-	@./rebar3 as test do release, ct $(CT_TEST_FLAGS) --sys_config config/test.config
+	$(eval EPOCH_PROCESSES := $(shell ps -fea | grep "bin/epoch" | grep -v grep | wc -l))
+	@if [ $(EPOCH_PROCESSES) -gt 0 ] ; then \
+		(echo "An Epoch node is already running, tests are skipped" >&2; exit 1);\
```
ThomasArts: Test are not skipped, but fail, right "..., tests fail". That's why you have exit 1. You want to actually tell that something is wrong.

PR #835, opened 2018-03-07
 Push docker "epoch" image to DockerHub from CircleCI

dincho:
After this change we'll push docker images to DockerHub from CircleCI.
- on master build we push `aetrnty/epoch:dev`
- on tag build we push `aetrnty/epoch:latest` and `aetrnty/epoch:$TAG`
- each PR build is still building the docker image without push

[example build](https://circleci.com/workflow-run/0a266b5e-3985-46ea-946e-5b27b6d77a0e) that pushed the [`aetrnty/epoch:test`](https://hub.docker.com/r/aetrnty/epoch/tags/)

[PT 155366081](https://www.pivotaltracker.com/story/show/155366081)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -376,6 +376,7 @@ workflows:
               ignore:
                 - env/dev1
                 - env/dev2
+                - master
 
       - deploy_api_docs:
           requires:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -297,9 +297,9 @@ jobs:
           name: Build and push Docker image to DockerHub
           command: |
             ~/bin/docker login -u $DOCKER_USER -p $DOCKER_PASS
-            ~/bin/docker build -t ${DOCKERHUB_REPO:?}:${CIRCLE_TAG:?} .
-            ~/bin/docker tag ${DOCKERHUB_REPO:?}:latest ${DOCKERHUB_REPO:?}:${CIRCLE_TAG:?}
+            ~/bin/docker build -t ${DOCKERHUB_REPO:?}:${CIRCLE_TAG:?} -t ${DOCKERHUB_REPO:?}:latest .
             ~/bin/docker push ${DOCKERHUB_REPO:?}:${CIRCLE_TAG:?}
+            ~/bin/docker push ${DOCKERHUB_REPO:?}:latest
 
   docker_push_dev:
     <<: *container_config
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,7 @@ references:
     working_directory: ~/epoch
     environment:
       DOCKER_CLIENT_VERSION: "17.09.0-ce"
+      DOCKERHUB_REPO: aetrnty/epoch
 
   setup_remote_docker: &setup_remote_docker
     setup_remote_docker:
@@ -24,12 +25,6 @@ references:
           https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_CLIENT_VERSION:?}.tgz
         tar -xz -C /tmp -f /tmp/docker-${DOCKER_CLIENT_VERSION:?}.tgz
         mkdir ~/bin && mv /tmp/docker/docker ~/bin
-
-  docker_build: &docker_build
-    run:
-      name: Build Docker image
-      command: |
-        ~/bin/docker build .
 
   merge_checkout: &merge_checkout
     run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -287,7 +282,37 @@ jobs:
...
+      - checkout
+      - *setup_remote_docker
+      - *install_docker_client
+      - run:
+          name: Build and push Docker image to DockerHub
+          command: |
+            ~/bin/docker login -u $DOCKER_USER -p $DOCKER_PASS
+            ~/bin/docker build -t ${DOCKERHUB_REPO:?}:${CIRCLE_TAG:?} .
+            ~/bin/docker tag ${DOCKERHUB_REPO:?}:latest ${DOCKERHUB_REPO:?}:${CIRCLE_TAG:?}
+            ~/bin/docker push ${DOCKERHUB_REPO:?}:dev
```
lucafavatella: Copy-paste error?

dincho: yes, addressed in c8119089

.circleci/config.yml
```diff
@@ -287,7 +282,37 @@ jobs:
...
+      - checkout
+      - *setup_remote_docker
+      - *install_docker_client
+      - run:
+          name: Build and push Docker image to DockerHub
+          command: |
+            ~/bin/docker login -u $DOCKER_USER -p $DOCKER_PASS
+            ~/bin/docker build -t ${DOCKERHUB_REPO:?}:${CIRCLE_TAG:?} .
+            ~/bin/docker tag ${DOCKERHUB_REPO:?}:latest ${DOCKERHUB_REPO:?}:${CIRCLE_TAG:?}
+            ~/bin/docker push ${DOCKERHUB_REPO:?}:${CIRCLE_TAG:?}
```
lucafavatella: I understand `docker push` changes remote state. Would it be possible to print remote state portion meant to be modified by state push before *and* after each push?

`docker images ...` may be the right command though I am not sure if its output discerns between image tagged-though-not-pushed-yet from image tagged-and-pushed.

.circleci/config.yml
```diff
@@ -287,7 +282,37 @@ jobs:
...
+    steps:
+      - checkout
+      - *setup_remote_docker
+      - *install_docker_client
+      - run:
+          name: Build and push Docker image to DockerHub
+          command: |
+            ~/bin/docker login -u $DOCKER_USER -p $DOCKER_PASS
+            ~/bin/docker build -t ${DOCKERHUB_REPO:?}:dev .
+            ~/bin/docker push ${DOCKERHUB_REPO:?}:dev
```
lucafavatella: Same comment as for other push - if possible print before and after push so to see change in remote state.

.circleci/config.yml
lucafavatella: How does the latest tag get pushed? Is a separate second distinct push needed or it gets pushed automagically with the related tag?

dincho: I'm currently investigating this

amir-hadi: I am not sure if you need to push both tags, something like:
`~/bin/docker push ${DOCKERHUB_REPO:?}:${CIRCLE_TAG:?} ${DOCKERHUB_REPO:?}:latest`.

I might be wrong about this.

.circleci/config.yml
amir-hadi: @lucafavatella Why do we need this? To see if the push worked? Docker push prints out the layer hashes, and reports if a layer already exists on the remote.

lucafavatella: @amir-hadi  Can you please quote lines that make this explicit? Looking at the output of the sample build that @dincho posted I could not find it (probably me not being familiar with docker push).

.circleci/config.yml
dincho: I don't find simple or build-in way to do that. As I see we could use the docker registry API, but it requires authentication, thus I find it out of scope of this PR.

.circleci/config.yml
amir-hadi: @lucafavatella I just checked the logs and I coundn't find it. It must be a problem how docker outputs the information (it has a fancy progress bar in newer versions etc).

This is how docker push looks like on my server:
```The push refers to a repository [docker.io/ahadi/znc-docker]
d7c212d822a6: Layer already exists
d05104117d00: Layer already exists
2edfb6fb8973: Layer already exists
6adf46d34d91: Layer already exists
d2c12f0055eb: Layer already exists
607b9bebb156: Layer already exists
00f3d8510106: Layer already exists
e4b57986d764: Layer already exists
a3a69de6a181: Layer already exists
f7127c748181: Layer already exists
48e4f637e500: Layer already exists
08d7a688e98d: Layer already exists
eccdee480abd: Layer already exists
102dfad69cf7: Layer already exists
480a87c93e34: Layer already exists
f8ee36ed8c16: Layer already exists
2f2b8a63dcc6: Layer already exists
8c92a215240e: Layer already exists
725b4ef7ffce: Layer already exists
41ef8cc0bccb: Layer already exists
100396c46221: Layer already exists
7b4b54c74241: Layer already exists
d17d48b2382a: Layer already exists
latest-36efa85: digest: sha256:6cc0e30907c0d5a85f6e826937d5e2fa21122cf14d0f366f49e1381aef228622e0 size: 5148```

File: .circleci/config.yml (hunk 2 above)
Comment (dincho) [reply .circleci/config.yml]: Should be fixed in 0fa66473

**Comment by dincho** [comment]
> I am checking how to improve logging #835 (comment)

Me too

**Comment by dincho** [comment]
@lucafavatella I sent them support ticket. We should not loose more time with CI bugs (IMO).

File: .circleci/config.yml [hunk 4]
```diff
@@ -351,6 +376,7 @@ workflows:
               ignore:
                 - env/dev1
                 - env/dev2
+                - master
```
lucafavatella: Well spotted.

I see that these 3 jobs are mutually exclusive:
- docker_image_build
- docker_push_dev
- docker_push_tag

dincho:
@lucafavatella WRT the output bug this is what i got:

> Hi Dincho,Thank you for the additional details. We've confirmed we are sending the output in the JSON, so it looks like a front-end bug. I've filed a report.I'm not sure what priority it will receive since the team is hard at work getting ready for the 1.0 EOL. As I don't have a timeframe for resolution, I'm going to mark this ticket solved, but it's still linked to the internal bug report and we can reach out when there's an update.

PR #838, opened 2018-03-08
Pt 153463059 endpoints grouping proposal

velzevur:
This PR is part of the story for enabling and disabling of groups of endpoints.

This PR splits the endpoints in groups and provides a setup through epoch.yaml
Groups are formed using swagger's definition for tags.

There are tons of autogenerated source. This is the product of swagger-codegen, producing handlers and api wrappers for every single tag added. This hopefully will be gone soon.


velzevur:
Also there are a lot of compile time warnings (which will be gone with swagger-coden as well...)

velzevur:
Using more than one tag results in swagger router being terribly broken.

This PR must be applied on top the one introducing a proper router.

velzevur:
Rebased on top of proper http router

config/swagger.yaml
```diff
@@ -13,6 +13,22 @@ tags:
...
   - name: internal
     description: Internal API
+  - name: gossip
+    description: Gossip protocol API
+  - name: name_service
+    description: Name resolution API
+  - name: chain 
+    description: Chain state inspection endpoints
+  - name: transactions
+    description: Transactions insection endpoints
```
lucafavatella: type inspection

config/swagger.yaml
```diff
@@ -1163,6 +1219,7 @@ paths:
     get:
       tags:
         - external
+        - debug
```
lucafavatella: I see this is still not dependent on enable_debug_endpoints. Ok.

velzevur: Yes, will be a subject to another PR

velzevur:
What is yet missing:
1. A system level tests that verify that nodes with various setups of the `epoch.yaml` are operational and can communicate as expected.
2. Debug enpoints - currently they're enabled/disabled by `application:get_env(aehttp, enable_debug_endpoints)`

apps/aehttp/src/aehttp_api_router.erl
```diff
@@ -37,7 +38,9 @@ path(Path0) ->
 method(Atom) ->
     list_to_binary(string:uppercase(atom_to_list(Atom))).
 
-is_enabled(Target, Tags) when is_atom(Target) ->
-    is_enabled(atom_to_binary(Target, utf8), Tags);
-is_enabled(Target, Tags) ->
-    lists:member(Target, Tags).
+is_enabled(Target, Tags, EnabledGroups) when is_atom(Target) ->
+    is_enabled(atom_to_binary(Target, utf8), Tags, EnabledGroups);
```
lucafavatella: Not sure why two clauses needed.

apps/aehttp/src/aehttp_api_router.erl
```diff
@@ -37,7 +38,9 @@ path(Path0) ->
...
+is_enabled(Target, Tags, EnabledGroups) ->
+    lists:member(Target, Tags) andalso
+    lists:any(fun(Tag) -> lists:member(Tag, EnabledGroups) end, Tags).
```
lucafavatella: So if multiple tags then if one enabled enables all? This shall be (confirmed and) documented in the description in the schema.

config/swagger.yaml
lucafavatella: I am not sure `/version` shall be put in same bucket `debug` as `/balances` because much lighter. Though can be refined later?

apps/aehttp/src/aehttp_api_router.erl
velzevur: This is on an endpoint level. If a certain endpoint has a couple of tags and if any of them is enabled - this specific endpoint is enabled.

PR #839, opened 2018-03-08
make 'epoch.log' level configurable

uwiger:
Since we don't use all lager log levels, I added the options "debug", "info", "warning", "error" and "none".

Due to start sequence non-determinism, some logging output may occur before the new level is set.

### DIFF
--- a/apps/aecore/src/aecore_app.erl
+++ b/apps/aecore/src/aecore_app.erl
@@ -85,6 +85,7 @@ set_level(L) when is_binary(L) ->
     case lists:member(Level, levels()) of
         true ->
             lager_set_level_env(Level),
+            adjust_sinks(Level),
             if_running(lager, fun() -> live_set_level(Level) end);
         false ->
             lager:error("Unknown log level: ~p", [Level]),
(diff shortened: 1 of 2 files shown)


dincho:
Can we have this also working for `epoch_mining.log` and `epoch_pow_cuckoo.log` as well (this is what we use currently)?

apps/aecore/src/aecore_app.erl
```diff
@@ -8,7 +8,7 @@
          prep_stop/1,
          stop/1]).
 -export([check_env/0]).
-
+-export([set_level/1]).
```
uwiger: A note: this was actually only exported for manual testing, but could possibly be useful otherwise. I had intended to remove the export.

uwiger:
@dincho the problem with `epoch_mining` and `epoch_pow_cuckoo` is that they have different log levels today (`info` vs `debug`), and it wasn't obvious to me what the semantics should be.

uwiger:
@happi this was also my thought. Perhaps we can consider that as a future improvement?

lucafavatella:
In scope of https://www.pivotaltracker.com/n/projects/2124891/stories/154838044

PR #848, opened 2018-03-12
Configure epoch log level per environment

dincho:
- UAT: warning
- tester: warning
- integration: debug
- dev1, dev2: debug

Tested on [dev1](https://circleci.com/gh/aeternity/epoch/3701), feel free to login to poke around.

[PT 154837975](https://www.pivotaltracker.com/story/show/154837975)

### DIFF
--- a/deployment/ansible/group_vars/integration/config.yml
+++ b/deployment/ansible/group_vars/integration/config.yml
@@ -1,0 +1,1 @@
+log_level: debug
--- a/deployment/ansible/deploy.yml
+++ b/deployment/ansible/deploy.yml
@@ -44,6 +44,8 @@
       keypair:
         dir: "keys"
         password: "{{ vault_keys_password|default('secret') }}"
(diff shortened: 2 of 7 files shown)


deployment/ansible/group_vars/integration/config.yml
```diff
@@ -0,0 +1 @@
+log_level: debug
```
lucafavatella: I suggest warning here too - as UAT - in order not to experience only in UAT that the warning+ logs are not enough.
dev1 and dev2 ok debug.

dincho: It sounds better to me that integration is in debug mode and UAT is not. While the integration contains unstable software, the UAT should contain stable releases only and debug shouldn't be needed except in extreme situations.

lucafavatella: ok

PR #852, opened 2018-03-13
Index height of headers to make lookup by height faster

gorillainduction:
Remove the prev hash index to save one index.

The successor function now use the height index instead.

https://www.pivotaltracker.com/n/projects/2124891/stories/155893232

### DIFF
--- a/apps/aecore/src/aec_db.erl
+++ b/apps/aecore/src/aec_db.erl
@@ -26,7 +26,7 @@
          find_block/1,
          find_header/1,
          find_chain_node/1,
-         find_chain_node_successors/1,
+         find_chain_nodes_at_height/1,
          get_block/1,
          get_header/1,
          get_genesis_hash/0,
@@ -80,7 +80,7 @@
 %% - one per state tree
 
 -record(aec_blocks             , {key, txs}).
--record(aec_headers            , {key, value, has_block = false, prev_hash}).
+-record(aec_headers            , {key, value, has_block = false, height}).
 -record(aec_contract_state     , {key, value}).
 -record(aec_tx                 , {key, tx}).
 -record(aec_chain_state        , {key, value}).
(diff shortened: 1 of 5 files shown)


apps/aecore/src/aec_db.erl
```diff
@@ -167,14 +167,14 @@ write_header(Header) ->
...
-    PrevHash = aec_headers:prev_hash(Header),
+    Height = aec_headers:height(Header),
     ?t(case mnesia:read(aec_headers, Hash) of
            [#aec_headers{value = Header, has_block = OldHasBlock} = H] ->
                NewHasBlock = (HasBlock or OldHasBlock),
                mnesia:write(H#aec_headers{has_block = NewHasBlock});
            [#aec_headers{value = Old}] -> error({header_exist, Header, Old});
            [] -> mnesia:write(#aec_headers{key = Hash, value = Header,
-                                           prev_hash = PrevHash,
+                                           height = Height,
```
lucafavatella: It looks like this change in DB schema is so small that a new node may manage to start with an old DB. We [document to the user not to reuse persisted blockchain ("Please ensure that you do not reuse a persisted blockchain produced by the previous releases "v0.8.x".")](https://github.com/aeternity/epoch/blob/4795b2c820e1f44c23a92cb62ef5c85ec47cbd9e/docs/release-notes/RELEASE-NOTES-0.9.0.md#about-this-release) so it is fine. (We did not [teach node to detect incompatible DB schema](https://www.pivotaltracker.com/story/show/155656073)).

gorillainduction: The schema is altered because of the new field. Will Mnesia really start up with the persisted DB?

But, I guess the node will not shut down as it should.

lucafavatella: > Will Mnesia really start up with the persisted DB?

I did not test.

PR #857, opened 2018-03-14
Pt 155428313 optimize retrieving blocks

ThomasArts:
We now fetch blocks from below and in parallel

```erlang
%% When we Ping a node with at least as much difficulty as we have, 
%% then we are going to sync with it.
%% We already agree upon the genesis block and need to find the highest common
%% block we agree upon. We use a binary search to find out.
%% From the height we agree upon, we start asking for blocks to add to the chain.
%% 
%% When an additional Ping arrives for which we agree upon genesis, we have
%% the following possibilities:
%% 1. It has worst top hash than our node, do not include in sync
%% 2. It has better top hash than our node
%%    We binary search for a node that we agree upon (could be genesis) 
%% We add new node to sync pool to sync agreed node upto top of new
%% 1. If we are already synchronizing, we ignore it,
%%    the ongoing sync will pick that up later
%% 2. If we are not already synchronizing, we start doing so.
%% 
%% We sync with several nodes at the same time and use as strategy 
%% to pick a random hash from the hashes in the pool.
```

### DIFF
--- a/apps/aecore/src/aec_headers.erl
+++ b/apps/aecore/src/aec_headers.erl
@@ -72,7 +72,7 @@ serialize_to_map(H = #header{}) ->
     {ok, Serialized}.
 
 
--spec deserialize_from_map(map()) -> {ok, header()}.
+-spec deserialize_from_map(map()) -> {ok, header()} | {error, deserialize}.
 deserialize_from_map(H = #{}) ->
       #{<<"height">> := Height,
         <<"prev_hash">> := PrevHash,
--- a/apps/aecore/src/aec_sync.erl
+++ b/apps/aecore/src/aec_sync.erl
@@ -37,6 +37,9 @@
 -compile([export_all, nowarn_export_all]).
 -endif.
 
+-define(MAX_HEADERS_PER_CHUNK, 20).
+-define(MAX_DIFF_FOR_SYNC, 50).
+
 -record(sync_peer, {difficulty, from, to, hash, uri}).
 
 %%%=============================================================================
@@ -128,8 +131,8 @@ fetch_mempool(Uri) ->
 schedule_ping(Uri) ->
     gen_server:cast(?MODULE, {schedule_ping, Uri}).
 
-fetch_next(Uri, HeightIn, HashIn) ->
-    gen_server:call(?MODULE, {fetch_chain, Uri, HeightIn, HashIn}).
+fetch_next(Uri, HeightIn, HashIn, Result) ->
+    gen_server:call(?MODULE, {fetch_next, Uri, HeightIn, HashIn, Result}).
 
 delete_from_pool(Uri) ->
     gen_server:cast(?MODULE, {delete_from_pool, Uri}).
@@ -137,6 +140,12 @@ delete_from_pool(Uri) ->
 %% Check whether this header is better than previous best
 new_header(Uri, Header, AgreedHeight, Hash) ->
     gen_server:call(?MODULE, {new_header, Uri, Header, AgreedHeight, Hash}).
+
+update_hash_pool(Hashes) ->
+  gen_server:call(?MODULE, {update_hash_pool, Hashes}).
+
+sync_in_progress(Uri) ->
+   gen_server:call(?MODULE, {sync_in_progress, Uri}).
 
 %%%=============================================================================
 %%% gen_server functions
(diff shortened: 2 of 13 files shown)


apps/aecore/src/aec_headers.erl
```diff
@@ -72,7 +72,7 @@ serialize_to_map(H = #header{}) ->
     {ok, Serialized}.
 
 
--spec deserialize_from_map(map()) -> {ok, header()}.
```
lucafavatella: Well spotted.

config/swagger.yaml
```diff
@@ -112,6 +112,71 @@ paths:
...
+          description: The headers found
+          schema:
+            type: array 
+            items:
+              $ref: '#/definitions/Header'
+        '400':
+          description: Invalid hash
+          schema:
+            $ref: '#/definitions/Error'
+        '401':
```
lucafavatella: Shall this rather be included in 400? 401 seems auth specific https://tools.ietf.org/html/rfc7235#section-3.1
So 400 would become with description "Invalid hash or number"

lucafavatella: >  So 400 would become with description "Invalid hash or number"
I mean description in the schema. Of course error text can stay distinct in the code.

apps/aecore/src/aec_sync.erl
```diff
@@ -294,7 +406,10 @@ process_job([{_T, Job}]) ->
         {forward, #{tx := Tx}, Uri} ->
             do_forward_tx(Tx, Uri);
         {start_sync, Uri, RemoteHash} ->
-            do_start_sync(Uri, RemoteHash);
+            case sync_in_progress(Uri) of
```
lucafavatella: I notice this as first improvement: it prevents spending effort on finding agreed height if sync for this URI already in progress (same check is done in do_start_sync function - see new_header call - after height agreed).

No change required for this in this PR, though note at https://github.com/aeternity/epoch/pull/829#discussion_r172653557 still applies.

apps/aecore/src/aec_sync.erl
```diff
@@ -431,64 +555,65 @@ agree_on_height(Uri, _, RH, LHeader, LH, Max, Min) when RH =/= LH ->
...
           lager:debug("Header fetched (~p): ~p", [Uri, pp(Header)]),
-          case aec_headers:prev_hash(Header) == HashFromH of
-            true ->
-              {ok, HeaderHash} = aec_headers:hash_header(Header),
-              fetch_chain_from_header(Uri, FromH + 1, HeaderHash); 
-            false ->
-              lager:info("Abort sync due to non-fitting header ~p =/= ", [Header]),
-              delete_from_pool(Uri),
-              {error, sync_abort}
+          case Chunk of
```
lucafavatella: Reading it does not seem obvious that this quantity is greater or equal to 1.

apps/aeutils/src/aeu_requests.erl
```diff
@@ -142,29 +143,57 @@ get_header_by_hash(Uri, Hash) ->
...
+            %% Should have been turned to {error, _} by swagger validation
+            lager:debug("unexpected response (~p): ~p", [Uri, Response]),
+            {error, unexpected_response}
+    end.
 
 %% Add API for header later... now use block
 -spec get_header_by_height(http_uri_uri(), non_neg_integer()) -> response(aec_headers:header()).
 get_header_by_height(Uri, Height) when is_integer(Height) ->
-    Response = process_request(Uri, 'GetBlockByHeight', [{"height", integer_to_list(Height)}]),
+    Response = process_request(Uri, 'GetHeaderByHeight', [{"height", integer_to_list(Height)}]),
```
lucafavatella: I see this means that this node cannot sync from node running v0.8.0 node. Could you please propose a line in the [releases notes](https://github.com/aeternity/epoch/blob/32722bae6d04dfcaf4b50d8e19a2eca5fdbded78/docs/release-notes/RELEASE-NOTES-0.9.0.md#about-this-release)?

lucafavatella: Anyway the consensus changed, though a line may be beneficial.

apps/aecore/src/aec_sync.erl
```diff
@@ -431,64 +555,65 @@ agree_on_height(Uri, _, RH, LHeader, LH, Max, Min) when RH =/= LH ->
...
-    aec_events:publish(chain_sync, {client_done, Uri});
-fetch_chain(Uri, FromH, ToH, HashFromH) when FromH < ToH ->
-    case aeu_requests:get_header_by_height(Uri, FromH + 1) of
-        {ok, Header} ->
+fill_pool(Uri, RemoteHeight, AgreedHeight, AgreedHash) ->
+    %% This should change to "get next ?MAX_HEADERS_PER_CHUNK blocks
+    %% starting from provided hash.
+    Chunk = min(RemoteHeight - AgreedHeight, ?MAX_HEADERS_PER_CHUNK),
+    case aeu_requests:get_block_by_height(Uri, AgreedHeight + Chunk) of
+        {ok, Block} ->
```
lucafavatella: Who checks that the height in the received block is the same height as requested?

lucafavatella: Who checks that the prev_hash in the received block makes sense for progressing towards the sync goal, and is not a random pruned/orphan block at that height?

config/swagger.yaml
ThomasArts: It doesn't really matter, does it. We are going to remove the http interface or are going to go over the swagger definitions, which are a bit broken anyway.

apps/aecore/src/aec_sync.erl
ThomasArts: Yep, there are one or two calls at the borders that are duplicated, but that's because the chain may have just forked at the top.

apps/aecore/src/aec_sync.erl
ThomasArts: That's part of the update_from_pool.

ThomasArts: L375 update_chain_from pool does that.

PR #859, opened 2018-03-14
[After Upgrade] Add function to get at most n headers forward from a hash in the main chain

gorillainduction:
https://www.pivotaltracker.com/n/projects/2124891/stories/155893290

### DIFF
--- a/apps/aecore/src/aec_chain.erl
+++ b/apps/aecore/src/aec_chain.erl
@@ -19,7 +19,8 @@
         , get_header/1
         , get_header_by_height/1
         , get_missing_block_hashes/0
-        , get_n_headers_from_hash/2
+        , get_n_headers_backwards_from_hash/2
+        , get_at_most_n_headers_forward_from_hash/2
         , get_top_N_blocks_time_summary/1
         , get_transactions_between/2
         , has_block/1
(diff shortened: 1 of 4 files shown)


lucafavatella:
I see https://www.pivotaltracker.com/n/projects/2124891/stories/155893290 labelled as next_release

apps/aecore/src/aec_chain.erl
```diff
@@ -255,19 +256,53 @@ find_common_ancestor(Hash1, Hash2) when is_binary(Hash1), is_binary(Hash2) ->
...
+get_at_most_n_headers_forward_from_hash(Hash, N) when is_binary(Hash), is_integer(N), N > 0 ->
+    case top_header() of
+        undefined -> error;
+        TopHeader ->
+            TopHeight = aec_headers:height(TopHeader),
+            case get_header(Hash) of
+                error -> error;
+                {ok, Header} ->
+                    Height = aec_headers:height(Header),
+                    Delta = min(TopHeight - Height, N - 1),
```
ThomasArts: I guess Delta could indeed be negative if there is a fork that is higher but has less difficulty

gorillainduction: Yes, exactly.

PR #864, opened 2018-03-15
Fix GetTx when a transaction exists in multiple places

gorillainduction:
A transaction can be in the mempool or in multiple forks at the same
time. GetTx can now handle this case.

### DIFF
--- a/apps/aecore/src/aec_chain.erl
+++ b/apps/aecore/src/aec_chain.erl
@@ -8,6 +8,7 @@
 
 %%% Chain API
 -export([ find_common_ancestor/2
+        , find_transaction_in_main_chain_or_mempool/1
         , genesis_block/0
         , genesis_hash/0
         , genesis_header/0
@@ -26,6 +27,7 @@
         , has_block/1
         , has_header/1
         , hash_is_connected_to_genesis/1
+        , hash_is_in_main_chain/1
         , top_block/0
         , top_block_hash/0
         , top_block_header/0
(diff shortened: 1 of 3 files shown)


apps/aecore/src/aec_chain.erl
```diff
@@ -251,6 +252,16 @@ genesis_header() ->
 find_common_ancestor(Hash1, Hash2) when is_binary(Hash1), is_binary(Hash2) ->
     aec_chain_state:find_common_ancestor(Hash1, Hash2).
 
+
+-spec hash_is_in_main_chain(binary()) -> boolean().
+hash_is_in_main_chain(Hash) when is_binary(Hash) ->
+    case get_header(Hash) of
+        none -> false;
```
lucafavatella: error rather than none

apps/aecore/src/aec_chain.erl
```diff
@@ -251,6 +252,16 @@ genesis_header() ->
...
+        {value, Header} ->
```
lucafavatella: and {ok, Header}

apps/aecore/src/aec_chain.erl
```diff
@@ -252,6 +273,16 @@ genesis_header() ->
...
     aec_chain_state:find_common_ancestor(Hash1, Hash2).
 
+
+-spec hash_is_in_main_chain(binary()) -> boolean().
+hash_is_in_main_chain(Hash) when is_binary(Hash) ->
+    case get_header(Hash) of
+        error -> false;
+        {ok, Header} ->
+            Height = aec_headers:height(Header),
+            {ok, Hash} =:= aec_chain_state:get_hash_at_height(Height)
```
lucafavatella: Fragile to return type change (`{ok, H}` vs. `H`) though a debatable cosmetic matter.

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -859,6 +861,81 @@ fork_get_by_height() ->
...
+    ?assertEqual(none,
+                 aec_chain:find_transaction_in_main_chain_or_mempool(HardTxHash)),
+
+    ok = write_blocks_to_chain(HardChain),
+    ?assertEqual({block_hash(HardButLastBlock), HardButLastCoinbase},
+                 aec_chain:find_transaction_in_main_chain_or_mempool(HardButLastTxHash)),
+    ?assertEqual({block_hash(HardTopBlock), HardCoinbase},
+                 aec_chain:find_transaction_in_main_chain_or_mempool(HardTxHash)),
+
+    %% Now, mock that the transaction is in the mempool as well.
```
lucafavatella: I see assumes coinbase potentially in mempool, that is ok for sake of testing.

PR #866, opened 2018-03-16
Pt 156003774 get successor hashes when fetching a long chain

ThomasArts:
With new function we can now get the headers on top of a certain hash. We use this instead of re-implementing this functionality.

Experiment on 500 blocks chain:
At the moment we can fetch 10 blocks per second when talking to 2 nodes to fetch from, more nodes, quicker fetching. We only fetch what we need, no duplications.

### DIFF
--- a/apps/aeutils/src/aeu_requests.erl
+++ b/apps/aeutils/src/aeu_requests.erl
@@ -37,7 +37,7 @@
 -export([ping/2,
          top/1,
          get_header_by_hash/2,
-         get_n_hashes/3,
+         get_n_successors/3,
          get_header_by_height/2,
          get_block_by_height/2,
          get_block/2,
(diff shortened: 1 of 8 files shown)


apps/aeutils/src/aeu_requests.erl
```diff
@@ -143,10 +143,13 @@ get_header_by_hash(Uri, Hash) ->
...
-get_n_hashes(Uri, Hash, N) when is_integer(N) ->
+%% Get the next n hashes and their heights
+-spec get_n_successors(http_uri_uri(),  binary(), non_neg_integer()) -> response([{integer(), binary()}]).
+get_n_successors(Uri, Hash, N) when is_integer(N) ->
     EncHash = aec_base58c:encode(block_hash, Hash),
-    Response = process_request(Uri, 'GetHeadersByHash', [{"hash", EncHash}, {"number", integer_to_list(N)}]),
+    Response = process_request(Uri, 'GetHeadersByHash', [{"hash", EncHash}, 
+                                                         {"number", integer_to_list(N+1)}, 
+                                                         {"direction", forward}]),
     case Response of
```
gorillainduction: To follow the spec (and please Dialyzer) you should change:
`{"direction", forward}` -> `{"direction", "forward"}`

ThomasArts: Thanks for pointing this out... took me also a while to find this

PR #868, opened 2018-03-16
Pt 156042138 cleanup calls to aec db

gorillainduction:
https://www.pivotaltracker.com/n/projects/2124891/stories/156042138

This is also fixing a similar problem as https://www.pivotaltracker.com/n/projects/2124891/stories/156009494

### DIFF
--- a/apps/aehttp/src/aehttp_dispatch_int.erl
+++ b/apps/aehttp/src/aehttp_dispatch_int.erl
@@ -783,17 +783,10 @@ get_block_range(GetFun, Req) when is_function(GetFun, 0) ->
     end.
 
 get_account_balance_at_hash(AccountPubkey, Hash) ->
-    case aec_db:find_block_state(Hash) of
-        none ->
-            {error, not_on_main_chain};
-        {value, Trees} ->
-            AccountsMPTree = aec_trees:accounts(Trees),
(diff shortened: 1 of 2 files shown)


apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -908,19 +901,15 @@ get_txs_and_headers(KeepTxTypes, DropTxTypes, Account) ->
...
-                    [{BlockHash, _}] ->
-                        {ok, H} = aec_chain:get_header(BlockHash),
-                        H
-                end,
-            {Header, SignedTx}
-        end,
-        Txs).
+    TxHashes = lists:usort([aetx:hash(aetx_sign:tx(SignedTx))
+                            || SignedTx <- Txs]),
+    [case aec_chain:find_transaction_in_main_chain_or_mempool(TxHash) of
```
uwiger: Shouldn't there be an outer transaction wrapping the `get_txs_and_headers/3` work?

gorillainduction: Agreed. It makes sense to get a proper snapshot and to avoid multiple transactions. The risk of forking in the middle of this call is relatively small, but the cost of that many transactions are probably not negligible.

PR #873, opened 2018-03-19
Pt 154867258 implement call aevm

happi:
[Finishes #154867258]

### DIFF
--- a/apps/aevm/src/aevm_eeevm.erl
+++ b/apps/aevm/src/aevm_eeevm.erl
@@ -495,17 +495,26 @@ loop(StateIn) ->
 		    State5 = aevm_eeevm_memory:write_area(Us1, CodeArea, State4),
 		    next_instruction(OP, State, State5);
 		?RETURNDATASIZE ->
-		    %% 0x3d RETURNDATASIZE
-		    %% Not in yellow paper
-		    error({opcode_not_implemented,
-			   lists:flatten(
-			     io_lib:format("~2.16.0B",[OP]))});
+		    %% 0x3d RETURNDATASIZE  δ=0 α=1
+		    %% Get size of output data from the previous call from the current
+		    %% environment.
+		    %% µ's[0] ≡ |µo|
+		    Val = byte_size(aevm_eeevm_state:return_data(State0)),
+		    State1 = push(Val, State0),
+		    next_instruction(OP, State, State1);
 		?RETURNDATACOPY ->
-		    %% 0x3e RETURNDATACOPY
-		    %% Not in yellow paper
-		    error({opcode_not_implemented,
-			   lists:flatten(
-			     io_lib:format("~2.16.0B",[OP]))});
+		    %% 0x3e RETURNDATACOPY δ=3 α=0
+		    %% Copy output data from the previous call to memory.
+		    %% ∀i∈{0...µs[2]−1}µ'm[µs[0] + i] ≡  µo[µs[1] + i] if µs[1] + i < |µo|
+		    %%                                   0                otherwise
+		    %% The additions in µs[1] + i are not subject to the 2^256 modulo.
+		    %% µ'i ≡ M(µi, µs[0], µs[2])
+		    {Us0, State1} = pop(State0), %% memOffset
+		    {Us1, State2} = pop(State1), %% dataOffset
+		    {Us2, State3} = pop(State2), %% length
+		    ReturnData = return_data_get_bytes(Us1, Us2, State3),
+		    State4 = aevm_eeevm_memory:write_area(Us0, ReturnData, State3),
+		    next_instruction(OP, State, State4);
 		%% No opcode 0x3f
 		16#3f -> throw({illegal_instruction, OP, State0});
 		%% 40s Block Information
@@ -1300,6 +1309,14 @@ data_get_val(Address, State) ->
 %% Get a binary of size Size bytes from input data.
 data_get_bytes(Address, Size, State) ->
     Data = aevm_eeevm_state:data(State),
+    try aevm_eeevm_utils:bin_copy(Address, Size, Data)
+    catch error:system_limit ->
+	    throw({out_of_memory, State})
+    end.
+
+%% Get a binary of size Size bytes from return data.
+return_data_get_bytes(Address, Size, State) ->
+    Data = aevm_eeevm_state:return_data(State),
     try aevm_eeevm_utils:bin_copy(Address, Size, Data)
     catch error:system_limit ->
 	    throw({out_of_memory, State})
--- a/apps/aevm/src/aevm_eeevm.erl
+++ b/apps/aevm/src/aevm_eeevm.erl
@@ -29,7 +29,30 @@
 %%
 %%
 eval(State) ->
-    loop(valid_jumpdests(State)).
+    NewState = loop(valid_jumpdests(State)),
+    Metaop = check_eval_state(NewState),
+    case Metaop of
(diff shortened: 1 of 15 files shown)


apps/aevm/src/aevm_eeevm.erl
```diff
@@ -1343,6 +1316,16 @@ inc_cp(Amount, State) ->
 %% GAS
 %% ------------------------------------------------------------------------
 
+spend_call_gas(State, OP) when OP =:= ?CALL;
```
gorillainduction: This should probably include `?DELEGATE_CALL`

gorillainduction: If this is indeed the cost for `?DELEGATECALL` that is.

apps/aevm/src/aevm_eeevm.erl
```diff
@@ -1424,3 +1408,92 @@ log_topics(AccountAddress, Topics) ->
...
+                 ?CALL -> aevm_eeevm_state:address(State8);
+                 ?CALLCODE -> aevm_eeevm_state:address(State8);
+                 ?DELEGATECALL -> aevm_eeevm_state:caller(State8)
+             end,
+    CallState = aevm_eeevm_state:prepare_for_call(Caller, Dest, CallGas, Value,
+                                                  Code, CallDepth,
+                                                  State8),
+    case aevm_eeevm_state:no_recursion(State8) of
+        true  -> %% TODO: Set callcreates
+            GasOut = GasAfterSpend + CallGas,
```
UlfNorell: `GasAfterSpend - CallGas`?

apps/aevm/src/aevm_eeevm.erl
```diff
@@ -1424,3 +1408,92 @@ log_topics(AccountAddress, Topics) ->
...
+            State9 = aevm_eeevm_state:set_gas(GasOut, State8),
+            State10 = aevm_eeevm_state:add_callcreates(#{ data => I
+                                                        , destination => Dest
+                                                        , gasLimit => CallGas
+                                                        , value => Value
+                                                        }, State9),
+            {1, State10};
```
UlfNorell: I don't quite understand this case. What is the purpose of the `no_recursion` flag? If we're not performing the call (due to `no_recursion`) the return memory will contain garbage. At the very least I would expect the return value to be `0`.

apps/aevm/src/aevm_eeevm.erl
```diff
@@ -1424,3 +1408,92 @@ log_topics(AccountAddress, Topics) ->
...
+            CallState1 = aevm_eeevm_memory:write_area(0, I, CallState),
+            {OutGas, OutState, R} =
+                case eval(CallState1) of
+                    {ok, OutState2} -> {aevm_eeevm_state:gas(OutState2), OutState2, 1};
+                    {error, out_of_gas,_RState2} -> {0, CallState1, 1}
+                end,
+            CallTrace = aevm_eeevm_state:trace(OutState),
+            %% Go back to the caller state.
+            ReturnState1 = aevm_eeevm_state:add_trace(CallTrace, State8),
+            GasAfterCall = GasAfterSpend + OutGas,
```
UlfNorell: `GasAfterSpend + OutGas - CallGas`?

apps/aevm/src/aevm_eeevm.erl
gorillainduction: As I remember it, the no_recursion flag is for testing. The add_callcreates is adding some test data that is verified in the test framework.

UlfNorell: Is this something we need? It would make more sense to use a different implementation of the chain API if we don't want to perform any calls but still make sure that they happen. That way we can also test contracts that actually look at the return value of a call.

gorillainduction: It is just the way the imported test suite works. I guess they are aiming at testing the setup of calls.

I agree that there probably are better ways of testing it.

apps/aevm/src/aevm_eeevm.erl
```diff
@@ -1424,3 +1408,92 @@ log_topics(AccountAddress, Topics) ->
...
+            State9 = aevm_eeevm_state:set_gas(GasOut, State8),
+            State10 = aevm_eeevm_state:add_callcreates(#{ data => I
+                                                        , destination => Dest
+                                                        , gasLimit => CallGas
+                                                        , value => Value
+                                                        }, State9),
+            {1, State10};
+        false ->
+            %% TODO: Should this not count towards mem gas?
+            CallState1 = aevm_eeevm_memory:write_area(0, I, CallState),
```
UlfNorell: Should we not put the calldata in the `data` field in the state and then have the called contract load it into memory using `CALLDATALOAD`/`CALLDATACOPY`?

apps/aevm/src/aevm_eeevm.erl
happi: No. GasAfterSpend is how much gas you would have if all the gas you give to the caller is used up. So we add back the unused amount after the call.

apps/aevm/src/aevm_eeevm.erl
happi: No. GasAfterSpend is how much gas you would have if all the gas you give to the caller is used up. So we add back the unused amount after the call.

apps/aevm/src/aevm_eeevm.erl
happi: Yes. (There was a TODO comment for it higher up...)
Fixed now.

apps/aevm/src/aevm_eeevm.erl
UlfNorell: `GasAfterSpend` is computed before `CallGas`, so I don't see how that could be. Also, if that is really the case then the check `GasAfterSpend >= CallGas` makes no sense at all.

As far as I can tell `GasAfterSpend` is the amount of gas you have after paying for the call instruction, computed by `spend_call_gas`.

happi: It is not so easy to see...
spend_call_gas includes the dynamic cost of the call, that is also the amount of gas you transfer to the call (i.e. the gas argument to the instruction). See aevm_gas.erl line 80 and forward.

happi: The check look suspicious though.

apps/aevm/src/aevm_eeevm.erl
happi: The documentation for the vmTests states that:
  "Since these tests are meant only as a basic test of VM operation, the ``CALL`` and 
``CREATE`` instructions are not actually executed. To provide the possibility of 
testing to guarantee they were actually run at all, a separate portion ``callcreates`` 
details each ``CALL`` or ``CREATE`` operation in the order they would have been executed. 
Furthermore, gas required is simply that of the VM execution: the gas cost for 
transaction processing is excluded."

But that turns out to not be completely true. Some of the vmTests actually performs the call.
Possibly one can determine this on whether the callcreats is [] or not in the spec.
I fixed the test suite so that it sends no_recursion = false for the tests that I have discovered that does the calls.

apps/aevm/src/aevm_eeevm.erl
happi: Rewritten to check for 0 and to not give the call stipend back.
Things should be closer to right now but this is just the beginning.
When we can hook in the chain for real we can start doing some real tests, and fix things like handling calls to new contracts.

PR #876, opened 2018-03-20
Abort node start if an unexpected persisted genesis block is found

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/155656097)
While starting a node, if persistence is `on` and node loads an unexpected genesis block from the persisted chain - the start is aborted and there is a proper message in the logs.

### DIFF
--- a/apps/aecore/src/aecore_app.erl
+++ b/apps/aecore/src/aecore_app.erl
@@ -17,7 +17,14 @@ start(_StartType, _StartArgs) ->
     ok = lager:info("Starting aecore node"),
     ok = application:ensure_started(mnesia),
     aec_db:load_database(),
-    aecore_sup:start_link().
+    case aec_db:persisted_valid_genesis_block() of
+        true ->
+            aecore_sup:start_link();
+        false ->
+            lager:error("Persisted chain has a different genesis block than "
+                        ++ "the one being expected. Aborting", []),
+            {error, inconsistent_database}
+    end.
 
 start_phase(create_metrics_probes, _StartType, _PhaseArgs) ->
     lager:debug("start_phase(create_metrics_probes, _, _)", []),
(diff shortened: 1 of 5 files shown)


apps/aecore/src/aecore_app.erl
```diff
@@ -17,7 +17,14 @@ start(_StartType, _StartArgs) ->
     ok = lager:info("Starting aecore node"),
     ok = application:ensure_started(mnesia),
     aec_db:load_database(),
-    aecore_sup:start_link().
+    case aec_db:persisted_valid_genesis_block() of
+        true ->
+            aecore_sup:start_link();
+        false ->
+            lager:error("Persisted chain has a different genesis block than "
```
lucafavatella: Printing the two hashes - expected and actual - may help user troubleshoot?

velzevur:
Added a log for an easier debugging when the situation with the inconsistent genesis blocks occurs

velzevur:
Added a corresponding [section](https://github.com/aeternity/epoch/wiki/Troubleshooting#node-wont-start-after-release-update) in the troubleshooting wiki page.
CC @lucafavatella

lucafavatella:
> Added a corresponding section in the troubleshooting wiki page.

> This is a symptom of running the node with an old persisted DB while there had been a non-backwards compatible change in the protocol.

I suggest "This is a symptom of running the node with an old persisted DB while there had been a non-backwards compatible change in the protocol *including the genesis block*." as not all protocol changes necessarily include a different genesis block. @velzevur How does that read?

lucafavatella:
I did not manage to test locally as I am in the middle of some other testing.

velzevur:
@lucafavatella Updated the wiki page with clarification concerning the symptoms at hand

PR #877, opened 2018-03-20
Enable unconditional xref runs in SA job

dincho:
https://www.pivotaltracker.com/story/show/153122308

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -222,6 +222,7 @@ jobs:
       - run: make dialyzer
       - run: make swagger-check
       - run: make rebar-lock-check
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -222,6 +222,7 @@ jobs:
       - run: make dialyzer
       - run: make swagger-check
       - run: make rebar-lock-check
+      - run: ./rebar3 xref || true
```
lucafavatella: Make target would be preferable.

dincho: To be honest I really don't like the current state (abuse) of the Makefile.
Make is build tool, not task runner.

PR #878, opened 2018-03-20
Stop tracking python tests swagger client

dincho:
https://www.pivotaltracker.com/story/show/155182731


fabiankrol:
> +3 −26,493

😍

PR #880, opened 2018-03-21
Empty accounts.json processing

velzevur:
Addressing some corner cases:

* Empty `accounts.json`

* `accounts.json` containing an empty js object: JSX has some curious behaviour with it ( `[{}] = jsx:decode(<<"{}">>).`) and it was causing crashes

Added EUnit tests to cover the functionality of the module

### DIFF
--- a/apps/aecore/src/aec_genesis_block_settings.erl
+++ b/apps/aecore/src/aec_genesis_block_settings.erl
@@ -1,26 +1,44 @@
 -module(aec_genesis_block_settings).
 
 -export([dir/0,
-         preset_accounts/0]).
+         preset_accounts/0,
+         read_presets/0]).
 
 dir() ->
     filename:join(aeu_env:data_dir(aecore), ".genesis").
 
+-spec preset_accounts() -> list().
 preset_accounts() ->
-    PresetAccountsFile = filename:join([dir(), "accounts.json"]),
-    case file:read_file(PresetAccountsFile) of
-        {error, _Err} ->
+    case read_presets() of
+        {error, {_Err, PresetAccountsFile}} ->
             % no setup, no preset accounts
             erlang:error({genesis_accounts_file_missing, PresetAccountsFile});
         {ok, JSONData} ->
+            DecodedData =
+                try jsx:decode(JSONData) of
+                    %% JSX decodes an empty json object as [{}]:
+                    %% [{}] = jsx:decode(<<"{}">>).
+                    [{}] -> [];
+                    L when is_list(L) -> L
+                catch
+                  error:_ ->
+                    erlang:error(invalid_accounts_json)
+                end,
             Accounts =
                 lists:map(
                     fun({EncodedPubKey, Amt}) ->
                         {ok, PubKey} = aec_base58c:safe_decode(account_pubkey, EncodedPubKey),
-                        %% PubKey = base64:decode(EncodedPubKey),
                         {PubKey, Amt}
                     end,
-                    jsx:decode(JSONData)),
+                    DecodedData),
             % ensure deterministic ordering of accounts
             lists:keysort(1, Accounts)
+    end.
+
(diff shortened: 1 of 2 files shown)


apps/aecore/src/aec_genesis_block_settings.erl
```diff
@@ -1,26 +1,46 @@
...
-    case file:read_file(PresetAccountsFile) of
-        {error, _Err} ->
+    case ?MODULE:read_presets() of
+        {error, _Err, PresetAccountsFile} ->
             % no setup, no preset accounts
             erlang:error({genesis_accounts_file_missing, PresetAccountsFile});
         {ok, JSONData} ->
+            DecodedData =
+                case JSONData of
+                    <<"{}">> ->
```
lucafavatella: This looks fragile e.g. for whitespaces/newlines before/in-middle/after the `{}`.

apps/aecore/src/aec_genesis_block_settings.erl
```diff
@@ -1,26 +1,44 @@
...
             % ensure deterministic ordering of accounts
             lists:keysort(1, Accounts)
     end.
+
+-spec read_presets() -> {ok, binary()}| {error, atom(), string()}.
+read_presets() ->
+    PresetAccountsFile = filename:join([dir(), "accounts.json"]),
+    case file:read_file(PresetAccountsFile) of
+        {ok, _} = OK -> OK;
+        {error, Err} -> {error, Err, PresetAccountsFile}
```
ThomasArts: Since dialyzer would not catch it, it might be better to stay to tuple notation {error, {Err, PresetAccountsFile}}

apps/aecore/src/aec_genesis_block_settings.erl
```diff
@@ -1,26 +1,44 @@
...
 
 dir() ->
     filename:join(aeu_env:data_dir(aecore), ".genesis").
 
+-spec preset_accounts() -> list().
 preset_accounts() ->
-    PresetAccountsFile = filename:join([dir(), "accounts.json"]),
-    case file:read_file(PresetAccountsFile) of
-        {error, _Err} ->
+    case ?MODULE:read_presets() of
```
ThomasArts: Why ?MODULE here... it's the same module anyway?

apps/aecore/test/aec_genesis_block_settings_tests.erl
```diff
@@ -0,0 +1,83 @@
...
+     end,
+     fun(ok) ->
+         meck:unload(?TEST_MODULE)
+     end,
+     [ {"Preset accounts parsing: broken file",
+        fun() ->
+            %% empty file
+            expect_accounts(<<"">>),
+            ?assertError(invalid_accounts_json, ?TEST_MODULE:preset_accounts()),
+            %% broekn json
```
ThomasArts: broken

apps/aecore/src/aec_genesis_block_settings.erl
velzevur: I know this is ugly but I could not find a way around it.
This module is being tested and in order to use `meck` for the `read_presets/1` - a fully qualified function call must be used. Other option would be meck-ing filename module...
I could add a comment for that if you find it useful

ThomasArts: I guess this is a misunderstanding of meck. You don't need to add the module in the call to make meck pick it up. Ask Adam for details.

eproxus: Just adding this here for posterity: Meck cannot mock local function calls in a module. The alternative is to use a fully qualified call (`Mod:Func(...)`) or even better, not to test using mock functions inside modules but at module borders only.

velzevur:
Opening a new PR

velzevur:
New PR #910

PR #882, opened 2018-03-21
Update dockerhub org name to aeternity

dincho:
The [builder image](https://hub.docker.com/r/aeternity/builder/tags/) is already pushed under the new org.
I also copied the `latest` and `v0.9.0` [docker images](https://hub.docker.com/r/aeternity/epoch/tags/) to the new org.

In scope of https://www.pivotaltracker.com/story/show/155940263

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,12 +5,12 @@ references:
     setup_remote_docker:
       docker_layer_caching: true
     docker:
-      - image: aetrnty/builder
+      - image: aeternity/builder
         user: builder
     working_directory: ~/epoch
     environment:
       DOCKER_CLIENT_VERSION: "17.09.0-ce"
-      DOCKERHUB_REPO: aetrnty/epoch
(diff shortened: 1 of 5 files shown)


lucafavatella:
> I also copied the latest and v0.9.0 docker images to the new org.

Could you push v0.8.0 too? I saw some Devs still using it.

.circleci/config.yml
```diff
@@ -5,12 +5,12 @@ references:
     setup_remote_docker:
       docker_layer_caching: true
     docker:
-      - image: aetrnty/builder
+      - image: aeternity/builder
```
lucafavatella: Dependent on changes in https://github.com/aeternity/docker-builder ?

dincho:
> Could you push v0.8.0 too? I saw some Devs still using it.

Interesting. Pushed.

.circleci/config.yml
dincho: Not really, but I opened a PR for that repo too.

PR #884, opened 2018-03-22
Use position in transaction list as key for the transaction root hash

gorillainduction:
This makes the transaction root hash dependent on the order of transactions

https://www.pivotaltracker.com/n/projects/2124891/stories/156130705

### DIFF
--- a/apps/aecore/src/aec_txs_trees.erl
+++ b/apps/aecore/src/aec_txs_trees.erl
@@ -28,7 +28,7 @@
 
 -spec from_txs([aetx_sign:signed_tx(), ...]) -> txs_tree().
 from_txs(Txs = [_|_]) ->
-    lists:foldl(fun enter_signed_tx/2, empty(), Txs).
+    from_txs(Txs, 0, aeu_mtrees:empty()).
 
 -spec root_hash(txs_tree()) -> {ok, root_hash()}.
 root_hash(TxsTree) ->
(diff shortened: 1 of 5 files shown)


docs/release-notes/RELEASE-NOTES-0.10.0.md
```diff
@@ -4,6 +4,7 @@
 It:
 * Improves chain handling by storing accumulated difficulty in the db rather than recomputing it. This impacts the persisted DB.
 * Enable user configuration of HTTP API acceptors pool
+* Restructure the transaction root hash in the block header for stricter validation. This impacts consensus and the persisted DB.
```
lucafavatella: `git diff v0.9.0 origin/master -- apps/aecore/include/blocks.hrl` returns empty so it looks like this PR has the honour of being the first to break consensus.
* Could you please bump block PROTOCOL_VERSION ?
* And could you also bump GENESIS_VERSION - as so far no support for hard fork without chain restart ?

Ref https://github.com/aeternity/epoch/blob/4e96464f8be73205919e68e237433b0e3df88c8c/apps/aecore/include/blocks.hrl#L3-L5

apps/aecore/src/aec_txs_trees.erl
```diff
@@ -38,13 +38,9 @@ root_hash(TxsTree) ->
...
-
-enter_signed_tx(SignedTx, TxsTree) ->
-    V = aetx_sign:serialize_to_binary(SignedTx),
-    K = aec_hash:hash(signed_tx, V),
-    enter(K, V, TxsTree).
+from_txs([],_Position, Tree) ->
+    Tree;
+from_txs([SignedTx|Left], Position, Tree) ->
+    Key = binary:encode_unsigned(Position),
+    Val = aetx_sign:serialize_to_binary(SignedTx),
```
lucafavatella: (Not in scope for this PR.)
Are signatures in the signed tx - if multiple signature - guaranteed to be sorted?

gorillainduction: I'm addressing this in the new serialization format.

PR #885, opened 2018-03-22
Pt 155933047 peer keys

hanssv:
Add peer keys (curve25519) to aec_keys
    
Since the node account keys are scheduled for removal there is lots of duplicated code for now.
Eventually only one copy should remain.


hanssv:
It could be nice (@dincho brought up this idea) to have the public peer key accessible through HTTP, that would be another task/PR though!

PR #888, opened 2018-03-22
Add yaml option to configure expected_mine_rate

dincho:
in scope of https://www.pivotaltracker.com/story/show/154634483

### DIFF
--- a/apps/aecore/src/aec_governance.erl
+++ b/apps/aecore/src/aec_governance.erl
@@ -22,12 +22,8 @@ blocks_to_check_difficulty_count() ->
     ?BLOCKS_TO_CHECK_DIFFICULTY_COUNT.
 
 expected_block_mine_rate() ->
-    case aeu_env:user_config([<<"mining">>, <<"expected_mine_rate">>]) of
-        {ok, MineRate} ->
-            MineRate;
-        undefined ->
-            application:get_env(aecore, expected_mine_rate, ?EXPECTED_BLOCK_MINE_RATE)
-    end.
+    aeu_env:user_config_or_env([<<"mining">>, <<"expected_mine_rate">>],
+                               aecore, expected_mine_rate, ?EXPECTED_BLOCK_MINE_RATE).
 
 block_mine_reward() ->
     ?BLOCK_MINE_REWARD.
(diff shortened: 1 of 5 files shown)


dincho:
The second commit has been tested on env/dev1 see https://github.com/aeternity/epoch/commit/f2bd26f08573b8fd8286a3d3422dbe76cd64290b

dincho:
> Not needed, but you could add maximum and minimum to the json schema.

I guess `0` should also work - meaning "do as fast as possible" ? What min/max do you suggest?

apps/aecore/src/aec_governance.erl
```diff
@@ -22,8 +22,12 @@ blocks_to_check_difficulty_count() ->
     ?BLOCKS_TO_CHECK_DIFFICULTY_COUNT.
 
 expected_block_mine_rate() ->
-    application:get_env(aecore, expected_mine_rate,
-                        ?EXPECTED_BLOCK_MINE_RATE).
+    case aeu_env:user_config([<<"mining">>, <<"expected_mine_rate">>]) of
```
uwiger: This would be equivalent to
```erlang
aeu_env:user_config_or_env([<<"mining">>, <<"expected_mine_rate">>],
                            aecore, expected_mine_rate,
                            ?EXPECTED_BLOCK_MINE_RATE)
```

dincho: Already changed locally, will push soon

dincho: in 0767c438

apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -209,6 +209,11 @@
                     "Maximum time (milliseconds) for each attempt to mine a block with a specific nonce.",
                     "type" : "integer"
                 },
+                "expected_mine_rate" : {
+                    "description" : "Expected mine rate (milliseconds) between blocks. Used in governance.",
+                    "type" : "integer",
+                    "default": 300000
```
lucafavatella: Minimum strictly greater then 0 beneficial though not mandatory.

dincho: just pushed 937da894

PR #889, opened 2018-03-22
Take into account that contract also spends gas now.

happi:


lucafavatella:
CI passed. I see two approvals. Merging.

PR #890, opened 2018-03-22
Pt 156157920 crash under sync recovery

ThomasArts:
When a worker that fetches blocks from another chain crashes unexpectetly, we should remve it from the pool of sync workers.

The difficult part of this patch is the test case.

### DIFF
--- a/apps/aecore/test/aecore_sync_SUITE.erl
+++ b/apps/aecore/test/aecore_sync_SUITE.erl
@@ -29,7 +29,8 @@
     ensure_tx_pools_empty/1,
     ensure_tx_pools_one_tx/1,
     report_metrics/1,
-    check_metrics_logged/1
+    check_metrics_logged/1,
+    crash_syncing_worker/1
    ]).
 
 -include_lib("common_test/include/ct.hrl").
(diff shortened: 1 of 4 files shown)


apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -310,6 +312,54 @@ check_no_system_metrics_sent() ->
...
+
+    %% Set the same mining_rate to validate target
+    application:set_env(aecore, expected_mine_rate, 10), 
+
+    aecore_suite_utils:start_node(Dev2, Config),
+    aecore_suite_utils:connect(N2),
+    ct:log("node connected ~p", [N2]),
+    ok = rpc:call(N2, application, set_env, [aecore, ping_interval_limits, {500, 1000}], 5000),
+
+    spawn_link(fun() -> kill_sync_worker(N2, <<"http://ThomasComputer:3013">>) end),
```
hanssv: This looks interesting - either the node name doesn't matter or else something is fishy?

hanssv:
I also have some slight concerns about the test being timing dependent, but it might be hard to avoid here...

apps/aecore/test/aecore_sync_SUITE.erl
velzevur: It doesn't really matter - it is going to be killed right away

ThomasArts: It matters... only when we kill the process that is syncing, we test the right thing. Updated this.

PR #891, opened 2018-03-23
New sync protocol

hanssv:
More or less the complete solution - see TODO below.

Removes the HTTP API for ping, and instead introduce a `sync_port` where the node will listen for incoming TCP connections. A secure Noise handshake is made (using the protocol version and genesis hash to ensure nodes are compatible) and then this connection is kept for ping, sync, gossip, etc.

This includes a major rewrite of `aec_peers` and the introduction of `aec_peer_connection` - the new module contains all the logic for P2P communication (previously scattered between `aeu_request`, `aehttp_dispatch_ext` etc...).

TODO:
- [ ] Document the changes to the configuration and the new overall sync behavior 
- [x] Implement message fragmentation (once serialized blocks gets bigger than 2^16 bytes we hit this)
- [x] Fix the tests for blocked nodes + aec_peers
- [x] Fix python tests
- [x] Fix the Peers endpoint (format changed)
- [x] Python test_use_cases (should use Peers endpoint) 
- ~~Eventually give up if connection/handshake works but we fail pinging - postpone this to upcoming work with peer discovery~~
- [x] Remove more dead-code (aeu_request, etc...)
- [x] Go through all the logging and remove and re-label them
- [ ] Synchronize with the work in serialization - currently using `msgpack`


thepiwo:
Is there some documentation on how overall sync works, including this?

hanssv:
Read the TODO :-)

apps/aecore/src/aec_sync.erl
```diff
@@ -414,52 +365,43 @@ process_job([{_T, Job}]) ->
...
+      || PeerInfo <- Peers ].
 
-await_aehttp_and_ping_peer(Uri) ->
+await_aehttp_and_ping_peer(PeerId) ->
     %% Don't ping until our own HTTP endpoint is up. This is not strictly
     %% needed for the ping itself, but given that a ping can quickly
     %% lead to a greater discovery, we should be prepared to handle pings
     %% ourselves at this point.
     %% The URI comes from aec_peers and is a valid http url.
     case await_aehttp() of
```
ThomasArts: Should we really still wait for aehttp?

hanssv: No, I don't think we need to do that anymore... Probably just a leftover from before!

apps/aecore/src/aec_peers.erl
```diff
@@ -655,43 +608,78 @@ ping_interval_limits() ->
...
+    case Prev of
+        undefined -> ok;
+        _ -> try erlang:cancel_timer(Prev, [{async, true}, {info, false}])
+             catch error:_ -> ok end
+    end,
+    Peer#peer{timer_tref = TRef}.
+
+backoff_timeout(#peer{ retries = Retries, trusted = Trusted }) ->
+    case Retries of
+        N when N < ?MAX_RETRIES ->
```
happi: How do you keep MAX_RETRIES and this list in sync?
Perhaps define BACKOFF_TIMES at the same place as MAX_RETRIES is defined?

apps/aecore/src/aec_sync.erl
```diff
@@ -69,105 +66,62 @@ connect_peer(Uri) ->
...
 %% block we agree upon. We use a binary search to find out.
 %% From the height we agree upon, we start asking for blocks to add to the chain.
-%% 
+%%
 %% When an additional Ping arrives for which we agree upon genesis, we have
 %% the following possibilities:
 %% 1. It has worst top hash than our node, do not include in sync
 %% 2. It has better top hash than our node
-%%    We binary search for a node that we agree upon (could be genesis) 
+%%    We binary search for a node that we agree upon (could be genesis)
```
happi: Is "node" the right word here? Is it not the "block" we search for?

apps/aecore/src/aec_peer_connection_listener.erl
```diff
@@ -0,0 +1,56 @@
...
+            gen_tcp:close(LSock)
+    end.
+
+acceptor(Parent, LSock) ->
+    Res = gen_tcp:accept(LSock),
+    case Res of
+        {ok, Sock} ->
+            LPort = aeu_env:user_config_or_env([<<"sync">>, <<"port">>], aecore, sync_port, ?DEFAULT_SYNC_PORT),
+            {ok, {Host, Port}} = inet:peername(Sock),
+            lager:debug("ZZPC: port ~p got tcp_connect ~p from ~p", [LPort, Sock, {Host, Port}]),
```
dincho: What `ZZPC` stands for?

apps/aecore/src/aec_peers.erl
```diff
@@ -270,178 +236,270 @@ unblock_all() ->
...
                }).
 
 start_link() ->
     gen_server:start_link({local, ?MODULE} ,?MODULE, ok, []).
 
 init(ok) ->
-    {Scheme, Host, Port} = aeu_env:local_peer(),
-    LocalPeer = #peer{scheme = Scheme, host = Host, port = Port},
+    {_Scheme, Host, _Port} = aeu_env:local_peer(),
+    Port = aeu_env:user_config_or_env([<<"sync">>, <<"port">>], aecore, sync_port, 1234),
```
dincho: I see this configuration related call is all over the place, and in this particular case it's even uses hardcoded (probably) wrong default port.

Isn't it better to wrap it small function like most of the other modules do?

apps/aecore/src/aec_sync.erl
hanssv: I just fixed the whitespace issue that Thomas introduced, but yes it looks wrong :-)

apps/aecore/src/aec_peer_connection_listener.erl
hanssv: ZZ for something that is easy to grep for, PC for "peer connection"

Most of the debug should go away and what should stay will not have this prefix in the final version!

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -0,0 +1,827 @@
...
+            RemotePub = enoise_keypair:pubkey(enoise_hs_state:remote_keys(FinalState)),
+            lager:debug("ZZPC: eniose:accept() successful remotepub ~p", [RemotePub]),
+            %% Report this to aec_peers!? And possibly fail?
+            %% Or, we can't do this yet we don't know the port?!
+            TRef = erlang:start_timer(?FIRST_PING_TIMEOUT, self(), first_ping_timeout),
+            {noreply, S#{ status => {connected, ESock}, r_pubkey => RemotePub,
+                          first_ping_tref => TRef }};
+        {error, _Reason} ->
+            %% What to do here? Close the socket and stop?
+            lager:debug("ZZPC: eniose:accept() failed - ~p", [_Reason]),
```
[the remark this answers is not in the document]
hanssv: The lager-logging is still WIP - lots should go away and some things should be info/error

This should be `info` I think - we can fail here if the peers disagree on genesis hash or version!

apps/aecore/src/aec_peers.erl
hanssv: Good comment - will fix!

apps/aecore/src/aec_peers.erl
hanssv: Yeah, this is lazy code - will try to remember to fix it...

And yes, we should decide on a standard listen port for sync!

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -0,0 +1,865 @@
...
+    send_send_tx(State, From, Hash);
+handle_call({send_block, Hash}, From, State) ->
+    send_send_block(State, From, Hash).
+
+handle_cast(retry, S = #{ host := Host, port := Port, status := error }) ->
+    Self = self(),
+    lager:debug("ZZPC: connecting to ~p", [{Host, Port}]),
+    Pid = spawn(fun() -> try_connect(Self, Host, Port, 1000) end),
+    {noreply, S#{ status => {connecting, Pid} }};
+handle_cast(owns_tcp_socket, S0 = #{ role := responder, tcp_sock := TcpSock }) ->
```
eproxus: I find the other cast messages straightforward because they are verbs, but I don't understand this name... perhaps it should be change to start with a verb?

hanssv: How about `give_tcp_socket_ownership`?

apps/aecore/src/aec_peers.erl
hanssv: DONE

apps/aecore/src/aec_peer_connection_listener.erl
hanssv: DONE

apps/aecore/src/aec_peer_connection.erl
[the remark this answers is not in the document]
hanssv: DONE

apps/aecore/src/aec_sync.erl
hanssv: DONE

apps/aecore/src/aec_peers.erl
```diff
@@ -270,178 +226,279 @@ unblock_all() ->
...
+    end.
+
+handle_cast({log_ping, Ok, PeerId, _Time}, State) ->
+    %% Log the ping and schedule another as long as we have a connection
+    update_ping_metrics(Ok),
+    case lookup_peer(PeerId, State) of
+        none ->
+            lager:debug("Reported ping event for unknown peer - ~p", [ppp(PeerId)]),
+            {noreply, State};
+        {value, _Hash, Peer = #peer{ connection = {connected, _PeerCon} }} ->
```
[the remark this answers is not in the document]
hanssv: It wasn't obvious to me what to do in this negative case (when we have an active Noise connection - i.e. genesis and version match but the other side doesn't want to respon properly to ping). I've deferred this case to the upcoming tasks where we more carefully design peer discovery and connection pools, etc.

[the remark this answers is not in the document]
hanssv: I'm adding a comment to the code.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -0,0 +1,825 @@
...
+
+local_ping_obj(S) ->
+    PingObj = aec_sync:local_ping_object(),
+    PingObj#{ <<"host">> => maps:get(local_host, S),
+              <<"port">> => maps:get(local_port, S) }.
+
+%% -- Get Header by Hash -----------------------------------------------------
+handle_get_header_by_hash(S, MsgObj) ->
+    Msg =
+        case maps:get(<<"hash">>, MsgObj, undefined) of
```
gorillainduction: Break this out to separate function.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -0,0 +1,825 @@
...
+
+%% -- Local functions --------------------------------------------------------
+get_request(S, Kind) ->
+    maps:get(Kind, maps:get(requests, S, #{}), none).
+
+set_request(S, Kind, From) ->
+    Rs = maps:get(requests, S, #{}),
+    S#{ requests => Rs#{ Kind => {Kind, From} } }.
+
+drop_request(S, Kind) ->
```
gorillainduction: `drop_request` sounds as if it wasn't handled. I think this could be better named for readability.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -0,0 +1,825 @@
...
+                {{T, _}, {T, _}} ->
+                    lager:debug("Same top blocks", []),
+                    aec_sync:server_get_missing_blocks(PeerId);
+                {{_, DL}, {_, DR}} when DL > DR ->
+                    lager:debug("Our difficulty is higher", []),
+                    aec_sync:server_get_missing_blocks(PeerId);
+                {{_, _}, {_, DR}} ->
+                    aec_sync:start_sync(PeerId, RTHash, DR)
+            end,
+            aec_peers:add_and_ping_peers(RPeers),
```
gorillainduction: Match on return to show it is used for side effects

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -0,0 +1,865 @@
...
+    call(PeerId, get_mempool).
+
+send_tx(PeerId, Tx) ->
+    call(PeerId, {send_tx, Tx}).
+
+send_block(PeerId, Tx) ->
+    call(PeerId, {send_block, Tx}).
+
+stop(PeerCon) ->
+    gen_server:cast(PeerCon, stop).
```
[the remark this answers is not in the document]
eproxus: It could be made synchronous with a `wait_for_exit(PeerCon)` after the cast, and an error if a timeout is reached.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -0,0 +1,825 @@
...
+        Err = {error, _} ->
+            Err
+    end.
+
+
+%% -- gen_server callbacks ---------------------------------------------------
+init([Opts]) ->
+    Version = <<?PROTOCOL_VERSION:16>>,
+    Genesis = aec_chain:genesis_hash(),
+    {ok, Opts#{ version => Version, genesis => Genesis }, 0}.
```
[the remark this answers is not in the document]
eproxus: And if really needed, the only safe way to do these things is using the new `gen_statem` and returning a pre-defined action from init.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -0,0 +1,825 @@
...
+                _ -> {error, no_connection}
+            end;
+        Err = {error, _} ->
+            Err
+    end.
+
+
+%% -- gen_server callbacks ---------------------------------------------------
+init([Opts]) ->
+    Version = <<?PROTOCOL_VERSION:16>>,
```
lucafavatella:
```
6> <<16#0000:16>>.
<<0,0>>
7> <<16#1000:16>>.
<<16,0>>
8> <<16#10000:16>>.
<<0,0>>
```

It looks like this shall cater for 8 bytes https://github.com/aeternity/epoch/blob/95b2d26ca66dc3c4f59b7bef2134afd82b1cee6c/apps/aecore/src/aec_headers.erl#L109 https://github.com/aeternity/protocol/pull/56/files#diff-3976273847c0bfced542c67100a7990bR25

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -0,0 +1,825 @@
...
+handle_cast(retry, S = #{ host := Host, port := Port, status := error }) ->
+    Self = self(),
+    Pid = spawn(fun() -> try_connect(Self, Host, Port, 1000) end),
+    {noreply, S#{ status => {connecting, Pid} }};
+handle_cast(give_tcp_socket_ownership, S0 = #{ role := responder, tcp_sock := TcpSock }) ->
+    S = #{ version := Version, genesis := Genesis,
+           seckey := SecKey, pubkey := PubKey } = ensure_genesis(S0),
+    NoiseOpts = [ {noise, <<"Noise_XK_25519_ChaChaPoly_BLAKE2b">>}
+                , {s, enoise_keypair:new(dh25519, SecKey, PubKey)}
+                , {prologue, <<Version/binary, Genesis/binary>>}
```
lucafavatella: I understand that if different prologue than handshake fails. This is passing genesis hash - that makes sense (and BTW contains the initial consensus protocol version) - and *current* consensus/block protocol version - that I do not understand fully. What I would expect here is a peer-to-peer protocol specific version, that changes completely independently from the consensus protocol.

lucafavatella: Is genesis hash here meant to be base58c encoded for human readability? (Just checking - I expect no.)

apps/aecore/src/aec_peer_connection_listener.erl
```diff
@@ -0,0 +1,51 @@
...
+    ?MODULE ! stop,
+    ok.
+
+pc_listener() ->
+    {_Scheme, Host, _Port} = aeu_env:local_peer(),
+    {ok, SecKey} = aec_keys:peer_privkey(),
+    {ok, PubKey} = aec_keys:peer_pubkey(),
+    Port = aec_peers:sync_port(),
+    lager:info("Starting peer_connection_listener at port ~p", [Port]),
+    {ok, LSock} = gen_tcp:listen(Port, [{active, false}, binary, {reuseaddr, true}]),
```
lucafavatella: Configurable max number of acceptors?
Configurable max timeout(s)?
Using ranch instead of gen_tcp for having only one code path for connections?

apps/aecore/src/aec_peer_connection.erl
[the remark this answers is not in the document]
hanssv: Please elaborate? This stop function is called from aec_peers:handle_call that is called from the aec_peer_connection that we are trying to stop?

apps/aecore/src/aec_peer_connection.erl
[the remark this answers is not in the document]
hanssv: I just want to continue as quickly as possible, but it isn't crucial so should be fine here...

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -0,0 +1,825 @@
...
+drop_request(S, Kind) ->
+    Rs = maps:get(requests, S, #{}),
+    S#{ requests => maps:remove(Kind, Rs) }.
+
+try_connect(Parent, Host, Port, Timeout) when is_binary(Host) ->
+    try_connect(Parent, binary_to_list(Host), Port, Timeout);
+try_connect(Parent, Host, Port, Timeout) ->
+    Res =
+        try
+            gen_tcp:connect(Host, Port, [binary, {reuseaddr, true}, {active, true}], Timeout)
```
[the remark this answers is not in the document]
hanssv: The protocol says that I should send first... But in general yes, will fix it.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -0,0 +1,825 @@
...
+        catch _E:R ->
+            {error, R}
+        end,
+    case Res of
+        {ok, Sock} -> gen_tcp:controlling_process(Sock, Parent);
+        {error, _} -> ok
+    end,
+    Parent ! {connected, self(), Res}.
+
+ensure_genesis(S = #{ genesis := G }) when is_binary(G) ->
```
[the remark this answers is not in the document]
hanssv: Yes, at some point I would like to say `wait_for_conductor` but in this particular instance I actually just need the genesis hash and nothing else - will think about it :-)

apps/aecore/src/aec_peer_connection_listener.erl
hanssv: This will be addressed further on in https://www.pivotaltracker.com/story/show/156222652

apps/aeutils/test/data/epoch_full.yaml
```diff
@@ -2,17 +2,21 @@
...
 
 # Pre-configured addresses of epoch nodes NOT to contact
 blocked_peers:
-    - "http://some-really-bad-peer-address:3013/"
+    - peer:
+        host: "some-really-bad-peer-address"
+        port: 3013
+        pubkey: "pp$2M9oPohzsWgJrBBCFeYi3PVT4YF7F2botBtq6J1EGcVkiutx3R"
 
 sync:
```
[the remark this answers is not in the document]
hanssv: Sorry, what do you mean? `sync_port`?

apps/aecore/src/aec_peer_connection.erl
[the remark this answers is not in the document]
eproxus: I could only find a call to `enoise` from this gen_server. Perhaps I'm mistaken...

apps/aeutils/test/data/epoch_full.yaml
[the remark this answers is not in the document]
lucafavatella: Yes - sync -> port. https://github.com/aeternity/epoch/pull/891/files#diff-3bd87451a027dcf4afb8802d11556894R74

apps/aecore/src/aec_peer_connection.erl
hanssv: Yes of course, throwing PROTOCOL_VERSION in there was just a hack that I forgot to fix this has little to do with consensus! Good catch!

hanssv: No I have not planned to base58c encode it.

hanssv:
Still most of the documentation is TODO... But I'd rather see the code running first!

apps/aecore/src/aec_peer_connection.erl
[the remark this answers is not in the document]
hanssv: Yes you are :-) 
For example line 212 `            case aec_peers:connect_peer(peer_id(S), self()) of`
and quite a few other places?!

apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -218,14 +219,26 @@ start_blocked_second(Config) ->
...
-    T0 = os:timestamp(),
     aecore_suite_utils:start_node(Dev2, Config),
     aecore_suite_utils:connect(N2),
     timer:sleep(2000),
-    ct:log("Peers on dev1: ~p", [rpc:call(N1, aec_peers, all, [], 5000)]),
-    ct:log("Unblocked peers on dev1: ~p",
-           [rpc:call(N1, aec_peers, get_random, [10], 5000)]),
-    await_sync_abort(T0, [N1, N2]).
+
+    %% Check that there is only one non-blocked peer (dev3) and no connected peers
```
lucafavatella: s/dev3/dev2 ?

hanssv: No, dev2 is blocked so dev3 is the non-blocked one!?

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -488,16 +488,12 @@ handle_request('GetAccountBalance', Req, _Context) ->
...
-                        #{uri => URI,
-                          last_seen => LastSeen}
-                    end,
-                    aec_peers:all()),
+            Peers = aec_peers:all(),
             Blocked = aec_peers:blocked(),
-            {200, [], #{peers => Peers,
-                        blocked => Blocked}};
+            Enc = fun(Xs) -> [ X#{ pubkey := aec_base58c:encode(peer_pubkey, PK) }
+                               || X = #{ pubkey := PK } <- Xs ] end,
```
lucafavatella: This list comprehension is an implicit filter. I would prefer hardcoding expectation that all peers have pubkey by using lists:map - though this is debatable so not need to change.

hanssv:
Rebased on master and (hopefully) fixed peer key deployment - we will need to deploy twice (initially) but otherwise it should be rock solid :-)

apps/aecore/src/aec_peer_connection_listener.erl
dincho: I would also like to see configurable `listen_address`, like all other listen sockets.

From ops perspective it's much easier to set the `listen_address` than configuring firewall rules. (e.g. in case of private setups)

apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -8,38 +8,77 @@
...
+                        }
+                    }
+                }
             }
         },
         "sync" : {
             "type" : "object",
             "additionalProperties" : false,
             "properties" : {
+                "port" : {
```
dincho: I'd like to see `listen_address` as well (see my other commend)

apps/aeutils/priv/extensions/get_peer_key.sh
```diff
@@ -0,0 +1,12 @@
...
+# get the peer pubkey
+CODE="{ok, PK} = aec_keys:peer_pubkey(),
+      binary_to_list(aec_base58c:encode(peer_pubkey, PK))."
+
+PK=$(relx_nodetool eval $CODE)
+
+# now print it out
+echo $PK
+
+exit $?
```
dincho: isn't this the exit status of the echo ?

deployment/ansible/templates/epoch.yaml.j2
```diff
@@ -1,7 +1,14 @@
...
+sync:
+    port: {{ config.sync.port }}
+
 peers:
     {% for host in groups[hosts_group] | difference([inventory_hostname]) %}
-- "http://{{ hostvars[host]['ansible_default_ipv4']['address'] }}:3013/"
+    {% if hostvars[host]['sync_public_key'] is defined %}
+- peer: {host: {{ hostvars[host]['ansible_default_ipv4']['address'] }}, port: 3015, pubkey: {{ hostvars[host]['sync_public_key'] }}}
+    {% else %}
+- peer: {host: {{ hostvars[host]['ansible_default_ipv4']['address'] }}, port: 3015, pubkey: "no-key-found-peer-will-be-ignored"}
```
dincho: Nice workaround

deployment/ansible/templates/epoch.yaml.j2
```diff
@@ -1,7 +1,14 @@
 ---
+sync:
+    port: {{ config.sync.port }}
+
 peers:
     {% for host in groups[hosts_group] | difference([inventory_hostname]) %}
-- "http://{{ hostvars[host]['ansible_default_ipv4']['address'] }}:3013/"
+    {% if hostvars[host]['sync_public_key'] is defined %}
+- peer: {host: {{ hostvars[host]['ansible_default_ipv4']['address'] }}, port: 3015, pubkey: {{ hostvars[host]['sync_public_key'] }}}
```
dincho: could you use `{{ config.sync.port }}` here as well? Does it make sense?

dincho: also multi-row peer object will be much more readable

hanssv: Absolutely! This was the last bug I fixed it still said 3013... Was too tired at that point.

apps/aeutils/priv/extensions/get_peer_key.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# get the peer pubkey
+CODE="{ok, PK} = aec_keys:peer_pubkey(),
+      binary_to_list(aec_base58c:encode(peer_pubkey, PK))."
+
+$(relx_nodetool eval $CODE)
```
lucafavatella: It looks like an explicit exit is needed? https://github.com/aeternity/epoch/blob/28f01e6b1c2923167047d7dbe503c411a0816983/apps/check_config/priv/extensions/check_config.sh#L10-L11

apps/aeutils/priv/epoch_config_schema.json
hanssv: adding to the backlog

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -0,0 +1,823 @@
...
+send_send_block(S = #{ status := {connected, _ESock} }, From, Block) ->
+    BlockMap = aec_blocks:serialize_to_map(Block),
+    send_msg(S, ?MSG_SEND_BLOCK, BlockMap),
+    {noreply, set_request(S, send_block, From)}.
+
+handle_send_block(S, MsgObj) ->
+    Msg =
+        case aec_blocks:deserialize_from_map(MsgObj) of
+            {ok, Block} ->
+                Header = aec_blocks:to_header(Block),
```
gorillainduction: Left over debug code?

hanssv: No we were logging that in the original design... I just kept it without thinking too much...

apps/aeutils/priv/extensions/get_peer_key.sh
```diff
@@ -1,9 +1,34 @@
...
+
 # get the peer pubkey
 CODE="{ok, PK} = aec_keys:peer_pubkey(),
       binary_to_list(aec_base58c:encode(peer_pubkey, PK))."
 
-relx_nodetool eval $CODE
+! relx_nodetool eval $CODE
+
+if [ $? -eq 1 ]; then
+    exit 0
```
lucafavatella: i.e. if first CODE invocation returned 0, return 0. Ok.

apps/aeutils/priv/extensions/get_peer_key.sh
```diff
@@ -1,9 +1,34 @@
...
+fi
+
+CODE="{ok, Bin} = file:read_file(filename:join(\"$KEYSDIR\", \"peer_key.pub\")),
```
lucafavatella: Not sure why this second code attempt is needed?

docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
     command: -aecore expected_mine_rate ${EPOCH_MINE_RATE:-15000}
     volumes:
       - ${PWD}/docker/epoch_node1_mean16.yaml:/home/epoch/epoch.yaml
+      - ${PWD}/docker/keys/node1:/home/epoch/node/keys
```
lucafavatella: How does this impact ability of user to specify mining key ? https://github.com/aeternity/epoch/blob/c4938333d80e0892cc43d265d5470fcadc1b9702/docs/docker.md#persisting-data

apps/aeutils/priv/extensions/get_peer_key.sh
hanssv: If the epoch node is not running/not started we read the key directly from the file. Will be good on failed deployments.

docker-compose.yml
hanssv: This is just for the docker-compose example (I hope) not sure exactly how it relates to the instructions in docker.md?

lucafavatella: You are right.

deployment/ansible/templates/epoch.yaml.j2
```diff
@@ -1,8 +1,21 @@
 ---
+sync:
+    port: {{ config.sync.port }}
+
 peers:
-    {% for host in groups[hosts_group] | difference([inventory_hostname]) %}
-- "http://{{ hostvars[host]['ansible_default_ipv4']['address'] }}:3013/"
-    {% endfor %}
+{% for host in groups[hosts_group] | difference([inventory_hostname]) %}
+{% if hostvars[host]['sync_public_key'] is defined %}
```
lucafavatella: @dincho It would be nice to have an improvement to the deployment script so to avoid that for some glitch/typo no node in deployment has peer key.

apps/aecore/test/aec_eunit_SUITE.erl
```diff
@@ -105,37 +104,37 @@ application_test(Config) ->
...
-    ct:log("trying to connect again --> job scheduled"),
-    timer:sleep(50),
-    SyncPid = whereis(aec_sync), %% Sync has not been restarted!
-    Peers = aec_peers:all(),
-    ct:log("tried twice and although it failed, it is a peer: ~p", [Peers]),
-    1 = length(Peers),
-
-    ok = application:stop(aecore),
-    app_stop(StartedApps -- ?TO_BE_STOPPED_APPS_BLACKLIST, TempDir).
+%% aec_sync_test(Config) ->
```
uwiger: Perhaps delete?

hanssv: I had some hope of recreating it... Let's keep it there as a reminder for me!?

PR #892, opened 2018-03-23
Fine-tune `make clean` so not to remove built rocksdb

lucafavatella:

----

When switching branches and having to clean working directory often, recompiling rocksdb takes significant time.


lucafavatella:
I will this locally a bit more then remove the WIP label.

lucafavatella:
Used for #893 , this speeds things up.

PR #896, opened 2018-03-26
Add fast_integration environment

dincho:
in scope of https://www.pivotaltracker.com/story/show/154634483

depend on https://github.com/aeternity/infrastructure/pull/33 (fast_integration is deployed manually already, so this can be merged)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -386,6 +386,12 @@ jobs:
       - DEPLOY_ENV: uat
     steps: *deploy_steps
 
+  deploy_fast_integration:
+    <<: *container_config
+    environment:
+      - DEPLOY_ENV: fast_integration
+    steps: *deploy_steps
+
   github_release_osx:
     <<: *container_config
     steps:
@@ -603,3 +609,23 @@ workflows:
                 - master
     jobs:
       - uat_chain_snapshot
+
+  nightly_fast_integration_deploy:
+    triggers:
+      - schedule:
+          cron: "0 23 * * *"
+          filters:
+            branches:
+              only:
+                - master
+    jobs:
+      - deploy_fast_integration:
+          requires:
+            - test
+            - eunit
+            - uat_tests
+            - static_analysis
+            - linux_package
+          filters:
+            branches:
+              only: master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -629,3 +629,22 @@ workflows:
           filters:
             branches:
               only: master
+
+  mock_fast_integration_deploy:
+    triggers:
+      - schedule:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -602,3 +608,23 @@ workflows:
...
+      - schedule:
+          cron: "0 23 * * *"
+          filters:
+            branches:
+              only:
+                - master
+    jobs:
+      - deploy_fast_integration:
+          requires:
+            - test
```
lucafavatella: Do these jobs get triggered automatically overnight because of this dependency?

dincho: Good question. I'll shall setup a mock

dincho: Theoretically, `test`, `eunit`, `static_analysis` and `linux_package` run on all branches and `uat_tests` run on `master`, so it looks like that should be fine.

dincho: I pushed a mock in 40145eb1 that shall run at 9:30 UTC (after ~10mins)

dincho:
```
We weren't able to start this workflow.
The following jobs are unreachable: deploy_fast_integration
At least one job in the workflow must have no dependencies.
deploy_fast_integration: requires job "static_analysis" but "static_analysis" is not part of this workflow.
deploy_fast_integration: requires job "eunit" but "eunit" is not part of this workflow.
deploy_fast_integration: requires job "linux_package" but "linux_package" is not part of this workflow.
deploy_fast_integration: requires job "test" but "test" is not part of this workflow.
```

dincho: @lucafavatella I pushed new mock 9be03d24

I guess we should just assume the master is in good shape and depend only on `linux_package`

dincho: https://circleci.com/workflow-run/7b17470a-4373-4f19-b68c-797cf02fde98

dincho: Passed. I prepared changes in 7e592f33.

I prefer to "risk" and rely on stable master than copy pasting all the jobs in this cron workflow.

.circleci/config.yml
```diff
@@ -602,3 +608,19 @@ workflows:
...
+      - schedule:
+          cron: "0 23 * * *"
+          filters:
+            branches:
+              only:
+                - master
+    jobs:
+      - deploy_fast_integration:
+          requires:
+            - linux_package
```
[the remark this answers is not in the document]
dincho: I thought I added it, at least it was in the mock/test workflow. I'll fix that in a while.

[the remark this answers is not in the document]
dincho: @lucafavatella fixed in 9a827ad9

PR #897, opened 2018-03-26
Add new binary serialization for chain objects

gorillainduction:
Use RLP encoding rather than message pack for low level serialization.

This changes the binaries on which root hashes are built for:
* accounts
* oracle query
* oracles
* contracts
* contract calls
* name service names
* name service commitments
* signed transactions
* coinbase transaction
* spend transaction
* oracle transactions
* name transactions
* contract transactions

Add basic validation of fields for spend transaction to ensure we can serialize safely

Change the message pack API in a non-backwards compatible way


gorillainduction:
@lucafavatella I can confirm that I didn't run the Python tests ;-)
Will do. Thanks for the heads up.

gorillainduction:
@lucafavatella 
* The HTTP API has indeed changed, but the documentation is under specified, so the documentation is still valid. I have been in contact with the aepps team. Will create a follow up ticket for the message pack interface. 
* The WS API is unchanged.
* The intended usage document is not affected.

gorillainduction:
@lucafavatella They would not be deserialized. Possibly, this is not protected enough so we would get unexpected crasches, but since the genesis and protocol versions are changed in this release, it should not cause any problems.

PR #898, opened 2018-03-26
Add system tests with docker backend

sylane:
* See https://github.com/aeternity/epoch/wiki/System-Testing
 * Set the environment variable EPOCH_DISABLE_NODE_CLEANUP=1
   to keep the containers and networks created during tests.
 * Disabled test `docker_keeps_data` because the last blocks
   of the chain seems to get lost when restarting a node.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -123,7 +123,15 @@
                                         "_build/prod/lib/aeutils/priv/epoch_config_schema.json",
                                         "data/epoch_config_schema.json"}]}
                            ]}
-                   ]}
+                   ]},
+            {system_test, [
+                {extra_src_dirs, ["system_test", "system_test/helpers"]},
+                {shell, [{apps, []}]},
+                {deps, [
+                    bbmustache,
+                    hackney
+                ]}
+            ]}
            ]
 }.
 
--- a/system_test/aest_sync_SUITE.erl
+++ b/system_test/aest_sync_SUITE.erl
@@ -1,0 +1,222 @@
+-module(aest_sync_SUITE).
+
+%=== EXPORTS ===================================================================
+
+% Common Test exports
+-export([all/0]).
+-export([init_per_testcase/2]).
+-export([end_per_testcase/2]).
+
+% Test cases
+-export([new_node_joins_network/1,
+         docker_keeps_data/1,
+         crash_and_continue_sync/1]).
+
+-import(aest_nodes, [
+    setup_nodes/2,
+    start_node/2,
+    stop_node/2, stop_node/3,
+    kill_node/2,
+    http_get/5,
+    request/4,
+    wait_for_height/4,
+    assert_synchronized/2
+]).
+
+%=== INCLUDES ==================================================================
+
+-include_lib("eunit/include/eunit.hrl").
+
+%=== MACROS ====================================================================
+
+-define(MINING_TIMEOUT, 1000).
+
+-define(NODE1, #{
+    name    => node1,
+    peers   => [node2],
+    backend => aest_docker,
+    % Change to a compatible fixed version when possible
+    source  => {pull, "aeternity/epoch:local"}
+}).
+
+-define(NODE2, #{
+    name    => node2,
+    peers   => [node1],
+    backend => aest_docker,
+    % Change to a compatible fixed version when possible
+    source  => {pull, "aeternity/epoch:local"}
+}).
+
+-define(NODE3, #{
+    name    => node3,
+    peers   => [node1],
+    backend => aest_docker,
+    source  => {pull, "aeternity/epoch:local"}
+}).
+
+%=== COMMON TEST FUNCTIONS =====================================================
+
+all() -> [
+    new_node_joins_network
+    % , docker_keeps_data
+    , crash_and_continue_sync
+].
+
+init_per_testcase(_TC, Config) ->
+    %% Some parameters depend on the speed and capacity of the docker containers:
+    aest_nodes:ct_setup([ {blocks_per_second, 3},
+                          {node_startup_time, 5000}, %% Time it takes to get the node to respond to http
+                          {node_stop_time, 20000}    %% Time it takes to get the node to stop cleanly
+                          | Config]).
+
+end_per_testcase(_TC, Config) ->
+    aest_nodes:ct_cleanup(Config).
+
+%=== TEST CASES ================================================================
+
+%% A few tests that verify that our assumptions are right for docker timings and
+%% API.
+
+%% A node with a newer version of the code can join and synchronize
+%% to a cluster of older nodes.
+new_node_joins_network(Cfg) ->
+    Length = 20,
+    NodeStartupTime = proplists:get_value(node_startup_time, Cfg), 
+
+    setup_nodes([?NODE1, ?NODE2, ?NODE3], Cfg),
+
+    %% Starts a chain with two nodes
+    start_node(node1, Cfg),
+    start_node(node2, Cfg),
+    wait_for_height(0, [node1, node2], NodeStartupTime, Cfg),
+
+    %% Mines for 20 blocks and calculate the average mining time
+    StartTime = os:timestamp(),
+    wait_for_height(Length, [node1, node2], Length * ?MINING_TIMEOUT, Cfg),
+    EndTime = os:timestamp(),
+    %% Average mining time per block plus 50% extra
+    MiningTime = round(timer:now_diff(EndTime, StartTime) * 1.5)
+                 div (1000 * Length),
+
+    Top1 = request(node1, [v2, 'top'], #{}, Cfg),
+    ct:log("Node 1 top: ~p~n", [Top1]),
+    Height1 = request(node1, [v2, 'block-by-height'], #{height => Length}, Cfg),
+    ct:log("Node 1 at height ~p: ~p~n", [Length, Height1]),
+    Height2 = request(node2, [v2, 'block-by-height'], #{height => Length}, Cfg),
+    ct:log("Node 2 at height ~p: ~p~n", [Length, Height2]),
+
+    %% Checks node 1 and 2 are synchronized
+    ?assertEqual(Height1, Height2),
+
+    %% Starts a third node and check it synchronize with the first two
+    start_node(node3, Cfg),
+    wait_for_height(0, [node3], NodeStartupTime, Cfg),
+    ct:log("Node 3 ready to go"),
+
+    %% Waits enough for node 3 to sync but not for it to build a new chain
+    wait_for_height(Length, [node3], MiningTime * 3, Cfg),
+    ct:log("Node 3 on same height"),
+    Height3 = request(node3, [v2, 'block-by-height'], #{height => Length}, Cfg),
+    ct:log("Node 3 at height ~p: ~p~n", [Length, Height3]),
+
+    %% Checks node 3 is synchronized with nodes 1 and 2
+    ?assertEqual(Height1, Height3),
+    ok.
+
+%% When we stop and restart a node we will be able to read the blocks
+%% that we had in the chain before stopping: data is persistent.
+docker_keeps_data(Cfg) ->
+    Length = 20,
+    NodeStartupTime = proplists:get_value(node_startup_time, Cfg),
+
+    setup_nodes([?NODE1], Cfg),
+
+    start_node(node1, Cfg),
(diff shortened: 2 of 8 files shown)


rebar.config
```diff
@@ -123,7 +123,15 @@
...
                                         "data/epoch_config_schema.json"}]}
                            ]}
-                   ]}
+                   ]},
+            {system_test, [
+                {extra_src_dirs, ["system_test", "system_test/helpers"]},
+                {shell, [{apps, []}]},
+                {deps, [
+                    bbmustache,
+                    {hackney, ".*", {git, "git://github.com/benoitc/hackney.git", {branch, "master"}}}
```
eproxus: We should use a stable version of Hackney from Hex here, unless we have a policy to depend on Git repositories?

system_test/helpers/aest_docker.erl
```diff
@@ -0,0 +1,260 @@
...
+      data_dir => DataDir,
+      temp_dir => TempDir,
+      net_id => NetId
+    }.
+
+
+-spec stop(BackendState) -> ok
+    when BackendState :: backend_state().
+
+stop(BackendState) ->
```
eproxus: I would prefer a more compact spec declaration format:

```
-spec stop(backend_state()) -> ok.
stop(BackendState) ->
    ...
```

It has less duplication (of variable names) and doesn't feel like a separate function declaration.

system_test/helpers/aest_docker.erl
```diff
@@ -0,0 +1,260 @@
...
+    hostname := atom(),         % Hostname of the container running the node
+    exposed_ports := #{service_label() => pos_integer()},
+    local_ports := #{service_label() => pos_integer()}
+}.
+
+
+%=== GENERIC API FUNCTIONS =====================================================
+
+-spec start(Options) -> State
+    when Options :: #{test_id => test_uid(),
```
eproxus: `Options` should probably an `options()` type?

system_test/aest_sync_SUITE.erl
```diff
@@ -0,0 +1,190 @@
...
+    wait_for_height(0, [node2], NodeStartupTime, Cfg),
+    ct:log("Node 2 ready to go"),
+
+    %% we are fetching blocks crash now
+    stop_node(node1, NodeStopTime div 1000, Cfg),
+    Top2 = request(node2, [v2, 'top'], #{}, Cfg),
+    ct:log("Node 2 top: ~p~n", [Top2]),
+    Height = maps:get(height, Top2),
+    case Height >= Length of
+         true -> {skip, already_synced_when_crashed};
```
eproxus: Shouldn't this be more deterministic? In some cases we might not be testing anything at all?

system_test/aest_sync_SUITE.erl
```diff
@@ -0,0 +1,190 @@
...
+    peers   => [node1],
+    backend => aest_docker,
+    source  => {pull, "aetrnty/epoch:v0.8.0"}
+}).
+
+-define(NODE3, #{
+    name    => node3,
+    peers   => [node1],
+    backend => aest_docker,
+    source  => {pull, "aetrnty/epoch:local"}
```
lucafavatella: Do I understand correctly this is meant to be current master hence this can be [aeternity:dev](https://hub.docker.com/r/aeternity/epoch/tags/) ? ([Current pushes happen on aeternity Docker org](https://github.com/aeternity/epoch/blob/e65c7e4d838db29b52f371ab97c96eaae84bf717/.circleci/config.yml#L13).)

I ask as I am thinking of attempting to hook this CI after this is merged.

eproxus: I would point to an image that is built locally, even on CI, so you know which exact revisions you are testing. If a locally built Docker image called `aetrnty/epoch:dev` exists locally on CI, then that's a good choice.

system_test/helpers/aest_docker.erl
sylane: For small functions like this one it looks a bit more verbose, but for complicated specs it looks way cleaner that way (at least for me), and I prefer uniformity... In addition, I think that having the variable names usually helps (for more complex functions)...

system_test/helpers/aest_docker.erl
sylane: It could, but it is only for the stop function so you would have to create a `start_option()` type then...

system_test/aest_sync_SUITE.erl
```diff
@@ -0,0 +1,190 @@
...
+    Height = request(node2, [v2, 'block-by-height'], #{height => Length}, Cfg),
+    ct:log("Node 2 at height ~p: ~p~n", [Length, Height]),
+    start_node(node3, Cfg),
+    wait_for_height(0, [node3], NodeStartupTime, Cfg),
+    ct:log("Node 3 ready to go"),
+
+    %% get them with N blocks a second
+    timer:sleep((Length div BlocksPerSecond) * 1000),
+
+    wait_for_height(Length, [node3], 1000, Cfg),
```
lucafavatella: I replaced `aetrnty/epoch:local` with `aeternity/epoch:dev` then I got:
```
=== Location: [{aest_nodes,assert_expiration,361},
              {aest_nodes,wait_for_height,367},
              {aest_sync_SUITE,new_node_joins_network,99},
```

system_test/aest_sync_SUITE.erl
sylane: This question would be for @ThomasArts , I didn't look much at this test.

system_test/aest_sync_SUITE.erl
sylane: Maybe it is actually catching a real synchronization issue between `v0.8.0` and `dev`! It fails for me too, but if I set the 3 nodes to `dev` the test passes.

lucafavatella: I see in logs:
```
2018-03-26 16:04:19.825 [debug] <0.1111.0>@aeu_requests:ping:73 Got an error return: Reason = <<"Different genesis blocks">>
2018-03-26 16:04:19.825 [debug] <0.1111.0>@aec_sync:ping_peer:455 ping result (<<"http://node1_61A8E9F439C197A4:3013">>): {error,protocol_violation}
```
Maybe test is meant not to run v0.8.0 with master/`dev` but rather a previous version e.g. v0.8.0 itself? Or all :dev?

system_test/aest_sync_SUITE.erl
```diff
@@ -0,0 +1,190 @@
...
+    Top1 = request(node1, [v2, 'top'], #{}, Cfg),
+    ct:log("Node 1 top: ~p~n", [Top1]),
+    Height = request(node2, [v2, 'block-by-height'], #{height => Length}, Cfg),
+    ct:log("Node 2 at height ~p: ~p~n", [Length, Height]),
+    start_node(node3, Cfg),
+    wait_for_height(0, [node3], NodeStartupTime, Cfg),
+    ct:log("Node 3 ready to go"),
+
+    %% get them with N blocks a second
+    timer:sleep((Length div BlocksPerSecond) * 1000),
```
lucafavatella: Is this test actually meant to fail if sync is slower than a threshold, or is this meant to be large timeout?
If performance-related, maybe such tests shall be tagged as "performance" or similar and skipped unless explicitly enabled.

system_test/aest_sync_SUITE.erl
sylane: This test is meant to ensure that the last version is able to sync with a cluster of older versions. If it does not pass, either we broke the compatibility and we know it (then we need to update the images) or we broke the protocol (or the test is bad to begin with).

rebar.config
lucafavatella: At least versioned i.e. package version / repo tag / repo commit.

system_test/aest_sync_SUITE.erl
lucafavatella: I see - so recent master shall be ok. Thanks.

system_test/helpers/aest_docker.erl
eproxus: For me it's just one more point where code rot can start to creep in, where you forget to change one name but not the other. Making it more compact would save 2-4 lines per function. Just my 2 cents 🙂

system_test/aest_sync_SUITE.erl
sylane: It is not performance related, I should probably change the timing to something like the `docker_keeps_data` test where I measure the average time for mining a block. The test is about having a node with the last code join a cluster of older version nodes and check it synchronize and not start a new chain.

system_test/aest_sync_SUITE.erl
lucafavatella: > Maybe it is actually catching a real synchronization issue between v0.8.0 and dev! It fails for me too, but if I set the 3 nodes to dev the test passes.

You are right. Recent versions are not sync compatible so this shall - at least at the moment - have all the same version.

----

Chain consensus protocol in release 0.8.0:
https://github.com/aeternity/epoch/blob/v0.8.0/apps/aecore/include/blocks.hrl#L3-L5

Chain consensus protocol in release 0.9.0:
https://github.com/aeternity/epoch/blob/v0.9.0/apps/aecore/include/blocks.hrl#L3-L5

Chain consensus protocol in current master:
https://github.com/aeternity/epoch/blob/e65c7e4d838db29b52f371ab97c96eaae84bf717/apps/aecore/include/blocks.hrl#L3-L5

system_test/helpers/aest_docker.erl
sylane: It would save lines only for small functions, for bigger one it makes them cleaner (IMHO). About code rote, dialyzer would not detect variable names being wrong, but it would detect function typing getting wrong forcing you to change the spec...

system_test/aest_sync_SUITE.erl
sylane: I'll change the test to use `aetrnty/epoch:dev` and change the wiki to have people build the dev image from the local sources.

system_test/helpers/aest_docker.erl
```diff
@@ -0,0 +1,260 @@
...
+allocate_ports([], Acc) -> Acc;
+allocate_ports([Label | Labels], Acc) ->
+    allocate_ports(Labels, Acc#{Label => free_port()}).
+
+
+format(Fmt, Args) ->
+    iolist_to_binary(io_lib:format(Fmt, Args)).
+
+write_template(TemplateFile, OutputFile, Context) ->
+    Template = bbmustache:parse_file(TemplateFile),
```
lucafavatella: It looks like this handles unexistent file silently. Not blocker for PR though.

I put together patch for making this more strict https://github.com/lucafavatella/epoch/commit/df66cc92637cea88227ef1d1c436d4b882694891
After this PR is merged to master I plan to open PR proposing this (though feel free to include in this if your want).

sylane: I'll include it.

system_test/helpers/aest_docker.erl
hanssv: In this case even `-spec stop(BackendState :: backend_state() -> ok.` fits in one line. Breaking it over several lines is _not_ helping.

And *no*, unfortunately, dialyzer won't find squat 90% of the time in this situation... Dialyzer will just use the success type!?

eproxus: @sylane That was exactly my point of code rot, I'm arguing for removing the variable names. Keeping the variable name in two places with no compiler check in-between is not good. It's enough just to use the type spec.

system_test/helpers/aest_docker.erl
eproxus: `start_options()` is fine.

system_test/aest_sync_SUITE.erl
eproxus: For the system tests on CI we're now using `local`? So this should not be changed?

system_test/helpers/aest_docker.erl
eproxus: @hanssv I usually go by the rule that if `typer` doesn't find anything to complain about, I just stick with the deduced types.

system_test/aest_sync_SUITE.erl
sylane: Yes, this comment was before we ended up deciding to keep using `local`. Not changing it.

sylane:
@dincho : done

PR #899, opened 2018-03-26
PT-156154873: Bump cowboy to 2.2.2

fabiankrol:
[Upgrade to new cowboy version](https://www.pivotaltracker.com/story/show/156154873)


apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -120,20 +120,27 @@ init(WaitingPid) ->
...
-websocket_handle({pong, _}, _ConnState, Regs) ->
-    {ok, Regs};
+    {close, normal, Regs}.
+
+websocket_handle({pong, Nonce}, _ConnState, Regs) ->
+    case get_registered(waiting_nonce, Regs) of
+        [Nonce] ->
+            Regs1 = delete_registered(Nonce, waiting_nonce, Regs),
+            [WaitingPid] = get_registered(waiting_connected, Regs1),
+            inform_registered(WaitingPid, websocket, connected),
```
lucafavatella: I see this change makes the WS considered connected after additional ping-pong initiated by client. It is ok. I wonder whether this fixes a symptom or race condition?

I read cowboy supports ping:
> Note that ping and pong frames require no action from the handler as Cowboy will automatically reply to ping frames.

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -120,20 +120,27 @@ init(WaitingPid) ->
...
+            Regs2 = delete_registered(WaitingPid, waiting_connected, Regs1),
+            {ok, Regs2};
+        _ ->
```
lucafavatella: Could this open match be changed to match with `[]` or anyway stricter match? (Mostly for readability - otherwise reader needs to understand why the other match is the way it is.)

fabiankrol: 👍

apps/aehttp/test/aehttp_ws_test_utils.erl
fabiankrol: @lucafavatella It's bc of a change in cowboy websocket behaviour. In cowboy's 2.* version the handshake is done before the return on `websocket_init/1` hence the client's execute `onconnect/2` without any delays (`jobs:ask/1`).

PR #900, opened 2018-03-26
Hook system tests to CI

lucafavatella:
Depends on #898

The only functional thing I changed is moving the Docker Hub login after the build&tag, that I do not think is relevant (login is still before push).

TODO:
* [x] Rebase on master once #898 merged
* [ ] Fix failure in test case init `{'EXIT',{case_clause,{error,enoent}}`
* [ ] Test pushing image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -378,6 +378,10 @@ jobs:
       - *setup_remote_docker
       - *install_docker_client
       - *docker_system_tests
+      - store_test_results:
+          path: system_test/logs
+      - store_artifacts:
+          path: system_test/logs
 
   docker_system_tests_dev:
     <<: *container_config
@@ -386,6 +390,10 @@ jobs:
       - *setup_remote_docker
       - *install_docker_client
       - *docker_system_tests
+      - store_test_results:
+          path: system_test/logs
+      - store_artifacts:
+          path: system_test/logs
 
   docker_system_tests_tag:
     <<: *container_config
@@ -394,6 +402,10 @@ jobs:
       - *setup_remote_docker
       - *install_docker_client
       - *docker_system_tests
+      - store_test_results:
+          path: system_test/logs
+      - store_artifacts:
+          path: system_test/logs
 
   uat_chain_snapshot:
     <<: *container_config
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -311,6 +311,7 @@ jobs:
     <<: *container_config
     steps:
       - checkout
+      - *merge_checkout
       - *setup_remote_docker
       - *install_docker_client
       - run:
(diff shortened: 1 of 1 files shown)


lucafavatella:
I arrived to saving system test logs on CI.

Tests fail in test case init ([sample]( https://1864-110719633-gh.circle-artifacts.com/0/home/builder/epoch/system_test/logs/ct_run.epoch_ct@localhost.2018-03-26_22.15.31/extras.system_test.logs/run.2018-03-26_22.15.31/aest_sync_suite.new_node_joins_network.html)) with `{'EXIT',{case_clause,{error,enoent}}}`, not sure why.

.circleci/config.yml
```diff
@@ -306,44 +312,100 @@ jobs:
...
+  docker_push_tag:
+    <<: *container_config
+    steps:
+      - *setup_remote_docker
+      - *install_docker_client
+      - run:
+          name: Push Docker image to DockerHub
           command: |
             ~/bin/docker login -u $DOCKER_USER -p $DOCKER_PASS
-            ~/bin/docker build -t ${DOCKERHUB_REPO:?}:${CIRCLE_TAG:?} -t ${DOCKERHUB_REPO:?}:latest .
```
dincho: Last time I checked the built image was not cached/shared between jobs/steps. Can you prove this is working now?

.circleci/config.yml
```diff
@@ -497,12 +569,25 @@ workflows:
             branches:
               only: master
 
+      - docker_image_build_dev:
```
dincho: What's the reason having two separate jobs: `docker_image_build_dev` and `docker_system_tests_dev` as the neither is reused?

.circleci/config.yml
```diff
@@ -568,11 +653,28 @@ workflows:
             tags:
               only: /^v.*$/
 
+      - docker_image_build_tag:
```
dincho: Same question as above.

.circleci/config.yml
lucafavatella: The reason is that I am not sure about Circle CI constraints on required jobs. I.e. I remember from Circle CI docs that required jobs need same branch/tag filter, hence I kept it explicit for being on the safe side.

.circleci/config.yml
lucafavatella: Not checked so I expect same error. Tests do not arrive at stage of using tag yet (earlier enoent error).

.circleci/config.yml
dincho: > I remember from Circle CI docs that required jobs need same branch/tag filter

Yes, it is.

lucafavatella:
Rebased on master.

lucafavatella:
Tracked as https://www.pivotaltracker.com/story/show/156683319

PR #905, opened 2018-03-28
Relax CI from using PR code merged to master

lucafavatella:


lucafavatella:
@dincho Biggest issue (independent of this PR) that may justify removal of merge_checkout completely is that each job in the workflow may checkout a different commit.

dincho:
> @dincho Biggest issue (independent of this PR) that may justify removal of merge_checkout completely is that each job in the workflow may checkout a different commit.

Yes, I see this PR give additional chance for race conditions between jobs

PR #906, opened 2018-03-28
Separate business logic from http layer

velzevur:
* Refactored the serialization to be separated from the node internals

* Moved business logic to a separate module `aehttp_logic`

* Moved signed transaction-specific logic (that is going to be deprecated) to a `aehttp_int_tx_logic module`

Still TODO: rebase on top of noise protocol changes and apply them accordingly

### DIFF
--- a/apps/aecore/src/aec_peer_connection.erl
+++ b/apps/aecore/src/aec_peer_connection.erl
@@ -504,7 +504,7 @@ handle_get_header_by_hash_rsp(S, none, MsgObj) ->
 handle_get_header_by_hash_rsp(S, {get_header_by_hash, From}, MsgObj) ->
     case maps:get(<<"result">>, MsgObj, <<"error">>) of
         <<"ok">> ->
-            case aec_headers:deserialize_from_map(maps:get(<<"header">>, MsgObj)) of
+            case aehttp_api_parser:decode(header, maps:get(<<"header">>, MsgObj)) of
                 {ok, Header} ->
                     gen_server:reply(From, {ok, Header});
                 Err = {error, _} ->
@@ -541,7 +541,7 @@ get_header(height, N) ->
 get_header(Fun, Arg) ->
     case Fun(Arg) of
         {ok, Header} ->
-            {ok, HH} = aec_headers:serialize_to_map(Header),
+            HH = aehttp_api_parser:encode(header, Header),
             #{result => ok, header => HH};
         error ->
             #{result => error, reason => <<"Block not found">>}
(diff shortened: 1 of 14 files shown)


apps/aehttp/src/aehttp_int_tx_logic.erl
```diff
@@ -0,0 +1,232 @@
...
+                        expires_at => aeo_oracles:expires(O)}
+            end,
+            Oracles),
+    {ok, FmtOracles}.
+
+get_oracle_questions(OracleId, From, Max) ->
+    {ok, Queries} = aec_chain:get_open_oracle_queries(OracleId, From, Max),
+    FmtQueries =
+        lists:map(
+            fun(Q) -> #{<<"query_id">> => aeo_query:id(Q),
```
gorillainduction: Should this be a binary string?

apps/aehttp/src/aehttp_int_tx_logic.erl
```diff
@@ -0,0 +1,232 @@
...
+                query_id => DecodedQueryId,
+                response => Response,
+                fee      => Fee})
+          end).
+
+get_oracles(From, Max) ->
+    {ok, Oracles} = aec_chain:get_oracles(From, Max),
+    FmtOracles =
+        lists:map(
+            fun(O) -> #{<<"address">> => aeo_oracles:id(O),
```
gorillainduction: Should this be a binary string?

apps/aehttp/src/aehttp_int_tx_logic.erl
velzevur: It is encoded later on

velzevur:
Rebased on top of master and incorporated `Noise` changes

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -541,7 +541,7 @@ get_header(height, N) ->
 get_header(Fun, Arg) ->
     case Fun(Arg) of
         {ok, Header} ->
-            {ok, HH} = aec_headers:serialize_to_map(Header),
+            HH = aehttp_api_parser:encode(header, Header),
```
lucafavatella: I see this keeps base58. Ok.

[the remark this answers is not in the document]
velzevur: I think all HTTP objects should not be part of the aec_peer_connection at all. So there should be a follow up PR fixing this.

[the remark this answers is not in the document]
lucafavatella: So I understand idea is that complex items are serialized in distinct module, while hashes can get base58c encoded where they are. Makes sense.

PR #908, opened 2018-03-28
Fine-tune make clean following system test framework merge

lucafavatella:
Not sure where `_build/system_test+test` comes from.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -228,7 +228,7 @@ clean:
 	( cd apps/aering/test/contracts && $(MAKE) clean; )
 	( cd $(HTTP_APP) && $(MAKE) clean; )
 	@$(MAKE) multi-distclean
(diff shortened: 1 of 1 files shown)


dincho:
> Not sure where _build/system_test+test comes from.

https://github.com/aeternity/epoch/blob/master/rebar.config#L127 ?

eproxus:
You should clean both `_build/system_test` and `_build/system_test+test`. The plus folders are generated by Rebar when several profiles are active, and when running any test related task in Rebar the profile `test` is always active. We've created one extra profile called `system_test` to be able to have optional dependencies which are only installed and used when running system tests.

lucafavatella:
No way this can impact CI. I see no conflict with other major open PRs.

PR #912, opened 2018-03-28
Stop docker container using epoch stop command

sylane:


system_test/helpers/aest_docker_api.erl
```diff
@@ -193,29 +237,34 @@ docker_post(Path) -> docker_post(Path, #{}).
...
         {error, _Reason} = Error -> Error;
-        {ok, BodyJson} -> decode(BodyJson)
+        {ok, Body} -> decode(Body, Type)
     end.
 
-decode(<<>>) -> {ok, undefined};
-decode(JsonStr) ->
-    try jsx:decode(JsonStr, [{labels, attempt_atom}, return_maps]) of
+decode(<<>>, _) -> {ok, undefined};
+decode(Data, raw) -> Data;
```
lucafavatella: I am testing locally patch with `{ok, Data}` instead of `Data` here.

system_test/helpers/aest_docker.erl
```diff
@@ -187,7 +188,12 @@ stop_node(NodeState) -> stop_node(NodeState, #{}).
 
 -spec stop_node(node_state(), stop_node_options()) -> node_state().
 stop_node(#{container_id := ID, hostname := Name} = NodeState, Opts) ->
-    aest_docker_api:stop_container(ID, Opts),
+    Timeout = maps:get(soft_timeout, Opts, ?EPOCH_STOP_TIMEOUT),
+    aest_docker_api:exec(ID, ["/home/epoch/node/bin/epoch", "stop"]),
```
dincho: It might be good idea to factor this path (dir) out (e.g. as constant)  as it might change in future

lucafavatella: Or assume that workdir in Docker image is already the one from which bin/epoch can be run.

sylane:
I rebased on top of master and adapted the sync pubkey parametrization from @hanssv .

system_test/aest_sync_SUITE.erl
```diff
@@ -29,27 +29,38 @@
...
+    pubkey  => <<29,110,222,56,140,83,227,182,7,100,207,18,240,52,200,151,221,151,247,213,94,191,198,219,184,33,139,118,35,95,157,120>>,
+    peers   => [old_node1],
+    backend => aest_docker,
+    source  => {pull, "aeternity/epoch:local"}
+}).
+
+-define(STANDALONE_NODE, #{
+    name    => standalone_node,
+    pubkey  => <<29,110,222,56,140,83,227,182,7,100,207,18,240,52,200,151,221,151,247,213,94,191,198,219,184,33,139,118,35,95,157,120>>,
+    peers   => [],
```
lucafavatella: I ran successfully twice the system tests replacing all `:local` with `:v0.10.0`.
I get some distinct failures by replacing all `:loca` with `v0.10.1` - I suspect https://github.com/aeternity/epoch/commit/49eb621a51f669cbc33fe1f47d8d6872bfe7d9b4

lucafavatella: It looks like it does pick testnet because of no peers key. Maybe making peers as set as empty list helps. Last time I tried our yamerl lib (0.5.0) was not distinguishing between empty YAML sequence and empty YAML mapping, hence making using empty sequence not practical. Release 0.6.0 appears to have addressed that.

lucafavatella: > I get some distinct failures by replacing all :local with v0.10.1 - I suspect 49eb621

... that implies that system tests would fail in master if this PR 912 merged as is (though not hooked to CI yet).

I managed to get system tests to pass locally by:
* Tagging as `:local` v0.10.1 plus PR #942
* Amending system tests with following patch: 
```
diff --git a/system_test/aest_sync_SUITE_data/epoch.yaml.mustache b/system_test/aest_sync_SUITE_data/epoch.yaml.mustache
index e4d57536..76b66345 100644
--- a/system_test/aest_sync_SUITE_data/epoch.yaml.mustache
+++ b/system_test/aest_sync_SUITE_data/epoch.yaml.mustache
@@ -1,11 +1,9 @@
 {{#epoch_config}}
 ---
-{{#has_peers}}
-peers:
+peers:{{^peers}} []{{/peers}}
 {{#peers}}
     - {{peer}}
 {{/peers}}
-{{/has_peers}}

 sync:
     port: {{services.sync.port}}
diff --git a/system_test/helpers/aest_docker.erl b/system_test/helpers/aest_docker.erl
index a9512272..811a6b72 100644
--- a/system_test/helpers/aest_docker.erl
+++ b/system_test/helpers/aest_docker.erl
@@ -146,7 +146,6 @@ setup_node(Spec, BackendState) ->
     RootVars = #{
         hostname => Name,
         ext_addr => format("http://~s:~w/", [Hostname, ?EXT_HTTP_PORT]),
-        has_peers => length(PeerVars) > 0,
         peers => PeerVars,
         services => #{
             sync => #{port => ?EXT_SYNC_PORT},
```

Shall OLD_NODE1 and OLD_NODE1 be fixed to v0.10.1 for the v0.11.0 sprint? If yes, the patch above ^^^ is not sufficient.

lucafavatella:
PR #942 merged.

PR #922, opened 2018-03-29
Remove EXTERNAL_PEER_ADDRESS from docker docs

dincho:


docs/docker.md
```diff
@@ -58,16 +58,8 @@ docker exec epoch_node0 tail log/epoch.log
...
-
-Your public IP address will be automatically determined and used in `peer_address` configuration option.
-Make sure you have a working port forwarding setup on your firewall to be able to fully participate in the testnet.
-
-Docker environment variable `EXTERNAL_PEER_ADDRESS` may be used in case external peer address shall be changed e.g. different port mapping or different inbound IP:
-
-```bash
-docker run -d -p 3013:3013 -e EXTERNAL_PEER_ADDRESS=http://1.2.3.4:3013/ aeternity/epoch
-```
+The container will use [Minimal testnet config](docs/examples/epoch_testnet.yaml) unless other configuration file is specified (see below), it mostly relies on configuration defaults.
```
lucafavatella: Bad link.

lucafavatella:
I see there is an epoch_testnet.yaml change in concurrent PR https://github.com/aeternity/epoch/pull/920/files#diff-0a42c60e21a7009f94398e883d713489R7

docker/entrypoint.sh
```diff
@@ -1,15 +1,11 @@
 #!/bin/bash
 set -e
 
-# If not provided set to public facing IP address of this node
-EXTERNAL_PEER_ADDRESS="${EXTERNAL_PEER_ADDRESS:-http://$(curl -s https://api.ipify.org):3013/}"
```
lucafavatella: I see not needed any longer as peer receiving ping will use IP from inbound TCP connection

docs/docker.md
```diff
@@ -58,16 +58,8 @@ docker exec epoch_node0 tail log/epoch.log
...
-
-Your public IP address will be automatically determined and used in `peer_address` configuration option.
-Make sure you have a working port forwarding setup on your firewall to be able to fully participate in the testnet.
-
-Docker environment variable `EXTERNAL_PEER_ADDRESS` may be used in case external peer address shall be changed e.g. different port mapping or different inbound IP:
-
-```bash
-docker run -d -p 3013:3013 -e EXTERNAL_PEER_ADDRESS=http://1.2.3.4:3013/ aeternity/epoch
-```
+The container will use [Minimal testnet config](apps/aeutils/test/data/epoch_testnet.yaml) unless other configuration file is specified (see below), it mostly relies on configuration defaults.
```
lucafavatella: Broken links.

docs/docker.md
```diff
@@ -58,16 +58,8 @@ docker exec epoch_node0 tail log/epoch.log
...
-
-Your public IP address will be automatically determined and used in `peer_address` configuration option.
-Make sure you have a working port forwarding setup on your firewall to be able to fully participate in the testnet.
-
-Docker environment variable `EXTERNAL_PEER_ADDRESS` may be used in case external peer address shall be changed e.g. different port mapping or different inbound IP:
-
-```bash
-docker run -d -p 3013:3013 -e EXTERNAL_PEER_ADDRESS=http://1.2.3.4:3013/ aeternity/epoch
-```
+The container will use [Minimal testnet config](../apps/aeutils/test/data/epoch_testnet.yaml) unless other configuration file is specified (see below), it mostly relies on configuration defaults.
```
lucafavatella: I think this paragraph is unnecessarily too complex, as this doc shall assume that reader came from release notes so already knows where to find the config reference (wiki - linked from release notes).

lucafavatella: I think simply reinstating `#### External Peer Address` and state what the behaviour is (by default the exposed port is 3015, that needs to be the the same on the Internet facing firewall) and how to change it (change the config file - see instructions below) would suffice.

dincho: Addressed in the last push.

PR #927, opened 2018-04-01
Check db for hash, not node

gorillainduction:
Tighten up some typ checks to avoid similar bugs in the future

https://www.pivotaltracker.com/n/projects/2124891/stories/156411649


apps/aecore/src/aec_chain_state.erl
```diff
@@ -263,7 +263,7 @@ internal_insert(Node, Original, State0) ->
     maps:remove(currently_adding, StateOut).
 
 internal_insert_1(Node, Original, State) ->
```
uwiger: Perhaps `internal_insert_1(#node{} = Node, Original, State)`?

PR #934, opened 2018-04-04
Fix failing UATs

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/156497687)
The python was importing (now gone) Ping structure. This structure is not needed anymore and it is removed from `swagger.yaml` so the python module being imported was missing. The module itself is not being used anymore so the import is removed.

How to test locally:
1. rm -rf py/tests/swagger_client/
2. make swagger
3. make python-env
4. make multi-build
5. make python-uats <- not crashing :)


dincho:
```

PR #939, opened 2018-04-05
Fix single node deployments (using limit)

dincho:
e.g. `ansible-playbook --limit="ae-uat-epoch-n0" .... deploy.yml`

### DIFF
--- a/deployment/ansible/templates/epoch.yaml.j2
+++ b/deployment/ansible/templates/epoch.yaml.j2
@@ -4,10 +4,11 @@ sync:
 
 peers:
 {% for host in groups[hosts_group] | difference([inventory_hostname]) %}
+{% set public_ip = hostvars[host]['ansible_ssh_host']|default(hostvars[host]['ansible_host']) -%}
 {% if hostvars[host]['sync_public_key'] is defined %}
-    - "aenode://{{ hostvars[host]['sync_public_key'] }}@{{ hostvars[host]['public_ip'] }}:{{ config.sync.port }}"
+    - "aenode://{{ hostvars[host]['sync_public_key'] }}@{{ public_ip }}:{{ config.sync.port }}"
 {% else %}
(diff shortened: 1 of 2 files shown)


deployment/ansible/templates/epoch.yaml.j2
```diff
@@ -4,10 +4,11 @@ sync:
 
 peers:
 {% for host in groups[hosts_group] | difference([inventory_hostname]) %}
+{% set public_ip = hostvars[host]['ansible_ssh_host']|default(hostvars[host]['ansible_host']) -%}
```
lucafavatella: This looks a duplicate of public_ip var in deploy.yml?

lucafavatella:
I see this PR has not assigned 2+ reviewers so I am taking the liberty of marking as WIP for tracking not expected to be merged soon.

deployment/ansible/templates/epoch.yaml.j2
dincho: Yes, however if the playbook is run with `--limit` option the [`set_fact`](https://github.com/aeternity/epoch/pull/939/files#diff-2ad1947fe9b869ffd23a6a86872e65e0L30) is executed only on the limited set of hosts specified.

Having this in the template in the `for` loop makes sure it get's the correct IP of all the hosts in the group/env.

dincho: Moreover, we can't use the [`public_ip`](https://github.com/aeternity/epoch/pull/939/files#diff-2ad1947fe9b869ffd23a6a86872e65e0R42) var in the template, but `set_fact` that was used before.

dincho: As the PR title suggest, this is fixing deployments to limited number of hosts (e.g. 1 host), otherwise if you always deploy/limit to the full group/env, the old version (before this PR) still works as expected.

dincho:
> I see this PR has not assigned 2+ reviewers so I am taking the liberty of marking as WIP for tracking not expected to be merged soon.

I just missed to add explicit reviewers. Should be ready for review.

PR #940, opened 2018-04-05
Polish and add tests for channel_create_tx

zp-sd:
* Polish channel_create_tx
* Common Tests for channel_create_tx
* Require signatures of all signers during tx verification

### DIFF
--- a/apps/aechannel/test/aesc_txs_SUITE.erl
+++ b/apps/aechannel/test/aesc_txs_SUITE.erl
@@ -12,7 +12,8 @@
          groups/0]).
 
 %% test case exports
--export([create/1]).
+-export([create/1,
+         create_negative/1]).
 
 -include_lib("common_test/include/ct.hrl").
 
@@ -27,7 +28,8 @@ groups() ->
     [
      {all_tests, [sequence], [{group, transactions}]},
      {transactions, [sequence],
-      [create]
+      [create,
+       create_negative]
      }
     ].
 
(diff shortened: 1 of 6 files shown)


apps/aecore/src/aec_keys.erl
```diff
@@ -239,12 +239,12 @@ handle_call({sign, Tx}, _From,
             SignedTx = aetx_sign:sign(Tx, PrivKey, CryptoMap),
             {reply, {ok, SignedTx}, State}
     end;
-handle_call({verify, Sigs, Tx}, _From, #state{crypto = C} = State) ->
-    Signers = aetx:signers(Tx),
+handle_call({verify, Signatures, Tx}, _From, #state{crypto = C} = State) ->
```
uwiger: Ok.

We will also need an API function that checks e.g. wether the `initiator` alone has signed. This would be used in the off-chain logic, where e.g. the `initiator` signs the `channel_create_tx`, then passes it to the `participant`. The `participant` should verify that the `initiator` signature is in fact present and valid, before adding its own signature.

apps/aechannel/test/aesc_txs_SUITE.erl
```diff
@@ -62,4 +64,75 @@ create(Cfg) ->
     PubKey2 = aesc_channels:participant(Ch),
     0       = aesc_channels:sequence_number(Ch),
     true    = aesc_channels:is_active(Ch),
+    {PubKey1, PubKey2, S3}.
+
+create_negative(Cfg) ->
```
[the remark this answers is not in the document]
zp-sd: Yes, this may indeed make tests more readable. I will refactor at some point.

apps/aecore/src/aec_keys.erl
zp-sd: OK, So this should have different arity then? Take additional param, which is account's pubkey that should have signed?

velzevur:
Test failed with non-related intermittent error of
```

%%% aehttp_integration_SUITE ==> {{badmatch,{error,socket_closed_remotely}},
 [{aehttp_integration_SUITE,'-populate_block/1-fun-0-',1,
                            [{file,"/home/builder/epoch/_build/test/lib/aehttp/test/aehttp_integration_SUITE.erl"},
                             {line,3852}]},
  {lists,foreach,2,[{file,"lists.erl"},{line,1338}]},
  {aehttp_integration_SUITE,populate_block,1,
                            [{file,"/home/builder/epoch/_build/test/lib/aehttp/test/aehttp_integration_SUITE.erl"},
                             {line,3850}]},
  {aehttp_integration_SUITE,add_spend_txs,0,
                            [{file,"/home/builder/epoch/_build/test/lib/aehttp/test/aehttp_integration_SUITE.erl"},
                             {line,3846}]},
  {lists,foreach,2,[{file,"lists.erl"},{line,1338}]},
  {aehttp_integration_SUITE,generic_block_tx_index_test,1,
                            [{file,"/home/builder/epoch/_build/test/lib/aehttp/test/aehttp_integration_SUITE.erl"},
                             {line,2134}]},
  {test_server,run_test_case_eval1,6,[{file,"test_server.erl"},{line,1062}]}]}
```
I've restarted it

PR #941, opened 2018-04-05
Remove Swagger redundancy

velzevur:


apps/aecontract/src/aect_call_tx.erl
```diff
@@ -206,7 +206,7 @@ for_client(#contract_call_tx{caller     = CallerPubKey,
                              gas        = Gas,
                              gas_price  = GasPrice,
                              call_data  = CallData}) ->
-    #{<<"data_schema">> => <<"ContractQueryTxObject">>, % swagger schema name
+    #{<<"data_schema">> => <<"ContractCallTxObject">>, % swagger schema name
```
lucafavatella: Good catch.

config/swagger.yaml
```diff
@@ -2467,181 +2467,51 @@ definitions:
...
-          amount: integer
-          fee: integer
-          nonce: integer
-        required:
-        - sender
-        - recipient
-        - amount
-        - fee
-        - nonce
+      - $ref: '#/definitions/SpendTx'
```
lucafavatella: Re this redundancy - I see the def SpendTx (I did not check the other ones) is more refined. Maybe this redundancy was in place because swagger-codegen was not able to generate the Python client - or the generated Python client was misbehaving? I am concerned that UAT tests may not work.

velzevur: Ran tests locally and they've passed

PR #943, opened 2018-04-05
Add swagger data_schema-s

velzevur:


velzevur:
Rebased on top of recent `feature-state-channels` branch

PR #944, opened 2018-04-05
Relocate account balance and transactions endpoints to external

e-user:
Fixes [PT-156367049](https://www.pivotaltracker.com/story/show/156367049).

### DIFF
--- a/config/swagger.yaml
+++ b/config/swagger.yaml
@@ -1094,7 +1094,7 @@ paths:
   /account/balance/{account_pubkey}:
     get:
       tags:
-        - internal
+        - external
         - chain
       operationId: GetAccountBalance
       description: 'Get accounts''s balance'
@@ -1132,7 +1132,7 @@ paths:
   /account/txs/{account_pubkey}:
     get:
       tags:
-        - internal
+        - external
         - transactions
       operationId: GetAccountTransactions
       description: 'Get accounts''s transactions included in blocks in the longest chain'
--- a/apps/aehttp/priv/swagger.json
+++ b/apps/aehttp/priv/swagger.json
@@ -1298,7 +1298,7 @@
     },
     "/account/balance/{account_pubkey}" : {
       "get" : {
-        "tags" : [ "internal", "chain" ],
+        "tags" : [ "external", "chain" ],
         "description" : "Get accounts's balance",
         "operationId" : "GetAccountBalance",
         "produces" : [ "application/json" ],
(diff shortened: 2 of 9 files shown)


lucafavatella:
It appears it makes some tests fail - I am taking the liberty of marking as WIP.

velzevur:
Thanks for the PR, AK!
It looks good but you should add some changes in the tests, as well. They currently are still expecting to find those APIs on the internal port
please change [this](https://github.com/aeternity/epoch/blob/PT156367049-relocate-endpoints/apps/aehttp/test/aehttp_integration_SUITE.erl#L2968) to 
```
    Host = external_address(),
```
Same stands true for the [account transactions](https://github.com/aeternity/epoch/blob/PT156367049-relocate-endpoints/apps/aehttp/test/aehttp_integration_SUITE.erl#L2973) as well.

Also it looks like you will have to do the same change [here](https://github.com/aeternity/epoch/blob/PT156367049-relocate-endpoints/apps/aehttp/test/aehttp_integration_SUITE.erl#L3332) and [here](https://github.com/aeternity/epoch/blob/PT156367049-relocate-endpoints/apps/aehttp/test/aehttp_integration_SUITE.erl#L3336)

You'll also have to move the [account transactions test](https://github.com/aeternity/epoch/blob/PT156367049-relocate-endpoints/apps/aehttp/test/aehttp_integration_SUITE.erl#L261) and the [balances tests](https://github.com/aeternity/epoch/blob/PT156367049-relocate-endpoints/apps/aehttp/test/aehttp_integration_SUITE.erl#L296-L297) to the `external_endpoints` group.

You can run tests locally using `make test` target but this will run all tests. In order to run only the APIs' tests you can use
```
SUITE=apps/aehttp/test/aehttp_integration make test
```

config/swagger.yaml
```diff
@@ -1094,7 +1094,7 @@ paths:
   /account/balance/{account_pubkey}:
```
lucafavatella: I see param account_pubkey is required - ok.

config/swagger.yaml
```diff
@@ -1132,7 +1132,7 @@ paths:
   /account/txs/{account_pubkey}:
```
lucafavatella: I see param account_pubkey is required - ok.

lucafavatella:
> @lucafavatella wrt. API usage update: aeternity/protocol#59

Merged too https://github.com/aeternity/protocol/pull/59

PR #949, opened 2018-04-06
Adapt aesc_close_mutual_tx to new spec

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/156570763)

Since the [description of channel_close_mutual](https://github.com/aeternity/protocol/blob/channels/channels/ON-CHAIN.md#channel_close_mutual) has changed not to include initiatior and participant, this PR is addressing that.

Most notable change is getting peers in the channel: now they're fetched from the DB (whenever possible - from state trees)

### DIFF
--- a/apps/aechannel/src/aesc_close_mutual_tx.erl
+++ b/apps/aechannel/src/aesc_close_mutual_tx.erl
@@ -23,7 +23,8 @@
          serialization_template/1,
          serialize/1,
          deserialize/2,
-         for_client/1
+         for_client/1,
+         verifiable/1
         ]).
 
 %%%===================================================================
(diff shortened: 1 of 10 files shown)


velzevur:
Since the spes is updated, I'm adapting this PR to it

apps/aechannel/src/aesc_close_mutual_tx.erl
```diff
@@ -69,134 +72,194 @@ fee(#channel_close_mutual_tx{fee = Fee}) ->
...
+                    ok_or_error(ChannelAmt >= Fee, fee_too_big)
+                end,
+                fun() -> % check amounts
+                    ChannelAmt = aesc_channels:initiator_amount(Channel) +
+                                 aesc_channels:participant_amount(Channel),
+                    ok_or_error(ChannelAmt =:= InitiatorAmount + ResponderAmount + Fee,
+                                wrong_amounts)
+                end,
+                fun() -> % check TTL
+                    ok_or_error(TTL >= Height, ttl_expired)
```
zp-sd: Are we sure this is what TTL is meant to do? And Isn't that standard transaction TTL that is supposed to be implemented for all blockchain transaction? Then shouldn't that value be just a positive number and be relative to current Height? 

In specs I can see only
```
Peers will want to set the ttl to a value greater than zero, to avoid uncertainty. If the fees included are low and transaction pressure is high, then the transaction might end up being stuck in the mempool for an extended period.
```

In oracles/naming service TTL was specific to these services, and described for how long oracle query/response/name/whatnot will live. Standard transaction TTL was not introduced there yet (AFAIK).

velzevur: I was pointed to [This pivotal story](https://www.pivotaltracker.com/n/projects/2124891/stories/153897559)

zp-sd: Good 👍

uwiger:
My old approval seems to stick past later commits. Very well, I still have no objections.

PR #951, opened 2018-04-08
Improve supervision tree for connection servers

gorillainduction:
* Make a new supervisor, aec_connection_sup to collect peer/sync
  servers
* Make a new supervisor, aec_peer_connection_sup, to keep track of all
  aec_peer_connection servers
* Cut off the links between individual aec_peer_connections and
  aec_peers to avoid bringing down aec_peers (and all other
  connections) if one connection fails
* Cut off the links between individual aec_peer_connections and
  aec_peer_connection_listener to avoid bringing it down if one connection fails
* Protect aec_sync from calling a aec_peer_connection that is no
  longer around
*  Make aec_peers aware of the aec_peer_connection by monitoring them and remove failing connections.

https://www.pivotaltracker.com/n/projects/2124891/stories/156460780

### DIFF
--- a/apps/aecore/src/aec_peer_connection.erl
+++ b/apps/aecore/src/aec_peer_connection.erl
@@ -18,7 +18,7 @@
         , ping/1
         , send_block/2
         , send_tx/2
-
+        , start_link/2
         , stop/1
         ]).
 
@@ -55,14 +55,16 @@
 
 %% -- API --------------------------------------------------------------------
 
-connect(Options) ->
+connect(#{} = Options) ->
     lager:debug("New peer connection to ~p", [Options]),
-    gen_server:start_link(?MODULE, [Options#{ role => initiator }], []).
+    aec_peer_connection_sup:start_peer_connection(Options#{ role => initiator}).
 
 accept(TcpSock, Options) ->
     {ok, {Host, _Port}} = inet:peername(TcpSock),
-    case gen_server:start_link(?MODULE, [Options#{ tcp_sock => TcpSock, role => responder,
-                                                   host => inet_parse:ntoa(Host)}], []) of
+    Options1 = Options#{ tcp_sock => TcpSock
+                       , role => responder
+                       , host => inet_parse:ntoa(Host)},
+    case aec_peer_connection_sup:start_peer_connection(Options1) of
         {ok, Pid} ->
             gen_tcp:controlling_process(TcpSock, Pid),
             gen_server:cast(Pid, give_tcp_socket_ownership),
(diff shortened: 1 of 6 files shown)


apps/aecore/src/aec_peer_connection.erl
```diff
@@ -71,6 +73,10 @@ accept(TcpSock, Options) ->
             Res
     end.
 
+start_link([{ext_sync_port, Port}], Opts) ->
```
hanssv: Ah, it is a `simple_one_for_one` - I had some trouble seeing how this function was called with 2 arguments from start_child...

lucafavatella:
I see WIP in the title - I took the liberty of labelling as WIP (as tag/label) so to ease filtering.

apps/aecore/src/aec_peer_connection_listener.erl
```diff
@@ -6,57 +7,103 @@
...
+-record(state, { lsock
+               , opts
+               , acceptor
+               }).
+
+%%====================================================================
+%% API functions
+%%====================================================================
+
+start_link([{sync_port, _}, {listen_address, _}] = Args) ->
```
sylane: I am a bit confused about this use of proplists. The order of the tuple is not supposed to be relevant but with these matches it is. If the order is enforced I wouldn't go with a proplist, but a normal list because I think this is confusing. If you want to match why not use a map ?

sylane:
Looks good beside the pattern matching on proplists. I would either remove the pattern matching in favor of `proplists:lookup` or use maps.

apps/aecore/src/aec_peer_connection_listener.erl
gorillainduction: It is just me being (over) careful. Since the way gen_server (and friends) are passing arguments, I want to make sure we don't accidentally mix the args up. It is not a proplist, and should not be treated as such. It is a way of ensuring that the order of args is preserved. The call points are static (in the supervisor).

gorillainduction: I don't like using maps or records in cases like this either, since I want to ensure full match.

But I agree that it looks a bit strange ;-)

gorillainduction:
Rebased and addressed the somewhat strange init args.

apps/aecore/src/aecore_sup.erl
```diff
@@ -1,3 +1,30 @@
...
+%%%           |
+%%%           ------------------------------------------------------------
+%%%           |         |         |           |            |             |
+%%%           |   aec_metrics  aec_keys  aec_tx_pool aec_conductor aec_subscribe
+%%%           |
+%%%   aec_connection_sup  (one_for_all)
+%%%           |
+%%%           -------------------------------------------------------
+%%%           |                                          |          |
+%%%   aec_peer_connection_sup (simple_one_for_one)   aec_peers   aec_sync
```
ThomasArts: Aec_connection_listener is not in this diagram?

PR #952, opened 2018-04-08
Feature state channels uw channel setup

uwiger:
This commit includes a test suite case that negotiates an off-chain channel, including committing to the chain and ensuring a `minimum_depth` number of blocks on top.

The key length was temporarily increased in the off-chain messages. This needs to be changed to use 32bytes (compressed) `Curve25519` key representation.

Next up: adding channel update messages for the continuous use of the channel. Also, documenting the configuration options of the channel setup.

The test case setup: The test process `aesc_fsm_SUITE:create_channel/1` starts two nodes, using the miner pubkeys of each as peers. However, both ends of the channel are set up on `dev1`, and the test case process sets itself as a (signing) client for both ends.

### DIFF
--- a/apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
+++ b/apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
@@ -1,7 +1,7 @@
 -module(aesc_fsm_min_depth_watcher).
 -behaviour(gen_server).
 
--export([start_link/2]).
+-export([start_link/3]).
 
 -export([init/1,
          handle_call/3,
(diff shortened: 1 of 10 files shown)


apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -0,0 +1,193 @@
...
+    {P1, _} = await_signing_request(funding_created, P),
+    await_funding_signed_i(I, P1).
+
+await_funding_signed_i(I, P) ->
+    receive_info(I, funding_signed),
+    await_funding_locked(I, P).
+
+await_funding_locked(I, P) ->
+    ct:log("mining blocks on dev1 for minimum depth", []),
+    aecore_suite_utils:mine_blocks(aecore_suite_utils:node_name(dev1), 4),
```
velzevur: I see this 4 = 1 + 3, where 
1 = include the multisigned tx on the chain
3 = minimum_depth

uwiger: We need to mine at least one block (hopefully just one, as there's only one tx) in order to get the tx on the chain, then `minimum_depth` blocks after that.

uwiger: so, yes :)

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -0,0 +1,147 @@
...
+               tx_hash := TxHash, min_depth := MinDepth} = St) ->
+    lager:debug("check_status(~p)", [Fsm]),
+    case min_depth_achieved(TxHash, MinDepth) of
+        true ->
+            lager:debug("min_depth achieved", []),
+            aesc_fsm:own_funding_locked(Fsm, ChanId),
+            {stop, normal, St};
+        _ ->
+            lager:debug("min_depth not yet achieved", []),
+            {noreply, St, 5000}
```
velzevur: The timeout intervals are too strict, please remove them.

uwiger: That timeout is just for debugging purposes. There should not be a timeout at all, eventually. Also, the gen_server/gen_statem debug tracing shall be removed.

apps/aechannel/src/aesc_session_noise.erl
```diff
@@ -134,10 +162,10 @@ tcp_opts(_Op, Opts) ->
 
 
 noise_defaults() ->
-    [{protocol, <<"Noise_XK_25519_ChaChaPoly_BLAKE2b">>}].
+    [{noise, <<"Noise_XK_25519_ChaChaPoly_BLAKE2b">>}].
 
 tcp_defaults() ->
-    [{active, once},
+    [{active, true},
```
hanssv: Probably not an issue here, but I've had trouble when I was delegating the socket and messages arrived at the wrong process (too early). Starting with active false and activate once setup is done is an alternative.

uwiger: Yes, `{active, once}` clearly didn't work. I decided that I don't care what `enoise` wants during the setup, and can set `{active, once}` after successful noise negotiation. I'll probably change to `{active, false}` in a future PR then.

PR #959, opened 2018-04-10
Adapt deposit and withdraw txs to new specs

zp-sd:
* Add common tests


velzevur:
I will address TTLs and [from's type](https://github.com/aeternity/epoch/blob/aba11f02099c62e9c5d601d4ca4c5668afbdd390/config/swagger.yaml#L2558) in a subsequent PR

PR #956, opened 2018-04-10
Channels: adapt aesc_create_tx to new spec

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/156646077)
Latest channel create tx as described in the [protocol](https://github.com/aeternity/protocol/blob/c02ac9f8ac8afa7a898c20486b114b6d6c083d01/channels/ON-CHAIN.md#channel_create)

PR #962, opened 2018-04-11
Channels txs: adapt close solo slash settle txs to new spec

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/156652453)

This PR is based on [PR 956](https://github.com/aeternity/epoch/pull/956) and must be merged after it.

TODO: add some tests

### DIFF
--- a/apps/aechannel/src/aesc_channels.erl
+++ b/apps/aechannel/src/aesc_channels.erl
@@ -30,7 +30,8 @@
          initiator_amount/1,
          participant_amount/1,
          channel_reserve/1,
-         sequence_number/1]).
+         sequence_number/1,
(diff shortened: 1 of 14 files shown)


apps/aechannel/src/aesc_channels.erl
```diff
@@ -28,8 +28,10 @@
          participant/1,
          total_amount/1,
          initiator_amount/1,
+         participant_amount/1,
          channel_reserve/1,
-         sequence_number/1]).
+         sequence_number/1,
```
uwiger: Are we going with `sequence_number`?
I've just been changing it to `round` in my branch...

PR #964, opened 2018-04-11
Generate nodes peer keys automatically for system tests

sylane:


system_test/aest_sync_SUITE_data/epoch.yaml.mustache
```diff
@@ -21,6 +21,7 @@ websocket:
         port: {{services.int_ws.port}}
 
 keys:
+    # If the password is changed, reflect the change in aest_docker.erl
     password: "top secret"
```
lucafavatella: Can this be passed as root var in template? So at least it is only in aest_docker.erl

system_test/helpers/aest_nodes.erl
```diff
@@ -66,7 +66,11 @@
 %% When `backend` is `aest_docker`:
 
     % The source of the docker image
-    source  := {pull, binary() | string()}
+    source  := {pull, binary() | string()},
+    % Public/private peer key can be specified explicity for the node.
+    % Both are required and will be saved, overriding any present keys.
+    pubkey := binary(),
```
lucafavatella: Shall this be => ?

system_test/helpers/aest_nodes.erl
```diff
@@ -66,7 +66,11 @@
...
+    privkey := binary()
```
lucafavatella: Shall this be => ?

system_test/helpers/aest_nodes.erl
ThomasArts: Indeed: http://erlang.org/doc/reference_manual/typespec.html

system_test/aest_sync_SUITE_data/epoch.yaml.mustache
sylane: We could do that.

system_test/helpers/aest_nodes.erl
sylane: Yes indeed, my bad.

apps/aecore/src/aec_keys.erl
```diff
@@ -25,7 +25,10 @@
          terminate/2, code_change/3]).
 
 -export([sign/1, pubkey/0, delete/0,
-         wait_for_pubkey/0]).
+         wait_for_pubkey/0,
+         setup_peer_keys/2,
```
hanssv: export for test only?

system_test/helpers/aest_docker.erl
```diff
@@ -101,6 +105,29 @@ stop(BackendState) ->
...
+-spec prepare_spec(node_spec(), backend_state()) -> node_spec().
+prepare_spec(#{pubkey := PubKey, privkey := PrivKey} = Spec, BackendState) ->
+    #{data_dir := DataDir} = BackendState,
+    #{name := Name} = Spec,
+    KeysDir = keys_dir(DataDir, Name),
+    Password = ?PEER_KEYS_PASSWORD,
+    {PubFile, PrivFile} = aec_keys:peer_key_filenames(KeysDir),
+    file:delete(PubFile),
+    file:delete(PrivFile),
+    aec_keys:save_peer_keys(Password, KeysDir, PubKey, PrivKey),
```
ThomasArts: I don't understand this. Would you not need to save these keys on the container? It looks like you are saving them on the local disk, since aest_nodes calls this function from outside the container.

Puzzled

sylane: Because later on this directory is mounted inside the container.

PR #965, opened 2018-04-11
Add network split recovery system test

sylane:
After rebasing this branch on top of master, the test `new_node_joins_network` do not pass anymore with `aeternity/epoch:v0.10.1`. Is it expected or is it a regression that needs to be fixed ?

### DIFF
--- a/system_test/aest_sync_SUITE.erl
+++ b/system_test/aest_sync_SUITE.erl
@@ -286,6 +286,11 @@ crash_and_continue_sync(Cfg) ->
             ?assertEqual(B1, B2)
     end.
 
+%% Test that two disconnected clusters of nodes are able to recover and merge
+%% there chain when connected back together.
+%% It tests both case of the chain being started from scratch in different
+%% network partitions, and that the network is partitiioned when the chain
+%% is already shared.
 net_split_recovery(Cfg) ->
     Length = 10,
 
--- a/system_test/aest_sync_SUITE.erl
+++ b/system_test/aest_sync_SUITE.erl
@@ -64,6 +64,9 @@
 }).
 
 
+%% By default, this node only connects to network `net1` even though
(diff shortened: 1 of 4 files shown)


sylane:
This was before the key generation PR, I will remove them when rebasing.

system_test/aest_sync_SUITE.erl
```diff
@@ -248,6 +286,89 @@ crash_and_continue_sync(Cfg) ->
...
+net_split_recovery(Cfg) ->
+    Length = 10,
+
+    setup_nodes([?NET1_NODE1, ?NET1_NODE2, ?NET2_NODE1, ?NET2_NODE2], Cfg),
+    start_node(net1_node1, Cfg),
+    start_node(net1_node2, Cfg),
+    start_node(net2_node1, Cfg),
+    start_node(net2_node2, Cfg),
+
+    %% Starts with a net split
```
ThomasArts: Since net1_node1 is connected to net2_node1 I don't see why this is building two different chains. In fact it starts with a joined network, as far as I can see

system_test/aest_sync_SUITE.erl
```diff
@@ -61,12 +63,48 @@
     source  => {pull, "aeternity/epoch:local"}
 }).
 
+
+-define(NET1_NODE1, #{
```
ThomasArts: Is it such that if you mention [net1] in this setup, then it won't be connected to other networks, but you may still have nodes to that network? A comment would be in place.
This is different semantics from when you mention a node that is not defined, because then you get an error message.

system_test/aest_sync_SUITE.erl
sylane: `net1_node1` as `net2_node2` as a peer BUT the container is not connected to the `net2` network. So it will fail to connect to the other node and end-up forking the chain.

PR #966, opened 2018-04-11
PT156692287 don't Multiplex in peer_connection

hanssv:
Adding some logic that was missing and cleaning up a bit in the meantime...

### DIFF
--- a/apps/aecore/src/aec_peer_connection.erl
+++ b/apps/aecore/src/aec_peer_connection.erl
@@ -219,7 +219,7 @@ handle_info({connected, Pid, {ok, TcpSock}}, S0 = #{ status := {connecting, Pid}
                 ok ->
                     {noreply, S#{ status => {connected, ESock} }};
                 {error, _} ->
-                    lager:debug("Dropping unnecessary connection to ", [maps:get(host, S)]),
+                    lager:debug("Dropping unnecessary connection to ~p", [maps:get(host, S)]),
                     enoise:close(ESock),
                     {stop, normal, S}
             end;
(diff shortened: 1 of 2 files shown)


apps/aecore/src/aec_peer_connection.erl
```diff
@@ -378,6 +370,35 @@ handle_msg(S, MsgType, true, Request, {ok, {MsgType, _Vsn, Msg}}) ->
         block   -> handle_get_block_rsp(S, Request, Msg)
     end.
 
+handle_request(S, {Request, Args}, From) ->
+    case get_request(S, Request) of
+        none ->
+            handle_request(S, Request, Args, From);
+        {Request, _From} ->
+            {reply, {error, request_already_in_progress}, S}
```
lucafavatella: Are all requests in this state guaranteed to eventually be removed e.g. in case of process crash or slow other TCP endpoint?

apps/aecore/src/aec_peer_messages.erl
```diff
@@ -26,11 +29,11 @@
 
 serialize_response(Type, {ok, Object}) ->
     SerObj = serialize(Type, Object),
-    serialize(response, #{ result => <<"ok">>,
+    serialize(response, #{ result => true,
```
velzevur: I see this is adaptation according to the [protocol spec](https://github.com/aeternity/protocol/blob/cfac4d31691f1358c00ce28dfb93ea590fcd7bd9/sync/p2p_messages.md#msg_p2p_response). Great!

PR #967, opened 2018-04-11
Plant support for hard forks

lucafavatella:
Supersedes https://github.com/aeternity/epoch/pull/960

This PR aims at finishing https://www.pivotaltracker.com/story/show/153778514

* Block validation old of chain is system tested
* Mining coinbase-only with old and new protocol is system tested though with mocked PoW
  * Block validation of blocks so mined is system tested though with mocked PoW

TODO:
* [x] Add smoke test for PoW at protocol switch
* [x] Test against latest UAT/testnet DB snapshot (i.e. [this](https://7305-99802036-gh.circle-artifacts.com/0/tmp/chain_snapshots/54.202.11.197/tmp/mnesia_ip-172-31-36-149_db_backup_1523491249.gz) of height 3455)
* [x] Mention in release process in wiki that before tagging height of new protocol needs to be changed
  * See recent changes in https://github.com/aeternity/epoch/wiki/Release-process
* [x] Fine tune height of switch to new protocol
* [ ] Refine release notes re this change
  * May do in separate PR
* [X] Bump deployment DB dir? Cc DevOps @acidsennui 
  * Done as db42

PR #970, opened 2018-04-12
Release preparation for 0.11.0

lucafavatella:
Depends on #967

TODO:
* [x] Smoke test configuration file in doc


lucafavatella:
I build local package considering also master and tested config in instructions and node started and mined blocks. It also attempted to establish some peer connections though connection handshake failed as expected.

PR #971, opened 2018-04-13
Properly deserialize older blocks

hanssv:
If we want to be backwards compatible we should understand old blocks...

PR #973, opened 2018-04-13
Fix backwards compatibility for block serialization

hanssv:


lucafavatella:
Earlier I saw CI completed successfully in #971 for exact same changes so merging.

PR #974, opened 2018-04-13
Make hard forking system test stricter

lucafavatella:


lucafavatella:
Merging - I do not see potential substantial conflicts with #978

PR #975, opened 2018-04-13
Pt 156688109 improve forks

gorillainduction:
Large rewrite of aec_chain_state: Remove ability to store headers in chain db
- No more headers as chain nodes in aec_chain_state
- No more difference between top_header_hash and top_block_hash
- When adding a block, the previous block either has state, or it is not connected to genesis.
- Use fork_id to find fork points and common ancestors
- Adapt to new chain interface
- Make sure adding a block is atomic with the state changes

https://www.pivotaltracker.com/n/projects/2124891/stories/156688109

### DIFF
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -1,3 +1,4 @@
+%%% -*- erlang-indent-level:4; indent-tabs-mode: nil -*-
 %%%-------------------------------------------------------------------
 %%% @copyright (C) 2017, Aeternity Anstalt
 %%% @doc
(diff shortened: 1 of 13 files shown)


gorillainduction:
Rebased and squashed commits.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -193,150 +170,64 @@ export_header(#node{header = Header}) ->
...
+            %% To preserve the invariants of the chain,
+            %% Only add the block if we can do the whole
+            %% transitive operation (i.e., calculate all the state
+            %% trees, and update the pointers)
+            Fun = fun() ->
+                          State = new_state_from_persistence(),
+                          State1 = State#{ currently_adding => hash(Node)},
+                          assert_not_new_genesis(Node, State1),
+                          ok = db_put_node(Block, hash(Node)),
+                          State2 = update_state_tree(Node, maybe_add_genesis_hash(State1, Node)),
```
lucafavatella: This addressed https://www.pivotaltracker.com/story/show/155851463

PR #976, opened 2018-04-13
Update config.yml

acidsennui:


lucafavatella:
Not needed any longer as deployment of 0.11.0 ok.
@acidsennui Thanks for the proposed changes though, may be reused for next release. Tracked as https://www.pivotaltracker.com/story/show/156798908

PR #979, opened 2018-04-16
PT-156728559 Cleanup requests, PeerId and convert hostname to binary

hanssv:
[x] When we drop the TCP connection error (and remove) all outstanding requests
[x] `aec_peers` had hostname and port in its PeerId - only PubKey is necessary
[x] `inet:parse/1` returns a `string()` it should be converted to a binary.

### DIFF
--- a/apps/aecore/src/aec_peers.erl
+++ b/apps/aecore/src/aec_peers.erl
@@ -35,7 +35,6 @@
          get_random/2,
          parse_peer_address/1,
          encode_peer_address/1,
-         set_local_peer_info/1,
          get_local_peer_info/0]).
 
 %% API only used in aec_sync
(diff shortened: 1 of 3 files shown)


apps/aecore/src/aec_peers.erl
```diff
@@ -149,7 +147,7 @@ remove(Peer) ->
 %%------------------------------------------------------------------------------
 %% Get list of all peers. The list may be big. Use with caution.
 %% Consider using get_random instead.
-%%------------------------------------------------------------------------------
+%------------------------------------------------------------------------------
```
gorillainduction: Cosmetic bug ;-)

uwiger: Why?

hanssv: New keyboard, I sometimes the wrong keys which gives interesting effects in vi/vim...

hanssv: New keyboard, I sometimes the wrong keys which gives interesting effects in vi/vim...

PR #983, opened 2018-04-16
Cleanup aec_conductor workers better and don't log lager:error

hanssv:
We should flush messages from workers that are pre-empted.

### DIFF
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -392,7 +392,7 @@ handle_worker_reply(Pid, Reply, State) ->
                                 },
             worker_reply(Tag, Reply, State1);
         error ->
-            epoch_mining:error("Got unsolicited worker reply: ~p",
+            epoch_mining:info("Got unsolicited worker reply: ~p",
                                [{Pid, Reply}]),
             State
     end.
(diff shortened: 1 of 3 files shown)


apps/aecore/src/aec_conductor.erl
```diff
@@ -385,14 +385,14 @@ handle_worker_reply(Pid, Reply, State) ->
...
-            cleanup_after_worker(Info),
+            cleanup_after_worker(Pid, Info),
             Tag = Info#worker_info.tag,
             State1 = State#state{workers = orddict:erase(Pid, Workers),
                                  blocked_tags = ordsets:del_element(Tag, Blocked)
                                 },
             worker_reply(Tag, Reply, State1);
         error ->
-            epoch_mining:error("Got unsolicited worker reply: ~p",
+            epoch_mining:info("Got unsolicited worker reply: ~p",
```
dincho: finally 👯‍♂️ 🍌 

tracked in https://www.pivotaltracker.com/story/show/155776152

hanssv: Yeah, sorry...

apps/aecore/src/aec_conductor.erl
```diff
@@ -443,18 +443,21 @@ update_chain_metrics(Block) ->
...
 
 
-cleanup_after_worker(Info) ->
+cleanup_after_worker(Pid, Info) ->
     case Info#worker_info.timer of
         no_timer -> ok;
         {t, TRef} -> timer:cancel(TRef)
     end,
     demonitor(Info#worker_info.mon, [flush]),
+    %% Flush messages from this worker.
```
gorillainduction: The flush is unnecessary when we have received the answer and there is a race for the shutdown in `kill_worker/3`. Better place this flush after the `exit(Pid, shutdown)` instead.

hanssv:
Moved the flush to the right place and also address the error-logging mentioned in https://www.pivotaltracker.com/n/projects/2124891/stories/155776152

hanssv:
And resolved a useless whitespace conflict...

PR #985, opened 2018-04-16
Pt 156370750 deploy sync keys

acidsennui:
PR including this: https://github.com/aeternity/epoch/pull/939 for `public_ip`
PT: https://www.pivotaltracker.com/story/show/156370750


deployment/ansible/files/peer_keys/13.250.162.250.pub
```diff
@@ -0,0 +1,7 @@
+$ANSIBLE_VAULT;1.1;AES256
```
dincho: we don't need to encrypt this - the pubkeys (it's just additional noise in the playbooks)

deployment/ansible/host_vars/13.250.162.250/vars.yml
```diff
@@ -0,0 +1,2 @@
+public_peer_key: files/peer_keys/{{ public_ip }}.pub
```
dincho: hmm, looking on this it's a lot of duplication, let's rework it:

create new group in a static inventory file `inventory/static.ini`:
```
[uat_seed_nodes]
52.10.46.160
18.195.109.60
13.250.162.250
31.13.249.70
```

Then create `group_vars/uat_seed_nodes/vars.yml` with the content of one of those 4 (duplicating) files.

deployment/ansible/deploy.yml
```diff
@@ -168,6 +166,24 @@
         when: genesis_accounts is defined
         tags: [config]
 
+      - name: "Copy public peer key"
```
dincho: Please use similar (same/copied) [task that deploys the chain keys](https://github.com/aeternity/epoch/pull/985/files#diff-2ad1947fe9b869ffd23a6a86872e65e0R146)

- you don't need user/group as it's already executing with the correct user
- mode should be `0600` instead of `0644`, otherwise it's visible to all users
- can be combined in one task thanks to `with_items:` - see above

deployment/ansible/files/peer_keys/13.250.162.250.pub
acidsennui: true...

deployment/ansible/host_vars/13.250.162.250/vars.yml
acidsennui: ok

deployment/ansible/deploy.yml
acidsennui: 644 was there on nodes with current files, thats why put 633

dincho: I guess we have to change the default mask at some point (additional ticket .. )

acidsennui:
@dincho looks like it is working now ;]

deployment/ansible/deploy.yml
```diff
@@ -166,6 +166,14 @@
         when: genesis_accounts is defined
         tags: [config]
 
+       - name: "Copy node keys to {{ project_root }}/{{ config.keypair.dir }}"
```
dincho: Could you please update the title as well, it's the same as the chain keys

`"Copy node peer keys to {{ project_root }}/{{ config.keypair.dir }}"` should do

deployment/ansible/deploy.yml
```diff
@@ -166,6 +166,14 @@
...
+         copy:
+           src: "{{ item }}"
+           dest: "{{ project_root }}/{{ config.keypair.dir }}"
+           mode: 0600
+         with_items: "{{ key_pairs }}"
+         when: key_pairs is defined
```
dincho: the variables is off, the chain keys are using `keypairs` ... what about `peer_keypairs`

dincho:
> @dincho looks like it is working now ;]

Looks OK to me too, however I have couple of debt comments

deployment/ansible/deploy.yml
dincho: also missing the `config` tag

acidsennui:
@dincho updated

deployment/ansible/deploy.yml
```diff
@@ -39,7 +39,7 @@
     project_root: "{{ ansible_env.HOME }}/node"
     genesis_accounts_path: "{{ project_root }}/data/aecore/.genesis/accounts.json"
     is_remote_package: false
-    public_ip: "{{ ansible_ssh_host|default(ansible_host) }}"
+    public_ip: "{{ ansible_ssh_host|default(ansible_host)|default(inventory_hostname) }}"
```
lucafavatella: It is not clear to me why this change needed. (I see commit message "fix public ip for static inventory" though unclear what issue was.)

deployment/ansible/inventory/static
```diff
@@ -0,0 +1,5 @@
+[tag_seed_uat]
+52.10.46.160
```
lucafavatella: I understand this file labels these 4 hosts with a group that then defines var `peer_keypairs`. Who guarantees that Ansible actually picks up this label for these 4 hosts? E.g. is it possible inventory is iterated over by host name rather than IP hence making this label not to be picked up (hence keys not instated if hosts created from scratch)?

deployment/ansible/deploy.yml
dincho: @lucafavatella `ansible_ssh_host` and  `ansible_host` are set by the dynamic inventories of OSS/AWS. If you use the static inventory alone (provided in this PR), we need `inventory_hostname`, although the automation does not use it, I often use it locally.

deployment/ansible/inventory/static
dincho: Good question, didn't found an explicit explanation in the docs, but I guess one can create inventory groups either by hostname or IP addresses.

This is tested already. Both in combination with dynamic inventory and this inventory alone.

PR #986, opened 2018-04-17
Send chat notification on uat_tests fail

dincho:
Example failing job: https://circleci.com/gh/aeternity/epoch/7812
Example full workflow: https://circleci.com/workflow-run/2593efdc-8b97-469d-a9f1-c066a2a7e966
Example failing `build` job: https://circleci.com/workflow-run/043a9e74-268a-4ac5-91a3-2973d8d5090a 
Example notification can be seen in the corresponding channel.

PT: https://www.pivotaltracker.com/story/show/156197975

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -174,7 +174,6 @@ jobs:
   test:
     <<: *container_config
     steps:
-      - run: "false"
       - checkout
       - *restore_rebar_cache
       - *restore_build_cache
@@ -192,7 +191,6 @@ jobs:
   eunit:
     <<: *container_config
     steps:
-      - run: "false"
       - checkout
       - *restore_rebar_cache
       - *restore_build_cache
@@ -217,7 +215,6 @@ jobs:
   static_analysis:
     <<: *container_config
     steps:
-      - run: "false"
       - checkout
       - *restore_rebar_cache
       - *restore_build_cache
@@ -243,7 +240,6 @@ jobs:
   linux_package:
     <<: *container_config
     steps:
-      - run: "false"
       - checkout
       - *set_package_path
       - *build_package
@@ -274,7 +270,6 @@ jobs:
       xcode: "9.0"
     working_directory: ~/epoch
     steps:
-      - run: "false"
       - checkout
       - run:
           name: Install required tools
@@ -301,7 +296,6 @@ jobs:
   uat_tests:
     <<: *container_config
     steps:
-      - run: "false"
       - checkout
       - *restore_rebar_cache
       - *restore_build_cache
@@ -355,7 +349,6 @@ jobs:
   docker_push_dev:
     <<: *container_config
     steps:
-      - run: "false"
       - checkout
       - *setup_remote_docker
       - *install_docker_client
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,6 +126,26 @@ references:
           owner=${CIRCLE_PROJECT_USERNAME} repo=${CIRCLE_PROJECT_REPONAME} \
           tag=${CIRCLE_TAG} ASSETS=${PACKAGES_DIR:?}/*
 
+  fail_notification: &fail_notification
+    run:
+      name: Send failure notification
+      environment:
+        HOOK_TEMPLATE: |
+          {
+            "text": "CircleCI job **%s** failed on branch **%s**",
+            "attachments": [
+              {
+                "title": "Build Link",
+                "title_link": "%s",
+                "color": "#FAD6D6"
+              }
+            ]
+          }
+      command: |
+        HOOK_DATA=$(printf "$HOOK_TEMPLATE" "${CIRCLE_JOB:?}" "${CIRCLE_BRANCH:?}" "${CIRCLE_BUILD_URL:?}")
+        curl -X POST -H 'Content-Type: application/json' ${ROCKET_HOOK_URL:?} --data "${HOOK_DATA:?}"
+      when: on_fail
+
 jobs:
   build:
     <<: *container_config
@@ -273,6 +293,7 @@ jobs:
       - run:
           name: UAT Tests
           command: |
+            false
             epmd -daemon
             make python-env && make swagger && make multi-build
             make python-uats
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -126,6 +126,26 @@ references:
...
+                "title": "Build Link",
+                "title_link": "%s",
+                "color": "#FAD6D6"
+              }
+            ]
+          }
+      command: |
+        HOOK_DATA=$(printf "$HOOK_TEMPLATE" "${CIRCLE_JOB:?}" "${CIRCLE_BRANCH:?}" "${CIRCLE_BUILD_URL:?}")
+        curl -X POST -H 'Content-Type: application/json' ${ROCKET_HOOK_URL:?} --data "${HOOK_DATA:?}"
+      when: on_fail
```
lucafavatella: I see - if any one of previous steps returned non-zero exit status:
https://circleci.com/docs/2.0/configuration-reference/#run
https://circleci.com/docs/2.0/configuration-reference/#the-when-attribute

.circleci/config.yml
```diff
@@ -126,6 +126,26 @@ references:
...
+            "attachments": [
+              {
+                "title": "Build Link",
+                "title_link": "%s",
+                "color": "#FAD6D6"
+              }
+            ]
+          }
+      command: |
+        HOOK_DATA=$(printf "$HOOK_TEMPLATE" "${CIRCLE_JOB:?}" "${CIRCLE_BRANCH:?}" "${CIRCLE_BUILD_URL:?}")
```
lucafavatella: I see job branch build_url at https://circleci.com/docs/2.0/env-vars/#build-specific-environment-variables - ok.

.circleci/config.yml
```diff
@@ -126,6 +126,26 @@ references:
...
+        curl -X POST -H 'Content-Type: application/json' ${ROCKET_HOOK_URL:?} --data "${HOOK_DATA:?}"
```
lucafavatella: @dincho ROCKET_HOOK_URL comes from Circle CI account setup - right?

.circleci/config.yml
```diff
@@ -126,6 +126,26 @@ references:
           owner=${CIRCLE_PROJECT_USERNAME} repo=${CIRCLE_PROJECT_REPONAME} \
           tag=${CIRCLE_TAG} ASSETS=${PACKAGES_DIR:?}/*
 
+  fail_notification: &fail_notification
+    run:
+      name: Send failure notification
+      environment:
+        HOOK_TEMPLATE: |
+          {
+            "text": "CircleCI job **%s** failed on branch **%s**",
```
lucafavatella: I suggest to name who did change (`CIRCLE_USERNAME`? did not try it) so to jump-start initial investigation.

.circleci/config.yml
```diff
@@ -290,6 +310,7 @@ jobs:
       - store_artifacts:
           path: _build/dev3/rel/epoch/log
           destination: node3/
+      - *fail_notification
```
lucafavatella: In order to finish ["Notify a dedicated chat channel for CI failures on master"](https://www.pivotaltracker.com/story/show/156197975) I expect all CI failure on master to be notified in chat channel - not only for job uat_tests.

.circleci/config.yml
dincho: @lucafavatella right, because it contains secrets (token) thus have to be in the env config of the CI

.circleci/config.yml
dincho: @lucafavatella The notification contains link to the CI job where you can find all the information needed.

I prefer to keep the notification params simple to easier maintenance and better stability.

.circleci/config.yml
dincho: I really don't see an easy way to do that without splitting (duplicating) all the jobs to be master specific as the fail_notification is run as a job step (no builtin CCI notification I can found for v2.0).

Thus I kept this only for uat_tests as it's the only job that runs only on master. Other jobs all run on PR builds too, so the probability to fail when merged is low IMO.


Let me know if you have better idea how to achieve your request.

lucafavatella: @dincho Have you considered making notification conditional based on CIRCLE_BRANCH env var?

dincho: @lucafavatella I just come up with the same idea on the terrace 🚬

.circleci/config.yml
```diff
@@ -126,10 +126,34 @@ references:
           owner=${CIRCLE_PROJECT_USERNAME} repo=${CIRCLE_PROJECT_REPONAME} \
           tag=${CIRCLE_TAG} ASSETS=${PACKAGES_DIR:?}/*
 
+  fail_notification: &fail_notification
+    run:
+      name: Send failure notification
+      environment:
+        NOTIFY_BRANCH: PT-156197975_failing_master_notifications
```
lucafavatella: ?

dincho: testing ATM

dincho:
@lucafavatella see updated description for more examples and the channel

.circleci/config.yml
lucafavatella: Only this needed to be changed then looks good.

.circleci/config.yml
```diff
@@ -174,6 +199,7 @@ jobs:
           command: |
             epmd -daemon
             make eunit
+      - *fail_notification
 
   aevm_tests:
```
lucafavatella: I suggest adding notification to aevm_tests too.

dincho:
> Minor addition suggested aevm_tests

I'll abstain from adding it right way, because I think it may need special case handling and testing. Looks like cronjob builds does not have "triggered by user" variable (it says unknown). So I'm not sure if the variable is set or just equals "unknown". see https://circleci.com/gh/aeternity/epoch/8026

Feel free to create new ticket for further testing and implementation.

.circleci/config.yml
dincho: done in 9b407b7

lucafavatella:
> Feel free to create new ticket for further testing and implementation.

Tracked as https://www.pivotaltracker.com/story/show/156877392 Thanks.

PR #988, opened 2018-04-17
More meaningful chain snapshots name

dincho:
- add environment name to the filename
- use node public IP address instead of hostname

example: https://circleci.com/gh/aeternity/epoch/7874#artifacts/containers/0

https://www.pivotaltracker.com/story/show/156612994


dincho:
WIP until test/example build

PR #989, opened 2018-04-17
Refactor peer connections to use ranch

sylane:


apps/aecore/src/aec_connection_sup.erl
```diff
@@ -85,3 +101,13 @@ sync_listen_address() ->
                 aecore, sync_listen_address, ?DEFAULT_SYNC_LISTEN_ADDRESS),
     {ok, IpAddress} = inet:parse_address(binary_to_list(Config)),
     IpAddress.
+
+acceptors() ->
+    aeu_env:user_config_or_env([<<"sync">>, <<"accetors">>],
```
lucafavatella: Does accetors pick correct value here?

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -37,27 +46,19 @@
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
 
+%% -- BAHAVIOUR ranch_protocol CALLBACKS -------------------------------------
```
lucafavatella: BEHAVIOUR?

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -149,14 +148,32 @@ handle_cast(give_tcp_socket_ownership, S0 = #{ role := responder, tcp_sock := Tc
...
+                   , version => Version
+                   , genesis => Genesis},
+    {ok, Opts1, 0}.
+
+handle_call(Request, From, State) ->
+    handle_request(State, Request, From).
+
+handle_cast(retry, S = #{ host := Host, port := Port, status := error }) ->
+    Self = self(),
+    Pid = spawn(fun() -> try_connect(Self, Host, Port, 1000) end),
```
lucafavatella: There are two calls to try_connect, both with hardcoded timeout 1000. Shall those/that be made configurable?

lucafavatella: It looks like the gen_tcp:connect embedded in this try_connect call has the possibility of passing send_timeout option. Shall that be configurable?

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -37,27 +46,19 @@
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
```
lucafavatella: I see above:
```
-define(FIRST_PING_TIMEOUT, 30000).
-define(NOISE_HS_TIMEOUT, 5000).
```
shall they be made configurable?

apps/aecore/src/aec_connection_sup.erl
dincho: I guess it should be `acceptors` (typo) ?

apps/aecore/src/aec_peer_connection.erl
sylane: I didn't think that outgoing connections were part of the feature.

sylane: I don't know, as I said before, I though outgoing wasn't part of the ticket.

apps/aecore/src/aec_peer_connection.erl
sylane: Same here. I understand it could be configurable, but it didn't seem part of the ticket.

apps/aecore/src/aec_connection_sup.erl
lucafavatella: It seems typo.

PR #990, opened 2018-04-17
remove test dependency from all `tag: only: /^v.*$/

acidsennui:
PT: https://www.pivotaltracker.com/story/show/156798908

i think this should not block release with failing `test` job, and it should be left there to run

https://github.com/aeternity/epoch/blob/master/.circleci/config.yml#L428

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -522,7 +522,6 @@ workflows:
 
       - deploy_uat:
           requires:
-            - test
             - eunit
             - static_analysis
             - linux_package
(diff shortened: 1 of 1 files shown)


dincho:
@lucafavatella if we remove the `test` dependancy, what's the point to have `eunit` and `static_analysis` ?

dincho:
Ironically intermittent test failure during building this PR :)

dincho:
I restarted failed jobs.

PR #991, opened 2018-04-17
Clarify that hard forking test is up to date to 0.11.0

lucafavatella:
Preparatory PR for next step in hard forking support https://www.pivotaltracker.com/story/show/156671329

A change that may affect test suite is non-deprecated HTTP APIs moving from internal to external https://github.com/aeternity/epoch/commit/bd8a10ba2b4d2102ad7ac9147352f220ef762d63 .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,7 +154,8 @@ all-tests: eunit test
 system-test:
 	@./rebar3 as system_test do ct --dir system_test --logdir system_test/logs --config system_test/cfg $(CT_TEST_FLAGS)
 
-system-test-deps: | system_test/aest_hard_fork_SUITE_data/db/mnesia_ae-uat-epoch_backup.tar ;
+system-test-deps: | system_test/aest_hard_fork_SUITE_data/db/mnesia_ae-uat-epoch_backup.tar
+	docker pull aeternity/epoch:v0.11.0
 
 system_test/aest_hard_fork_SUITE_data/db/mnesia_ae-uat-epoch_backup.tar: system_test/aest_hard_fork_SUITE_data/db/mnesia_ae-uat-epoch_backup
 	tar -c -C $(<D) -f $@ $(<F)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -154,7 +154,8 @@ all-tests: eunit test
 system-test:
 	@./rebar3 as system_test do ct --dir system_test --logdir system_test/logs --config system_test/cfg $(CT_TEST_FLAGS)
 
-system-test-deps: | system_test/aest_hard_fork_SUITE_data/db/mnesia_ae-uat-epoch_backup.tar ;
+system-test-deps: | system_test/aest_hard_fork_SUITE_data/db/mnesia_ae-uat-epoch_backup.tar
+	docker pull aeternity/epoch:v0.11.0
```
jur0: What if we do also `docker tag aeternity/epoch:v0.11.0 aeternity/epoch:local`? It would make `:local` alias of `:v0.11.0` so then other changes in the code are not necessary.

lucafavatella: Because other tests may expect :local to be local - I am thinking of sync tests https://github.com/aeternity/epoch/blob/a9798667c3a7e49926ae4caa874d7625b2b1e58f/system_test/aest_sync_SUITE.erl#L56

dincho: Maybe you could use different tag for the fork tests? I like the @jur0 idea.

system_test/aest_hard_fork_SUITE.erl
```diff
@@ -79,7 +79,7 @@
           name    => new_node1,
           peers   => [],
           backend => aest_docker,
-          source  => {pull, "aeternity/epoch:local"},
+          source  => {pull, "aeternity/epoch:v0.11.0"},
```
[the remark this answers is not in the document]
lucafavatella: `-define(TESTED_DOCKER_IMAGE, "aeternity/epoch:v0.11.0").` - preparing patch.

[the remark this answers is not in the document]
lucafavatella: Pushed update.
Running suite locally I see already a test using new image passed so macro works.

PR #994, opened 2018-04-18
0.11.1 release preparation

lucafavatella:


lucafavatella:
@dincho I force pushed bumping Swagger version to be in line with VERSION, and rebased.

PR #997, opened 2018-04-18
Pt 156874858 sync fixes 0.11.1

hanssv:


apps/aecore/src/aec_peer_connection.erl
```diff
@@ -237,11 +237,12 @@ handle_info({noise, _, <<Type:16, Payload/binary>>}, S) ->
     end;
 handle_info({tcp_closed, _}, S) ->
     lager:debug("Peer connection got tcp_closed", []),
+    S1 = cleanup_connection(S),
```
lucafavatella: I see this and rebar.config/lock from https://github.com/aeternity/epoch/pull/993/files - ok

PR #998, opened 2018-04-18
Updated test_use_cases.py to reveal more shortcomings

hanssv:


py/tests/integration/test_use_cases.py
```diff
@@ -137,14 +137,15 @@ def test_node_discovery():
     root_dir = tempfile.mkdtemp()
 
     # Alice's config: no peers
+    alice_peers = ' {peers, [<<"aenode://pp$28uQUgsPcsy7TQwnRxhF8GMKU4ykFLKsgf4TwDwPMNaSCXwWV8@localhost:3025">>]}, '
     alice_sys_config = make_peers_config(root_dir, "alice.config",
-                            alice_peer_url, "node1", "3015", '{peers, []},', mining=True)
-    print("\nAlice has address " + alice_peer_url + " and no peers")
+                            alice_peer_url, "node1", "3015", alice_peers, mining=True)
+    print("\nAlice has address " + alice_peer_url + " and peers [" + bob_peer_url + "]")
     # Bob's config: only peer is Alice
```
lucafavatella: I see this changes the nature of the test (maybe it is different test) so not fit for merging. Though I am considering this for validation of 0.11.1

py/tests/integration/test_use_cases.py
```diff
@@ -137,14 +137,15 @@ def test_node_discovery():
     root_dir = tempfile.mkdtemp()
 
     # Alice's config: no peers
+    alice_peers = ' {peers, [<<"aenode://pp$28uQUgsPcsy7TQwnRxhF8GMKU4ykFLKsgf4TwDwPMNaSCXwWV8@localhost:3025">>]}, '
```
lucafavatella: I see this is meant to be Bob (same pubkey string as few lines below).

py/tests/integration/test_use_cases.py
lucafavatella: This looks a new test:
> Assuming Carol's node only knows about Bob upon startup and that Alice's node knows Bob, Carol's node should be able to discover Alice and sync with her node.

Suitable against master as new test.

lucafavatella: Tracked in https://www.pivotaltracker.com/story/show/156879823

lucafavatella:
Tracked in https://www.pivotaltracker.com/story/show/156879823

PR #1003, opened 2018-04-19
Set connection to pending upon retry

hanssv:
PR1000 should also go onto master.


gorillainduction:
Restarting tests on known intermittent error
```
%%% aehttp_integration_SUITE ==> {{badmatch,{error,socket_closed_remotely}},
```

PR #1004, opened 2018-04-19
Test re sync 0.11.1 fixes

lucafavatella:


py/tests/integration/common.py
```diff
@@ -46,12 +46,20 @@ def external_api(name):
 def internal_api(name):
     return INT_API[name]
 
-def node_online(api):
-    try:
-        top = api.get_top()
-        return top.height > -1
-    except Exception as e:
-        return False
+def node_online(ext_api, int_api):
```
hanssv: is python `and` short circuiting? Not that it is a big deal but if the first check fails the second can be skipped.

lucafavatella: It is.

From https://docs.python.org/3/reference/expressions.html#boolean-operations
> The expression x and y first evaluates x; if x is false, its value is returned; otherwise, y is evaluated and the resulting value is returned.

lucafavatella:
> How to merge: preferably "Squash and merge" with commit message being list of commit messages (excluding enabling/disabling UAT tests in any branch).

I should have squashed... though it is fine.

PR #1007, opened 2018-04-19
Remove invalid CircleCI configuration

dincho:
https://www.pivotaltracker.com/story/show/156886481

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,8 +2,6 @@ version: 2
 
 references:
   container_config: &container_config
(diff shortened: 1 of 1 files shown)


dincho:
> I have seen this only when re-running a job - so maybe the config is picked up even if undocumented? Please monitor impact on speed of CI after merging.
> 
> Also maybe putting config in correct location would actually enable a speedup?

I guess I'll even open a support question to CircleCI as the docs are not fully clear to me.

PR #894, opened 2018-03-26
PT-154173362 VM/chain API

UlfNorell:
This PR implements a bare-bones interface between the chain and a VM. It currently contains `spend` and `call_contract` functions. All the required information about the chain is kept in a state that's abstract to the VM.

https://www.pivotaltracker.com/n/projects/2159793/stories/154173362

PR #1008, opened 2018-04-19
Pt 154173362 vm chain api  rebased on master

lucafavatella:
This is PR #894 without the consensus protocol version changes (to be addressed in separate PR re hard fork). I tested in [this CI workflow](https://circleci.com/workflow-run/44fa8dea-5e7b-4c15-aaa7-64f80ad13ae1) including Python UAT tests and aevm-test.

I am running further check as this is meant to be the first consensus breaking change since v0.11.0.


lucafavatella:
I checked that node with this PR is able to verify testnet chain up to height 5254 (I did not check further - current height is 5718).

deployment/ansible/deploy.yml
```diff
@@ -61,7 +61,7 @@
           port: 3114
       chain:
         persist: true
-        db_path: "./db43"
+        db_path: "./db44"
```
lucafavatella: This will conflict with https://github.com/aeternity/epoch/pull/1018/files#diff-2ad1947fe9b869ffd23a6a86872e65e0R64

lucafavatella: Removed.

PR #1009, opened 2018-04-19
Enable back backward compatibility sync system-test

sylane:
* Parametrize the dependencies not removed by make clean

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,9 +25,6 @@ export AEVM_EXTERNAL_TEST_DIR=aevm_external
 HTTP_APP = apps/aehttp
 SWTEMP := $(shell mktemp -d 2>/dev/null || mktemp -d -t 'mytmpdir')
 
-## Dependency `rocksdb` takes long to build, the others are just to shave off some seconds...
-CLEAN_DEPS_EXCEPTIONS=rocksdb sha3 enacl erlexec
-
 all:	local-build
 
 console:
@@ -246,7 +243,7 @@ clean:
 	@$(MAKE) multi-distclean
 	@rm -rf _build/system_test+test _build/system_test system_test/aest_hard_fork_SUITE_data/db system_test/logs _build/test _build/prod _build/local
 	@rm -rf _build/default/plugins
-	@rm -rf $$(ls -d _build/default/lib/* | grep -Ev "$$(sep=""; for d in $(CLEAN_DEPS_EXCEPTIONS); do echo -n "$$sep[^_]$$d"; sep="|"; done)")
+	@rm -rf $$(ls -d _build/default/lib/* | grep -v '[^_]rocksdb') ## Dependency `rocksdb` takes long to build.
 
 distclean: clean
 	( cd apps/aecuckoo && $(MAKE) distclean; )
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,9 @@ export AEVM_EXTERNAL_TEST_DIR=aevm_external
 HTTP_APP = apps/aehttp
 SWTEMP := $(shell mktemp -d 2>/dev/null || mktemp -d -t 'mytmpdir')
 
+## Dependency `rocksdb` takes long to build, the others are just to shave off some seconds...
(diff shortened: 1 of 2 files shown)


system_test/aest_sync_SUITE.erl
```diff
@@ -38,15 +38,15 @@
     peers   => [old_node2],
     backend => aest_docker,
     % Change to a compatible fixed version when possible
```
lucafavatella: So the comment can be deleted or updated?

system_test/aest_sync_SUITE.erl
```diff
@@ -38,15 +38,15 @@
...
-    source  => {pull, "aeternity/epoch:local"}
+    source  => {pull, "aeternity/epoch:v0.11.1"}
 }).
 
 -define(OLD_NODE2, #{
     name    => old_node2,
     peers   => [old_node1],
     backend => aest_docker,
     % Change to a compatible fixed version when possible
```
lucafavatella: So the comment can be deleted or updated?

Makefile
```diff
@@ -242,7 +246,7 @@ clean:
 	@$(MAKE) multi-distclean
 	@rm -rf _build/system_test+test _build/system_test system_test/aest_hard_fork_SUITE_data/db system_test/logs _build/test _build/prod _build/local
 	@rm -rf _build/default/plugins
-	@rm -rf $$(ls -d _build/default/lib/* | grep -v '[^_]rocksdb') ## Dependency `rocksdb` takes long to build.
+	@rm -rf $$(ls -d _build/default/lib/* | grep -Ev "$$(sep=""; for d in $(CLEAN_DEPS_EXCEPTIONS); do echo -n "$$sep[^_]$$d"; sep="|"; done)")
```
lucafavatella: I did not have a chance of trying this locally - in the middle of other task - though this looks ok and POSIX compliant.

system_test/aest_sync_SUITE.erl
sylane: I'll do that right after the standup

system_test/aest_sync_SUITE.erl
```diff
@@ -37,16 +37,14 @@
     name    => old_node1,
     peers   => [old_node2],
     backend => aest_docker,
-    % Change to a compatible fixed version when possible
-    source  => {pull, "aeternity/epoch:local"}
+    source  => {pull, "aeternity/epoch:v0.11.1"}
```
lucafavatella: If we agreed that new releases shall keep backwards compatibility of P2P network this may be updated to :latest though for now good enough.

Makefile
```diff
@@ -25,6 +25,9 @@ export AEVM_EXTERNAL_TEST_DIR=aevm_external
 HTTP_APP = apps/aehttp
 SWTEMP := $(shell mktemp -d 2>/dev/null || mktemp -d -t 'mytmpdir')
 
+## Dependency `rocksdb` takes long to build, the others are just to shave off some seconds...
+CLEAN_DEPS_EXCEPTIONS=rocksdb sha3 enacl erlexec
```
ThomasArts: This seems dangerous to me. If we update one of them, we won't notice and it only takes a few seconds.

This optimation takes more time than it saves if we in a few weeks start using a new version of erlexec and people do a make clean to get the latest version in their env.

sylane: Are you talking about erlexec in particular or about all of them (sha3, enacl and erlexec) ?

Makefile
dincho: Why all of this magic instead of `rocksdb|sha3|enacl|erlexec` (kind of) ?

Makefile
dincho: I somehow agree with @ThomasArts `make clean` is used to clean **all** the deps.
However, we should not speedup the *build after clean*, but we rather should fix all the broken makefiles.

`make any-target` should perfectly work regardless if you updated `erlexec` (or any other vendor lib) or not.

However, this is just my rant about our builds, not strictly related to this PR.

lucafavatella:
When merging, please choose squashing among the GitHub-presented options.

PR #1012, opened 2018-04-20
Swagger HTTP API client for testing

jur0:
It's meant to be used in integration and system tests.


apps/aehttp/test/aehttp_client.erl
```diff
@@ -0,0 +1,100 @@
...
+    Op = endpoints:operation(OpId),
+    {Method, Interface} = operation_spec(Op),
+    BaseUrl = proplists:get_value(Interface, Cfg),
+    Params1 = maybe_encode_params(Params, Cfg),
+    Path = operation_path(Method, OpId, Params1),
+    request(Method, BaseUrl, Path, Params1, [], [], []).
+
+request(get, BaseUrl, Path, QueryParams, Headers, HttpOpts, Opts) ->
+    Url = make_url(BaseUrl, Path, QueryParams),
+    ct:log("GET ~p", [Url]),
```
lucafavatella: I see ct:log hardcoded. Ok as used only in test.

apps/aehttp/test/aehttp_client.erl
```diff
@@ -0,0 +1,100 @@
...
+operation_spec(#{get := GetSpec}) ->
+    {get, operation_interface(GetSpec)};
+operation_spec(#{post := PostSpec}) ->
+    {post, operation_interface(PostSpec)}.
+
+operation_interface(#{tags := Tags}) ->
+    IsExt = lists:member(<<"external">>, Tags),
+    IsInt = lists:member(<<"internal">>, Tags),
+    case {IsExt, IsInt} of
+        {true, _} -> ext_addr;
```
lucafavatella: I see that, as a client, if I want to invoke an HTTP API on an old version of the node I can rather use request/7 so looks ok. (I am thinking e.g. to bd8a10ba2b4d2102ad7ac9147352f220ef762d63 .)

jur0: Or you can just pass `ext_addr` and `int_addr` of the old node in `Cfg` param of `request/3`.

PR #1019, opened 2018-04-23
Minimize repetition of consensus protocol versions in repo

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/156960045

Addresses https://github.com/aeternity/epoch/pull/1015#issuecomment-383529414

Remaining consensus protocol versions are in:
* blocks.hrl inevitable for macros
* aec_governance.erl - inevitable for default height of each hard fork (maybe could be in config - though shall be somewhere)
* aest_hard_fork_SUITE.erl - ~inevitable for tests as config instantiated~ minimized in PR #1015

### DIFF
--- a/apps/aeutils/priv/epoch_config_schema.json
+++ b/apps/aeutils/priv/epoch_config_schema.json
@@ -227,24 +227,8 @@
                     "description" : "The consensus protocol versions with respective effective heights",
                     "type" : "object",
                     "additionalProperties" : false,
-                    "required" : [
-                        "9",
-                        "10",
-                        "11"
-                    ],
-                    "properties" : {
-                        "9" : {
-                            "description" : "Minimum height at which protocol is effective",
-                            "type" : "integer",
-                            "minimum" : 0,
-                            "maximum" : 0
-                        },
-                        "10" : {
-                            "description" : "Minimum height at which protocol is effective",
-                            "type" : "integer",
-                            "minimum" : 0
-                        },
-                        "11" : {
+                    "patternProperties" : {
+                        "^[1-9][0-9]*$": {
                             "description" : "Minimum height at which protocol is effective",
                             "type" : "integer",
                             "minimum" : 0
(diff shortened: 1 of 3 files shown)


apps/aeutils/test/aeu_env_tests.erl
```diff
@@ -10,17 +10,30 @@
...
 
+extra_checks_test_() ->
+    {setup,
+     fun() -> ok = meck:new(setup, [passthrough]), setup() end,
+     fun(R) -> teardown(R), ok = meck:unload(setup) end,
+     [{"Example user configuration files pass checks further to the schema", %% For enabling files to be linked from wiki as examples.
+       [fun() ->
+                {ok, {UserMap, UserConfig}} = aeu_env:check_config(Config),
+                ok = mock_user_config(UserMap, UserConfig),
+                ?assertEqual(ok, aec_hard_forks:check_env())
```
dincho: I'm not too much into the test suites, but I guess this is a "unit test"? Isn't calling additional module an antipattern ?
Also mocking in unit tests.

eproxus: Mocking in unit tests is fine I think, it's about isolating the "unit". However, I still think mocking in general is an anti pattern and that interfaces that are unit tested should be structured in such a way that mocking is not needed (parameterization, dependency injection, pure functional code etc.)

apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -227,24 +227,8 @@
...
-                            "minimum" : 0,
-                            "maximum" : 0
-                        },
-                        "10" : {
-                            "description" : "Minimum height at which protocol is effective",
-                            "type" : "integer",
-                            "minimum" : 0
-                        },
-                        "11" : {
+                    "patternProperties" : {
```
eproxus: I find this name to be very generic, to the point where I don't understand at all what it is meant to signify. Maybe something more specific is better, like `minHeights` (just an example, I'm not familiar with this data enough)?

eproxus: Ignore this, it's JSON schema and not our config 🤦‍♂️

PR #1023, opened 2018-04-24
PT156910906 - Make sync target a fork rather than individual peers

hanssv:
When starting to sync against a PeerId we try really hard to identify the chain this Peer
is on -  and then we organise the work in sync tasks _per chain_

### DIFF
--- a/apps/aecore/src/aec_sync.erl
+++ b/apps/aecore/src/aec_sync.erl
@@ -6,15 +6,16 @@
 %%%=============================================================================
 -module(aec_sync).
 
+-include("common.hrl").
+-include("blocks.hrl").
+
 -behaviour(gen_server).
 
 -import(aeu_debug, [pp/1]).
 -import(aec_peers, [ppp/1]).
 
 %% API
--export([connect_peer/1,
-         start_link/0
-        ]).
+-export([start_link/0]).
 
 %% API called from strongly connected component aec_peers
 -export([schedule_ping/1]).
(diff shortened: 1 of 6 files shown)


apps/aecore/src/aec_sync.erl
```diff
@@ -230,132 +203,198 @@ terminate(_, _) ->
...
-      || PeerInfo <- Peers ].
+delayed_run_job(PeerId, Task, Queue, Fun, Delay) ->
+    OldWorker = self(),
+    spawn(fun() -> NewWorker = self(),
+                   handle_worker(Task, {change_worker, PeerId, OldWorker, NewWorker}),
+                   timer:sleep(Delay),
+                   jobs:run(Queue, Fun)
+          end).
+
+run_job(Queue, Fun) ->
```
ThomasArts: I cannot really oversee what the difference is between using the jobs package and spawning workers.
Do we possibly loose anything in this way, such as load regulation, zombie process cleanup, things like that? You still call jobs, but via an extra spawn?

apps/aecore/src/aec_sync.erl
```diff
@@ -230,132 +203,198 @@ terminate(_, _) ->
...
+    equal;
+match_chains(_, _) ->
+    different.
+
+do_handle_worker({new_worker, PeerId, Pid}, ST = #sync_task{ workers = Ws }) ->
+    case lists:keyfind(PeerId, 1, Ws) of
+        false    -> ok;
+        {_, Old} -> lager:info("Peer ~p already has a worker (~p)", [ppp(PeerId), Old])
+    end,
+    erlang:monitor(process, Pid),
```
gorillainduction: The worker should be linked rather than monitored so that the workers are brought down if the server goes down. That would also mean that the server needs to trap exits (but not the workers).

Since the workers would die if the server dies, there is probably no need for the `terminate/1` function to clean anything up.

An alternative is to add the workers to a simple_one_for_one supervisor, but the only benefit I can see is to avoid the full exit signal from the server ending up in the crash message for the workers.

apps/aecore/src/aec_sync.erl
```diff
@@ -230,132 +203,198 @@ terminate(_, _) ->
...
-            lager:debug("unknown job", [])
-    end.
-
-enqueue(Op, Msg) ->
-    Peers = aec_peers:get_random(all),
-    [ jobs:enqueue(sync_jobs, {Op, Msg, aec_peers:peer_id(PeerInfo)})
-      || PeerInfo <- Peers ].
+delayed_run_job(PeerId, Task, Queue, Fun, Delay) ->
+    OldWorker = self(),
+    spawn(fun() -> NewWorker = self(),
```
gorillainduction: You might want to use `proc_lib:spawn` in all these places to avoid excessive error logging when the supervision tree is brought down, and to get nicer error reports.

config/dev1/sys.config
```diff
@@ -34,12 +34,11 @@
...
-     [{sync_jobs, [passive]},
-      {sync_workers, [{regulators,
-                       [{counter, [{limit, 10}]}
-                       ]},
-                      {producer, {aec_sync, sync_worker, []}}
-                     ]},
+     [{ping_workers, [{regulators, [{counter, [{limit, 3}]}]}]},
+      {mempool_workers, [{regulators, [{counter, [{limit, 1}]}]}]},
+      {sync_task_workers, [{regulators, [{counter, [{limit, 10}]}]}]},
+      {gossip_workers, [{regulators, [{counter, [{limit, 10}]}]}]},
```
gorillainduction: I would like to see markers that tell me these workers are sync workers. Perhaps prefix sync_ping_workers or something like it. Especially for mempool_workers which might be confusing.

PR #1028, opened 2018-04-24
PT-156751861: Test sync speeds

eproxus:


system_test/aest_sync_SUITE.erl
```diff
@@ -31,7 +34,9 @@
 
 %=== MACROS ====================================================================
 
--define(MINING_TIMEOUT, 2000).
+-define(MINING_TIMEOUT,   2 * 1000).
+-define(STARTUP_TIMEOUT, 10 * 1000).
+-define(SYNC_TIMEOUT,     1 * 1000).
```
hanssv: Is this per block? Then it is very large, 50ms per block or something sounds reasonable?!

system_test/helpers/aest_nodes.erl
```diff
@@ -561,14 +561,14 @@ mgr_prepare_specs(NodeSpecs, State) ->
...
         NewPeers = lists:map(fun
             (Addr) when is_binary(Addr) -> Addr;
             (Name) when is_atom(Name) ->
                 case maps:find(Name, AllAddrs) of
                     {ok, Addr} -> Addr;
                     _ -> error({peer_not_found, Name})
                 end
-        end, Peers),
-        Spec#{peers := NewPeers}
+        end, maps:get(peers, Spec, [])),
```
lucafavatella: Shall spec in [aest_nodes:node_spec](https://github.com/aeternity/epoch/blob/1134b14ba9eacbb68e59ca4948d50ab4af8abb5b/system_test/helpers/aest_nodes.erl#L71) be updated then?

system_test/aest_sync_SUITE.erl
```diff
@@ -384,8 +390,35 @@ net_split_recovery(Cfg) ->
...
+    InitialNodes = [node_name(I) || I <- lists:seq(1, 4)],
+    NextNodes = [node_name(I) || I <- lists:seq(5, 6)],
+    Height = 1000,
+
+    NodeSpecs = [#{
+        name => N,
+        backend => aest_docker,
+        source => {pull, "aeternity/epoch:v0.11.1"},
+        mine_rate => 100
+    } || N <- InitialNodes ++ NextNodes],
```
lucafavatella: I see based on code inspection that absence of peers key in node spec means `peers: []` i.e. each node is alone.

system_test/helpers/aest_nodes.erl
eproxus: I guess it's not needed, a spec of `[Thing]` can be an empty list AFAIK

system_test/aest_sync_SUITE.erl
```diff
@@ -384,8 +390,35 @@ net_split_recovery(Cfg) ->
...
+
+    setup_nodes(NodeSpecs, Cfg),
+    [start_node(N, Cfg) || N <- InitialNodes],
+    wait_for_value({height, Height}, InitialNodes, ?MINING_TIMEOUT * Height, Cfg),
```
lucafavatella: Each of these 4 nodes is mining in isolation, right?

system_test/aest_sync_SUITE.erl
```diff
@@ -384,8 +390,35 @@ net_split_recovery(Cfg) ->
...
+
+    start_node(node_name(5), Cfg),
+    wait_for_value({height, 0}, [node_name(5)], ?STARTUP_TIMEOUT, Cfg),
+    wait_for_value({height, Height}, [node_name(5)], ?SYNC_TIMEOUT * Height, Cfg),
```
lucafavatella: This node is not syncing but mining in isolation, right?

system_test/aest_sync_SUITE.erl
```diff
@@ -384,8 +390,35 @@ net_split_recovery(Cfg) ->
...
+
+    start_node(node_name(6), Cfg),
+    wait_for_value({height, 0}, [node_name(5)], ?STARTUP_TIMEOUT, Cfg),
```
lucafavatella: Here you mean node 6?

system_test/aest_sync_SUITE.erl
```diff
@@ -384,8 +390,35 @@ net_split_recovery(Cfg) ->
...
+    wait_for_value({height, Height}, [node_name(6)], ?SYNC_TIMEOUT * Height, Cfg),
```
lucafavatella: This node is not syncing but mining in isolation, right?

system_test/aest_sync_SUITE.erl
eproxus: 🤦‍♂️

system_test/aest_sync_SUITE.erl
eproxus: @lucafavatella Addressed in c575ff2

system_test/aest_sync_SUITE.erl
```diff
@@ -384,8 +390,43 @@ net_split_recovery(Cfg) ->
...
+        || N <- NextNodes
+    ],
+
+    setup_nodes(InitialNodeSpecs ++ NextNodeSpecs, Cfg),
+    [start_node(N, Cfg) || N <- InitialNodes],
+    wait_for_value({height, Height}, InitialNodes, ?MINING_TIMEOUT * Height, Cfg),
+
+    start_node(node_name(5), Cfg),
+    wait_for_value({height, 0}, [node_name(5)], ?STARTUP_TIMEOUT, Cfg),
+    wait_for_value({height, Height}, [node_name(5)], ?SYNC_TIMEOUT * Height, Cfg),
```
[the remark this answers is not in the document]
lucafavatella: Putting :local, at this time locally node5 had arrived at height 818

system_test/aest_sync_SUITE.erl
```diff
@@ -384,8 +390,43 @@ net_split_recovery(Cfg) ->
...
     ok.
 
+sync_speed(Cfg) ->
+    InitialNodes = [node_name(I) || I <- lists:seq(1, 4)],
+    NextNodes = [node_name(I) || I <- lists:seq(5, 6)],
+    Height = 1000,
+
+    Spec = #{
+        backend => aest_docker,
+        source => {pull, "aeternity/epoch:v0.11.1"},
```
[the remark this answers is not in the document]
eproxus: You're right, this should be `:local` of course.

eproxus:
> Analyzing ticket PT-156751861 I see:
> 
> > We want to test that adding a new node to the network works well in an existing net (reasonable sync speed)
> 
> I see the test is testing speed of syncing relative to ?SYNC_TIMEOUT. Will probably required fine-tuning / parametrization when running on CI.

This is a tricky part. I'm not convinced putting more strict timeouts is beneficially, because of the risk of getting intermittently failing tests. I think these timeouts are more a warning flag if things go really wrong. Real performance tracking should be done in a different way, open for ideas here!

> I have not run test locally though it appears ok.
> 
> > Measure how quickly a new node catches up with a 8-node network with 10k blocks. Sync speed should be faster with more nodes, i.e. 8 nodes should sync a 9th node twice as fast as 4 nodes.
> 
> What is the plan for this portion? This portion of the test would be probably more robust.

The blocker here is to figure out how to to emulate this locally. One problem is that all Docker containers share the CPU equally, so adding one more essentially makes all nodes slower. I could change the test to have a chain of 10 000 blocks instead, but I'm not sure it would make a big difference...

system_test/aest_sync_SUITE.erl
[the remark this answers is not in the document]
eproxus: Two sets of blocks are a fork?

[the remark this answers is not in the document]
lucafavatella: I did not check whether the two sets of blocks were belonging to distinct forks. They could have been well portions of the same fork at different height based only on the logs (for some reason). I did not check the hashes.

lucafavatella:
> One problem is that all Docker containers share the CPU equally

Maybe you can instantiate Docker containers passing as param a fraction of the total CPUs e.g. 0.25 CPU per container?

system_test/aest_sync_SUITE.erl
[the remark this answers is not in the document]
eproxus: @lucafavatella I guess this would show if I compare the blocks at height 1000 in the end as well, even if there is no timeout. If what you observed was a fork, what could be the cause? Getting a fork with only 5 nodes doesn't seem so nice...

eproxus:
@lucafavatella Latest changes include a timeout fix for the old test cases, and block verification plus use of `:local` image

PR #1030, opened 2018-04-24
Add system test for epoch commands

sylane:


rebar.config
```diff
@@ -49,7 +49,7 @@
 {swagger_endpoints, [{src, "config/swagger.yaml"}, {dst, "apps/aeutils/src/endpoints.erl"}]}.
 
 {relx, [{release, { epoch, "version value comes from VERSION" },
-         [runtime_tools, lager, setup, sext, rocksdb, mnesia_rocksdb, gproc,
+         [runtime_tools, sasl, lager, setup, sext, rocksdb, mnesia_rocksdb, gproc,
```
lucafavatella: Could you please add small comment in PR on why and expected impact?

lucafavatella: Thinking more about this - could you please split sasl-changes in separate PR? Reason is that I would like to avoid possibility of creating unexpected volume or format of logs close to stabilization of master.

Makefile
```diff
@@ -158,8 +158,9 @@ system-test:
 	@./rebar3 as system_test do ct --dir system_test --logdir system_test/logs --config system_test/cfg $(CT_TEST_FLAGS)
 
 system-test-deps: | system_test/aest_hard_fork_SUITE_data/db/mnesia_ae-uat-epoch_backup.tar
-	docker pull aeternity/epoch:v0.11.1
 	$(info The downloaded DB backup is not checked for freshness.)
+	docker pull aeternity/epoch:v0.11.1
+	docker build . -t aeternity/epoch:local
```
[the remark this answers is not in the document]
sylane: It definitely builds it every time, I don't know how we could prevent that. I didn't know it was a requirement.

rebar.config
sylane: Adding SASL was to fix the command that wasn't working in the first place. I understand that you may want to wait before merging but the PR doesn't make sense without that.

lucafavatella: Which command?
If @dincho is ok to include sasl (for amount and format of logs), I am ok.

eproxus: Lager should take over SASL events. Including SASL in a production system is very good practice IMO.

sylane: `epoch versions` is not working without sasl.

Makefile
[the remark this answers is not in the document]
eproxus: I would also not include it at this time, until we have a solution that avoids too many rebuilds. It is a wider issue though, not just for these tests.

system_test/aest_commands_SUITE.erl
```diff
@@ -0,0 +1,84 @@
...
+
+%=== TEST CASES ================================================================
+
+epoch_commands(Cfg) ->
+    setup_nodes([?NODE1], Cfg),
+    start_node(node1, Cfg),
+    wait_for_value({height, 0}, [node1], ?STARTUP_TIMEOUT, Cfg),
+    wait_for_value({height, 10}, [node1], ?MINING_TIMEOUT * 10, Cfg),
+    {0, Output1} = aest_nodes:run_cmd_in_node_dir(node1, ["bin/epoch", "versions"], Cfg),
+    ?assertMatch({match, _}, re:run(Output1, ?EPOCH_VERSION_OUTPUT_RX)),
```
eproxus: I think having the output regex defined somewhere else makes the test harder to read. It's better if all the test data is inside the test itself. This comment applies to all macros used in the asserts below as well.

system_test/aest_hard_fork_SUITE.erl
```diff
@@ -638,9 +638,10 @@ restore_db_backup_on_node(NodeName, TarBin, Content, DestDir, Cfg) ->
     ct:log("Restoring DB backup of byte size ~p on node ~s",
            [byte_size(TarBin), NodeName]),
     _ = aest_nodes:run_cmd_in_node_dir(NodeName, ["mkdir", DestDir], Cfg),
-    "" = aest_nodes:run_cmd_in_node_dir(NodeName, ["ls", DestDir], Cfg),
+    {_, ""} = aest_nodes:run_cmd_in_node_dir(NodeName, ["ls", DestDir], Cfg),
```
eproxus: These should probably match on the exit code as well? Same for the calls to `run_cmd_in_node_dir` below.

Makefile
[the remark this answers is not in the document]
sylane: I'll remove it. I was using the command only when I wanted to rebuild the local image so I was fine with it being build every time :)

system_test/aest_hard_fork_SUITE.erl
sylane: There wasn't any check of the status code before, I just kept the exact same behavior.

lucafavatella:
> I think @lucafavatella meant that the PR should mention the change, not a comment in the code itself.

Correct - I meant comment in PR on github not code - though fine anyway.

eproxus:
The comment seems to be gone now, as far as I can see

rebar.config
```diff
@@ -49,7 +49,9 @@
 {swagger_endpoints, [{src, "config/swagger.yaml"}, {dst, "apps/aeutils/src/endpoints.erl"}]}.
 
 {relx, [{release, { epoch, "version value comes from VERSION" },
-         [runtime_tools, lager, setup, sext, rocksdb, mnesia_rocksdb, gproc,
+         % sasl is required for the command `epoch versions` to work,
+         % it is disabled in `sys.config` though.
```
eproxus: This comment should be removed, I don't think it adds anything considering the review comment was about the PR itself, not the code.

PR #1031, opened 2018-04-24
Plant support for hard forking with new tx types/versions

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/156959557

[CI workflow passing](https://circleci.com/workflow-run/a9880b4e-31f1-472b-a831-05e4d1c884e0) (including Python UAT tests and aevm_tests).

This PR contains no functional changes - it is meant to be used by spend tx with payload.

PR #1013, opened 2018-04-20
Spend tx: add payload

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/156942570)

In order to verify that this is a backwards compatible change, you can:
1) Checkout the branch
2) Configure config/sys.config so the peers section has one of the integration node's peers
```                                             
      {peers, [                                                                                       
        <<"aenode://pp$2Jx7N2qDUqu3ZNfd831Byiq4WofebBsSRHNTD79Tcsr9VTTt54@31.13.249.73:3015">> %% integration n0
      ]}, 
```   
3) Adapt `data/aecore/.genesis/accounts.json` to have the [correct integration preset account](https://github.com/aeternity/epoch/blob/master/deployment/ansible/group_vars/tag_env_integration/accounts.yml)

4) Build and start the node

5) Observe as the node syncs correctly with the existing network

6) Verify that there are _some_ spend transactions
```
(epoch@localhost)2> [Tx || {aec_tx, {_, _}, {signed_tx, {aetx, Type, _, _}, _}}= Tx   <-mnesia:dirty_select(aec_tx, [{'_',[],['$_']}]), Type =/= coinbase_tx].
[{aec_tx,{<<166,125,217,115,13,250,74,254,129,143,244,146,
            87,156,129,18,163,114,230,235,1,81,159,37,211,
            ...>>,
          <<177,27,82,183,23,159,84,91,184,8,30,114,95,47,222,137,
            208,221,76,95,210,47,64,173,...>>},
         {signed_tx,{aetx,spend_tx,aec_spend_tx,
                          {spend_tx,<<4,44,86,165,206,116,7,63,150,169,2,93,238,212,
                                      158,238,155,...>>,
                                    <<4,112,190,180,76,136,225,221,31,4,225,193,217,251,226,2,
                                      ...>>,
                                    42,10,1,<<>>,1}},
                    [<<48,69,2,33,0,230,58,191,240,128,198,90,255,98,24,12,
                       17,109,88,38,51,...>>]}},
 {aec_tx,{<<162,220,6,156,207,97,226,193,118,49,23,149,134,
            38,43,66,60,198,112,246,223,24,42,179,...>>,
          <<177,27,82,183,23,159,84,91,184,8,30,114,95,47,222,137,
            208,221,76,95,210,47,64,...>>},
         {signed_tx,{aetx,spend_tx,aec_spend_tx,
                          {spend_tx,<<4,44,86,165,206,116,7,63,150,169,2,93,238,212,
[...]

PR #1032, opened 2018-04-25
Pt 156942570 spend payload  hard fork support

lucafavatella:
Supersedes PR #1013

Depends on PR #1031 (included in this PR).

The lack of capability for the user via the user API to specify creation of old version of spend tx causes major issues in tests.
* The `make test` tests and Python UAT tests may be catered for by setting these tests - via user config epoch.yaml `hard_forks` - to switch earlier (block 3 or so) to the protocol version mining spend tx with payload.
* The aest_hard_fork_SUITE may be catered for by creating old txs only on old node, and relying on backward compatible sync to pass those txs to new node.

TODO:
* [ ] Update `make test`
* [ ] Update Python UAT tests
  * Meant to have been addressed (7e6595f9) - CI is checking
* [ ] Update aest_hard_fork_SUITE
* [ ] Remove commit enabling Python UAT tests
* [ ] Remove commit enabling aevm tests
* [ ] Rebase on master


lucafavatella:
Preparing rebased version.

PR #1035, opened 2018-04-25
Use proc_lib:spawn in aec_sync

hanssv:
Forgot to include this commit in the Sync PR yesterday - sorry for the noise.


hanssv:
There are issues building docker image at CircleCI - merging anyway.

PR #1036, opened 2018-04-25
Bump enoise to commit 14b6f46

hanssv:
Fixed a couple of crashes and this will give us less noise in the logs.


hanssv:
There are issues building docker images at CircleCI - merging anyway

PR #1038, opened 2018-04-25
Race in cleanup and missing case in aec_sync

hanssv:
Observed a crash on fast-integration network (and some unnecessary logging)


hanssv:
Still problems on CircleCI - no docker_image_build...

PR #1039, opened 2018-04-25
PT-156268183 sophia chain interaction

UlfNorell:
This PR adds primitives to Sophia for accessing the chain environment, making contract calls and spend transactions.

See `apps/aesophia/test/contracts/environment.aes` for the environment primitives and `remote_call.aes` and `factorial.aes` for examples calling other contracts. The primitive for contract calls is a bit bare-bones at the moment, since we don't have contract-as-classes implemented in the type checker yet.

This also fixes some problems with the Sophia ABI. It's now possible to have complex types in arguments and results of contract calls.


UlfNorell:
Hmmm, "All checks have failed."

> ===> Compiling /home/builder/epoch/_build/default/lib/erlexec/c_src/*.cpp
> ===> g++: fatal error: cannot specify -o with -c, -S or -E with multiple files
> compilation terminated.

Anyone have any insights as to what might have gone wrong?

apps/aecontract/src/aect_evm.erl
```diff
@@ -105,7 +105,7 @@ execute_call(#{ code := CodeAsHexBinString
            pre => #{}},
     TraceSpec =
         #{ trace_fun =>
-               fun(S,A) -> io_lib:format(S,A) end
+               fun(S,A) -> io:format(S,A) end
```
happi: Was this intentional?

UlfNorell: Yes. The `trace_fun` is used as `io:format/2` rather than `io_lib:format/2` (see `aevm_eeevm_state:trace_format/3`), so using `io_lib:format/2` is equivalent to `fun(_, _) -> ok end`.

PR #1041, opened 2018-04-25
[WIP] mine key & micro blocks - tests not ready

sennui:
don't merge, missing validation/instantiation flow, block drift, buffer

### DIFF
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -12,7 +12,7 @@
         , get_hash_at_height/1
         , hash_is_connected_to_genesis/1
         , hash_is_in_main_chain/1
-        , insert_block/2
+        , insert_block/1
         ]).
 
 %% For tests
@@ -34,9 +34,9 @@
 get_hash_at_height(Height) when is_integer(Height), Height >= 0 ->
     get_hash_at_height(Height, new_state_from_persistence()).
 
--spec insert_block(#block{}, block_type()) -> 'ok' | {'error', any()}.
-insert_block(Block, Type) ->
-    Node = wrap_block(Block, Type),
+-spec insert_block(#block{}) -> 'ok' | {'error', any()}.
+insert_block(Block) ->
+    Node = wrap_block(Block),
     try internal_insert(Node, Block)
     catch throw:?internal_error(What) -> {error, What}
     end.
@@ -81,8 +81,7 @@ hash_is_in_main_chain(Hash) ->
 new_state_from_persistence() ->
     Fun = fun() ->
                   #{ type                  => ?MODULE
-                   , top_block_hash    => aec_db:get_top_block_hash()
-                   , top_key_block_height  => aec_db:get_top_block_height()
+                   , top_block_hash        => aec_db:get_top_block_hash()
                    , genesis_block_hash    => aec_db:get_genesis_hash()
                    }
           end,
@@ -96,8 +95,7 @@ persist_state(State) ->
             case get_top_block_hash(State) of
                 undefined -> ok;
                 TopBlockHash ->
-                    aec_db:write_top_block_hash(TopBlockHash),
-                    aec_db:write_top_block_height(get_top_block_height(State))
+                    aec_db:write_top_block_hash(TopBlockHash)
             end
     end.
 
@@ -107,8 +105,6 @@ internal_error(What) ->
     throw(?internal_error(What)).
 
 get_genesis_hash(#{genesis_block_hash := GH}) -> GH.
-
-get_top_block_height(#{top_block_height := Height}) -> Height.
 
 get_top_block_hash(#{top_block_hash := H}) -> H.
 set_top_block_hash(H, State) when is_binary(H) -> State#{top_block_hash => H}.
@@ -121,6 +117,7 @@ set_top_block_hash(H, State) when is_binary(H) -> State#{top_block_hash => H}.
               , hash    :: binary()
               , height  :: pos_integer()
               , type    :: block_type()
+              , key_hash :: binary()
               }).
 
 hash(#node{hash = Hash}) -> Hash.
@@ -163,13 +160,13 @@ node_is_genesis(Node, #{genesis_block_hash := undefined}) ->
 node_is_genesis(Node, State) ->
     hash(Node) =:= get_genesis_hash(State).
 
-wrap_block(Block, Type) ->
+wrap_block(Block) ->
     Header = aec_blocks:to_header(Block),
     {ok, Hash} = aec_headers:hash_header(Header),
     #node{ header = Header
          , hash = Hash
          , height = aec_headers:height(Header)
-         , type = Type
+         , type = aec_blocks:type(Block)
          }.
 
 wrap_header(Header) ->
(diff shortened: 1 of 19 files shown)


apps/aecore/src/aec_chain_state.erl
```diff
@@ -34,9 +34,9 @@
 get_hash_at_height(Height) when is_integer(Height), Height >= 0 ->
     get_hash_at_height(Height, new_state_from_persistence()).
 
--spec insert_block(#block{}) -> 'ok' | {'error', any()}.
-insert_block(Block) ->
-    Node = wrap_block(Block),
+-spec insert_block(#block{}, block_type()) -> 'ok' | {'error', any()}.
+insert_block(Block, Type) ->
```
gorillainduction: I would prefer finding the type out from the block itself.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -151,19 +163,28 @@ node_is_genesis(Node, #{genesis_block_hash := undefined}) ->
...
 
 wrap_header(Header) ->
     {ok, Hash} = aec_headers:hash_header(Header),
     #node{ header = Header
          , hash = Hash
+         , height = aec_headers:height(Header)
+         , type = aec_blocks:type()
          }.
 
+set_height(#node{type = micro} = Node, #{top_key_block_height := Height}) ->
```
gorillainduction: This is not safe. You don't know that this micro block is at the top. It might even be on a different fork.

If the height is supposed to be inherited from the keyblock, you should check that there is a chain of microblocks to a keyblock and then get the height from there. I don't see why the height is not given in the micro blocks, and then is checked to match in a micro chain back to the key block.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -232,6 +255,8 @@ internal_insert(Node, Block) ->
         {ok, Old} -> internal_error({same_key_different_content, Node, Old})
     end.
 
+%% NG-INFO: height asserted at the beginning to be equal to key block height
+assert_previous_height(#node{type = micro}) -> ok;
```
gorillainduction: See comment above.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -245,6 +270,8 @@ assert_previous_height(Node) ->
 %% To assert the target calculation we need DeltaHeight headers counted
 %% backwards from the node we want to assert. If Height <= DeltaHeight
 %% we will need all headers back to genesis.
+%% NG-INFO: micro blocks are signed, so we don't validate the target
```
gorillainduction: But then the signature probably should be checked? But that might have happened elsewhere.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -364,11 +391,12 @@ update_next_state_tree_children([{Child, ForkId}|Left], Trees, Difficulty, Max,
 get_state_trees_in(Node, State) ->
     case node_is_genesis(Node, State) of
         true  -> {ok, aec_block_genesis:populated_trees(), 0, hash(Node)};
-        false -> db_find_state(prev_hash(Node))
+        false -> db_find_state(prev_hash(Node)) %% NG-INFO: this holds for NG without changes.
+                                                %% It causes micro-forks that we don't GC for now
```
gorillainduction: Comment is probably not at the correct place.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -440,6 +468,7 @@ find_fork_point(_Hash1, _Res1,_Hash2,_Res2) ->
 db_put_node(#block{} = Block, Hash) when is_binary(Hash) ->
     ok = aec_db:write_block(Block).
 
+%% NG-INFO Heigh/Hahs queries have sense in context of key blocks. For non-key height = 0
```
gorillainduction: Spelling of Heigh/Hahs

apps/aecore/src/aec_chain_state.erl
```diff
@@ -450,6 +479,7 @@ db_get_node(Hash) when is_binary(Hash) ->
     {ok, Node} = db_find_node(Hash),
     Node.
 
+%% NG-INFO Heigh/Hahs queries have sense in context of key blocks. For non-key height = 0
```
gorillainduction: Spelling.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -499,6 +529,7 @@ db_children(#node{} = Node) ->
      || Header <- aec_db:find_headers_at_height(Height + 1),
         aec_headers:prev_hash(Header) =:= Hash].
 
+db_node_has_sibling_blocks(#node{type = micro}) -> false;
```
gorillainduction: This is not true. You might have micro forks.

gorillainduction: Oh, but I also note that the function is not correctly implemented. Once it is fixed it should handle micro blocks just fine.

gorillainduction: There is no bug, I was too hasty.

gorillainduction:
In general, not having height on micro blocks is problematic from the aec_chain_state point of view. We need to have another way of finding the children of a micro-block. The current code only works if the blocks arrive in perfect order.

apps/aecore/src/aec_conductor.erl
```diff
@@ -625,9 +659,91 @@ retry_mining_with_new_nonce(Candidate, State) ->
...
+    aec_events:publish(start_micro_signing, Info),
+    Fun = fun() ->
+                  {aec_keys:sign(HeaderBin), HeaderBin}
+          end,
+    dispatch_worker(micro_signing, Fun, State).
+
+handle_micro_signing_reply(_Reply, #state{micro_block_candidate = undefined} = State) ->
+    %% Something invalidated the block candidate already.
+    start_micro_sleep(State);
+handle_micro_signing_reply({ok, {[Signature], HeaderBin}}, #state{} = State) ->
```
jur0: `aec_keys:sign(HeaderBin)` returns `{ok, {HeaderBin, Singatures}}` so this should be changed to:
```
handle_micro_signing_reply({ok, {HeaderBin, [Signature]}}, #state{} = State)
```

apps/aecore/src/aec_mining.erl
```diff
@@ -18,23 +18,39 @@
...
-                             list(aec_headers:header())) ->
+-spec create_micro_block_candidate(aec_blocks:block(), aec_blocks:block(), aec_trees:trees()) ->
                                     {ok, aec_blocks:block(), aec_pow:nonce()} |
                                     {error, term()}.
-create_block_candidate(TopBlock, TopBlockTrees, AdjHeaders) ->
-    create_block_candidate(get_txs_to_mine_in_pool(),
-                           TopBlock, TopBlockTrees,
-                           AdjHeaders).
+create_micro_block_candidate(TopBlock, CurrentKeyBlock, TopBlockTrees) ->
+    create_micro_block_candidate(get_txs_to_mine_in_pool(), TopBlock, CurrentKeyBlock, TopBlockTrees).
```
jur0: If this was `created_signed_micro_block_candidate(...)`, then quite a lot of code could be removed from `aec_conductor`, such as `start_micro_signing` and `handle_micro_signing_reply`. It's already done like that for [coinbase tx](https://github.com/aeternity/epoch/blob/98db894d5d02b513e83ff1f69f2747f038590bb6/apps/aecore/src/aec_mining.erl#L31-L47).

sennui: @jur0 please proceed with refactoring

sennui:
superseded by https://github.com/aeternity/epoch/compare/master...ng-develop-3

PR #1049, opened 2018-04-26
System tests hard fork adapted for spend tx with payload

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/157115203

### DIFF
--- a/system_test/aest_hard_fork_SUITE.erl
+++ b/system_test/aest_hard_fork_SUITE.erl
@@ -186,7 +186,7 @@ groups() ->
      {hard_fork_all_with_tx,
       [sequence],
       [
-       {group, hard_fork_old_chain_with_tx},
+       %% {group, hard_fork_old_chain_with_tx}, Not applicable as is in 0.12.0 because new node's user API creates only new spend tx.
        {group, hard_fork_new_chain_with_tx}
       ]},
      {hard_fork_old_chain_with_tx,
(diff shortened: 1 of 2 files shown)


system_test/aest_hard_fork_SUITE.erl
```diff
@@ -192,8 +192,8 @@ groups() ->
      {hard_fork_old_chain_with_tx,
       [sequence],
       [
-       new_node_can_receive_short_old_chain_from_old_node,
-       new_node_can_mine_spend_tx_on_old_chain_using_old_protocol
+       new_node_can_receive_short_old_chain_from_old_node
+       %% new_node_can_mine_spend_tx_on_old_chain_using_old_protocol Not applicable as is in 0.12.0 because new node's user API creates only new spend tx.
```
jur0: It'd be better to comment `{group, hard_fork_old_chain_with_tx}` in `hard_fork_all_with_tx` group.

lucafavatella: I now understand - thanks for clarifying - commenting the whole hard_fork_old_chain_with_tx as the left only test new_node_can_receive_short_old_chain_from_old_node is preparatory.

lucafavatella: @jur0 Addressed in force-pushed version of commit.

PR #1053, opened 2018-04-27
Improve function for running erl commands on docker

jur0:
* Return value is converted to erlang terms, so easier to pattern match
  * New function for formatting erl commands

### DIFF
--- a/system_test/aest_hard_fork_SUITE.erl
+++ b/system_test/aest_hard_fork_SUITE.erl
@@ -494,9 +494,10 @@ new_node_can_mine_old_spend_tx_without_payload_using_new_protocol(Cfg) ->
     {true, _} = {HeightToBeMinedWithOldProtocol > TopHeight,
                  {check_at_least_a_block_to_mine_using_old_protocol,
                   HeightToBeMinedWithOldProtocol}},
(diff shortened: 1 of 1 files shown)


system_test/aest_hard_fork_SUITE.erl
```diff
@@ -755,14 +751,22 @@ to_forms(DotEndingTokenLists) ->
...
+    lists:flatten(io_lib:format(Fmt, Params)).
+
+run_erl_cmd_on_node(NodeName, ErlCmd, Timeout, Cfg) ->
+    ct:log("Running Erlang command on node ~s:~n~s", [NodeName, ErlCmd]),
     Cmd = ["bin/epoch", "eval", ErlCmd],
     {0, Output} = aest_nodes:run_cmd_in_node_dir(NodeName, Cmd, Timeout, Cfg),
-    ?assertEqual(ExpectedOutput, string:trim(Output)),
-    ct:log("Run Erlang command on node ~s", [NodeName]),
-    ok.
+    Result = eval_expression(Output),
```
ThomasArts: Let us hope we don't accidentally have "file:delete_all()" as output of the docker...

PR #1054, opened 2018-04-27
PT-156914951 Swagger HTTP client improvements

jur0:
These changes are needed for using the client in system tests (so far it's used locally in a smart contract system test).

### DIFF
--- a/apps/aecore/test/aecore_sync_SUITE.erl
+++ b/apps/aecore/test/aecore_sync_SUITE.erl
@@ -97,7 +97,7 @@ init_per_suite(Config) ->
     ct:log("Environment = ~p", [[{args, init:get_arguments()},
                                  {node, node()},
                                  {cookie, erlang:get_cookie()}]]),
-    Forks = aecore_suite_utils:forks(), 
+    Forks = aecore_suite_utils:forks(),
     DefCfg = #{<<"metrics">> =>
                    #{<<"rules">> =>
                          [#{<<"name">> => <<"ae.epoch.system.**">>,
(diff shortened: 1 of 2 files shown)


apps/aehttp/test/aehttp_client.erl
```diff
@@ -2,67 +2,57 @@
...
     end;
-process_response({error, _} = Error) ->
+process_response({error, _} = Error, _Cfg) ->
     Error.
 
-maybe_encode_params(Params, Cfg) ->
-    case proplists:get_value(encode_base58c, Cfg) of
-        true -> encode_params(Params);
-        _ -> Params
+log(Fmt, Params, Cfg) ->
```
sylane: Why not get a logging function?

jur0: Do you mean passing log function in `Cfg`? That could be an option, too. But currently this client is going to be used in tests where there is only CT logging or no logging (QuichCheck tests).

sylane: Just saying because this how system test actually works.

apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -511,12 +511,11 @@ new_tx(#{node1 := N1, node2 := N2, amount := Am, fee := Fee} = M) ->
                     [ [<<"http">>, <<"internal">>, <<"port">>],
                       aehttp, [internal, port], 8143], 5000),
     Params = #{sender_pubkey => PK1,
```
velzevur: Since this is the internal API and sender is not required - why do we send it at all? I see we are sending the raw binary...

jur0: I've just used the new HTTP client here. I keep all the params as they were before.

PR #1059, opened 2018-04-27
Asking the wrong chain for info

hanssv:
Logic was backwards, it works until we find a fork that is further than 1 block apart :-(


lucafavatella:
I see aehttp_integration_SUITE socket_closed_remotely in https://circleci.com/gh/aeternity/epoch/9763 - restarting as unrelated to PR.

PR #1060, opened 2018-04-27
Add log checking to system tests

sylane:


system_test/helpers/aest_nodes_mgr.erl
```diff
@@ -139,6 +146,27 @@ mgr_cleanup(State) ->
...
+mgr_scan_logs_for_errors(#{nodes := Nodes} = State) ->
+    maps:fold(fun(NodeName, {Backend, NodeState}, Result) ->
+        LogPath = Backend:get_log_path(NodeState),
+        LogFile = binary_to_list(filename:join(LogPath, "epoch.log")),
+        case filelib:is_file(LogFile) of
+            false -> Result;
+            true ->
+                Command = "grep '\\[error\\]' '" ++ LogFile ++ "'"
+                       %% Ingore errors from watchdog/eper due to dead process
+                       ++ "| grep -v 'emulator Error in process <[0-9.]*> "
```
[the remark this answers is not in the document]
sylane: It is a problem, the issue is that the more specific content of the error is in a new line, it makes it harder to grep for it...

system_test/helpers/aest_nodes_mgr.erl
```diff
@@ -139,6 +146,27 @@ mgr_cleanup(State) ->
             State#{backends => [], nodes => #{}}
     end.
 
+mgr_scan_logs_for_errors(#{nodes := Nodes} = State) ->
```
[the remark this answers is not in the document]
sylane: Because this is not a performance issue and the result is a lot simpler...

system_test/aest_commands_SUITE.erl
```diff
@@ -43,7 +43,7 @@ init_per_testcase(_TC, Config) ->
     aest_nodes:ct_setup(Config).
 
 end_per_testcase(_TC, Config) ->
-    aest_nodes:ct_cleanup(Config).
+    ok = aest_nodes:ct_cleanup(Config).
```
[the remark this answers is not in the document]
sylane: I can do that...

system_test/helpers/aest_nodes_mgr.erl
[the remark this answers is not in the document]
sylane: I'll do it and you will see :)

sylane:
@eproxus I tried to clean up a bit and fix the timeout sec/ms confusion

system_test/helpers/aest_nodes.erl
```diff
@@ -355,3 +345,21 @@ decode_json(Data) ->
...
     end.
+
+scan_logs_for_errors() ->
+    Logs = aest_nodes_mgr:get_log_paths(),
+    maps:fold(fun(NodeName, LogPath, Result) ->
+        LogFile = binary_to_list(filename:join(LogPath, "epoch.log")),
+        case filelib:is_file(LogFile) of
+            false -> Result;
+            true ->
+                Command = "grep '\\[error\\]' '" ++ LogFile ++ "'",
```
lucafavatella: This is a start though I am concerned that check relies a lot on this regexp/string. Not sure if there are any other meaningful strings.

I wonder if there are alternatives:
* Review that all nodes are configured to log crash.log, and check it is empty file (so not need to grep);
* As Operator may want, enable by config logging all log entries not meant to exist (error or higher?) in separate log file then check that log file is empty?

system_test/helpers/aest_docker_api.erl
```diff
@@ -117,14 +117,12 @@ start_container(ID) ->
 
 stop_container(ID, Opts) ->
     STDefault = ?CONTAINER_STOP_TIMEOUT,
-    {Query, HTDefault} = case maps:get(soft_timeout, Opts, STDefault) of
+    {Query, ReqTimeout} = case maps:get(soft_timeout, Opts, STDefault) of
         infinity -> {#{t => ?CONTAINER_STOP_INFINTY}, infinity};
-        STSecs ->
-            {#{t => STSecs}, STSecs + ?CONTAINER_KILL_TIMEOUT_EXTRA}
-    end,
-    ReqTimeout = case maps:get(hard_timeout, Opts, HTDefault) of
```
lucafavatella: As hard_timeout now unused it shall be removed from https://github.com/aeternity/epoch/pull/1060/files#diff-e2b7668405d6906af2515a075fe330c1R93 ?

system_test/helpers/aest_nodes.erl
```diff
@@ -41,6 +41,68 @@
 -define(NODE_TEARDOWN_TIMEOUT, 0).
 -define(DEFAULT_HTTP_TIMEOUT, 3000).
 
+%% AWK script to keep only error, critical, alert and emergency log lines with
+%% all the extra lines following the log lines.
+%% In addition, it filter out the error lines caused by eper/watchdog.
+%% Update when https://www.pivotaltracker.com/story/show/157215253 is fixed.
```
lucafavatella: Could you please ensure the update of these scripts is tracked in Pivotal?

eproxus: Should the scripts even cater to existing bugs? Shouldn't we just highlight the bugs with the tests instead?

sylane: If they don't it would just sit around until these bugs are fixed... Or do you suggest to merge something in master that would make system tests fail randomly?

system_test/helpers/aest_docker_api.erl
sylane: Good catch, it is not that hard_timeout is not used anymore, it was a something wrong in the PR.

system_test/helpers/aest_nodes.erl
eproxus: I would suggest that, to highlight the problems and to avoid having to remember to fix it later when the bugs are fixed. But I don't have a strong opinion on it 🙂

PR #1061, opened 2018-04-27
Pt 155251857 test solidity on chain

happi:
Define ABI in vm_version and dispatch contract calls on this.
Allso run init code in create_contract transaction.


happi:
When merged to master we should also merge https://github.com/aeternity/protocol/pull/61

PR #1062, opened 2018-04-28
Pt 156600645 resolve name at current height

thepiwo:
Resolves names for SpendTx at current height while mining and validating.
 - account_pubkey pointer from naming system is used
 - if name is found but no account_pubkey, the transaction is rejected
 - if name is not valid, the transaction is rejected

https://www.pivotaltracker.com/story/show/156600645

### DIFF
--- a/apps/aecore/src/aec_chain.erl
+++ b/apps/aecore/src/aec_chain.erl
@@ -134,7 +134,7 @@ name_entry(Name) ->
                                         {error, atom()}.
 resolve_name(Type, Name) ->
     case get_top_state() of
-        {ok, Trees} -> aens:resolve(Type, Name, aec_trees:ns(Trees));
+        {ok, Trees} -> aens:resolve_encoded(Type, Name, aec_trees:ns(Trees));
         error -> {error, no_state_trees}
     end.
 
--- a/apps/aecore/src/aec_spend_tx.erl
+++ b/apps/aecore/src/aec_spend_tx.erl
@@ -90,19 +90,23 @@ payload(#spend_tx{payload = Payload}) ->
 
 -spec check(tx(), aetx:tx_context(), aec_trees:trees(), height(), non_neg_integer()) -> {ok, aec_trees:trees()} | {error, term()}.
 check(#spend_tx{} = SpendTx, _Context, Trees0, Height, _ConsensusVersion) ->
-    RecipientPubkey = recipient(SpendTx),
+    RecipientPubkeyOrName = recipient(SpendTx),
     Checks = [fun check_tx_fee/3,
               fun check_sender_account/3],
     case aeu_validation:run(Checks, [SpendTx, Trees0, Height]) of
+        {error, _Reason} = Error -> Error;
         ok ->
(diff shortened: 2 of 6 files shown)


thepiwo:
as discussed with @velzevur, maybe we should specify explicitly if we want to send a spend_tx (or others using name resolvement) to a pubkey or name, as the binary handling could lead to conflicts.

sennui:
@thepiwo ...that would be huge change. We are just after adding payload to spend. Maybe we can do short call about it? I might be to fast with that feature.

sennui:
@thepiwo contracts have more logic and that would be another mechanism to drive it. For regular transaction I don't see the point.

thepiwo:
@sennui @happi ok, I'd propose to squash and keep this to the scope it is. And then have various other resolvements as follow up task for me, so could you review?

apps/aecore/src/aec_chain.erl
```diff
@@ -134,7 +134,7 @@ name_entry(Name) ->
                                         {error, atom()}.
 resolve_name(Type, Name) ->
     case get_top_state() of
-        {ok, Trees} -> aens:resolve(Type, Name, aec_trees:ns(Trees));
+        {ok, Trees} -> aens:resolve_encoded(Type, Name, aec_trees:ns(Trees));
```
sennui: please apply changes to aeo_query_tx. With this change it may revceive human readable string instead of translated hash

lucafavatella:
I understand not suitable to merge yet, and that @thepiwo may address comment. I take the liberty of labeling as WIP for easing filtering.

thepiwo:
Yes, thank you @lucafavatella I will clarify my state tomorrow in the call.

apps/aecore/src/aec_spend_tx.erl
```diff
@@ -114,16 +118,20 @@ signers(#spend_tx{sender = SenderPubKey}) -> [SenderPubKey].
...
                   fee = Fee,
                   nonce = Nonce}, _Context, Trees0, Height, _ConsensusVersion) ->
     AccountsTrees0 = aec_trees:accounts(Trees0),
+    NSTrees0 = aec_trees:ns(Trees0),
 
     {value, SenderAccount0} = aec_accounts_trees:lookup(SenderPubkey, AccountsTrees0),
     {ok, SenderAccount} = aec_accounts:spend(SenderAccount0, Amount + Fee, Nonce, Height),
     AccountsTrees1 = aec_accounts_trees:enter(SenderAccount, AccountsTrees0),
 
+    % possible errors have been filtered previously which check -> ensure_account_at_height
```
zp-sd: with check?

thepiwo:
so, I will close this for now and include all possible name resolvements and change requests in a future PR

apps/aecore/src/aec_spend_tx.erl
thepiwo: the check/5 function ensures that all accounts that could possibly be resolved, are initialized in the accounts tree already, otherwise this would fail

PR #1064, opened 2018-04-30
(WIP) Pt 156915425 store calls in separate state tree

happi:
WIP since it is based on unmerged branch.

[Finishes #156915425,  #156915336]


UlfNorell:
What are the consequences of only keeping calls for one block? Does it mean you can't get at the result after more blocks are added? What are the benefits?

happi:
The benefit is that we are not filling up the state space with call/return values.
The plan is that you can put a web socket watcher on the node and get the return value when the transaction goes through (not implemented).
Ethereum have no return values for top level calls to save space. Now we can have return values and still save space.

happi:
We might not need height in the call state tree any more.

PR #1066, opened 2018-04-30
PT 157189489 Channels close TCP connections upon FSM death

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/157189489)
Test opens a channel, updates it and closes the WSs. This should result in FSMs death and release of the socket being used. The state channel itself is not closed. Then test opens a second channel on the same port and host to verify that the port is available.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -602,10 +602,10 @@ error_binary(E) when is_atom(E) ->
     atom_to_binary(E, latin1).
 
 
-terminate(_Reason, _State, Data) ->
+terminate(Reason, _State, Data) ->
     #data{session = Sn} = Data,
     aesc_session_noise:close(Sn),
-    report_info(died, Data),
+    report_info({died, Reason}, Data),
     ok.
 
 code_change(_OldVsn, OldState, OldData, _Extra) ->
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -603,8 +603,6 @@ error_binary(E) when is_atom(E) ->
 
 
 terminate(Reason, _State, Data) ->
-    #data{session = Sn} = Data,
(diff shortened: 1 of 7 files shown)


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -207,6 +208,11 @@ shutdown(Fsm) ->
     lager:debug("shutdown(~p)", [Fsm]),
     gen_statem:call(Fsm, shutdown).
 
+client_died(Fsm) ->
+    %TODO: possibility for reconnect
+    lager:debug("client died(~p)", [Fsm]),
+    ok = gen_statem:stop(Fsm).
```
ThomasArts: This is still the nice way of stopping it, a exit(Pid, kill) would be rougher. But since you don't care about whether or not you can recover the state nicely, this should work fine as a test.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -588,7 +594,10 @@ error_binary(E) when is_atom(E) ->
     atom_to_binary(E, latin1).
 
 
-terminate(_Reason, _State, _Data) ->
+terminate(_Reason, _State, Data) ->
```
ThomasArts: I doubt you get here when the FSM really crashes.

velzevur:
Rebased on top of latests feature-state-channels and resolved conflicts

velzevur:
@ThomasArts WRT to linking, actually `aesc_session_noise:connect/3` and `aesc_session_noise:connect/2` follow the `enoise` interface: there are `enoise:connect/3` and `enoise:accept/2` and the `aesc_session_noise` functions are basically wrapper around the enoise. All of those spawn linked processes (even though their names lack this)

velzevur:
Added a monitoring of the FSM so if it dies even with a `normal` reason, the aesc_session_noise is informed and dies as well. This resolves the lingering TCP connections.

PR #1067, opened 2018-05-01
Pt 156915425 store calls in separate state tree

happi:
Related to https://github.com/aeternity/protocol/pull/65


apps/aecore/src/aec_db.erl
```diff
@@ -321,6 +328,12 @@ find_oracles_cache_node(Hash) ->
     end.
 
 
+find_calls_node(Hash) ->
+    case ?t(mnesia:read(aec_calls_state, Hash)) of
```
lucafavatella: aec_call_state ?

happi: Good catch Ill see if I can add a test case.

PR #1069, opened 2018-05-01
Pt 156915425 store calls in separate state tree v2

happi:
Cleaned up branch.


apps/aecore/src/aec_db.erl
```diff
@@ -320,6 +327,11 @@ find_oracles_cache_node(Hash) ->
         [] -> none
     end.
 
+find_calls_node(Hash) ->
+    case ?t(mnesia:read(aec_call_state, Hash)) of
```
lucafavatella: I see addresses https://github.com/aeternity/epoch/pull/1067#discussion_r185198756

lucafavatella:
I see this PR had superseded #1064

PR #1074, opened 2018-05-02
Fix peer keys deployment

dincho:
- always run facts setup regardless of ansible tags used
- fix peer keys deployment destination

Before the fix the keys was deployed to wrong destination: `keys/{ip}.pub` and respective prv key file, while it should be `peer_key.pub` and respective prv key.


dincho:
> Please confirm static&documented testnet peer pubkeys deployed and picked up by running testnet nodes at correct IPs.

I just redeployed eu-east seed node and it looks fine. Other seed nodes should be untouched ATM, and this changes should not change their keys (because it's the same)

PR #1078, opened 2018-05-03
Improve intergration test of contracts

jur0:
The test case is dependent on the previous test cases, as it's expected the miner has some balance at the beginning of the test case. All the rpc calls are removed apart from getting the nonce, signing the tx and starting/stopping miner.

### DIFF
--- a/apps/aehttp/test/aehttp_integration_SUITE.erl
+++ b/apps/aehttp/test/aehttp_integration_SUITE.erl
@@ -397,7 +397,7 @@ init_per_suite(Config) ->
     ct:log("Environment = ~p", [[{args, init:get_arguments()},
                                  {node, node()},
                                  {cookie, erlang:get_cookie()}]]),
-    Forks = aecore_suite_utils:forks(), 
+    Forks = aecore_suite_utils:forks(),
 
     aecore_suite_utils:create_configs(Config1, #{<<"chain">> =>
                                                  #{<<"persist">> => true,
@@ -647,8 +647,8 @@ contract_transactions(_Config) ->
     ValidEncoded = #{ owner => MinerAddress,
                       code => Code,
                       vm_version => 1,
-                      deposit => 5,
-                      amount => 10,
+                      deposit => 2,
+                      amount => 1,
                       gas => 30,
                       gas_price => 3,
                       fee => 1,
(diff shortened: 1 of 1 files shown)


apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -705,14 +701,22 @@ contract_transactions(_Config) ->
                                fun get_contract_call/1,
                                fun aect_call_tx:new/1, MinerPubkey),
 
+    {ok, 200, #{<<"tx">> := EncodedUnsignedContractCallTx,
+                <<"tx_hash">> := ContractCallTxHash}} = get_contract_call(ContractCallEncoded),
+    sign_and_post_tx(EncodedUnsignedContractCallTx),
+
+    % mine a block
+    aecore_suite_utils:mine_blocks(aecore_suite_utils:node_name(?NODE), 2),
+    tx_is_mined_test(MinerAddress, ContractCallTxHash),
```
lucafavatella: I see contract call tx with already encoded call data was not mined in chain - now it is posted to node and mined in chain.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -727,6 +731,15 @@ contract_transactions(_Config) ->
...
                                fun aect_call_tx:new/1, MinerPubkey),
 
+    {ok, 200, #{<<"tx">> := EncodedUnsignedContractCallComputeTx,
+                <<"tx_hash">> := ContractCallComputeTxHash}} =
+        get_contract_call_compute(ComputeCCallEncoded),
+    sign_and_post_tx(EncodedUnsignedContractCallComputeTx),
+
+    % mine a block
+    aecore_suite_utils:mine_blocks(aecore_suite_utils:node_name(?NODE), 2),
+    tx_is_mined_test(MinerAddress, ContractCallComputeTxHash),
```
lucafavatella: I see contract call tx created with function and arguments (non-encoded call data) was not mined in chain - now it is posted to node and mined in chain.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -3788,13 +3774,19 @@ open_websockets_count() ->
...
+sign_and_post_tx(EncodedUnsignedTx, Opts) ->
     {ok, SerializedUnsignedTx} = aec_base58c:safe_decode(transaction, EncodedUnsignedTx),
     UnsignedTx = aetx:deserialize_from_binary(SerializedUnsignedTx),
     {ok, SignedTx} = rpc(aec_keys, sign, [UnsignedTx]),
     SerializedTx = aetx_sign:serialize_to_binary(SignedTx),
     {ok, 200, _} = post_tx(aec_base58c:encode(transaction, SerializedTx)),
     % create_contract_tx is in mempool
-    {ok, [SignedTx]} = rpc(aec_tx_pool, peek, [infinity]), % same tx
+    case proplists:get_value(mempool_check, Opts) of
+        true -> {ok, [SignedTx]} = rpc(aec_tx_pool, peek, [infinity]); % same tx
```
lucafavatella: This may be replaced with call(s) to user HTTP API. I am thinking of [get_account_transactions with pending txs](https://github.com/aeternity/epoch/blob/e2589afebe6acefaddb425ca77768bdc87db9a33/config/swagger.yaml#L1179) checking [height `-1`](https://github.com/aeternity/epoch/blob/e2589afebe6acefaddb425ca77768bdc87db9a33/apps/aetx/src/aetx_sign.erl#L158).

jur0: Replaced in 45c0d3d.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -3787,16 +3769,27 @@ open_websockets_count() ->
...
+    #{<<"hash">> := TxHash} = aetx_sign:serialize_for_client_pending(json, SignedTx),
+    %% Check tx is in mempool.
+    {ok, [TxHash]} = aec_test_utils:wait_for_it_or_timeout(fun() -> mempool_txs(AccountPubKey) end,
+                                                           [TxHash], 5000),
     ok.
 
+mempool_txs(AccountPubKey) ->
+    {ok, 200, #{<<"transactions">> := Txs}} =
+        get_account_transactions(AccountPubKey, tx_encoding_param(json)),
+    lists:filtermap(fun(#{<<"block_hash">> := <<"none">>, <<"hash">> := TxHash}) ->
```
lucafavatella: I see https://github.com/aeternity/epoch/blob/e2589afebe6acefaddb425ca77768bdc87db9a33/apps/aetx/src/aetx_sign.erl#L163

PR #1080, opened 2018-05-03
PT-156609771: Test synchronizing far ahead node

eproxus:
A new performance suite is added, and the existing sync speed test is moved there. A new startup speed test is added.

Faster `smoke-test` target will be added in a separate PR.

### DIFF
--- a/system_test/helpers/aest_docker_api.erl
+++ b/system_test/helpers/aest_docker_api.erl
@@ -23,6 +23,7 @@
 -export([exec/3]).
 -export([extract_archive/3]).
 -export([container_logs/1]).
+-export([commit/2]).
 
 %=== MACROS ====================================================================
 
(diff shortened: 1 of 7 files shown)


sylane:
Maybe some separate testing for long-running performance tests ? `make perf-tests` ?

system_test/helpers/aest_docker_api.erl
```diff
@@ -207,6 +208,11 @@ container_logs(ID) ->
             throw({docker_error, maps:get(message, decode_json(Response))})
     end.
 
+commit(ID, Tag) ->
```
ThomasArts: Probably a comment what commit is used for? Of course, I could read the manual of dockers.

ThomasArts:
We have 8 hours for all system tests together (nightly run). If we smartly re-use this docker container, we could benefit from creating a long chain once.

system_test/helpers/aest_docker_api.erl
eproxus: Commit is the standard command in Docker to create an image from a container file system

eproxus:
@sylane No, it's coming in a separate PR

eproxus:
See the description

sylane:
I should really be more attentive...

PR #1081, opened 2018-05-04
PT-154733661 switch to curve25519

hanssv:


apps/aecore/src/aec_keys.erl
```diff
@@ -278,131 +240,132 @@ peer_key_filenames(KeysDir) -> p_gen_peer_filename(KeysDir).
...
-    <<0:(Pad*8), Bin/binary>>.
+p_save_keys(Pwd, PubFile, PubKey, PrivFile, PrivKey) ->
+    EncPub = encrypt_key(Pwd, PubKey),
+    EncPriv = encrypt_key(Pwd, PrivKey),
+    ok = to_local_dir(PubFile, EncPub),
+    ok = to_local_dir(PrivFile, EncPriv),
+    {EncPub, EncPriv}.
 
-p_gen_new_peer(Password, PeerPubFile, PeerPrivFile) ->
-    KeyPair = enoise_keypair:new(dh25519),
```
lucafavatella: I understand peer keys not changed.

docs/release-notes/RELEASE-NOTES-0.13.0.md
```diff
@@ -2,6 +2,8 @@
 
 [This release][this-release] is focused on TODOFILLMEIN.
 It:
+* Change the cryptographic signing to use Curve25519. This means all old node/signing keys are not useable
+  anymore (`key/key.pub` is replaced by `sign_key/sign_key.pub`).
```
lucafavatella: Lower in this document, 'even if you keep your key pair" suggests it may be possible to keep keypair, that may be misleading.

hanssv: Yes, I saw that comment. I think temporarily removing it for this release would make sense.

hanssv:
The comment is gone in Ulf's working branch. It meant rather that it wouldn't work with 65 byte addresses and that all is good with 32 byte addresses!

PR #1084, opened 2018-05-04
Make node stop more reliable

sylane:
* Fix `new_node_joins_network` test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ system-test:
 
 system-test-deps: | system_test/aest_hard_fork_SUITE_data/db/mnesia_ae-uat-epoch_backup.tar
 	$(info The downloaded DB backup is not checked for freshness.)
-	docker pull aeternity/epoch:v0.11.1
+#	docker pull aeternity/epoch:latest
 
 system_test/aest_hard_fork_SUITE_data/db/mnesia_ae-uat-epoch_backup.tar: system_test/aest_hard_fork_SUITE_data/db/mnesia_ae-uat-epoch_backup
 	tar -c -C $(<D) -f $@ $(<F)
--- a/system_test/helpers/aest_docker.erl
+++ b/system_test/helpers/aest_docker.erl
@@ -415,14 +415,6 @@ write_template(TemplateFile, OutputFile, Context) ->
     ok = filelib:ensure_dir(OutputFile),
     file:write_file(OutputFile, Data).
 
-wait_stopped(Id, Timeout) -> wait_stopped(Id, Timeout, os:timestamp()).
-
-wait_stopped(Id, Timeout, StartTime) ->
-    case is_running(Id) of
-        false -> ok;
-        true -> maybe_continue_waiting(Id, Timeout, StartTime)
-    end.
-
 is_running(Id) -> is_running(Id, 5).
 
 is_running(_Id, 0) -> error(retry_exausted);
@@ -453,7 +445,8 @@ attempt_epoch_stop(#{container_id := ID, hostname := Name} = NodeState, Timeout)
     ok.
 
 %% Sometime the stop command do not succeed...
-retry_epoch_stop(_NodeState, _ID, _Cmd, _Opts, 0) -> error;
+retry_epoch_stop(_NodeState, _ID, _Cmd, _Opts, 0) ->
+    error(retry_exausted);
 retry_epoch_stop(NodeState, ID, Cmd, Opts, Retry) ->
     #{container_id := ID, hostname := Name} = NodeState,
     case aest_docker_api:exec(ID, Cmd, Opts) of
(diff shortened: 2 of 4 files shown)


system_test/helpers/aest_docker.erl
```diff
@@ -450,6 +448,19 @@ attempt_epoch_stop(#{container_id := ID, hostname := Name} = NodeState, Timeout)
     end,
     ok.
 
+%% Sometime the stop command do not succeed...
+retry_epoch_stop(_NodeState, _ID, _Cmd, _Opts, 0) -> error;
```
eproxus: Where does the atom `error` go? Is it checked for anywhere? If this is the public API we should use an exception instead

system_test/helpers/aest_docker_api.erl
```diff
@@ -144,6 +145,17 @@ kill_container(ID) ->
             throw({docker_error, maps:get(message, Response)})
     end.
 
+wait_stopped(ID, Timeout) ->
+    Query = #{condition => "not-running"},
+    PostOpts = #{timeout => Timeout},
+    case docker_post([containers, ID, wait], Query, #{}, PostOpts) of
+        {error, timeout} -> timeout;
```
eproxus: Same as `error` above, I don't see any code checking this. Should be a throw instead?

system_test/helpers/aest_docker.erl
sylane: I understand I'll throw an error.

system_test/helpers/aest_docker_api.erl
sylane: The timeout result is actually used in `aest_docker:291`.

eproxus: Ah, missed that. Looks okay to me

system_test/helpers/aest_docker.erl
```diff
@@ -450,15 +440,18 @@ attempt_epoch_stop(#{container_id := ID, hostname := Name} = NodeState, Timeout)
...
-maybe_continue_waiting(Id, infinity, StartTime) ->
-    timer:sleep(100),
-    wait_stopped(Id, infinity, StartTime);
-maybe_continue_waiting(Id, Timeout, StartTime) ->
-    case timer:now_diff(os:timestamp(), StartTime) > (1000 * Timeout) of
-        true -> timeout;
-        false ->
-            timer:sleep(200),
-            wait_stopped(Id, Timeout, StartTime)
+%% Sometime the stop command do not succeed...
```
eproxus: Personally I would remove this comment 🙂

sylane: I thought one could wonder why we are retrying the stop command...

eproxus: The comment doesn't actually explain why though

sylane: Because I don't really know why :/ Do we want to spend extra time to figure out why the command sometimes fail without returning any value?

eproxus: No, not really 🙂  Keep it then

sylane:
The test is not related. @ThomasArts test is triggering a crash that is not resolved by this. This PR is the result of investigating the crash, while trying to reproduce it with an already existing chain I found that sometime the stop command would do nothing and this could lead to a timeout.

Makefile
```diff
@@ -164,7 +164,7 @@ system-test:
 
 system-test-deps: | system_test/aest_hard_fork_SUITE_data/db/mnesia_ae-uat-epoch_backup.tar
 	$(info The downloaded DB backup is not checked for freshness.)
-	docker pull aeternity/epoch:v0.11.1
+#	docker pull aeternity/epoch:latest
```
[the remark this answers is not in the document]
sylane: `epoch:local` cannot be pulled, this is commented out because right now the test doesn't sync with latest because the genesis block changed. When this is not the case anymore this should be uncommented and latest should be used in `new_node_joins_network` again.

system_test/helpers/aest_docker.erl
```diff
@@ -450,6 +448,19 @@ attempt_epoch_stop(#{container_id := ID, hostname := Name} = NodeState, Timeout)
...
     ok.
 
+%% Sometime the stop command do not succeed...
+retry_epoch_stop(_NodeState, _ID, _Cmd, _Opts, 0) -> error;
+retry_epoch_stop(NodeState, ID, Cmd, Opts, Retry) ->
+    #{container_id := ID, hostname := Name} = NodeState,
+    case aest_docker_api:exec(ID, Cmd, Opts) of
+        {ok, 137, <<"ok\r\n">>} -> ok;
+        {ok, Status, Res} ->
+            CmdStr = lists:join($ , Cmd),
```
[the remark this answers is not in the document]
sylane: This is no my code, I just refactored it. This is a space character, if I quote it with `'` it would become an atom and if I quote it with `"` it would become a list.

PR #1085, opened 2018-05-04
[RFC] Application of txs only post genesis

lucafavatella:
This is a preparatory PR in scope of https://www.pivotaltracker.com/story/show/157214052 and I am opening it for receiving comments.

The problem is that when removing coinbase tx (to be done in future PR for https://www.pivotaltracker.com/story/show/157214052) the reward of the miner needs to be assigned as part of trees transformation, hence attempting to apply the genesis txs on the genesis trees would change the trees.

I recommend not to merge this PR yet, because:
* It does not bring any benefit in its own;
* I did not perform extensive testing of it yet (e.g. is attempt to post genesis block from a peer covered by tests?);
* A handful of tests fail (I did not validate yet if the tests or the code are wrong);
* The branch contains a couple of test-only commits that shall be removed (enabling aevm_tests and Python UAT tests).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -480,6 +480,8 @@ workflows:
           requires:
             - build
 
+      - aevm_tests: {}
+
       - static_analysis:
           requires:
             - build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -479,9 +479,6 @@ workflows:
       - uat_tests:
           requires:
             - build
-          filters:
-            branches:
-              only: master
 
       - static_analysis:
           requires:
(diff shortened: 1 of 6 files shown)


gorillainduction:
Looks reasonable.

sennui:
@lucafavatella but what is wrong with rewarding miner of the genesis block?

We had ticket to research mining vs hashing the genesis block - now with NG it's clear that we have to mine it (send account.json in micro block and then approve it with another key block). Why we need special case for that?

To sum up tho, this is good new functionality. If ae must not mine (due to legal reasons) that will fix such default behaviour.

lucafavatella:
Thanks for the comments.

PR #1086, opened 2018-05-05
Integrate system tests to CircleCI

dincho:
https://www.pivotaltracker.com/story/show/156683319

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,12 +2,23 @@ version: 2
 
 references:
   container_config: &container_config
+    setup_remote_docker:
+      docker_layer_caching: true
     docker:
       - image: aeternity/builder
         user: builder
     working_directory: ~/epoch
     environment:
+      # CircleCI supported versions: https://circleci.com/docs/2.0/building-docker-images/#docker-version
+      # Official docker versions: https://download.docker.com/linux/static/stable/x86_64/
       DOCKER_CLIENT_VERSION: "17.09.0-ce"
+      DOCKERHUB_REPO: aeternity/epoch
+
+  machine_config: &machine_config
+    machine:
+      docker_layer_caching: true
+    working_directory: ~/epoch
+    environment:
       DOCKERHUB_REPO: aeternity/epoch
 
   setup_remote_docker: &setup_remote_docker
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -2,14 +2,25 @@ version: 2
 
 references:
   container_config: &container_config
+    setup_remote_docker:
```
lucafavatella: I see [this is step](https://circleci.com/docs/2.0/configuration-reference/#setup_remote_docker) so unclear how this here.

.circleci/config.yml
```diff
@@ -24,6 +33,40 @@ references:
         tar -xz -C /tmp -f /tmp/docker-${DOCKER_CLIENT_VERSION:?}.tgz
         mkdir ~/bin && mv /tmp/docker/docker ~/bin
 
+  install_os_deps: &install_os_deps
+    run:
+      name: Install OS dependancies
+      command: |
+        sudo apt-get -qq -y install git curl \
```
lucafavatella: Could you please add to each of these install_* portions a comment with versioned link of where this setup is taken from? E.g. in this line https://github.com/aeternity/docker-builder/blob/0db5b300f251460e51e63ad5530a07896864f46b/Dockerfile#L4-L5

This would prevent Dev/DevOps people to amend this file as if it were the canonical copy.

.circleci/config.yml
```diff
@@ -527,6 +608,13 @@ workflows:
             branches:
               only: master
 
+      - docker_system_tests:
```
lucafavatella: I expect this to be removed before merge.

.circleci/config.yml
```diff
@@ -440,6 +440,7 @@ jobs:
             sudo -u ubuntu -E -H make system-test-deps
       - run:
           name: System Tests
+          no_output_timeout: 2h
```
lucafavatella: I see https://circleci.com/docs/2.0/configuration-reference/#run
Ok.

.circleci/config.yml
dincho: cleanup in 41feeaf

.circleci/config.yml
dincho: Added comment in dffc9bf

PR #1087, opened 2018-05-05
Pt 154664084 save contract state in tree

happi:
Saving store as a MPtree. Comments welcome.


apps/aecontract/src/aect_contracts.erl
```diff
@@ -96,6 +100,11 @@
 id(C) ->
   pubkey(C).
 
+-spec store_id(contract()) -> id().
+store_id(C) ->
+    CId = pubkey(C),
+    << CId/binary, ?STORE_PREFIX>>.
```
gorillainduction: The `?STORE_PREFIX` placed as a suffix of the contract ID. I see why, but is should have a comment explaining it.

apps/aecontract/src/aect_state_tree.erl
```diff
@@ -51,24 +52,74 @@ insert_contract(Contract, Tree = #contract_tree{ contracts = CtTree }) ->
...
-    Tree#contract_tree{ contracts = CtTree1 }.
+    CtTree2    = insert_store(Contract, CtTree1),
+    Tree#contract_tree{ contracts = CtTree2 }.
+
+insert_store(Contract, CtTree) ->
+    Id = aect_contracts:store_id(Contract),
+    Store = aect_contracts:state(Contract),
+    insert_store_nodes(Id, Store, CtTree).
+
+insert_store_nodes(Prefix, Store, CtTree) ->
```
gorillainduction: If I remember correctly, writing `0` to the store, will end up deleting the store position. If this is is implemented by removing the store position in the map, this insertion into the MPT will preserve the deleted value (as the previous value).

This needs to be special treated if the semantics is to be preserved. One way of doing this would be to keep the position in the map and use the `<<>>` value, as this will delete the value in the MPT. This only works if the value `<<>>` isn't a valid value for the store, but if it is a valid value, this in turn needs to be special treated.

apps/aecontract/src/aect_state_tree.erl
```diff
@@ -51,24 +52,74 @@ insert_contract(Contract, Tree = #contract_tree{ contracts = CtTree }) ->
...
-    Tree#contract_tree{ contracts = CtTree1 }.
+    CtTree2    = enter_store(Contract, CtTree1),
+    Tree#contract_tree{ contracts = CtTree2 }.
+
+enter_store(Contract, CtTree) ->
+    Id = aect_contracts:store_id(Contract),
+    Store = aect_contracts:state(Contract),
+    enter_store_nodes(Id, Store, CtTree).
+
+enter_store_nodes(Prefix, Store, CtTree) ->
```
gorillainduction: The same here as above. Actually, for the above, it might work since it is a fresh contract, but here it might fail.

apps/aevm/test/aevm_dummy_chain.erl
```diff
@@ -8,18 +8,22 @@
 
 -module(aevm_dummy_chain).
 
--behaviour(aec_vm_chain_api).
+%% -behaviour(aec_vm_chain_api).
```
UlfNorell: Why no more `-behaviour`?

apps/aevm/test/aevm_ethereum_test_chain.erl
```diff
@@ -8,21 +8,49 @@
 
 -module(aevm_ethereum_test_chain).
 
--behaviour(aec_vm_chain_api).
+%-behaviour(aec_vm_chain_api).
```
UlfNorell: Also ?

apps/aecontract/src/aect_state_tree.erl
gorillainduction: Exactly, but since the tree node of the contract will still point to the old store in the MPT, we need to explicitly delete these positions (or equivalently overwrite them with `<<>>`). Otherwise they will remain in the tree. The subtree to the contract is not fresh in every run.

apps/aevm/src/aevm_eeevm_state.erl
```diff
@@ -108,21 +124,24 @@ init(#{ env  := Env
...
-          , cp        => 0
-          , logs      => []
-          , memory    => #{}
-          , return_data => <<>>
-          , stack     => []
-          , storage   => Store
-          }.
+            maps:get(code, Exec, #{}),
+            maps:get(mem, Exec, #{mem_size => 0}),
+            maps:get(store, Exec, #{})).
```
UlfNorell: The store shouldn't be part of `Exec`, it's already part of the `ChainState`. Use `ChainAPI:get_store(ChainState)` here.

apps/aecontract/src/aect_dispatch.erl
```diff
@@ -43,97 +43,140 @@ encode_call_data(_, _, _, _) ->
...
-			, height     := Height
-			, trees      := Trees
-			}) ->
-
+call_AEVM_01_Sophia_01(#{ contract   := ContractPubKey
+                        , height     := Height
+                        , trees      := Trees
+			} = CallDef) ->
+    Env = set_env(ContractPubKey, Height, Trees, aec_vm_chain),
+    Store = get_store(ContractPubKey, Trees),
```
UlfNorell: This is not the place to get the store. See note in `aevm_eeevm_state`.

apps/aecontract/src/aect_dispatch.erl
```diff
@@ -43,97 +43,140 @@ encode_call_data(_, _, _, _) ->
...
+        end,
+    Store.
+
+
+call_AEVM_01_Solidity_01(#{ contract   := ContractPubKey
+                          , height     := Height
+                          , trees      := Trees
+                          } = CallDef) ->
+    Env = set_env(ContractPubKey, Height, Trees, aec_vm_chain),
+    Store = get_store(ContractPubKey, Trees),
```
UlfNorell: This is not the place to get the store. See note in `aevm_eeevm_state`.

apps/aecontract/src/aect_state_tree.erl
gorillainduction: You can compute the diff here and update the trees accordingly. But you still need to explicitly delete the removed store positions.

apps/aevm/test/aevm_dummy_chain.erl
happi: It can not find behavior if you run just one test suite

apps/aecontract/src/aect_state_tree.erl
happi: Fixed in enter_store...

apps/aecontract/src/aect_state_tree.erl
```diff
@@ -51,24 +51,85 @@ insert_contract(Contract, Tree = #contract_tree{ contracts = CtTree }) ->
...
 %% @doc Update an existing contract.
 -spec enter_contract(aect_contracts:contract(), tree()) -> tree().
 enter_contract(Contract, Tree = #contract_tree{ contracts = CtTree }) ->
     Id         = aect_contracts:id(Contract),
     Serialized = aect_contracts:serialize(Contract),
     CtTree1    = aeu_mtrees:enter(Id, Serialized, CtTree),
-    Tree#contract_tree{ contracts = CtTree1 }.
+    OldContract = get_contract(Id, Tree),
+    OldStore = aect_contracts:state(OldContract),
+    CtTree2    = enter_store(Contract, OldStore, CtTree1),
```
UlfNorell: `OldStore`? Shouldn't this be the new store? If not how is the store ever updated?

happi: Contract contains new store.
OldStore is used to see if there where keys existing in the store before that has been deleted so that they can be deleted from the tree.

UlfNorell: Nevermind..

PR #1090, opened 2018-05-07
Check that fee is more than minimum fee for all transactions

gorillainduction:
https://www.pivotaltracker.com/n/projects/2124891/stories/157308158

### DIFF
--- a/apps/aetx/src/aetx.erl
+++ b/apps/aetx/src/aetx.erl
@@ -159,8 +159,13 @@ signers(#aetx{ cb = CB, tx = Tx }) ->
 -spec check(Tx :: tx(), Trees :: aec_trees:trees(), Height :: non_neg_integer(),
             ConsensusVersion :: non_neg_integer()) ->
     {ok, NewTrees :: aec_trees:trees()} | {error, Reason :: term()}.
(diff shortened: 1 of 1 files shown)


apps/aetx/src/aetx.erl
```diff
@@ -159,8 +159,13 @@ signers(#aetx{ cb = CB, tx = Tx }) ->
 -spec check(Tx :: tx(), Trees :: aec_trees:trees(), Height :: non_neg_integer(),
             ConsensusVersion :: non_neg_integer()) ->
     {ok, NewTrees :: aec_trees:trees()} | {error, Reason :: term()}.
-check(#aetx{ cb = CB, tx = Tx }, Trees, Height, ConsensusVersion) ->
-    CB:check(Tx, aetx_transaction, Trees, Height, ConsensusVersion).
+check(#aetx{ cb = CB, tx = Tx } = O, Trees, Height, ConsensusVersion) ->
+    case is_coinbase(O) orelse (CB:fee(Tx) >= aec_governance:minimum_tx_fee()) of
```
sennui: it looks a bit like a "zero", what does the abbreviation mean, Opaque?

cc: @lucafavatella FYI, new place doing check on coinbase tx, so up for removal in scope of your tasks.

PR #1091, opened 2018-05-07
Pt 156609603 system test improvements

ThomasArts:
Some small improvements needed to execute QuickCheck tests
as well as updated sync tests to meet new release

### DIFF
--- a/system_test/aest_sync_SUITE.erl
+++ b/system_test/aest_sync_SUITE.erl
@@ -22,7 +22,6 @@
     kill_node/2,
     connect_node/3, disconnect_node/3,
     http_get/5,
-    request/4,
     wait_for_value/4,
     assert_synchronized/2
 ]).
@@ -121,6 +120,8 @@ new_node_joins_network(Cfg) ->
     NodeStartupTime = proplists:get_value(node_startup_time, Cfg),
 
     Compatible = "aeternity/epoch:local", %% Latest version it should be compatible with
+                                          %% Change if comptibility with previous version 
+                                          %% should be guaranteed
     ct:log("Testing compatiblity of epoch:local with ~p", [Compatible]),
 
     OldNode1 = #{ 
@@ -205,7 +206,7 @@ docker_keeps_data(Cfg) ->
     A = [get_block(standalone_node, H, Cfg) || H <- lists:seq(1, Length)],
 
     stop_node(standalone_node, infinity, Cfg), %% Is this triggering PT-155851463 ?
-    timer:sleep(8000),
+
     start_node(standalone_node, Cfg),
     wait_for_value({height, 0}, [standalone_node], NodeStartupTime, Cfg),
 
@@ -308,7 +309,7 @@ stop_and_continue_sync(Cfg) ->
     case Height >= Length of
          true -> {skip, already_synced_when_stopped};
          false ->
-            timer:sleep(8000),
+            timer:sleep(4 * ?MINING_TIMEOUT), %% make sure we do a bit of forking on the new chain
             start_node(node1, Cfg),
             wait_for_value({height, Length}, [node2], (Length - Height) * MiningTime, Cfg),
             {ok, 200, B2} = request(node2, 'GetBlockByHeight', #{height => Length}),
--- a/system_test/aest_sync_SUITE.erl
+++ b/system_test/aest_sync_SUITE.erl
@@ -37,34 +37,12 @@
 -define(STARTUP_TIMEOUT, 10000).
 -define(SYNC_TIMEOUT,      100).
 
--define(OLD_NODE1, #{
-    name    => old_node1,
-    peers   => [old_node2],
-    backend => aest_docker,
-    source  => {pull, "aeternity/epoch:v0.11.1"}
-}).
-
--define(OLD_NODE2, #{
-    name    => old_node2,
-    peers   => [old_node1],
-    backend => aest_docker,
-    source  => {pull, "aeternity/epoch:v0.11.1"}
-}).
-
--define(NEW_NODE1, #{
-    name    => new_node1,
-    peers   => [old_node1],
-    backend => aest_docker,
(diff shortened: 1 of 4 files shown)


system_test/aest_sync_SUITE.erl
```diff
@@ -207,6 +206,7 @@ docker_keeps_data(Cfg) ->
     A = [get_block(standalone_node, H, Cfg) || H <- lists:seq(1, Length)],
 
     stop_node(standalone_node, infinity, Cfg), %% Is this triggering PT-155851463 ?
+    timer:sleep(8000),
```
hanssv: What makes 8 seconds special? A comment perhaps?

system_test/aest_sync_SUITE.erl
```diff
@@ -144,7 +121,29 @@ new_node_joins_network(Cfg) ->
     Length = 20,
     NodeStartupTime = proplists:get_value(node_startup_time, Cfg),
 
-    setup_nodes([?OLD_NODE1, ?OLD_NODE2, ?NEW_NODE1], Cfg),
+    Compatible = "aeternity/epoch:local", %% Latest version it should be compatible with
```
eproxus: Is it intended to test `local` compatibility with `local` itself?

system_test/aest_sync_SUITE.erl
```diff
@@ -444,9 +454,14 @@ sync_speed(Cfg) ->
...
-get_block(NodeName, Height, Cfg) ->
-    Query = #{height => Height},
-    case http_get(NodeName, ext_http, [v2, 'block-by-height'], Query, Cfg) of
+get_block(NodeName, Height, _Cfg) ->
+    case request(NodeName, 'GetBlockByHeight', #{height => Height}) of
         {ok, 200, Block} -> Block;
         {ok, 404, _} -> undefined
     end.
+
+request(Node, Id, Params) ->
```
hanssv: This is a new definition of request, but where was the old one? I expected it to be removed?

system_test/aest_sync_SUITE.erl
```diff
@@ -444,9 +454,14 @@ sync_speed(Cfg) ->
...
+    aehttp_client:request(Id, Params,
```
eproxus: I'm not opposed to sharing the HTTP client from the integration tests, but I'm worried we're having yet another way to check for similar things (like height). Should we make some consolidation effort?

system_test/aest_sync_SUITE.erl
```diff
@@ -383,10 +393,10 @@ net_split_recovery(Cfg) ->
     wait_for_value({height, Length * 7}, [net1_node1, net1_node2, net2_node1, net2_node2],
                     Length * 2 * ?MINING_TIMEOUT, Cfg),
 
-    D1 = request(net1_node1, [v2, 'block-by-height'], #{height => Length * 7}, Cfg),
-    D2 = request(net1_node2, [v2, 'block-by-height'], #{height => Length * 7}, Cfg),
-    D3 = request(net2_node1, [v2, 'block-by-height'], #{height => Length * 7}, Cfg),
-    D4 = request(net2_node2, [v2, 'block-by-height'], #{height => Length * 7}, Cfg),
+    {ok, 200, D1} = request(net1_node1, 'GetBlockByHeight', #{height => Length * 7}),
```
eproxus: Shouldn't these use the `get_block` function instead?

system_test/aest_sync_SUITE.erl
hanssv: Ah, it is `imported`, that import should be removed...

ThomasArts: Thanks missed the import.

system_test/aest_sync_SUITE.erl
ThomasArts: For this version yes, but I liked Sebastiens solution to call it latest_released better. Just that I didn't get it when he called it latest.

system_test/aest_sync_SUITE.erl
ThomasArts: Yes, I consolidate towards using the swagger endpoints. This would have rescued us from continuing to use a depracated endpoint in tests.

system_test/aest_sync_SUITE.erl
ThomasArts: No. get_block is remainder of the past and I didn't see why it is used. Just didn't want to remove its uses straight away.

system_test/aest_sync_SUITE.erl
eproxus: What is `latest`? A Docker image that already exist?

system_test/aest_sync_SUITE.erl
eproxus: What do we do about it? New chore in Pivotal?

system_test/aest_sync_SUITE.erl
eproxus: What do we do about it? New chore in Pivotal?

system_test/aest_sync_SUITE.erl
lucafavatella: @eproxus From https://github.com/aeternity/epoch/blame/89bb66e68e2058466a865645e16c0ead590e431f/docs/docker.md#L13
> Docker image is automatically build and published on [DockerHub](https://hub.docker.com/r/aeternity/epoch/). All tagged source code versions have their own docker image tag as well. Latest tagged ("stable") version is tagged with `latest`.

eproxus: @lucafavatella @ThomasArts So shouldn't we just use that version here then?

lucafavatella: > @lucafavatella @ThomasArts So shouldn't we just use that version here then?

@eproxus  Assuming that master is meant to be backwards compatible - that is currently the case - yes.

PR #1092, opened 2018-05-07
Add smoke test and docker target

eproxus:
This change adds a `smoke-test` and `docker` target to the make file. The smoke test runs the `aest_sync_SUITE` and `aest_commands_SUITE` to achieve a shorter system test run for local tests. In addition, it depends on the (new) `docker` and `system-test-deps` targets to always have an up-to-date build. Because this target is intended to be run on PRs before committing, it is better to be safe and always rebuild the `local` Docker tag.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,6 +159,12 @@ eunit:
 
 all-tests: eunit test
 
+docker:
+	@docker build . -t aeternity/epoch:local
+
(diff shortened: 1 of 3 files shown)


eproxus:
@dincho This is a good candidate to run on PRs, what do you think?

PR #1093, opened 2018-05-07
Refactoring: aetx's signers/1 -> signers/2

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/157357793)
Channels' transactions don't contain the signers in the tx body so we must take those from the state trees instead. This PR addresses this by providing the aetx's callback signers/1 the trees (thus making it signers/2)

Since the signers may or may not be in the state trees at all (because of the channel might not be there) we are having a clear distinction between those two cases.

### DIFF
--- a/apps/aechannel/src/aesc_slash_tx.erl
+++ b/apps/aechannel/src/aesc_slash_tx.erl
@@ -19,12 +19,11 @@
          check/5,
          process/5,
          accounts/1,
-         signers/1,
+         signers/2,
          serialization_template/1,
          serialize/1,
          deserialize/2,
-         for_client/1,
-         is_verifiable/1
+         for_client/1
         ]).
 
 %%%===================================================================
@@ -119,13 +118,14 @@ process(#channel_slash_tx{channel_id = ChannelId,
 -spec accounts(tx()) -> list(pubkey()).
 accounts(#channel_slash_tx{payload = Payload}) ->
     case deserialize_from_binary(Payload) of
-        {ok, SignedState, _StateTx} -> aetx:signers(aetx_sign:tx(SignedState));
+        {ok, SignedState, _StateTx} -> aetx:signers(aetx_sign:tx(SignedState),
+                                                    aec_trees:new());
         {error, _Reason}            -> []
     end.
 
--spec signers(tx()) -> list(pubkey()).
-signers(#channel_slash_tx{from = FromPubKey}) ->
-    [FromPubKey].
+-spec signers(tx(), aec_trees:trees()) -> {ok, list(pubkey())}.
+signers(#channel_slash_tx{from = FromPubKey}, _) ->
+    {ok, [FromPubKey]}.
 
 -spec serialize(tx()) -> {vsn(), list()}.
 serialize(#channel_slash_tx{channel_id = ChannelId,
@@ -183,13 +183,6 @@ serialization_template(?CHANNEL_SLASH_TX_VSN) ->
     , {nonce     , int}
     ].
 
--spec is_verifiable(tx()) -> boolean().
-is_verifiable(#channel_slash_tx{channel_id = ChannelId}) ->
-    case aec_chain:get_channel(ChannelId) of
-        {ok, _Channel} -> true;
-        {error, not_found} -> false
-    end.
-
 %%%===================================================================
 %%% Internal functions
 %%%===================================================================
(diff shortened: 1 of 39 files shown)


velzevur:
EUnit failed, I am investigating it; adding the WIP label temporarily untill tests are fixed

velzevur:
Since signatures' validation is no logner performed in `aec_blocks:validate/1` - I have to add a test that `aec_trees:apply_signed_txs/4` rejects transactions with broken signatures

apps/aechannel/src/aesc_slash_tx.erl
```diff
@@ -119,13 +118,14 @@ process(#channel_slash_tx{channel_id = ChannelId,
 -spec accounts(tx()) -> list(pubkey()).
 accounts(#channel_slash_tx{payload = Payload}) ->
     case deserialize_from_binary(Payload) of
-        {ok, SignedState, _StateTx} -> aetx:signers(aetx_sign:tx(SignedState));
+        {ok, SignedState, _StateTx} -> aetx:signers(aetx_sign:tx(SignedState),
```
uwiger: But `aetx:signers/2` returns `{ok, list[pubkey()]}`.
Doesn't Dialyzer complain about this?

apps/aecore/src/aec_blocks.erl
```diff
@@ -111,7 +111,7 @@ new_with_state(LastBlock, Txs, Trees1) ->
     %% We should not have any transactions with invalid signatures for
     %% creation of block candidate, as only txs with validated signatures should land in mempool.
     %% Let's hardcode this expectation for now.
-    {Txs, _} = {aetx_sign:filter_invalid_signatures(Txs), Txs},
+    {Txs, _} = {aetx_sign:filter_invalid_signatures(Txs, Trees1), Txs},
```
gorillainduction: I see no point in keeping this call as the signatures are checked in the call to `aec_trees:apply_signed_txs/4`. The `aetx_sign:filter_invalid_signatures/1` function is then not needed anymore, which is good since it is no longer correct.

apps/aecore/test/aec_test_utils.erl
```diff
@@ -188,7 +188,9 @@ start_chain_db() ->
     ok = mnesia:start(),
     ok = aec_db:initialize_db(ram),
     Tabs = [Tab || {Tab, _} <- aec_db:tables(ram)],
-    ok = mnesia:wait_for_tables(Tabs, 5000).
+    ok = mnesia:wait_for_tables(Tabs, 5000),
```
uwiger: Given my little extension with hooks allowing e.g. `aechannels` to create extra tables (which it currently doesn't), this will not be sufficient in the future.

apps/aechannel/src/aesc_slash_tx.erl
velzevur: Great catch!

apps/aecore/test/aec_blocks_tests.erl
```diff
@@ -85,14 +85,21 @@ network_serialization_test_() ->
 
 validate_test_() ->
     {setup,
-     fun aec_test_utils:aec_keys_setup/0,
-     fun aec_test_utils:aec_keys_cleanup/1,
+     fun() ->
+             ok = meck:new(aec_chain, [passthrough]),
+             meck:expect(aec_chain, get_top_state, 0, {ok, aec_trees:new()}),
+             aec_test_utils:aec_keys_setup()
+     end,
```
gorillainduction: I don't think you need this anymore, since the signature check went away.

apps/aecore/test/aec_blocks_tests.erl
```diff
@@ -85,14 +85,21 @@ network_serialization_test_() ->
...
+             meck:unload(aec_chain),
+             ok = aec_test_utils:aec_keys_cleanup(TmpKeysDir)
+     end,
      [ {"Multiple coinbase txs in the block",
         fun validate_test_multiple_coinbase/0}
      , {"Malformed txs merkle tree hash",
         fun validate_test_malformed_txs_root_hash/0}
-     , {"Malformed tx signature",
-        fun validate_test_malformed_tx_signature/0}
+     %, {"Malformed tx signature",
```
gorillainduction: Remove the code instead of commenting.

apps/aecore/test/aec_test_utils.erl
uwiger: Since we have only one node, I imagine it would be safe to use
`Tabs = mnesia:system_info(tables) -- [schema]`

apps/aecore/test/aec_blocks_tests.erl
velzevur: Yes, I am adding a corresponding test to verify this is covered by aec_trees. This comment will remove this test as well

velzevur:
As discussed with @gorillainduction  - channel transaction signers are to be determined by the state trees. Thus validating signatures without applying transactions to the trees makes no sense (as channels can be created or closed and thus making subsequents transactions valid/invalid). This is addressed by removing `aetx_sign:filter_invalid_signatures/2`

velzevur:
All tests passed, I am restarting the `test` JOB in order to verify that the race condition is solved.

apps/aecore/test/aec_blocks_tests.erl
velzevur: Well we shall either `meck` signed coinbases in this module or use keys...

gorillainduction: Yes, but you don't need the mecking of the chain, I think. No big deal, though.

velzevur: Addressed

PR #1095, opened 2018-05-08
Pt 157308075 ensure contiguous nonces

gorillainduction:
https://www.pivotaltracker.com/n/projects/2124891/stories/157308075

### DIFF
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -333,12 +333,19 @@ check_nonce_at_hash(Tx, BlockHash) ->
             {error, illegal_nonce};
         true ->
             case aec_chain:get_account_at_hash(Pubkey, BlockHash) of
-                {value, Account} ->
-                    aetx_utils:check_nonce(Account, TxNonce);
                 {error, no_state_trees} ->
                     ok;
                 none ->
-                    ok
+                    ok;
+                {value, Account} ->
+                    case aetx_utils:check_nonce(Account, TxNonce) of
+                        ok -> ok;
+                        {error, account_nonce_too_low} ->
+                            %% This can be ok in the future
+                            ok;
+                        {error, account_nonce_too_high} = E ->
+                            E
+                    end
             end
     end.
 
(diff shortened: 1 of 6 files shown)


apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -1136,7 +1136,8 @@ unsigned_tx_positive_test(Data, Params, HTTPCallFun, NewFun, Pubkey) ->
 tx_is_mined_test(AccountPubKey, TxHash) ->
     {ok, 200, #{<<"transactions">> := Txs}} =
         get_account_transactions(AccountPubKey, tx_encoding_param(json)),
-    true = lists:any(fun(#{<<"hash">> := H}) when H =:= TxHash -> true;
+    true = lists:any(fun(#{<<"block_hash">> := BH, <<"hash">> := H})
+                           when H =:= TxHash, BH =/= <<"none">> -> true;
```
lucafavatella: If `BH` binds to `<<"block_hash">>` unclear how it may be not `=/= <<"none">>`

apps/aecore/src/aec_tx_pool.erl
```diff
@@ -333,12 +333,19 @@ check_nonce_at_hash(Tx, BlockHash) ->
             {error, illegal_nonce};
         true ->
             case aec_chain:get_account_at_hash(Pubkey, BlockHash) of
-                {value, Account} ->
-                    aetx_utils:check_nonce(Account, TxNonce);
                 {error, no_state_trees} ->
```
lucafavatella: I see two calls of check_nonce_at_hash:
* One when tx is put in mempool, via check_nonce;
* One when building block candidate.

When building block candidate, no_state_trees shall not allow ok i.e. tx in block candidate.

When putting tx in mempool, not sure what behaviour shall be.

apps/aehttp/test/aehttp_integration_SUITE.erl
gorillainduction: The signalling of the transaction being in mempool is to put `<<"none">>` there. A bit strange, but that is how the implementation is currently.

apps/aecore/src/aec_tx_pool.erl
gorillainduction: `no_state_trees` means that the nodeis not started, i.e., is in unit tests. Worst case would be that the transaction would be added while the node is started, and then we get a possibly invalid transaction in the mempool. This transaction will not be added to a block anyway, so it is fine.

gorillainduction:
Rebased and added release notes as suggested.

gorillainduction:
Restarted test on known flaky test error 'socket_closed_remotely`

PR #1096, opened 2018-05-08
Pt 157386342 storage tests

happi:


apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -231,3 +245,108 @@ state_tree(_Cfg) ->
...
+    Ct1    = ?call(get_contract, Ct1),
+    Store1 = #{ <<0>> => <<42>>
+              , <<1>> => <<0>>
+              , <<2>> => <<>> },
+    Ct2    = aect_contracts:set_state(Store1, Ct1),
+    Ct2    = ?call(enter_contract, Ct2),
+    %% Empty values are removed in state tree.
+    Ct3    = ?call(get_contract, Ct2),
+    Store2 = #{ <<0>> => <<42>>
+              , <<1>> => <<0>>},
```
gorillainduction: Are you supposed to be able to store 0 in the store? I see that it is not `<<>>` as the numerical value `0` would correspond to, so I see why the test succeeds, but it should really be an impossible value.

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -231,3 +245,108 @@ state_tree(_Cfg) ->
...
+              , <<1>> => <<0>>
+              , <<2>> => <<>> },
+    Ct2    = aect_contracts:set_state(Store1, Ct1),
+    Ct2    = ?call(enter_contract, Ct2),
+    %% Empty values are removed in state tree.
+    Ct3    = ?call(get_contract, Ct2),
+    Store2 = #{ <<0>> => <<0>>
+              , <<1>> => <<>>
+              , <<2>> => <<42>> },
+    Ct4    = aect_contracts:set_state(Store2, Ct3),
```
gorillainduction: The same as above.

apps/aecontract/test/aecontract_SUITE.erl
happi: In the PMT handling there can be the value 0, in the aevm any 0 value is striped away. Other VMs/ABIs might want to allow value 0.

PR #1100, opened 2018-05-08
Fix include path.

happi:


apps/aecontract/src/aect_call_state_tree.erl
```diff
@@ -19,7 +19,7 @@
 
 -export_type([tree/0]).
 
--include_lib("aecore/include/common.hrl").
+-include_lib("apps/aecore/include/common.hrl").
```
lucafavatella: What was the symptom? Maybe this include can be completely deleted?

happi: Robert was building "locally" in the dir, not with rebar, then the include path can not be found. I think there are many places where apps are missing as a prefix. It only affects building without rebar I think.

PR #1103, opened 2018-05-08
Sync improvements

hanssv:
The PR fixes several issues in one go, mainly
- https://www.pivotaltracker.com/story/show/157215225
- https://www.pivotaltracker.com/story/show/157218434
- and some further improvements w.r.t. enoise

### DIFF
--- a/apps/aecore/include/aec_peer_messages.hrl
+++ b/apps/aecore/include/aec_peer_messages.hrl
@@ -13,3 +13,19 @@
 -define(MSG_MEMPOOL, 14).
 -define(MSG_GET_HEADER_BY_HEIGHT, 15).
 
+-define(RESPONSE_VSN, 1).
+-define(PING_VSN, 1).
+-define(GET_MEMPOOL_VSN, 1).
+-define(MEMPOOL_VSN, 1).
+-define(GET_HEADER_BY_HASH_VSN, 1).
+-define(GET_HEADER_BY_HEIGHT_VSN, 2).
+-define(HEADER_VSN, 1).
+-define(GET_N_SUCCESSORS_VSN, 2).
+-define(HEADER_HASHES_VSN, 1).
+-define(GET_BLOCK_VSN, 1).
+-define(BLOCK_VSN, 1).
+-define(TX_VSN, 1).
+-define(PEER_VSN, 1).
+
+-define(VSN_1, 1).
+
--- a/apps/aecore/src/aec_peer_connection.erl
+++ b/apps/aecore/src/aec_peer_connection.erl
@@ -42,6 +42,7 @@
 -define(DEFAULT_CONNECT_TIMEOUT, 1000).
 -define(DEFAULT_FIRST_PING_TIMEOUT, 30000).
 -define(DEFAULT_NOISE_HS_TIMEOUT, 5000).
+-define(REQUEST_TIMEOUT, 5000).
 
 %% gen_server callbacks
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
@@ -95,7 +96,7 @@ stop(PeerCon) ->
 
 
 call(PeerCon, Call) when is_pid(PeerCon) ->
-    try gen_server:call(PeerCon, Call)
+    try gen_server:call(PeerCon, Call, ?REQUEST_TIMEOUT + 2000)
     catch exit:{noproc, _} -> {error, no_connection}
     end;
 call(PeerId, Call) when is_binary(PeerId) ->
@@ -166,8 +167,8 @@ init([Opts]) ->
     Version = <<?P2P_PROTOCOL_VSN:64>>,
     Genesis = aec_chain:genesis_hash(),
     Opts1 = Opts#{ role => initiator
-                   , version => Version
-                   , genesis => Genesis},
+                 , version => Version
+                 , genesis => Genesis},
     {ok, Opts1, 0}.
 
 handle_call(Request, From, State) ->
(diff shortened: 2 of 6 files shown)


apps/aecore/src/aec_peer_connection.erl
```diff
@@ -143,7 +143,7 @@ accept_init(Ref, TcpSock, ranch_tcp, Opts) ->
                 , {timeout, HSTimeout} ],
     %% Keep the socket passive until here to avoid premature message
     %% receiving...
-    inet:setopts(TcpSock, [{active, true}]),
+    inet:setopts(TcpSock, [{active, true}, {send_timeout, 1000}, {send_timeout_close, true}]),
```
lucafavatella: I see [send_timeout is related to send being completely accepted by underlying TCP stack and that it 
default timeout is infinity](https://github.com/erlang/otp/blob/OTP-20.2.3/lib/kernel/doc/src/inet.xml#L1057-L1063). I am concerned of 1s (this is ms, right?) hardcoded, that may be tight in real world scenarios. Would it make sense to make this [configurable](https://github.com/aeternity/epoch/blob/86772227ad547e05d0df21337b0f171e16ed258b/apps/aeutils/test/data/epoch_full.yaml#L29-L34)?

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -226,7 +226,7 @@ handle_info({connected, Pid, {ok, TcpSock}}, S0 = #{ status := {connecting, Pid}
...
```
lucafavatella: Same as above.

rebar.config
```diff
@@ -27,7 +27,7 @@
         {parse_trans, "3.2.0"},
         {mnesia_rocksdb, ".*", {git, "https://github.com/aeternity/mnesia_rocksdb.git", {ref,"618870f"}}},
         {enacl, ".*", {git,"https://github.com/aeternity/enacl.git", {ref,"2f50ba6"}}},
-        {enoise, ".*", {git, "https://github.com/aeternity/enoise.git", {ref, "e3baa21"}}},
+        {enoise, ".*", {git, "https://github.com/aeternity/enoise.git", {ref, "7347906"}}},
```
lucafavatella: Wrt https://github.com/aeternity/enoise/compare/e3baa21...7347906#diff-93af256510f79796918e0ff06b07fbb2R109
I see it relies on being linked but when controlling process changes who does the new linking? (I did not arrive at checking yet.)

apps/aecore/src/aec_peer_connection.erl
hanssv: I think 1s is a rather sensible default (after all we are talking about having micro-blocks in Bitcoin-NG with intervals in a handful(??) of seconds). Not sure it makes (much) sense to have it configurable it should at least always be significantly shorter than the default gen_server timeout or else there is a race!?

rebar.config
hanssv: This is the safety mechanism monitoring from the inside. I had envisioned the linking to be handled by the controlling process (automatically on accept/connect and later by hand if giving away the control) - but of course you could think of it being done automatically?

apps/aecore/include/aec_peer_messages.hrl
```diff
@@ -13,3 +13,19 @@
 -define(MSG_MEMPOOL, 14).
 -define(MSG_GET_HEADER_BY_HEIGHT, 15).
 
+-define(RESPONSE_VSN, 1).
+-define(PING_VSN, 1).
+-define(GET_MEMPOOL_VSN, 1).
+-define(MEMPOOL_VSN, 1).
+-define(GET_HEADER_BY_HASH_VSN, 1).
+-define(GET_HEADER_BY_HEIGHT_VSN, 2).
```
lucafavatella: I see this bumped from 1. Ok.

apps/aecore/include/aec_peer_messages.hrl
```diff
@@ -13,3 +13,19 @@
...
+-define(HEADER_VSN, 1).
+-define(GET_N_SUCCESSORS_VSN, 2).
```
lucafavatella: I see this bumped from 1. Ok.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -613,29 +613,52 @@ get_header(Fun, Arg) ->
...
-                         {error, block_not_found}
-                 end;
+            #{ hash := Hash, n := N } when Vsn == ?VSN_1 ->
+                do_get_n_successors(Hash, N);
+            #{ from_hash := FromHash, target_hash := TargetHash, n := N }
+                when Vsn == ?GET_N_SUCCESSORS_VSN ->
+                Res = do_get_n_successors(FromHash, N),
+                case aec_chain:hash_is_in_main_chain(TargetHash) of
+                    true  -> Res;
+                    false -> {error, not_on_chain}
```
lucafavatella: I see this as new error code if request is new version. Ok.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -613,29 +613,52 @@ get_header(Fun, Arg) ->
...
+    S;
+handle_get_header_by_height(S, ?GET_HEADER_BY_HEIGHT_VSN,
+                            #{ height := H, top_hash := TopHash}) ->
+    case {aec_chain:get_header_by_height(H),
+          aec_chain:hash_is_in_main_chain(TopHash)} of
+        {{ok, Header}, true} ->
+            SerHeader = aec_headers:serialize_to_binary(Header),
+            send_response(S, header, {ok, #{ hdr => SerHeader }});
+        {_, false} ->
+            send_response(S, header, {error, not_on_chain});
```
lucafavatella: I see this new error if request new version. Ok.

apps/aecore/src/aec_peer_connection.erl
hanssv: Not really, this is the error you get if the peer has switched fork.

lucafavatella:
> It is unclear to me how the new node determines whether to send to an old node new or old versions of the requests.

I see that only aec_peer_messages:serialize/2 is used hence always sending out new request.

hanssv:
Yes, there is no attempt to be fully backwards compatible... A new node cannot sync with an old node, but a new node can serve responses to an old node.

PR #1104, opened 2018-05-08
Check keys length before using enacl

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/157389930)

Guard the `enacl` nif from keys with unexpected sizes.
Added some tests.

### DIFF
--- a/apps/aetx/src/aetx_sign.erl
+++ b/apps/aetx/src/aetx_sign.erl
@@ -58,9 +58,9 @@ sign(Tx, PrivKey) when is_binary(PrivKey) ->
   sign(Tx, [PrivKey]);
 sign(Tx, PrivKeys) when is_list(PrivKeys) ->
     Bin = aetx:serialize_to_binary(Tx),
-    case lists:any(fun(PrivKey) -> not (?VALID_PRIVK(PrivKey)) end, PrivKeys) of
-        true -> erlang:error(invalid_priv_key);
-        false -> pass
+    case lists:filter(fun(PrivKey) -> not (?VALID_PRIVK(PrivKey)) end, PrivKeys) of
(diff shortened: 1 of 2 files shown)


apps/aetx/src/aetx_sign.erl
```diff
@@ -56,7 +59,8 @@ sign(Tx, PrivKey) when is_binary(PrivKey) ->
   sign(Tx, [PrivKey]);
 sign(Tx, PrivKeys) when is_list(PrivKeys) ->
     Bin = aetx:serialize_to_binary(Tx),
-    Signatures = [ enacl:sign_detached(Bin, PrivKey) || PrivKey <- PrivKeys ],
+    Signatures = [ enacl:sign_detached(Bin, PrivKey)
+                   || PrivKey <- PrivKeys, ?VALID_PRIVK(PrivKey)],
```
lucafavatella: This is silently filtering. May I suggest an explicit lists:filter if this is intention?

hanssv: I doubt you want to filter here, rather fail in a controlled way?

velzevur:
The main point was not crashing when verifying incoming keys (because it could be an attack vector)

What relates to a node signing with a broken key - this probably is a symptom for a bug so modified the silent producing of broken _signed_ transactions into a loud bang.

lucafavatella:
I see comment clarified. Removing WIP label.

velzevur:
Addressed comments:

* coinbases transactions in tests are spend transactions so one less dependency in removing the coinbase

* error emitted by `aetx_sign:sign2` does contain the invalid key(s)

* `aetx_sign:verify_one_pubkey/4` head that was processing 2 errors, now is split into two distinct heads for better visibility

PR #1105, opened 2018-05-08
Address rebase issues

velzevur:
This PR addresses 2 issues exposed by recent tx pool changes:
1) When channel is closed (`aesc_close_mutual_tx`) the channel is removed from state trees and the `origin` is unobtainable; this causes problems with nonce validation _after_ the transaction had been included in a block (and the channel had been closed).

2) There is a subtle difference between `aec_account:earn/3` and `aec_account:spend/4` is that the latter bumpes the account's nonce. All transaction types use the `spend/4` at least for paying the miner the fee. Only transacton that has no explicit payer of the fee is `aec_close_solo_tx` (the fee is split between parties) so no spend is done there, hence no nonce bump.


velzevur:
Addressed indentation comment

velzevur:
`docker_image_build` failed with
```
Step 9/18 : RUN apt-get -qq update && apt-get -qq -y install openssl curl     && ldconfig     && rm -rf /var/lib/apt/lists/*
 ---> Running in 3af34ba515df
Too long with no output (exceeded 10m0s)
```
Since it looks like a network issue not related with changes in the PR, I am restarting this JOB

PR #1107, opened 2018-05-09
Pt 157420596 spend from sophia contracts

happi:


apps/aevm/src/aevm_eeevm_state.erl
```diff
@@ -182,18 +183,33 @@ init_vm(State, Code, Mem, Store) ->
...
-            {ok, Return, GasSpent, set_chain_state(ChainState1, State)};
-        {error, Err} -> {error, Err}
-    catch K:Err ->
-        lager:error("~w:call_contract(~w, ~w, ~w, ~w, ~w, _) crashed with ~w:~w",
-                    [TargetKey, CallGas, Value, Data, CallStack, K, Err]),
-        {error, Err}
+    case vm_version(State) of
+        ?AEVM_01_Sophia_01 when Target == 0 ->  %% Primitive call
+            case Data of    %% TODO: use aeso_data:from_binary() (but it doesn't take a base address at the moment)
+                <<64:256, ?PRIM_CALL_SPEND:256, Recipient:256>> ->
```
happi: What is this magic 64?

UlfNorell: This is the pointer to the pair `(?PRIM_CALL_SPEND, Recipient)`.

UlfNorell: As the comment says, it would be nicer to use `aeso_data:from_binary()`, but it doesn't yet know about base addresses (32 for calldata).

PR #1109, opened 2018-05-09
Improvements to building Docker container and smoke tests

eproxus:
Also ignores crash dumps in the root directory.

If you can come up with other ideas of things that can be ignored safely, I can amend the PR.


eproxus:
@lucafavatella I saw it in the `.gitignore` file, but since I didn't have it I didn't want to add it

Makefile
```diff
@@ -162,7 +162,13 @@ all-tests: eunit test
 docker:
 	@docker build . -t aeternity/epoch:local
 
-smoke-test: system-test-deps docker
+docker-clean:
+	@docker image prune -f
+	@docker container prune -f
```
lucafavatella: This removes stopped containers.

Makefile
```diff
@@ -162,7 +162,13 @@ all-tests: eunit test
 docker:
 	@docker build . -t aeternity/epoch:local
 
-smoke-test: system-test-deps docker
+docker-clean:
+	@docker image prune -f
```
lucafavatella: This removes unused images.

sylane: You may want to do the following to cleanup any system test containers and network:
```
docker rm -f $(docker ps -a | grep 'aeternity/epoch' | cut -d ' ' -f 1)
docker network prune -f
```

Makefile
eproxus: This might be too much? The target has to be run manually though...

dincho:
> Some kind of filter of what images and containers being deleted would be very helpful.

👍

eproxus:
@dincho I'll update with filters.

@lucafavatella Why would it differ if the order was different?

lucafavatella:
> @lucafavatella Why would it differ if the order was different?

@eproxus Because I expect that a stopped container still makes its image count as used.

eproxus:
@dincho @lucafavatella New strategy: I'm adding a tag to all containers and images created and only cleaning those

PR #1111, opened 2018-05-09
Pt 157391502 reliable smoke tests

ThomasArts:
Smoke test failed before, now passes more reliably. sometimes a test is skipped because precondition for good testing is not fulfilled within time frame we have.


eproxus:
@ThomasArts Was it intentional to remove the hard fork suite and all related code?

PR #1115, opened 2018-05-10
Redirect sync log to there own log file

sylane:
* Do we want to redirect ONLY the debug logging to the sync log file?
* Do we want to include the log from `aec_peers` and the rest of the log from `aec_peer_connections`?


lucafavatella:
I am taking the liberty to label as WIP for the moment just for reminding us not to be merged yet.

config/sys.config
```diff
@@ -87,6 +87,14 @@
                   {formatter_config, [time, " ", message, "\n"]}
                ]}
              ]}
+           ]},
+           {epoch_sync_lager_event, [
+             {handlers, [
+               {lager_file_backend, [
+                  {file, "log/epoch_sync.log"}, {level, debug},
```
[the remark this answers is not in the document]
dincho: yes, but I guess we can update datadog configuration after this is merged too

config/sys.config
```diff
@@ -87,6 +87,14 @@
                   {formatter_config, [time, " ", message, "\n"]}
                ]}
              ]}
+           ]},
+           {epoch_sync_lager_event, [
```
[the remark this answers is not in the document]
sylane: They pass, I added the config in devs sys files anyway.

lucafavatella:
Normal review&merge rules apply. Removing WIP label.

sylane:
@lucafavatella: I missed the changes in `aecore_app`, I'll add it; I tried adding`epoch_sync.log` in `aecore_suitre_utils` module but some tests do not seem to sync and that makes the tests fail; I'll add the mining log to system tests check; I'll update `ISSUE_TEMPLATE.md`.

config/sys.config
dincho: Is this rules reroute only debug logs or all logs is lower than debug level?
I wonder if this changes affect the `lager_console_backend` handler above and docker logs?

sylane: As I understand it, all the log done using `epoch_sync:XXX` functions end up in `epoch.sync.log`, this configuration is to filter them.

dincho: Does not sound good, please don't merge until I test it locally

sylane:
@dincho How is this degrading docker experience ?

dincho:
@sylane [`docker logs`](https://docs.docker.com/engine/reference/commandline/logs/) and `docker run -it ...` will stop producing meaningful output. 

It's not just docker however, IMO each unix process running in "foreground" should stream it's output to the stderr/stdout. [More about the reasoning](https://jdebp.eu/FGA/unix-daemon-design-mistakes-to-avoid.html#DoNotUseSyslog).

What if you add `lager_console_backend` as extra backend like in the [default handlers](https://github.com/aeternity/epoch/pull/1115/files#diff-a20e447d0b40a861040c0b922625585cR56) ?

sylane:
@dincho  I could add `{lager_console_backend, [{level, info}]}` like in the default handler, but that would not change much given most of the log generated are `debug` lines. @uwiger, you requested the feature, what do you think about @dincho concerns?

sylane:
@dincho  I added console backend for info log lines.

config/sys.config
```diff
@@ -90,7 +90,8 @@
            ]},
            {epoch_sync_lager_event, [
              {handlers, [
-               {lager_file_backend, [
+                {lager_console_backend, [{level, info}]},
```
lucafavatella: Also in epoch_mining_lager_event ?

sylane:
@dincho I though one of your concern was related to using docker/compose and these are used for that.

config/sys.config
sylane: I don't know, we could; this is not related to the changes in that PR though.

sylane:
@dincho, do you have more issues with this PR ?

dincho:
> @dincho I though one of your concern was related to using docker/compose and these are used for that.

Can you point me where those `dev{1..3}/sys.config` is used in docker compose? AFAIK it's not, and it's used for development and testing.

dincho:
@sylane Still the PR is good to go from my perspective and I already approved it, just with additional (optional) comment.

PR #1118, opened 2018-05-10
PT-157121522 Oracle response TTL bug

hanssv:
Fixes https://www.pivotaltracker.com/story/show/157121522

### DIFF
--- a/apps/aeoracle/test/aeo_test_utils.erl
+++ b/apps/aeoracle/test/aeo_test_utils.erl
@@ -21,6 +21,7 @@
         , setup_new_account/1
         , setup_new_account/3
         , trees/1
+        , ttl_defaults/0
         ]).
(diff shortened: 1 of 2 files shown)


apps/aeoracle/src/aeo_query.erl
```diff
@@ -8,7 +8,8 @@
 -module(aeo_query).
 
 %% API
--export([ deserialize/1
+-export([ add_response/3
```
lucafavatella: set_response/2 is now_unused and it shall be deleted? (alongside tests for this test).

lucafavatella:
I am preparing PR #1124 addressing my two comments .

lucafavatella:
Tests were fixed in distinct commit hence I squashed before merging.

PR #1124, opened 2018-05-11
Pt 157121522 oracle response ttl bug - Comments

lucafavatella:


lucafavatella:
Rebased on master - now that #1118 merged.

apps/aeoracle/src/aeo_query.erl
```diff
@@ -221,10 +220,6 @@ set_oracle_address(X, I) ->
 set_query(X, I) ->
     I#query{query = assert_field(query, X)}.
 
--spec set_response(oracle_response(), query()) -> query().
```
hanssv: it is a getter/setter API seems strange to arbitrarily remove one function?

hanssv: (and no, not working I promise :-) )

PR #1126, opened 2018-05-14
Add support for channels to docker compose testnet

mpowaga:


dincho:
@mpowaga Could you please add a line in the [docs](https://github.com/aeternity/epoch/blob/master/docs/docker.md#localnet) mentioning this new endpoint. 

I'll be also grateful if you mention the `websocket` endpoint in the docs too.

dincho:
Thanks

docker/nginx-default.conf
```diff
@@ -39,14 +39,14 @@ server {
...
-        proxy_connect_timeout 15s;
-        proxy_send_timeout 2h;
-        proxy_read_timeout 2h;
-        proxy_pass_request_headers on;
+    }
+
+    location /channel {
+        include cors.conf;
+        include ws.conf;
+        proxy_pass http://node1:3014;
```
lucafavatella: node1?

docker/nginx-default.conf
```diff
@@ -65,13 +65,13 @@ server {
...
-        proxy_send_timeout 2h;
-        proxy_read_timeout 2h;
-        proxy_pass_request_headers on;
     }
-}
+
+    location /channel {
+        include cors.conf;
+        include ws.conf;
+        proxy_pass http://node1:3014;
```
lucafavatella: node1?

mpowaga:
Good catch @lucafavatella! Fixed it.

PR #1127, opened 2018-05-14
resolve name at current height

thepiwo:
names will be resolved for the following transactions:
 - spend recipient (account_pubkey)
 - name transfer recipient (account_pubkey)
 - oracle query oracle address (oracle_pubkey)

Faulty resolvement while processing post requests was removed. Names will be persisted in the transaction and resolvement and affecting state happens during processing on state while mining/verifying.

Get name requests include the owner.

Faulty is_name function from aens is removed, as it is not used any more and resolvement uses full name verification to ensure a given binary is a name.

Json encoded responses now include the clear text name instead of invalid encoding of the name.


lucafavatella:
@thepiwo I understand this is meant to deliver https://www.pivotaltracker.com/story/show/156600645 ?

thepiwo:
@lucafavatella yes exactly, but including the resolvemements implementation as described in the PR description

thepiwo:
@lucafavatella I will

apps/aeoracle/src/aeo_extend_tx.erl
```diff
@@ -98,10 +102,13 @@ signers(#oracle_extend_tx{oracle = OraclePK}, _) ->
...
 
 -spec process(tx(), aetx:tx_context(), aec_trees:trees(), height(), non_neg_integer()) -> {ok, aec_trees:trees()}.
-process(#oracle_extend_tx{oracle = OraclePK, nonce = Nonce, fee = Fee, ttl = TTL},
+process(#oracle_extend_tx{oracle = OraclePKOrName, nonce = Nonce, fee = Fee, ttl = TTL},
         _Context, Trees0, Height, _ConsensusVersion) ->
     AccountsTree0 = aec_trees:accounts(Trees0),
     OraclesTree0  = aec_trees:oracles(Trees0),
+    NamesTree0 = aec_trees:ns(Trees0),
+
+    {ok, OraclePK} = aens:resolve_decoded(oracle_pubkey, OraclePKOrName, NamesTree0),
```
lucafavatella: This affects the originator of the tx, that is responsible for signing and whose nonce is affected. This seems too involved.

* As both signature and nonce are closely coupled with tx originator, why would originator use a name? I may be missing the use case.
* If mempool checked tx before accepting it in pool (assuming it does - did not check), how could it reliably check that signature is from correct tx originator?
* If attempting some sort of ordering of tx for the originatorpubkey-nonce contraints, this makes it impossible to reason on this contraints distinctly from creation of block candidate. (That could be ok - not sure.) cc @gorillainduction

thepiwo: I agree @lucafavatella the case of name resolvement for extend tx can be problematic and is not beneficial to user experience, I propose to remove name resolvement here.

config/swagger.yaml
```diff
@@ -2245,7 +2247,7 @@ definitions:
     type: object
     properties:
       oracle_pubkey:
-        type: string
+        $ref: '#/definitions/EncodedHashOrName'
```
lucafavatella: I see [serialization documented as binary without size](https://github.com/aeternity/protocol/blob/0918c2a3262bb6e39e9eb589da380b248d006f3a/serializations.md#oracle-query-transaction). So no need to change that. Ok.

config/swagger.yaml
```diff
@@ -2245,7 +2247,7 @@ definitions:
     type: object
     properties:
       oracle_pubkey:
```
lucafavatella: This property name may be misleading (pubkey while it can be name)?

config/swagger.yaml
lucafavatella: The [oracle consensus doc describes this field as "The oracle (address)"](https://github.com/aeternity/protocol/blob/0918c2a3262bb6e39e9eb589da380b248d006f3a/oracles/oracle_transactions.md#oracle-query-transaction): that seems clashing with name?

config/swagger.yaml
```diff
@@ -2399,7 +2401,7 @@ definitions:
       name_hash:
         type: string
       recipient_pubkey:
```
lucafavatella: This property name may be misleading (pubkey while it can be name)?

config/swagger.yaml
```diff
@@ -2399,7 +2401,7 @@ definitions:
...
-        type: string
+        $ref: '#/definitions/EncodedHashOrName'
```
lucafavatella: [Serialization doc mentions this as binary without size](https://github.com/aeternity/protocol/blob/0918c2a3262bb6e39e9eb589da380b248d006f3a/serializations.md#name-service-transfer-transaction) though the [AENS spec details size 32](https://github.com/aeternity/protocol/blob/0918c2a3262bb6e39e9eb589da380b248d006f3a/AENS.md#transfer). How does that reconcile with fitting a name?

config/swagger.yaml
thepiwo: it is allowed to be an encoded pubkey or a name string

config/swagger.yaml
thepiwo: should clarify on what the contents is allowed to be

thepiwo: so, should we adjust spec, this is basically the feature this PR implements

config/swagger.yaml
thepiwo: so, should we adjust spec, this is basically the feature this PR implements

config/swagger.yaml
thepiwo: this is what this property is

lucafavatella:
This needs rebase on recent master for re-reviewing conformance with https://github.com/aeternity/protocol/pull/91

thepiwo:
@lucafavatella I'll close this for now, as we will have to add a type byte to the binary representation as discussed with @gorillainduction during the dev meetup, afterwards I will rebase, adjust to namehash and reopen this PR

PR #1128, opened 2018-05-14
Add key and micro blocks initial validation

zp-sd:
* To be fixed: tests.


apps/aecore/src/aec_validation.erl
```diff
@@ -0,0 +1,27 @@
...
+
+%%%===================================================================
+%%% API
+%%%===================================================================
+
+%% TODO Create envelope that tells which kind of block/header is being processed.
+%% Then Move specific validation to aec_key_block.erl and aec_micro_block.erl,
+%% depending on block/header kind.
+%% This will be changed in/after jur0's changes.
+validate(Block, LeaderKey) ->
```
lucafavatella: module:function name seems too generic.

apps/aecore/src/aec_blocks.erl
```diff
@@ -292,32 +293,43 @@ hash_internal_representation(B = #block{}) ->
     aec_headers:hash_header(to_header(B)).
 
 
-%% TODO: implement validation of microblocks
--spec validate(block(), binary()) -> ok | {error, term()}.
-validate(Block, _LeaderKey) ->
+validate_key_block(Block) ->
+    case aec_headers:validate_key_block_header(to_header(Block)) of
+        ok ->
+            Validators = [fun validate_no_txs/1,
```
lucafavatella: I would expect the txs absence to be checked during serialization, by not having txs in the key block structure?

zp-sd: Yes, this should go away after @jur0 's refactoring?

jur0: Correct, txs won't be there anymore in key blocks.

zp-sd: @lucafavatella @jur0 Since this PR should go in before we have big @jur0 's refactoring I wanted this check to be here so that we don't forget. Objections?

apps/aecore/src/aec_blocks.erl
```diff
@@ -292,32 +293,43 @@ hash_internal_representation(B = #block{}) ->
...
+    end.
+
+validate_no_txs(#block{txs = Txs}) ->
+    case Txs =:= [] of
+        true  -> ok;
+        false -> {error, txs_in_key_block}
+    end.
+
+validate_empty_txs_hash(#block{txs_hash = TxsHash}) ->
+    case TxsHash =:= <<0:?TXS_HASH_BYTES/unit:8>> of
```
gorillainduction: A would prefer having a callback in aec_txs_trees to check if this is the empty hash.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -247,70 +257,94 @@ internal_insert(Node, Block) ->
...
 
-assert_calculated_target(Node, PrevNode, Delta, Height) when Delta >= Height ->
-    %% We only need to verify that the target is equal to its predecessor.
-    case {node_difficulty(Node), node_difficulty(PrevNode)} of
+assert_difficulty_equal_to_prev(Node, PrevNode) ->
+    PrevKeyNode = case is_key_block(PrevNode) of
+                      true  -> PrevNode;
+                      false -> db_find_node(node_key_hash(PrevNode))
+                  end,
+    case {node_difficulty(Node), node_difficulty(PrevKeyNode)} of
```
gorillainduction: We should probably compare targets instead of difficulties to avoid comparing floats for equality.

apps/aecore/src/aec_blocks.erl
lucafavatella: > Objections?

No.

apps/aecore/src/aec_chain_state.erl
zp-sd: Good point.

apps/aecore/src/aec_blocks.erl
gorillainduction: Nevermind, this will go away later.

PR #1129, opened 2018-05-14
Lock down last stable ethereum test suite for aevm tests

gorillainduction:
https://www.pivotaltracker.com/n/projects/2124891/stories/157550448

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,7 @@ PYTHON_TESTS = $(PYTHON_DIR)/tests
 PIP = $(PYTHON_BIN)/pip
 
 export AEVM_EXTERNAL_TEST_DIR=aevm_external
+export AEVM_EXTERNAL_TEST_VERSION=348b0633f4a6ee3c100368bf0f4fca71394b4d01
 
 HTTP_APP = apps/aehttp
 SWTEMP := $(shell mktemp -d 2>/dev/null || mktemp -d -t 'mytmpdir')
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -193,6 +194,11 @@ aevm-test: aevm-test-deps
 	@./rebar3 eunit --application=aevm
 
 aevm-test-deps: $(AEVM_EXTERNAL_TEST_DIR)/ethereum_tests
+aevm-test-deps:
+	$(eval VER=$(shell git -C "$(AEVM_EXTERNAL_TEST_DIR)/ethereum_tests" log -1 --pretty=format:"%H"))
+	@if [ "$(VER)" != "$(AEVM_EXTERNAL_TEST_VERSION)" ]; then \
```
dincho: I'm not sure why we'd need this explicit check, I guess git does that internal and much faster 🤔

gorillainduction: You are right. I just wanted to be a bit more silent if everything was ok.

Makefile
```diff
@@ -193,6 +194,11 @@ aevm-test: aevm-test-deps
 	@./rebar3 eunit --application=aevm
 
 aevm-test-deps: $(AEVM_EXTERNAL_TEST_DIR)/ethereum_tests
+aevm-test-deps:
+	$(eval VER=$(shell git -C "$(AEVM_EXTERNAL_TEST_DIR)/ethereum_tests" log -1 --pretty=format:"%H"))
```
lucafavatella: `git rev-parse HEAD`?

gorillainduction: As this is a stopgap, I'll not update. But indeed. That might have been more clear.

PR #1130, opened 2018-05-14
Add aevm tests to PR and master workflows

dincho:
Instead of running nightly AEVM tests and send fail notification, I think it's better to have this tests run on PR and master branches already.

https://www.pivotaltracker.com/story/show/156877392

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -476,6 +476,17 @@ workflows:
             tags:
               only: /^v.*$/
 
+      - aevm_tests:
+          requires:
+            - build
+          filters:
+            branches:
+              ignore:
+                - env/dev1
+                - env/dev2
+            tags:
+              only: /^v.*$/
+
       - uat_tests:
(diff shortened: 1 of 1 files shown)


dincho:
> Addition of fail_notification step in aevm_tests job?

Why? They fail the build.

dincho:
> Addition of fail_notification step in aevm_tests job?
> Why? They fail the build.

Ah, because of master. Makes sense, added in 029fdf50

lucafavatella:
From [build job](https://circleci.com/gh/aeternity/epoch/12249?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link):
```
===> Compiling erlexec
===> Compiling /home/builder/epoch/_build/default/lib/erlexec/c_src/*.cpp
===> g++: fatal error: cannot specify -o with -c, -S or -E with multiple files
compilation terminated.
```

PR #1132, opened 2018-05-15
Pt 157349203 only signed tx hashes

gorillainduction:
https://www.pivotaltracker.com/n/projects/2124891/stories/157349203

### DIFF
--- a/apps/aecore/src/aec_db.erl
+++ b/apps/aecore/src/aec_db.erl
@@ -44,8 +44,8 @@
         , add_tx/1
         , add_tx_hash_to_mempool/1
         , is_in_tx_pool/1
-        , find_transaction_in_main_chain_or_mempool/1 %% NOTE: Legacy tx hash
         , find_tx_location/1
+        , find_tx_with_location/1
         , remove_tx_from_mempool/1
         , remove_tx_location/1
         ]).
@@ -85,7 +85,6 @@
 
 %% indexing callbacks
 -export([ ix_acct2tx/3
-        , ix_tx2stx/3
         ]).
 
 -include("common.hrl").
@@ -129,7 +128,7 @@ tables(Mode) ->
    , ?TAB(aec_channel_state)
    , ?TAB(aec_name_service_cache)
    , ?TAB(aec_name_service_state)
-   , ?TAB(aec_signed_tx, [{index, [{acct2tx}, {tx2stx}]}])
+   , ?TAB(aec_signed_tx, [{index, [{acct2tx}]}])
    , ?TAB(aec_tx_location)
    , ?TAB(aec_tx_pool)
     ].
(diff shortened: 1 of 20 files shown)


config/swagger.yaml
```diff
@@ -3156,9 +3156,6 @@ definitions:
       tx:
         description: Unsigned transaction object
         $ref: '#/definitions/EncodedHash'
-      tx_hash:
```
lucafavatella: Could you please re-scan the whole swagger.yaml for `tx_hash`? It looks like it is in there. E.g. 2 lines below it is listed as required field.

lucafavatella: Following removal of this item, shall (encoded) hash of (posted) signed tx be returned by `PostTx` request? (So that user can easily call `GetTx`.)

apps/aecore/src/aec_db.erl
```diff
@@ -429,6 +428,24 @@ find_tx_location(STxHash) ->
            [#aec_tx_location{value = BlockHash}] -> BlockHash
        end).
 
+-spec find_tx_with_location(binary()) ->
+                                   'none'
+                                 | {'mempool', aetx_sign:signed_tx()}
+                                 | {binary(), aetx_sign:signed_tx()}.
+find_tx_with_location(STxHash) ->
```
lucafavatella: (Mostly out of scope for this PR so optional for this PR.) Is the [order of the signers documented](https://github.com/aeternity/protocol/blob/414051131a29b808bd829525f773e8fc3ae75d2d/serializations.md#signed-transaction) anywhere to be deterministic?

config/swagger.yaml
gorillainduction: I will fix

gorillainduction: I was considering this, but I think it makes sense for the user of the API to construct it themselves. It is easy, and the hash should probably not be trusted from the node anyway.

apps/aecore/src/aec_db.erl
gorillainduction: The order is defined in the code, but it is probably not documented.

apps/aehttp/priv/swagger.json
```diff
@@ -4093,15 +4093,11 @@
     },
     "UnsignedTx" : {
       "type" : "object",
-      "required" : [ "tx", "tx_hash" ],
```
lucafavatella: Spurious `, "tx_hash" : null` in `example`?

gorillainduction: Fixed

gorillainduction:
Rebased and squashed

gorillainduction:
@lucafavatella I have checked the relevant API documentation in Protocol, and concluded that apart from the API change that is captured by the generated swagger docs (removal of tx_hash in unsigned transactions), the changes are conforming to the old documentation. The old hashes (of unsigned transactions) have simply been replaced by the hashes of the signed transactions.

apps/aecore/src/aec_db.erl
```diff
@@ -129,7 +128,7 @@ tables(Mode) ->
    , ?TAB(aec_channel_state)
    , ?TAB(aec_name_service_cache)
    , ?TAB(aec_name_service_state)
-   , ?TAB(aec_signed_tx, [{index, [{acct2tx}, {tx2stx}]}])
+   , ?TAB(aec_signed_tx, [{index, [{acct2tx}]}])
```
lucafavatella: Shall DB folder of environments (deploy.yml) be bumped? Mind conflicts.

gorillainduction:
I will add the signed transaction hash in the response from the endpoint where the signed transaction is submitted. Then you can wait for that transaction hash. It is still a good idea to verify the hash by computing it in the SDK as well and compare it, but since it is a small effort to return it, I can add it.

apps/aecore/src/aec_db.erl
lucafavatella: Opened https://github.com/aeternity/protocol/pull/83 with separate Pivotal ticket.

PR #1133, opened 2018-05-15
quick fix for missing check

sennui:


apps/aecore/src/aec_chain_state.erl
```diff
@@ -260,15 +260,15 @@ internal_insert(Node, Block) ->
 
 %% NG-INFO: micro blocks inherit the height from the last key block
 assert_previous_height(Node) ->
-    case is_key_block(Node) of
-        true ->
-            {ok, PrevNode} = db_find_node(prev_hash(Node)),
+    case {db_find_node(prev_hash(Node)), is_key_block(Node)} of
+        {error, _} ->
+            ok;
```
lucafavatella: It means that if block not connected to genesis, height is not checked and height shall be checked when block gets connected to genesis. Where is block height checked when it gets connected to genesis?

sennui: @lucafavatella in this case it should allow to `error` only when the node is genesis. If I understand correctly header removal, we don't let out of order nodes in

sennui: to sum up after chat discussion - we allow out of order nodes, but they won't reach that code area

PR #1134, opened 2018-05-15
Ignore aevm_external in docker context

dincho:
This is the aevm external tests repo witch is about 600MB and useless in docker context.


ThomasArts:
I hope we don't need the tests to be around in the docker used for CI?

dincho:
> I hope we don't need the tests to be around in the docker used for CI?

Nope

dincho:
> We should probably move to whitelist.

Good idea, https://www.pivotaltracker.com/story/show/157624989

PR #1136, opened 2018-05-15
Fallback to old version of port compiler

ThomasArts:
Fix compilation problems in quick and dirty way.
We need a ticket upstream

Somewhere above version 1.6.0 of pc (port_compiler) the environment variables are passed down to pc. This causes error in compilation stating that -o flag cannot have multiple out files.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -45,6 +45,7 @@
         {idna, {git, "https://github.com/aeternity/erlang-idna", {ref, "24bf647"}}}
        ]}.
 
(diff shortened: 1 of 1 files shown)


lucafavatella:
CI workflow on first commit of this PR had passed https://circleci.com/workflow-run/b86a7134-5304-4543-b2a2-6e144f1493fb

I do not see how a comment in rebar.config could break CI so I am considering merging. I see two approvals.

eproxus:
(For future reference, [reviews can be dismissed](https://help.github.com/articles/dismissing-a-pull-request-review/) to make the checks pass).

PR #1138, opened 2018-05-15
Pt 157355749 path and query parameters

ThomasArts:
Use the fix in the swagger_endpoints plugin version 0.2.1 to create URL from path and query maps.

And remove deprecated APIs from the swagger.yaml file as well as the tests related to it.

### DIFF
--- a/config/swagger.yaml
+++ b/config/swagger.yaml
@@ -106,61 +106,7 @@ paths:
           description: Header not found
           schema:
             $ref: '#/definitions/Error'
-      security:
-  /block-by-height:
-    get:
-      tags:
-        - external
-        - gossip
-      operationId: GetBlockByHeightDeprecated
-      description: 'Get a block by height'
-      produces:
-        - application/json
-      parameters:
-        - in: query
-          name: height
-          description: 'Height of the block to fetch'
-          required: true
-          type : integer
-      responses:
-        '200':
-          description: The block being found
-          schema:
-            $ref: '#/definitions/Block'
-        '404':
-          description: Block not found
-          schema:
-            $ref: '#/definitions/Error'
-      security:
-  /block-by-hash:
-    get:
-      tags:
-        - external
-        - gossip
-      operationId: GetBlockByHashDeprecated
-      description: 'Get a block by hash'
-      produces:
-        - application/json
-      parameters:
-        - in: query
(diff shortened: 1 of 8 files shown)


config/swagger.yaml
```diff
@@ -175,7 +122,7 @@ paths:
       parameters:
         - in: body
           name: body
-          description: Put a new block to node
+          description: A new block
```
jur0: I see some have description in single quotes, some without.

config/swagger.yaml
```diff
@@ -201,7 +148,7 @@ paths:
       parameters:
         - in: body
           name: body
-          description: Put a new transaction to node
+          description: a new transaction
```
jur0: A small typo: "A new ..."

ThomasArts: There is no verb in this description and hence it is not a sentence, therefore I was not using a capital A here.

config/swagger.yaml
ThomasArts: I left the quotation as it was originally. Swagger does not demand one or the other.

velzevur: I think quotation is required only when there are special symbols like `'`

config/swagger.yaml
uwiger: Still, it would be consistent with "A new block" above and "A spend transaction" below to use a capital A.

system_test/aest_perf_SUITE.erl
```diff
@@ -182,8 +182,11 @@ wait_for_sync(Nodes, Height, Cfg) ->
     SyncTimeout = proplists:get_value(sync_timeout, Cfg),
     wait_for_value({height, Height}, Nodes, SyncTimeout * Height, Cfg).
 
-get_block(NodeName, Height, Cfg) ->
-    request(NodeName, [v2, 'block-by-height'], #{height => Height}, Cfg).
+get_block(NodeName, Height) ->
+    case request(NodeName, 'GetBlockByHeight', #{height => Height}) of
+        {ok, 200, Block} -> Block;
+        _ -> error(undefined)
```
eproxus: Can this raise some more error information to make the test cases easier to debug? Something like `Error -> error({error_getting_block, NodeName, Height, Error})`

PR #1140, opened 2018-05-15
series of fixes to get int tests green (not yet...)

sennui:


apps/aecore/src/aec_mining.erl
```diff
@@ -28,7 +28,7 @@ create_micro_block_candidate(TopBlock, CurrentKeyBlock, TopBlockTrees) ->
 -spec create_key_block_candidate(aec_blocks:block(), aec_blocks:block(), aec_trees:trees(),
     list(aec_headers:header())) -> {ok, aec_blocks:block(), aec_pow:nonce()} | {error, term()}.
 create_key_block_candidate(TopBlock, CurrentKeyBlock, TopBlockTrees, AdjHeaders) ->
-    Height = aec_blocks:height(TopBlock) + 1,
+    Height = aec_blocks:height(CurrentKeyBlock) + 1,
```
zp-sd: `TopBlock` height should be same as `CurrentKeyBlock`? Or not?

sennui: It may be. We need info from the current key block and it may be current top (if no micro is present)

PR #1143, opened 2018-05-16
Handle more error cases in sync/peer_connection

hanssv:
This addresses https://www.pivotaltracker.com/story/show/157218156 and https://www.pivotaltracker.com/story/show/157621616


apps/aecore/src/aec_peer_connection.erl
```diff
@@ -436,7 +436,7 @@ handle_request(S, {Request, Args}, From) ->
     case get_request(S, Request) of
         none ->
             handle_request(S, Request, Args, From);
-        {MappedRequest, _From} ->
+        {MappedRequest, _From, _TRef} ->
```
lucafavatella: I see [setter](https://github.com/aeternity/epoch/blob/9736a29a264819e9474539e09f4905f057f03ae5/apps/aecore/src/aec_peer_connection.erl#L302) and other usages. Ok.

Any chance of having state with type specs?
(BTW Also I currently prefer record over map for process state - as more structured - though debatable.)

rebar.lock
```diff
@@ -21,7 +21,7 @@
   0},
  {<<"enoise">>,
   {git,"https://github.com/aeternity/enoise.git",
-       {ref,"7347906f29c86456d0e144456b524014722424c7"}},
+       {ref,"c3bdb02e53706cafceccb109f8a5ace9a37356f4"}},
```
lucafavatella: I see https://github.com/aeternity/enoise/compare/7347906...c3bdb02
ok

hanssv:
Rebased and resolved conflict in release notes... Sigh...

PR #1150, opened 2018-05-17
Work around intermittent `socket_closed_remotely` test error

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/157656809


lucafavatella:
System tests do this https://github.com/aeternity/epoch/blob/cb10ac4bf7b01dbb8759b0b9ddbe68b39f533fcc/apps/aehttp/test/aehttp_client.erl#L17-L19

lucafavatella:
I split improvements in test client that already uses this workaround in PR #1152 so to prevent unnecessary delay of this - that shall reduce intermittent test failures.

PR #1152, opened 2018-05-17
Improve usage of httpc profile in integration and system tests

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/157656809

TODO:
* [x] Ensure system tests not broken by this change - confirmed by @ThomasArts

### DIFF
--- a/apps/aehttp/test/aehttp_client.erl
+++ b/apps/aehttp/test/aehttp_client.erl
@@ -14,22 +14,22 @@ request(OpId, Params, Cfg) ->
     BaseUrl = string:trim(proplists:get_value(Interface, Cfg), trailing, "/"),
     Path = endpoints:path(Method, OpId, Params),
     Query = endpoints:query(Method, OpId, Params),
-    inets:start(httpc, [{profile, test_browser}]),
-    Response = request(Method, BaseUrl, Path, Query, Params, [], [{timeout, 15000}], [], Cfg),
-    inets:stop(httpc, test_browser),
+    {ok, ClientPid} = inets:start(httpc, [{profile, test_browser}], stand_alone),
+    Response = request(Method, BaseUrl, Path, Query, Params, [], [{timeout, 15000}], [], ClientPid, Cfg),
+    ok = inets:stop(stand_alone, ClientPid),
     Response.
 
-request(get, BaseUrl, Path, Query, _Params, Headers, HttpOpts, Opts, Cfg) ->
+request(get, BaseUrl, Path, Query, _Params, Headers, HttpOpts, Opts, Profile, Cfg) ->
     Url = binary_to_list(iolist_to_binary([BaseUrl, Path, Query])),
     log("GET ~p", [Url], Cfg),
-    Resp = httpc:request(get, {Url, Headers}, HttpOpts, Opts, test_browser),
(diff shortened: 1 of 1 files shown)


apps/aehttp/test/aehttp_client.erl
```diff
@@ -14,22 +14,22 @@ request(OpId, Params, Cfg) ->
     BaseUrl = string:trim(proplists:get_value(Interface, Cfg), trailing, "/"),
     Path = endpoints:path(Method, OpId, Params),
     Query = endpoints:query(Method, OpId, Params),
-    inets:start(httpc, [{profile, test_browser}]),
-    Response = request(Method, BaseUrl, Path, Query, Params, [], [{timeout, 15000}], [], Cfg),
-    inets:stop(httpc, test_browser),
+    {ok, ClientPid} = inets:start(httpc, [{profile, test_browser}], stand_alone),
```
ThomasArts: could you now also leave out {profile, test_browser} altogether? Not needed, but you don't seem to need it any more.

lucafavatella: > could you now also leave out {profile, test_browser} altogether?

[Not possible](https://github.com/erlang/otp/blob/OTP-20.2.3/lib/inets/src/http_client/httpc.erl#L456-L459).

And it makes sense not to use the default profile, as otherwise [standalone process would attempt to re-create (hence failing) the table of keepalive sessions](https://github.com/erlang/otp/blob/OTP-20.2.3/lib/inets/src/http_client/httpc_manager.erl#L421) of the default profile.

----

I performing a few unrelated fine-tuning.

lucafavatella:
I cannot run system tests with this PR now, so keeping labelled as WIP.

apps/aehttp/test/aehttp_client.erl
```diff
@@ -14,22 +14,22 @@ request(OpId, Params, Cfg) ->
...
+    Response = request(Method, BaseUrl, Path, Query, Params, [], [{timeout, 15000}], [], ClientPid, Cfg),
+    ok = gen_server:stop(ClientPid, normal, infinity),
     Response.
 
-request(get, BaseUrl, Path, Query, _Params, Headers, HttpOpts, Opts, Cfg) ->
+request(get, BaseUrl, Path, Query, _Params, Headers, HttpOpts, Opts, Profile, Cfg) ->
```
sylane: Is the `Profile` parameter always required now? is there no meaningful default value?
Couldn't it be shoved inside `Opts` and extracted using:
```
Profile = proplists:get_value(profile, Opts, test_browser),
Opts2 = lists:keydelete(profile, 1, Opts),
```

lucafavatella: > Is the Profile parameter always required now?

Yes. Though this is internal function so shall be ok.

> is there no meaningful default value?

Using a default value would hinder the benefits this commit adds - namely enhanced error handling and potential parallel clients. From commit message:

> * Linking httpc_manager process to test process - rather than to
>   httpc_profile_sup process;
> * Not registering httpc_manager process by (profile-based) name - so
>   to enable future concurrent clients;

I suggest to leave this as is and add default value if/when required.

> Couldn't it be shoved inside Opts ...

`Opts` is passed to `httpc:request` so I would prefer not to reuse it.

PR #1153, opened 2018-05-17
Channels: broadcast on chain transactions to participants

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/157621529)
This PR:

* introduces new channels tag `on_chain_tx` for broadcasting transactions that ar going into mempool

* refactors WS tests so they're easier to write (most notably tests are auto unsubscribed to channels' events upon a test end)


velzevur:
Will prepare a PR in protocol addressing changes.
@lucafavatella since changes are user visible, do you think it is worth mentioning in release notes?

lucafavatella:
> @lucafavatella since changes are user visible, do you think it is worth mentioning in release notes?

@velzevur I do.

velzevur:
I see 2 confirmations. Once I've posted a PR for the protocol repo, I'll merge this PR.

PR #1160, opened 2018-05-18
Increase miner reward - smallest unit is atto-Aeternity-token

hanssv:
For now go with 10^-18, will be easy to adjust if we decide on something different.

### DIFF
--- a/py/tests/integration/setup.yaml
+++ b/py/tests/integration/setup.yaml
@@ -28,7 +28,7 @@ nodes:
       external_api: 3033
       internal_api: 3133
       internal_ws: 3134
-coinbase_reward: 10
+coinbase_reward: 10000000000000000000
 tests: # test specific settings
   test_use_cases:
     test_syncing:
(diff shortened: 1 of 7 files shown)


py/tests/integration/setup.yaml
```diff
@@ -107,7 +107,7 @@ tests: # test specific settings
     test_contract_create:
       nodes:
         node: dev1
-      blocks_to_mine: 115
+      blocks_to_mine: 5
```
lucafavatella: Could you please re-decrease timeout in  2 py/tests/integration/common.py wait_until_height from 600s to something more sensible e.g. previous 120s ?

https://github.com/aeternity/epoch/commit/4ef90d62a5d0e9affc79e29a1e83e6beaa06a552#diff-aaa0ff120e51b18e0a5d4dc49ea74dd8R189

velzevur: Great!

hanssv:
Updated the timeout + added release notes and DB-path bump.

PR #1161, opened 2018-05-18
Fix allow showing transactions to a new accounts

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/157316134)

If an account address is new (not yet persisted in trees) - we used to not show account transactions.
#### Ex. scenario
Alice has an account, Bob has not yet.
Alice posts a `spend_tx` to Bob. Transaction is not yet mined but rather in mempool.
Bob wants to get all transactions related to his address


velzevur:
Not sure if worth mentioning in the release notes

PR #1165, opened 2018-05-18
[RFC] Add miner to header (part of moving coinbase to header)

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/157214052

I approached moving coinbase to header in two steps:
1. Add miner to header and use it in the code - still keeping coinbase
2. Remove coinbase (this will be the main step and break many tests)

This PR is a request for comments on the first step - it cannot be merged to master yet as the coinbase is still present in the block though not verified.

TODO:
* [ ] Ensure CI passes (I believe I fixed all tests though I still need to check a full passing CI build)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -568,9 +568,6 @@ workflows:
       - uat_tests:
           requires:
             - build
-          filters:
-            branches:
-              only: master
 
       - static_analysis:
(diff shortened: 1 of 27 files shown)


lucafavatella:
I will open updated one.

PR #1167, opened 2018-05-21
Fix crash in eper - use aeternity fork

hanssv:
Also remove the specaial case for this crash in system tests.


hanssv:
Closes: https://www.pivotaltracker.com/story/show/157215253

rebar.lock
```diff
@@ -24,8 +24,8 @@
        {ref,"c3bdb02e53706cafceccb109f8a5ace9a37356f4"}},
   0},
  {<<"eper">>,
-  {git,"git://github.com/massemanet/eper.git",
-       {ref,"805a58728a3258e78de7ae732397ffc74c91423c"}},
+  {git,"git://github.com/aeternity/eper.git",
+       {ref,"03d4be3ce7dd75ae332fc21ba7d5c18678affc5b"}},
```
lucafavatella: I see https://github.com/aeternity/eper/compare/805a58728a3258e78de7ae732397ffc74c91423c...03d4be3ce7dd75ae332fc21ba7d5c18678affc5b

Only comment is why catch `_:_` rather than `error:badarg` though debatable. Ok.

I see PR upstream. Ok.

hanssv:
Hmpf... Different Erlang versions means different versions of `runtime_tools` - new version with proper filtering!

PR #1168, opened 2018-05-21
Don't close the socket to early, the same port can be returned

ThomasArts:
We need 4 different TCP ports per container to map configured ports to actually used ports.

The way to get a port is to open a socket and use inets:port on this socket. That should give a unique port for that moment. However, when the socket is closed, that port may be re-used when inets:port is called again.

This does not solve the complete port unicity. Since we create nodes in setup, we may in the second setup accidently get the same port.


sylane:
I would only remove the `gen_tcp:close(Socket)` from the `free_port` function and maybe check the result of `gen_tcp:close` in `start_node` (that would have detected the mistake).

PR #1171, opened 2018-05-22
PT-155764172 Sophia oracles

UlfNorell:
Sophia front-end part of Oracle interactions. Nothing yet to hook them up to, so mostly untested.

Should only affects the Sophia compiler.


lucafavatella:
I took the liberty of moving WIP from title to tag - so to ease PR filtering.

UlfNorell:
Will make a new PR when chain side is in place.

PR #1173, opened 2018-05-22
PT-157032168: Add quick start/stop test

eproxus:


eproxus:
For me it's not failing. Let's see on other machines 🙂

PR #1174, opened 2018-05-22
Release preparation

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/157112368

TODO:
* [x] Rebase on current master
* [x] Polish release notes


lucafavatella:
I went through release notes as user would do (installation, configuration, operation), from build from code in this branch (that is on recent master). Hence I found the pending API small error in the doc - corrected in this PR.

docs/operation.md
```diff
@@ -7,8 +7,7 @@ This document describes how to start your epoch node installed using a release b
 The instructions below assume that:
 * The node is deployed in directory `/tmp/node`;
 * No custom peers are specified under the `peers:` key in the config. If the `peers:` key is undefined, the *testnet* seed peers (built-in in the package source) are used.
-* The external HTTP endpoint of the user API of the node can be contacted at 127.0.0.1 port 3003;
-* The internal HTTP endpoint of the user API of the node can be contacted at 127.0.0.1 port 3103.
+* The external HTTP endpoint of the user API of the node can be contacted at 127.0.0.1 port 3003.
```
dincho: I thought the default port of external API is 3013?

https://github.com/aeternity/epoch/blob/master/config/sys.config#L10

lucafavatella: For as long as this document operation.md is internally consistent, it does not matter.

Also, configuration doc suggests https://github.com/aeternity/epoch/blob/49a57291c2a9e9176a4abdc34d452730d1385191/docs/configuration.md#instructions

dincho: I guess you're assuming the user is following the configuration.md ?

However, most of the users are really confused about 3003/3013 (the tickets/questions I've seen), what about using default values in configuration.md ?

docs/operation.md
```diff
@@ -7,8 +7,7 @@ This document describes how to start your epoch node installed using a release b
 The instructions below assume that:
 * The node is deployed in directory `/tmp/node`;
 * No custom peers are specified under the `peers:` key in the config. If the `peers:` key is undefined, the *testnet* seed peers (built-in in the package source) are used.
-* The external HTTP endpoint of the user API of the node can be contacted at 127.0.0.1 port 3003;
-* The internal HTTP endpoint of the user API of the node can be contacted at 127.0.0.1 port 3103.
```
dincho: I see the internal API is not used in this docs, thus the removal of the line?

lucafavatella: Yes.

docs/operation.md
lucafavatella: > However, most of the users are really confused about 3003/3013 (the tickets/questions I've seen), what about using default values in configuration.md ?

Ok. I will propose change for that.

lucafavatella: @dincho Addressed in new commit. Could you please re-review?

(I am preparing PR also for https://github.com/aeternity/protocol )

PR #1175, opened 2018-05-23
PT-157781241: Enhance quick start/stop system test

eproxus:
This PR makes it so the test verifies that the restarted node behaves well by making sure it restarts/syncs to the old height. Further, it changes `wait_for_value/4` to return the values waited for (e.g. for height, it returns the blocks at those heights) so they can be used for verification. It also introduces the `wait_for_startup/3` shortcut from `aest_perf_SUITE` to the testing library so the same configuration can be used in all test suites.

### DIFF
--- a/system_test/aest_perf_SUITE.erl
+++ b/system_test/aest_perf_SUITE.erl
@@ -67,6 +67,7 @@ init_per_suite(Cfg) ->
         {mine_interval, {seconds, 30}}, % Interval to check mining height
         {mine_rate, MineRate},          % Mine rate configuration for nodes
         {mine_timeout, MineRate * 2},   % Per block
+        {startup_timeout, 20000},       % Timeout until HTTP API responds
         {sync_timeout, 1000}            % Per block
     ].
 
--- a/system_test/aest_perf_SUITE.erl
+++ b/system_test/aest_perf_SUITE.erl
@@ -32,6 +32,7 @@
     wait_for_value/4,
     wait_for_time/4,
     get_block/2,
+    wait_for_startup/3,
     time_to_ms/1
 ]).
 
@@ -190,10 +191,6 @@ sync_node(Node, Height, Cfg) ->
     % Sanity check that syncing does not take longer than mining
     ?assert(End - Start =< time_to_ms(?cfg(mine_time))),
     Block.
-
-wait_for_startup(Nodes, Height, Cfg) ->
-    StartupTimeout = proplists:get_value(startup_timeout, Cfg),
-    wait_for_value({height, Height}, Nodes, StartupTimeout, Cfg).
 
 wait_for_sync(Nodes, Height, Cfg) ->
     SyncTimeout = proplists:get_value(sync_timeout, Cfg),
--- a/system_test/aest_perf_SUITE.erl
+++ b/system_test/aest_perf_SUITE.erl
@@ -67,7 +67,6 @@ init_per_suite(Cfg) ->
         {mine_interval, {seconds, 30}}, % Interval to check mining height
         {mine_rate, MineRate},          % Mine rate configuration for nodes
         {mine_timeout, MineRate * 2},   % Per block
-        {startup_timeout, 20000},       % Timeout until HTTP API responds
         {sync_timeout, 1000}            % Per block
     ].
 
--- a/system_test/aest_perf_SUITE.erl
+++ b/system_test/aest_perf_SUITE.erl
@@ -67,7 +67,7 @@ init_per_suite(Cfg) ->
         {mine_interval, {seconds, 30}}, % Interval to check mining height
         {mine_rate, MineRate},          % Mine rate configuration for nodes
         {mine_timeout, MineRate * 2},   % Per block
-        {startup_timeout, 20000},       % Timeout until HTTP API responds
+        {node_startup_timeout, 20000},  % Timeout until HTTP API responds
         {sync_timeout, 1000}            % Per block
     ].
 
@@ -118,12 +118,12 @@ startup_speed_mining(Cfg) ->
     int_startup_speed(true, Cfg).
 
 sync_speed(Cfg) ->
-    % This tests starts 4 nodes at the pre-mined height, then adds and syncs two
+    % This tests starts 3 nodes at the pre-mined height, then adds and syncs two
     % more one at a time. Syncing is verified to not take longer than mining and
     % then the measured time is logged.
     [Height, MineRate] = ?cfg([height, mine_rate]),
 
-    InitialNodes = [n1, n2, n3, n4],
+    InitialNodes = [n1, n2, n3],
 
     InitialNodeSpecs = [
         spec(N, InitialNodes -- [N], #{
@@ -154,23 +154,21 @@ sync_speed(Cfg) ->
     [?assertEqual(AB, BB) || {AN, AB} <- Blocks, {BN, BB} <- Blocks, AN =/= BN].
 
 stay_in_sync(Cfg) ->
-    % This test starts 6 nodes at the pre-mined height and then mines for an
+    % This test starts N nodes at the pre-mined height and then mines for an
     % additional 20% of the original mining time. It asserts that the network
     % does not fork.
     Height = ?cfg(height),
-    Nodes = [n1, n2, n3, n4, n5, n6],
+    Nodes = [n1, n2, n3, n4, n5],
     setup_nodes(cluster(Nodes, #{
         mine_rate => ?cfg(mine_rate),
         source => ?cfg(source)
     }), Cfg),
     [start_node(N, Cfg) || N <- Nodes],
     wait_for_startup(Nodes, Height, Cfg),
-    Fifth = floor(time_to_ms(?cfg(mine_time)) / 5),
-    Reached = wait_for_time(height, Nodes, Fifth, #{
-        interval => ?cfg(mine_interval)
-    }),
 
-    Blocks = [{N, get_block(N, Reached)} || N <- Nodes],
+    wait_for_value({height, Height + 41}, Nodes, 41*2000, Cfg),
+
+    Blocks = [{N, get_block(N, Height + 40)} || N <- Nodes],
     [?assertEqual(AB, BB) || {AN, AB} <- Blocks, {BN, BB} <- Blocks, AN =/= BN].
 
 %=== INTERNAL FUNCTIONS ========================================================
--- a/system_test/aest_sync_SUITE.erl
+++ b/system_test/aest_sync_SUITE.erl
@@ -27,6 +27,7 @@
     stop_node/3,
     connect_node/3, disconnect_node/3,
     wait_for_value/4,
+    wait_for_startup/3,
     get_block/2,
     get_top/1,
     request/3
@@ -243,7 +244,7 @@ new_node_joins_network(Cfg) ->
 %% that we had in the chain before stopping: data is persistent.
 docker_keeps_data(Cfg) ->
     Length = 20,
-    NodeStartupTime = proplists:get_value(node_startup_time, Cfg),
+    NodeStartupTime = proplists:get_value(startup_timeout, Cfg),
     NodeShutdownTime = proplists:get_value(node_shutdown_time, Cfg),
 
     setup_nodes([?STANDALONE_NODE], Cfg),
@@ -706,13 +707,15 @@ net_split_mining_power(Cfg) ->
     ok.
 
 quick_start_stop(Cfg) ->
-    setup_nodes(cluster([n1, n2], #{}), Cfg),
-    start_node(n2, Cfg),
-    start_node(n1, Cfg),
+    Nodes = [n1, n2],
+    setup_nodes(cluster(Nodes, #{}), Cfg),
+    [start_node(N, Cfg) || N <- Nodes],
+    Blocks = wait_for_value({height, 5}, Nodes, 5 * ?MINING_TIMEOUT, Cfg),
     stop_node(n2, 2000, Cfg),
     timer:sleep(2000),
     start_node(n2, Cfg),
-    ok.
+    NewBlocks = wait_for_startup(Nodes, 5, Cfg),
+    ?assertEqual(Blocks, NewBlocks).
 
 %% helper functions
 
--- a/system_test/aest_sync_SUITE.erl
+++ b/system_test/aest_sync_SUITE.erl
@@ -163,7 +163,6 @@ patron() ->
 %% A node with a newer version of the code can join and synchronize
 %% to a cluster of older nodes.
 new_node_joins_network(Cfg) ->
-    NodeStartupTime = proplists:get_value(node_startup_time, Cfg),
     Compatible = "aeternity/epoch:local", %% Latest version it should be compatible with
                                           %% Change if comptibility with previous version
                                           %% should be guaranteed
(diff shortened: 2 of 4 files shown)


system_test/aest_sync_SUITE.erl
```diff
@@ -428,13 +426,15 @@ net_split_recovery(Cfg) ->
...
 
 quick_start_stop(Cfg) ->
-    setup_nodes(cluster([n1, n2], #{}), Cfg),
-    start_node(n2, Cfg),
-    start_node(n1, Cfg),
+    Nodes = [n1, n2],
+    setup_nodes(cluster(Nodes, #{}), Cfg),
+    [start_node(N, Cfg) || N <- Nodes],
+    Blocks = wait_for_value({height, 5}, Nodes, 5 * ?MINING_TIMEOUT, Cfg),
     stop_node(n2, 2000, Cfg),
```
sylane: This will never fail because the node got time to start properly. What I tested was:
 - Start both nodes; stop node 2 right away (should crash); restart node 2; check they synchronize.
 - Start both nodes; let them build some chain; stop node 2; start node 2 and stop it right away (couldn't make it crash); restart node 2; check they synchronize.

eproxus: I don't follow. What actual changes do you request?

sylane: In your test, as I understand it, you build a chain before stopping node2; this gives time for node 2 to start so it will never crash.

sylane: I described the scenarios I think that should be test.

lucafavatella:
I see ongoing comments so I take liberty of marking as WIP to ease filtering.

eproxus:
Current state for me is that the `abrupt_stop_new_node` case either:

* passes
* fails with log errors
* fails to synchronize at all

The second test `abrupt_stop_mining_node` mostly fails with log errors.

@ThomasArts This sounds like a good area for QuickCheck, to determine the stop timeout to trigger errors?

eproxus:
@ThomasArts @sylane Is this still something you think should be merged?

system_test/aest_sync_SUITE.erl
```diff
@@ -667,14 +667,44 @@ net_split_mining_power(Cfg) ->
...
+    % Let them build some chain
+    Blocks = wait_for_value({height, 5}, Nodes, 5 * ?MINING_TIMEOUT, Cfg),
+    assert_in_sync(Blocks),
+    % Stop node 2
+    stop_node(n2, NodeShutdownTime, Cfg),
+    % Start node 2
+    start_node(n2, Cfg),
+    % Stop node 2 abruptly
     stop_node(n2, 2000, Cfg),
-    timer:sleep(2000),
```
ThomasArts: I think this sleep was needed to really disconnect (otherwise node 2 was not really stopped). But this might have been fixed in the docker framwork

ThomasArts:
System tests pass locally for slower machines and reveal real bug on faster machines that have an issue with inet:getaddr.
It's not in the way for a merge, because on CI these tests passed and will not bother people more than needed.

```
%%% aest_channels_SUITE ==> test_simple_same_node_channel: OK
%%% aest_channels_SUITE ==> test_simple_different_nodes_channel: OK
%%% aest_commands_SUITE ==> epoch_commands: OK
%%% aest_peers_SUITE ==> test_peer_discovery: OK
%%% aest_peers_SUITE ==> test_inbound_limitation: OK
%%% aest_perf_SUITE ==> long_chain.startup_speed: OK
%%% aest_perf_SUITE ==> long_chain.startup_speed_mining: OK
%%% aest_perf_SUITE ==> long_chain.sync_speed: OK
%%% aest_perf_SUITE ==> long_chain.stay_in_sync: OK
%%% aest_sync_SUITE ==> new_node_joins_network: OK
%%% aest_sync_SUITE ==> docker_keeps_data: OK
%%% aest_sync_SUITE ==> stop_and_continue_sync: OK
%%% aest_sync_SUITE ==> tx_pool_sync: OK
%%% aest_sync_SUITE ==> net_split_recovery: OK
%%% aest_sync_SUITE ==> net_split_mining_power: OK
%%% aest_sync_SUITE ==> abrupt_stop_new_node: SKIPPED
%%% aest_sync_SUITE ==> {tc_user_skip,database_restart_needs_fix}
%%% aest_sync_SUITE ==> abrupt_stop_mining_node: SKIPPED
%%% aest_sync_SUITE ==> {tc_user_skip,database_restart_needs_fix}
```

system_test/helpers/aest_nodes.erl
```diff
@@ -1,5 +1,7 @@
 -module(aest_nodes).
 
+-include_lib("eunit/include/eunit.hrl").
```
lucafavatella: Do you mean `-include_lib("stdlib/include/assert.hrl").` - for `?assertEqual`? Anyway this file is test framework so not an issue.

system_test/aest_perf_SUITE.erl
```diff
@@ -157,19 +158,17 @@ stay_in_sync(Cfg) ->
     % additional 20% of the original mining time. It asserts that the network
     % does not fork.
     Height = ?cfg(height),
-    Nodes = [n1, n2, n3, n4, n5, n6],
+    Nodes = [n1, n2, n3, n4, n5],
```
lucafavatella: Comment above mentions 6 nodes.

lucafavatella: Is this independent change for reducing resources needed by running tests?

system_test/aest_perf_SUITE.erl
```diff
@@ -157,19 +158,17 @@ stay_in_sync(Cfg) ->
...
     }), Cfg),
     [start_node(N, Cfg) || N <- Nodes],
     wait_for_startup(Nodes, Height, Cfg),
-    Fifth = floor(time_to_ms(?cfg(mine_time)) / 5),
-    Reached = wait_for_time(height, Nodes, Fifth, #{
-        interval => ?cfg(mine_interval)
-    }),
 
-    Blocks = [{N, get_block(N, Reached)} || N <- Nodes],
+    wait_for_value({height, Height + 41}, Nodes, 41*2000, Cfg),
```
lucafavatella: Comment above explicitly mentions 20%, this change loses this correlation.

system_test/aest_perf_SUITE.erl
```diff
@@ -157,19 +158,17 @@ stay_in_sync(Cfg) ->
...
+
+    Blocks = [{N, get_block(N, Height + 40)} || N <- Nodes],
```
lucafavatella: It looks like the amendment you want to keep is only `get_block(N, Reached - 1)` (notice `- 1`) for solving ties?

system_test/aest_sync_SUITE.erl
```diff
@@ -705,14 +707,48 @@ net_split_mining_power(Cfg) ->
...
+abrupt_stop_new_node(_Cfg) ->
+    {skip, database_restart_needs_fix};
+abrupt_stop_new_node(Cfg) ->
+    RepairTimeout = 30000, % Time allowed for node to repair DB and finish sync
+    Nodes = [n1, n2],
+    setup_nodes(cluster(Nodes, #{}), Cfg),
+    % Start both nodes
+    [start_node(N, Cfg) || N <- Nodes],
+    % Stop node 2 abruptly
+    stop_node(n2, 2000, Cfg),
```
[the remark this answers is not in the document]
eproxus: Using stop will effectively kill the container when the timeout expires. The alternative equivalent to this would be:
```
timer:sleep(2000),
stop_node(n2, 0, Cfg),
```
If we kill it too early, the node will not have time to start.

system_test/aest_sync_SUITE.erl
```diff
@@ -201,8 +204,8 @@ new_node_joins_network(Cfg) ->
     StartTime = os:timestamp(),
 
 
-    inject_spend_txs(old_node1, patron(), 20, 1, 100),
-    inject_spend_txs(old_node2, patron(), 20, 21, 100),
+    inject_spend_txs(old_node1, patron(), 5, 1, 100),
```
lucafavatella: I see these reductions in number of spend txs are independent change for making test faster. Ok.

system_test/aest_sync_SUITE.erl
```diff
@@ -262,11 +264,11 @@ docker_keeps_data(Cfg) ->
     %% Get all blocks before stopping
     A = [get_block(standalone_node, H) || H <- lists:seq(1, Length)],
 
-    stop_node(standalone_node, NodeShutdownTime, Cfg),
+    stop_node(standalone_node, ShutdownTimeout, Cfg),
```
lucafavatella: I see this is waiting 20s for bin/epoch stop. Ok.

system_test/aest_sync_SUITE.erl
```diff
@@ -625,18 +627,18 @@ net_split_mining_power(Cfg) ->
...
 
-    Net1Nodes = [net1_node1, net1_node2],
-    Net2Nodes = [net2_node1, net2_node2, net2_node3, net2_node4],
+    Net1Nodes = [net1_node1],
+    Net2Nodes = [net2_node1, net2_node2, net2_node3],
     AllNodes = Net1Nodes ++ Net2Nodes,
 
-    setup_nodes([?NET1_NODE1, ?NET1_NODE2, ?NET2_NODE1,
-                 ?NET2_NODE2, ?NET2_NODE3, ?NET2_NODE4], Cfg),
+    setup_nodes([?NET1_NODE1, ?NET1_NODE2,
```
lucafavatella: Why net1_node2 setup but not started? Leftover?

system_test/aest_sync_SUITE.erl
```diff
@@ -625,18 +627,18 @@ net_split_mining_power(Cfg) ->
...
     ExtraLength = 3,
 
-    Net1Nodes = [net1_node1, net1_node2],
-    Net2Nodes = [net2_node1, net2_node2, net2_node3, net2_node4],
+    Net1Nodes = [net1_node1],
+    Net2Nodes = [net2_node1, net2_node2, net2_node3],
     AllNodes = Net1Nodes ++ Net2Nodes,
 
-    setup_nodes([?NET1_NODE1, ?NET1_NODE2, ?NET2_NODE1,
-                 ?NET2_NODE2, ?NET2_NODE3, ?NET2_NODE4], Cfg),
```
lucafavatella: As now unused, may I suggest deleting NET2_NODE4 macro definition?

system_test/aest_sync_SUITE.erl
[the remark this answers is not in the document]
lucafavatella: > Using stop will effectively kill the container when the timeout expires.

On a fast machine the timeout may not expire - i.e. clean stop may complete before 2s. Hence this test would stop gently, not abruptly (test case name is `abrupt_stop_new_node`).

> The alternative equivalent to this would be:
> 
> timer:sleep(2000),
> stop_node(n2, 0, Cfg),

It is quite difficult to define equivalent with timeout outside the call. Also the [stop retries 5 times with the same timeout](https://github.com/aeternity/epoch/blob/acc7fd0651a3e6ac79e5195126494519f25fbe6a/system_test/helpers/aest_docker.erl#L460).

> If we kill it too early, the node will not have time to start.

Yes, test shall check node started before stopping it.


It feels like test here shall do:
```
wait_for_startup(n2, 0, Cfg),
aest_nodes:kill_node(n2, Cfg)
```

system_test/aest_sync_SUITE.erl
```diff
@@ -705,14 +707,48 @@ net_split_mining_power(Cfg) ->
...
+    ShutdownTimeout = proplists:get_value(node_shutdown_time, Cfg),
+    Nodes = [n1, n2],
+    setup_nodes(cluster(Nodes, #{}), Cfg),
+    % Start both nodes
+    [start_node(N, Cfg) || N <- Nodes],
+    % Let them build some chain
+    Blocks = wait_for_value({height, 5}, Nodes, 5 * ?MINING_TIMEOUT, Cfg),
+    assert_in_sync(Blocks),
+    % Stop node 2
+    stop_node(n2, ShutdownTimeout, Cfg),
```
lucafavatella: Re usage of stop_node in this test case - same as above. To be clarified how abruptly node shall be stopped.

system_test/aest_sync_SUITE.erl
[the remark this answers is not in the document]
eproxus: That would be the exact opposite of an abrupt start. We currently trigger the DB corruption problem on most machines with this setup. I suggest we keep it until we come up with something better.

system_test/aest_sync_SUITE.erl
```diff
@@ -705,14 +707,48 @@ net_split_mining_power(Cfg) ->
...
+    % Stop node 2 abruptly
+    stop_node(n2, 2000, Cfg),
+    % Restart node 2
+    start_node(n2, Cfg),
+    % Check that they synchronize
+    Blocks = wait_for_value({height, 5}, Nodes, RepairTimeout, Cfg),
+    assert_in_sync(Blocks).
+
+abrupt_stop_mining_node(_Cfg) ->
+    {skip, database_restart_needs_fix};
```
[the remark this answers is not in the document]
ThomasArts: No, this is PT-159518268. When that is solved, we can actually restart the system more reliably.

system_test/aest_sync_SUITE.erl
```diff
@@ -705,14 +707,48 @@ net_split_mining_power(Cfg) ->
 
     ok.
 
-quick_start_stop(Cfg) ->
-    setup_nodes(cluster([n1, n2], #{}), Cfg),
+abrupt_stop_new_node(_Cfg) ->
+    {skip, database_restart_needs_fix};
```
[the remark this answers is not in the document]
ThomasArts: Same here PT-159518268.

system_test/aest_perf_SUITE.erl
ThomasArts: The comment is outdated :-).

system_test/aest_sync_SUITE.erl
ThomasArts: It's not a left-over, the network is made for 6 nodes and they try to find each others addresses. It would crash hard if the node looking for is not configured, but for not starting it the penalty is low.

In this way we can easily increase the nr of nodes if one has a fast machine and needs to experiment a bit.

system_test/aest_sync_SUITE.erl
ThomasArts: No, cannot delete it, because it is refered to in the other nodes.

lucafavatella: I do not see net2_node4 referred to in the whole https://github.com/aeternity/epoch/blob/acc7fd0651a3e6ac79e5195126494519f25fbe6a/system_test/aest_sync_SUITE.erl

system_test/aest_sync_SUITE.erl
ThomasArts: Not just making them faster, also to make sure we don't spend too much time posting, giving the chain the possibility to grow over the target height.

lucafavatella:
I see CI job test [failed](https://circleci.com/gh/aeternity/epoch/26784):
```
%%% aecore_sync_SUITE ==> all_nodes.two_nodes.ensure_tx_pools_empty: FAILED
%%% aecore_sync_SUITE ==> {test_case_failed,{retry_exhausted,#{info =>
                                         {297,ensure_tx_pools_n_txs_,0,
                                          [epoch_dev1@localhost,
                                           epoch_dev2@localhost]},
                                     prev => undefined,retries => 5,
                                     tries => 0}}}
```

sylane:
I run sync suite 10 times. Sometimes 3 tests are skipped, sometimes only 2, but no failure...

PR #1180, opened 2018-05-23
SC deposit support

uwiger:
First working test case for state channel deposits. Erlang API only.

See https://www.pivotaltracker.com/story/show/157506334

### DIFF
--- a/apps/aechannel/src/aesc_deposit_tx.erl
+++ b/apps/aechannel/src/aesc_deposit_tx.erl
@@ -16,6 +16,7 @@
          fee/1,
          nonce/1,
          origin/1,
+         amount/1,
          check/5,
          process/5,
          accounts/1,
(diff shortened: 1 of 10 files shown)


lucafavatella:
I understand not targeted for this release so I am taking the liberty of labelling as WIP so to ease filtering.

apps/aechannel/src/aesc_deposit_tx.erl
```diff
@@ -204,3 +213,23 @@ check_channel(ChannelId, FromPubKey, Trees) ->
...
+        true ->
+            N;
+        false ->
+            erlang:error(invalid_ttl)
+    end;
+adjust_ttl(_) ->
+    erlang:error(invalid_ttl).
+
+
+cur_height() ->
```
gorillainduction: This makes CI break, since the chain is not running on the CT node.

PR #1181, opened 2018-05-23
PT-157807187 fixed incorrect account_pubkey decoding instead of oracle_pubkey

thepiwo:
https://www.pivotaltracker.com/n/projects/2124891/stories/157807187

### DIFF
--- a/apps/aehttp/src/aehttp_dispatch_ext.erl
+++ b/apps/aehttp/src/aehttp_dispatch_ext.erl
@@ -242,7 +242,7 @@ handle_request('PostOracleResponse', #{'OracleResponseTx' := Req}, _Context) ->
     ParseFuns = [parse_map_to_atom_keys(),
                  read_required_params([oracle, query_id,
                                        response, fee]),
-                 base58_decode([{oracle, oracle, account_pubkey},
+                 base58_decode([{oracle, oracle, oracle_pubkey},
                                {query_id, query_id, oracle_query_id}]),
                  get_nonce(oracle),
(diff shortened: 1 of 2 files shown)


apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -242,7 +242,7 @@ handle_request('PostOracleResponse', #{'OracleResponseTx' := Req}, _Context) ->
     ParseFuns = [parse_map_to_atom_keys(),
                  read_required_params([oracle, query_id,
                                        response, fee]),
-                 base58_decode([{oracle, oracle, account_pubkey},
+                 base58_decode([{oracle, oracle, oracle_pubkey},
```
lucafavatella: I see this affects https://github.com/aeternity/epoch/blob/25a5501ec5dd8bec1e28f3b473134b079a36f02e/config/swagger.yaml#L2287 that is used in at least two paths.

thepiwo: but oracle_pubkey is EncodedHash as well, right?

lucafavatella: It looks like [not decoded for `PostOracleResponseTx`](https://github.com/aeternity/epoch/blob/78e3bad9f0a06c2832240e8adda8ddf4baf72a42/apps/aehttp/src/aehttp_dispatch_int.erl#L105) that uses same Swagger definition?

PR #1183, opened 2018-05-23
Pt 155764127 oracle integration in vm

happi:
@UlfNorell I have managed to get aec_vm_chain to implement the oracle interface.

There are a number of things that should be fixed or moved to tickets:

- [x] Add oracle contract tests to aehttp_integration_SUITE [#158023312]
- [x] Remove the debug IO.
- [x] Think through the correct handling of fees when doing oracle transactions from contracts. (and fix code accordingly.) [#158023291]
- [x] Handle oracles on other accounts than the current contract -- including a safe way of doing signing. [#158023244]
- [x] Reduce one step of decoding-encoding when passing "sophia type" arguments VM->erlang->chain and chain->erlang->VM. (On the chain the query/response types are encoded as sophia terms and the query and the response are also encoded as sophia data but when passing through vm_chain_api they are encoded in Erlang terms.) [#158023173]


apps/aevm/src/aevm_ae_primops.erl
```diff
@@ -0,0 +1,116 @@
...
+-define(PRIMOP_BASE_ADDRESS, 96). %% Byte offset for data in primops
+-spec call( non_neg_integer(), binary(), aevm_eeevm_state:state()) ->
+                  {ok, binary(), non_neg_integer(), aevm_eeevm_state:state()}
+                      | {error, any()}.
+call(Value, Data, State) ->
+    %% TODO: use aeso_data:from_binary() (but it doesn't take a base address at the moment)
+    case Data of
+        <<?BASE_ADDRESS:256, ?PRIM_CALL_SPEND:256, Recipient:256>> ->
+            spend(Recipient, Value, State);
+	<<?PRIMOP_BASE_ADDRESS:256, _:256, Type:256, Argument/binary>> ->
```
happi: @UlfNorell This was not the encoding I expected.

apps/aesophia/test/contracts/oracles.aes
```diff
@@ -4,9 +4,9 @@ contract Oracles =
   function registerOracle(acct : address,
                           sign : int,   // Signed account address
                           fee  : int,
-                          ttl  : int) : oracle(string, int) =
+                          ttl  : int) : oracle(int, int) = // oracle(string, int) =
```
happi: @UlfNorell 
Using the type oracle(string, int) crashes at contract run time.
I will investigate further.

apps/aevm/src/aevm_ae_primops.erl
UlfNorell: Matching on the pointers is not a good idea (`?PRIMOP_BASE_ADDRESS`). The encoding scheme doesn't uniquely determine the memory layout of the components. In this case the argument tuple happens to start at address 96. We really should parse the data.

happi: OK, thanks I’ll do that then.

apps/aecontract/src/aect_create_tx.erl
```diff
@@ -275,11 +275,11 @@ run_contract(#contract_create_tx{ owner      = Caller
 initialize_contract(#contract_create_tx{owner      = OwnerPubKey,
                                         nonce      = Nonce,
                                         vm_version = VmVersion,
-                                        amount     = _Amount,
-                                        gas        = _Gas,
+                                        amount     =_Amount,
```
hanssv: So the missing spaces are done on purpose? I should maybe stop "fixing" them then?!

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -384,6 +387,28 @@ sophia_spend(_Cfg) ->
...
+%%  - Handling of fees
+%%  - Failing calls
+sophia_oracles(_Cfg) ->
+    state(aect_test_utils:new_state()),
+    Acc               = ?call(new_account, 1000000),
+    Ct = <<CtId:256>> = ?call(create_contract, Acc, oracles, {}, #{amount => 100000}),
+    QueryFee          = 100,
+    TTL               = 15,
+    CtId              = ?call(call_contract, Acc, Ct, registerOracle, word, {CtId, 0, 10, QueryFee, TTL}),
+    Question          = <<"Manchester United vs Brommapojkarna">>,
```
hanssv: Very realistic scenario...

apps/aecore/src/aec_vm_chain.erl
```diff
@@ -71,14 +86,218 @@ spend(Recipient, Amount, State = #state{ trees   = Trees,
...
+    Nonce = aec_accounts:nonce(Account) + 1,
+    BinaryQuerySpec = aeso_data:to_binary(QuerySpec, 0),
+    BinaryResponseSpec = aeso_data:to_binary(ResponseSpec, 0),
+    Spec =
+        #{account       => AccountKey,
+          nonce         => Nonce,
+          query_spec    => BinaryQuerySpec,
+          response_spec => BinaryResponseSpec,
+          query_fee     => QueryFee,
+          oracle_ttl    => {delta, TTL},
```
hanssv: Having the same TTL in two places does look strange?!

apps/aecore/src/aec_vm_chain.erl
```diff
@@ -71,14 +86,218 @@ spend(Recipient, Amount, State = #state{ trees   = Trees,
...
+    {value, Account} = aec_accounts_trees:lookup(ContractKey, AT),
+    Nonce = aec_accounts:nonce(Account) + 1,
+
+    {ok, Tx} = aeo_response_tx:new(
+                 #{oracle   => Oracle,
+                   nonce    => Nonce,
+                   query_id => QueryId,
+                   response => aeso_data:to_binary(Response, 0),
+                   fee      => 0,
+                   ttl      => 1
```
hanssv: This means block height <= 1

apps/aecontract/src/aect_create_tx.erl
happi: Yes, Tobias likes it that way, I am not so sure myself...

apps/aesophia/src/aeso_abi.erl
```diff
@@ -52,6 +54,8 @@ encode_function(_Contract, Function) ->
 
 %% TODO: Handle all sophia data types.
 get_type(I) when is_integer(I) -> int;
+get_type(none) -> nil;
+get_type({some, X}) -> {option, get_type(X)};
```
lucafavatella: I see `option`, `None` and `Some` first mentioned [at this point in the spec](https://github.com/aeternity/protocol/blob/cde08c291750614a5908973100657ccbd69bf92f/contracts/sophia.md#algebraic-data-types). Maybe they shall be defined before that?

apps/aesophia/src/aeso_ast_to_icode.erl
```diff
@@ -235,6 +237,15 @@ ast_body({record,Attrs,{typed,_,Record,RecType={record_t,Fields}},Update}) ->
 ast_body({typed, _, Body, _}) ->
     ast_body(Body).
 
+%% Implemented as a contract call to the contract with address 0.
+prim_call(Prim, Amount, Args, ArgTypes, OutType) ->
+    #prim_call_contract{ gas      = #integer{ value = 0 },
+                         address  = #integer{ value = 0 },
```
lucafavatella: I see address 0 mentioned in (technical) documentation in module [`aevm_ae_primops`](https://github.com/aeternity/epoch/pull/1183/files#diff-a14025d96267bdad00e5afcaf65f98c8R5).

Is it address in ROM or memory or storage or (unlikely) contract tree under consensus?

It seems this address 0 not documented in protocol repo? Maybe shall fit [here](https://github.com/aeternity/protocol/blob/cde08c291750614a5908973100657ccbd69bf92f/contracts/sophia.md#contract-primitives)?

Maybe the hardcoded addresses shall be preferred in the code as macros or functions (rather than hardcoded values)?

apps/aeoracle/src/aeo_query_tx.erl
```diff
@@ -117,12 +117,16 @@ origin(#oracle_query_tx{sender = SenderPubKey}) ->
...
+                       fee = Fee} = QTx, Context, Trees, Height, _ConsensusVersion) ->
     Checks =
         [fun() -> aetx_utils:check_account(SenderPubKey, Trees, Nonce, Fee + QFee) end,
-         fun() -> aeo_utils:check_ttl_fee(Height, QTTL, Fee - ?ORACLE_QUERY_TX_FEE) end,
-         fun() -> check_oracle(OraclePubKey, Trees, QFee, Height, QTTL, RTTL) end,
+         fun() -> check_oracle(OraclePubKey, Trees, QFee, Height, OTTL, RTTL) end,
          fun() -> check_query(QTx, Trees, Height) end
+         | case Context of
+               aetx_contract -> [];
+               _ ->
```
lucafavatella: I would suggest to specify this clause with `aetx_transaction` so to prevent errors to slip through in case of refactoring or typos.

apps/aeoracle/src/aeo_response_tx.erl
```diff
@@ -93,19 +93,22 @@ origin(#oracle_response_tx{oracle = OraclePubKey}) ->
...
-                 fun() -> check_query(I, OraclePubKey) end,
-                 fun() -> aetx_utils:check_account(OraclePubKey, Trees,
-                                                   Nonce, Fee - QueryFee) end,
-                 fun() -> aeo_utils:check_ttl_fee(Height, ResponseTTL,
-                                                  Fee - ?ORACLE_RESPONSE_TX_FEE) end
-                ],
+                 fun() -> check_query(I, OraclePubKey) end |
+                 case Context of
+                     aetx_contract -> []; %% TODO: Handle fees from contracts right.
+                     _ ->
```
lucafavatella: I would suggest to specify this clause with `aetx_transaction` so to prevent errors to slip through in case of refactoring or typos.

apps/aesophia/src/aeso_ast_to_icode.erl
happi: There is a ticket for documenting this https://www.pivotaltracker.com/story/show/158023467

PR #1188, opened 2018-05-24
Move block candidate generation

hanssv:
Also refactor things out of aec_trees/aec_mining and collect in aec_block_candidate.

The change is complicated by the fact that we want to be able to update block candidates
this we have to a bit careful when applying miner fee etc.

This will make candidate generation a separate functionality, it will listen for "new top block" and "new tx" events and re-calculate a candidate accordingly. It will emit "candidate_block" events that anyone interested can listen to (at the moment only the conductor is interested :-))


sennui:
@hanssv I see that this is a push model from the new generator to conductor. Its quite opposite from what we talked about ;) (with Tobias)

Conductor will maintain two worker loops (one for key blocks, one for micro blocks).
If the micro loop sleeps for 3 sec (to satisfy NG protocol) and we get a candidate,
we will cache it in state (thats fine), but then we dispatch it and it may not be the latest.
If in the mean time we get another, isn't it extra communication in the background?

Or maybe add maintain X-seconds batching to the generator?

hanssv:
If you view the generator as "a worker" the ideas are quite similar I think. Of course you need to apply changes for NG. Either you introduce two generators or let one handle both key- and micro-block. In normal operation I guess you want to switch off micro-block generation since you only need those when you have solved the key-block. 

Without NG we always try to provide the best possible block; the blocks are not huge so the internal communication should be manageble. With NG we still want the best possible key block (including the latest micro-block seen); and now the key-blocks are really tiny so sharing an extra key-block should not be an issue. For the micro-blocks, since the timing is predictable, you could make something slightly more optimized than just "always push the best possible block".

apps/aecore/test/aec_tx_pool_tests.erl
```diff
@@ -79,7 +79,7 @@ tx_pool_test_() ->
                {GenesisBlock, _} = aec_block_genesis:genesis_block_with_state(),
                aec_test_utils:start_chain_db(),
                ok = aec_chain_state:insert_block(GenesisBlock),
-               {TopBlock, TopState} = aec_chain:top_block_with_state(),
+               {TopBlock, _TopState} = aec_chain:top_block_with_state(),
```
gorillainduction: Use `aec_chain:top_block/0` to not get the state since you don't need it anymore.

apps/aecore/test/aec_tx_pool_tests.erl
```diff
@@ -219,13 +217,12 @@ tx_pool_test_() ->
...
```
gorillainduction: Use `aec_chain:top_block/0` to not get the state since you don't need it anymore.

gorillainduction:
@hanssv @sennui To mitigate the message size, the candidate generator could just sent a simple event when it has found a new candidate, and the conductor could get it from the candidate generator when it pleases.

This will take care of unnecessary block copying, but it will also introduce a small time overhead when getting the candidate.

For NG, this would allow the conductor to set its timeout when the first candidate arrives, and then get the most current one when it decides to sign it.

hanssv:
Fixed the conflict...

hanssv:
And made dialyzer happy...

apps/aecore/src/aec_conductor.erl
```diff
@@ -193,10 +193,13 @@ handle_call({post_block, Block},_From, State) ->
 handle_call(stop_mining,_From, State) ->
     epoch_mining:info("Mining stopped"),
     State1 = kill_all_workers(State),
+    aec_block_generator:stop_generation(),
     {reply, ok, State1#state{mining_state = 'stopped',
+                             new_candidate_available = false,
```
gorillainduction: This is a race. The new candidate event might be in the message queue.

apps/aecore/src/aec_conductor.erl
```diff
@@ -217,6 +220,7 @@ handle_call(reinit_chain, _From, State1) ->
                 State3 = kill_all_workers(State2),
                 epoch_mining:info("Mining started"),
                 start_mining(State3#state{mining_state = running,
+                                          new_candidate_available = false,
```
gorillainduction: You need to stop the candidate generation here as well. And prevent the race.

gorillainduction: Or rather, kill the candidate geration to get a fresh start, and prevent the race.

apps/aecore/src/aec_conductor.erl
gorillainduction: Ok, this is fixed by ignoring candidates when not running the mining.

apps/aecore/src/aecore_sup.erl
```diff
@@ -60,6 +60,7 @@ init([]) ->
                                  ?CHILD(aec_metrics_rpt_dest, 5000, worker),
                                  ?CHILD(aec_keys, 5000, worker),
                                  ?CHILD(aec_tx_pool, 5000, worker),
+                                 ?CHILD(aec_block_generator, 5000, worker),
```
lucafavatella: Could you please update (or delete) representation of supervision tree in comment above in module?

apps/aecore/src/aec_conductor.erl
```diff
@@ -229,8 +234,32 @@ handle_cast(Other, State) ->
...
+            end;
+        #candidate{} ->
+            State1 = State#state{ new_candidate_available = true },
+            {noreply, State1}
+    end;
 handle_info(init_continue, State) ->
     %% Continue the initialization by (possibly) starting the miner
+    case State#state.mining_state of
+        running ->
+            aec_block_generator:start_generation();
```
lucafavatella: Assuming `autostart` option, process `aec_conductor` in `init_continue` calls `gen_server:cast(aec_block_generator, start_generation)` that returns `ok` even if `aec_block_generator` not registered then waits for event `candidate_block`. Considering that process `aec_block_candidate` initializes with `generating = false` and that `aecore_sup` uses `one_for_one` strategy, if process `aec_block_generator` terminated abnormally after `aec_conductor` had called `gen_server:cast(aec_block_generator, start_generation)` then process `aec_conductor` would wait forever (I do not see timeout in `aec_conductor` for `candidate_block` event though I did not thoroughly look for it).

lucafavatella: I would expect `aecore_sup` supervision strategy refined with intermediate `rest_for_one` with children - in the order - `aec_block_generator` and `aec_conductor`.

apps/aecore/src/aec_conductor.erl
```diff
@@ -582,101 +620,15 @@ handle_mining_reply({{error, {runtime, Reason}}, _}, State) ->
...
-create_block_candidate(State) ->
-    epoch_mining:info("Creating block candidate"),
-    Fun = fun() ->
-                  {TopBlock, TopBlockState} = aec_chain:top_block_with_state(),
-                  AdjChain = get_adjustment_headers(TopBlock),
-                  {aec_mining:create_block_candidate(TopBlock, TopBlockState,
-                                                     AdjChain),
-                   State#state.seen_top_block_hash}
-          end,
-    dispatch_worker(create_block_candidate, Fun, State).
```
lucafavatella: Documentation at top of module seems to assume that creation of block candidate is still worker - may benefit from rephrasing though optional.
https://github.com/aeternity/epoch/blob/76053d4e6149078f20f6a129ca119d4a62b0ecdc/apps/aecore/src/aec_conductor.erl#L10-L11

apps/aecore/src/aec_block_generator.erl
```diff
@@ -0,0 +1,272 @@
...
+    {noreply, State};
+handle_info({gproc_ps_event, Event, #{info := Info}}, State) ->
+    lager:debug("got event ~p", [Event]),
+    State1 =
+        case Event of
+            block_created -> preempt_generation(State, Info);
+            top_changed   -> preempt_generation(State, Info);
+            tx_created    -> add_new_tx(State, Info);
+            tx_received   -> add_new_tx(State, Info);
+            _             -> State
```
lucafavatella: As the process subscribes only for certain events, I expect this clause not to be exercised unless of typos. Maybe logging a warning here? (Totally optional.)

apps/aecore/src/aec_conductor.erl
```diff
@@ -417,8 +446,7 @@ preempt_if_new_top(#state{seen_top_block_hash = OldHash} = State, Publish) ->
             maybe_publish_top(Publish, NewHash),
             State1 = State#state{seen_top_block_hash = NewHash},
             State2 = kill_all_workers_with_tag(mining, State1),
-            State3 = kill_all_workers_with_tag(create_block_candidate, State2),
```
lucafavatella: How does `aec_block_generator` know that top changed when top changed because of block received from network? I could not find it. (I did not assume that distinct blocks necessarily have distinct set of txs, as that would seem a fragile assumption to me, so I did not explore that possibility for answering myself.)

hanssv: Since blocks received from another peer is still added to the chain through `aec_conductor` - `top_changed` means top changed due to block from network and `block_created` means the block was mined locally.

apps/aecore/src/aec_conductor.erl
hanssv: Great catch, I was worried about the opposite scenario when the conductor crashed and made sure that it worked. The opposite totally slipped my mind... I have refined the supervision tree (and updated the graph accordingly!))

apps/aecore/src/aec_conductor.erl
hanssv: Rephrased.

PR #1190, opened 2018-05-25
Remove docker compose keys volume

dincho:
https://www.pivotaltracker.com/story/show/157838332

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,4 @@ swagger/swagger-codegen-cli-*.jar
 *.retry
 system_test/logs
 system_test/aest_hard_fork_SUITE_data/db
+docker/keys/*/sign_key*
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -53,3 +53,4 @@ swagger/swagger-codegen-cli-*.jar
 *.retry
 system_test/logs
 system_test/aest_hard_fork_SUITE_data/db
+docker/keys/*/sign_key*
```
lucafavatella: I see this for not committing [these keys](https://github.com/aeternity/epoch/blob/093c2368419e771e76eb88e8361f8f8092c7606c/apps/aecore/src/aec_keys.erl#L46-L47) while keeping [the sync keys](https://github.com/aeternity/epoch/commit/4259366d7c33e8e9adf3d22638749671c7888f14).

PR #1199, opened 2018-05-28
SC fsm events as maps, separating ch id fr temp ch id

uwiger:
See https://www.pivotaltracker.com/story/show/157871516

The messages sent from the SC fsm to the WS handler is now on the form
`{aesc_fsm, FsmPid, #{type => Type, tag => Tag, info => Info}}`

The channel id is represented as `temp_channel_id => TmpChanId` initially. Then, as a persistent channel id exists `channel_id => ChanId` is added (the temp channel id is kept in the map).


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1728,13 +1727,31 @@ default_report_flags() ->
...
+            Client ! {?MODULE, self(), Msg};
+       true  ->
+            lager:debug("report_info(~p, ~p, ~p)", [DoRpt, Tag, Info]),
+            ok
     end,
     ok.
 
+rpt_message(Type, Tag, Info, #data{on_chain_id = undefined,
+                                   channel_id  = TmpChId}) ->
+    #{ temp_channel_id => TmpChId
```
velzevur: I see this is not used by client. Shall we addapt the WS to propagate it some how?

uwiger: Perhaps best to simply remove it if it won't break anything.

PR #1201, opened 2018-05-28
Fix mempool to delete channel transactions when included

velzevur:
[Pivotal story](https://www.pivotaltracker.com/story/show/157919113)
Txs are put into mempool's ETS using a complex key, including the `aetx:origin/1`. If a channel is closed, it will not be present in channels' tree so the origin is unobtainable and thus we can not construct the ETS key.

This PR includes a fourth element in the key, being the tx hash and a function to find a key using it. It is worth noting that the transaction can be not in the mempool at all.

### DIFF
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -15,7 +15,7 @@
 -include("common.hrl").
 
 -define(MEMPOOL, mempool).
--define(KEY_NONCE_PATTERN(Sender), {{'_', Sender, '$1'}, '_'}).
+-define(KEY_NONCE_PATTERN(Sender), {{'_', Sender, '$1', '_'}, '_'}).
 
 %% API
 -export([ start_link/0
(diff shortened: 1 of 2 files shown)


apps/aecore/src/aec_tx_pool.erl
```diff
@@ -206,7 +207,16 @@ pool_db_key(SignedTx) ->
...
     %%       * negative fee places high profit transactions at the beginning
     %%       * ordered_set type enables implicit overwrite of the same txs
-    {-aetx:fee(Tx), aetx:origin(Tx), aetx:nonce(Tx)}.
+    {-aetx:fee(Tx), aetx:origin(Tx), aetx:nonce(Tx), aetx_sign:hash(SignedTx)}.
+
+-spec select_pool_db_key_by_hash(pool_db(), binary()) -> {ok, pool_db_key()} | not_in_ets.
+select_pool_db_key_by_hash(Mempool, TxHash) ->
+    case sel_return(ets_select(Mempool, [{ {{'$1', '$2', '$3', '$4'}, '_'},
+                                           [{'=:=','$4', TxHash}],
+                                           [{{'$1', '$2', '$3', '$4'}}] }], infinity)) of
```
velzevur: The nested tuple is intended, from [docs](http://erlang.org/doc/man/ets.html)
```
If tuples are to be constructed, one has to write a tuple of arity 1 where the single element in the tuple is the tuple one wants to construct (as an ordinary tuple can be mistaken for a Guard).
```

PR #1202, opened 2018-05-28
Pt 157873700 serialization for proof of inclusion

gorillainduction:
https://www.pivotaltracker.com/n/projects/2124891/stories/157873700
https://github.com/aeternity/protocol/pull/106
NOTE: This PR also changes the way we calculate the root hash of state trees.

TODO:
- [x] Update protocol with the changes to serialization
- [ ] ~Update protocol with the changes to root hash calculation~ (This is not documented. Should be at some point)
- [ ] ~Implement the proof of inclusion top level for state channel trees~ (Leaving this for follow up)
- [x] Make a note in the release notes

### DIFF
--- a/apps/aecore/src/aec_object_serialization.erl
+++ b/apps/aecore/src/aec_object_serialization.erl
@@ -78,6 +78,7 @@ tag(channel_slash_tx) -> 55;
 tag(channel_settle_tx) -> 56;
 tag(channel_offchain_tx) -> 57;
 tag(channel) -> 58;
+tag(trees_poi) -> 60;
 tag(block) -> 100.
 
 rev_tag(10) -> account;
@@ -109,4 +110,5 @@ rev_tag(55) -> channel_slash_tx;
 rev_tag(56) -> channel_settle_tx;
 rev_tag(57) -> channel_offchain_tx;
 rev_tag(58) -> channel;
+rev_tag(60) -> trees_poi;
 rev_tag(100) -> block.
--- a/apps/aecore/test/aec_trees_tests.erl
+++ b/apps/aecore/test/aec_trees_tests.erl
@@ -81,3 +81,132 @@ make_spend_tx(Sender, Recipient) ->
                                        nonce => 1,
                                        payload => <<>>}),
     SpendTx.
+
+poi_test_() ->
+    [ {"POI for one account",
+       fun() ->
+               AccountPubkey = <<123:?MINER_PUB_BYTES/unit:8>>,
+
+               %% POI from an empty tree cannot be constructed.
+               Trees0 = aec_test_utils:create_state_tree(),
+               Poi0 = ?TEST_MODULE:new_poi(Trees0),
+               ?assertEqual(?TEST_MODULE:hash(Trees0),
+                            ?TEST_MODULE:poi_hash(Poi0)),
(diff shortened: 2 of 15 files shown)


apps/aecore/test/aec_trees_tests.erl
```diff
@@ -82,3 +82,132 @@ make_spend_tx(Sender, Recipient) ->
...
+                             ?TEST_MODULE:verify_poi(accounts, Pubkey3,
+                                                     Serialized3, Poi2)),
+                ?assertMatch(ok,
+                             ?TEST_MODULE:verify_poi(accounts, Pubkey3,
+                                                     Serialized3, Poi3)),
+
+                %% Check serialization/deserialization of PoI
+                %% NOTE: The deserialized poi contains a gb_tree,
+                %% so it is operation order dependent.
+                %% The serialized version is canonical, though.
```
lucafavatella: Thanks for the note.

apps/aecore/src/aec_object_serialization.erl
```diff
@@ -78,6 +78,7 @@ tag(channel_slash_tx) -> 55;
 tag(channel_settle_tx) -> 56;
 tag(channel_offchain_tx) -> 57;
 tag(channel) -> 58;
+tag(trees_poi) -> 60;
```
lucafavatella: I understand the proof of inclusion is not part of consensus. May I suggest marking it with a comment in this file (e.g. `tag(trees_poi) -> 60 %% Not part of consensus.`)?

This also means no need to amend [protocol](https://github.com/aeternity/protocol/blob/fb7059ee91617718ce8c8356c18685db13ca64d6/serializations.md#table-of-object-tags).

apps/aecore/src/aec_object_serialization.erl
```diff
@@ -78,6 +78,7 @@ tag(channel_slash_tx) -> 55;
...
 tag(block) -> 100.
```
lucafavatella: I understand the block is not part of consensus, so correctly not documented in [protocol](https://github.com/aeternity/protocol/blob/fb7059ee91617718ce8c8356c18685db13ca64d6/serializations.md#table-of-object-tags). May I suggest marking it with a comment in this file (e.g. `tag(block) -> 100 %% Not part of consensus.`)?

apps/aecore/src/aec_object_serialization.erl
velzevur: The proof of inclusion will be part of the channels' closing transactions so it will be used in the consesus. If a comment would be added - it will probably be short lived

gorillainduction:
Rebased to fix conflict, and squashed

apps/aecore/test/aec_trees_tests.erl
```diff
@@ -81,3 +81,132 @@ make_spend_tx(Sender, Recipient) ->
                                        nonce => 1,
                                        payload => <<>>}),
     SpendTx.
+
+poi_test_() ->
```
lucafavatella: It would be nice to have a `poi_serialization_test_` exemplifying the serialized PoI up to nibble level. This comes mostly from https://github.com/aeternity/protocol/pull/106#discussion_r191706391
Though totally optional for this PR.

PR #1206, opened 2018-05-29
Prevent mining on stale block candidate

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/157929767

TODO:
* [x] Add unit test if not too much effort
* [x] Check Python UAT tests and system tests pass
  * [CI workflow](https://circleci.com/workflow-run/7b9eb091-d8dd-438e-88f3-2457586cb4dc) on commit 3f828580 equivalent to 0eb5b447
    * UPDATE: Python UAT tests passed
    * UPDATE: System tests [~failed~ passed except for known issue](https://3837-110719633-gh.circle-artifacts.com/0/home/circleci/epoch/system_test/logs/ct_run.epoch_ct@localhost.2018-05-30_12.31.07/extras.system_test.logs/run.2018-05-30_12.31.07/suite.log.html) addressed in PR #1208
* [ ] Consider restructuring events UPDATE: Refactoring avoided as too involved. If needed for changing behaviour of any modules / subsystems it can be done in future PRs.
* [ ] Squash-and-merge

### DIFF
--- a/apps/aecore/src/aec_events.erl
+++ b/apps/aecore/src/aec_events.erl
@@ -16,9 +16,10 @@
 
 -include("blocks.hrl").
 
--type event() :: block_created
-               | start_mining
-               | top_changed
+-type event() :: start_mining
+               | block_created %% We created (mined) a new block. It usually becomes the new top block, though not necessarily.
+               | top_synced %% We received a block, pulled from a network peer, that became the new top.
+               | top_changed %% We received a block, pushed by a network peer, that became the new top.
                | tx_created
                | tx_received
                | candidate_block
--- a/apps/aecore/test/aec_conductor_tests.erl
+++ b/apps/aecore/test/aec_conductor_tests.erl
@@ -183,7 +183,8 @@ chain_test_() ->
      end,
      [
       {"Start mining add a block.", fun test_start_mining_add_block/0},
-      {"Test preemption of mining", fun test_preemption/0},
+      {"Test preemption of mining by block pushed by a network peer", fun test_preemption_pushed/0},
+      {"Test preemption of mining by block pulled from a network peer", fun test_preemption_pulled/0},
       {"Test chain genesis state" , fun test_chain_genesis_state/0},
       {timeout, 20, {"Test block publishing", fun test_block_publishing/0}}
      ]}.
@@ -201,7 +202,7 @@ test_start_mining_add_block() ->
       fun () -> aec_chain:top_header() end,
       BH2).
 
-test_preemption() ->
+test_preemption_pushed() ->
     %% Assert preconditions
     assert_stopped_and_genesis_at_top(),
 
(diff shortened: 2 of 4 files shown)


apps/aecore/test/aec_conductor_tests.erl
```diff
@@ -235,6 +236,39 @@ test_preemption() ->
...
+
+    wait_for_start_mining(Hash1),
+
+    %% Sync the rest of the chain, which will take over.
+    [?TEST_MODULE:add_synced_block(B) || B <- Chain2],
+    wait_for_top_block_hash(Hash2),
+
+    %% The mining should now have been preempted
+    %% and started over with the new top block hash
+    wait_for_start_mining(Hash2),
```
lucafavatella: Cherry-picking this commit on master (aa2d03e8) - i.e. code without fix in this PR - I got error as expected:
```
    aec_conductor_tests: chain_test_ (Test preemption of mining by block pulled from a network peer)...*failed*
in function aec_conductor_tests:wait_for_gproc/2 (/Users/luca/dev/aeternity/epoch/_build/test/lib/aecore/test/aec_conductor_tests.erl, line 474)
in call from aec_conductor_tests:wait_for_start_mining/1 (/Users/luca/dev/aeternity/epoch/_build/test/lib/aecore/test/aec_conductor_tests.erl, line 465)
in call from aec_conductor_tests:test_preemption_pulled/0 (/Users/luca/dev/aeternity/epoch/_build/test/lib/aecore/test/aec_conductor_tests.erl, line 268)
**error:{timeout,block_created}
  output:<<"">>
```

apps/aecore/src/aec_events.erl
```diff
@@ -16,9 +16,10 @@
 
 -include("blocks.hrl").
 
--type event() :: block_created
-               | start_mining
-               | top_changed
+-type event() :: start_mining
+               | block_created %% We created (mined) a new block. It usually becomes the new top block, though not necessarily.
+               | top_synced %% We received a block, pulled from a network peer, that became the new top.
+               | top_changed %% We received a block, pushed by a network peer, that became the new top.
```
lucafavatella: @uwiger @velzevur FYI In this WIP PR I documented the current semantics of `block_created` and `top_changed` internal events. I also added a new `top_synced` event for symptom of block candidate generation not being preempted when node pulling blocks from network (at bootstrap or node restart).

In the commit message I listed other modules that may benefit from this internal event, though so far I do not plan to change behaviour of those (e.g. on bootstrap lots of blocks get pulled from network so it would be a big difference).

lucafavatella:
> One idea I had when looking at this would be to refactor the events a bit. top_changed could be really that the top changed, and we could have one event that means "we should spread this block" (current top_changed + block_created) etc... But this will work!

@hanssv I considered that too though I feel it may be a bit too involved (I even see we have [tests checking on these events](https://github.com/aeternity/epoch/blob/aa2d03e887b36432a7ce5c1b1abd067d604764f7/apps/aecore/test/aec_conductor_tests.erl#L271-L313)). I suggest not to go with that refactoring in this PR.

lucafavatella:
@hanssv @uwiger @velzevur I consider this ready to be reviewed.

lucafavatella:
I am investigating why [system tests failed for this branch](https://circleci.com/gh/lucafavatella/epoch/3837).

apps/aecore/src/aec_events.erl
uwiger: This would imply that `aesc_fsm_min_depth_watcher.erl` should also be extended to monitor `top_synced`?
https://github.com/aeternity/epoch/blob/master/apps/aechannel/src/aesc_fsm_min_depth_watcher.erl#L36

lucafavatella: Opened as separate PR #1210

PR #1207, opened 2018-05-30
Pt 153897559 add transaction TTL

hanssv:


config/swagger.yaml
```diff
@@ -2368,6 +2387,9 @@ definitions:
       fee:
         type: integer
         format: int64
+      ttl:
+        type: integer
+        format: int64
```
lucafavatella: This may impact (broken) code generators from swagger spec, though out Python client is the only one I am aware of.

hanssv: I think the swagger file would benefit from a normalization, the "same" things are expressed in a multitude of ways today.

hanssv:
https://github.com/aeternity/protocol/pull/107

lucafavatella:
Will need rebase.

hanssv:
Rebased

lucafavatella:
Was in scope of https://www.pivotaltracker.com/story/show/153897559

config/swagger.yaml
```diff
@@ -2210,6 +2210,9 @@ definitions:
       fee:
         type: integer
         format: int64
+      ttl:
```
lucafavatella: Not added to required.

lucafavatella: Addressed in https://github.com/aeternity/epoch/pull/1217

config/swagger.yaml
```diff
@@ -3005,6 +3044,10 @@ definitions:
         description: Transaction fee
         type: integer
         minimum: 0
+      ttl:
```
lucafavatella: Missing in require as noticed by @hanssv

lucafavatella: Addressed in https://github.com/aeternity/epoch/pull/1217

PR #1208, opened 2018-05-30
#157969094 - Add missing case to aec_block_generator

hanssv:


lucafavatella:
I see in scope of https://www.pivotaltracker.com/story/show/157969094

PR #1212, opened 2018-05-31
Review block-related events to simplify channel testing

lucafavatella:
Delivers:
* https://www.pivotaltracker.com/story/show/157998007
* Also https://www.pivotaltracker.com/story/show/157960330

Please review also [corresponding protocol repo PR](https://github.com/aeternity/protocol/pull/108).

TODO:
* [x] Adapt tests.
* [x] Check Python UAT tests pass.
  * [CI job](https://circleci.com/gh/lucafavatella/epoch/3929) on 687699de that is equivalent to 6cfe3755
* [x] Check system tests pass.
  * [CI job](https://circleci.com/gh/lucafavatella/epoch/3923) on 687699de that is equivalent to 6cfe3755
* [ ] Squash-and-merge

### DIFF
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -450,6 +450,7 @@ preempt_if_new_top(#state{seen_top_block_hash = OldHash} = State, Publish) ->
         OldHash -> no_change;
         NewHash ->
             ok = aec_tx_pool:top_change(OldHash, NewHash),
+            aec_events:publish(top_changed, NewHash),
             maybe_publish_top(Publish, NewHash),
             {ok, NewBlock} = aec_chain:get_block(NewHash),
             aec_metrics:try_update([ae,epoch,aecore,chain,height],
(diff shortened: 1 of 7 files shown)


apps/aecore/src/aec_conductor.erl
```diff
@@ -481,8 +481,10 @@ maybe_publish_block(block_received,_Block) ->
     %% changes the top.
     ok;
 maybe_publish_block(block_created = T, Block) ->
+    aec_events:publish(T, Block),
     %% This is a block we created ourselves. Always publish.
-    aec_events:publish(T, Block).
+    {ok, BlockHash} = aec_blocks:hash_internal_representation(Block),
```
hanssv: Since we are only using block_to_publish in one place and there we use the full block it might be better to just include the block in the message?

lucafavatella: ok

lucafavatella: Addressed.

PR #1159, opened 2018-05-17
change dockerignore to whitelist

acidsennui:
pt: https://www.pivotaltracker.com/story/show/157624989

tested, syncing with UAT after docker build.

PR #1215, opened 2018-06-05
Compile eunit/CT parse transform only when needed

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/157624989

Triggered by https://github.com/aeternity/epoch/pull/1159/files#r192683104

TODO:
* [ ] Checking locally EUNIT_INCLUDED


lucafavatella:
I did not manage to make `make test` work with and without `EUNIT_INCLUDED` so closing this.

PR #1216, opened 2018-06-06
Add /api as endpoint for retrieving yaml spec

ThomasArts:
This new endpoint at /api provides the yaml spec in json format. It is not part of the yaml file itself, because then it would take the configured base path from there /v2/api. Since the base path may change (in version update or so), this endpoint is not related to that path.


lucafavatella:
I see delivers https://www.pivotaltracker.com/story/show/154725991

PR #1218, opened 2018-06-06
Separate system test logs per test case

eproxus:
Duplicate node names in tests should no longer merge the log files. Creates a structure like this:

```
.
├── extras.system_test.aest_commands_SUITE.logs
│   └── run.2018-06-06_13.03.38
│       ├── log_private
│       ├── log_private.1
│       │   ├── config
│       │   └── node1_logs
│       ├── log_private.cfg.3
│       └── log_private.cfg.4
└── extras.system_test.aest_sync_SUITE.logs
    └── run.2018-06-06_12.59.08
        ├── log_private
        ├── log_private.1
        │   ├── config
        │   ├── new_node1_logs
        │   ├── old_node1_logs
        │   └── old_node2_logs
        ├── log_private.2
        │   ├── config
        │   └── standalone_node_logs
        ├── log_private.3
        │   ├── config
        │   ├── node1_logs
        │   └── node2_logs
        ├── log_private.4
        │   ├── config
        │   ├── net1_node1_logs
        │   ├── net1_node2_logs
        │   ├── net2_node1_logs
        │   └── net2_node2_logs
        ├── log_private.5
        │   ├── config
        │   ├── n1_logs
        │   └── n2_logs
        ├── log_private.cfg.1
        └── log_private.cfg.2

```


eproxus:
I removed the old config file because it was only used by the now deleted hard fork suite.

rebar.config
```diff
@@ -136,7 +136,8 @@
                     bbmustache,
                     {hackney, "1.12.1"},
                     {websocket_client, ".*", {git, "git://github.com/aeternity/websocket_client", {ref, "a4fb3db"}}}
-                ]}
+                ]},
+                {ct_opts, [{create_priv_dir, auto_per_tc}]}
```
lucafavatella: I see http://www.rebar3.org/docs/configuration#section-common-test and http://erlang.org/doc/man/ct.html#run_test-1

Makefile
```diff
@@ -2,6 +2,7 @@ CORE = rel/epoch/bin/epoch
 VER := $(shell cat VERSION)
 
 EUNIT_VM_ARGS = $(CURDIR)/config/eunit.vm.args
+CT_FLAGS = --dir system_test --logdir system_test/logs
```
lucafavatella: A more specific var name as this is system test specific?

rebar.config
eproxus: What do you mean?

lucafavatella: I mean that I saw that the change makes sense.

Actually a better reference is http://erlang.org/doc/apps/common_test/write_test_chapter.html#id78698

PR #1223, opened 2018-06-07
Make transaction TTL optional

hanssv:
Internally represent no-ttl/infinity with 0 to keep serialisation short.

### DIFF
--- a/apps/aecore/src/aec_spend_tx.erl
+++ b/apps/aecore/src/aec_spend_tx.erl
@@ -36,7 +36,7 @@
           recipient = <<>>          :: aec_keys:pubkey(),
           amount    = 0             :: non_neg_integer(),
           fee       = 0             :: non_neg_integer(),
-          ttl       = 0             :: aec_blocks:height(),
+          ttl       = 0             :: aetx:tx_ttl(),
           nonce     = 0             :: non_neg_integer(),
           payload   = <<>>          :: binary()}).
 
@@ -49,21 +49,18 @@ new(#{sender := SenderPubkey,
       recipient := RecipientPubkey,
       amount := Amount,
       fee := Fee,
-      ttl := TTL,
       nonce := Nonce,
-      payload := Payload}) when is_integer(Amount), Amount >= 0,
-                                is_integer(Nonce), Nonce >= 0,
-                                is_integer(Fee), Fee >= 0,
-                                is_integer(TTL), TTL >= 0,
-                                is_binary(SenderPubkey),
-                                is_binary(RecipientPubkey),
-                                is_binary(Payload)
-                                ->
+      payload := Payload} = Args) when is_integer(Amount), Amount >= 0,
+                                       is_integer(Nonce), Nonce >= 0,
+                                       is_integer(Fee), Fee >= 0,
+                                       is_binary(SenderPubkey),
+                                       is_binary(RecipientPubkey),
+                                       is_binary(Payload) ->
     Tx = #spend_tx{sender = SenderPubkey,
                    recipient = RecipientPubkey,
                    amount = Amount,
                    fee = Fee,
-                   ttl = TTL,
+                   ttl = maps:get(ttl, Args, 0),
                    nonce = Nonce,
                    payload = Payload},
     {ok, aetx:new(?MODULE, Tx)}.
@@ -76,7 +73,7 @@ type() ->
 fee(#spend_tx{fee = F}) ->
     F.
 
--spec ttl(tx()) -> aec_blocks:height().
+-spec ttl(tx()) -> aetx:tx_ttl().
 ttl(#spend_tx{ttl = TTL}) ->
     TTL.
 
--- a/apps/aehttp/src/aehttp_dispatch_ext.erl
+++ b/apps/aehttp/src/aehttp_dispatch_ext.erl
@@ -158,8 +158,8 @@ handle_request('PostTx', #{'Tx' := Tx} = Req, _Context) ->
 handle_request('PostContractCreate', #{'ContractCreateData' := Req}, _Context) ->
     ParseFuns = [parse_map_to_atom_keys(),
                  read_required_params([owner, code, vm_version, deposit,
-                                       amount, gas, gas_price, fee, ttl,
-                                       call_data]),
+                                       amount, gas, gas_price, fee, call_data]),
+                 read_optional_params([{ttl, ttl, '$no_value'}]),
                  base58_decode([{owner, owner, account_pubkey}]),
                  get_nonce(owner),
                  hexstrings_decode([code, call_data]),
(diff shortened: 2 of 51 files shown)


hanssv:
edoc failed :-(

apps/aetx/src/aetx.erl
```diff
@@ -28,6 +28,8 @@
 -export([tx/1]).
 -endif.
 
+-define(MAXINT, 16#FFFFffffFFFFffff).
```
dincho: Interesting formatting hack ❤️

hanssv: Not invented here, saw it elsewhere and thought it improved readability!

hanssv:
Added a note in the Release notes

docs/release-notes/RELEASE-NOTES-0.16.0.md
```diff
@@ -2,6 +2,7 @@
 
 [This release][this-release] is focused on TODOFILLMEIN.
 It:
+* Transaction TTL is now optional (giving no TTL means the transaction is valid "forever").
```
lucafavatella: "Makes the transaction TTL optional ..." would have been more readable but ok.

apps/aetx/src/aetx.erl
lucafavatella: It looks this MAXINT is the internal representation of ttl meant to never expire.
* Why not using an atom? (I do not see why this must be int.)
* How is that this is "the max"? How is range defined? It looks to me that [serialized tx may have larger TTL as unbound](https://github.com/aeternity/protocol/blob/d89aea62813f9a39d946f440b2c640838799afdf/serializations.md#dynamic-size-object-serialization).

apps/aetx/src/aetx.erl
```diff
@@ -168,11 +178,21 @@ accounts(#aetx{ cb = CB, tx = Tx }) ->
 signers(#aetx{ cb = CB, tx = Tx }, Trees) ->
     CB:signers(Tx, Trees).
 
+%% Internally we let 0 represent no TTL/infinity in order to keep the
+%% serialization as short as possible. Since there are no transactions in the
```
lucafavatella: Do you mean serialization for DB here? (as previous line says "internally" rather than "externally", that I would expect if this sentence refers to serialization for consensus)

apps/aecore/src/aec_spend_tx.erl
```diff
@@ -49,21 +49,18 @@ new(#{sender := SenderPubkey,
...
+                                       is_integer(Fee), Fee >= 0,
+                                       is_binary(SenderPubkey),
+                                       is_binary(RecipientPubkey),
+                                       is_binary(Payload) ->
     Tx = #spend_tx{sender = SenderPubkey,
                    recipient = RecipientPubkey,
                    amount = Amount,
                    fee = Fee,
-                   ttl = TTL,
+                   ttl = maps:get(ttl, Args, 0),
```
lucafavatella: This is removing the check that value must be integer. Though :new shall be called on already sanitized data so shall be ok.

apps/aetx/src/aetx.erl
hanssv: This is the max block height we can have but if one wants to put an even larger number in there it is certainly possible...

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -312,7 +321,9 @@ handle_request('PostSpend', #{'SpendTx' := Req}, _Context) ->
     ParseFuns = [parse_map_to_atom_keys(),
                  read_required_params([sender,
                                        {recipient_pubkey, recipient},
-                                       amount, fee, ttl, payload]),
+                                        amount, fee, payload]),
+                 read_optional_params([{ttl, ttl, '$no_value'}]),
+                 read_optional_params([{ttl, ttl, '$no_value'}]),
```
lucafavatella: Twice?

hanssv: No, rebasing gone wrong. But no harm in doing it twice ;-)

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -419,7 +419,7 @@ read_channel_options(Params) ->
          Read(<<"initiator_amount">>, initiator_amount, #{type => integer}),
          Read(<<"responder_amount">>, responder_amount, #{type => integer}),
          Read(<<"channel_reserve">>, channel_reserve, #{type => integer}),
-         Read(<<"ttl">>, ttl, #{type => integer}),
+         Read(<<"ttl">>, ttl, #{type => integer, mandatory => false}),
```
lucafavatella: Update needed at https://github.com/aeternity/protocol/blob/d89aea62813f9a39d946f440b2c640838799afdf/epoch/api/channels_api_usage.md#channel-parameters ?

apps/aehttp/src/aehttp_helpers.erl
```diff
@@ -330,6 +330,7 @@ ok_response(Fun) ->
 unsigned_tx_response(NewTxFun) when is_function(NewTxFun, 1) ->
     RespFun =
         fun(State) ->
+            lager:debug("ZZZ ~p", [State]),
```
lucafavatella: ?

apps/aehttp/src/aehttp_dispatch_ext.erl
hanssv: Will fix of course

apps/aehttp/src/aehttp_helpers.erl
hanssv: Move along, nothing to see ;-)

apps/aetx/src/aetx.erl
```diff
@@ -77,14 +77,22 @@
...
-%%      smart contract. In the latter case the fee logic is different.
 -type tx_context() :: aetx_transaction | aetx_contract.
+%% Where does this transaction come from? Is it a top level transaction
+%% or was it created by a smart contract. In the latter case the fee
+%% logic is different.
+
+-type tx_ttl() :: 0 | aec_blocks:height().
+%% A transaction TTL is either an absolute block height, or the
+%% transaction does not have a TTL. The latter is represented as 0
+%% internally to get a small serialization. `aetx:ttl/1' returns MAXINT
```
lucafavatella: MAXINT obsolete?

hanssv: yes, and another internally :-(

apps/aetx/src/aetx.erl
```diff
@@ -168,11 +175,21 @@ accounts(#aetx{ cb = CB, tx = Tx }) ->
...
+        N -> N
+    end.
+
 -spec check(Tx :: tx(), Trees :: aec_trees:trees(), Height :: non_neg_integer(),
             ConsensusVersion :: non_neg_integer()) ->
     {ok, NewTrees :: aec_trees:trees()} | {error, Reason :: term()}.
-check(#aetx{ cb = CB, tx = Tx }, Trees, Height, ConsensusVersion) ->
-    case {CB:fee(Tx) >= aec_governance:minimum_tx_fee(), CB:ttl(Tx) >= Height} of
+check(AeTx = #aetx{ cb = CB, tx = Tx }, Trees, Height, ConsensusVersion) ->
+    case {CB:fee(Tx) >= aec_governance:minimum_tx_fee(), ttl(AeTx) >= Height} of
```
lucafavatella: `max_ttl > 1234` ?

hanssv:
```
1> max_ttl > 1243.
true
```

lucafavatella: From http://erlang.org/doc/reference_manual/expressions.html#id81948
> number < atom < reference < fun < port < pid < tuple < map < nil < list < bit string

So `max_ttl` will always be greater than a block height. Though may impact readability.

PR #1224, opened 2018-06-07
Pt 158175735 remove account index

gorillainduction:
As discussed at the Stockholm meetup:

https://www.pivotaltracker.com/n/projects/2124891/stories/158175735

* Removes the `/account/{account_pubkey}/txs` endpoint
* Removes the (broken) index that kept track of txs that affected an account
* Removes the `accounts/1` callback from the `aetx` behavior.

(WIP since it is not to be included in the current release, but it is really done)


gorillainduction:
Rebased on master and added release notes

apps/aecore/src/aec_db.erl
```diff
@@ -581,7 +546,7 @@ add_backend_plugins(_) ->
     ok.
 
 add_index_plugins() ->
-    mnesia_schema:add_index_plugin({acct2tx}, aec_db, ix_acct2tx).
```
lucafavatella: Shall otp_patches and any reference to it (e.g. rebar.config, `-pa patches/ebin` in vm.args files) be contextually removed?

gorillainduction: We can leave it for a while before we determine that we don't need it. There is infrastructure in state channels to use them (although it is not currently used).

kryztoval:
If you would be so kind and check this https://github.com/aeternity/epoch/issues/1729 I would really appreciate it.

Thank you.

PR #1226, opened 2018-06-07
Maps as list of tuple example contract.

happi:


happi:
@lucafavatella , At the moment it is not hooked in it is just an example so that the AEpps team can see how Oracles are used. @rvirding is working on a compiler test suite that will run all example contracts trough the compiler and also execute them on the chain.

PR #1230, opened 2018-06-08
[RFC] Give gas fees to miner (no mempool changes yet)

lucafavatella:
This is the initial implementation of proposed solution for "Reward miner with gas used (times gas price) after having processed contract-related transaction" as described at https://www.pivotaltracker.com/story/show/154664693/comments/190227856 ; it does not implement solution for "Enhance mempool to consider reward miner might get by processing contract-related transaction" yet.

I am opening this PR for requesting comments while I address the remaining TODO items mentioned in the commit messages in this PR. Hence this PR is not meant to be merged.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -575,9 +575,6 @@ workflows:
       - uat_tests:
           requires:
             - build
-          filters:
-            branches:
-              only: master
 
       - static_analysis:
           requires:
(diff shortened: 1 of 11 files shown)


gorillainduction:
I think the approach is valid.

lucafavatella:
Thanks for the comments - I will increase test coverage, rebase, bump consensus, add release notes line, then present as new PR.

PR #1233, opened 2018-06-11
Increase unit test coverage of contract create and call txs

lucafavatella:
Preparatory PR for https://www.pivotaltracker.com/story/show/154664693 that introduced gas-related reward to miner.

Please also review https://github.com/aeternity/protocol/pull/116

### DIFF
--- a/apps/aecontract/test/aecontract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_SUITE.erl
@@ -29,6 +29,7 @@
         ]).
 
 -include_lib("common_test/include/ct.hrl").
+-include_lib("stdlib/include/assert.hrl").
 
 -include_lib("apps/aecore/include/blocks.hrl").
 -include_lib("apps/aecontract/include/contract_txs.hrl").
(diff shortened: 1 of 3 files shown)


apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -123,8 +124,47 @@ create_contract(_Cfg) ->
...
+    ?assertMatch([_], aect_call_state_tree:to_list(aect_test_utils:calls(S2))),
+    InitCallId = aect_call:id(PubKey, aetx:nonce(Tx), ContractKey),
+    {value, InitCall} = aect_call_state_tree:lookup_call(ContractKey, InitCallId, aect_test_utils:calls(S2)),
+    %% Check that the created init call has the correct details from the contract create tx
+    ?assertEqual(PubKey, aect_call:caller_address(InitCall)),
+    ?assertEqual(aetx:nonce(Tx), aect_call:caller_nonce(InitCall)),
+    %% Check that the created init call has the correct details not from the contract create tx
+    ?assertEqual(ContractKey, aect_call:contract_address(InitCall)),
+    _ = aect_call:height(InitCall), %% Unclear if this needed.
+    _ = aect_call:gas_used(InitCall), %% No init in contract - shall value be zero?
```
lucafavatella: Shall value be zero? (I checked and it is positive.)

lucafavatella: @happi Does this relate to ["Make sure init is executed at contract create"](https://www.pivotaltracker.com/story/show/158109766) ?

happi: Yes. In the Sophia ABI an init function is required in the contract and create must call the init function.

UlfNorell: If no init is defined the compiler inserts a stub, so the gas cost will be positive, but small.

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -123,8 +124,47 @@ create_contract(_Cfg) ->
...
+    ?assertEqual(ok, aect_call:return_type(InitCall)),
+    _ = aect_call:return_value(InitCall), %% No init in contract - what shall value be?
```
UlfNorell: The value is the unit value, which the compiler represents as a 0 word.

PR #1234, opened 2018-06-12
Revert "change dockerignore to whitelist"

dincho:
Reverts aeternity/epoch#1159

Since the build process generates artefacts all over the source tree (not good IMO), it will be quite hard to figure our the correct whitelist+blacklist mix of .dockerignore. See [.gitignore](https://github.com/aeternity/epoch/blob/revert-1159-PT157624989/.gitignore)


dincho:
Alternative PR https://github.com/aeternity/epoch/pull/1235

dincho:
closing in favour of #1235

PR #1237, opened 2018-06-12
Pt 158071477 tagged identifiers

gorillainduction:
This PR introduces the typed identifiers in `aec_id` which is a good place to start for the review.

This PR is aimed to introduce the tagged identifiers in:
* The Erlang objects to get some type checking.
* The serialized format to get all the protocol changes needed (and to get tagged identifiers of course)

This PR is not hooking in the identifiers outside of the objects themselves. Since the serialization types allows for any type of identifier, we are future proof when other types of identifiers (e.g., names) starts to appear in transactions (e.g., names in spend tx).

We should introduce the typed identifiers in other code as well, so that we can track the types of identiers from the API (base58 decoding should keep the tags) and from the contracts (based on Sophia types). This is out of scope for this ticket.

PT: https://www.pivotaltracker.com/n/projects/2124891/stories/158071477
Protocol PR: https://github.com/aeternity/protocol/pull/117

TODO:

- [x] Bump protocol if needed.
- [x] Bump deploy directory.


lucafavatella:
> Bump protocol if needed.

I see consensus protocol apps/aecore/include/blocks.hrl not bumped since v0.15.0 so I understand needed.

apps/aecore/src/aec_id.erl
```diff
@@ -0,0 +1,104 @@
...
+%%%===================================================================
+
+-spec create(tag(), val()) -> id().
+create(Tag, Val) when ?IS_TAG(Tag), ?IS_VAL(Val) ->
+    #id{ tag = Tag
+       , val = Val};
+create(Tag, Val) when ?IS_VAL(Val) ->
+    error({illegal_tag, Tag});
+create(Tag, Val) when ?IS_TAG(Tag)->
+    error({illegal_val, Val}).
```
hanssv: What, if it is neither a tag nor a value?

apps/aecore/src/aec_accounts.erl
```diff
@@ -33,11 +33,12 @@
 
 -spec new(aec_keys:pubkey(), non_neg_integer()) -> account().
```
lucafavatella: So this PR does not impact accounts genesis (accounts.json). Ok.

apps/aecore/src/aec_id.erl
gorillainduction: You get a lovely `function_clause` ;-)

gorillainduction:
Rerunning tests on spurious error:
```
%%% aeu_requests_SUITE ==> {errors_in_lager_log,["13:25:17.209 [error] Supervisor {<0.32686.0>,poolboy_sup} had child ws_task_worker started with ws_task_worker:start_link([ws_task_worker_args]) at undefined exit with reason killed in context shutdown_error\n\n"]}
```

PR #1239, opened 2018-06-12
Check that create contract for Sophia ABI actually calls the init fun…

happi:
…ction.


apps/aecontract/src/aect_sophia.erl
```diff
@@ -138,7 +137,7 @@ on_chain_call(ContractKey, Function, Argument) ->
                     , gas => 100000000000000000
                     , gasPrice => 1
                     , origin => 0
-                    , value => 0
+                    , value => Amount
```
lucafavatella: Not sure what is impact of this.

happi: None, Amount is 0 defined a few lines above but not used.

lucafavatella: 🤦🏻‍♂️

PR #1242, opened 2018-06-13
Enhance mempool to consider gas fees to miner

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/154664693

Follows from #1232 

TODO:
* [x] Check minimum gas price before adding tx to mempool
* [x] Enhance mempool to consider reward miner might get by processing contract-related transaction
  * [x] Tests
  * [x] Code
* [x] Ensure Python UAT tests pass
  * [CI job passed](https://circleci.com/gh/lucafavatella/epoch/4178) on f2a46806 equivalent to 7e6a954f
* [x] Ensure system tests pass
  *  [CI job passed](https://circleci.com/gh/lucafavatella/epoch/4174) on f2a46806 equivalent to 7e6a954f 
  * (System tests had [passed](https://circleci.com/gh/aeternity/epoch/16911) on master 5a138412 )

### DIFF
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -13,7 +13,13 @@
 -behaviour(gen_server).
 
 -define(MEMPOOL, mempool).
--define(KEY_NONCE_PATTERN(Sender), {{'_', Sender, '$1', '_'}, '_'}).
+-define(KEY(NegFee, Origin, Nonce, TxHash), {NegFee, Origin, Nonce, TxHash}).
+-define(VALUE_POS, 2).
+-define(KEY_AS_MATCH_SPEC_RESULT(NegFee, Origin, Nonce, TxHash),
+        { %% Tuple of arity 1 where the single element is the mempool key tuple.
+         ?KEY(NegFee, Origin, Nonce, TxHash)
+        }).
+-define(KEY_NONCE_PATTERN(Sender), {?KEY('_', Sender, '$1', '_'), '_'}).
 
 %% API
 -export([ start_link/0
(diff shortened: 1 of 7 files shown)


apps/aecore/src/aec_tx_pool.erl
```diff
@@ -357,3 +371,21 @@ check_minimum_fee(Tx,_Hash) ->
...
+int_check_minimum_gas_price(GasPrice) ->
+    case GasPrice >= aec_governance:minimum_gas_price() of
+        true  -> ok;
+        false -> {error, too_low_gas_price}
+    end.
+
+int_gas_price(Tx) ->
+    case aetx:lookup_gas_price(Tx) of
+        none -> ?PSEUDO_GAS_PRICE;
+        {value, GP} when is_integer(GP), GP >= 0 -> GP
```
hanssv: This should only be called for proper Txs so this check is overly cautious?

deployment/ansible/deploy.yml
```diff
@@ -56,7 +56,7 @@
           port: 3114
       chain:
         persist: true
-        db_path: "./db64"
+        db_path: "./db65"
```
hanssv: Is the mempool sorting part of consensus? But better bump one time too many!

lucafavatella: Mmm... I was thinking the mempool key. Though that table is not persisted so I could have avoided.

One more no issue though.

apps/aecore/src/aec_tx_pool.erl
lucafavatella: Correct.

PR #1243, opened 2018-06-13
Blue/green deployments

dincho:
Example deployment to *dev1* https://circleci.com/workflow-run/da29f7ae-63bb-43be-b817-f12fb4f7687f

Blue: n0
Green: n1

Finishes https://www.pivotaltracker.com/story/show/157217781

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -485,14 +485,14 @@ jobs:
   deploy_uat_blue:
     <<: *container_config
     environment:
-      - DEPLOY_ENV: dev1
+      - DEPLOY_ENV: uat
       - DEPLOY_COLOR: blue
     steps: *deploy_steps
 
   deploy_uat_green:
     <<: *container_config
     environment:
-      - DEPLOY_ENV: dev1
+      - DEPLOY_ENV: uat
       - DEPLOY_COLOR: green
     steps: *deploy_steps
 
@@ -657,7 +657,7 @@ workflows:
           type: approval
           filters:
             branches:
-              only: 157217781_blue_green_deployments
+              ignore: /.*/
             tags:
               only: /^v.*$/
 
@@ -665,7 +665,7 @@ workflows:
           type: approval
           filters:
             branches:
-              only: 157217781_blue_green_deployments
+              ignore: /.*/
             tags:
               only: /^v.*$/
 
@@ -677,9 +677,7 @@ workflows:
             - hodl_blue
           filters:
             branches:
-              only:
-                - env/uat_blue
-                - 157217781_blue_green_deployments
+              only: env/uat_blue
             tags:
               only: /^v.*$/
 
@@ -689,9 +687,7 @@ workflows:
             - hodl_green
           filters:
             branches:
-              only:
-                - env/uat_green
-                - 157217781_blue_green_deployments
+              only: env/uat_green
             tags:
               only: /^v.*$/
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -496,20 +496,6 @@ jobs:
       - DEPLOY_COLOR: green
     steps: *deploy_steps
 
-  manual_deploy_uat_blue:
-    <<: *container_config
-    environment:
-      - DEPLOY_ENV: dev1
-      - DEPLOY_COLOR: blue
-    steps: *deploy_steps
-
-  manual_deploy_uat_green:
-    <<: *container_config
-    environment:
-      - DEPLOY_ENV: dev1
-      - DEPLOY_COLOR: green
-    steps: *deploy_steps
-
   deploy_fast_integration:
     <<: *container_config
     environment:
@@ -554,8 +540,6 @@ workflows:
               ignore:
                 - env/dev1
                 - env/dev2
-                - env/uat_green
-                - env/uat_blue
                 - master
 
       - deploy_api_docs:
@@ -704,24 +688,6 @@ workflows:
               ignore: /.*/
             tags:
               only: /^v.*$/
-
-      - manual_deploy_uat_blue:
-          requires:
-            - eunit
-            - static_analysis
-            - linux_package
-          filters:
-            branches:
-              only: env/uat_blue
-
-      - manual_deploy_uat_green:
-          requires:
-            - eunit
-            - static_analysis
-            - linux_package
-          filters:
-            branches:
-              only: env/uat_green
 
       - github_release_linux:
           requires:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,9 +153,13 @@ references:
         environment:
           PACKAGES_DIR: *packages_workspace
         command: |
-          export PACKAGE=${PACKAGES_DIR:?}/epoch-$(cat VERSION)-ubuntu-x86_64.tar.gz
+          PACKAGE=${PACKAGES_DIR:?}/epoch-$(cat VERSION)-ubuntu-x86_64.tar.gz
+          LIMIT="tag_role_epoch:&tag_env_${DEPLOY_ENV:?}"
+          if [ -n "$DEPLOY_COLOR" ]; then
+            LIMIT+=":&tag_color_${DEPLOY_COLOR}"
+          fi
           cd deployment/ansible
-          ansible-playbook --limit="tag_role_epoch:&tag_env_${DEPLOY_ENV:?}" \
+          ansible-playbook --limit="${LIMIT:?}" \
             --extra-vars "package=${PACKAGE:?} hosts_group=tag_env_${DEPLOY_ENV:?} env=${DEPLOY_ENV:?}" \
             deploy.yml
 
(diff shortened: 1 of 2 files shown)


lucafavatella:
For reference - I read in https://www.pivotaltracker.com/story/show/157217781
> So the new tag workflow would be tag -> build -> hold -> deploy blue -> hold -> deploy green

docs/operation.md
```diff
@@ -64,7 +64,7 @@ Verify that your node sees the same longest blockchain as the testnet.
 
 Inspect the current top of the blockchain as seen by the testnet:
 ```bash
-curl http://31.13.249.70:3013/v2/top
```
lucafavatella: Is `||` mean to be or on exit status?
* Not sure about exit status of curl (e.g. no `-f`) in all error conditions
* User may copy paste and not understand what is going on

May I suggest replacing `||` with `##` i.e. comment? Though change optional.

.circleci/config.yml
```diff
@@ -639,23 +653,41 @@ workflows:
...
           filters:
             branches:
-              ignore: /.*/
+              only: env/uat_blue
+            tags:
+              only: /^v.*$/
+
+      - deploy_uat_green:
+          requires:
+            - deploy_uat_blue
```
lucafavatella: So blue is deployed first, then green.

.circleci/config.yml
```diff
@@ -687,7 +719,7 @@ workflows:
           requires:
             - eunit
             - static_analysis
-            - hodl
+            - hodl_blue
```
lucafavatella: As blue is deployed before green, this means that docker image (versioned and `latest`) pushed before green deployed. Ok. (Just confirming.)

.circleci/config.yml
```diff
@@ -676,7 +708,7 @@ workflows:
             - eunit
             - static_analysis
             - osx_package
-            - hodl
+            - hodl_blue
```
lucafavatella: As blue is deployed before green, this means that macOS binary uploaded before green deployed. Ok. (Just confirming.)

.circleci/config.yml
```diff
@@ -664,7 +696,7 @@ workflows:
             - eunit
             - static_analysis
             - linux_package
-            - hodl
+            - hodl_blue
```
lucafavatella: As blue is deployed before green, this means that Ubuntu binary uploaded before green deployed. Ok. (Just confirming.)

.circleci/config.yml
```diff
@@ -639,23 +653,41 @@ workflows:
...
+            tags:
+              only: /^v.*$/
+
+      - deploy_uat_green:
+          requires:
+            - deploy_uat_blue
+            - hodl_green
+          filters:
+            branches:
+              only: env/uat_green
```
lucafavatella: How can this (I mean triggering job `deploy_uat_green` by pushing branch `env/uat_green`) work, considering that job `deploy_uat_blue` - that this job `deploy_uat_green` depends on - is not triggered on push to branch `env/uat_green`?

on which job deploy_uat_blue depends on - is not triggered?

.circleci/config.yml
```diff
@@ -639,23 +653,41 @@ workflows:
...
           requires:
             - eunit
             - static_analysis
             - linux_package
-            - hodl
+            - hodl_blue
           filters:
             branches:
-              ignore: /.*/
+              only: env/uat_blue
```
lucafavatella: How can this work - considering that required job `hodl_blue` is not triggered on push to branch `env/uat_blue`?

.circleci/config.yml
```diff
@@ -639,23 +653,41 @@ workflows:
...
+            tags:
+              only: /^v.*$/
+
+      - deploy_uat_green:
+          requires:
+            - deploy_uat_blue
+            - hodl_green
```
lucafavatella: Let us say that, on `v.*` tag:
* Job `deploy_uat_blue` succeeds;
* `hodl_green` is not complete (clicked...) yet;
* Validating blue deployment, validating pushed Docker image, validating uploaded macOS package, take some time (so `hodl_green` not clicked yet).

Does `hodl_green` and the rest of the workflow not time out in - let us say - 3 days? (Checking what the workflow is meant to be.)

----

A related question is whether such a partially-completed deployment shall be completed by pushing branch `env/uat_green` (and whether that works - see comment below).

docs/operation.md
dincho: > Is || mean to be or on exit status?

yes

I see that `curl` have sophisticated error/exit codes https://linux.die.net/man/1/curl
This makes me thing that it exit with 0 really only on success, I think it will cover 99% is the cases, while we're looking really for simple cases where the IP is available but not HTTP listener on it, or some HTTP error code during node boot phase.


> May I suggest replacing || with ## i.e. comment? Though change optional.

I could, but I'm not sure what's better. As you wrote users are copy paste and go on, I guess most of the users don't care (and don't understand) what's going on as long as it works.

I'm concerned that additional comment with additional command will confuse them more ? 🤔

.circleci/config.yml
```diff
@@ -153,9 +153,13 @@ references:
         environment:
           PACKAGES_DIR: *packages_workspace
         command: |
-          export PACKAGE=${PACKAGES_DIR:?}/epoch-$(cat VERSION)-ubuntu-x86_64.tar.gz
+          PACKAGE=${PACKAGES_DIR:?}/epoch-$(cat VERSION)-ubuntu-x86_64.tar.gz
+          LIMIT="tag_role_epoch:&tag_env_${DEPLOY_ENV:?}"
+          if [ -n "$DEPLOY_COLOR" ]; then
+            LIMIT+=":&tag_color_${DEPLOY_COLOR}"
```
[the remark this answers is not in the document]
dincho: It's bash 3.1 feature: http://tldp.org/LDP/abs/html/bashver3.html

dincho:
> At least mentioning what is meant to come first - blue or green (it can be confusing), how many clicks expected to do (two unhold - so not to leave partially completed deployment), at which point check which artifacts (Docker images, binary packages) uploaded, what to do in case of half deployed environment (e.g. Circle CI workflow timed out or corrupted after half environment deployed - sentence "ask DevOps" is a good text if such were the case :) ).

I have added couple of sentences to the Wiki: https://github.com/aeternity/epoch/wiki/Release-process/_compare/c953b52b0f223797fcdc4941a75318edee4e27f1...74f96f819fd263886ad98d6c1e75a06cc5d8abc6

However, as this is visual workflow showing all the steps and unhold buttons, I doubt anyone of the team will get confused. Moreover, the changes conceptually are actually very small and simple - one additional step compliments the previous workflow.

docs/operation.md
lucafavatella: You convinced me. Please leave.

.circleci/config.yml
dincho: > Does hodl_green and the rest of the workflow not time out in - let us say - 3 days? (Checking what the workflow is meant to be.)

I sent the question to CircleCI support as I can't find it documented.

dincho: > A related question is whether such a partially-completed deployment shall be completed by pushing branch env/uat_green (and whether that works - see comment below).

Partially-completed deployment should be retried with workflow retry.

.circleci/config.yml
dincho: I decided that we don't need manual blue/green UAT triggers, because I don't see a use case to deploy random source tree on UAT, it's should be sequential following the release process.

The source code to be release could be tagged on the previous day to trigger release workflow and deploy on blue env., then on the next (release) day, the green deployment should be unhold.

see 1e3363e

.circleci/config.yml
dincho: see 1e3363e

dincho:
@lucafavatella After my comment on [manual deployments](https://github.com/aeternity/epoch/pull/1243#discussion_r195383671), I've one more question:

Shall we push GH release binaries and docker image after the second (green) approval ?

.circleci/config.yml
dincho: > I sent the question to CircleCI support as I can't find it documented.

Just received an reply, we're all good. We can leave it waiting for a week or so. The timeout is only on job execution.

lucafavatella: > Partially-completed deployment should be retried with workflow retry.

Workflow retry or [job retry](https://github.com/aeternity/epoch/wiki/Release-process/_compare/c953b52b0f223797fcdc4941a75318edee4e27f1...74f96f819fd263886ad98d6c1e75a06cc5d8abc6#diff-991db8edfb34dd9dafc5302987e97019R28)?

lucafavatella:
> Shall we push GH release binaries and docker image after the second (green) approval ?

GH release binaries could be uploaded straight after the first (blue) approval, as GH release has concept of pre-release vs. release.

Re Docker image - would it be possible to push tagged image after first (blue) approval and "latest" image after second (green) approval?

Other question is the swagger file upload, that would be better be uploaded after second (green) approval.

dincho:
> Re Docker image - would it be possible to push tagged image after first (blue) approval and "latest" image after second (green) approval?

Last time I checked we were unable to reliable share the built image between jobs, thus that would mean double image builds that I don't like.

dincho:
> Other question is the swagger file upload, that would be better be uploaded after second (green) approval.

@lucafavatella addressed in e84b8d7d

lucafavatella:
> Last time I checked we were unable to reliable share the built image between jobs, thus that would mean double image builds that I don't like.

Fetch of known tag, then tag as latest, then push?

.circleci/config.yml
dincho: Yes, job retry.

dincho:
> Fetch of known tag, then tag as latest, then push?

We could try to split and optimize our docker related builds based on this, but I'd prefer to separate it from this ticket.

PR #1245, opened 2018-06-13
Introduce efficient TX-pool sync

hanssv:
Previously we sent the entire mempool after every ping message - this is
obviuosly not scaling. Here we introduce a proper sync mechanism.
TX-pool synchronization is done between a pair of peers. Since the node
will still work if it miss a transaction or two it is better to keep it
somewhat simple.

The sync is handled by aec_tx_pool_sync which allows for 1 outgoing and
5 incoming simultaneous synchronizations. The synchronization is done by
building an MP tree on each side and only sending the parts of the trees
that differ - thus minimizing the bandwidth used.


hanssv:
Corresponding updates to the protocol repo: https://github.com/aeternity/protocol/pull/118

docs/release-notes/RELEASE-NOTES-0.16.0.md
```diff
@@ -2,6 +2,8 @@
 
 [This release][this-release] is focused on TODOFILLMEIN.
 It:
+* Improves TX-pool synchronization. It is now synced (with only missing transactions being
```
dincho: What if the node gets out of sync for some reason (e.g. temporary connection issue)? Does it need to be restarted to catch up?

hanssv: First, it will still work, although potentially missing some transactions is an issue if it becomes the leader (otherwise it will pull missing transactions when a block contains them). 

The tricky thing is for the node to realize it is out of synch, if it can do this it should be straightforward to re-synch (adding an API for this is 10 minutes work)... Might be worth thinking about.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -414,21 +432,28 @@ handle_msg(S, _MsgType, true, {RequestFld, From, _TRef}, {error, Reason}) ->
...
-        mempool -> handle_get_mempool_rsp(S, Request, Msg);
-        header  -> handle_header_rsp(S, Request, Msg);
+        ping          -> handle_ping_rsp(S, Request, Msg);
+        header        -> handle_header_rsp(S, Request, Msg);
         header_hashes -> handle_header_hashes_rsp(S, Request, Msg);
-        block   -> handle_get_block_rsp(S, Request, Msg)
+        block         -> handle_get_block_rsp(S, Request, Msg);
+        txps_init     -> handle_tx_pool_sync_rsp(S, init, Request, Msg);
+        txps_unfold   -> handle_tx_pool_sync_rsp(S, unfold, Request, Msg);
+        txs           -> handle_tx_pool_sync_rsp(S, get, Request, Msg);
```
ThomasArts: I would expect to see txps_get here instead?

hanssv: Why? The response to a `txps_get` is a `txs` The others happen to have the same response as the request...

system_test/aest_sync_SUITE.erl
```diff
@@ -334,6 +337,103 @@ stop_and_continue_sync(Cfg) ->
...
+
+    %% Add 5 valid spend transactions
+    ValidTxs = add_spend_txs(node1, Patron, 5, 1),
+    %% Add 10 invalid (nonce_too_high) spend transactions
+    InvalidTxs = add_spend_txs(node1, Patron, 5, 7),
+    add_spend_txs(node1, Patron, 5, 15),
+
+
+    %% Check that the valid transactions made it to the chain.
+    #{ reciever := RecvAccount, amount := Amount } = lists:last(ValidTxs),
```
lucafavatella: "receiver". Though as typo consistent, ok...

system_test/aest_sync_SUITE.erl
```diff
@@ -334,6 +337,103 @@ stop_and_continue_sync(Cfg) ->
...
+    wait_for_value({balance, aec_base58c:encode(account_pubkey, RecvAccount), Amount},
+                   [node1], 5 * ?MINING_TIMEOUT, Cfg),
+
+    %% Check that the mempool has the other transactions
+    {ok, 200, MempoolTxs1} = request(node1, 'GetTxs', #{}),
+    true = (length(MempoolTxs1) == 10),
```
lucafavatella: May I suggest making assertion more descriptive in case of failure so to enable easier investigation? E.g. `{10, _} = {length(MempoolTxs1), MempoolTxs1}` or `?assertMatch(X when length(X) =:= 10, MempoolTxs1)`

system_test/aest_sync_SUITE.erl
```diff
@@ -334,6 +337,103 @@ stop_and_continue_sync(Cfg) ->
...
+    %% Add 5 valid spend transactions
+    ValidTxs = add_spend_txs(node1, Patron, 5, 1),
+    %% Add 10 invalid (nonce_too_high) spend transactions
+    InvalidTxs = add_spend_txs(node1, Patron, 5, 7),
+    add_spend_txs(node1, Patron, 5, 15),
+
+
+    %% Check that the valid transactions made it to the chain.
+    #{ reciever := RecvAccount, amount := Amount } = lists:last(ValidTxs),
+    wait_for_value({balance, aec_base58c:encode(account_pubkey, RecvAccount), Amount},
```
lucafavatella: The receiver account is distinct in each spend, so this is checking that last tx included. And the last tx has highest nonce, hence its inclusion implies inclusion of txs with lower nonce too. Ok.

system_test/aest_sync_SUITE.erl
```diff
@@ -334,6 +337,103 @@ stop_and_continue_sync(Cfg) ->
...
+    %% Start 2nd node and let it sync
+    start_node(node2, Cfg),
+    wait_for_value({height, 0}, [node2], NodeStartupTime, Cfg),
+
+    %% Give the sync a moment to finish
+    #{ height := Height1 } = get_top(node1),
+    wait_for_value({height, Height1 + 5}, [node2], 5 * ?MINING_TIMEOUT, Cfg),
+
+    {ok, 200, MempoolTxs2} = request(node2, 'GetTxs', #{}),
+    true = (length(MempoolTxs2) == 10),
```
lucafavatella: Semilarly - more descriptive error in case of failure maybe?

system_test/aest_sync_SUITE.erl
```diff
@@ -334,6 +337,103 @@ stop_and_continue_sync(Cfg) ->
...
+
+    %% Stop node1
+    stop_node(node1, 8000, Cfg),
```
lucafavatella: I see this means graceful shutdown. Eventually this hardcoded 8000 shall be removed though ok.

system_test/aest_sync_SUITE.erl
```diff
@@ -334,6 +337,103 @@ stop_and_continue_sync(Cfg) ->
...
+    %% TODO: Automate check that _only_ this Tx is synced.
+    start_node(node1, Cfg),
+    wait_for_value({height, 0}, [node1], NodeStartupTime, Cfg),
+
+    %% Give the sync a moment to finish
+    #{ height := Height2 } = get_top(node2),
+    wait_for_value({height, Height2 + 5}, [node1], 5 * ?MINING_TIMEOUT, Cfg),
+
+    {ok, 200, MempoolTxs1B} = request(node1, 'GetTxs', #{}),
+    true = (length(MempoolTxs1B) == 11),
```
lucafavatella: As above.

system_test/aest_sync_SUITE.erl
hanssv: That is code-completetion at its worst :-( 

As an Erlang-programmer receive should come naturally you'd think...

apps/aecore/src/aec_tx_pool_sync.erl
```diff
@@ -0,0 +1,477 @@
...
+    lager:warning("Ignoring unknown cast message: ~p", [Msg]),
+    {noreply, State}.
+
+handle_info({local_action, Ref, TimerRef, Action, Result}, State) ->
+    erlang:cancel_timer(TimerRef),
+    %% We might have already gotten the message :-(
+    receive {timeout, TimerRef, _} -> ok
+    after 0 -> ok end,
+    {noreply, handle_local_action(State, Action, Ref, Result)};
+handle_info({timeout, _TimerRef, {local_action, Ref, Worker, Action}}, State) ->
```
lucafavatella: Who adds this `{timeout, _}` wrapper tuple?

----

```
$ erl
Erlang/OTP 20 [erts-9.2.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false]

Eshell V9.2.1  (abort with ^G)
1> Timeout = 1000, TimeoutMsg = {local_action, a_ref, a_pid, an_action}, ok.
ok
2> erlang:send_after(Timeout, self(), TimeoutMsg).
#Ref<0.959383622.920387587.49031>
3> flush().
Shell got {local_action,a_ref,a_pid,an_action}
ok
```

apps/aecore/src/aec_tx_pool_sync.erl
```diff
@@ -0,0 +1,477 @@
...
+handle_cast({finish, PeerId, _Status}, State) ->
+    {noreply, handle_finish(State, PeerId)};
+handle_cast(Msg, State) ->
+    lager:warning("Ignoring unknown cast message: ~p", [Msg]),
+    {noreply, State}.
+
+handle_info({local_action, Ref, TimerRef, Action, Result}, State) ->
+    erlang:cancel_timer(TimerRef),
+    %% We might have already gotten the message :-(
+    receive {timeout, TimerRef, _} -> ok
```
lucafavatella: Who adds this `{timeout, _}` wrapper tuple?

----

```
$ erl
Erlang/OTP 20 [erts-9.2.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false]

Eshell V9.2.1  (abort with ^G)
1> Timeout = 1000, TimeoutMsg = {local_action, a_ref, a_pid, an_action}, ok.
ok
2> erlang:send_after(Timeout, self(), TimeoutMsg).
#Ref<0.959383622.920387587.49031>
3> flush().
Shell got {local_action,a_ref,a_pid,an_action}
ok
```

apps/aecore/src/aec_tx_pool_sync.erl
hanssv: `erlang:start_timer/3,4`

apps/aecore/src/aec_tx_pool_sync.erl
hanssv: `erlang:start_timer/3,4`

apps/aecore/src/aec_tx_pool_sync.erl
```diff
@@ -0,0 +1,477 @@
...
+serialize_unfolds(Us) ->
+    [ serialize_unfold(U) || U <- Us ].
+
+deserialize_unfolds(SUs) ->
+    [ deserialize_unfold(SU) || SU <- SUs ].
+
+serialize_unfold({node, Path, Node}) ->
+    aeu_rlp:encode(
+        aec_serialization:encode_fields(
+            [{type, int}, {path, binary}, {node, binary}],
```
lucafavatella: I see that `aec_peer_connection` calls `aec_blocks:serialize_to_binary/1` for serializing block that in turn relies on [`aec_object_serialization` recognizing internal tag `block` mapped to `100`](https://github.com/aeternity/epoch/blob/aa2017a6e94164ae19940565c29aea4b39a32b12/apps/aecore/src/aec_object_serialization.erl#L69) that is - correctly - not documented among [serializations meant to affect consensus](https://github.com/aeternity/protocol/blob/faf1815bcc8900a8c0020d09904b47cb9daf8640/serializations.md#block). Maybe these new serializations shall follow same path?

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -711,23 +751,57 @@ handle_get_block_rsp(S, {get_block, From, _TRef}, Msg) ->
...
+
+handle_tx_pool_sync_rsp(S, Action, {tx_pool, From, _TRef}, MsgObj) ->
+    case Action of
+        init ->
+            gen_server:reply(From, ok);
+        unfold ->
+            gen_server:reply(From, {ok, maps:get(unfolds, MsgObj, [])});
+        get ->
+            Txs = [ aetx_sign:deserialize_from_binary(SerTx)
+                    || SerTx <- maps:get(txs, MsgObj, []) ],
```
lucafavatella: Why default value `[]`? Is `txs` field not guaranteed by its deserialization both [in code](https://github.com/aeternity/epoch/pull/1245/files#diff-c1692ca9a93deb9e4ca76237ebbc01fdR209) and [in spec](https://github.com/aeternity/protocol/pull/118/files#diff-8fff5937ba66ea026ab1b2fd51a3d518R133)? I find it hinders readability.

apps/aecore/src/aec_tx_pool_sync.erl
```diff
@@ -0,0 +1,477 @@
...
+unfold({node, Path, Node}, Tree) ->
+    aeu_mp_trees:unfold(Path, Node, Tree);
+unfold({leaf, Path}, _Tree) ->
+    {leaf, Path};
+unfold({subtree, Path}, Tree) ->
+    get_subtree(Path, Tree);
+unfold({key, Key}, _Tree) ->
+    [{key, Key}].
+
+get_subtree(Key, _Tree) when bit_size(Key) =:= 32 * 8 ->
```
lucafavatella: s/32/`?TXS_HASH_BYTES` ?

apps/aecore/src/aec_tx_pool_sync.erl
```diff
@@ -0,0 +1,477 @@
...
+    {leaf, deserialize_niblets(Path)};
+deserialize_unfold(?SUBTREE_TAG, Flds) ->
+    [{subtree, Path}] = aec_serialization:decode_fields([{subtree, binary}], Flds),
+    {subtree, deserialize_niblets(Path)};
+deserialize_unfold(?KEY_TAG, Flds) ->
+    [{key, Key}] = aec_serialization:decode_fields([{key, binary}], Flds),
+    {key, Key}.
+
+%% Paths are not necessarily even bytes, this is required by RLP so we have to
+%% pad this using parity information.
```
gorillainduction: Internally in the `aeu_mp_trees.erl` this is handled here:

https://github.com/aeternity/epoch/blob/aa2017a6e94164ae19940565c29aea4b39a32b12/apps/aeutils/src/aeu_mp_trees.erl#L831

apps/aecore/src/aec_tx_pool_sync.erl
```diff
@@ -0,0 +1,477 @@
...
+
+handle_unfold(State, PeerId, SerUnfolds) ->
+    case lists:keyfind(PeerId, #sync.peer_id, State#state.remote) of
+        false ->
+            {error, no_active_sync};
+        #sync{ tree = undefined } ->
+            {error, tree_not_ready};
+        #sync{ tree = Tree } ->
+            Unfolds = deserialize_unfolds(SerUnfolds),
+            NewUnfolds = lists:append([ unfold(Unfold, Tree) || Unfold <- Unfolds ]),
```
lucafavatella: As `SerUnfolds` comes from the network and is not (deeply) sanitized, among the items in the list `Unfolds` there could be items `{error, term()}` that would make function `unfold/2` err for function clause hence causing `aec_tx_pool_sync` to restart. Expected?

apps/aecore/src/aec_tx_pool_sync.erl
```diff
@@ -0,0 +1,477 @@
...
+        STx = aec_db:get_signed_tx(TxHash),
+        [STx]
+    catch _:_ ->
+        []
+    end.
+
+unfold({node, Path, Node}, Tree) ->
+    aeu_mp_trees:unfold(Path, Node, Tree);
+unfold({leaf, Path}, _Tree) ->
+    {leaf, Path};
```
lucafavatella: `[{leaf, Path}]`?

That begs the question of whether this clause can ever be reached (I did not check).

----

```
$ erl
Erlang/OTP 20 [erts-9.2.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false]

Eshell V9.2.1  (abort with ^G)
1> lists:append([{leaf, a_path}, [{key, a_key}]]).
** exception error: bad argument
     in operator  ++/2
        called as {leaf,a_path} ++ [{key,a_key}]
     in call from lists:append/1 (lists.erl, line 127)
```

apps/aecore/src/aec_tx_pool_sync.erl
lucafavatella: This clause looks an optimization for performing 0 lookups instead of two. Ok. (I would prefer not having the clause at all for maintainability though debatable.)

apps/aecore/src/aec_tx_pool_sync.erl
```diff
@@ -0,0 +1,477 @@
...
+    {node, deserialize_niblets(Path), Node};
+deserialize_unfold(?LEAF_TAG, Flds) ->
+    [{leaf, Path}] = aec_serialization:decode_fields([{leaf, binary}], Flds),
+    {leaf, deserialize_niblets(Path)};
+deserialize_unfold(?SUBTREE_TAG, Flds) ->
+    [{subtree, Path}] = aec_serialization:decode_fields([{subtree, binary}], Flds),
+    {subtree, deserialize_niblets(Path)};
+deserialize_unfold(?KEY_TAG, Flds) ->
+    [{key, Key}] = aec_serialization:decode_fields([{key, binary}], Flds),
+    {key, Key}.
```
lucafavatella: In order to guarantee invariant of key is [non-empty](https://github.com/aeternity/epoch/pull/1245/files#diff-0c07dd015037eb446bbd6e7298a61532R372), shall this point not check that and, if not the case, terminate abnormally as deserialization failed? (Caller `deserialize_unfold` handler runtime errors.)

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -711,23 +751,57 @@ handle_get_block_rsp(S, {get_block, From, _TRef}, Msg) ->
...
+    aec_tx_pool_sync:sync_finish(peer_id(S), {done, maps:get(done, MsgObj, false)}),
+    send_response(S, txps_finish, {ok, #{ done => maps:get(done, MsgObj, false) }}),
+    S.
+
+handle_tx_pool_sync_rsp(S, Action, {tx_pool, From, _TRef}, MsgObj) ->
+    case Action of
+        init ->
+            gen_server:reply(From, ok);
+        unfold ->
+            gen_server:reply(From, {ok, maps:get(unfolds, MsgObj, [])});
```
lucafavatella: Why default `[]`? Is this not guaranteed by https://github.com/aeternity/epoch/pull/1245/files#diff-c1692ca9a93deb9e4ca76237ebbc01fdR217 ? I find that unnecessary default hinders readability (though I understand may be debatable).

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -711,23 +751,57 @@ handle_get_block_rsp(S, {get_block, From, _TRef}, Msg) ->
...
+    case aec_tx_pool_sync:accept(peer_id(S), self()) of
+        ok ->
+            send_response(S, txps_init, {ok, #{}});
+        Err = {error, _} ->
+            send_response(S, txps_init, Err)
+    end,
+    S.
+
+handle_tx_pool_sync_unfold(S, MsgObj) ->
+    Unfolds = maps:get(unfolds, MsgObj, []),
```
lucafavatella: Similarly - why `[]`? Is this not guaranteed by https://github.com/aeternity/epoch/pull/1245/files#diff-c1692ca9a93deb9e4ca76237ebbc01fdR217 ?

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -711,23 +751,57 @@ handle_get_block_rsp(S, {get_block, From, _TRef}, Msg) ->
...
+            send_response(S, txps_unfold, Err)
+    end,
+    S.
 
-handle_get_mempool(S, _MsgObj) ->
-    {ok, Txs0} = aec_tx_pool:peek(infinity),
-    Txs = [ aetx_sign:serialize_to_binary(T) || T <- Txs0 ],
-    send_response(S, mempool, {ok, #{ txs => Txs }}),
+handle_tx_pool_sync_get(S, MsgObj) ->
+    TxHashes = maps:get(tx_hashes, MsgObj, []),
```
lucafavatella: Similarly - why `[]`? It shall be guaranteed by https://github.com/aeternity/epoch/pull/1245/files#diff-c1692ca9a93deb9e4ca76237ebbc01fdR221

apps/aecore/src/aec_tx_pool_sync.erl
```diff
@@ -0,0 +1,477 @@
...
+                             [Delay + 1000]),
+            do_local_action(Sync, {unfold, OldUnfolds, Delay + 1000}),
+            State;
+        {unfold, {ok, tree_not_ready, _OldUnfolds, _Delay}} ->
+            epoch_sync:debug("Remote tree permanent failure, aborting", []),
+            do_local_action(Sync, {finish, false}),
+            State;
+        {unfold, {ok, NewUnfolds}} ->
+            #sync{ data = OldGets } = Sync,
+            {NewUnfolds1, NewGets} = analyze_unfolds(NewUnfolds, Sync#sync.tree),
```
lucafavatella: This may terminate abnormally because [deserialize_unfolds](https://github.com/aeternity/epoch/pull/1245/files#diff-0c07dd015037eb446bbd6e7298a61532R290) may return, among the items in the returned list, `{error, _}` (as input from network, and not deeply sanitized because only sanitized up to simple [deserialization](https://github.com/aeternity/epoch/pull/1245/files#diff-c1692ca9a93deb9e4ca76237ebbc01fdR217)). Expected?

apps/aecore/src/aec_peer_connection.erl
hanssv: Yes, I agree, that default value is not necessary...

apps/aecore/src/aec_tx_pool_sync.erl
hanssv: It is actually necessary since the iterator on aeu_mp_trees will start from a Path *not including Path itself*.

apps/aecore/src/aec_peer_connection.erl
hanssv: Yes, should never be used

apps/aecore/src/aec_tx_pool_sync.erl
hanssv: No, that should never be reached, we shouldn't call unfold with a leaf... It will just be in the return value of an unfold. But for completeness it makes sense ot be able to call this local unfold for all "values" of the "unfold type".

apps/aecore/src/aec_tx_pool_sync.erl
hanssv: It is not sanitized, but it is created by serialize_unfolds (and sent over an encrypted channel) thus it better deserialize properly...

apps/aecore/src/aec_tx_pool_sync.erl
```diff
@@ -0,0 +1,477 @@
...
+analyze_unfolds(Us, Tree) ->
+    analyze_unfolds(Us, Tree, [], []).
+
+analyze_unfolds([], _Tree, NewUs, NewGets) ->
+    {lists:reverse(NewUs), lists:reverse(NewGets)};
+analyze_unfolds([{key, Key} | Us], Tree, NewUs, NewGets) ->
+    analyze_unfolds(Us, Tree, NewUs, [Key | NewGets]);
+analyze_unfolds([{leaf, Path} | Us], Tree, NewUs, NewGets) ->
+    case aeu_mp_trees:get(Path, Tree) of
+        <<>> -> analyze_unfolds(Us, Tree, NewUs, [Path | NewGets]);
```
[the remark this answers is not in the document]
hanssv: Yes, the local tree does not have this key (TxHash) thus we need to get this Tx.

apps/aecore/src/aec_tx_pool_sync.erl
hanssv: I actually never expect deserialize_unfold to return {error, _} since all data it is called on is produced by serialize_unfold which is (or at least should be) symmetric...

apps/aecore/src/aec_tx_pool_sync.erl
hanssv: The key cannot be empty? Keys are only returned by get_subtree and keys should represent TxHashes? - Not sure I understand the comment fully?

apps/aecore/src/aec_tx_pool_sync.erl
hanssv: Will make the error case explicit.

apps/aecore/src/aec_tx_pool_sync.erl
hanssv: Making the error explicit and handled.

hanssv:
Since Luca spent lots of effort on reviewing I made another commit where I address (and add more type spec's) the comments. 

Message size should not be an issue we do have internal framing (on top of Noise) thus we should be able to handle messages up to 4 GByte - and I really hope the mempool wont reach that size... But for other reasons it might be a good idea to split the getting Txs into smaller chunks.

hanssv:
But Luca's comment on scaling made me think of a potential bottleneck, so last commit moves the handling of Txs to the worker and alots it a bit more time...

apps/aecore/src/aec_tx_pool_sync.erl
lucafavatella: The peer may send an empty one? It is untrusted, right?

PR #1246, opened 2018-06-14
Pt 158318830 secure priv key

gorillainduction:
https://www.pivotaltracker.com/n/projects/2124891/stories/158318830

### DIFF
--- a/apps/aecore/src/aec_keys.erl
+++ b/apps/aecore/src/aec_keys.erl
@@ -11,8 +11,6 @@
 
 -behaviour(gen_server).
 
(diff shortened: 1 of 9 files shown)


apps/aecore/src/aec_keys.erl
```diff
@@ -149,35 +189,45 @@ check_key_pair() ->
 %%                     {stop, Reason}
 %% @end
 %%--------------------------------------------------------------------
+
+-ifdef(TEST).
```
hanssv: If only you could ifdef separate clauses...

PR #1249, opened 2018-06-15
Security vulnerability

ThomasArts:
Probably a bit paranoic, but reading the private key in memory in order to see if it already exists is not needed.

### DIFF
--- a/apps/aecore/src/aec_keys.erl
+++ b/apps/aecore/src/aec_keys.erl
@@ -507,22 +507,13 @@ p_gen_filename(KeysDir, PubFile0, PrivFile0) ->
     {PubFile, PrivFile}.
 
 p_save_keys(Pwd, PubFile, PubKey, PrivFile, PrivKey) ->
+    %% We only get here after reading the file. If it already exists at this point
+    %% the user must have set write but not read permissions or so.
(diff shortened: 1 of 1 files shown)


apps/aecore/src/aec_keys.erl
```diff
@@ -514,12 +514,10 @@ p_save_keys(Pwd, PubFile, PubKey, PrivFile, PrivKey) ->
     {EncPub, EncPriv}.
 
 to_local_dir(NewFile, Bin) ->
-    case file:read_file(NewFile) of
-        {error, enoent} ->
-            {ok, IODevice} = file:open(NewFile, [write, binary, raw]),
-            ok = file:write_file(NewFile, Bin),
-            ok = file:close(IODevice);
-        {ok, _Out} ->
+    case filelib:is_file(NewFile) of
```
lucafavatella: Ok http://erlang.org/doc/man/filelib.html#is_file-1

```
1> filelib:is_file("foo").
false
2> filelib:is_file("apps").
true
```

hanssv:
You could just remove the check... 

We only reach here if we have already failed to read the keys? And if the user has his file not readable but writeable he deserves to have it overwritten!

PR #1251, opened 2018-06-15
[RFC] User API for retrieving contract call object for contract create init

lucafavatella:
I open this PR for getting feedback on how [the user API for fetching the call object resulting from the execution of the initialization function in the contract create tx](https://www.pivotaltracker.com/story/show/158152982) will look like. The main point is that I plan to overload this on the existing path for fetching the call object for a contract call tx - see swagger.yaml file.

This PR is not meant to be merged yet.

TODO:
* [ ] Code
* [ ] Test coverage for contract create init failure
* [ ] Ensure Python UAT tests do not break


lucafavatella:
Thanks for the comments, I will put together new PR.

PR #1255, opened 2018-06-18
PT-157316323 mempool gc

hanssv:
Introduce garbage collection of TXs 

Transactions where TTL expire or nonce gets stale are removed after 5 blocks. Transactions that are not applicable for 256 consecutive blocks are also removed.

### DIFF
--- a/apps/aecore/src/aec_block_candidate.erl
+++ b/apps/aecore/src/aec_block_candidate.erl
@@ -46,16 +46,17 @@ create(Block) ->
                       aec_blocks:height(), non_neg_integer()) ->
         {ok, list(aetx_sign:signed_tx()), aec_trees:trees()}.
 apply_block_txs(Txs, Miner, Trees, Height, Version) ->
-    {ok, Txs1, Trees1, _} = int_apply_block_txs(Txs, Miner, Trees, Height, Version, false),
-    {ok, Txs1, Trees1}.
+    {ok, ValidTxs, _, Trees1, _} =
+        int_apply_block_txs(Txs, Miner, Trees, Height, Version, false),
+    {ok, ValidTxs, Trees1}.
 
 -spec apply_block_txs_strict(list(aetx_sign:signed_tx()), aec_keys:pubkey(),
                              aec_trees:trees(), aec_blocks:height(), non_neg_integer()) ->
         {ok, list(aetx_sign:signed_tx()), aec_trees:trees()} | {error, term()}.
 apply_block_txs_strict(Txs, Miner, Trees, Height, Version) ->
     case int_apply_block_txs(Txs, Miner, Trees, Height, Version, true) of
-        Err = {error, _}      -> Err;
-        {ok, Txs1, Trees1, _} -> {ok, Txs1, Trees1}
+        Err = {error, _}             -> Err;
+        {ok, ValidTxs, _, Trees1, _} -> {ok, ValidTxs, Trees1}
     end.
 
 -spec create_with_state(aec_blocks:block(), aec_keys:pubkey(),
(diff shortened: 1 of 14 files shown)


apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -0,0 +1,141 @@
...
+    %% Add a bunch of transactions...
+    add_spend_tx(N1, 1000, 1,  1,  10), %% Ok
+    add_spend_tx(N1, 1000, 1,  2,  10), %% Should expire ?EXPIRE_TX_TTL after
+                                        %% next TX is on chain = 2 + 2 = 4
+    add_spend_tx(N1, 1000, 10, 2,  10), %% Duplicate should be preferred
+    add_spend_tx(N1, 1000, 1,  3,  10), %% Ok
+
+    add_spend_tx(N1, 1000, 1,  5,  10), %% Non consecutive nonce
+    add_spend_tx(N1, 1000, 1,  7,  10), %% Non consecutive nonce
+    add_spend_tx(N1, 1000, 1,  8,  4),  %% Short TTL - expires at 4 + 2 = 6
```
lucafavatella: I see that the specified 4 is the TTL as absolute height, and the magic 2 comes from [here](https://github.com/aeternity/epoch/pull/1255/commits/249b6a3e538029cc020f8f07f1b2a7f976933a5d?utf8=%E2%9C%93&diff=split&w=1#diff-2e95a69d7723aebad303814cb3f38e7eR80). Ok.

apps/aecore/src/aec_block_candidate.erl
```diff
@@ -183,14 +184,16 @@ int_create_block(PrevBlockHash, PrevBlock, Trees, Miner, Txs) ->
     Height = PrevBlockHeight + 1,
     Version = aec_hard_forks:protocol_effective_at_height(Height),
 
-    {ok, Txs1, Trees2, Fees} =
+    {ok, ValidTxs, InvalidTxs, Trees2, Fees} =
         int_apply_block_txs(Txs, Miner, Trees, Height, Version, false),
 
-    TxsTree = aec_txs_trees:from_txs(Txs1),
+    aec_tx_pool:invalid_txs([ aetx_sign:hash(STx) || STx <- InvalidTxs ], Height),
```
lucafavatella: Reading this line it feels it does not fit the responsibility of the exported function. Maybe `InvalidTxs` shall be returned by this function, in turn returned by the only caller function used by the system that is `aec_block_candidate:create/1`, and this line of code shall be in `aec_block_generator:create_block_candidate/1`? Totally optional.

----

I am trying to understand synchronous/asynchronous operations for potential intermittent test failures introduced by this PR (still getting my head around it). This is where this comment is coming from.

hanssv: Hmm, yes, good point. Since `aec_tx_pool:invalid_txs` is asynchronous it should not make a differenc functionality-wise but doing it here is definitely slightly lazy approach!

lucafavatella: Any reasons why `aec_tx_pool:invalid_txs` is not called also in code path around block candidate update?

hanssv: Small (not-obvious) optimization - garbage collect is only run after `create` thus there is no point in adding invalid_txs at this point. If they are still invalid in `create` they will be added, and otherwise we did save some work!

apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -25,6 +25,20 @@
                 "pattern": "^aenode://pp\$[a-zA-Z0-9]+@[^:\\.\"!#$%^&*()',/]+(\\.[^:\\.\"!#$%^&*()',/]+)*:[0-9]+/*$"
             }
         },
+        "mempool" : {
```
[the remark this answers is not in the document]
hanssv: Yes, it is updated??!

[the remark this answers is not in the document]
lucafavatella: it is, had not noticed it sorry.

hanssv:
Non-mining nodes will not do GC. This will change once NG lands and this mechanism is properly updated...

PR #1258, opened 2018-06-18
Replace channel slash transaction with extended solo close transaction

gorillainduction:
The difference in semantics of the two transactions were only when
they could be used. Now the solo close transaction can be used as
a slash transaction instead.

https://www.pivotaltracker.com/n/projects/2124891/stories/158421035


apps/aechannel/src/aesc_close_solo_tx.erl
```diff
@@ -216,28 +216,45 @@ serialization_template(?CHANNEL_CLOSE_SOLO_TX_VSN) ->
...
+    InitiatorPubKey   = aesc_channels:initiator(Channel),
+    ResponderPubKey   = aesc_channels:responder(Channel),
+    ChannelAmount     = aesc_channels:total_amount(Channel),
+    case aesc_utils:accounts_in_poi([InitiatorPubKey, ResponderPubKey], PoI) of
+        {error, _} = Err -> Err;
+        {ok, [PoIInitiatorAcc, PoIResponderAcc]} ->
+            PoIInitiatorAmt = aec_accounts:balance(PoIInitiatorAcc),
+            PoIResponderAmt = aec_accounts:balance(PoIResponderAcc),
+            PoIAmount       = PoIInitiatorAmt + PoIResponderAmt,
+            case ChannelAmount =:= PoIAmount of
```
uwiger: Would it not be acceptable here with `ChannelAmount >= PoIAmount`?

gorillainduction: Yes. This is a restriction that we have discussed to lift, but I aimed at not changing the semantics for now.

lucafavatella:
I took the liberty of labelling as WIP for the time being - it eases filtering.

gorillainduction:
Closing this PR because of the reasons @lucafavatella gives above.

PR #1259, opened 2018-06-18
Enable user to fetch call object from user API

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/158152982

Supersedes #1251


lucafavatella:
Rebased on master for release notes conflict.

PR #1260, opened 2018-06-19
ng rebuild

sennui:
Known issues:
- [ ] - micro and key blocks re-use the same record 
- [ ] - still no beneficiary (next sprint)
- [ ] - no proof of fraud (next sprint)

Further issues to discuss:
- [ ] - default values for hash fields in blocks. Zeros may not be the best

### DIFF
--- a/apps/aehttp/test/aehttp_integration_SUITE.erl
+++ b/apps/aehttp/test/aehttp_integration_SUITE.erl
@@ -592,7 +592,7 @@ contract_transactions(_Config) ->
                 <<"contract_address">> := EncodedContractPubKey}} =
         get_contract_create(ValidEncoded),
     {ok, ContractPubKey} = aec_base58c:safe_decode(contract_pubkey, EncodedContractPubKey),
-    ContractCreateTxHash = sign_and_post_tx(MinerAddress, EncodedUnsignedContractCreateTx),
+    ContractCreateTxHash = sign_and_post_tx(EncodedUnsignedContractCreateTx),
 
     % mine a block
     aecore_suite_utils:mine_blocks(aecore_suite_utils:node_name(?NODE), 2),
@@ -626,7 +626,7 @@ contract_transactions(_Config) ->
                                fun aect_call_tx:new/1, MinerPubkey),
 
     {ok, 200, #{<<"tx">> := EncodedUnsignedContractCallTx}} = get_contract_call(ContractCallEncoded),
-    ContractCallTxHash = sign_and_post_tx(MinerAddress, EncodedUnsignedContractCallTx),
+    ContractCallTxHash = sign_and_post_tx(EncodedUnsignedContractCallTx),
 
     %% Try to get the call object while in mempool
     {ok, 400, #{<<"reason">> := <<"Tx not mined">>}} =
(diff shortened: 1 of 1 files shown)


apps/aecore/src/aec_conductor.erl
```diff
@@ -632,16 +661,139 @@ handle_mining_reply({{error, {runtime, Reason}}, _}, State) ->
...
+                    seen_top_block_hash = SeenHash} = State) when MicroBlockHash =/= SeenHash ->
+    %% Candidate generated with stale top hash.
+    %% Regenerate the candidate.
+    State#state{micro_block_candidate = undefined};
+start_micro_signing(#state{consensus = #consensus{leader = true},
+                           micro_block_candidate = MicroCandidate} = State) ->
+    epoch_mining:info("Starting signing round"),
+    HeaderBin = MicroCandidate#candidate.bin,
+    Info      = [{top_block_hash, State#state.seen_top_block_hash}],
+    aec_events:publish(start_micro_signing, Info),
```
hanssv: Who is listening to this?

apps/aecore/src/aec_txs_trees.erl
```diff
@@ -30,8 +30,12 @@
 %%% API
 %%%===================================================================
 
--spec from_txs([aetx_sign:signed_tx()]) -> txs_tree().
-from_txs(Txs) ->
+-spec from_txs([aetx_sign:signed_tx(), ...] | []) -> txs_tree().
+
+from_txs([]) ->
+    %% NG-INFO: its fine to have empty transaction list (key-block)
```
hanssv: Should go away

apps/aetx/src/aetx_sign.erl
```diff
@@ -57,18 +60,24 @@ new(Tx, Signatures) ->
 %% @doc Given a transaction Tx, a private key or list of keys,
 %% return the cryptographically signed transaction using the default crypto
 %% parameters.
--spec sign(aetx:tx(), list(binary()) | binary()) -> signed_tx().
+-spec sign(aetx:tx()|binary(), list(binary()) | binary()) -> signed_tx() | tuple().
```
hanssv: I don't like this spec-change... This function is a signing a transaction, not a general binary.

apps/aecore/src/aec_blocks.erl
```diff
@@ -58,6 +65,12 @@ target(Block) ->
 difficulty(Block) ->
     aec_pow:target_to_difficulty(target(Block)).
 
+-spec is_key_block(block()) -> boolean().
+is_key_block(#block{miner = Miner, height = Height}) ->
```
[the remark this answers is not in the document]
sennui: yes, it will into fall patterns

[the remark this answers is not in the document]
jur0: Yes. There will be separate records for key and micro blocks.

apps/aetx/src/aetx_sign.erl
jur0: I will fix this. I can move signing of blocks into `aec_blocks` and the param will be a block, not a binary.

apps/aecore/src/aec_conductor.erl
```diff
@@ -177,7 +184,7 @@ handle_call({add_synced_block, Block},_From, State) ->
     {reply, Reply, State1};
 handle_call(get_block_candidate,_From, State) ->
     Res =
-        case State#state.block_candidate of
+        case State#state.micro_block_candidate of
```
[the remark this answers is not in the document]
sennui: true, only makes sense when we are a leader. New check would improve API

PR #1263, opened 2018-06-19
Contracts proof of inclusion HTTP endpoint

velzevur:
[HTTP Endpoint for contract PoI](https://www.pivotaltracker.com/story/show/158289180)
Endpoint providing proof of inclusion for a contract on the current chain top.

### DIFF
--- a/config/swagger.yaml
+++ b/config/swagger.yaml
@@ -778,6 +778,34 @@ paths:
           description: Invalid transaction
           schema:
             $ref: '#/definitions/Error'
+  /poi/contract/{contract}:
+    get:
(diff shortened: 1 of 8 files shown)


config/swagger.yaml
```diff
@@ -3122,7 +3150,14 @@ definitions:
             type: string
         required:
         - contract_address
-
+  PoI:
+    type: object
+    properties:
+      poi:
+        description: Proof of inclusion 
+        $ref: '#/definitions/EncodedHash'
```
lucafavatella: Maybe this type is misleading?

PR #1268, opened 2018-06-20
Update install and build docs to support Ubuntu 18.04

dincho:
https://www.pivotaltracker.com/story/show/158453189

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -1,42 +1,43 @@
 # Build from source
 
-This document describes how to build an epoch node from source on current Ubuntu 16.04.4 LTS.
+This document describes how to build an epoch node from source on current Ubuntu 16.04.4 LTS or Ubuntu 18.04 LTS.
 The commands below assume you are logged in with `sudo` user.
 
 ## Dependencies installation
 
-### Tools and libraries
+### Common tools and libraries
 
-Make sure your Ubuntu version and it's packages are up to date, then install required tools and libs:
+Make sure your Ubuntu version and it's packages are up to date, then install required tools and libraries:
 ```bash
 sudo apt-get -qq update \
 && sudo apt-get -y upgrade \
-&& sudo apt-get -qq -y install \
-git curl autoconf build-essential ncurses-dev libssl-dev
+&& sudo apt-get -qq -y install git curl autoconf build-essential ncurses-dev libssl-dev
 ```
 
 ### OTP install
 
-Required Erlang OTP version is `20.2.3` it could be installed from source running below commands:
+Required Erlang OTP version is `20.1`.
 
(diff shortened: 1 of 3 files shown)


docs/installation.md
```diff
@@ -11,14 +11,40 @@ In order to install an epoch node using a release binary, you need to:
...
 
-The Ubuntu 16.04 package requires a libsodium shared library (v1.0.16) in `/usr/local/lib/libsodium.so.23` (`wget https://download.libsodium.org/libsodium/releases/libsodium-1.0.16.tar.gz`, unpack, then do `./configure && make && sudo make install && sudo ldconfig`).
+The package requires a libsodium v1.0.16 as `libsodium.so.23` shared object/library.
+
+#### Ubuntu 18.04
+
+Ubuntu 18.04 ships with libsodium 1.0.16, thus it can be installed with `apt` package manager:
+
+```bash
+sudo apt-get install libsodium23
```
lucafavatella: I am considering how to reduce the testing effort for this user experience i.e. package built on 16.04 but run on 18.04. May I suggest to update the Dockerfile in this epoch repo so that, when releasing, the testing of the Docker image implicitly tests this? (The testing that the package built on 16.04 can be run on 16.04 is on testnet nodes.)

I see this PR does not change the Dockerfile in this epoch repo, that builds package on aeternity/builder - that is based on ubuntu:16.04 - and then actually prepares the image from ubuntu:16.04 .
Can you confirm that [the copying of `/usr/local/lib` from the builder](https://github.com/aeternity/epoch/blob/3019a2bf09e4da87fa8ae45e974984904ec02d2b/Dockerfile#L18) is only for [libsodium](https://github.com/aeternity/docker-builder/blob/a143797b693c31ba928b4c60390859e33b151225/Dockerfile#L22-L28)?

dincho:
Tracking as https://www.pivotaltracker.com/story/show/158490458

docs/installation.md
dincho: Please see 3821683a

docs/build.md
```diff
@@ -1,42 +1,43 @@
...
 && sudo apt-get -y upgrade \
-&& sudo apt-get -qq -y install \
-git curl autoconf build-essential ncurses-dev libssl-dev
+&& sudo apt-get -qq -y install git curl autoconf build-essential ncurses-dev libssl-dev
 ```
 
 ### OTP install
 
-Required Erlang OTP version is `20.2.3` it could be installed from source running below commands:
+Required Erlang OTP version is `20.1`.
```
hanssv: Didn't follow the discussion, but 20.1 is a step backwards?

dincho: @hanssv It's already what we require in [rebar.config](https://github.com/aeternity/epoch/blob/master/rebar.config#L19)

PR #1270, opened 2018-06-20
Pt 158185357 Sophia data decode endpoint - Required fields and status code

lucafavatella:


apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -547,10 +549,72 @@ test_encode_decode_sophia_data2(_Config) ->
...
+    %% Field valid according to schema but invalid for handler.
+    {ok, 400, #{<<"reason">> := <<"bad_type">>}} = get_contract_decode_data(#{'sophia-type' => <<"foo">>, data => D}),
+    {ok, 400, #{<<"reason">> := <<"bad argument">>}} = get_contract_decode_data(#{'sophia-type' => T, data => <<"foo">>}),
+    {ok, 200,
+     #{<<"data">> :=
+           #{<<"type">> := <<"tuple">>,
+             <<"value">> :=
+                 [#{<<"type">> := <<"string">>, <<"value">> := <<"bar">>},
+                  #{<<"type">> := <<"word">>  , <<"value">> := 192} ]}}
+    } = %% Why not `{ok, 400, #{<<"reason">> := <<"bad_type">>}}`, with potentially different reason?
```
lucafavatella: @happi This is unclear to me.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -547,10 +549,72 @@ test_encode_decode_sophia_data2(_Config) ->
...
+         get_contract_decode_data(#{'sophia-type' => <<"(string, (int))">>, data => D}),
+    {ok, 200,
+     #{<<"data">> :=
+           #{<<"type">> := <<"tuple">>,
+             <<"value">> :=
+                 [#{<<"type">> := <<"string">>, <<"value">> := <<"bar">>},
+                  #{<<"type">> := <<"word">>, <<"value">> := 192} ]}}
+    } = %% Why not `{ok, 400, #{<<"reason">> := <<"bad_type">>}}`, with potentially different reason?
```
lucafavatella: @happi This is unclear to me.

apps/aehttp/test/aehttp_integration_SUITE.erl
happi: The Sophia data (on the EVM) does not contain any type information so the decode will try to decode according to the given type spec which in some cases will succeed even though the spec is not matching the data since the data can be interpreted in many ways.

PR #1271, opened 2018-06-20
Handle contract store in POI.

happi:


apps/aecontract/src/aect_state_tree.erl
```diff
@@ -149,18 +149,67 @@ root_hash(#contract_tree{contracts = CtTree}) ->
...
+            {ok, aect_contracts:serialize_for_poi(aect_contracts:set_state(Store, Contract)), StorePoi}
+    end.
+
+add_store_keys_poi(_, '$end_of_table', _, Store, Poi, _) ->
+    {Store, Poi};
+add_store_keys_poi(Id, {PrefixedKey, Val, Iter}, PrefixSize, Store, Poi, CtTree) ->
+    case PrefixedKey of
+        <<Id:PrefixSize/binary, Key/binary>> = KeyId ->
+            Store1 = Store#{ Key => Val},
+            case aec_poi:add_poi(KeyId, CtTree, Poi) of
```
lucafavatella: I see this line is enriching the PoI proof DB for the contracts tree with the contract store item (key value). Ok.

apps/aecontract/src/aect_state_tree.erl
```diff
@@ -149,18 +149,67 @@ root_hash(#contract_tree{contracts = CtTree}) ->
                      {'ok', binary(), aec_poi:poi()}
                    | {'error', 'not_present' | 'wrong_root_hash'}.
 add_poi(Id, #contract_tree{contracts = CtTree}, Poi) ->
-    aec_poi:add_poi(Id, CtTree, Poi).
+    case aec_poi:add_poi(Id, CtTree, Poi) of
+        {ok, Contract, ContractPoi} ->
+            add_store_to_poi(
+              aect_contracts:deserialize(Id, Contract),
+              CtTree,
+              ContractPoi);
```
lucafavatella: I see this `ContractPoi` is the same variable that would have been returned earlier (as third element of the tuple). Ok.

lucafavatella:
I opened lighter PR #1272 

The ability to deserialize contract store from contract PoI is tracked as https://www.pivotaltracker.com/story/show/158489781

apps/aecontract/src/aect_state_tree.erl
lucafavatella: (with added items for store)

PR #1273, opened 2018-06-20
Release preparation

lucafavatella:
TODO:
* [x] Test locally the user documentation


lucafavatella:
@dincho I am considering that:
* docs/cuda-miner.md shall not be updated as it reports a tested configuration
* deployment/README.md does not need to be updated, as it documents deployment instructions for the DevOps scripts

PR #1275, opened 2018-06-21
Pt 158489781 contract store poi

happi:
Some more negative tests of state in POI would be nice.

### DIFF
--- a/apps/aecontract/src/aect_state_tree.erl
+++ b/apps/aecontract/src/aect_state_tree.erl
@@ -197,18 +197,10 @@ add_store_keys_poi(Id, {PrefixedKey, Val, Iter}, PrefixSize, Store, Poi, CtTree)
                         'ok' | {'error', term()}.
 verify_poi(Id, PoiSerializedContract, Poi) ->
     Contract = aect_contracts:deserialize_from_poi(Id, PoiSerializedContract),
-    SerializedContract = aect_contracts:serialize(Contract),
-    case aec_poi:verify(Id, SerializedContract, Poi) of
-        ok ->
-            Prefix = aect_contracts:store_id(Contract),
-            maps:fold(fun(K, V, ok) ->
-                              KeyId = <<Prefix/binary, K/binary>>,
-                              aec_poi:verify(KeyId, V, Poi);
-                         (_,_, {error, _} = E) -> E
-                      end,
-                      ok,
-                      aect_contracts:state(Contract));
-        {error, _} = E  -> E
+    case lookup_poi(Id, Poi) of
+        {ok, Contract} -> ok;
+        {ok, _} -> {error, bad_proof};
+        {error, _} = E -> E
     end.
 
 -spec lookup_poi(aect_contracts:id(), aec_poi:poi()) ->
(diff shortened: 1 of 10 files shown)


apps/aecontract/src/aect_state_tree.erl
```diff
@@ -149,21 +149,103 @@ root_hash(#contract_tree{contracts = CtTree}) ->
...
                         'ok' | {'error', term()}.
-verify_poi(Id, SerializedContract, Poi) ->
-    aec_poi:verify(Id, SerializedContract, Poi).
+verify_poi(Id, PoiSerializedContract, Poi) ->
+    Contract = aect_contracts:deserialize_from_poi(Id, PoiSerializedContract),
+    SerializedContract = aect_contracts:serialize(Contract),
+    case aec_poi:verify(Id, SerializedContract, Poi) of
+        ok ->
+            Prefix = aect_contracts:store_id(Contract),
+            maps:fold(fun(K, V, ok) ->
```
gorillainduction: This is not enough. This checks that the store from the `PoiSerializedContract` is included in the PoI, but it doesn't check that it is the whole store. I think you should construct the full store from the PoI and compare the stores instead.

apps/aecontract/src/aect_state_tree.erl
```diff
@@ -149,21 +149,103 @@ root_hash(#contract_tree{contracts = CtTree}) ->
...
+add_store_from_poi(Id, {PrefixedKey, Val, Iter}, PrefixSize, Store, Poi) ->
+    case PrefixedKey of
+        <<Id:PrefixSize/binary, Key/binary>> ->
+            Store1 = Store#{ Key => Val},
+            case aec_poi:iterator_next(Iter) of
+                {error, _} = E -> E;
+                Next ->
+                    add_store_from_poi(Id, Next, PrefixSize, Store1, Poi)
+            end;
+         _ -> {error, bad_proof}
```
gorillainduction: You should match hard on the clause above instead. The iterator guarantees that you always get the prefix.

apps/aecontract/src/aect_state_tree.erl
```diff
@@ -149,21 +149,92 @@ root_hash(#contract_tree{contracts = CtTree}) ->
...
+        {error, _} = E -> E;
+        {Store, StorePoi} ->
+            {ok, aect_contracts:serialize_for_poi(aect_contracts:set_state(Store, Contract)), StorePoi}
+    end.
+
+add_store_keys_poi(_, '$end_of_table', _, Store, Poi, _) ->
+    {Store, Poi};
+add_store_keys_poi(Id, {PrefixedKey, Val, Iter}, PrefixSize, Store, Poi, CtTree) ->
+    case PrefixedKey of
+        <<Id:PrefixSize/binary, Key/binary>> = KeyId ->
```
lucafavatella: I see that `Key` may get bound to empty binary. Wanted?

```
1> <<Id:1/binary, Key/binary>> = <<"1">>.
<<"1">>
2> Id.
<<"1">>
3> Key.
<<>>
```

PR #1279, opened 2018-06-22
Check size of keys when decoding

velzevur:
[Protect HTTP API from keys with wrong size](https://www.pivotaltracker.com/story/show/158383367)

### DIFF
--- a/apps/aecore/src/aec_base58c.erl
+++ b/apps/aecore/src/aec_base58c.erl
@@ -2,7 +2,8 @@
 
 -export([encode/2,
          decode/1,
-         safe_decode/2]).
+         safe_decode/2,
+         byte_size_for_type/1]).
 
 -type known_type() :: block_hash
                     | block_tx_hash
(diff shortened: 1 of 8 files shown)


apps/aecore/src/aec_base58c.erl
```diff
@@ -107,6 +123,11 @@ pfx2type(<<"nm">>) -> name;
 pfx2type(<<"st">>) -> state;
 pfx2type(<<"pi">>) -> poi.
 
+type2size(poi) -> not_set;
+type2size(name) -> not_set;
+type2size(transaction) -> not_set;
+type2size(_) -> 32.
```
lucafavatella: May I suggest explicitly listing each and every one of the types? This catch all may lead to surprising errors - especially if new types are introduced or size of any of the existing types changes.

lucafavatella: Is size of type `signature` 32? I did not find that.

velzevur:
Revisited key check and made it more assertive regarding new types.
Added integration tests for _shorter_ keys
Got a confirmation by @hanssv that signature size is 64 bytes.

velzevur:
Rebased on top of master

velzevur:
Rebased on top of NG changes

PR #1281, opened 2018-06-25
158591718 compact state db representation

gorillainduction:
Make a more compact representation for the state trees of a block. This becomes more important for NG, but it is good for the current system as well.

I add nothing new in Protocol for this, since it is a node local change, and is not visible in the communication protocol.

https://www.pivotaltracker.com/n/projects/2124891/stories/158591718

### DIFF
--- a/apps/aecore/src/aec_trees.erl
+++ b/apps/aecore/src/aec_trees.erl
@@ -27,6 +27,10 @@
          set_oracles/2
         ]).
 
+-export([ deserialize_from_db/1
+        , serialize_for_db/1
+        ]).
+
 -export([ensure_account/2]).
 
 -export([apply_txs_on_state_trees/4,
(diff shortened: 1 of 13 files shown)


apps/aecore/src/aec_trees.erl
```diff
@@ -204,38 +208,111 @@ contracts(Trees) ->
...
+
+db_serialization_template(?AEC_TREES_VERSION) ->
+    [ {contracts_hash,     [binary]}
+    , {calls_hash,         [binary]}
+    , {channels_hash,      [binary]}
+    , {ns_hash,            [binary]}
+    , {ns_cache_hash,      [binary]}
+    , {oracles_hash,       [binary]}
+    , {oracles_cache_hash, [binary]}
+    , {accounts_hash,      [binary]}
```
lucafavatella: I see (from 3 lines below, and from removal of hash padding) that it is list of binary as it could be one hash or none - in case tree empty. Ok.

PR #1285, opened 2018-06-26
Docker compose smoke tests

dincho:
This PR introduce smoke testing of docker-compose configuration by building the branch source code, setup 3 nodes (localnet) and perform basic health checks - responsible API endpoints and number of peers per node.

It also pushes the built image to DockerHub on master branch.

https://www.pivotaltracker.com/story/show/155124244

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -392,11 +392,11 @@ jobs:
       - run:
           name: Build latest docker image
           command: |
-            docker-compose build node1
+            IMAGE_TAG=${CIRCLE_BRANCH:?} docker-compose build node1
       - run:
           name: Create localnet with docker-compose
           command: |
-            sudo -u ubuntu -E -H docker-compose up -d
+            sudo -u ubuntu -E -H IMAGE_TAG=${CIRCLE_BRANCH:?} docker-compose up -d
       - run:
           name: Healthcheck
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -672,10 +672,11 @@ workflows:
             - test
             - eunit
             - aevm_tests
+            - uat_tests
             - static_analysis
           filters:
             branches:
-              only: 155124244_docker_compose_smoke
+              only: master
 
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -538,7 +551,7 @@ workflows:
             tags:
               only: /^v.*$/
 
-      - docker_image_build:
+      - docker_smoke_tests:
```
lucafavatella: I am not convinced on this job being run on all branches but master.

I see master used to exclude docker_image_build job from master as other job, building Docker image, was built https://github.com/aeternity/epoch/blob/859918480239f6b3c914061448415be947f996cf/.circleci/config.yml#L643

Now instead master has image built, but non-master branches receive extra testing. Maybe it shall be other way round? E.g.:
* Master: docker_push_dev
* Non-master: docker_image_build (only build)
* Master twice a day: docker_smoke_tests

?

dincho: I'm not sure I understand your concerns, the "extra testing" adds just 8s. The master should not receive direct changes, thus it sounds OK to have more testing in PRs than in master.

The job you're referring to is actually not a test job, but artifacts publishing job, the build is a side effect.

We do have some extra tests in master (uat_test and OSX), but only for slow jobs that are not feasible to run for each PR.

dincho: > We do have some extra tests in master (uat_test and OSX), but only for slow jobs that are not feasible to run for each PR.

(We can also run those in PR's for a cost for more CircleCI workers that needs to be provisioned.)

lucafavatella: > The master should not receive direct changes, thus it sounds OK to me to have more testing in PRs than in master.

@dincho  If PR A passes additional testing and concurrent PR B passes additional testing, then we cannot infer that master with merged A and B passes the additional testing. Hence I am considering additional testing (i.e. Docker compose smoke testing) on master (once a day?).

dincho: I add master builds too.

PR #1286, opened 2018-06-26
Don't preempt mining on new micro-block in same generation

hanssv:


apps/aecore/src/aec_conductor.erl
```diff
@@ -214,7 +217,7 @@ handle_call(reinit_chain, _From, State1) ->
     %% NOTE: ONLY FOR TEST
     ok = reinit_chain_state(),
     TopBlockHash = aec_chain:top_block_hash(),
-    State2 = State1#state{seen_top_block_hash = TopBlockHash},
+    State2 = State1#state{top_block_hash = TopBlockHash},
```
lucafavatella: I am not sure why not resetting top_key_block_hash too

apps/aecore/src/aec_conductor.erl
```diff
@@ -889,31 +906,22 @@ flush_candidate() ->
...
+        case Origin of
+            block_created when IsLeader ->
+                aec_block_generator:start_generation(),
+                start_micro_signing(State1);
+            block_received when not IsLeader ->
+                aec_block_generator:stop_generation(),
+                State1;
+            micro_block_created when IsLeader ->
+                start_micro_sleep(State1);
+            _Other ->
```
lucafavatella: I wonder if it were possible to avoid the open ending case clause i.e. explicitly mention all potential values, as this would avoid typos or refactorings to change semantics of this code.

PR #1287, opened 2018-06-27
Handle miner fees for NG

gorillainduction:
* Split the fees by generation
* Store fees in the block state
* Handle delay of miner rewards

There are certainly more optimizations that can be done here, and some more testing.

### DIFF
--- a/apps/aecore/src/aec_block_key_candidate.erl
+++ b/apps/aecore/src/aec_block_key_candidate.erl
@@ -6,47 +6,28 @@
 %%%=============================================================================
 -module(aec_block_key_candidate).
 
--export([ create/2
-        , create_with_state/4
+-export([ create/1
         ]).
 
 -ifdef(TEST).
 -export([adjust_target/2]).
 -endif.
 
--export_type([block_info/0]).
-
 -include("blocks.hrl").
 
--type fees_info() :: #{txs := non_neg_integer(),
-                       gas := non_neg_integer()}.
--opaque block_info() :: #{trees := aec_trees:trees(),
-                          fees := fees_info(),
-                          txs_tree := aec_txs_trees:txs_tree(),
-                          adj_chain := [aec_headers:header()]}.
-
 %% -- API functions ----------------------------------------------------------
--spec create(aec_blocks:block() | aec_blocks:block_header_hash(), fees_info()) ->
-        {ok, aec_blocks:block(), block_info()} | {error, term()}.
-create(BlockHash, FeesInfo) when is_binary(BlockHash) ->
+-spec create(aec_blocks:block() | aec_blocks:block_header_hash()) ->
+        {ok, aec_blocks:block()} | {error, term()}.
+create(BlockHash) when is_binary(BlockHash) ->
     case aec_chain:get_block(BlockHash) of
         {ok, Block} ->
-            int_create(BlockHash, Block, FeesInfo);
+            int_create(BlockHash, Block);
         error ->
             {error, block_not_found}
     end;
-create(Block, FeesInfo) ->
+create(Block) ->
     {ok, BlockHash} = aec_blocks:hash_internal_representation(Block),
-    int_create(BlockHash, Block, FeesInfo).
-
--spec create_with_state(aec_blocks:block(), aec_keys:pubkey(),
-                        aec_trees:trees(), fees_info()) ->
-        {aec_blocks:block(), aec_trees:trees()}.
-create_with_state(Block, Miner, Trees, FeesInfo) ->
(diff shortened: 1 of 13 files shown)


apps/aecore/src/aec_block_key_candidate.erl
```diff
@@ -6,47 +6,28 @@
...
--opaque block_info() :: #{trees := aec_trees:trees(),
-                          fees := fees_info(),
-                          txs_tree := aec_txs_trees:txs_tree(),
-                          adj_chain := [aec_headers:header()]}.
-
 %% -- API functions ----------------------------------------------------------
--spec create(aec_blocks:block() | aec_blocks:block_header_hash(), fees_info()) ->
-        {ok, aec_blocks:block(), block_info()} | {error, term()}.
-create(BlockHash, FeesInfo) when is_binary(BlockHash) ->
+-spec create(aec_blocks:block() | aec_blocks:block_header_hash()) ->
```
hanssv: This is legacy from pre-NG block creation it should be enough with the version that takes a hash... The other variant is only used in tests!??

apps/aecore/src/aec_block_key_candidate.erl
```diff
@@ -66,44 +47,38 @@ adjust_target(Block, AdjHeaders) ->
...
             Error
     end.
 
-int_create(PrevBlockHash, PrevBlock, FeesInfo, Trees, Miner, AdjChain) ->
-    {ok, Block, BlockInfo} = int_create_block(PrevBlockHash, PrevBlock, FeesInfo, Trees, Miner),
+int_create(PrevBlockHash, PrevBlock, Miner, AdjChain) ->
+    {ok, Trees} = aec_chain_state:calculate_keyblock_state(PrevBlockHash, Miner),
+    Block = int_create_block(PrevBlockHash, PrevBlock, Miner, Trees),
     case adjust_target(Block, AdjChain) of
-        {ok, AdjBlock} -> {ok, AdjBlock, BlockInfo#{ adj_chain => AdjChain }};
```
hanssv: Premature optimization - we will need that adjacent chain when implementing an optimized update for keyblocks...

Easy to re-introduce though.

apps/aecore/src/aec_block_key_candidate.erl
```diff
@@ -66,44 +47,38 @@ adjust_target(Block, AdjHeaders) ->
...
-            int_create(BlockHash, Block, FeesInfo, Trees, Miner, AdjChain);
+            int_create(BlockHash, Block, Miner, AdjChain);
         {error, _} = Error ->
             Error
     end.
 
-int_create(PrevBlockHash, PrevBlock, FeesInfo, Trees, Miner, AdjChain) ->
-    {ok, Block, BlockInfo} = int_create_block(PrevBlockHash, PrevBlock, FeesInfo, Trees, Miner),
+int_create(PrevBlockHash, PrevBlock, Miner, AdjChain) ->
+    {ok, Trees} = aec_chain_state:calculate_keyblock_state(PrevBlockHash, Miner),
```
hanssv: Aren't we doing the job twice here? We have already performed `aec_chain_state:insert_block` for `PrevBlock`?

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -734,6 +734,7 @@ fork_get_transaction() ->
     RecipientPubKey = <<42:32/unit:8>>,
     PresetAccounts = [{SenderPubKey, 100}],
     meck:expect(aec_genesis_block_settings, preset_accounts, 0, PresetAccounts),
+    meck:expect(aec_governance, miner_reward_delay, 0, 100),
```
hanssv: 😛

apps/aecore/src/aec_block_key_candidate.erl
sennui: yeah, that's good it's shared with aec chain state, tho, so easier to refactor

gorillainduction: You are misreading. The `calculate_keykblock_state` is calculating what the state is if there is a keyblock added on top of the hash. Should be clarified in the code, and possibly have a better name.

Anyway, the job is done once (here) and if the block is mined, it is performed again with the actual block.

apps/aecore/src/aec_block_key_candidate.erl
gorillainduction: Please reintroduce it if needed. Currently it is not used.

apps/aecore/test/aec_chain_tests.erl
gorillainduction: :man_shrugging:

PR #1290, opened 2018-06-27
NG: fix aesc_fsm_SUITE

fabiankrol:


apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -390,12 +383,12 @@ set_chid_if_undefined(R, Msg) ->
 receive_info(R, Msg, Debug) ->
     receive_from_fsm(info, R, Msg, ?LONG_TIMEOUT, Debug).
 
-receive_from_fsm(Tag, #{role := Role, fsm := Fsm}, Info, Timeout, Debug)
+receive_from_fsm(Tag, #{role := Role, fsm := Fsm} = R, Info, Timeout, Debug)
```
uwiger: Could you ensure that all calls to `receive_info/3` (and, by extension, to `receive_from_fsm/5`), expect an updated `I` or `R` map, and that the function indeed always returns a map? (from my reading, it will still sometimes return `ok`)

uwiger: Sorry, didn't see that it had been closed. :)

PR #1294, opened 2018-06-29
Pt 158665685 contract balance http

velzevur:
[HTTP API: Contract's balance](https://www.pivotaltracker.com/story/show/158665685)

### DIFF
--- a/apps/aechannel/test/aesc_fsm_SUITE.erl
+++ b/apps/aechannel/test/aesc_fsm_SUITE.erl
@@ -551,6 +551,7 @@ wait_for_signed_transaction_in_block(Node, SignedTx) ->
                 aecore_suite_utils:mine_key_blocks(aecore_suite_utils:node_name(Node), 2),
                 case rpc(Node, aec_chain, find_tx_location, [TxHash]) of
                     none -> erlang:error(tx_not_found);
+                    not_found -> erlang:error(tx_not_found);
                     mempool -> Try(Attempts - 1);
                     BlockHash when is_binary(BlockHash) -> ok
                 end
--- a/apps/aechannel/test/aesc_fsm_SUITE.erl
+++ b/apps/aechannel/test/aesc_fsm_SUITE.erl
@@ -453,8 +453,8 @@ prep_responder(#{pub := IPub, balance := IBal} = _Initiator, Node) ->
     NodeName = aecore_suite_utils:node_name(Node),
     #{ public := Pub, secret := Priv } = enacl:sign_keypair(),
     Amount = IBal div 10,
-    aecore_suite_utils:spend(NodeName, IPub, Pub, Amount),
-    mine_blocks(Node, 2), % one extra for good measure (TODO: ensure tx in block)
+    {ok, SignedTx} = aecore_suite_utils:spend(NodeName, IPub, Pub, Amount),
+    ok = wait_for_signed_transaction_in_block(Node, SignedTx),
     {ok, Amount} = rpc(Node, aehttp_logic, get_account_balance, [Pub]),
     #{role => responder,
       priv => Priv,
@@ -542,3 +542,17 @@ role(Pubkey, #{pub               := MyKey
         {_OtherKey, initiator} -> responder;
         {_OtherKey, responder} -> initiator
     end.
+
+wait_for_signed_transaction_in_block(Node, SignedTx) ->
+    TxHash = aetx_sign:hash(SignedTx),
+    MineTx =
+        fun Try(0) -> did_not_mine;
+            Try(Attempts) ->
+                aecore_suite_utils:mine_key_blocks(aecore_suite_utils:node_name(Node), 2),
+                case rpc(Node, aec_chain, find_tx_location, [TxHash]) of
+                    none -> erlang:error(tx_not_found);
+                    mempool -> Try(Attempts - 1);
+                    BlockHash when is_binary(BlockHash) -> ok
+                end
+        end,
+    ok = MineTx(5).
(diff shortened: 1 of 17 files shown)


config/swagger.yaml
```diff
@@ -2088,6 +2088,36 @@ paths:
...
+      operationId: 'GetContractBalance'
+      description: 'Get Contract balance'
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters:
+        - in: path
+          name: contract_address
+          description: Arguments in sophia
```
lucafavatella: ?

config/swagger.yaml
```diff
@@ -2088,6 +2088,36 @@ paths:
...
+      tags:
+        - external
+        - chain
+      operationId: 'GetContractBalance'
+      description: 'Get Contract balance'
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters:
```
lucafavatella: I guess requirement [at height and at hash](https://github.com/aeternity/epoch/blob/0bdbb40a4da982f6fab39ff9b3c38146e266b35c/config/swagger.yaml#L1288-L1298) will come? At hash at least. (Not sure how to interpret height in Bitcoin-NG.)

velzevur: Probably. Not requested yet

velzevur:
Refactored: now account balance HTTP endpoint returns balances both for normal accounts and for contract ones.

apps/aehttp/src/aehttp_helpers.erl
```diff
@@ -554,3 +580,62 @@ safe_binary_to_atom(Binary) when is_binary(Binary) ->
     catch
         error:badarg -> {error, non_existing}
     end.
+
+get_block_hash_optionally_by_hash_or_height(PutKey) ->
```
hanssv: Is this NG-safe? There are multiple blocks at the same height?!

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -436,28 +437,22 @@ handle_request('GetContractPoI', Req, _Context) ->
...
-                      case aehttp_logic:get_account_balance_at_hash(AccountPubkey, Hash) of
+    ParseFuns = [read_required_params([address]),
+                 base58_decode([{address, address, [account_pubkey,
+                                                    contract_pubkey]}]),
+                 get_block_hash_optionally_by_hash_or_height(block_hash),
+                 fun(_, #{address := Pubkey, block_hash := AtHash}) ->
+                      case aehttp_logic:get_account_balance_at_hash(Pubkey, AtHash) of
                           {error, account_not_found} ->
-                              {404, [], #{reason => <<"Account not found">>}};
+                              {error, {404, [], #{reason => <<"Account not found">>}}};
```
ThomasArts: must be some magic going on, but in your test you match on {ok, {404, ...}}... but that's the request function manipulating it, I guess.

apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -542,3 +542,17 @@ role(Pubkey, #{pub               := MyKey
...
         {_OtherKey, responder} -> initiator
     end.
+
+wait_for_signed_transaction_in_block(Node, SignedTx) ->
+    TxHash = aetx_sign:hash(SignedTx),
+    MineTx =
+        fun Try(0) -> did_not_mine;
+            Try(Attempts) ->
+                aecore_suite_utils:mine_key_blocks(aecore_suite_utils:node_name(Node), 2),
+                case rpc(Node, aec_chain, find_tx_location, [TxHash]) of
```
gorillainduction: NOTE: Behavior changed of 'aec_chain:find_tx_location' in new tx GC PR.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -267,17 +267,22 @@ get_hash_at_height(Height, State) when is_integer(Height), Height >= 0 ->
...
             end
     end.
 
-first_hash_in_main_chain([Node|Left], TopHash) ->
-    case hash_is_in_main_chain(hash(Node), TopHash) of
-        true  -> {ok, hash(Node)};
-        false -> first_hash_in_main_chain(Left, TopHash)
+keyblock_hash_in_main_chain([Node|Left], TopHash) ->
+    NodeFound =
+        case hash_is_in_main_chain(hash(Node), TopHash) of
```
gorillainduction: Change order of these. hash_is_in_main_chain is much more expensive.

gorillainduction: Also, leave a `TODO` that this can change when key blocks and micro blocks are indexed separately.

apps/aechannel/test/aesc_fsm_SUITE.erl
hanssv: Yes, there is now both 'none' and 'not_found' both should crash the test...

apps/aehttp/src/aehttp_dispatch_ext.erl
velzevur: `aehttp_helpers:process_request/2` runs a list of functions on top of the incoming Req and aggregates state in a map of its own.
If any function ever returns an `{error, CowboyResponse}` - the chain is broken (the rest functions are not being applied at all) and the `CowboyResponse` is sent to the client.

The `ok` in the test has nothing to do with this whatsoever - it just states that we've managed to make a request to the URL...

apps/aecore/src/aec_chain_state.erl
```diff
@@ -268,17 +268,24 @@ get_hash_at_height(Height, State) when is_integer(Height), Height >= 0 ->
...
             end
     end.
 
-first_hash_in_main_chain([Node|Left], TopHash) ->
-    case hash_is_in_main_chain(hash(Node), TopHash) of
-        true  -> {ok, hash(Node)};
-        false -> first_hash_in_main_chain(Left, TopHash)
+%% TODO: this function relies that both key blocks and micro blocks share the
+%% same index; this can change when they are indexed separately
+keyblock_hash_in_main_chain([Node|Left], TopHash) ->
```
gorillainduction: A bit nit-picky, but a suggestion:
```erlang
case is_key_block(Node) andalso hash_is_in_main_chain(hash(Node), TopHash) of
  true  -> {ok, hash(Node)};
  false -> keyblock_hash_in_main_chain(Left, TopHash)
end;
```

PR #1303, opened 2018-07-02
Cuckoo sync

dincho:
finishes https://www.pivotaltracker.com/story/show/154520242

It has been tested on AWS GPU instance.


dincho:
> Since it is a client visible feature (and probably a highly desired one) probably it is worth a line in the release notes?
> Cc @lucafavatella

Dropped a line in 1e168b78

PR #1308, opened 2018-07-04
Add basic channel delegate support

uwiger:
See [PT #15770430](https://www.pivotaltracker.com/story/show/158770430)

* Add support for a list of delegates (pubkeys) in the channel state
* Slash transactions can be posted by a registered delegate
* Other channel-related transactions (e.g. `close_solo`) cannot

No WebSocket API support for now.

### DIFF
--- a/apps/aechannel/src/aesc_channels.erl
+++ b/apps/aechannel/src/aesc_channels.erl
@@ -14,6 +14,7 @@
          is_solo_closing/2,
          new/1,
          peers/1,
+         delegates/1,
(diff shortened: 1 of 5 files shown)


uwiger:
@ThomasArts the list of delegates, for now, needs to be set in the `channel_create_tx` (but defaults to `[]`). There is currently no support for changing such parameters afterwards. The only change in the close/slash Erlang-side API for now is that an error message changed (`account_not_peer_or_delegate`.)

PR #1309, opened 2018-07-04
Adjust ping interval and fail instead of crash end_per_testcase

ThomasArts:
The system tests were not very realistic, because mining speed and ping interval are related, but only mining speed was decreased a lot in system test, without taking case of ping interval.

By looking for errors in mining.log the system test fails on present master, because there are errors that need to be fixed in block generation!

[finishes PT-158819238]

### DIFF
--- a/system_test/aest_commands_SUITE.erl
+++ b/system_test/aest_commands_SUITE.erl
@@ -45,10 +45,7 @@ init_per_testcase(_TC, Config) ->
     aest_nodes:ct_setup(Config).
 
 end_per_testcase(_TC, Config) ->
-    try aest_nodes:ct_cleanup(Config)
-    catch _:Reason ->
-        {fail, Reason}
-    end.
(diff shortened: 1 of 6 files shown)


system_test/aest_commands_SUITE.erl
```diff
@@ -45,8 +45,10 @@ init_per_testcase(_TC, Config) ->
     aest_nodes:ct_setup(Config).
 
 end_per_testcase(_TC, Config) ->
-    aest_nodes:ct_cleanup(Config).
-
+    try aest_nodes:ct_cleanup(Config)
```
eproxus: We could move the try catch into `ct_cleanup` since the function is intended to be run from Common Test anyway

ThomasArts: I was unsure if it was used differently. Will do

PR #1276, opened 2018-06-21
Pt 158452869 channel ws events deposit withdraw

velzevur:
[deposit/withdraw WS events](https://www.pivotaltracker.com/story/show/158452869)

Still TODO
- [x] Add a line in the release notes
- [x] Add description in the protocol

PR #1311, opened 2018-07-04
Fix intermittent failure in FSM SUITE

velzevur:
[Fix FSM_SUITE test failures](https://www.pivotaltracker.com/story/show/158821214)
#### Problem
A `create_channel_tx` is produced by FSM and ended up in the mempool. Then `minimal_depth + 1` key blocks are mined. If for race conditions the first generation does not contain a micro block that included the `create_channel_tx` - the minimal depth height is never reached and the test fails intermittently.

#### Solution
Ensure that the `create_channel_tx` is mined and then mine `minimal_depth` more blocks on top of it.

Note: Same issue is present with deposit and withdrawal, will address in the appropriate branch. Thus the commented pieces of code.

### DIFF
--- a/apps/aechannel/test/aesc_fsm_SUITE.erl
+++ b/apps/aechannel/test/aesc_fsm_SUITE.erl
@@ -192,6 +192,9 @@ deposit(Cfg) ->
     check_info(),
     {I1, _} = await_signing_request(deposit_tx, I),
     {_R1, _} = await_signing_request(deposit_created, R),
+    %SignedTx = await_on_chain_report(I1, ?TIMEOUT),
+    %SignedTx = await_on_chain_report(R, ?TIMEOUT), % same tx
+    %wait_for_signed_transaction_in_block(dev1, SignedTx),
     mine_blocks(dev1, 4),
     ct:log("I2 = ~p", [I1]),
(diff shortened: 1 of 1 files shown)


apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -192,6 +192,9 @@ deposit(Cfg) ->
     check_info(),
     {I1, _} = await_signing_request(deposit_tx, I),
     {_R1, _} = await_signing_request(deposit_created, R),
+    %SignedTx = await_on_chain_report(I1, ?TIMEOUT),
```
velzevur: Will uncomment in #1276

apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -212,6 +215,9 @@ withdraw(Cfg) ->
     check_info(),
     {I1, _} = await_signing_request(withdraw_tx, I),
     {_R1, _} = await_signing_request(withdraw_created, R),
+    %SignedTx = await_on_chain_report(I1, ?TIMEOUT),
```
velzevur: Will uncomment in #1276

PR #1314, opened 2018-07-04
Pt 158825246 update system tests to new api

ThomasArts:
System tests pass on new API


apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -795,9 +795,8 @@ contract_transactions(_Config) ->    % miner has an account
     PoI = aec_trees:deserialize_poi(PoIBin),
     {ok, ContractInPoI} = aec_trees:lookup_poi(contracts, ContractPubKey, PoI),
     {ok, Trees} = rpc(aec_chain, get_top_state, []),
-    ContractInTree = rpc(aect_state_tree, get_contract, [ContractPubKey,
```
hanssv: This change will make it harder to understand a failure as you would just get a badmatch - the code was written like this on purpose. An alternative would be to use EUnit asserts?!

PR #1319, opened 2018-07-05
Move signing with exposed private key to test utils

gorillainduction:
https://www.pivotaltracker.com/n/projects/2124891/stories/158725460


gorillainduction:
Rebased on master, and removed some changes that was accidentally committed to `aec_chain_state.erl`

PR #1320, opened 2018-07-05
PT-158762346 Sophia variant types

UlfNorell:
Adds support for non-recursive variant types to Sophia.

Unresolved question: how to handle variant types when creating call data and decoding return values? You need the type definitions for this.


UlfNorell:
Found some missing things... stay tuned

UlfNorell:
Ok, should be good to go now.

PR #1323, opened 2018-07-05
PT-158105016 Add beneficiary to key blocks

zp-sd:
Python UAT tests result:
```
----------------------------------------------------------------------
XML: nosetests.xml
----------------------------------------------------------------------
Ran 15 tests in 441.296s

OK
```

System tests result:
```
%%% aest_commands_SUITE ==> epoch_commands: OK
%%% aest_perf_SUITE ==> long_chain.startup_speed: OK
%%% aest_perf_SUITE ==> long_chain.sync_speed: OK
%%% aest_perf_SUITE ==> long_chain.stay_in_sync: OK
%%% aest_sync_SUITE ==> new_node_joins_network: OK
%%% aest_sync_SUITE ==> docker_keeps_data: OK
%%% aest_sync_SUITE ==> stop_and_continue_sync: OK
%%% aest_sync_SUITE ==> tx_pool_sync: OK
%%% aest_sync_SUITE ==> net_split_recovery: OK
%%% aest_sync_SUITE ==> quick_start_stop: OK
All 10 tests passed.
```


zp-sd:
@hanssv 
1) Regarding
```
This PR does not remove the static signing keys, is there a follow up ticket that takes care of introducing throw-away signing keys?
```
There is not, but I'd be happy to create one. It will be quite big change because:
* Our `UAT` tests still heavily use internal APIs, which uses signing keys for creating transactions.
* I tricked most of `eunit`, `CT` and `UAT` tests to mimic beneficiary pubkey, so that it is the same as miner's signing pubkey. Otherwise fixing test would be much more "fun".. (mimicking looks a bit fishy especially in `UAT`s where I created files with signing keys, but I didn't have any smarter idea).

2) I will make changes to the following:
* `PROTOCOL_VERSION`,
* `epoch_full.yaml`,
* `RELEASE_NOTES`,
thanks! 🙇

zp-sd:
@hanssv I think I have covered `PROTOCOL_VERSION`,`epoch_full.yaml` and`RELEASE_NOTES`.

I will squash commits when I have 2 approvals.

hanssv:
#1319 makes the testing update task slightly more feasible I think!

zp-sd:
@hanssv I created `PT-158863688 Remove static signing keys` (estimated initially to 2 points 😄 ).

apps/aeutils/test/data/epoch_full.yaml
```diff
@@ -94,6 +94,8 @@ mining:
     attempt_timeout: 3600000
     # Expected mine rate (milliseconds) between blocks. Used in governance.
     expected_mine_rate: 300000
+    # Public key of beneficiary account that will receive fees from mining on a node.
```
dincho: This description sounds better than the one in the schema file, could you please update the schema description to match this one?

apps/aeutils/test/data/epoch_testnet.yaml
```diff
@@ -10,3 +10,5 @@ chain:
 
 mining:
     autostart: true
+    # Public key of beneficiary account that will receive fees from mining on a node.
+    beneficiary: "encoded_beneficiary_pubkey_to_be_replaced"
```
dincho: This config should contain the bare minimum to join the testnet. Thus, I think the `beneficiary` should not be added in.

config/sys.config
```diff
@@ -27,6 +27,7 @@
   ]},
 
   {aecore, [
+      {beneficiary, <<"ak$tjnw1KcmnwfqXvhtGa9GRjanbHM3t6PmEWEWtNMM3ouvNKRu5">>},
```
dincho: I think that adding default public key is no go, as if users don't configure it they'll loose their rewards

apps/aeutils/test/data/epoch_testnet.yaml
zp-sd: @dincho BTW I made an assumption that, in order to force users to set it, encodable `beneficiary` must be present in either user config or sys config to start a node. Otherwise `aec_conductor` will fail (this is our "main" application process).

dincho: In case of mandatory `beneficiary` is makes sense.

zp-sd:
@lucafavatella To me this looks ready for re-review. I have put UAT and system tests results in PR description. 

/cc @gorillainduction @velzevur @dincho

apps/aecore/src/aec_headers.erl
```diff
@@ -145,6 +152,7 @@ serialize_to_binary(key, Header) ->
       (Header#header.prev_hash)/binary,
       (Header#header.root_hash)/binary,
       (Header#header.miner)/binary,
+      (Header#header.beneficiary)/binary,
```
lucafavatella: Please update https://github.com/aeternity/protocol/blob/7898261efc6b5ee8d6dc666b1eaf84a71fe2e540/serializations.md#key-block

config/dev.config
```diff
@@ -0,0 +1,105 @@
+[
```
lucafavatella: I see changes in this file compared to sys.config are:

```
@@ -23,16 +21,11 @@
                         {listen_address, <<"0.0.0.0">>}
                       ]}
           ]},
-      {enable_debug_endpoints, false} %% CAUTION: Debug endpoints may be inefficient
+      {enable_debug_endpoints, true} %% CAUTION: Debug endpoints may be inefficient
   ]},

-  {aecore, [
-      {peers, [
-        <<"aenode://pp$QU9CvhAQH56a2kA15tCnWPRJ2srMJW8ZmfbbFTAy7eG4o16Bf@52.10.46.160:3015">>, %% us-west-2
-        <<"aenode://pp$2vhFb3HtHd1S7ynbpbFnEdph1tnDXFSfu4NGtq46S2eM5HCdbC@18.195.109.60:3015">>, %% eu-central-1
-        <<"aenode://pp$27xmgQ4N1E3QwHyoutLtZsHW5DSW4zneQJ3CxT5JbUejxtFuAu@13.250.162.250:3015">>,%% ap-southeast-1
-        <<"aenode://pp$2i8N6XsjCGe1wkdMhDRs7t7xzijrjJDN4xA22RoNGCgt6ay9QB@31.13.249.70:3015">> %% eu-east
-      ]},
+  {aecore, [{beneficiary, <<"ak$tjnw1KcmnwfqXvhtGa9GRjanbHM3t6PmEWEWtNMM3ouvNKRu5">>},
+      {peers, []},
       {password, <<"secret">>},
       {db_path, "."},
       {persist, false},
```

Ok. (I do not like an extra file though I do not have a better idea without increasing complexity.)

apps/aecore/src/aec_block_genesis.erl
```diff
@@ -68,6 +68,8 @@ height() -> ?GENESIS_HEIGHT.
 
 miner() -> <<0:?MINER_PUB_BYTES/unit:8>>.
 
+beneficiary() -> <<0:?BENEFICIARY_PUB_BYTES/unit:8>>.
```
lucafavatella: Could you please update the reference to `aec_block_genesis:miner()` in [`aec_test_utils`](https://github.com/aeternity/epoch/pull/1323/files#diff-e14138de459cdd890333dfad3bd83f4cR72) even if commented out? I expect this exercise to lead to the removal of export of `aec_block_genesis:miner` and addition of export of `aec_block_genesis:beneficiary`

system_test/aest_commands_SUITE.erl
```diff
@@ -75,6 +75,11 @@ epoch_commands(Cfg) ->
     Result6 = aest_nodes:run_cmd_in_node_dir(node1, ["bin/epoch", "status"], Cfg),
     ?assertMatch({0, _}, Result6),
 
+    {0, Output7} = aest_nodes:run_cmd_in_node_dir(node1, ["bin/epoch", "keys_gen", "secret password"], Cfg),
+    ?assertMatch({match, _}, re:run(Output7, "Generated keypair with encoded pubkey:[\r\n]*")),
```
dincho: I'd suggest matching for hardcoded (known) output.

apps/aecore/src/aec_headers.erl
zp-sd: @lucafavatella PR https://github.com/aeternity/protocol/pull/140.

system_test/aest_commands_SUITE.erl
dincho: Ah my bad, sorry. It's random.

zp-sd:
@dincho 
* Deployment is not full (missing beneficiary in deployment/ansible/templates/epoch.yaml), thus failed to deploy on env/dev1 - FIXED
* Reading the documentation it becomes clear that users cannot run a Docker container before somehow obtain access to already deployed node to run the keys_gen. Of course they could just throw in some fake (but with valid format) public key in the config just for testing. This command can be run in the container itself with some additional pain, but it's not documented tho. - FIXED
* missing system_test in aest_commands_SUITE - FIXED
* keys_gen accepts empty password - FIXED
* keys_gen is missing .erl extension - this is run by escript, without module definition, so it cannot be a part of Erlang sources with erl extension (compilation would fail); it is to be used only by keys_gen.sh
* keys_gen is not in the src dir, but it's erl source - see above
* keys_gen don't exit with code != 0 on error - FIXED

@lucafavatella I think your comments have been addressed.

dincho:
I'm approving from configuration/deployment perspective. Conceptually I still disagree with the additional hassle that the mandatory beneficiary will introduce to users.

zp-sd:
`deploy_dev1` passed.

PR #1324, opened 2018-07-06
PT-158827776 Add first batch of new HTTP endpoints

jur0:
The PR adds endpoints:
```
/blocks/top
/key-blocks/*
/micro-blocks/*
/generations/*
```
We still keep the old endpoints. We plan to create follow up PRs for:
* adding of the rest of new API enpoints
* cleaning up old endpoints

### DIFF
--- a/apps/aehttp/src/aehttp_api_parser.erl
+++ b/apps/aehttp/src/aehttp_api_parser.erl
@@ -2,6 +2,7 @@
 
 -export([encode/2,
          encode/3,
+         encode_client_readable_key_block/2,
          encode_client_readable_block/2,
          decode/2]).
 
@@ -13,13 +14,18 @@
                                     fun(Pow) ->
                                         {ok, aec_headers:deserialize_pow_evidence(Pow)}
                                     end}},
+                     {<<"key_hash">>, block_hash},
                      {<<"txs_hash">>, block_tx_hash},
                      {<<"signature">>, signature},
-                     {<<"key_hash">>, block_hash},
                      {<<"miner">>, {fun(Val) -> aec_base58c:encode(account_pubkey, Val) end,
                                     fun decode_miner/1}}]).
 -define(OBJECTS, #{header_map => ?HEADER_OBJ,
                    block_map => [ {<<"transactions">>, {list, tx}} | ?HEADER_OBJ],
+                   block_hash =>  {fun(Block) ->
+                                       {ok, Hash} = aec_blocks:hash_internal_representation(Block),
+                                       aec_base58c:encode(block_hash, Hash)
+                                   end,
+                                   fun(_) -> {error, not_implemented} end},
                    block => {fun(Block) ->
                                 BMap = aehttp_logic:cleanup_genesis(
                                             aec_blocks:serialize_to_map(Block)),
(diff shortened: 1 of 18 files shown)


apps/aehttp/src/aehttp_api_parser.erl
```diff
@@ -42,10 +48,10 @@
                                     micro ->
                                         {ok, HMap0} = aec_headers:serialize_to_map(aec_blocks:to_header(Block)),
 %% currently the pending endpoint can return a pending microblock. The
```
velzevur: Since the `GetPendingBlock` is changed to `GetPendingKeyBlock` - the endpoint is no longer returning a micro block. We are safe to remove this comment and the check bellow for block signatures being `undefined`

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -59,7 +171,7 @@ handle_request('GetBlockLatest', Req, _Context) ->
 handle_request('GetBlockPending', Req, _Context) ->
```
velzevur: Since we have a `GetBlockPending` do we really need this anymore?

apps/aehttp/src/aehttp_api_parser.erl
velzevur: `GetBlockPending` is still around, if its removal will be a subject to another PR - so shall this comment and the workaround

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -59,7 +171,7 @@ handle_request('GetBlockLatest', Req, _Context) ->
...
     get_block(fun aehttp_logic:get_block_pending/0, Req, json, false);
 
-handle_request('GetKeyBlockByHeight', Req, _Context) ->
+handle_request('GetKeyBlockByHeightObsolete', Req, _Context) ->
```
velzevur: This is a duplicated endpoint. Will its removal be a subject to another PR? (I strongly suspect the path `/block/height/{height}` is being used by Python tests so its removal will require some changes there)

config/swagger.yaml
```diff
@@ -37,22 +37,355 @@ paths:
...
-          description: successful operation
+          description: 'Successful operation'
           schema:
             $ref: '#/definitions/Top'
       security:
+  /key-blocks/current:
+    get:
+      tags:
+        - external
+        - gossip
```
velzevur: As discussed - we need to address endpoint tags. Tracked in [Revisit swagger tags](https://www.pivotaltracker.com/story/show/158872647)

apps/aehttp/src/aehttp_dispatch_ext.erl
jur0: This will be removed in separate PR.

config/swagger.yaml
```diff
@@ -37,6 +37,22 @@ paths:
 ## External APIs definitions
 ###
 
+  /top:
```
fabiankrol: I'm reintroducing `/top` endpoint for full backward compatibility. It'll be removed with other legacy endpoints

fabiankrol: @jur0 @velzevur ^

velzevur: I see @jur0 had fixed the python UATs, do we still need it?

PR #1326, opened 2018-07-09
Reintroduce /top endpoint

fabiankrol:
fixes deploy_fast_integration, deploy_integration, docker_test_push_master CI jobs

### DIFF
--- a/apps/aehttp/priv/swagger.json
+++ b/apps/aehttp/priv/swagger.json
@@ -43,6 +43,23 @@
   } ],
   "schemes" : [ "http" ],
   "paths" : {
+    "/top" : {
+      "get" : {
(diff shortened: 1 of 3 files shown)


dincho:
> Wouldn't it be better to replace /top with /blocks/top in the places where it fails as we'll have to do it anyway?

In this case, also update the docs.

fabiankrol:
@dincho @jur0 I prefer to remove it with the rest of endpoints. This way we're able to give some transition time to possible clients.

dincho:
> @dincho @jur0 I prefer to remove it with the rest of endpoints. This way we're able to give some transition time to possible clients.

Indeed we have to release with both endpoints working, so that once we change the deploy scripts they work both with integration (master) and uat (latest tag).

PR #1327, opened 2018-07-09
AWS artifacts

dincho:
In scope of https://www.pivotaltracker.com/story/show/158550692

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -543,7 +543,7 @@ jobs:
       - DEPLOY_DOWNTIME: 1800 #15M
     steps: *deploy_steps
 
-  github_release_osx:
+  upload_packages_osx:
     <<: *container_config
     steps:
       - checkout
@@ -551,7 +551,7 @@ jobs:
           at: *packages_workspace
       - *upload_github_release_asset
 
-  github_release_linux:
+  upload_packages_linux:
     <<: *container_config
     steps:
       - checkout
@@ -746,7 +746,7 @@ workflows:
             tags:
               only: /^v.*$/
 
-      - github_release_linux:
+      - upload_packages_linux:
           requires:
             - eunit
             - static_analysis
@@ -758,7 +758,7 @@ workflows:
             tags:
               only: /^v.*$/
 
-      - github_release_osx:
+      - upload_packages_osx:
           requires:
             - eunit
             - static_analysis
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -111,14 +113,8 @@ references:
...
-      command: |
-        if [ "$(uname -s)" == "Darwin" ]; then
-            PKG_SUFFIX="osx-$(sw_vers -productVersion)"
-        elif [ "$(uname -s)" == "Linux" ]; then
-            PKG_SUFFIX="ubuntu-"`uname -m`
-        fi
-        export PACKAGE_TARBALL=${PACKAGES_DIR:?}/epoch-$(cat VERSION)-${PKG_SUFFIX}.tar.gz
-        echo "export PACKAGE_TARBALL=${PACKAGE_TARBALL}" >> $BASH_ENV
+      # construct and export PACKAGE_TARBALL environment variable available to all jobs
+      command: .circleci/set_package_path.sh
```
lucafavatella: The code moved to a separate file used to do two things:
1. export PACKAGE_TARBALL
2. append a string to a file

Now that the code is in separate (non-builtin) utility `.circleci/set_package_path.sh`, side effect 2 (append to file) still works while side effect 1 (export) does not: that [`export PACKAGE_TARBALL`](https://github.com/aeternity/epoch/pull/1327/files#diff-5f4850fe41b44cee2c6fcbcf61a7599eR16) is equivalent to using a non-exported variable. Unless the `.circleci/set_package_path.sh` is doing something I am not noticing that manages to change environment of caller?

There are two cases:
* A. export was needed, and its neutralization means there is a bug in this PR whose symptoms I am not sure about
  * If that is the case, `. .circleci/set_package_path.sh` shall be used (with no need for executable permission - I think - and better to re-consider `set -e` and shebang)
* B. export was not needed. In that case, I recommend to remove "export PACKAGE_TARBALL" from script (using plain variable) so to reduce complexity for next person reading script and wondering why there is export.

dincho: I guess it's B.

lucafavatella:
I am taking the liberty of labelling as WIP as I understand @dincho may add change.

.circleci/config.yml
dincho: `export` removed in efe3c33a

PR #1328, opened 2018-07-10
alternative way of setting release version from file

tsloughter:
Was just looking around and noticed the use of a `rebar.config.script` for this instead of relx's `cmd` option. I should probably make a better way in relx to do this that doesn't require the `tr -d ...`, but yea, just thought I'd mention it.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -49,7 +49,7 @@
 {plugins, [{swagger_endpoints, {git, "https://github.com/aeternity/swagger_endpoints", {tag, "0.2.1"}}}, {pc, "1.6.0"}]}.
 {swagger_endpoints, [{src, "config/swagger.yaml"}, {dst, "apps/aeutils/src/endpoints.erl"}]}.
 
-{relx, [{release, { epoch, "version value comes from VERSION" },
+{relx, [{release, { epoch, {cmd, "cat VERSION | tr -d '[:space:]'"}},
          % sasl is required for the command `epoch versions` to work,
          % it is disabled in `sys.config` though.
          [runtime_tools, sasl, lager, setup, sext, rocksdb, mnesia_rocksdb, gproc,
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -1,8 +1,0 @@
-Relx0 = lists:keyfind(relx, 1, CONFIG),
-{relx, [{release, {epoch, _}, Opts} | T]} = Relx0,
-{ok, VersionBin} = file:read_file(<<"VERSION">>),
-Version = string:trim(binary_to_list(VersionBin)),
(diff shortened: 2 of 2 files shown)


rebar.config.script
```diff
@@ -1,8 +0,0 @@
-Relx0 = lists:keyfind(relx, 1, CONFIG),
```
lucafavatella: I see references to this file in:
* Dockerfile
* .dockerignore

rebar.config
```diff
@@ -49,7 +49,7 @@
 {plugins, [{swagger_endpoints, {git, "https://github.com/aeternity/swagger_endpoints", {tag, "0.2.1"}}}, {pc, "1.6.0"}]}.
 {swagger_endpoints, [{src, "config/swagger.yaml"}, {dst, "apps/aeutils/src/endpoints.erl"}]}.
 
-{relx, [{release, { epoch, "version value comes from VERSION" },
+{relx, [{release, { epoch, {cmd, "cat VERSION | tr -d '[:space:]'"}},
```
lucafavatella: I see this is implemented at https://github.com/erlware/relx/blob/v3.24.4/src/rlx_config.erl#L347

@tsloughter Is that cmd option documented? I did not find it documented in the erlware/relx repo readme or wiki. (My concern is that if undocumented it may change in the future abruptly.)

rebar.config.script
lucafavatella: I removed these in https://github.com/aeternity/epoch/pull/1330 There I also started additional CI checks.

(BTW CI checks did not run for this PR - I tracked this as https://www.pivotaltracker.com/story/show/158938024 )

rebar.config
tsloughter: It isn't. I haven't really been doing relx docs separate from rebar3 so for now I'll just be adding it to the docs here http://www.rebar3.org/docs/releases today.

rebar.config.script
tsloughter: K, I'll get rid of rebar.config.script from dockerfile and dockerignore.

tsloughter: Oh, nevermind, just saw you already did that :). I'll get the docs up sometime this morning and comment in that PR.

lucafavatella:
I am taking liberty of labelling as WIP - did not see update in http://www.rebar3.org/docs/releases yet.

See also #1330

tsloughter:
I've updated the getting started section to include the options for release version, http://www.rebar3.org/docs/releases#section-getting-started

lucafavatella:
Closing in favour of #1330 (that is the same as this plus one small follow-up commit).

PR #1330, opened 2018-07-10
alternative way of setting release version from file - with adjustments

lucafavatella:
Supersedes #1328

TODO:
* [x] Ensure Python UAT tests, macOS package build, Docker-based system tests pass
  * [Auxiliary CI workflow](https://circleci.com/workflow-run/3626b62f-02f9-4981-bfc8-e2ad0c4907f6) on 9a7a0c99 that is equivalent to 5086adb3
    * Passed, though unrelated Docker-based system tests failure.
* [x] Clarify cmd option documentation. Refer to discussion in #1328

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,7 +3,6 @@
 !rebar.lock
 !Makefile
 !rebar3
-!rebar.config.script
 !VERSION
 !apps
 !.git
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM aeternity/builder as builder
 
 # Add required files to download and compile only the dependencies
-ADD rebar.config rebar.lock Makefile rebar3 rebar.config.script VERSION /app/
+ADD rebar.config rebar.lock Makefile rebar3 VERSION /app/
 RUN cd /app && make prod-compile-deps
 # Add the whole project and compile epoch itself.
 ADD . /app


tsloughter:
I've updated the getting started section to include the options for release version, http://www.rebar3.org/docs/releases#section-getting-started

lucafavatella:
> I've updated the getting started section to include the options for release version, http://www.rebar3.org/docs/releases#section-getting-started

@tsloughter Thanks!

lucafavatella:
I noticed that `{ epoch, {cmd, "cat VERSION | tr -d '[:space:]'"}}` introduces the regression that if VERSION file is missing then build does not abort. I attempted to replace with `{ epoch, {cmd, "V=$(cat VERSION | tr -d '[:space:]'); case \"$V\" in '') exit 1 ;; *) printf \"%b\" \"$V\" ;; esac"}}` but it appears that the exit status has no control of the outcome (I suspect because of `os:cmd`).

lucafavatella:
I tried replacing `"cat VERSION | tr -d '[:space:]'"` with `"cat VERSIONFOO | tr -d '[:space:]'"` i.e. simulating that the well-known name of VERSION file changes and this reference is not updated, and I got that tarball is created anyway.
```
===> Resolved epoch-cat: VERSIONFOO: No such file or directory

===> tarball /Users/luca/dev/aeternity/epoch/_build/prod/rel/epoch/epoch-cat: VERSIONFOO: No such file or directory
.tar.gz successfully created!
```

It may be caught by other layers (e.g. copy in CI, if no wildcard is used ?) those seems a relaxation of the guarantees. I am considering this.

tsloughter:
Thanks, that is good to know as that should be the expected behaviour. I'll be fixing this in relx.

PR #1331, opened 2018-07-10
Ignore log errors from aec_conductor:dispatch_worker

sylane:


system_test/helpers/aest_nodes.erl
```diff
@@ -53,10 +53,19 @@
 
 %% AWK script to keep only error, critical, alert and emergency log lines with
 %% all the extra lines following the log lines.
+%% FIXME: Temporarily ignore dispatch_worker errors, remove when fixed.
 -define(EPOCH_LOG_SCAN_AWK_SCRIPT, "
+    /^.*\\[error\\].*aec_conductor:dispatch_worker.*Disallowing dispatch of additional.*$/ {
```
lucafavatella: From https://21008-99802036-gh.circle-artifacts.com/0/home/circleci/epoch/system_test/logs/ct_run.epoch_ct@localhost.2018-07-10_12.28.43/extras.system_test.logs/run.2018-07-10_12.28.43/aest_sync_suite.new_node_joins_network.html on master I read:

> 2018-07-10 13:03:18.551 [error] <0.1270.0>@aec_conductor:dispatch_worker:394 Disallowing dispatch of additional create_key_block_candidate worker

ok

system_test/helpers/aest_nodes.erl
```diff
@@ -53,10 +53,19 @@
 
 %% AWK script to keep only error, critical, alert and emergency log lines with
 %% all the extra lines following the log lines.
+%% FIXME: Temporarily ignore dispatch_worker errors, remove when fixed.
```
lucafavatella: @sylane Before merging we shall ensure we at least track when to identify and address somehow log message.

system_test/helpers/aest_nodes.erl
```diff
@@ -53,10 +53,22 @@
...
+    /^.*\\[error\\].*aec_conductor:dispatch_worker.*Disallowing dispatch of additional.*$/ {
       matched = 1
-      buff = $0
+      if (buff != \"\") {
+        buff = \"\"
+      }
+    }
+    /^.*\\[(error|critical|alert|emergency)\\].*$/ {
+      if (!matched) {
+        matched = 1
```
lucafavatella: I see you moved this line from before the `if` to here. I guess it is because otherwise the "aec_conductor:dispatch_worker.*Disallowing dispatch of additional" string would match again hence causing `buff` to be not empty string but `$0`?

sylane: The `if` is added because there is now two pattern that matches the same line, this is so if it matches the first one it doesn't add the line to the buffer. With the `match = 1` before the if, it was a bug, no error lines would ever be found...

PR #1332, opened 2018-07-11
Pt 158586283 use id in txs

gorillainduction:
* Extend the usage of `aec_id:id()` to the interface in all transactions.
* Connect the `base58c` encoded information in the API to the creation of  transactions.
* Resolve name hashes while processing transactions at the correct tree height instead of in the API layer (resolves https://www.pivotaltracker.com/n/projects/2124891/stories/156600645)

Main pivotal tracker ticket: https://www.pivotaltracker.com/n/projects/2124891/stories/158586283

TODO:

- [x] Make more tests for spending to names
- [ ] ~~Implement querying an oracle by name~~ EDIT: This involves changing the subscription as well, so out of scope of this PR. 
- [x] Implement transferring a name to a named account
- [ ] ~~Add release notes~~ Nothing here is visible outside the node. I suggest adding release notes in follow up PRs.
- [ ] ~~Amend the protocol description with relevant API changes.~~ Again, should be in follow up PR.
- [ ] ~~Make tests for spending to different account types~~ Follow up PR.


lucafavatella:
Among TODO items I would add:
* Check whether spec/api doc changes in protocol repo necessary, and if so prepare new PR superseding https://github.com/aeternity/protocol/pull/91

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -414,9 +415,9 @@ handle_request('PostSpend', #{'SpendTx' := Req}, _Context) ->
                                        {recipient_pubkey, recipient},
                                         amount, fee, payload]),
                  read_optional_params([{ttl, ttl, '$no_value'}]),
-                 base58_decode([{sender, sender, account_pubkey},
-                                {recipient, recipient, account_pubkey}]),
-                 get_nonce(sender),
+                 base58_decode([{sender, sender, id_hash},
```
lucafavatella: For anyone reviewing this commit in isolation from the rest of this PR - this change seems to relax check on param though this is made again as tight in [this following commit](https://github.com/aeternity/epoch/pull/1332/commits/767265eec4b14bc21b8aeade1873e70f58ce3d8a#diff-453a86322d2489c390db7fde31659cceR417).

apps/aecore/src/aec_spend_tx.erl
```diff
@@ -82,15 +97,27 @@ nonce(#spend_tx{nonce = Nonce}) ->
...
 recipient(#spend_tx{recipient = Recipient}) ->
-    aec_id:specialize(Recipient, account).
+    Recipient.
+
+resolve_recipient(#spend_tx{recipient = Recipient}, Trees) ->
+    case aec_id:specialize(Recipient) of
+        {account,  RecipientPubkey} -> {ok, RecipientPubkey};
+        {oracle,   RecipientPubkey} -> {ok, RecipientPubkey};
+        {contract, RecipientPubkey} -> {ok, RecipientPubkey};
+        {name, NameHash} -> aens:resolve_from_hash(account_pubkey, NameHash, aec_trees:ns(Trees))
```
lucafavatella: Re-using review effort at [now superseded PR](https://github.com/aeternity/epoch/pull/1127/files#diff-89b8ec7c2f28f3e3c7d21ca9bf070bb7):
As in case of name this is not a pubkey but a hash, shall the Swagger param be renamed from [recipient_pubkey](https://github.com/aeternity/epoch/blob/5a72c277eb0f8619d3c7f8af5c94e63263db10df/config/swagger.yaml#L2471).

apps/aeoracle/src/aeo_query_tx.erl
```diff
@@ -80,17 +93,19 @@ response_ttl(#oracle_query_tx{response_ttl = ResponseTTL}) ->
...
-      oracle        := OraclePubKey,
+      oracle        := Oracle,
       query         := Query,
       query_fee     := QueryFee,
       query_ttl     := QueryTTL,
       response_ttl  := ResponseTTL,
       fee           := Fee} = Args) ->
-    Tx = #oracle_query_tx{sender        = aec_id:create(account, SenderPubKey),
+    account = aec_id:specialize_type(SenderAccount),
+    oracle  = aec_id:specialize_type(Oracle), %% TODO: Should also be 'name'
```
lucafavatella: When enabling oracle query target to be name by specifying its hash, Swagger param [`oracle_pubkey`](https://github.com/aeternity/epoch/blob/5a72c277eb0f8619d3c7f8af5c94e63263db10df/config/swagger.yaml#L2558) shall probably be renamed (as a name hash is not a pubkey).

apps/aens/src/aens_transfer_tx.erl
```diff
@@ -181,19 +193,22 @@ for_client(#ns_transfer_tx{nonce             = Nonce,
...
 %%% Getters
 %%%===================================================================
 
--spec recipient_account(tx()) -> aec_keys:pubkey().
-recipient_account(#ns_transfer_tx{recipient_account = Recipient}) ->
+recipient(#ns_transfer_tx{recipient_account = Recipient}) ->
+    Recipient.
+
+recipient_pubkey(#ns_transfer_tx{recipient_account = Recipient}) ->
+    %% TODO: Resolve name
```
lucafavatella: When enabling name transfer receiver to be name, the Swagger param [`recipient_pubkey`](https://github.com/aeternity/epoch/blob/5a72c277eb0f8619d3c7f8af5c94e63263db10df/config/swagger.yaml#L2727) shall probably be renamed as name is (?) expressed as hash hence not pubkey.

sennui:
The only thing that bothers me is around previous work in aec_id. We have `specialize/1`, `specialize/2` and `specialize_type/1`. Somehow it is hard to follow the pattern. Why not `specialize_tag/1` and `specialize_val/2` ?

``
-spec specialize(id()) -> {tag(), val()}.
specialize(#id{tag = Tag, val = Val}) ->
    {Tag, Val}.

-spec specialize(id(), tag()) -> val().
specialize(#id{tag = Tag, val = Val}, Tag) when ?IS_TAG(Tag), ?IS_VAL(Val) ->
    Val.

-spec specialize_type(id()) -> tag().
specialize_type(#id{tag = Tag}) when ?IS_TAG(Tag) ->
Tag.
``

PR #1344, opened 2018-07-15
Pt 157558948 contract chain test suite

rvirding:
This is the first proper version of the contract tests. It has been rebased on master a number of times and has incorporated all the necessary changes until now and works with them.

### DIFF
--- a/apps/aehttp/test/aehttp_contracts_SUITE.erl
+++ b/apps/aehttp/test/aehttp_contracts_SUITE.erl
@@ -28,7 +28,7 @@
 	 spending_3/1,
 	 identity_contract_1/1,
 	 identity_contract_2/1,
-	 identity_contract_3/1,
+	 simple_storage/1,
 	 dutch_auction_contract_1/1,
 	 dutch_auction_contract_2/1,
 	 null/1
@@ -38,6 +38,16 @@
 -define(NODE, dev1).
 -define(DEFAULT_TESTS_COUNT, 5).
 -define(WS, aehttp_ws_test_utils).
+
+%% Contract return values for some predefined numbers. Conversion
+%% functions will make this unnecessary.
+
+-define(VALUE_STATEP, <<"0x0000000000000000000000000000000000000000000000000000000000000020">>).
+-define(VALUE_21, <<"0x0000000000000000000000000000000000000000000000000000000000000015">>).
+-define(VALUE_42, <<"0x000000000000000000000000000000000000000000000000000000000000002A">>).
+-define(VALUE_84, <<"0x0000000000000000000000000000000000000000000000000000000000000054">>).
+-define(VALUE_126, <<"0x000000000000000000000000000000000000000000000000000000000000007E">>).
+-define(VALUE_168, <<"0x00000000000000000000000000000000000000000000000000000000000000A8">>).
 
 all() ->
     [
@@ -53,7 +63,7 @@ groups() ->
        spending_3,
        identity_contract_1,
        identity_contract_2,
-       identity_contract_3,
+       simple_storage,
        %% dutch_auction_contract_1,
        %% dutch_auction_contract_2,
        null					%This allows to end with ,
@@ -318,14 +328,14 @@ spending_3(Config) ->
     ok.
 
 %% identity_contract_1(Config)
-%%  Run the identity contract by miner.
+%%  Create the Identity contract by miner and call by miner.
 
 identity_contract_1(Config) ->
     NodeName = proplists:get_value(node_name, Config),
     %% Get miner balance.
     {ok, 200, _Btop} = get_balance_at_top(),
     {ok, 200, #{<<"pub_key">> := MinerAddress}} = get_miner_pub_key(),
-    {ok, MinerPubkey} = aec_base58c:safe_decode(account_pubkey, MinerAddress),
+    %% {ok,MinerPubkey} = aec_base58c:safe_decode(account_pubkey, MinerAddress),
 
     %% Compile contract "identity.aes"
     ContractString = aeso_test_utils:read_contract("identity"),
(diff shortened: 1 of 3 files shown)


apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -0,0 +1,1504 @@
...
+    ct:log("Events during TC: ~p", [[{N, aecore_suite_utils:all_events_since(N, Ts0)}
+                                     || {_,N} <- ?config(nodes, Config)]]),
+    ok.
+
+%% ============================================================
+%% Test cases
+%% ============================================================
+
+%% null(Config)
+%%  Does nothing an always succeeds.
```
velzevur: This is probably a typo (`an` -> `and`?)

apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -0,0 +1,1504 @@
...
+		 priv_key := APrivkey},
+      acc_b := #{pub_key := BPubkey}} = proplists:get_value(accounts, Config),
+
+    %% Check initial balances.
+    ABal0 = get_balance(APubkey),
+    BBal0 = get_balance(BPubkey),
+    ct:pal("Balances 0: ~p, ~p\n", [ABal0,BBal0]),
+
+    %% Add tokens to both accounts.
+    %% {ok,200,_} = post_spend_tx(APubkey, 500, 1),
```
velzevur: Why is this commented out?

apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -0,0 +1,1504 @@
...
+
+get_contract_call_object(TxHash) ->
+    Host = external_address(),
+    http_request(Host, get, "tx/"++binary_to_list(TxHash)++"/contract-call", []).
+
+get_contract_decode_data(Request) ->
+    Host = external_address(),
+    http_request(Host, post, "contract/decode-data", Request).
+
+%% get_spend(Data) ->
```
velzevur: Probably we don't need the commented out functions now?

apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -0,0 +1,1234 @@
...
+		 acc_b => #{pub_key => BPubkey,
+			    priv_key => BPrivkey,
+			    start_amt => StartAmt},
+		 acc_c => #{pub_key => CPubkey,
+			    priv_key => CPrivkey,
+			    start_amt => StartAmt},
+		 acc_d => #{pub_key => DPubkey,
+			    priv_key => DPrivkey,
+			    start_amt => StartAmt}},
+    [{accounts,Accounts},{node_name,NodeName}|Config];
```
lucafavatella: It seems that this init does not guarantee that Alice/Bert/Carl/Diana have StartAmt tokens, as it puts one spend tx per account, having a sender the mining node, in the mempool though without checking inclusion in block. (This is neither good nor bad on its own - it depends what test case assumes.)

apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -0,0 +1,1234 @@
...
+
+    %% Get balances after mining.
+    ABal1 = get_balance(APubkey),
+    BBal1 = get_balance(BPubkey),
+    ct:pal("Balances 1: ~p, ~p\n", [ABal1,BBal1]),
+
+    %% Transfer money from Alice to Bert.
+    TxHash = spend_tokens(APubkey, APrivkey, BPubkey, 200, 5),
+    MineUntil = fun () -> tx_in_chain(TxHash) end,
+    aecore_suite_utils:mine_blocks_until(NodeName, MineUntil, 10),
```
lucafavatella: I see this guarantees tx in chain.

apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -0,0 +1,1234 @@
...
+
+    %% Check initial balances.
+    ABal0 = get_balance(APubkey),
+    BBal0 = get_balance(BPubkey),
+    ct:pal("Balances 0: ~p, ~p\n", [ABal0,BBal0]),
+
+    %% Add tokens to both accounts.
+    {ok,200,_} = post_spend_tx(APubkey, 500, 1),
+    {ok,200,_} = post_spend_tx(BPubkey, 500, 1),
+    aecore_suite_utils:mine_blocks(NodeName, 2),
```
lucafavatella: It seems that this does not guarantee that the two spend tx at the lines just above are in the chain. (This is neither good nor bad on its own - it depends what the rest of the test case assumes.)

apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -0,0 +1,1234 @@
...
+    %% Check that tx has succeeded.
+    ?assert(tx_in_chain(TxHash)),
+
+    %% Check balances after sending.
+    ABal2 = get_balance(APubkey),
+    BBal2 = get_balance(BPubkey),
+    ct:pal("Balances 2: ~p, ~p\n", [ABal2,BBal2]),
+
+    %% Check that the balances are correct, don't forget the fee.
+    ABal2 = ABal1 - 200 - 5,
```
lucafavatella: This line is subject to intermittent test failure. This line assumes that `ABal1` reflects [the tx adding 500 tokens](https://github.com/aeternity/epoch/pull/1344/files#diff-c661915907572e4a612ffbdd1c98696eR183) though that is not guaranteed, as that tx may have been included only [while mining for the A->B transaction](https://github.com/aeternity/epoch/pull/1344/files#diff-c661915907572e4a612ffbdd1c98696eR195).

apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -0,0 +1,1234 @@
...
+
+    %% Mine 10 times to decrement value by 5000.
+    {ok,_} = aecore_suite_utils:mine_blocks(NodeName, 10),
+
+    %% Call the contract bid function by Bert.
+    BidValue = call_func(NodeName, BPubkey, BPrivkey, EncodedContractPubkey,
+			 HexCode, <<"bid">>, <<"()">>, #{amount => 100000}),
+    ct:pal("Bid value ~p\n", [BidValue]),
+
+    %% These values look reasonable.
```
lucafavatella: Could you please clarify whether the values of the balances are meant to be part of the passed/failed outcome of the test case? See also https://github.com/aeternity/epoch/wiki/Testing-general#debugging

rvirding: This should have been the same type of test as in spending_1 but the gas complicated comparing the values. This will be fixed.

lucafavatella:
I took the liberty of labelling as WIP as I understand this PR may receive some changes.

apps/aehttp/test/aehttp_contracts_SUITE.erl
rvirding: I have converted it to behave in exactly the same way is in the integration_SUITE.

apps/aehttp/test/aehttp_contracts_SUITE.erl
rvirding: This has been fixed with a mine_until.

apps/aehttp/test/aehttp_contracts_SUITE.erl
rvirding: There are now relevant tests with the balances checking that they have the correct values.

rvirding:
@lucafavatella I have fixed your issues and cleaned up the code and started working with maps. I will commit this so it is ready and you can decide when it is best to merge it. I will continue with it until then.

PR #1350, opened 2018-07-18
PT-159135120 Store docker localnet beneficiary keypair in repo

zp-sd:


docs/docker.md
```diff
@@ -133,6 +133,9 @@ docker run -d -p 3013:3013 \
 Small local network (not connected to testnet) can be created with `docker-compose`.
 It runs three nodes using the `mean16s-generic` miner (fastest generic miner) and a proxy server to allow CORS.
 
+All local network nodes have `ak$25MZX3BXYP32YDPGWsJqYZ6CgWnqD93VdpCYaTk6KsThEbeFJX` set as node beneficiary (for more details on beneficiary see [configuration documentation](configuration.md#beneficiary-account)).
+Public-private keypair of `ak$25MZX3BXYP32YDPGWsJqYZ6CgWnqD93VdpCYaTk6KsThEbeFJX` beneficiary can be found [here](https://github.com/aeternity/epoch/tree/master/docker/keys/beneficiary).
```
lucafavatella: I am a bit concerned of the pointer to master here. As we tag&release, and a user may fetch docker compose from tag, how does that link behave?

zp-sd: That always master I think. We may change it after tag & release? 🤔 I see no better solution..

lucafavatella: We refer to this document [only as online reference in tag](https://github.com/aeternity/epoch/blame/efe39cc79e7da76c6525e63b59b51cb976c028e8/docs/release-notes/RELEASE-NOTES-0.18.0.md#L44): please replace with local reference (`/docker/keys/beneficiary`?).

lucafavatella: > We may change it after tag & release?

I would prefer avoiding release burden. See my alternative solution.

dincho: Can we just publish the private key in the docs?

lucafavatella: > Can we just publish the private key in the docs?

That would be fine too. Or comment in config.

zp-sd: @lucafavatella @dincho Local reference works just fine, thanks!

PR #1354, opened 2018-07-18
Put Base58-encoded private key into docker docs

zp-sd:


docs/docker.md
```diff
@@ -135,6 +135,7 @@ It runs three nodes using the `mean16s-generic` miner (fastest generic miner) an
 
 All local network nodes have `ak$25MZX3BXYP32YDPGWsJqYZ6CgWnqD93VdpCYaTk6KsThEbeFJX` set as node beneficiary (for more details on beneficiary see [configuration documentation](configuration.md#beneficiary-account)).
 Public-private keypair of `ak$25MZX3BXYP32YDPGWsJqYZ6CgWnqD93VdpCYaTk6KsThEbeFJX` beneficiary can be found [here](/docker/keys/beneficiary).
+Base58-encoded form of private key, which matches above beneficiary public key: `6gLDkfCvx6emMtT7Fxnq9hZ4PModX8jfsmtHmJfxDc6eJADnun69DsW8H3pYDzWULopyhhhDmQxuAuYfUMmd2bZRQW85u`.
```
lucafavatella: base58? base58check? Maybe `ak$` prefix?

zp-sd: I don't think `ak$` prefix for private key is necessary. @dadaphl do you need it?

zp-sd: @lucafavatella Yes, changed to `Base58Check`, this is more accurate.

PR #1356, opened 2018-07-18
Update initial operational steps in user doc

lucafavatella:


docs/operation.md
```diff
@@ -86,7 +87,7 @@ $ curl http://31.13.249.70:3013/v2/top
 
 This is the hash of the block being at the top of the chain of the node and it should be same as the hash in `prev_hash` of the block you're currently mining:
 ```bash
-$ curl http://127.0.0.1:3013/v2/block/pending
+$ curl http://127.0.0.1:3013/v2/key-blocks/pending
 {...,"height":... ,"prev_hash":"bh$2UWBL9BciGC1w2FUukJZinchGRrCuwEuFTkcVvpZcfcpjiAbUy", ...}
 ```
 Height would be +1 of what is in the `/top` of the remote node but this is not
```
lucafavatella: @jur0 Does this statement still stand? (Considering that microblock may be returned by `/top`.)

jur0: @lucafavatella `/key-blocks/pending` returns a key block which is not on the chain (yet). `/top` returns the last block which is on the chain. The last block on chain can be:
* key block - the next key block (pending) has height + 1
* micro block - has the same height as the last key block, the next key block (pending) has height + 1

jur0: @lucafavatella so to sum up, yes, the statement still stands.

PR #1357, opened 2018-07-18
PT-159137219 CORS headers for 404 returns in nginx config in docker-compose

dadaphl:
This will set CORS headers also for requests resulting in a 404 in the docker-compose setup's proxy. This is necessary because errors and missing content usually returns a 404 from the API. Clients would not have access to the error messages.

### DIFF
--- a/docker/nginx-cors.conf
+++ b/docker/nginx-cors.conf
@@ -4,31 +4,31 @@
 #
 
 if ($request_method = 'OPTIONS') {
-    add_header 'Access-Control-Allow-Origin' '*';
-    add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS';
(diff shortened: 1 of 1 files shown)


docker/nginx-cors.conf
```diff
@@ -4,31 +4,31 @@
 #
 
 if ($request_method = 'OPTIONS') {
-    add_header 'Access-Control-Allow-Origin' '*';
-    add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS';
+    add_header 'Access-Control-Allow-Origin' '*' always;
```
lucafavatella: I see http://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header so seems fine.

PR #1367, opened 2018-07-20
Add export command to epoch

sylane:
* The command exports all the blocks to an Erlang disk_log.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -73,6 +73,7 @@
                    {copy, "apps/check_config/priv/extensions/check_config.sh", "bin/extensions/check_config"},
                    {copy, "apps/aeutils/priv/extensions/keys_gen.sh", "bin/extensions/keys_gen"},
                    {copy, "apps/aeutils/priv/extensions/get_peer_key.sh", "bin/extensions/peer_key"},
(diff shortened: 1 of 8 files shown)


rebar.config
```diff
@@ -84,7 +85,8 @@
         {extended_start_script_extensions, [
               {check_config, "extensions/check_config"},
               {keys_gen, "extensions/keys_gen"},
-              {peer_key, "extensions/peer_key"}
+              {peer_key, "extensions/peer_key"},
+              {export, "extensions/export_chain"}
```
lucafavatella: May I suggest it has a smoke test among [system tests](https://github.com/aeternity/epoch/blob/2c7db84032756f8c9eb28c7fcc934b04fc6fa590/system_test/aest_commands_SUITE.erl)?

sylane: @lucafavatella a smoke test is not simple without an import command. I would need to implement the import command in some way in the test. My idea was to add a test when implementing the import, but if you think it must have one I could add one. There would be two options, mocking the chain or doing a system test, exporting the chain and then reading the file in the test itself assuming its format.

lucafavatella:
Smoke test and rebase needed so I am taking liberty of marking as WIP.

PR #1373, opened 2018-07-23
Use infrastructure container to run Ansible playbooks

dincho:
In scope of https://www.pivotaltracker.com/story/show/158550692
Example run: https://circleci.com/workflow-run/91689ead-7e3b-4602-95a6-9b764ca484d3


deployment/DB_VERSION
```diff
@@ -0,0 +1 @@
+69
```
dincho: I shall make sure this is in sync with master before merge

.circleci/config.yml
```diff
@@ -157,24 +143,16 @@ references:
...
-          LIMIT="tag_role_epoch:&tag_env_${DEPLOY_ENV:?}"
-          if [ -n "$DEPLOY_COLOR" ]; then
-            LIMIT+=":&tag_color_${DEPLOY_COLOR}"
-          fi
-          cd deployment/ansible
-          ansible-playbook --limit="${LIMIT:?}" \
-            --extra-vars "package=${PACKAGE_TARBALL:?} hosts_group=tag_env_${DEPLOY_ENV:?} env=${DEPLOY_ENV:?} downtime=${DEPLOY_DOWNTIME:-0}" \
-            deploy.yml
+          export PACKAGE=${PACKAGE_TARBALL:?}
+          export DEPLOY_DB_VERSION=$(cat deployment/DB_VERSION)
```
lucafavatella: I see infrastructure repo [wants Ansible var db_version defined (potentially empty)](https://github.com/aeternity/infrastructure/blob/05f1a91d2499a2acf3c729db8a319f54b124aba0/ansible/deploy.yml#L61) though infrastructure repo [Makefile defaults undefined var `DEPLOY_DB_VERSION` as empty string](https://github.com/aeternity/infrastructure/blob/05f1a91d2499a2acf3c729db8a319f54b124aba0/Makefile#L32). This makes link between the two repos implicit and a bit fragile.

Would it be possible to strengthen infrastructure repo Makefile to check that the expected var is set-and-not-empty-string?

deployment/DB_VERSION
lucafavatella: 71 in master now.

.circleci/config.yml
```diff
@@ -157,24 +143,16 @@ references:
...
+          cd /infrastructure && make deploy
```
lucafavatella: I see https://github.com/aeternity/infrastructure/blob/05f1a91d2499a2acf3c729db8a319f54b124aba0/Dockerfile#L28
Ok.

.circleci/config.yml
dincho: I really see this kind of DB version as hack and prefer to keep the infrastructure DB version is.

The "client" of the infrastructure should be responsible to pass the version or not, moreover it should be able to pass an empty string meaning "I don't care for it"

deployment/DB_VERSION
dincho: sync in 2ea24f18

.circleci/config.yml
dincho: I realised that this is indeed dangerous, please see https://github.com/aeternity/infrastructure/pull/77

PR #1374, opened 2018-07-23
Add a system test to check mining power vs synchronized chain

sylane:
* Add a test that verify that when two cluster are synchronizing, the
chain from the cluster with the highest mining power wins.
 * Fix a bug in the peer pool for a corner case.

### DIFF
--- a/apps/aecore/src/aec_peers.erl
+++ b/apps/aecore/src/aec_peers.erl
@@ -864,7 +864,7 @@ on_peer_accepted(PeerAddr, PeerInfo, Pid, State0) ->
                                     "from ~s to ~s; rejecting connection",
                                     [ppp(PeerId), format_address(OldPeer),
                                      format_address({PeerAddr, Port})]),
-                    {error, trusted_address_changed}
+                    {{error, trusted_address_changed}, State}
             end
     end.
 
--- a/system_test/aest_sync_SUITE.erl
+++ b/system_test/aest_sync_SUITE.erl
@@ -16,7 +16,8 @@
     stop_and_continue_sync/1,
     tx_pool_sync/1,
     net_split_recovery/1,
-    quick_start_stop/1
+    quick_start_stop/1,
+    net_split_mining_power/1
 ]).
 
 -import(aest_nodes, [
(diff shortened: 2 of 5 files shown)


system_test/aest_sync_SUITE.erl
```diff
@@ -571,6 +678,17 @@ quick_start_stop(Cfg) ->
 
 %% helper functions
 
+node_mined_retries(Nodes) ->
+    Metric = "ae.epoch.aecore.mining.retries.value",
```
lucafavatella: Not sure about this metric:
https://github.com/aeternity/epoch/blob/2edd9fbe4e95bc010d57e043954d930ac7beba7f/apps/aecore/src/aec_conductor.erl#L693
https://github.com/aeternity/epoch/blob/2edd9fbe4e95bc010d57e043954d930ac7beba7f/apps/aecore/src/aec_conductor.erl#L698

lucafavatella: I see - nonce attempted without success. Thanks @sylane @dincho

system_test/aest_sync_SUITE.erl
```diff
@@ -560,6 +582,91 @@ net_split_recovery(Cfg) ->
...
+    end, [], Net2Nodes),
+
+    [N2A1 | N2As] = Net2BlocksA,
+    ?assertNotEqual(undefined, N2A1),
+    lists:foreach(fun(A) -> ?assertEqual(N2A1, A) end, N2As),
+
+    %% Check that the chains are different
+    ?assertNotEqual(N1A1, N2A1),
+
+    % Check that the 6 node cluster has more mining power.
```
lucafavatella: "Check that the 4 node cluster has more mining power than the 2 node cluster."?

system_test/aest_sync_SUITE.erl
```diff
@@ -560,6 +582,91 @@ net_split_recovery(Cfg) ->
...
+    Net1MinedBlocks1 = node_mined_retries(Net1Nodes),
+    Net2MinedBlocks1 = node_mined_retries(Net2Nodes),
+    ?assert(Net1MinedBlocks1 < Net2MinedBlocks1),
```
lucafavatella: In case this happens often (as it may happen for process scheduling etc...), we may declare test case skipped rather than failing. Ok for now.

system_test/aest_sync_SUITE.erl
```diff
@@ -560,6 +582,91 @@ net_split_recovery(Cfg) ->
...
+
+                [B1 | Bs] = BlocksB,
+                ?assertNotEqual(undefined, B1),
+                lists:foreach(fun(B) -> ?assertEqual(B1, B) end, Bs)
+            end),
+
+    {ok, 200, #{height := Top2}} = request(net1_node1, 'GetTop', #{}),
+    ct:log("Height reached ~p", [Top2]),
+
+    % Check that the 6 node cluster has still more mining power.
```
lucafavatella: "Check that the 4 node cluster has still more mining power than the 2 node cluster."?

apps/aecore/src/aec_peers.erl
```diff
@@ -864,7 +864,7 @@ on_peer_accepted(PeerAddr, PeerInfo, Pid, State0) ->
                                     "from ~s to ~s; rejecting connection",
                                     [ppp(PeerId), format_address(OldPeer),
                                      format_address({PeerAddr, Port})]),
-                    {error, trusted_address_changed}
+                    {{error, trusted_address_changed}, State}
```
lucafavatella: Not sure of impact of this hence whether release note line needed.

sylane: @lucafavatella this fixes a bug in a corner case where a known trusted peer would change its IP. It would crash without this fix.

PR #1385, opened 2018-07-26
Pt 157506439 create and call contracts off-chain

velzevur:
PT: [SC off-chain initialization and call of contract](https://www.pivotaltracker.com/story/show/157506439)
Initial version of contract's executution inside channels. Focing progress on-chain will be followed in a subsequent PR.

TODO:
- [x] Line in release notes
- [x] Bump DB version
- [x] Update protocol accordingly


lucafavatella:
Unit tests seen [failing](https://circleci.com/gh/aeternity/epoch/23322?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

Error [`%%% aehttp_contracts_SUITE ==> {{badmatch,{error,<<"argument_syntax_error">>}},` in test job](https://circleci.com/gh/aeternity/epoch/23320) is unrelated to this PR (known, being worked on).

apps/aecontract/src/aect_channel_contract.erl
```diff
@@ -0,0 +1,104 @@
...
+    ContractsTree  = aec_trees:contracts(Trees0),
+    Contract = aect_state_tree:get_contract(ContractPubKey, ContractsTree),
+    Owner = aect_contracts:owner(Contract),
+    Code = aect_contracts:code(Contract),
+    CallStack = [], %% TODO: should we have a call stack for create_tx also
+                    %% when creating a contract in a contract.
+    VmVersion = aect_contracts:vm_version(Contract),
+    CallDef = #{ caller     => Owner
+               , contract   => ContractPubKey
+               , gas        => 10000000
```
velzevur: There is a really tight balance there. This gas is never consumed from `onwer` account but yet - if the contract creation requires too much gas - we can still run out of gas nevertheless. This is a mechanism that protects the user that the contract is not too heavy to post on-chain. Same goes with gas amount in `run/7`

Probably I shall expose this to API? Shall it be optional?

apps/aecontract/src/aect_channel_contract.erl
```diff
@@ -0,0 +1,104 @@
...
+               , gas_price  => 1
```
velzevur: Gas is not used but we probably need it in order to verify the gas consumption does no make the contract non-postable on-chain because of high gas cost. Same goes with gas price in `run/7`

Probably I shall expose this to API? Shall it be optional as well?

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -3827,6 +3832,212 @@ sc_ws_withdraw_and_close(Config, Origin) when Origin =:= initiator
...
+            CallRes = GetCallResult(AckConnPid),
+            #{<<"caller_address">>    := CallerAddress,
+              <<"caller_nonce">>      := CallRound, 
+              <<"contract_address">>  := ContractAddress,
+              <<"gas_price">>         := _,
+              <<"gas_used">>          := _,
+              <<"height">>            := CallRound,
+              <<"return_type">>       := <<"ok">>,
+              <<"return_value">>      := ReturnValue} = CallRes,
+            {ok, 200, #{<<"data">> := #{<<"type">> := <<"word">>, <<"value">> := DecodedCallResult00}}} =
```
lucafavatella: This `<<"word">>` assumes which contracts this test helper function is being called with? Shall this be explicitly dependent on the test name instead?

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -3827,6 +3832,212 @@ sc_ws_withdraw_and_close(Config, Origin) when Origin =:= initiator
...
+                get_contract_decode_data(#{'sophia-type' => Type, data => ReturnValue}),
+            DecodedCallResult00
+      end,
+    % trigger call contract
+
+    contract_calls_(TestName, ContractPubKey, Code, SenderConnPid, UpdateVolley,
```
lucafavatella: Is the contract called only by the same party that created it?

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: Addressed

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: Yes, addressed

velzevur:
Enhanced unparsing of contract's call return value.
Enhanced tests to cover both participants being able to call a contract (it used to be tested just the owner)
Added a corresponding protocol section

velzevur:
Rebased on top of master, resolved conflict in release notes

PR #1372, opened 2018-07-23
PT-158827776 Add second batch of the new HTTP API endpoints

fabiankrol:
In this PR we're adding another batch of new endpoints. Following PRs are going to remove and clean up obsolete endpoints. 

TODO: 

- [x] docker smoke test #1363 d5a55f2f2d49b56909e04aa08ada0716232977f6
- [x] `aehttp_integration_SUITE ==> all.internal_endpoints.list_oracle_queries`
- [x] `aehttp_integration_SUITE ==> *.get_pending_key_block` 

TODO covered in following PRs:

- [ ] Improve test coverage for contracts, channels, names
- [ ] rename `/ng-transactions` to `/transactions`
- [ ] consult `/status` return with eapps 
- [ ] update release notes 
- [ ] update protocol repo

PR #1386, opened 2018-07-26
Correct type annotation of query field in oracle query

lucafavatella:


lucafavatella:
I prepared this while reviewing #1372

PR #1365, opened 2018-07-20
PT-159192048 Remove key_hash from micro blocks

zp-sd:
1) Protocol repo reference - https://github.com/aeternity/protocol/pull/150.

2) System tests:
```
===> Running Common Test suites...
%%% aest_commands_SUITE ==> epoch_commands: OK
%%% aest_peers_SUITE ==> test_peer_discovery: OK
%%% aest_peers_SUITE ==> test_inbound_limitation: OK
%%% aest_perf_SUITE ==> long_chain.startup_speed: OK
%%% aest_perf_SUITE ==> long_chain.startup_speed_mining: OK
%%% aest_perf_SUITE ==> long_chain.sync_speed: OK
%%% aest_perf_SUITE ==> long_chain.stay_in_sync: OK
%%% aest_sync_SUITE ==> new_node_joins_network: OK
%%% aest_sync_SUITE ==> docker_keeps_data: OK
%%% aest_sync_SUITE ==> stop_and_continue_sync: OK
%%% aest_sync_SUITE ==> tx_pool_sync: OK
%%% aest_sync_SUITE ==> net_split_recovery: OK
%%% aest_sync_SUITE ==> quick_start_stop: OK
Passed 13 tests.
```

3) UAT tests:
```
----------------------------------------------------------------------
XML: nosetests.xml
----------------------------------------------------------------------
Ran 15 tests in 645.184s

OK
```

PR #1393, opened 2018-07-31
PT-159432477 Update key_hash correctly when handling orphaned chain

zp-sd:
Scenario this fixes:

Assume chain:
`K1 -> M1 -> (...) -> K2 -> M3 -> K3`

`(...)` is to be replaced with `M2` to build a chain, and `M2` arrives with a delay.

Without this fix `K2`, `M3`, `K3` will have `K1` set as `key_hash` in the db, where `key_hash` should be `K2` for `K2` and `M3`, and `K3` for `K3`).

__

This is follow-up after https://github.com/aeternity/epoch/pull/1365.


apps/aecore/include/blocks.hrl
```diff
@@ -1,6 +1,6 @@
 -include("pow.hrl").
 
--define(PROTOCOL_VERSION, 19).
+-define(PROTOCOL_VERSION, 20).
```
lucafavatella: Already bumped since 0.18.0 so unneeded. Though does not harm.

Also this PR does not impact consensus, right? So only DB_VERSION bump needed. Though does not harm.

zp-sd: Ok, I was not sure what is the relation between protocol version and db bump, so did that just in case.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -424,7 +424,7 @@ assert_micro_signature(PrevNode, Node, KeyHash) ->
             ok
     end.
 
-%% Transitively compute new state trees iff
+%% Transitively compute new state trees if
```
sennui: why soften the condition?

PR #1394, opened 2018-07-31
Add funcionality for Channel PoI with accounts and contracts

velzevur:
PT: [SC WS: Get PoI](https://www.pivotaltracker.com/story/show/159406178)

### DIFF
--- a/apps/aehttp/src/sc_ws_handler.erl
+++ b/apps/aehttp/src/sc_ws_handler.erl
@@ -331,8 +331,9 @@ process_incoming(#{<<"action">> := <<"get">>,
     end;
 process_incoming(#{<<"action">> := <<"get">>,
                    <<"tag">> := <<"poi">>,
-                   <<"payload">> := #{<<"accounts">>  := AccountsE,
-                                      <<"contracts">>  := ContractsE}} = R, State) ->
+                   <<"payload">> := Filter} = R, State) ->
+    AccountsE   = maps:get(<<"accounts">>, Filter, []),
+    ContractsE  = maps:get(<<"contracts">>, Filter, []),
     Parse =
         fun(T, Keys) ->
             try {ok, lists:foldr(
@@ -344,6 +345,10 @@ process_incoming(#{<<"action">> := <<"get">>,
                 error:_ ->
                     {error, invalid_pubkey}
             end
+        end,
+    BrokenEncodingReply =
+        fun(What) ->
+            {reply, error_response(R, <<"broken_encoding: ", What/binary>>)}
         end,
     case {Parse(account_pubkey, AccountsE), Parse(contract_pubkey, ContractsE)} of
         {{ok, Accounts0}, {ok, Contracts0}} ->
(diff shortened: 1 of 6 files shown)


velzevur:
Will add a corresponding description in the protocol.

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -329,6 +329,38 @@ process_incoming(#{<<"action">> := <<"get">>,
...
+                {ok, PoI} ->
+                    Resp = #{action => <<"get">>,
+                             tag => <<"poi">>,
+                             payload => #{
+                                <<"poi">> => aec_base58c:encode(poi, aec_trees:serialize_poi(PoI))}},
+                    {reply, Resp};
+                {error, Reason} ->
+                    {reply, error_response(R, Reason)}
+            end;
+        _ -> {reply, error_response(R, broken_hexcode)}
```
lucafavatella: I suspect user would prefer a more specific error - i.e. on which of the two values the error is. Though seems good enough.

velzevur:
While improving tests, found an issue.

velzevur:
Improved error message for broken encoding to be more descriptive
Improved behaviour that keus `accounts` and `contracts` are no longer mandatorary in the payload
Fixed a small bug
Improved tests with negative paths

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -329,6 +329,45 @@ process_incoming(#{<<"action">> := <<"get">>,
...
+                              [Res | Acc]
+                      end, [], Keys)}
+            catch
+                error:_ ->
+                    {error, invalid_pubkey}
+            end
+        end,
+    BrokenEncodingReply =
+        fun(What) ->
+            {reply, error_response(R, <<"broken_encoding: ", What/binary>>)}
```
uwiger: I don't think the error responses from failed `safe_decode` are entirely consistent throughout this API. Perhaps something to track in a PT?

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -4764,9 +4764,57 @@ sc_ws_contract_(Config, TestName, Owner) ->
...
+                            }),
+
+                    {ok, #{<<"reason">> := R}} = ?WS:wait_for_channel_event(ConnPid, error),
+                    R
+                end,
+
+    EncodedPoI = GetPoI(SenderConnPid),
+    EncodedPoI = GetPoI(AckConnPid),
+
+    NegativePoiTests =
```
uwiger: Perhaps these should be broken out into a separate test? Could also be done some other time.

apps/aehttp/src/sc_ws_handler.erl
velzevur: Tracked [SC WS: revisit broken encoding responses](https://www.pivotaltracker.com/story/show/159459679)

apps/aehttp/test/aehttp_integration_SUITE.erl
velzevur: Tracked [SC WS Tests: revisit tests](https://www.pivotaltracker.com/story/show/159459719)

PR #1396, opened 2018-08-01
Fix encoding of query and response field in oracle api

fabiankrol:


apps/aecore/src/aec_base58c.erl
```diff
@@ -140,7 +137,9 @@ type2pfx(contract_pubkey)  -> <<"ct">>;
 type2pfx(transaction)      -> <<"tx">>;
 type2pfx(tx_hash)          -> <<"th">>;
 type2pfx(oracle_pubkey)    -> <<"ok">>;
-type2pfx(oracle_query_id)  -> <<"oq">>;
+type2pfx(oracle_query)     -> <<"oq">>;
```
ThomasArts: for backward compatibility, don't change the oq to oi tag... I cannot oversee whether it harms, but it may be troublesome for some hard coded tests. I see that the naming would be better, but anyhow.
I propose keep oq for oracle_query_id and make qv (query value) for oracle_query. Better proposals are accepted :-).

lucafavatella: I agree.

fabiankrol: To be honest that was my biggest concern in this pr.
I still prefer to change it and fix all the failing test than stuck for (a)eternity ;) especially while we're still pre-mainnet 

@lucafavatella @velzevur @jur0 any thoughts?

jur0: I'd check if there are any tests that use the prefix (it's unlikely). I'd rather change the test and keep the API clean.

ThomasArts: Because you like the first character to be refering to oracles, I guess? Because i for query_id is not really that logical. What about ov (oracle_query_value).

fabiankrol: 👍

PR #1398, opened 2018-08-01
Change PoW to 2^30 node graph.

dincho:
- [ ] update [Wiki](https://github.com/aeternity/epoch/wiki/Troubleshooting#node-failing-mining-attempts) after merge
- [x] trigger system tests run (EDIT: https://circleci.com/gh/aeternity/epoch/23960)

PT: https://www.pivotaltracker.com/story/show/154875588

### DIFF
--- a/docs/cuda-miner.md
+++ b/docs/cuda-miner.md
@@ -56,13 +56,13 @@ Compilation of CUDA miner is done by invoking:
 
 ```bash
 cd apps/aecuckoo && make c_src/.git
-cd c_src/src && make libblake2b.so cuda28
+cd c_src/src && make libblake2b.so cuda30
 ```
 
 Finally the actual installation of the miner binary is copying it to the node corresponding path, the documentation assumes the `epoch` node is installed in `~/node` directory.
 
 ```bash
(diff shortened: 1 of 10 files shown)


docs/cuda-miner.md
```diff
@@ -56,13 +56,13 @@ Compilation of CUDA miner is done by invoking:
 
 ```bash
 cd apps/aecuckoo && make c_src/.git
-cd c_src/src && make libblake2b.so cuda28
+cd c_src/src && make libblake2b.so cuda30
```
lucafavatella: Maybe tag `tags/v0.17.0` above shall be updated to `tags/v0.19.0` because of new cuckoo commit?

lucafavatella:
I am taking the liberty of marking as WIP while system tests on CI server running.

docs/cuda-miner.md
dincho: @lucafavatella I don't find it relevant to this PR/Commits. I can do it in separate one or in the release prep. as always.

Maybe we should start maintaining `tags/latest` ?

lucafavatella: > the release prep.

I will. Thanks.

PR #1399, opened 2018-08-01
Ease investigation of peer pool test failure

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/159461760


sylane:
Or completely remove this test in favor of a more appropriate QC test.

PR #1400, opened 2018-08-01
add note to release_notes

uwiger:
Note about race detection changes and added error_code field in some messages


lucafavatella:
No way this can impact CI. Merging.

PR #1401, opened 2018-08-01
Allow more than one suite to be specified

eproxus:
Usage example:
```
$ make test SUITE=apps/aecontract/test/aecontract,test/aevm_chain,apps/aecore/test/aecore_txs
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,15 @@ EUNIT_TEST_FLAGS ?=
 CT_TEST_FLAGS ?=
 ST_CT_FLAGS = --dir system_test --logdir system_test/logs
 
+null  :=
+space := $(null) # space
+comma := ,
+
+comma-separate = $(subst ${space},${comma},$(strip $1))
+space-separate = $(subst ${comma},${space},$(strip $1))
+
 ifdef SUITE
-CT_TEST_FLAGS += --suite=$(SUITE)_SUITE
+CT_TEST_FLAGS += --suite=$(call comma-separate,$(foreach suite,$(call space-separate,${SUITE}),${suite}_SUITE))
 unexport SUITE
 endif
 


lucafavatella:
I see strange docker_smoke_tests [failure on CI](https://circleci.com/gh/aeternity/epoch/24065):
```
Building node1
ERROR: Couldn't connect to Docker daemon at http+docker://localunixsocket - is it running?

If it's at a non-standard location, specify the URL with the DOCKER_HOST environment variable.
Exited with code 1
```

dincho:
> Not sure why CI failed again in this PR (cc @dincho @acidsennui ) though for being sure blocking this for now.

@eproxus could you please try changing the branch name (remove the `/`). I had docker related errors because of that.

Also I see that latest build of master passed the `docker_smoke_tests`

eproxus:
@dincho Thanks, #1405 is a test with a new name.

eproxus:
@dincho All checks passed on this _and_ the other test branch 🙁 Seems Circle CI is really flaky...

@lucafavatella Let the record show that tests have passed at least once ;-)

dincho:
> Seems Circle CI is really flaky...

I tried to debug it yesterday and indeed the issue was weird. CircleCI is relatively more stable than other services I've used in the past and on other projects.

dincho:
Still sometimes the status checks are updated by references regardless of the branch name. What I'm saying is that #1405 maybe updated both PRs status checks.

This is because we used the git branch name as docker image name where extra `/` is not OK.

eproxus:
@dincho So the test run and status for this PR's branch might actually be for the other branch?!

dincho:
> @dincho So the test run and status for this PR's branch might actually be for the other branch?!

Yes, AFAIK the status checks are based on commit SHAs

eproxus:
@dincho True, if it's the same SHA it's technically the same code being tested.

@lucafavatella I think this checks out okay now, merge at your discretion.

PR #1404, opened 2018-08-01
Dispatch refactor

eproxus:
Strictly a refactoring that SHOULD NOT change any behavior. Also removing an unrelated case statement that is not needed.


lucafavatella:
Based on experience in [other](https://github.com/aeternity/epoch/pull/1401) [PR](https://github.com/aeternity/epoch/pull/1405) I see that the [Docker test failure in this PR](https://circleci.com/gh/aeternity/epoch/24231) appears to be depending on the `/` in the branch name. So I ignore that failure.

PR #1405, opened 2018-08-02
Allow more than one suite to be specified

eproxus:
Temporary replacement for #1401 with new branch name


lucafavatella:
Merged as #1401

PR #1406, opened 2018-08-02
Plant release notes skeleton for next release

lucafavatella:


docs/cuda-miner.md
```diff
@@ -7,7 +7,7 @@ The release packages do not ship with a CUDA miner, but you can build it yoursel
 - Epoch node configuration
 
 The documentation below is tested on:
-- Epoch version 0.19.0
```
lucafavatella: @dincho I took the liberty of bumping this too.

lucafavatella:
Labelling as WIP as not meant to be merged yet - after release.

docs/cuda-miner.md
dincho: well, this will not be technically correct, because no one would test it on each release

velzevur: I assume CI tests are still tests. Code and documentation shall be corresponding to each other, so if we've missed a thingy - that would certainly be a bug. Iagree on the bumping.

dincho: True, but the CI does not test CUDA install and deployment, neither we deploy GPU miners on out testnet (yet?).

I see this line more like disclaimer, thus my comment.

PR #1408, opened 2018-08-06
Add support for state channels in system test

sylane:


system_test/aest_channels_SUITE.erl
```diff
@@ -0,0 +1,181 @@
...
+all() -> [
+    test_simple_same_node_channel,
+    test_simple_different_nodes_channel
+].
+
+init_per_suite(Config) ->
+    [
+        {node_startup_time, 20000}, %% Time may take to get the node to respond to http
+        {node_shutdown_time, 20000}, %% Time it may take to stop node cleanly
+        %% FIXME: Remove this when this is fixed:
```
eproxus: This FIXME could be made into a Pivotal task that is blocked by the referenced ticket.

system_test/helpers/aest_api.erl
```diff
@@ -0,0 +1,420 @@
+-module(aest_api).
```
eproxus: I like this API module. We should refactor the other suites to use it as well (and add to it).

system_test/aest_channels_SUITE.erl
sylane: I already added a comment on the task to change this when done.

PR #1410, opened 2018-08-06
Review stdout logs in VM and add further logs in CT tests

lucafavatella:
I prepared this while investigating failures in contract calling contract calling primops for https://www.pivotaltracker.com/story/show/158023291


lucafavatella:
Tracked as https://www.pivotaltracker.com/story/show/159690240

PR #1411, opened 2018-08-06
Use fixed seed in peer pool unit tests

sylane:


eproxus:
Which minimum Erlang version are we targeting? `exsplus` is backwards compatible to 19.0, but `exrop` is the default from 20.0 onwards.

sylane:
@ThomasArts do you mean a comment in the test or in the commit message (or both)?

dincho:
> Which minimum Erlang version are we targeting? exsplus is backwards compatible to 19.0, but exrop is the default from 20.0 onwards.

[20.1](https://github.com/aeternity/epoch/blob/master/rebar.config#L19)

apps/aecore/test/aec_peers_pool_tests.erl
```diff
@@ -33,16 +33,31 @@
...
+%% validate the bucket indexes cover enough of the bucket space.
+%% These tests may fail randomly for some seeds so fixing the seed make the
+%% test more reliable on CI.
 -define(USE_FIXED_SEED, true).
+
+-ifdef(USE_FIXED_SEED).
+
 -define(RANDOM_PROCESS_SEED, {1533,565972,58584}).
 -define(RANDOM_POOL_SEED, {1533,565990,602018}).
+-define(POOL_OPTS, [{disable_strong_random, true},
```
lucafavatella: It looks like all unit tests use this set of options. Though in actual node [no options are specified (taken from sys.config that has no options)](https://github.com/aeternity/epoch/blob/10c1b1c455e9a620766eb6cba9622d3e4bd23ef7/apps/aecore/src/aec_peers.erl#L375). So the pool setup used in actual node is now not unit tested - it appears to me.

May I suggest adding a very basic smoke test with no statistical checks (hence reproducible though basic) with no options passed to pool?

sylane: It is true that these options remove the testing of the stong random index... a tradeoff between predictability and coverage... I am open to any suggestion.

eproxus:
@sylane Since we're targeting 20.1+ I think the `exrop` algorithm should be used.

sylane:
@eproxus I already changed it to use `exrop`.

PR #1412, opened 2018-08-06
Ensure code paths are in place when running keys_gen script

lrascao:
It depends on enacl and base58, simply add all sibling libraries
to the escript's code path on start.

### DIFF
--- a/apps/aeutils/priv/keys_gen
+++ b/apps/aeutils/priv/keys_gen
@@ -8,6 +8,9 @@ main([[]]) ->
     io:fwrite("Usage: keys_gen PASSWORD~n"),
     halt(1);
 main([Password]) ->
+    % add all the sibling library `ebin` dirs to the code path
+    % in order to ensure runtime dependencies are met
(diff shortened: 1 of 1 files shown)


ThomasArts:
The keys_gen script is intended to run by using ./bin/epoch keys_gen PASSWD
in a production environment. I have just tried this for release 0.19.0 and that works perfectly well.

Unsure what you have been using that makes the path not correct?
Idea is: `make prod-package` and then deploy the generated tar.gz file somewhere. In that deployment you should be able to run ./bin/epoch with all provided commands.

lrascao:
You're right, it does run fine with the packaged release.
I came across this when generating the release (`rebar3 release`) and then `_build/default/rel/bin/epoch keys_gen`

lrascao:
changing the `include_erts` rebar.config entry to `true` makes it work, this points either to `relx` or OTP's release handler, i'll close this ticket and post any interesting findings regarding this here

PR #1417, opened 2018-08-07
Fix naming test failure + improve stability

hanssv:
Fix an intermittent test failure in naming test and improve stability of other tests by replacing the slightly error prone `mine_blocks_until` with `mine_blocks_until_tx_on_chain`. The latter will only check whole generations backwards from a key-block and should therefore be micro fork resistant (to some extent, with multiple miners the problem is still the same...)


apps/aecore/test/aecore_suite_utils.erl
```diff
@@ -198,46 +198,63 @@ mine_blocks(Node, NumBlocksToMine, MiningRate, Type) ->
...
+            case txs_not_on_chain(Node, Block, TxHashes) of
+                []        -> {ok, NewAcc};
+                TxHashes1 -> mine_blocks_until_txs_on_chain_loop(Node, TxHashes1, Max - 1, NewAcc)
             end;
         {error, _} = Error -> Error
     end.
 
+txs_not_on_chain(Node, Block, TxHashes) ->
+    {ok, BlockHash} = aec_blocks:hash_internal_representation(Block),
+    case rpc:call(Node, aec_chain, get_prev_generation, [BlockHash]) of
```
lucafavatella: Not sure why using RPC rather than using user API as in [other test](https://github.com/aeternity/epoch/blob/d90447069f135d08a5ef2ca0544c4aed148a3982/apps/aehttp/test/aehttp_integration_SUITE.erl#L6019-L6027) though seems good enough.

apps/aens/src/aens_claim_tx.erl
```diff
@@ -195,6 +195,7 @@ for_client(#ns_claim_tx{nonce     = Nonce,
                         ttl       = TTL,
                         account   = Account}) ->
     #{<<"vsn">>       => version(),
+      <<"data_schema">> => <<"NameClaimTxObject">>, % swagger schema name
```
velzevur: All of those are great catches. This corresponds to [swagger def](https://github.com/aeternity/epoch/blob/master/config/swagger.yaml#L3833)

apps/aecore/test/aecore_suite_utils.erl
hanssv: Everything in aecore_suite_utils uses RPC would be really strange to start using HTTP there?!

PR #1419, opened 2018-08-07
Plant partial specification of code style

lucafavatella:


lucafavatella:
> trim_trailing_whitespace = true. Not sure how EditorConfig is usually implemented, does it trim trailing whitespace for unchanged code?

I tried emacs and intellij idea. Both do. (I tested on aehttp_spec_SUITE.erl that has 2 lines with trailing whitespace.)

.editorconfig
```diff
@@ -0,0 +1,8 @@
+root = true
+
+[*.{erl,hrl}]
+indent_style = space
```
lucafavatella: @eproxus I tested indentation in aect_dispatch.erl (that has more than one line with tab as indentation) - amending one does not change all lines (tested emacs and intellij idea).

lucafavatella:
@eproxus This shall be ready for re-review now.

PR #1424, opened 2018-08-08
PT-159273652 Fix generation of fork_id in case of micro forks

zp-sd:
* New fork_id was not generated in case of micro fork taking over (same fork_id was used), which lead to illegal_micro_fork error


zp-sd:
Yes, it may be safer to bump DB version.

I will add short line in release notes as well.

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -605,6 +605,7 @@ forking_test_() ->
      , {"Test if hash is connected to genesis", fun fork_is_connected_to_genesis/0}
      , {"Test if hash is in main chain", fun fork_is_in_main_chain/0}
      , {"Get a transaction from the right fork", fun fork_get_transaction/0}
+     , {"Fork on micro-block", fun fork_on_micro_block/0}
```
lucafavatella: By commenting code changes in this PR this test fails with `illegal_micro_fork`. Good!

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -782,6 +783,50 @@ fork_get_transaction() ->
...
+                     Tx2 = aec_test_utils:sign_tx(make_spend_tx(PubKey, 2, PubKey, 1), PrivKey),
+                     [Tx1, Tx2];
+                (_) ->
+                     []
+             end,
+    [B0, B1, B2, _, _, _] =
+        Chain0 = gen_block_chain_with_state_by_target(
+                   PresetAccounts,
+                   [?HIGHEST_TARGET_SCI, ?HIGHEST_TARGET_SCI, ?HIGHEST_TARGET_SCI], 1, TxsFun),
+    [KB0, KB1, MB1, MB2, KB2, KB3] = blocks_only_chain(Chain0),
```
lucafavatella: Assertion that in this generated test date actually the KB* are key blocks and the MB* are micro blocks would help readability of test (e.g. not sure how microblocks are 2 and not 1 - with 2 txs).

apps/aecore/test/aec_test_utils.erl
```diff
@@ -359,18 +359,24 @@ next_block_with_state([{PB, PBS} | _] = Chain, Target, Time0, TxsFun, Nonce,
                       PubKey, PrivKey, BeneficiaryPubKey) ->
     Height = aec_blocks:height(PB) + 1,
     Txs = TxsFun(Height),
-    %% NG: if a block X used to have Txs, now put them in a microblock just before
-    %% the key-block at height X.
+    %% NG: if a block X used to have Txs, now put them in micro-blocks just before
+    %% the key-block at height X. Every transaction is put in separate micro-block.
```
lucafavatella: I see this constrain of one tx per microblock is a simplification of the test helper as it is what needed so far in test cases. Ok.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -851,7 +851,14 @@ db_node_has_sibling_blocks(Node) ->
     Height   = node_height(Node),
     PrevHash = prev_hash(Node),
     %% NOTE: Micro blocks have the same height.
-    length([1 || Header <- aec_db:find_headers_at_height(Height)
-                     ++ aec_db:find_headers_at_height(Height - 1),
+    %% For key blocks siblings at Height and Height - 1
+    %% and for micro blocks siblings at Height and Height + 1
+    OtherHeight =
+        case node_type(Node) of
```
lucafavatella: Good catch!

PR #1427, opened 2018-08-09
Pt 159648248 configure micro block cycle

ThomasArts:
Make micro_block_cycle configurable

And add a test to test the configurability, which now demonstrates that the NG implementation does not use the micro_block_cycle correctly!

### DIFF
--- a/apps/aecore/test/aecore_txs_SUITE.erl
+++ b/apps/aecore/test/aecore_txs_SUITE.erl
@@ -27,14 +27,24 @@ init_per_suite(Config) ->
     ok = application:ensure_started(erlexec),
     DataDir = ?config(data_dir, Config),
     TopDir = aecore_suite_utils:top_dir(DataDir),
+    MicroBlockCycle = 100,
     Config1 = [{symlink_name, "latest.txs"},
                {top_dir, TopDir},
-               {test_module, ?MODULE}] ++ Config,
+               {test_module, ?MODULE},
+               {micro_block_cycle, MicroBlockCycle}] ++ Config,
     aecore_suite_utils:make_shortcut(Config1),
     ct:log("Environment = ~p", [[{args, init:get_arguments()},
                                  {node, node()},
                                  {cookie, erlang:get_cookie()}]]),
-    aecore_suite_utils:create_configs(Config1),
+    DefCfg = #{
+        <<"chain">> => #{
+            <<"persist">> => false
+        },
+        <<"mining">> => #{
+            <<"micro_block_cycle">> => MicroBlockCycle
+        }
+    },
+    aecore_suite_utils:create_configs(Config1, DefCfg),
     aecore_suite_utils:make_multi(Config1),
     [{nodes, [aecore_suite_utils:node_tuple(dev1)]} | Config1].
 
@@ -120,7 +130,7 @@ txs_gc(Config) ->
 
 
 micro_block_cycle(Config) ->
-    MBC = aec_governance:micro_block_cycle(),
+    MBC = ?config(micro_block_cycle, Config),
     aecore_suite_utils:start_node(dev1, Config),
     N1 = aecore_suite_utils:node_name(dev1),
     aecore_suite_utils:connect(N1),
@@ -131,10 +141,12 @@ micro_block_cycle(Config) ->
     StartRes = rpc:call(N1, aec_conductor, start_mining, [], 5000),
     ct:log("aec_conductor:start_mining() (~p) -> ~p", [N1, StartRes]),
 
+    timer:sleep(1000), %% Make sure we're leader
+
     [ begin
         add_spend_tx(N1, 1000, 1,  Nonce, 10000),
-        timer:sleep(MBC div 100)
-      end || Nonce <- lists:seq(1,20) ],
+        timer:sleep(MBC div 3)
+      end || Nonce <- lists:seq(1,30) ],
 
     timer:sleep(2*MBC),
     StopRes = rpc:call(N1, aec_conductor, stop_mining, [], 5000),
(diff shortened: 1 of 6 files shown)


apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -117,6 +118,34 @@ txs_gc(Config) ->
...
+    [ begin
+        add_spend_tx(N1, 1000, 1,  Nonce, 10000),
+        timer:sleep(MBC div 100)
+      end || Nonce <- lists:seq(1,20) ],
+
+    timer:sleep(2*MBC),
+    StopRes = rpc:call(N1, aec_conductor, stop_mining, [], 5000),
+    ct:log("aec_conductor:stop_mining() (~p) -> ~p", [N1, StopRes]),
+
+    MicroBlocks = aecore_suite_utils:events_since(N1, micro_block_created, ?config(tc_start, Config)),
```
lucafavatella: Event delivery seems keeping order of publishing, and proxy performs appends to the dict entry for that event (hence keeping order). So seems fine.

ThomasArts: I have one moment been considering to sort the list but found it unnecessary.

apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -126,6 +155,22 @@ add_spend_tx(Node, Amount, Fee, Nonce, TTL) ->
     STx = aec_test_utils:sign_tx(Tx, maps:get(privkey, patron())),
     rpc:call(Node, aec_tx_pool, push, [STx]).
 
+timediff(Ms, []) ->
+  ok;
+timediff(Ms, [_]) ->
+  ok;
+timediff(Ms, [{T1, H1}, {T2, H1} | Rest]) ->
```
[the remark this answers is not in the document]
sennui: 👍

apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -117,6 +118,34 @@ txs_gc(Config) ->
...
+        timer:sleep(MBC div 100)
+      end || Nonce <- lists:seq(1,20) ],
+
+    timer:sleep(2*MBC),
+    StopRes = rpc:call(N1, aec_conductor, stop_mining, [], 5000),
+    ct:log("aec_conductor:stop_mining() (~p) -> ~p", [N1, StopRes]),
+
+    MicroBlocks = aecore_suite_utils:events_since(N1, micro_block_created, ?config(tc_start, Config)),
+    %% Below fails until micro_block_cycle is correctly implemented
+    %% ok = timediff(MBC, [ {aec_blocks:time_in_msecs(B), aec_blocks:height(B)} || #{info := B} <- MicroBlocks ]),
```
zp-sd: 👍

hanssv: Working on that...

apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -117,6 +118,34 @@ txs_gc(Config) ->
 
     ok = aecore_suite_utils:check_for_logs([dev1], Config).
 
+
+micro_block_cycle(Config) ->
+    MBC = aec_governance:micro_block_cycle(),
```
hanssv: Since you don't this value I assume it is the default 3 seconds? With fast mining we have (more than) 1 key block per second, how is this ever going to test the micro_block_cycle time?

apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -117,6 +118,34 @@ txs_gc(Config) ->
...
+
+    %% Mine a block to get some funds. Height=1
+    aecore_suite_utils:mine_key_blocks(N1, 1),
+
+    StartRes = rpc:call(N1, aec_conductor, start_mining, [], 5000),
+    ct:log("aec_conductor:start_mining() (~p) -> ~p", [N1, StartRes]),
+
+    [ begin
+        add_spend_tx(N1, 1000, 1,  Nonce, 10000),
+        timer:sleep(MBC div 100)
```
hanssv: So you sleep for a 1/100th of the micro block sleep time, with 20 Txs, will that ever span over more than 1 key block? I would suggest 1/2 of the micro block sleep time or something?!

apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -117,6 +118,34 @@ txs_gc(Config) ->
...
+      end || Nonce <- lists:seq(1,20) ],
```
hanssv: Nonces 1-5 are already used by the previous test??

apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -117,6 +118,34 @@ txs_gc(Config) ->
...
 
+
+micro_block_cycle(Config) ->
+    MBC = aec_governance:micro_block_cycle(),
+    aecore_suite_utils:start_node(dev1, Config),
+    N1 = aecore_suite_utils:node_name(dev1),
+    aecore_suite_utils:connect(N1),
+
+    %% Mine a block to get some funds. Height=1
+    aecore_suite_utils:mine_key_blocks(N1, 1),
```
hanssv: This tests runs on top of the previous test, thus this is redundant.

apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -126,6 +155,22 @@ add_spend_tx(Node, Amount, Fee, Nonce, TTL) ->
     STx = aec_test_utils:sign_tx(Tx, maps:get(privkey, patron())),
     rpc:call(Node, aec_tx_pool, push, [STx]).
 
+timediff(Ms, []) ->
+  ok;
+timediff(Ms, [_]) ->
+  ok;
+timediff(Ms, [{T1, H1}, {T2, H1} | Rest]) ->
+  Delta = Ms div 10,  %% 10% off is fine
```
hanssv: Why would 10% off be ok, this is exact science :-)

PR #1428, opened 2018-08-10
Pluggable PoW consensus engines

lrascao:
To facilitate research/experimentation

### DIFF
--- a/apps/aecore/src/aec_pow.erl
+++ b/apps/aecore/src/aec_pow.erl
@@ -10,6 +10,7 @@
          pick_nonce/0,
          next_nonce/1,
          target_to_difficulty/1,
+         engine/0,
 
          scientific_to_integer/1,
          integer_to_scientific/1]).
@@ -22,11 +23,15 @@
 
 %% 10^24, approx. 2^80
 -define(NONCE_RANGE, 1000000000000000000000000).
--define(POW_MODULE, aec_pow_cuckoo).
+-define(DEFAULT_POW_ENGINE, aec_pow_cuckoo).
 
 %% 0..?MAX_NONCE
 -type nonce() :: 0..16#ffffffffffffffff.
 -export_type([nonce/0]).
+
+-type engine() :: aec_pow_cuckoo |
+                  aec_pow_test.
+-export_type([engine/0]).
 
 %%------------------------------------------------------------------------------
 %%                      Target threshold and difficulty
@@ -131,6 +136,12 @@ pick_nonce() ->
 -spec next_nonce(aec_pow:nonce()) -> aec_pow:nonce().
 next_nonce(N) ->
     (N + 1) band ?MAX_NONCE.
+
+-spec engine() -> aec_pow:engine().
+engine() ->
+    aeu_env:user_config_or_env([<<"mining">>, <<"engine">>],
+                               aecore, engine,
+                               ?DEFAULT_POW_ENGINE).
 
 %%------------------------------------------------------------------------------
 %% Test if binary is under the target threshold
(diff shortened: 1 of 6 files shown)


dincho:
@lrascao by default config @hanssv probably meant: `apps/aeutils/test/data/epoch_full.yaml`

apps/aecore/src/aec_pow.erl
```diff
@@ -132,6 +137,12 @@ pick_nonce() ->
 next_nonce(N) ->
     (N + 1) band ?MAX_NONCE.
 
+-spec engine() -> aec_pow:engine().
+engine() ->
+    aeu_env:user_config_or_env([<<"mining">>, <<"engine">>],
```
dincho: The configuration will return a string that should be converted to atom?

apps/aecore/src/aec_pow_test.erl
```diff
@@ -0,0 +1,42 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2017, Aeternity Anstalt
+%%% @doc
+%%%    A library providing PoW generation and verification for testing purposes
+%%%    The solution is always 42.
+%%%
+%%% @end
+%%%-------------------------------------------------------------------
+-module(aec_pow_test).
```
ThomasArts: Test files should be placed in test directory. If this file is to be used "in production", then the "_test" suffix is not well chosen.

dincho: @ThomasArts I think the meaning of this file is `pow test engine` the same meaning as `pow cyckoo engine`, or I'm just wrong.

lrascao: yeah, that was the intended meaning, i can rename it to something else though ( maybe `aec_test_pow`)

dincho:
> i've added a fixup on top of the branch addressing all the comments except the renaming of aec_pow_test

thanks

apps/aecore/src/aec_pow_test.erl
```diff
@@ -0,0 +1,42 @@
...
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+-include("pow.hrl").
+
+%%%=============================================================================
+%%% API
+%%%=============================================================================
+-spec generate(Data :: aec_hash:hashable(), Target :: aec_pow:sci_int(),
+               Nonce :: aec_pow:nonce()) -> aec_pow:pow_result().
+generate(_Data, _Target, Nonce) ->
```
sennui: The refactoring is too shallow. A generic PoW framework, should make any PoW pluggable. The Test PoW still respects restrictions native to Cuckoo, which may make no sense in the new Test context.

lrascao: is it ok in terms of interface though? i see the purpose of a generic PoW framework as being generating data and verifying said data, so the thing missing here is a way to generically parameterize the invocation of these methods, am i reading it right?

apps/aecore/src/aec_pow_test.erl
ThomasArts: aec_test_pow would be a good name. The test framework picks up all _test as test cases and will start running them. Even if that's not the case, the code reader will think it's a test.

Since the module name already indicated that this is a test pow, you could easily well just call it pow().

evdoks:
@lrascao We have decided not to support pluggable PoW pluggable PoW for now. Besides, once stratum support is released, it will be very easy to implement pluggable PoW in the stratum code base.
I would suggest first opening an issue where a feature is first discussed and only then doing a PR.

PR #1429, opened 2018-08-10
Fix micro block cycle time

hanssv:
Don't sign blocks during micro_sleep period
    
Also simplify aec_conductor slightly by doing the signing in place rather than in a worker.


hanssv:
This PR, as well as dd0da34, fixes bugs. After these changes the code will behave as already documented?!

hanssv:
However, we are about to also fix the validation of micro blocks (to check that the block cycle time is adhered to) and then it makes sense to put a note in the release notes.

PR #1435, opened 2018-08-13
fix release/build files permissions

acidsennui:


.circleci/config.yml
```diff
@@ -133,12 +133,16 @@ references:
   upload_build_artifacts: &upload_build_artifacts
     run:
       name: Upload build artifacts to central repository (AWS)
-      command: aws s3 cp ${PACKAGE_TARBALL:?} s3://${S3_BUILDS_BUCKET:?}/
+      command: |
+        aws s3 cp ${PACKAGE_TARBALL:?} s3://${S3_BUILDS_BUCKET:?}/
+        aws s3api put-object-acl --acl public-read --bucket ${S3_BUILDS_BUCKET:?} --key ${PACKAGE_TARBALL_FILE:?}
```
dincho: I prefer `basename ${PACKAGE_TARBALL:?}` instead of additional redundant variable that also pollutes the environment variables.

.circleci/config.yml
```diff
@@ -133,12 +133,16 @@ references:
   upload_build_artifacts: &upload_build_artifacts
     run:
       name: Upload build artifacts to central repository (AWS)
-      command: aws s3 cp ${PACKAGE_TARBALL:?} s3://${S3_BUILDS_BUCKET:?}/
+      command: |
+        aws s3 cp ${PACKAGE_TARBALL:?} s3://${S3_BUILDS_BUCKET:?}/
+        aws s3api put-object-acl --acl public-read --bucket ${S3_BUILDS_BUCKET:?} --key `basename ${PACKAGE_TARBALL:?}`
```
lucafavatella: Is this line restricting or relaxing permissions?

Setting ACL after having uploaded artefact seems strange to me, as execution of this script may terminate in between commands, leaving package uploaded but in unexpected state. Why not using [`cp` option `--acl`](https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html#options)?

.circleci/config.yml
```diff
@@ -133,12 +133,16 @@ references:
...
 
   upload_release_artifacts: &upload_release_artifacts
```
lucafavatella: I see this step only part of `upload_packages_linux` job, only run on tag. Ok.

.circleci/config.yml
```diff
@@ -133,12 +133,16 @@ references:
...
     run:
       name: Upload release artifacts to central repository (AWS)
-      command: aws s3 cp ${PACKAGE_TARBALL:?} s3://${S3_RELEASES_BUCKET:?}/
+      command: |
+        aws s3 cp ${PACKAGE_TARBALL:?} s3://${S3_RELEASES_BUCKET:?}/
+        aws s3api put-object-acl --acl public-read --bucket ${S3_RELEASES_BUCKET:?} --key `basename ${PACKAGE_TARBALL:?}`
```
lucafavatella: Same as ACL comment above.

.circleci/config.yml
```diff
@@ -133,12 +133,16 @@ references:
   upload_build_artifacts: &upload_build_artifacts
```
lucafavatella: I see this step is in `linux_package` job that runs on all branches and tags. I see that package name on tags has version, on non-tag the commit hash.

And I see the bucket is distinct from the release one. Ok.

.circleci/config.yml
dincho: It's relaxing.

However, good point about `cp --acl`. @acidsennui Any reason it's separate API call?

acidsennui:
closing it's duplicate of https://github.com/aeternity/epoch/pull/1437

PR #1442, opened 2018-08-14
Fix some intermittent test failures

hanssv:


py/tests/integration/test_api_usage_contracts_counter.rst
```diff
@@ -83,6 +83,8 @@ th$...
 
 Bob ensures that the published contract create transaction is included in the chain.
 
+>>> import time
+>>> time.sleep(5)
```
lucafavatella: I see proper wait tracked as https://www.pivotaltracker.com/n/projects/2124891/stories/159762808 - ok

PR #1443, opened 2018-08-14
Contract executable documentation stabilization and speed

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/159762808


lucafavatella:
I run `while ( make python-single-uat TEST_NAME=test_api_usage_contracts_counter.rst; ); do echo Passed; done` for a while and it seems issue is not here anymore.

PR #1445, opened 2018-08-15
Block validation - key and micro block timestamp validation

hanssv:
Corresponding update to protocol: https://github.com/aeternity/protocol/pull/167

Key blocks should have `Timestamp < NOW + 10m` and `Timestamp > MedianTimestamp(11)` (the median timestamp of the last 11 blocks, or larger than Genesis timestamp for the first 11 blocks).

Micro blocks should have `Timestamp < NOW + 10m` and **between** microblocks there should be 3 seconds. The first micro block following a key block is only required to have `Timestamp > PrevTimestamp`.


apps/aecore/src/aec_governance.erl
```diff
@@ -28,10 +30,12 @@
          }).
 
 -define(BLOCKS_TO_CHECK_DIFFICULTY_COUNT, 10).
+-define(TIMESTAMP_MEDIAN_BLOCKS, 11).
 -define(EXPECTED_BLOCK_MINE_RATE, 300000). %% 60secs * 1000ms * 5 = 300000msecs
 -define(BLOCK_MINE_REWARD, 10000000000000000000).
 -define(MICRO_BLOCK_CYCLE, 3000). %% in msecs
 
+-define(ACCEPTED_FUTURE_BLOCK_TIME_SHIFT, 10 * 60 * 1000). %% 30 min
```
lucafavatella: 30 min ?

apps/aecore/src/aec_chain_state.erl
```diff
@@ -42,6 +43,22 @@ get_n_key_headers_backward_from(Header, N) ->
     Node = wrap_header(Header),
     get_n_key_headers_from(Node, N).
 
+%% Compute the median timestamp for last aec_governance:median_timestamp_key_blocks()
+-spec median_timestamp(aec_headers:header()) -> {ok, integer()} | 'error'.
+median_timestamp(Header) ->
+    TimeStampKeyBlocks = aec_governance:median_timestamp_key_blocks(),
+    case aec_headers:height(Header) =< TimeStampKeyBlocks + 1 of
```
lucafavatella: So when validating block at height 12 this is not computing the median? This seems in conflict with [spec](https://github.com/aeternity/protocol/pull/167/files#diff-c423cd627c304e382e23fb48bcb356dfR143).

lucafavatella:
I assume this aims at delivering https://www.pivotaltracker.com/story/show/153390572 ?

hanssv:
https://www.pivotaltracker.com/story/show/153390572

hanssv:
Fixed comments and rebased

apps/aecore/src/aec_chain_state.erl
```diff
@@ -42,6 +43,22 @@ get_n_key_headers_backward_from(Header, N) ->
     Node = wrap_header(Header),
     get_n_key_headers_from(Node, N).
 
+%% Compute the median timestamp for last aec_governance:median_timestamp_key_blocks()
+-spec median_timestamp(aec_headers:header()) -> {ok, integer()} | 'error'.
+median_timestamp(Header) ->
+    TimeStampKeyBlocks = aec_governance:median_timestamp_key_blocks(),
+    case aec_headers:height(Header) =< TimeStampKeyBlocks of
```
lucafavatella: If block being validated has height 11 then returns genesis time. Ok.

apps/aecore/src/aec_governance.erl
```diff
@@ -28,10 +30,12 @@
          }).
 
 -define(BLOCKS_TO_CHECK_DIFFICULTY_COUNT, 10).
+-define(TIMESTAMP_MEDIAN_BLOCKS, 11).
```
sennui: maybe the name can express more the number? TIMESTAMP_MEDIAN_BLOCKS_RANGE?

apps/aecore/src/aec_chain_state.erl
```diff
@@ -42,6 +43,22 @@ get_n_key_headers_backward_from(Header, N) ->
...
+        true ->
+            {ok, ?GENESIS_TIME};
+        false ->
+            case get_n_key_headers_from(wrap_header(Header), TimeStampKeyBlocks + 1) of
```
lucafavatella: If e.g. Header is header of key block and has height 12, then get_n_key_headers_from is called with the (wrapped) Header and 12 (11+1).
Then get_n_key_headers_from returns list of headers - exactly 11 - where the last is header at height 12 itself and first (list head) is header at height 2 (note: header at height 1 is missing). Then line below takes tail (`tl`) - that keeps headers from 3 to 12 (note: 12 is kept in computation of median for verification of 12 itself).

I feel I may be reading this code incorrectly.

lucafavatella: I see tl replaced with lists:droplast - ok.

hanssv: Yes, well spotted.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -752,6 +786,13 @@ do_find_micro_fork_point(Hash1, Hash2) ->
         false -> do_find_micro_fork_point(db_get_prev_hash(Hash1), Hash2)
     end.
 
+median(Xs) ->
+    Sorted = lists:sort(Xs),
+    Length = length(Sorted),
+    Mid = Length div 2,
+    Rem = Length rem 2,
+    (lists:nth(Mid+Rem, Sorted) + lists:nth(Mid+1, Sorted)) div 2.
```
lucafavatella: 6 is median of 11. Ok.

```
1> Sorted = lists:seq(1, 11).
[1,2,3,4,5,6,7,8,9,10,11]
2> 11 = Length = length(Sorted).
11
3> Mid = Length div 2.
5
4> Rem = Length rem 2.
1
5> lists:nth(Mid+Rem, Sorted).
6
6> lists:nth(Mid+1, Sorted).
6
7> (lists:nth(Mid+Rem, Sorted) + lists:nth(Mid+1, Sorted)) div 2.
6
```

apps/aecore/src/aec_headers.erl
```diff
@@ -272,10 +275,37 @@ validate_pow({#header{nonce        = Nonce,
...
-                           ok | {error, block_from_the_future}.
-validate_time({#header{time = Time}, _}) ->
-    MaxAcceptedTime = aeu_time:now_in_msecs() + ?ACCEPTED_FUTURE_KEY_BLOCK_TIME_SHIFT,
+-spec validate_median_time({header(), aec_governance:protocols()}) ->
+        ok | {error, block_from_the_past}.
+validate_median_time({Header = #header{time = Time}, _}) ->
+    case aec_chain_state:median_timestamp(Header) of
+        {ok, MTime} when Time > MTime -> ok;
+        {ok, _MTime}                  -> {error, block_from_the_past};
+        error                         -> ok %% We can't know yet - checked later
```
lucafavatella: I see that "later" is done by assert_key_block_time alongside other checks when block connected to genesis. Ok.

apps/aecore/src/aec_headers.erl
```diff
@@ -272,10 +275,37 @@ validate_pow({#header{nonce        = Nonce,
...
+    end.
+
+-spec validate_micro_block_cycle_time({header(), aec_governance:protocols()}) ->
+        ok | {error, bad_micro_block_interval}.
+validate_micro_block_cycle_time({#header{time = Time, prev_hash = PrevHash}, _}) ->
+    case aec_chain:get_header(PrevHash) of
+        {ok, PrevHeader = #header{ time = PrevTime }} ->
+            MinAccepted =
+                case aec_headers:type(PrevHeader) of
+                    micro -> PrevTime + aec_governance:micro_block_cycle();
```
lucafavatella: I see lack of +1 here may be surprising [when reading the doc](https://github.com/aeternity/protocol/pull/167#discussion_r210232363).

lucafavatella: Though it looks to me this shall be more precisely specified in doc - rather than changing code.

apps/aecore/src/aec_headers.erl
```diff
@@ -272,10 +275,37 @@ validate_pow({#header{nonce        = Nonce,
...
+                    key   -> PrevTime + 1
+                end,
+            case Time >= MinAccepted of
+                true  -> ok;
+                false -> {error, bad_micro_block_interval}
+            end;
+        error ->
+            ok %% We don't know yet - checked later
```
lucafavatella: The structure of this case statement suggests that if microblock to be validated does not have previous key block in DB yet, then "later" this validation will be performed. Though in assert_micro_block_time I do not see this validation (i.e. how microblock time relates to immediately preceding key block's time).

I may be missing it though blocking just being sure.

apps/aecore/src/aec_headers.erl
hanssv: Yes, it is a matter of taste if we want to accept X ms difference or if we want to require X+1 ms...

lucafavatella:
I see pushed patch:
```
diff --git a/apps/aecore/src/aec_chain_state.erl b/apps/aecore/src/aec_chain_state.erl
index 93a2c0e1..0db9d208 100644
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -427,11 +427,18 @@ assert_key_block_time(_PrevNode, Node) ->
     end.

 assert_micro_block_time(PrevNode, Node) ->
-    case is_micro_block(Node) andalso is_micro_block(PrevNode) of
+    case is_micro_block(Node) of
         true ->
-            case time_diff_greater_than_minimal(Node, PrevNode) of
-                true  -> ok;
-                false -> internal_error(micro_block_time_too_low)
+            case is_micro_block(PrevNode) of
+                true ->
+                    case time_diff_greater_than_minimal(Node, PrevNode) of
+                        true  -> ok;
+                        false -> internal_error(micro_block_time_too_low)
+                false ->
+                    case node_time(Node) > node_time(PrevNode) of
+                        true  -> ok;
+                        false -> internal_error(micro_block_time_too_low)
+                    end
             end;
         false -> ok
     end.
```

apps/aecore/test/aec_blocks_tests.erl
```diff
@@ -53,10 +53,13 @@ validate_test_() ->
              TmpKeysDir = aec_test_utils:aec_keys_setup(),
              meck:new(enacl, [passthrough]),
              meck:expect(enacl, sign_verify_detached, 3, {ok, <<>>}),
+             meck:new(aec_chain, [passthrough]),
+             meck:expect(aec_chain, get_header, 1, error),
```
lucafavatella: So previous header assumed unknown hence microblock cycle time check skipped. Seems ok.

PR #1440, opened 2018-08-14
Fix compiler bug with builtin map functions

UlfNorell:

PR #1446, opened 2018-08-15
Check for generation not only key block in aec_sync

hanssv:
When identifying blocks that do not need to be synced we have to look
at the whole generation, not only the keyblock.

PT - https://www.pivotaltracker.com/story/show/159791019


lucafavatella:
Auxiliary [CI workflow with system tests](https://circleci.com/workflow-run/e149fec0-751a-4ddf-b96a-753f1a9afe52) started on f6509de0 equivalent to 8b2a331f

lucafavatella:
@hanssv When merging, please let us merge this keeping merge commit so PR #1440 does not need further rebase.

PR #1447, opened 2018-08-15
Release preparation PR

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/159197480


lucafavatella:
I followed user doc as user would (from package built by `make prod-package` on 9c2d4e7c that is equivalent to master 6de59006), including mining locally, and all seems ok.

PR #1448, opened 2018-08-15
Trunc old data from channel's state on each round

velzevur:
[SC: Prune accounts and contracts](https://www.pivotaltracker.com/story/show/159794040)
Truncate old state out of channel's state trees - accounts and contracts


velzevur:
Not required for release 20.0, so marked as WIP

PR #1449, opened 2018-08-16
aec_conductor should check for generation when inserting sync block

hanssv:


hanssv:
This is fixing (hopefully completely this time) https://www.pivotaltracker.com/story/show/159791019

lucafavatella:
Started [auxiliary system tests job on CI server](https://circleci.com/gh/aeternity/epoch/26298) on a8c4b3e equivalent to 484a912b

apps/aecore/src/aec_conductor.erl
```diff
@@ -833,14 +833,14 @@ as_hex(S) ->
 
 handle_add_block(#{ key_block := KeyBlock } = Block, #state{} = State, Origin) ->
     Header = aec_blocks:to_header(KeyBlock),
-    handle_add_block(Header, Block, State, Origin);
+    handle_add_block(Header, fun aec_sync:has_generation/1, Block, State, Origin);
```
lucafavatella: I see that this means that `Block` (that is compound of a key block and microblocks - plus direction forward/backward) will be (attempted to be) inserted in chain if any of:
* Key block `KeyBlock`
* Most recent key block before `KeyBlock`
* Any microblocks in between

are absent from DB.

In this respect, I am considering whether this line assumes that the `dir` in `Block` is `backward`.

lucafavatella: If the `dir` in `Block` were `forward` - meaning that `Block` is meant to contain a keyblock and a list of subsequent microblocks - then:
* (A) If `KeyBlock` or most recent preceding key block or any microblocks in between are absent from DB, then inserting `Block` is attempted. All good.
* (B) If `KeyBlock` and preceding microblocks down to most recent preceding keyblock (included) are present, microblocks included in `Block` are not attempted to be included.

Though B was present also before this patch, and it was more likely.

PR #1458, opened 2018-08-20
Add transaction hash to aetx:process/5

velzevur:
PT [Expand aetx:process callback with TxHash](https://www.pivotaltracker.com/story/show/159887336)
This is a requirement for forcing of progress where the call's id will use the transaction hash for its contract id.

Tracked as a separate PT for easier review

### DIFF
--- a/apps/aetx/src/aetx.erl
+++ b/apps/aetx/src/aetx.erl
@@ -15,8 +15,8 @@
         , new/2
         , nonce/1
         , origin/1
-        , process_no_tx_hash/4
-        , process/5
+        , process/4
+        , process_with_tx_hash/5
         , process_from_contract/4
         , serialize_for_client/1
         , serialize_to_binary/1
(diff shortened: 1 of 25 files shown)


apps/aetx/src/aetx.erl
```diff
@@ -219,18 +221,25 @@ check(AeTx = #aetx{ cb = CB, tx = Tx }, Trees, Height, ConsensusVersion) ->
...
+process(#aetx{ cb = CB, tx = Tx }, Trees, Height, ConsensusVersion, TxHash) ->
+    CB:process(Tx, aetx_transaction, Trees, Height, ConsensusVersion, TxHash).
 
 -spec process_from_contract(Tx :: tx(), Trees :: aec_trees:trees(), Height :: non_neg_integer(),
                             ConsensusVersion :: non_neg_integer()) ->
     {ok, NewTrees :: aec_trees:trees()}.
 
 process_from_contract(#aetx{ cb = CB, tx = Tx }, Trees, Height, ConsensusVersion) ->
-    CB:process(Tx, aetx_contract, Trees, Height, ConsensusVersion).
+    CB:process(Tx, aetx_contract, Trees, Height, ConsensusVersion, <<>>).
```
lucafavatella: I am concerned about passing as no-value a value that can be misused easily as valid value. May I suggest a special value e.g. atom `no_tx_hash`?

apps/aetx/src/aetx.erl
```diff
@@ -219,18 +221,25 @@ check(AeTx = #aetx{ cb = CB, tx = Tx }, Trees, Height, ConsensusVersion) ->
...
 
--spec process(Tx :: tx(), Trees :: aec_trees:trees(), Height :: non_neg_integer(),
+
+-spec process_no_tx_hash(Tx :: tx(), Trees :: aec_trees:trees(), Height :: non_neg_integer(),
               ConsensusVersion :: non_neg_integer()) ->
     {ok, NewTrees :: aec_trees:trees()}.
-process(#aetx{ cb = CB, tx = Tx }, Trees, Height, ConsensusVersion) ->
-    CB:process(Tx, aetx_transaction, Trees, Height, ConsensusVersion).
+process_no_tx_hash(Tx, Trees, Height, ConsensusVersion) ->
+    process(Tx, Trees, Height, ConsensusVersion, <<>>).
```
lucafavatella: I am concerned about passing as no-value a value that can be misused easily as valid value. May I suggest a special value e.g. atom `no_tx_hash`?

lucafavatella:
> PT Expand aetx:process callback with TxHash

It seems that newly passed argument is unused. Could you please make this preparatory PT a blocker for PT that will actually use this passed argument?

velzevur:
>Could you please confirm that the tx that is meant to create call object with tx hash as id is meant to be only tx - i.e. not meant to be created by a contract builtin/primop? (I am concerned of merging argument still unused to master, and then realizing that will not be used.)

Yes. It will be used by a new tx type (aesc_force_progess_tx) and shall not be used by a contract builtin/primop

velzevur:
Modifying the empty value to be `no_tx_hash`

PR #1460, opened 2018-08-21
PT-152843085 target calculation

hanssv:
Corresponding update to `protocol` is WIP


hanssv:
Finishes https://www.pivotaltracker.com/story/show/152843085

hanssv:
Will need a rebase on #1459

hanssv:
And fixing the UAT-tests.

apps/aecore/src/aec_block_key_candidate.erl
```diff
@@ -34,13 +34,13 @@ create(Block, Beneficiary) ->
                            {ok, aec_blocks:block()} | {error, term()}.
 adjust_target(Block, AdjHeaders) ->
     Header = aec_blocks:to_header(Block),
-    DeltaHeight = aec_governance:key_blocks_to_check_difficulty_count(),
+    DeltaHeight = aec_governance:key_blocks_to_check_difficulty_count() + 1,
```
sennui: we have mysterious +1 (or -1) all over the place. Why not to bump it in governance?

gorillainduction: I believe this is because of the difference between validating the target and generating next target, so it probably cannot be fixed in one place.

hanssv: Yes, exactly, we are using the targets of the last 17 blocks, but also the solvetimes of the last 17 blocks. However, the latter require 18 blocks to compute... I prefer the constant to be 17 then since that is the `N` in the algorithm.

hanssv:
Rebased, fixed tests etc.

https://github.com/aeternity/protocol/pull/171

PR #1462, opened 2018-08-22
Make aecore_txs_SUITE properly depending on chain height

hanssv:
Avoids yet another intermittent test failure...

https://www.pivotaltracker.com/story/show/156154502

### DIFF
--- a/apps/aecore/test/aecore_txs_SUITE.erl
+++ b/apps/aecore/test/aecore_txs_SUITE.erl
@@ -72,54 +72,73 @@ txs_gc(Config) ->
 
     %% Mine a block to get some funds. Height=1
     aecore_suite_utils:mine_key_blocks(N1, 1),
+    Height0 = 1,
 
     %% Add a bunch of transactions...
-    add_spend_tx(N1, 1000, 1,  1,  10), %% Ok
-    add_spend_tx(N1, 1000, 1,  2,  10), %% Should expire ?EXPIRE_TX_TTL after
-                                        %% next TX is on chain = 2 + 2 = 4
-    add_spend_tx(N1, 1000, 10, 2,  10), %% Duplicate should be preferred
-    add_spend_tx(N1, 1000, 1,  3,  10), %% Ok
+    {ok, TxH1} = add_spend_tx(N1, 1000, 1,  1,  10), %% Ok
+    {ok, _}    = add_spend_tx(N1, 1000, 1,  2,  10), %% Should expire ?EXPIRE_TX_TTL after
+                                                     %% next TX is on chain = 2 + 2 = 4
+    {ok, TxH2} = add_spend_tx(N1, 1000, 10, 2,  10), %% Duplicate should be preferred
+    {ok, TxH3} = add_spend_tx(N1, 1000, 1,  3,  10), %% Ok
 
-    add_spend_tx(N1, 1000, 1,  5,  10), %% Non consecutive nonce
-    add_spend_tx(N1, 1000, 1,  7,  10), %% Non consecutive nonce
-    add_spend_tx(N1, 1000, 1,  8,  5),  %% Short TTL - expires at 5 + 2 = 7
+    {ok, TxH5} = add_spend_tx(N1, 1000, 1,  5,  10), %% Non consecutive nonce
+    {ok, _}    = add_spend_tx(N1, 1000, 1,  7,  10), %% Non consecutive nonce
+    {ok, _}    = add_spend_tx(N1, 1000, 1,  8,  5),  %% Short TTL - expires at 5 + 2 = 7
 
     %% Now there should be 7 transactions in mempool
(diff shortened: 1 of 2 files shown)


apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -72,54 +72,73 @@ txs_gc(Config) ->
...
+    {ok, _}    = add_spend_tx(N1, 1000, 1,  8,  5),  %% Short TTL - expires at 5 + 2 = 7
 
     %% Now there should be 7 transactions in mempool
     {ok, Txs1} = rpc:call(N1, aec_tx_pool, peek, [infinity]),
     {7, _} = {length(Txs1), Txs1},
 
-    %% Mine two key blocks to get a micro-block in between. Height=3
-    aecore_suite_utils:mine_key_blocks(N1, 2),
+    %% Mine to get TxH1-3 onto chain
+    {ok, Blocks1} = aecore_suite_utils:mine_blocks_until_txs_on_chain(N1, [TxH1, TxH2, TxH3], 10),
```
lucafavatella: This only counts key blocks. Ok.

apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -72,54 +72,73 @@ txs_gc(Config) ->
...
+    Height1 = Height0 + length(Blocks1), %% Very unlikely to be > 4...
 
-    %% Now there should be 4 transactions in mempool
+    %% Now there should be 3 or 4 transactions in mempool
     {ok, Txs2} = rpc:call(N1, aec_tx_pool, peek, [infinity]),
-    {4, _} = {length(Txs2), Txs2},
+    case Height1 of
+        3              -> {4, _} = {length(Txs2), Txs2};
+        N1 when N1 > 3 -> ct:log("Skipping assert; micro fork advanced height to far...")
```
lucafavatella: Is the idea that a TTL may have already kicked in hence dropping tx from mempool?

apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -72,54 +72,73 @@ txs_gc(Config) ->
...
+
+    Height2 = max(Height1, 4),
 
-    %% Now there should be 3 transactions in mempool
+    %% Now unless Height2 > 4, there should be 3 transactions in mempool
     {ok, Txs3} = rpc:call(N1, aec_tx_pool, peek, [infinity]),
-    {3, _} = {length(Txs3), Txs3},
+    case Height2 of
+        4              -> {3, _} = {length(Txs3), Txs3};
+        N2 when N2 > 4 -> ct:log("Skipping assert; micro fork advanced height to far...")
```
lucafavatella: Same story - because of potential TTL, right?

apps/aecore/test/aecore_txs_SUITE.erl
lucafavatella: How is that min height is 3? May it not be 2?

apps/aecore/test/aecore_txs_SUITE.erl
hanssv: Yes, `mine_blocks_until_tx(s)_on_chain` only returns key-blocks.

apps/aecore/test/aecore_txs_SUITE.erl
hanssv: Yes, if we have a microfork inserting TxH1-3 we might have already reached height 4.

apps/aecore/test/aecore_txs_SUITE.erl
hanssv: Yes.

apps/aecore/test/aecore_txs_SUITE.erl
hanssv: No, that is impossible. We mine to 1 before adding txs. And then we mine until txs are on chain - that cannot happen before we mine block 2 (we are not the leader) and then we need the key block at height 3 to close the generation so the check in `mine_blocks_until_txs_on_chain` can succeed.

lucafavatella: Node dev1 is the only node. It mines key block 1, hence becomes leader. Then transactions are posted. Generation 1 may include txs. Key block 2 may be the one closing the generation. (Or I am missing something.)

So this case may intermittently crash.

hanssv: No, after it has mined key block 1 mining is switched off and it is no longer the leader. Then the transactions are posted. Hence it has to first mine block 2 before it can include the transactions in a micro block. This is the same in almost _every_ test case we have...

lucafavatella: I see https://github.com/aeternity/epoch/blob/785232a123b57f42d7d6e758589d837dee2baeba/apps/aecore/src/aec_conductor.erl#L219

PR #1465, opened 2018-08-22
Pt 159829816 inet blocks aec_peers

ThomasArts:
If inet:getaddr takes too much time, we spend overtime in aec_sync or aec_peers loops. This has caused a number of timeout errors.

This in one go solves tickets reported in:
PT-159829816, PT-159648052, and PT-159490846 

System tests pass locally as well as on a faster machine.

### DIFF
--- a/apps/aecore/src/aec_peer_connection.erl
+++ b/apps/aecore/src/aec_peer_connection.erl
@@ -635,7 +635,7 @@ handle_first_ping(S, RemotePingObj) ->
             end,
 
             NewS = maps:remove(first_ping_tref, S#{ port => Port }),
-            case aec_peers:peer_accepted(PeerInfo, self()) of
+            case aec_peers:peer_accepted(PeerInfo, maps:get(address, NewS), self()) of
                 {error, _} = Error ->
                     gen_server:cast(self(), stop),
                     {Error, NewS};
--- a/apps/aecore/src/aec_peers.erl
+++ b/apps/aecore/src/aec_peers.erl
@@ -983,7 +983,7 @@ on_resolve_hostname(Ref, Host, #state{ hostnames = HostMap } = State) ->
 -spec on_add_peer(inet:ip_address(), peer(), state()) -> state().
 on_add_peer(SourceAddr, Peer, State0) ->
     #state{ hostnames = HostMap } = State0,
-    State = maybe_unblock(State0#state{hostnames = maps:remove(Peer#peer.host, HostMap)}),
+    State = maybe_unblock(State0#state{hostnames = maps:remove(to_list(Peer#peer.host), HostMap)}),
     PeerId  = peer_id(Peer),
     #peer{ pubkey = PPK, address = PA, port = PP } = Peer,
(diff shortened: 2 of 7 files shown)


apps/aecore/src/aec_peers.erl
```diff
@@ -973,38 +967,23 @@ on_resolve_peer(SourceAddr, PeerInfo, IsTrusted, State) ->
...
         _ ->
             State
     end.
 
 %% Handles event adding a new peer to the pool.
 -spec on_add_peer(inet:ip_address(), peer(), state()) -> state().
 on_add_peer(SourceAddr, Peer, State0) ->
-    State = maybe_unblock(State0),
+    #state{ hostnames = HostMap } = State0,
+    State = maybe_unblock(State0#state{hostnames = maps:remove(Peer#peer.host, HostMap)}),
```
sylane: Multiple peers could have the same hostname, wouldn't that cancel them to be added?

apps/aecore/src/aec_peers.erl
```diff
@@ -559,22 +544,31 @@ update_ping_metrics(Outcome) ->
...
+                 {error, nxdomain} ->
+                     epoch_sync:info("Peer ~p - failed to resolve hostname ~p; "
+                                     "retrying later", [ppp(PeerInfo), Host]),
+                     gen_server:cast(?MODULE, {resolve_peer, SourceAddr0,
+                                               PeerInfo, IsTrusted});
+                 {ok, Addr} when SourceAddr0 == undefined ->
+                     epoch_sync:info("PEER undefined address ~p ~p -> ~p", [Host, SourceAddr0, Addr]),
+                     Peer = peer(Addr, PeerInfo, IsTrusted),
+                     gen_server:cast(?MODULE, {add_peer, Addr, Peer});
+                 {ok, Addr} when SourceAddr0 =/= undefined ->
```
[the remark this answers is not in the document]
ThomasArts: The refcatoring is from:
```
            SourceAddr = case SourceAddr0 of
                undefined -> Peer#peer.address;
                _ -> SourceAddr0
            end,
            gen_server:cast(?MODULE, {add_peer, SourceAddr, Peer})
```
Which states, if SourceAddr0 not equal undefined, then take SourceAddr0.

apps/aecore/src/aec_peers.erl
ThomasArts: This seems fine, the original code alo removed the host from the hostnames to be resolved when one of the resolves succeeded.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -635,7 +635,7 @@ handle_first_ping(S, RemotePingObj) ->
             end,
 
             NewS = maps:remove(first_ping_tref, S#{ port => Port }),
-            case aec_peers:peer_accepted(PeerInfo, self()) of
+            case aec_peers:peer_accepted(PeerInfo, maps:get(address, NewS), self()) of
```
lucafavatella: I see this comes from `inet:peername/1` in init. Ok.

PR #1466, opened 2018-08-22
Fix calculating fees in db and granting too big rewards to beneficiaries

zp-sd:
Currently for chain:
```
(...) -> K1 -> M1 -> M2 -> M3 -> K2 -> (...)
```
fees in `#fork_info{}` in DB are as follows:
```
fees(M1) = M1_fees
fees(M2) = 2*M1_fees + M2_fees
fees(M3) = 2*(2*M1_fees + M2_fees) + M3_fees
```

This changes it so that:
```
fees(M1) = M1_fees
fees(M2) = M1_fees + M2_fees
fees(M3) = M1_fees + M2_fees + M3_fees
```


apps/aecore/include/blocks.hrl
```diff
@@ -1,6 +1,6 @@
 -include("pow.hrl").
 
--define(PROTOCOL_VERSION, 22).
```
hanssv: I've already bumped the version once for the next release, no point doing it multiple times?

zp-sd: I'm not sure, isn't this needed for integration envs? Isn't the most recent `master` automatically deployed somewhere and we need a hard fork? I don't know. @lucafavatella ?

lucafavatella: I see protocol version already bumped in master compared to tag v0.20.0 (protocol version bump is needed between releases e.g. between v0.20.0 and to-be-tagged v0.21.0).

In order to make our environments use a fresh DB folder in deployment of master, you need to bump deployment/DB_VERSION

docs/release-notes/RELEASE-NOTES-0.21.0.md
```diff
@@ -4,6 +4,7 @@
 It:
 * Adds support for type aliases to the Sophia compiler.
 * Changes the target (difficulty) calculation algorithm to use [DigiShield v3][digishield_v3]. This impacts consensus.
+* Fixes fees calculations, adding which has been too generous. This impacts consensus.
```
hanssv: This doesn't read right... "Fixes miner fee reward calculations, was too generous before" perhaps?

PR #1467, opened 2018-08-23
New HTTP API

fabiankrol:
Let's use this PR as a reference point. We're going to create smaller pr during the day. 
cc @jur0


jur0:
Why is there just 1 huge PR? Didn't we agree this will be split into multiple PRs? At least these:
* IDs
* msg pack removal
* API cleanup
* distinct prefixes for key/micro blocks
* fix of the swagger tags

Also, I wouldn't have squashed:
* the IDs into 1 commit
* the API cleanup/removal commits

fabiankrol:
@jur0 you're right `ids` and `message pack` will be uncouple from the main api cleanups.

PR #1469, opened 2018-08-23
SC: Forcing progress on-chain

velzevur:
PT [Channels: Force progress on-chain](https://www.pivotaltracker.com/story/show/159405568)
Provides `aesc_force_progress_tx`

### Description
A participant provides sufficent information and calls an off-chain created contract on-chain. Contract is executed and resulting `call` object is put in the on-chain state trees. Since contract id is not present on-chain the hash of the `aesc_force_progress_tx` transaction is usead instead.

Fees and gas are payed by the participant that forces progress.

### Provided data
Forcer provides:
+ channel_id - channel to be force pushed
+ from - forcer
+ payload - the state of the channel that we're pushing. That can be:
  - co-signed `aesc_offchain_tx` that has the same channel id, correct signatures and a round greater the one persited on-chain. Its `state_hash` and `round` are used
  - empty binary - the last on-chain persisted state is used - the on-chain stored `state_hash` and `round` are used
+ poi - proof of inclusion of accounts and contracts. Its `state_hash` must be the same as the one in the `payload`. Currently *all* accounts and contracts must be provided in order to reproduce them on-chain. To be optimizied in a subsequent PR
+ addresses - list of all accounts and contracts provided in the `poi`. Probably not required, depending on the optimization of `poi`.
+ solo_payload - a unilaterally signed `aesc_offchain_tx` that has only one element in its `updates` and it is an contract call. Its `state_hash` is the what would be the result root of the trees after applying the contract call to the `poi`. Its `round` is the bumped by one `round` provided by the `payload`
+ ttl
+ fee
+ nonce

### Result
If everything is fine with the transaction and the contract can be called - it is. This consumes gas fees. If the produced  `state_hash` is the same as the one provided in the `solo_payload` - it is persisted on-chain as well as the `round`.

#### A closing channel
We can also force push a channel that is no longer active. This acts as a slash transaction and it bumps the lock timer (the one before the settle tx). It also updates on-chain representation of channel's participants balances on-chain. This way we can update closing balances accordingly.

### Co-signed vs computed state
If provided a conflict between on-chain persisted forced progressed state and a co-signed state having the same `round` - the co-signed state wins. After each force progress there is a block height timer interval so the other party can post a co-signed state (snapshot tx for example) in order to dispute the forcing of progress.

Main problem addressed here is one participant producing a couple of subsequent force pushes (starting from a previous round) and thus - producing a _longer chain_ and overwrite the one off-chain. This is when the timer kicks in for the other participant can dispute those with providing co-signed off-chain state with the same/greater round.


velzevur:
I believe we have the most important tests now. I will add a line in the release notes and remove the `WIP` label

apps/aechannel/src/aesc_utils.erl
```diff
@@ -341,9 +442,229 @@ process_solo_close_slash(ChannelId, FromPubKey, Nonce, Fee,
...
+    {ContractPubkey, Caller} = aesc_offchain_update:extract_call(Update),
+    %% use in gas payment
+    PoITrees0 = trees_from_poi(Addresses, PoI),
+    Reserve = aesc_channels:channel_reserve(Channel),
+    NextRound = aesc_offchain_tx:round(SoloPayloadTx),
+    PoITrees =
+        try aesc_offchain_update:apply_on_trees(Update, PoITrees0, NextRound,
+                                                      Reserve) of
+            UpdatedPoITrees -> UpdatedPoITrees
+        catch _:_ ->
```
velzevur: If anything goes wrong with the contract call updating the state tree provided - we go here. Then a new contract call is created (with an `error` return type and `invalid_call` as a value) consuming all gas provided.

This might happen if the contract call refers to another address (either a contract or an account) that is not present in the trees for example. Since we are not running the contract in the check - we don't catch those cases.

velzevur:
Will have conflicts with #1485; Will rebase after RP #1485 is merged

velzevur:
Will conflict with #1490

velzevur:
Doing a rebase on top of a current master;
will also `apps/aechannel/src/aesc_force_progress.erl` -> `apps/aechannel/src/aesc_force_progress_tx.erl`

apps/aechannel/src/aesc_force_progress.erl
```diff
@@ -0,0 +1,238 @@
...
+                                   nonce        = Nonce} = Tx, _Context,
+        Trees, Height, _ConsensusVersion, TxHash) when is_binary(TxHash) ->
+    ChannelId  = channel_hash(Tx),
+    FromPubKey = from_pubkey(Tx),
+    aesc_utils:process_force_progress(ChannelId, FromPubKey, Nonce, Fee,
+                                      Payload, SoloPayload, Addresses,
+                                      PoI, TxHash, Height, Trees).
+
+-spec signers(tx(), aec_trees:trees()) -> {ok, list(aec_keys:pubkey())}.
+signers(#channel_force_progress_tx{} = Tx, _) ->
```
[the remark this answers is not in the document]
velzevur: This is part of the [aetx behavior](https://github.com/aeternity/epoch/blob/master/apps/aetx/src/aetx.erl#L123)

apps/aechannel/src/aesc_channels.erl
```diff
@@ -46,17 +49,19 @@
...
+                  responder           :: aec_id:id(),
+                  delegates           :: [aec_id:id()],
+                  channel_amount      :: amount(),
+                  initiator_amount    :: amount(),
+                  responder_amount    :: amount(),
+                  channel_reserve     :: amount(),
+                  state_hash          :: binary(),
+                  round               :: seq_number(),
+                  lock_period         :: non_neg_integer(),
+                  force_blocked_till  :: aec_blocks:height(),
```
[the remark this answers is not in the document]
velzevur: renamed

velzevur:
I've missed updating the swagger doc. Cooking it now

velzevur:
Tests failed with an unrelated failure
```

%%% aeu_requests_SUITE ==> application_test: FAILED
%%% aeu_requests_SUITE ==> {errors_in_lager_log,["09:30:26.878 [error] CRASH REPORT Process <0.1729.1> with 0 neighbours exited with reason: {aborted,{no_exists,aec_chain_state}} in mnesia:wrap_trans/6 line 493\n\n",
                      "09:30:26.878 [error] Supervisor aecore_sup had child aec_tx_pool started with aec_tx_pool:start_link() at undefined exit with reason {aborted,{no_exists,aec_chain_state}} in context start_error\n\n",
                      "09:30:26.879 [error] CRASH REPORT Process <0.1721.1> with 0 neighbours exited with reason: {{shutdown,{failed_to_start_child,aec_tx_pool,{aborted,{no_exists,aec_chain_state}}}},{aecore_app,start,[normal,[]]}} in application_master:init/4 line 134\n\n"]}
```
Restarting the test job

apps/aechannel/src/aesc_utils.erl
```diff
@@ -341,9 +445,229 @@ process_solo_close_slash(ChannelPubKey, FromPubKey, Nonce, Fee,
...
+    [Update] = aesc_offchain_tx:updates(SoloPayloadTx),
+    {ContractPubkey, Caller} = aesc_offchain_update:extract_call(Update),
+    %% use in gas payment
+    PoITrees0 = trees_from_poi(Addresses, PoI),
+    Reserve = aesc_channels:channel_reserve(Channel),
+    NextRound = aesc_offchain_tx:round(SoloPayloadTx),
+    PoITrees =
+        try aesc_offchain_update:apply_on_trees(Update, PoITrees0, NextRound,
+                                                      Reserve) of
+            UpdatedPoITrees -> UpdatedPoITrees
```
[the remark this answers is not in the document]
velzevur: Addressed ;-)

apps/aechannel/src/aesc_utils.erl
```diff
@@ -341,9 +445,228 @@ process_solo_close_slash(ChannelPubKey, FromPubKey, Nonce, Fee,
...
+                    {error, update_not_call}
+            end;
+        [] ->
+            {error, no_update};
+        _ ->
+            {error, more_than_one_update}
+    end.
+
+-spec trees_from_poi([aec_id:id()], aec_trees:poi()) -> aec_trees:trees().
+trees_from_poi(IDs, PoI) ->
```
gorillainduction: I think you should check that the root hash of the constructed tree is the same as the root hash of the PoI. This is a safe way of checking that the PoI indeed represents the full tree, which in turn is a prerequisite for guaranteeing that the new computed root hash is the correct one.

Otherwise, you can strip the tree of certain parts that isn't touched by the contract call (e.g., another contract, or another account).

velzevur:
Rebased on top of endpoints and resolved conflicts

velzevur:
Since this is not making it in release 21, I am adding a `WIP` tag

docs/release-notes/RELEASE-NOTES-0.21.0.md
```diff
@@ -7,6 +7,8 @@ It:
 * Fixes miner fee reward calculations, was too generous before. This impacts consensus.
 * Modifies the minimum static component of the fee of oracle transactions to `1` - as for all other transactions. This impacts consensus.
 * Increases beneficiary reward delay to 180 key blocks / generations. This impacts consensus.
+* Adds on-chain forcing of progress for channel off-chain contracts. This
```
lucafavatella: Copied in https://github.com/aeternity/epoch/wiki/Release-notes-next/_compare/b373e5dfb8f3b15c113b71e88e436393942654c8...df1f035f5162e74e25265af132a96dd3b2d07182

lucafavatella: Change to this file shall be reverted.

lucafavatella:
I see that not much changes in master since 14ac3af2 - that is commit this PR is based on.

PR #1470, opened 2018-08-23
Oracle txs minimum static fee component

lucafavatella:
Delivers https://www.pivotaltracker.com/n/projects/2124891/stories/158423604
See also related cleanup PR #1468

Please refer to commit message for details.

----

Not TODO:
* Detailing documentation of dynamic component of oracle txs based on TTL of objects in state trees. That is in scope of https://www.pivotaltracker.com/story/show/159134621

PR #1471, opened 2018-08-23
Pt 157551442 split block on type

gorillainduction:
https://www.pivotaltracker.com/story/show/157551442


deployment/DB_VERSION
```diff
@@ -1 +1 @@
-79
+80
```
lucafavatella: Master seems [already at 80](https://github.com/aeternity/epoch/blob/452fcbd65e9db42f81bf222325023e88b2eaad03/deployment/DB_VERSION#L1) - blocking so we remember.

BTW there will be conflicts e.g. https://github.com/aeternity/epoch/pull/1470

lucafavatella:
This seems will conflict with #1467 too cc @fabiankrol @jur0

gorillainduction:
@lucafavatella Rebased on master and resolved conflict.
I think the conflict with #1467 will be trivial to resolve.
(cc @fabiankrol @jur0 )

apps/aecore/src/aec_blocks.erl
```diff
@@ -1,7 +1,10 @@
+%%% -*- erlang-indent-level:4; indent-tabs-mode: nil -*-
```
lucafavatella: We may consider slowly moving away from these using [editorconfig](https://github.com/aeternity/epoch/blob/78269fd936122b74fd00be617320ad4758b698b0/.editorconfig#L4-L5).

apps/aecore/src/aec_subscribe.erl
```diff
@@ -82,7 +82,10 @@ handle_info({gproc_ps_event, Event = micro_block_created, #{ info := Block }}, S
 
 notify_subscribers(top_changed = Event, Block, State = #state{ subscribed = Subs }) ->
     notify_chain_subscribers(Event, Block, Subs),
-    notify_tx_subscribers(aec_blocks:txs(Block), Subs),
+    case aec_blocks:type(Block) of
+        micro -> notify_tx_subscribers(aec_blocks:txs(Block), Subs);
+        key   -> ok
```
lucafavatella: I see this does not change behaviour (the line above calls `notify_chain_subscribers`). Ok.

PR #1475, opened 2018-08-23
Closes #159923409 - support "soft-reject" via competing update req

uwiger:
Covers race detection when an update is requested after a signing
request has been sent. This can also be used to "soft-reject" an update
(or deposit/withdraw) request, until a proper reject reply is implemented.


velzevur:
Test failed with an unrelared error and I've restarted them (due to a wrong button - all of them :) )

PR #1478, opened 2018-08-24
PT-158105051 Only gossip txhashes

hanssv:
This PR changes the format for micro blocks in P2P - instead of including whole Txs we include just TX hashes. If a node hasn't seen one (or more) transactions it will make a query for them.

In the future it could be a configuration option whether to gossip "light" micro blocks with just hashes or or full micro blocks.

Finishes https://www.pivotaltracker.com/story/show/158105051


apps/aecore/src/aec_peer_connection.erl
```diff
@@ -943,22 +926,144 @@ send_send_block(#{ status := error }, _Block) ->
...
+
+handle_get_block_txs(S, Msg) ->
+    Hash = maps:get(hash, Msg),
+    TxHashes = maps:get(tx_hashes, Msg),
+    case find_txs(TxHashes) of
+        {ok, Txs} ->
+            SerTxs = [ aetx_sign:serialize_to_binary(Tx) || Tx <- Txs ],
+            send_response(S, block_txs, {ok, #{ hash => Hash, txs => SerTxs }});
+        Err = {error, _} ->
+            send_response(S, txs, Err)
```
ThomasArts: block_txs ?

hanssv:
Will rebase this on on top of https://github.com/aeternity/epoch/pull/1481 (probably after the weekend!)

PR #1483, opened 2018-08-24
Fix test intermittent failure for reused variable N1

lucafavatella:


apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -96,8 +96,8 @@ txs_gc(Config) ->
     %% Now there should be 3 or 4 transactions in mempool
     {ok, Txs2} = rpc:call(N1, aec_tx_pool, peek, [infinity]),
     case Height1 of
-        3              -> {4, _} = {length(Txs2), Txs2};
-        N1 when N1 > 3 -> ct:log("Skipping assert; micro fork advanced height to far...")
+        3                -> {4, _} = {length(Txs2), Txs2};
+        HH1 when HH1 > 3 -> ct:log("Skipping assert; micro fork advanced height to far...")
```
velzevur: I see this is the actual fix, other changes are just complimentary. Nice catch.

hanssv: If only Erlang had sensible scoping... All of them of course was named `N` from the beginning since it is an unused varible. But then you get an "unsafe variable in case..." Mumble, mumble...

Good catch Luca!

PR #1484, opened 2018-08-24
Check configured beneficiary

sylane:
* Make beneficiary required in the configuration file.
 * Give a better error message when the schema given to check_config
   doesn't exists.


sylane:
@dincho There is no "default" beneficiary, this is just an example. But I see your point, the user may copy past it... we could make it more obvious this is an example value.

PR #1487, opened 2018-08-27
PT-158827776 API: 3/5 endpoints cleanup

fabiankrol:


fabiankrol:
Rebased on master

config/swagger.yaml
```diff
@@ -1447,7 +1319,7 @@ paths:
           name: body
           required: true
           schema:
-            $ref: '#/definitions/ChannelWithdrawalTx'
+            $ref: '#/definitions/ChannelWithdrawTx'
```
lucafavatella: I see minor withdraw/withdrawal conflict with PR #1469

py/tests/integration/test_spend_tx.py
```diff
@@ -73,7 +76,7 @@ def get_alice_balance(height):
     print("Coinbase reward is " + str(coinbase_reward) + ", had mined " +
           str(blocks_mined) + " blocks")
     expected_balance = bob_balance.balance + coinbase_reward * blocks_mined
-    bob_new_balance = get_bob_balance(checked_height)
+    bob_new_balance = get_bob_balance()
     print("Bob's balance (with coinbase rewards) is now " + str(bob_new_balance.balance))
     assert_equals(bob_new_balance.balance, expected_balance)
```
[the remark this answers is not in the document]
fabiankrol: You're right. We need to figure out better test scenario. I prefer to avoid adding new **debug/test** endpoints if possible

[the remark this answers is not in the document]
lucafavatella: If beneficiary were another keypair - e.g. Carol - and not Bob, then test would be not intermittent.

So idea would be that at beginning of test Carol transfers some tokens to Bob - though not enough to pay for the spend. Then rest of test would stay unchanged - except for Bob not gaining coinbase.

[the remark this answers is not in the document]
lucafavatella: Test was intermittent also in master:
* When [checking that Bob had not sufficient balance](https://github.com/aeternity/epoch/blob/51fe262865292340088517f7a96aebbbf59b786d/py/tests/integration/test_spend_tx.py#L42) - Bob is already mining
* When later checking that Alice not credited and Bob not debited - Bob may have reached enough tokens

So does not seem severe regression.

docs/release-notes/RELEASE-NOTES-0.20.0.md
```diff
@@ -72,10 +72,10 @@ For running a node as part of the testnet by using the Docker image, please cons
 The core nodes of the public test network are accessible from the Internet.
 
 Information, e.g. height, of the top block of the longest chain as seen by these core nodes of the testnet can be obtained by opening in the browser any of the following URLs:
-* http://52.10.46.160:3013/v2/top
-* http://18.195.109.60:3013/v2/top
-* http://13.250.162.250:3013/v2/top
-* http://31.13.249.70:3013/v2/top
+* http://52.10.46.160:3013/v2/blocks/top
```
lucafavatella: Wrong file? Shall be 0.21.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -163,8 +160,10 @@ handle_request('GetGenerationByHeight', Params, _Context) ->
     end;
 
 handle_request('GetAccountByPubkey', Params, _Context) ->
-    case aec_base58c:safe_decode(account_pubkey, maps:get(pubkey, Params)) of
-        {ok, Pubkey} ->
+    AllowedTypes = [account_pubkey, contract_pubkey],
```
lucafavatella: The lookup of contract account used to be in `GetAccountBalance`. Now it is here. Ok. Cc @happi @rvirding

fabiankrol: That's correct, `GetAccountBalance` is no longer with us 🙏

py/tests/integration/test_spend_tx.py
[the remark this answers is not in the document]
fabiankrol: 👍

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -15,328 +34,399 @@
...
+                ],
+    process_request(ParseFuns, Req);
+
+handle_request('PostNameTransfer', #{'NameTransferTx' := Req}, _Context) ->
+    ParseFuns = [parse_map_to_atom_keys(),
+                 read_required_params([account_id, name_id, recipient_id, fee]),
+                 read_optional_params([{ttl, ttl, '$no_value'}]),
+                 base58_decode([{account_id, account_id, {id_hash, [account_pubkey]}},
+                                {recipient_id, recipient_id,
+                                 {id_hash, [account_pubkey, name]}},
```
lucafavatella: I see this means name transfer to account or name. Ok.

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -15,328 +34,399 @@
...
+                ],
+    process_request(ParseFuns, Req);
+
+handle_request('PostOracleQuery', #{'OracleQueryTx' := Req}, _Context) ->
+    ParseFuns = [parse_map_to_atom_keys(),
+                 read_required_params([sender_id, oracle_id, query,
+                                       query_fee, fee, query_ttl, response_ttl]),
+                 read_optional_params([{ttl, ttl, '$no_value'}]),
+                 base58_decode([{sender_id, sender_id, {id_hash, [account_pubkey]}},
+                                {oracle_id, oracle_id, {id_hash, [oracle_pubkey]}}]),
```
lucafavatella: I see this means recipient of query only oracle pubkey - e.g. not name. cc @sennui @zp-sd

lucafavatella: That is in line with https://github.com/aeternity/protocol/blame/f622f21f6b41fb5860fd9f37e8df2d5ea7bdb705/oracles/oracle_transactions.md#L68 so looks ok.

PR #1488, opened 2018-08-27
PT-158827776 API: 4/5 micro and key block prefixes

fabiankrol:
PR is created agains `PT-158827776-3-endpoints-cleanup` branch after merging it i'll change base to master


apps/aehttp/src/ws_int_dispatch.erl
```diff
@@ -82,3 +90,17 @@ do_execute(chain, SubUnSub, SubscribeData)
     end;
 do_execute(_, _, _) -> % a catch all for a prettier error when action is missing
     {error, <<"Missing action">>}.
+
+%% TODO: add a function in base58c to find out what type the hash is?
+decode_block_hash(Hash) ->
```
gorillainduction: Strange type for this function, and it doesn't seem to be used properly.

apps/aehttp/src/ws_int_dispatch.erl
```diff
@@ -34,21 +34,29 @@ do_execute(chain, get, QueryPayload) ->
             #{<<"height">> := Height} ->
                 {aec_chain:get_key_block_by_height(Height), {height, Height}};
             #{<<"hash">> := Hash0} ->
-                {ok, Hash} = aec_base58c:safe_decode(block_hash, Hash0),
+                Hash = decode_block_hash(Hash0),
```
gorillainduction: If the user asks for a specific block type, this information is lost here. What should happen if it is of the other type?

jur0: Since both block types are currently in 1 table, [this info is not lost](https://github.com/aeternity/epoch/blob/master/apps/aecore/src/aec_db.erl#L270). When there are separate tables for key/micro blocks, this part will need refactoring too.

fabiankrol:
rebased

apps/aecore/src/aec_headers.erl
```diff
@@ -311,6 +312,37 @@ serialize_to_map(#mic_header{} = Header) ->
...
+     };
+serialize_for_client(#mic_header{} = Header, PrevBlockType) ->
+    {ok, Hash} = hash_header(Header),
+    #{<<"hash">>       => encode_block_hash(micro, Hash),
+      <<"height">>     => Header#mic_header.height,
+      <<"prev_hash">>  => encode_block_hash(PrevBlockType, Header#mic_header.prev_hash),
+      <<"state_hash">> => aec_base58c:encode(block_state_hash, Header#mic_header.root_hash),
+      <<"txs_hash">>   => aec_base58c:encode(block_tx_hash, Header#mic_header.txs_hash),
+      <<"time">>       => Header#mic_header.time,
+      <<"version">>    => Header#mic_header.version
```
lucafavatella: [Signature](https://github.com/aeternity/protocol/blob/f622f21f6b41fb5860fd9f37e8df2d5ea7bdb705/serializations.md#micro-block-header)?

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -44,45 +44,152 @@
...
+                            BlockType = aec_blocks:type(Block),
+                            PrevBlockType = aec_blocks:type(PrevBlock),
+                            Header = aec_blocks:to_header(Block),
+                            Result =
+                                case BlockType of
+                                    key ->
+                                        aec_headers:serialize_for_client(Header, PrevBlockType);
+                                    micro ->
+                                        R = aec_headers:serialize_for_client(Header, PrevBlockType),
+                                        R#{<<"signature">> => aec_blocks:signature(Block)}
```
lucafavatella: It seems to me this code assumes signature in microblock and not microheader. That seems outdated. Cc @gorillainduction

lucafavatella: Also it seems like microblock signature shall have some sort of encoding? (Not sure if the same of transaction signature.)

apps/aehttp/src/aehttp_helpers.erl
```diff
@@ -525,28 +523,6 @@ get_block_from_chain(Fun) when is_function(Fun, 0) ->
             {404, [], #{reason => <<"Not mining, no pending block">>}}
     end.
 
-get_block(Fun, Req) ->
-    get_block(Fun, Req, true).
-
-get_block(Fun, _Req, AddHash) when is_function(Fun, 0) ->
-    case get_block_from_chain(Fun) of
-        {ok, Block} ->
-            Resp0 = aehttp_logic:cleanup_genesis(
```
lucafavatella: aehttp_logic:cleanup_genesis is now unused and shall be deleted?

Are there tests on whether and how the following field are serialized for the client for the genesis block for the various APIs?
- prev_hash
- miner
- beneficiary
- pow
- nonce

lucafavatella:
> PR is created agains PT-158827776-3-endpoints-cleanup branch after merging it i'll change base to master

I see #1487 merged, I am changing base branch of this to master.

apps/aehttp/src/aehttp_dispatch_ext.erl
fabiankrol: That's correct. `Signature` will be part of `aec_headers:serialize_for_client/2` result.

fabiankrol: Encoding handled in `aec_headers:serialize_for_client/2`

apps/aecore/src/aec_headers.erl
```diff
@@ -311,6 +312,38 @@ serialize_to_map(#mic_header{} = Header) ->
...
+      <<"nonce">>       => Header#key_header.nonce,
+      <<"time">>        => Header#key_header.time,
+      <<"version">>     => Header#key_header.version
+     };
+serialize_for_client(#mic_header{} = Header, PrevBlockType) ->
+    {ok, Hash} = hash_header(Header),
+    #{<<"hash">>       => encode_block_hash(micro, Hash),
+      <<"height">>     => Header#mic_header.height,
+      <<"prev_hash">>  => encode_block_hash(PrevBlockType, Header#mic_header.prev_hash),
+      <<"signature">>  => aec_base58c:encode(signature, Header#mic_header.signature),
```
lucafavatella: I see signature. Ok.

lucafavatella:
aec_blocks:serialize_to_map seems unused though can be delete in separate PR so not to trigger need to re-approve.

apps/aecore/src/aec_headers.erl
```diff
@@ -311,6 +312,38 @@ serialize_to_map(#mic_header{} = Header) ->
...
+serialize_for_client(#key_header{} = Header, PrevBlockType) ->
+    {ok, Hash} = hash_header(Header),
+    #{<<"hash">>        => encode_block_hash(key, Hash),
+      <<"height">>      => Header#key_header.height,
+      <<"prev_hash">>   => encode_block_hash(PrevBlockType, Header#key_header.prev_hash),
+      <<"state_hash">>  => aec_base58c:encode(block_state_hash, Header#key_header.root_hash),
+      <<"miner">>       => aec_base58c:encode(account_pubkey, Header#key_header.miner),
+      <<"beneficiary">> => aec_base58c:encode(account_pubkey, Header#key_header.beneficiary),
+      <<"target">>      => Header#key_header.target,
+      <<"pow">>         => serialize_pow_evidence(Header#key_header.pow_evidence),
```
lucafavatella: I see field of genesis would become 42 zeros. Ok.

apps/aehttp/src/aehttp_helpers.erl
lucafavatella: Could you please add ticket (post-release) tracking adding tests re serialization of genesis in user API? So that we can confidently document this if we want to (e.g. as comment in swagger.yaml).

fabiankrol: PT: https://www.pivotaltracker.com/story/show/160148655

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -44,45 +42,141 @@
...
+                    end
+            end;
+        undefined ->
+            {404, [], #{reason => <<"Block not found">>}}
+    end;
 
-handle_request('GetCurrentKeyBlock', Req, _Context) ->
-    get_block(fun() -> aec_chain:top_key_block() end, Req);
+handle_request('GetCurrentKeyBlock', _Req, _Context) ->
+    case aec_chain:top_key_block() of
```
zp-sd: There seems to be exactly the same logic as in `GetTopBlock`, apart from difference in `aec_chain:top_block()` call. We may want to consider to extract this to some function.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -434,10 +531,34 @@ get_generation(Hash) ->
...
+encode_generation(KeyBlock, MicroBlocks, PrevBlockType) ->
+    Header = aec_blocks:to_header(KeyBlock),
+    #{key_block => aec_headers:serialize_for_client(Header, PrevBlockType),
+      micro_blocks => [begin
+                           {ok, Hash} = aec_blocks:hash_internal_representation(M),
+                           aec_base58c:encode(micro_block_hash, Hash)
+                       end || M <- MicroBlocks]}.
+
+
+deserialize_transaction(Tx) ->
```
zp-sd: 👍

apps/aehttp/src/aehttp_dispatch_ext.erl
fabiankrol: You're right -> future refactor 💪

PR #1493, opened 2018-08-27
PT-157931784 Sophia TTL type

UlfNorell:
Closes [PT-157931784](https://www.pivotaltracker.com/story/show/157931784).

Adds a builtin datatype
```
datatype Chain.ttl = RelativeTTL(int) | FixedTTL(int)
```
and changes the Oracle API to take values of this type instead of plain `int`s.

### DIFF
--- a/apps/aecontract/test/aecontract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_SUITE.erl
@@ -724,19 +724,22 @@ sophia_typed_calls(_Cfg) ->
 %%  - Failing calls
 sophia_oracles(_Cfg) ->
     state(aect_test_utils:new_state()),
+    RelativeTTL       = fun(Delta)  -> {variant, 0, [Delta]} end,
+    FixedTTL          = fun(Height) -> {variant, 1, [Height]} end,
     Acc               = ?call(new_account, 1000000),
     Ct = <<CtId:256>> = ?call(create_contract, Acc, oracles, {}, #{amount => 100000}),
     QueryFee          = 100,
     TTL               = 15,
-    CtId              = ?call(call_contract, Acc, Ct, registerOracle, word, {CtId, 0, QueryFee, TTL}),
+    CtId              = ?call(call_contract, Acc, Ct, registerOracle, word, {CtId, 0, QueryFee, FixedTTL(TTL)}),
     Question          = <<"Manchester United vs Brommapojkarna">>,
-    QId               = ?call(call_contract, Acc, Ct, createQuery, word, {Ct, Question, QueryFee, 5, 5}, #{amount => QueryFee}),
+    QId               = ?call(call_contract, Acc, Ct, createQuery, word,
+                                {Ct, Question, QueryFee, RelativeTTL(5), RelativeTTL(5)}, #{amount => QueryFee}),
     Question          = ?call(call_contract, Acc, Ct, getQuestion, string, {CtId, QId}),
     QueryFee          = ?call(call_contract, Acc, Ct, queryFee, word, Ct),
     none              = ?call(call_contract, Acc, Ct, getAnswer, {option, word}, {CtId, QId}),
     {}                = ?call(call_contract, Acc, Ct, respond, {tuple, []}, {CtId, QId, 0, 4001}),
     {some, 4001}      = ?call(call_contract, Acc, Ct, getAnswer, {option, word}, {CtId, QId}),
-    {}                = ?call(call_contract, Acc, Ct, extendOracle, {tuple, []}, {Ct, 0, TTL + 10}),
+    {}                = ?call(call_contract, Acc, Ct, extendOracle, {tuple, []}, {Ct, 0, FixedTTL(TTL + 10)}),
 
     %% Test complex answers
     Ct1 = ?call(create_contract, Acc, oracles, {}, #{amount => 100000}),
--- a/apps/aecontract/test/aecontract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_SUITE.erl
@@ -852,7 +852,7 @@ ttl_scenario_create_and_extend(Start0, Extend0, Stop0) ->
       || Start <- List(Start0), Extend <- List(Extend0), Stop <- List(Stop0),
          TTLo <- [ T || Extend == false, T <- ttls(Start, [Stop]) ] ++
                  [ T || Extend /= false, T <- ttls(Start, [Extend + 5]) ],
-         TTLe = {delta, _} <- [ 0 || Extend == false ] ++
+         TTLe = {delta, _} <- [ {delta, 0} || Extend == false ] ++
                               [ T || Extend /= false, T <- ttls(ttl_height(Start, TTLo), [Stop]) ] ].
 
 %% Base scenario for failing or unnecessary extends.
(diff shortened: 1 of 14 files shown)


apps/aecore/src/aec_vm_chain.erl
```diff
@@ -193,7 +193,7 @@ oracle_extend(Oracle,_Sign, TTL, State) ->
     {ok, Tx} =
         aeo_extend_tx:new(#{oracle_id  => aec_id:create(oracle, Oracle),
                             nonce      => Nonce,
-                            oracle_ttl => {delta, TTL},
+                            oracle_ttl => TTL,
```
lucafavatella: Extend transaction TTL can only be absolute ([spec](https://github.com/aeternity/protocol/blame/f622f21f6b41fb5860fd9f37e8df2d5ea7bdb705/oracles/oracle_transactions.md#L46), [deserialization](https://github.com/aeternity/epoch/blob/e3296af5398432e937965c81239cffe0a81ac47d/apps/aeoracle/src/aeo_extend_tx.erl#L158)).

If extend from contract can take absolute TTL (as indicated by this line and by [proposed spec update](https://github.com/aeternity/protocol/pull/178/files#diff-9ddeae469112ac9dd44768c45511c6e9R328)), I would expect [internal representation](https://github.com/aeternity/epoch/blob/e3296af5398432e937965c81239cffe0a81ac47d/apps/aeoracle/src/aeo_extend_tx.erl#L40) needs to reflect that. Cc @hanssv  @gorillainduction

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -730,6 +750,170 @@ sophia_oracles(_Cfg) ->
...
+       Query <- List(Query0),
+       QTTL  <- ttls(Query, List(QTTL0)),
+       RTTL  <- List(RTTL0) ].
+
+%% Extending an oracle must be done before expiry.
+ttl_scenario_extend_after_expiry() ->
+    [ [ {10, {create, TTLo}},
+        {20, {extend, TTLe}} ]
+    || TTLo <- ttls(10, [15]),
+       TTLe <- ttls(20, [25]) ].
```
lucafavatella: If I read this correctly, the following cases are not tested:
* Oracle extend VM primop called with absolute expiry TTL strictly smaller than current height;
* Oracle extend VM primop called with absolute expiry TTL equal to current height;
* Oracle extend VM primop called with absolute expiry TTL greater than current height though strictly smaller than planned expiration of oracle;
* Oracle extend VM primop called with absolute expiry TTL greater than current height though equal to planned expiration of oracle.

lucafavatella:
Labelling as WIP so to filter this out.

lucafavatella:
Keeping labelled as WIP, assigning to @UlfNorell for feedback.

apps/aecore/src/aec_vm_chain.erl
UlfNorell: I can certainly create and execute extend transactions with absolute TTLs, and it looks to me like the http endpoint allows it too (`ttl_decode` rather than `relative_ttl_decode`):

https://github.com/aeternity/epoch/blob/40bd3dd9cf1293bcf252881b8610b385c85f1c18/apps/aehttp/src/aehttp_dispatch_int.erl#L372

I'm not changing anything about oracles, just adding a Sophia interface to what's currently implemented. If we need to restrict extend transactions to relative TTLs, I think that should be done as a separate ticket.

apps/aecontract/test/aecontract_SUITE.erl
UlfNorell: I've added tests for this (0adce03).

UlfNorell:
I've updated the protocol PR to say that extend TTL must be relative, but I'm keeping the tests of absolute TTLs in this PR. Once the extend transaction actually enforces relative TTL we can update the tests accordingly.

apps/aecore/src/aec_vm_chain.erl
lucafavatella: > https://github.com/aeternity/epoch/blob/40bd3dd9cf1293bcf252881b8610b385c85f1c18/apps/aehttp/src/aehttp_dispatch_int.erl#L372

Deeper, when serializing tx, I expect that HTTP endpoint fails if TTL is not relative:

https://github.com/aeternity/epoch/blob/4124f84531e40b33857e40c92adcc04865c41630/apps/aeoracle/src/aeo_extend_tx.erl#L144

And when tx is deserialized, relative delta is enforced too:

https://github.com/aeternity/epoch/blob/4124f84531e40b33857e40c92adcc04865c41630/apps/aeoracle/src/aeo_extend_tx.erl#L159

docs/release-notes/RELEASE-NOTES-0.21.0.md
```diff
@@ -6,6 +6,7 @@ It:
 * Makes HTTP API more consistent by using IDs in all blockchain components (accounts, contracts, oracles, ...) and unifying endpoint paths.
 * Cleans up unused HTTP endpoints that signed transactions inside node.
 * Adds support for type aliases and typed contract calls to the Sophia compiler.
+* Makes it possible to use absolute TTLs in Sophia oracle API.
```
lucafavatella: Wrong file.

lucafavatella:
> I've updated the protocol PR to say that extend TTL must be relative, but I'm keeping the tests of absolute TTLs in this PR. Once the extend transaction actually enforces relative TTL we can update the tests accordingly.

@UlfNorell I tested deserialization of extend tx, used when receiving signed tx from either network or user API, enforces relative TTL - see PR #1537 . So it looks to me it shall not be possible to extend oracle from primop using absolute TTL.

I plan to clarify user API too. Up to you on whether you prefer merging this PR (1493) then revise code re absolute TTL in extend primop.

UlfNorell:
I guess the question is who is responsible for checking that the TTL is relative. Compare the query transaction where the response TTL also must be relative. In that case the VM primop doesn't do any checking and transactions with absolute TTL are ruled out by `aeo_query_tx:check/5`. Specifically by

https://github.com/aeternity/epoch/blob/9bcc9bb2e3106a6e7db97de5ba8f7598d71c9753/apps/aeoracle/src/aeo_query.erl#L313

I would be inclined to leave the extend primop as it is and fix `aeo_extend_tx:check/5` in a separate PR.

lucafavatella:
> I guess the question is who is responsible for checking that the TTL is relative. ..

I see. Many `check` callbacks do not check everything, e.g. that integers are non-negative, by relying on deserialization properties. Though aec_vm_chain does not go through deserialization. I tracked the generic review as https://www.pivotaltracker.com/story/show/160350682

I kept PR #1537 for clarification of external boundaries re extend tx TTL.

lucafavatella:
> I would be inclined to leave the extend primop as it is and fix aeo_extend_tx:check/5 in a separate PR.

@UlfNorell  I started work on this follow-up PR - see #1539. Some tests still failing.

test/contract_sophia_bytecode_aevm_SUITE.erl
```diff
@@ -332,28 +332,28 @@ oracles(_Cfg) ->
     Env0 = initial_state(#{}),
     Env1 = create_contract(101, Code, "()", Env0),
     QFee = 100,
-    {101, Env2} = successful_call(101, word, registerOracle, "(101, 3, "++ integer_to_list(QFee) ++", 10)", Env1),
-    {Q, Env3}   = successful_call(101, word, createQuery, "(101, \"why?\", "++ integer_to_list(QFee) ++", 10, 11)", Env2, #{value => QFee}),
+    {101, Env2} = successful_call(101, word, registerOracle, "(101, 3, "++ integer_to_list(QFee) ++", (0, 10))", Env1),
```
lucafavatella: I understand from aeso_icode that 0 means relative. Shall this (and 1 for fixed) be documented in https://github.com/aeternity/protocol/pull/178 e.g. in section "Data memory layout"?

UlfNorell: https://github.com/aeternity/protocol/pull/178/commits/c22456366c7debd9ba53b169b92ed8a9b3e6509b

PR #1496, opened 2018-08-28
Pt 160014862 missing db table

ThomasArts:
Sporadically on CI we seem to not be able to find a database table in the test case aeu_requests. This test is a remainder from early days added to allert us when a started system had errors in its logs.

I cannot understand why this happens. Locally I have run this test with combinations of other tests in an attempt to reproduce it. No luck. I tried reproducing it on CI by pushing it a few times, no luck.

I wrote a completely new test that can replace the old one and also added some more info in the test case to help when the test fails again. In the long run, we can remove the test, but I would like to understand the cause of this.

Made use of this patch to cleanup aec_db a bit.

### DIFF
--- a/apps/aecore/src/aec_db.erl
+++ b/apps/aecore/src/aec_db.erl
@@ -558,25 +558,12 @@ fold_mempool(FunIn, InitAcc) ->
 
 load_database() ->
     lager:debug("load_database()", []),
-    try
-        wait_for_tables()
-    catch
-        error:E ->
-            erlang:error({E, erlang:get_stacktrace()});
-        exit:E ->
-            exit({E, erlang:get_stacktrace()})
-    end.
+    wait_for_tables().
 
 wait_for_tables() ->
     Tabs = mnesia:system_info(tables) -- [schema],
     lager:debug("wait_for_tables (~p)", [Tabs]),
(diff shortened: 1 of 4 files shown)


apps/aecore/src/aec_db.erl
```diff
@@ -588,7 +575,11 @@ wait_for_tables(Tabs, Sofar, Period, Max) when Sofar < Max ->
             wait_for_tables(NotLoaded, Sofar + Period, Period, Max)
     end;
 wait_for_tables(Tabs, Sofar, _, _) ->
-    {timeout, Sofar, Tabs}.
+    lager:error("Tables not loaded after ~p minutes: ~p", [Sofar, Tabs]),
+    %% This is serious and user intervention needed. Stop the system instead 
+    %% of keeping retrying, but also raise an error for the crash log.
+    init:stop(),
+    error({tables_not_loaded, Tabs}).
```
lucafavatella: If `init:stop()` were synchronous, this would not be reached. But it is currently implemented as async, so this code is sometimes reached. And it is improvement from before. Ok.

lucafavatella:
Happened again in recent master https://circleci.com/gh/aeternity/epoch/28827

PR #1503, opened 2018-08-30
Include contract's accont in contract's PoI endpoint

velzevur:
PT [Include contract account in contract proof of inclusion](https://www.pivotaltracker.com/story/show/158967632)


lucafavatella:
Removing WIP label as seems ready for review.

PR #1505, opened 2018-08-30
SC HTTP API: handle broken proof of inclusion

velzevur:
PT [HTTP API shall not crash when PoI is broken](https://www.pivotaltracker.com/story/show/158691726)

### DIFF
--- a/apps/aehttp/src/aehttp_helpers.erl
+++ b/apps/aehttp/src/aehttp_helpers.erl
@@ -356,8 +356,13 @@ parse_map_to_atom_keys() ->
 poi_decode(PoIKey) ->
     fun(_Req, State) ->
         PoIBin = maps:get(PoIKey, State),
-        PoI = aec_trees:deserialize_poi(PoIBin),
-        {ok, maps:put(PoIKey, PoI, State)}
+        try aec_trees:deserialize_poi(PoIBin) of
+            PoI ->
+                {ok, maps:put(PoIKey, PoI, State)}
+        catch
+            _:_ ->
+                {error, {400, [], #{<<"reason">> => <<"Invalid proof of inclusion">>}}}
+        end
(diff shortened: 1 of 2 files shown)


velzevur:
This is not in the scope or release 21 but rather 22. Once release notes for release 22 are planted, I will amend them with a line for this change.
Until then this PR is marked as a `WIP`.

lucafavatella:
I removed WIP label as seems ready to review.

apps/aehttp/src/aehttp_helpers.erl
```diff
@@ -356,8 +356,13 @@ parse_map_to_atom_keys() ->
...
     fun(_Req, State) ->
         PoIBin = maps:get(PoIKey, State),
-        PoI = aec_trees:deserialize_poi(PoIBin),
-        {ok, maps:put(PoIKey, PoI, State)}
+        try aec_trees:deserialize_poi(PoIBin) of
+            PoI ->
+                {ok, maps:put(PoIKey, PoI, State)}
+        catch
+            _:_ ->
+                {error, {400, [], #{<<"reason">> => <<"Invalid proof of inclusion">>}}}
```
jur0: Is the error in swagger spec, too?

velzevur: So this is used by the
* close solo and it has [code 400 handled](https://github.com/aeternity/epoch/blob/master/config/swagger.yaml#L1440)
* slash and it has [code 400 handled](https://github.com/aeternity/epoch/blob/master/config/swagger.yaml#L1465)

It will be used in the force progress once there is an endpoint for it as well :)

velzevur:
A line had been added in the [release notes in wiki](https://github.com/aeternity/epoch/wiki/Release-notes-next)

PR #1507, opened 2018-08-30
Release preparation PR

lucafavatella:
TODO:
* [x] Add line re API changes
* [x] Test locally user documentation


lucafavatella:
Does not conflict with PR #1509 that shall go in too.

PR #1509, opened 2018-08-30
Update configuration docs

fabiankrol:
ref: #1507


lucafavatella:
No way this can break CI.

PR #1516, opened 2018-08-31
Only allow forward generation on chain

hanssv:
Make the call to `get_generation_by_hash` fail if the hash is not on the longest chain - and make this an explicit error in HTTP response.

Also make sure we can validate forward-generations at GenesisHeight.


hanssv:
https://www.pivotaltracker.com/story/show/160089564

PR #1517, opened 2018-08-31
Pt 155700135 Proof of Fraud

sennui:
Implementation of Proof of Fraud. Both generating the event and interpretation on another nodes.

Protocol:
- PoF can be reported only by the next leader, only in the first microblock
- PoF is generated in case of microfork. Microfork is a presence of two microblocks at the same height from the same parent block
- PoF is set to micro header as two headers of microblocks that form microfork described above
- PoF award is given to the leader that reported PoF. It is set now to 1% of coinbase
- Malicious leader that signed microblocks in microfork doesn't get Coinbase and fees

TODO:
- [x] tests that exercise validation code from PoF module 
- [x] swap datastructure for PoF in conductor, so it overwrites frauds from the same malicious leaders (and doesn't grow)


zp-sd:
More TODO: run system tests.

apps/aecore/include/aec_db.hrl
```diff
@@ -5,7 +5,7 @@
 -record(aec_call_state         , {key, value}).
 -record(aec_contract_state     , {key, value}).
 -record(aec_chain_state        , {key, value}).
--record(aec_block_state        , {key, value, difficulty, fork_id, fees, key_hash}).
+-record(aec_block_state        , {key, value, difficulty, fork_id, fees, key_hash, fraud}).
```
zp-sd: Perhaps we may want to introduce separate storage to indicate fraud key blocks, so that aec_block_state remains lightweight? 🤔

apps/aecore/src/aec_chain_state.erl
```diff
@@ -177,9 +179,9 @@ set_top_block_hash(H, State) when is_binary(H) -> State#{top_block_hash => H}.
 %%% Internal ADT for differing between blocks and headers
 %%%-------------------------------------------------------------------
 
--record(node, { header  :: aec_headers:header()
-              , hash    :: binary()
-              , type    :: block_type()
+-record(node, { header :: aec_headers:header()
+                          , hash   :: binary()
+                                      , type   :: block_type()
```
zp-sd: indentation

apps/aecore/src/aec_chain_state.erl
```diff
@@ -898,18 +930,33 @@ db_children(#node{} = Node) ->
...
+    [Header || Header <- aec_db:find_headers_at_height(Height),
+               aec_headers:prev_hash(Header) =:= PrevHash].
+
+db_mark_malicious_leader(Node) ->
+    Height = node_height(Node),
+    {ok, MaliciousKeyBlock} = aec_chain:get_key_block_by_height(Height - 1),
+    MaliciousHeader         = aec_blocks:to_header(MaliciousKeyBlock),
+    {ok, MaliciousHash}     = aec_headers:hash_header(MaliciousHeader),
+
+    {value, Trees, Difficulty, ForkId, Fees, KeyHash, _Fraud}
```
zp-sd: We may want to consider separate structure for fraud key blocks, without extension of `#fork_info{}`.

apps/aecore/src/aec_headers.erl
```diff
@@ -66,6 +68,7 @@
           root_hash    = <<0:?STATE_HASH_BYTES/unit:8>>        :: state_hash(),
           signature    = <<0:?BLOCK_SIGNATURE_BYTES/unit:8>>   :: block_signature(),
           txs_hash     = <<0:?TXS_HASH_BYTES/unit:8>>          :: txs_hash(),
+          pof          = no_fraud                              :: aec_pof:pof(),
```
zp-sd: `aec_pof:pof()` should be a map, and here the type should be `no_fraud | aec_pof:pof()`

apps/aecore/src/aec_governance.erl
```diff
@@ -33,6 +34,7 @@
 -define(TIMESTAMP_MEDIAN_BLOCKS, 11).
 -define(EXPECTED_BLOCK_MINE_RATE, 3 * 60 * 1000). %% 60secs * 1000ms * 3 = 180000msecs
 -define(BLOCK_MINE_REWARD, 10000000000000000000).
+-define(POF_REWARD       , 10000000000000000). %% ?BLOCK_MINE_REWARD / 1000
```
sennui: here something between 1-10% and relative to coinbase

apps/aecore/src/aec_headers.erl
```diff
@@ -222,6 +225,12 @@ set_nonce(Header, Nonce) ->
 set_prev_hash(#key_header{} = H, Hash) -> H#key_header{prev_hash = Hash};
 set_prev_hash(#mic_header{} = H, Hash) -> H#mic_header{prev_hash = Hash}.
 
+-spec set_pof(micro_header(), tuple()) -> micro_header().
+set_pof(#mic_header{} = H, PoF) -> H#mic_header{pof = PoF}.
+
+-spec pof(micro_header()) -> tuple().
```
hanssv: Wrong type

apps/aecore/src/aec_pof.erl
```diff
@@ -0,0 +1,126 @@
...
+    Height1 = aec_headers:height(Header1),
+    Height2 = aec_headers:height(Header2),
+    Prev1 = aec_headers:prev_hash(Header1),
+    Prev2 = aec_headers:prev_hash(Header2),
+
+    if (Height1 =:= Height2) and (Prev1 =:= Prev2) -> ok;
+       true -> {error, not_siblings}
+    end.
+
+check_if_first_microblock(PrevBlock) ->
```
hanssv: This can just be a check if `prev_hash == prev_key_hash` once we have added `prev_key_hash`

apps/aecore/src/aec_conductor.erl
```diff
@@ -854,14 +859,27 @@ handle_add_block(Header, CheckFun, Block, State, Origin) ->
                             maybe_publish_block(Origin, Block),
                             case preempt_if_new_top(State, Origin) of
                                 no_change ->
-                                    {ok, State};
+                                    {ok, State#state{pof = no_fraud}};
```
hanssv: Don't we loose the PoF here? In the case we are not yet the leader but becomes one a couple of micro_blocks later? Not 100% sure though...

zp-sd: Well, solid tests are a must ... 😛

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -1166,6 +1166,129 @@ key_hash_orphaned_chain() ->
...
+    ok = insert_block(KB1),
+
+    %% Write orphaned micro blocks
+    ok = insert_block(MB2),
+    ok = insert_block(MB3),
+
+    %% Write missing micro block, which recognizes a fraud
+    FraudHeader1 = aec_blocks:to_header(MB2),
+    FraudHeader2 = aec_blocks:to_header(MB3),
+    {ok, {FraudHeader2, FraudHeader1}} = insert_block(MB1),
```
zp-sd: This test is failing.

zp-sd: 1) This is a bug that PoF is not recognized in case of delayed micro-blocks.

In the following state with orphaned `M2` and `M3`:
```
K1 -> __ -> M2
         -> M3
```
When `M1` is inserted a fraud should be recognized for `M2` and `M3` headers, but it is not:
```
K1 -> M1 -> M2
       \
         -> M3
```
This is proven in this test.

To fix this we should look for micro-forks when traversing children of an added block -  https://github.com/aeternity/epoch/blob/pt-155700135-pof-v2-bkp/apps/aecore/src/aec_chain_state.erl#L513. But this does not solve the problem that I describe below 😛 

2. Another bug is when a PoF is recognized after an insert of a micro block in the fork, that is not current top.
Given the following state, with `K4` being the top of the chain with the biggest difficulty (conductor is mining of top of `K4`):
```
K1 -> M1 -> K2 -> M3 -> K3 -> K4
       \
         -> K5 -> M4
```
Then assume `M5` arrives, that is a micro-fork on `K5`, but this does not change a top, i.e.:
```
K1 -> M1 -> K2 -> M3 -> K3 -> K4
       \
         -> K5 -> M4
              \
               -> M5
```

Now `aec_chain_state:insert_block/1` returns `{ok, {M4, M5}}` to the conductor, which is a fraud, but not relevant when mining on a top of K4! This could become relevant when chain on top of either `M4` or `M5` gets the biggest difficulty, and takes over.

To fix this either:
1) `aec_chain_state:insert_block/1` should return a key-block of a generation, in which a fraud has been found; then conductor should compare this returned key-block with top key-block of a generation, on which it is mining,

or

2) new db cache `{key-block, {micro-block-1, micro-block-2}}` should be introduced, which will track where exactly a fraud has been found; this case could be then used by conductor to find if there are any frauds recognized in the current generation

@sennui @hanssv @gorillainduction Does this make any sense?

sennui: So, both are fine as we discussed, and tagging fraud with key block is required. The new cache solution may be a bit better in regards to managing bloating state of frauds if they happen. Without the cache conductors state would grow.

zp-sd: I actually realized that `This could become relevant when chain on top of either M4 or M5 gets the biggest difficulty, and takes over.` is not the case we should worry about. If one of these chains takes over, a fraud should have been recognized already by the miners that posted new key-blocks in these chains. So fix proposed in A) should be sufficient.

I will think about it for a while.

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -1060,6 +1060,129 @@ fees_delayed_reward() ->
...
+    ok = write_blocks_to_chain(Chain),
+
+    %% Write micro-block, and recognize a fraud
+    FraudHeader1 = aec_blocks:to_header(MB2),
+    FraudHeader2 = aec_blocks:to_header(MB3),
+    {ok, {FraudHeader2, FraudHeader1}} = insert_block(MB3),
+
+    ok.
+
+pof_delayed_micro_block() ->
```
zp-sd: This test is not correct anymore.

sennui:
Thanks for fixes and refactoring. GH rules don't let me approve it, but I approve ;)

apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -1049,7 +1049,7 @@ prep_initiator(Node) ->
     {ok, PubKey} = rpc(Node, aec_keys, pubkey, []),
     {ok, PrivKey} = rpc(Node, aec_keys, sign_privkey, []),
     ct:log("initiator Pubkey = ~p", [PubKey]),
-    aecore_suite_utils:mine_key_blocks(aecore_suite_utils:node_name(Node), 1),
+    aecore_suite_utils:mine_key_blocks(aecore_suite_utils:node_name(Node), 3),
     ct:log("initiator: 1 block mined on ~p", [Node]),
```
hanssv: 1 -> 3

apps/aecore/src/aec_chain_state.erl
```diff
@@ -112,20 +112,22 @@ median_timestamp(Header) ->
             end
     end.
 
--spec insert_block(aec_blocks:block() | map()) -> 'ok' | {'error', any()}.
+-spec insert_block(aec_blocks:block() | map()) -> 'ok' | {'pof', aec_pof:pof()} | {'error', any()}.
 insert_block(#{ key_block := KeyBlock, micro_blocks := MicroBlocks, dir := forward }) ->
     %% First insert key_block
     case insert_block(KeyBlock, sync) of
```
hanssv: Do we need to handle `{pof, ...}`

apps/aecore/src/aec_chain_state.erl
```diff
@@ -112,20 +112,22 @@ median_timestamp(Header) ->
...
         ok ->
-            lists:foldl(fun(MB, ok) -> insert_block(MB, sync);
+            lists:foldl(fun(MB, ok) -> insert_block(MB);
```
hanssv: `{pof, ...}`?

apps/aecore/src/aec_chain_state.erl
gorillainduction: yes

apps/aecore/src/aec_chain_state.erl
gorillainduction: yes

gorillainduction:
Squashed, rebased, etc.

gorillainduction:
I'm dropping this PR and I'll make a new one to allow @sennui to review properly, and to disallow me to review . I rewrote quite a lot.

PR #1520, opened 2018-08-31
Fix state channel's client issues

velzevur:
PT [Invesitgate aesc_session_noise process relations](https://www.pivotaltracker.com/story/show/159293763)

I've found a couple of cases where one process dies and another one later tries stopping/informing it.

Please check commit messages for more info


velzevur:
Right before merging this PR will add the following line in the [release notes](https://github.com/aeternity/epoch/wiki/Release-notes-next):

* Improves channels' WebSocket client stability ([ref](https://github.com/aeternity/epoch/pull/1520))

PR #1526, opened 2018-09-05
SC: clean up push_amount

velzevur:
PT [SC: Remove push_amount](https://www.pivotaltracker.com/story/show/160287427)

There are some remnants of `push_amount` that had been left hanging in code.
@ThomasArts had a great catch with those.

A protocol change is upcoming.


velzevur:
Since we took a radically different approach (keeping the `push_amount`), I am closing this PR.
Ref - #1531

PR #1529, opened 2018-09-05
Pt 160166518 add prev key hash

gorillainduction:
https://www.pivotaltracker.com/story/show/160166518


gorillainduction:
Restarted tests on spurious error.

apps/aecore/src/aec_block_micro_candidate.erl
```diff
@@ -113,6 +108,10 @@ int_create_block(PrevBlockHash, PrevBlock, KeyBlock, Trees, Txs) ->
     {ExpectedPrevBlockVersion, _} = {aec_blocks:version(PrevBlock),
                                      {expected, ExpectedPrevBlockVersion}},
 
+    PrevKeyHash = case aec_blocks:type(PrevBlock) of
```
hanssv: This logic is duplicated... But only twice, so not very important to fix ;-)

PR #1533, opened 2018-09-06
Fix build docs

dincho:
Install Erlang from source on Ubuntu 16.04.
Erlang solution repository would be better, but we don't support latest erlang
(21) yet.

https://www.pivotaltracker.com/story/show/159649299

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -18,9 +18,26 @@ sudo apt-get -qq update \
 
 Required Erlang OTP version is `20.1`.
 
-A compatible OTP version can be installed from a package both on Ubuntu 16.04 and Ubuntu 18.04:
+#### Ubuntu 18.04
+
 ```bash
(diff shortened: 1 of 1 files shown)


docs/build.md
```diff
@@ -18,11 +18,28 @@ sudo apt-get -qq update \
...
+
 ```bash
 sudo apt-get install erlang
 ```
 
+#### Ubuntu 16.04
+
+Ubuntu 16.04 ships with outdated erlang version. Version 20 can be installed from source:
+
+```bash
```
lucafavatella: I see in line with tested snippet https://github.com/aeternity/docker-builder/blob/58b939c933304330938251cd58e31f6be1978de1/Dockerfile#L10-L17

lucafavatella:
I see two approvals. Merging.

PR #1534, opened 2018-09-06
Fix bug in fragmentation logic

hanssv:
https://www.pivotaltracker.com/story/show/160330133

### DIFF
--- a/apps/aecore/src/aec_peer_connection.erl
+++ b/apps/aecore/src/aec_peer_connection.erl
@@ -1091,30 +1091,41 @@ send_response(S, Type, Response) ->
     Msg = aec_peer_messages:serialize_response(Type, Response),
     send_msg(S, response, Msg).
 
-%% If the message is more than 65533 bytes it won't fit in a single Noise
-%% message and we need to fragment it.
+%% If the message is more than 0xFFFF - 2 - 16 bytes (2 bytes for the length at
+%% Noise protocol level and 16 bytes for the encryption MAC since we are using
+%% ChaChaPoly) it won't fit in a single Noise message and we need to fragment
+%% it.
 %%
 %% Fragments have the format <<?MSG_FRAGMENT:16, N:16, M:16,
-%% PayLoad:65529/binary>> (where the last fragment has a smaller Payload),
-%% saying that this is fragment N out of M fragments.
+%% PayLoad:65527/binary>> (where the last fragment has a smaller Payload, and
+%% where 16 bytes of the payload is the encryption MAC), saying that this is
+%% fragment N out of M fragments.
 %%
-%% For testing purpose - set the max packet size to 16#FF while testing
-
+%% For testing purpose - set the max packet size to 16#1FF while testing
+-define(NOISE_PACKET_LENGTH_SIZE, 2).
 -ifndef(TEST).
--define(MAX_PACKET_SIZE, 16#FFFF).
+-define(MAX_PACKET_SIZE, 16#FFFF - ?NOISE_PACKET_LENGTH_SIZE).
 -else.
--define(MAX_PACKET_SIZE, 16#1FF).
+-define(MAX_PACKET_SIZE, 16#1FF - ?NOISE_PACKET_LENGTH_SIZE).
 -endif.
--define(FRAGMENT_SIZE, (?MAX_PACKET_SIZE - 6)).
-do_send(ESock, Msg) when byte_size(Msg) < ?MAX_PACKET_SIZE - 2 ->
+
+-define(CHACHAPOLY_MAC_SIZE, 16).
+-define(MAX_PAYLOAD_SIZE, (?MAX_PACKET_SIZE - ?CHACHAPOLY_MAC_SIZE)).
+
+-define(FRAGMENT_HEADER_SIZE, 2 + 2 + 2).
+-define(MAX_FRAGMENT_PAYLOAD_SIZE,
+        (?MAX_PACKET_SIZE - ?FRAGMENT_HEADER_SIZE - ?CHACHAPOLY_MAC_SIZE)).
+
+do_send(ESock, Msg) when byte_size(Msg) =< ?MAX_PAYLOAD_SIZE ->
     enoise_send(ESock, Msg);
 do_send(ESock, Msg) ->
-    NChunks = (?FRAGMENT_SIZE + byte_size(Msg) - 1) div ?FRAGMENT_SIZE,
+    NChunks = (?MAX_FRAGMENT_PAYLOAD_SIZE + byte_size(Msg) - 1)
+                div ?MAX_FRAGMENT_PAYLOAD_SIZE,
     send_chunks(ESock, 1, NChunks, Msg).
 
 send_chunks(ESock, M, M, Msg) ->
     enoise:send(ESock, <<?MSG_FRAGMENT:16, M:16, M:16, Msg/binary>>);
-send_chunks(ESock, N, M, <<Chunk:?FRAGMENT_SIZE/binary, Rest/binary>>) ->
+send_chunks(ESock, N, M, <<Chunk:?MAX_FRAGMENT_PAYLOAD_SIZE/binary, Rest/binary>>) ->
     enoise_send(ESock, <<?MSG_FRAGMENT:16, N:16, M:16, Chunk/binary>>),
     send_chunks(ESock, N + 1, M, Rest).
 


apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1091,22 +1091,24 @@ send_response(S, Type, Response) ->
     Msg = aec_peer_messages:serialize_response(Type, Response),
     send_msg(S, response, Msg).
 
-%% If the message is more than 65533 bytes it won't fit in a single Noise
-%% message and we need to fragment it.
+%% If the message is more than 65535 - 2 - 16 bytes (2 bytes for the length and
```
lucafavatella: 65535 is 0xFFFF

hanssv: Yes, I was debating with myself if I was to change the representation.

lucafavatella: The 16 comes from [here](https://github.com/aeternity/enoise/blob/71aac15db7b934e0aecac87ff0175abf40aaa2f2/src/enoise_connection.erl#L177)? Maybe in a next PR the hardcoded numbers may be replaced with calls to some functions exported by enoise_connection for readability.

lucafavatella: Re 16 - I see 16 bytes mentioned in [doc](https://noiseprotocol.org/noise.html#message-format). I see [MAX_AD_LEN](https://github.com/aeternity/enoise/blob/71aac15db7b934e0aecac87ff0175abf40aaa2f2/include/enoise.hrl#L2) though unused.

hanssv: Yes, good suggestion except then I can't use it as a guard 🙄 Perhaps some local constants with a reference to the noise protocol? (This 2 byte length is a top level protocol detail.)

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1091,22 +1091,24 @@ send_response(S, Type, Response) ->
...
+%% 16 bytes for the encryption MAC) it won't fit in a single Noise message and
+%% we need to fragment it.
 %%
 %% Fragments have the format <<?MSG_FRAGMENT:16, N:16, M:16,
```
lucafavatella: I see these are 3 couples of bytes. Total 6 bytes. Plus 65529 leads to 65535. Ok.

apps/aecore/src/aec_peer_connection.erl
hanssv: I'll make a note of these and try polish up Enoise as part of https://www.pivotaltracker.com/story/show/157143024

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1091,22 +1091,24 @@ send_response(S, Type, Response) ->
...
 -else.
 -define(MAX_PACKET_SIZE, 16#1FF).
 -endif.
--define(FRAGMENT_SIZE, (?MAX_PACKET_SIZE - 6)).
-do_send(ESock, Msg) when byte_size(Msg) < ?MAX_PACKET_SIZE - 2 ->
+-define(FRAGMENT_SIZE, (?MAX_PACKET_SIZE - 6 - 16)).
+do_send(ESock, Msg) when byte_size(Msg) < ?MAX_PACKET_SIZE - 2 - 16 ->
     enoise_send(ESock, Msg);
 do_send(ESock, Msg) ->
     NChunks = (?FRAGMENT_SIZE + byte_size(Msg) - 1) div ?FRAGMENT_SIZE,
```
lucafavatella: At line 1118 - is it meant to enoise:send or enoise_send ?

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1091,22 +1091,24 @@ send_response(S, Type, Response) ->
...
-%% If the message is more than 65533 bytes it won't fit in a single Noise
-%% message and we need to fragment it.
+%% If the message is more than 65535 - 2 - 16 bytes (2 bytes for the length and
+%% 16 bytes for the encryption MAC) it won't fit in a single Noise message and
+%% we need to fragment it.
 %%
 %% Fragments have the format <<?MSG_FRAGMENT:16, N:16, M:16,
-%% PayLoad:65529/binary>> (where the last fragment has a smaller Payload),
-%% saying that this is fragment N out of M fragments.
+%% PayLoad:65529/binary>> (where the last fragment has a smaller Payload, and
```
lucafavatella: Shall not each fragment consider `- 2 - 16`?

apps/aecore/src/aec_peer_connection.erl
hanssv: `enoise_send` wrapper does some context aware error handling

apps/aecore/src/aec_peer_connection.erl
lucafavatella: > where 16 bytes of the payload is the encryption MAC

I see 16 considered. Not sure about 2.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1091,22 +1091,24 @@ send_response(S, Type, Response) ->
...
 
 -ifndef(TEST).
 -define(MAX_PACKET_SIZE, 16#FFFF).
 -else.
 -define(MAX_PACKET_SIZE, 16#1FF).
 -endif.
--define(FRAGMENT_SIZE, (?MAX_PACKET_SIZE - 6)).
-do_send(ESock, Msg) when byte_size(Msg) < ?MAX_PACKET_SIZE - 2 ->
+-define(FRAGMENT_SIZE, (?MAX_PACKET_SIZE - 6 - 16)).
+do_send(ESock, Msg) when byte_size(Msg) < ?MAX_PACKET_SIZE - 2 - 16 ->
```
lucafavatella: This could become `=<` rather than `<` - right?

apps/aecore/src/aec_peer_connection.erl
hanssv: Yes you are right, I think the ` - 2` should really go into `MAX_PACKET_SIZE` that would make the description better and the code easier to follow?!

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1091,22 +1091,24 @@ send_response(S, Type, Response) ->
...
+%% For testing purpose - set the max packet size to 16#1FF while testing
 
 -ifndef(TEST).
 -define(MAX_PACKET_SIZE, 16#FFFF).
 -else.
 -define(MAX_PACKET_SIZE, 16#1FF).
 -endif.
--define(FRAGMENT_SIZE, (?MAX_PACKET_SIZE - 6)).
-do_send(ESock, Msg) when byte_size(Msg) < ?MAX_PACKET_SIZE - 2 ->
+-define(FRAGMENT_SIZE, (?MAX_PACKET_SIZE - 6 - 16)).
```
lucafavatella: It looks to me this shall have `-2` as [2 bytes length added by enoise](https://github.com/aeternity/enoise/blob/71aac15db7b934e0aecac87ff0175abf40aaa2f2/src/enoise_connection.erl#L177).

apps/aecore/src/aec_peer_connection.erl
hanssv: Yes indeed! Will update.

apps/aecore/src/aec_peer_connection.erl
lucafavatella: Yes.

lucafavatella: I see 65529 now 65527. Ok.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1091,30 +1091,41 @@ send_response(S, Type, Response) ->
...
+
+-define(FRAGMENT_HEADER_SIZE, 2 + 2 + 2).
+-define(MAX_FRAGMENT_PAYLOAD_SIZE,
+        (?MAX_PACKET_SIZE - ?FRAGMENT_HEADER_SIZE - ?CHACHAPOLY_MAC_SIZE)).
+
+do_send(ESock, Msg) when byte_size(Msg) =< ?MAX_PAYLOAD_SIZE ->
     enoise_send(ESock, Msg);
 do_send(ESock, Msg) ->
-    NChunks = (?FRAGMENT_SIZE + byte_size(Msg) - 1) div ?FRAGMENT_SIZE,
+    NChunks = (?MAX_FRAGMENT_PAYLOAD_SIZE + byte_size(Msg) - 1)
```
lucafavatella: I am somehow not convinced by this assignment.

lucafavatella: Where does `byte_size(Msg)` come from? It seems to me it is not serialized, and not expected in receiving.

lucafavatella: I would expect `NChunks = (byte_size(Msg) div ?MAX_FRAGMENT_PAYLOAD_SIZE) + case (byte_size(Msg) rem ?MAX_FRAGMENT_PAYLOAD_SIZE) of 0 -> 0; X when X > 0 -> 1 end`, not sure about this line.

lucafavatella: I see equivalent.

PR #1536, opened 2018-09-07
PT-159224343 Return key blocks to mine via HTTP API

jur0:


apps/aecore/src/aec_conductor.erl
```diff
@@ -209,8 +205,8 @@ handle_call(stop_mining,_From, State = #state{ consensus = Cons }) ->
     epoch_mining:info("Mining stopped"),
     [ aec_tx_pool:garbage_collect() || is_record(Cons, consensus) andalso Cons#consensus.leader ],
     State1 = kill_all_workers(State),
-    {reply, ok, State1#state{mining_state = 'stopped',
-                             key_block_candidate = undefined}};
+    State2 = State1#state{mining_state = stopped},
```
hanssv: I think you want to set `key_block_candidate = undefined` here or else you will return a stale key-block while creating the new one?

apps/aecore/src/aec_conductor.erl
```diff
@@ -508,9 +504,8 @@ preempt_if_new_top(#state{ top_block_hash = OldHash,
                                     micro -> KeyHash;
                                     key   -> NewHash
                                 end,
-                    State5 = State4#state{ top_key_block_hash = NewTopKey,
-                                           key_block_candidate = undefined },
-                    {changed, NewBlock, State5}
+                    State5 = State4#state{ top_key_block_hash = NewTopKey},
```
hanssv: same as above

PR #1538, opened 2018-09-07
Replace tabs with spaces

velzevur:
We had some tabs. Since our code policy is identation with spaces, I've fixed those.


velzevur:
Rebased on top of code frozen master + some small manual tweaks

PR #1539, opened 2018-09-07
PT-157931784 Sophia TTL type -- relative oracle extension TTL

lucafavatella:
TODO:
* [x] Fix tests in aecontract_SUITE using domain-specific language for defining test cases and state machine for executing them. @UlfNorell Could you please help adapting those?

### DIFF
--- a/apps/aecontract/test/aecontract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_SUITE.erl
@@ -803,8 +803,9 @@ step_ttl(St = #{ account := Acc, contract := Ct }, Height, Cmd) ->
             #{ oracle := Oracle, oracle_ttl := TTLo } = St,
             Res = ?call(call_contract, Acc, Ct, extendOracle, {tuple, []}, {Oracle, 0, Enc(TTL)}, #{ height => Height }),
             NewTTLo = ttl_height(TTLo, TTL),
-            %% Can't extend if expired, and new expiry must be > old expiry
-            case TTLo >= Height andalso NewTTLo > TTLo of
+            %% Can't extend if expired, and new expiry must be > old expiry,
+            %% and TTL must be relative.
+            case TTLo >= Height andalso NewTTLo > TTLo andalso element(1, TTL) == delta of
                 true  -> {} = Res, St#{ oracle_ttl => NewTTLo }; %% Extend relative to previous expiry
                 false -> Error = Res, St
             end;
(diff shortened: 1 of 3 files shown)


apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -758,7 +758,7 @@ sophia_oracles(_Cfg) ->
     none              = ?call(call_contract, Acc, Ct, getAnswer, {option, word}, {CtId, QId}),
     {}                = ?call(call_contract, Acc, Ct, respond, {tuple, []}, {CtId, QId, 0, 4001}),
     {some, 4001}      = ?call(call_contract, Acc, Ct, getAnswer, {option, word}, {CtId, QId}),
-    {}                = ?call(call_contract, Acc, Ct, extendOracle, {tuple, []}, {Ct, 0, FixedTTL(TTL + 10)}),
+    {}                = ?call(call_contract, Acc, Ct, extendOracle, {tuple, []}, {Ct, 0, RelativeTTL(TTL + 10)}),
```
UlfNorell: Should be `RelativeTTL(10)`

hanssv:
Though dialyzer isn't happy...

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -851,8 +852,8 @@ ttl_scenario_create_and_extend(Start0, Extend0, Stop0) ->
       || Start <- List(Start0), Extend <- List(Extend0), Stop <- List(Stop0),
          TTLo <- [ T || Extend == false, T <- ttls(Start, [Stop]) ] ++
                  [ T || Extend /= false, T <- ttls(Start, [Extend + 5]) ],
-         TTLe <- [ 0 || Extend == false ] ++
-                 [ T || Extend /= false, T <- ttls(ttl_height(Start, TTLo), [Stop]) ] ].
+         TTLe = {delta, _} <- [ 0 || Extend == false ] ++
```
lucafavatella: I see this filters only the ones matching `{delta, _}` hence leaving out the ones matching `{block, _}`. Ok.

UlfNorell: Oh, but the `[ 0 || Extend == false ]` will never match, so we won't get any test cases that doesn't extend. This should be `[ {delta, 0} || Extend == false ]` instead.

lucafavatella:
This shall be ready now. CI running.

PR #1541, opened 2018-09-10
Validate light micro blocks before assembly

hanssv:
When we receive a light micro block via gossip, we better make some
validation before starting to put together a full block.

Validate:
  - micro block header
  - micro block connected to chain
  - micro block height within "gossip_allowed_height_from_top"
  - micro block signature

https://www.pivotaltracker.com/story/show/160093084

### DIFF
--- a/apps/aecore/src/aec_peer_connection.erl
+++ b/apps/aecore/src/aec_peer_connection.erl
@@ -954,22 +954,33 @@ handle_new_micro_block(S, Msg) ->
                 Header = aec_blocks:to_header(MicroBlock),
                 {ok, HH} = aec_headers:hash_header(Header),
                 epoch_sync:debug("Got new block: ~s", [pp(HH)]),
+                %% in the full micro block case - conductor will do the necessary checks
                 aec_conductor:post_block(MicroBlock);
             true ->
                 {ok, #{ header := Header, tx_hashes := TxHashes }} =
                     deserialize_light_micro_block(maps:get(micro_block, Msg)),
-                case get_micro_block_txs(TxHashes) of
-                    {all, Txs} ->
-                        MicroBlock = aec_blocks:new_micro_from_header(Header, Txs),
+                %% Before assembling the block, check if it is known, and valid
+                case pre_assembly_check(Header) of
+                    known ->
+                        ok;
+                    E = {error, _} ->
                         {ok, HH} = aec_headers:hash_header(Header),
-                        epoch_sync:debug("Got new block: ~s", [pp(HH)]),
-                        aec_conductor:post_block(MicroBlock);
-                    {some, TxsAndTxHashes} ->
-                        epoch_sync:info("Missing txs: ~p",
-                                        [[ TH || TH <- TxsAndTxHashes, is_binary(TH) ]]),
-                        cast(self(), {expand_micro_block, #{ header => Header,
-                                                             tx_data => TxsAndTxHashes }}),
-                        ok
+                        epoch_sync:info("Got invalid light micro_block (~s): ~p", [pp(HH), E]),
+                        ok;
+                    ok ->
+                        case get_micro_block_txs(TxHashes) of
+                            {all, Txs} ->
+                                MicroBlock = aec_blocks:new_micro_from_header(Header, Txs),
+                                {ok, HH} = aec_headers:hash_header(Header),
+                                epoch_sync:debug("Got new block: ~s", [pp(HH)]),
+                                aec_conductor:post_block(MicroBlock);
+                            {some, TxsAndTxHashes} ->
+                                epoch_sync:info("Missing txs: ~p",
+                                                [[ TH || TH <- TxsAndTxHashes, is_binary(TH) ]]),
+                                cast(self(), {expand_micro_block, #{ header => Header,
+                                                                     tx_data => TxsAndTxHashes }}),
+                                ok
+                        end
                 end
         end
     catch _:_ ->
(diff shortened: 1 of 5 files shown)


apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1234,3 +1245,51 @@ get_micro_block_txs([TxHash | TxHashes], State, Acc) ->
...
+
+pre_assembly_check(MicroHeader) ->
+    {ok, Hash} = aec_headers:hash_header(MicroHeader),
+    case aec_db:has_block(Hash) of
+        true -> known;
+        false ->
+            Validators = [fun validate_micro/1,
+                          fun validate_connected_to_chain/1,
+                          fun validate_delta_height/1,
+                          fun validate_micro_signature/1],
```
gorillainduction: You might as well check that the previous key hash is correct as well.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1234,3 +1245,51 @@ get_micro_block_txs([TxHash | TxHashes], State, Acc) ->
...
+                false -> {error, too_far_below_top}
+            end
+    end.
+
+validate_micro(MicroHeader) ->
+    aec_headers:validate_micro_block_header(MicroHeader).
+
+validate_micro_signature(MicroHeader) ->
+    case aec_db:find_header(aec_headers:prev_key_hash(MicroHeader)) of
+        none -> {error, signer_not_found}; %% This is virtually impossible!
```
gorillainduction: It is not impossible since you didn't check that the prev_key_hash is pointing correctly. It is impossible if you do check that.

hanssv: Fixed
|
|
v

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1234,3 +1245,72 @@ get_micro_block_txs([TxHash | TxHashes], State, Acc) ->
...
+                false -> {error, too_far_below_top}
+            end
+    end.
+
+validate_prev_key_block(MicroHeader) ->
+    case aec_db:find_header(aec_headers:prev_hash(MicroHeader)) of
+        {value, PrevHeader} ->
+            case aec_headers:type(PrevHeader) of
+                key ->
+                    {ok, PrevHash} = aec_header:hash_header(PrevHeader),
```
gorillainduction: I think you can trust the db, and you don't need to check the hash again. If there is a keyblock, then it is ok.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1234,3 +1245,77 @@ get_micro_block_txs([TxHash | TxHashes], State, Acc) ->
         none         -> get_micro_block_txs(TxHashes, some, [TxHash | Acc]);
         {value, STx} -> get_micro_block_txs(TxHashes, State, [STx | Acc])
     end.
+
+pre_assembly_check(MicroHeader) ->
+    {ok, Hash} = aec_headers:hash_header(MicroHeader),
+    case aec_db:has_block(Hash) of
+        true -> known;
+        false ->
+            Validators = [fun validate_micro/1,
```
lucafavatella: I see this function runs check `aec_headers:validate_micro_block_cycle_time` that skips (!) check if prev hash not in DB. I assume following check `validate_connected_to_chain` does check that prev hash is in DB (it runs `aec_chain_state:hash_is_connected_to_genesis(aec_headers:prev_hash(...))` - I did not check that this actually checks that prev hash is in DB). So it looks fine but, as DB updates may be performed while this function runs, there is slight possibility that `validate_micro` check does not find prev hash in DB (hence passes without checking time) then concurrently prev hash is inserted in DB so when check `validate_connected_to_chain` runs it finds the prev hash. This leaves the timestamp not fully checked.

I consider this a minor issue though it hinders a bit readability. Any strong opinions on swapping `validate_micro` and `validate_connected_to_chain`?

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1234,3 +1245,77 @@ get_micro_block_txs([TxHash | TxHashes], State, Acc) ->
...
+                true  -> ok;
+                false -> {error, too_far_below_top}
+            end
+    end.
+
+validate_prev_key_block(MicroHeader) ->
+    case aec_db:find_header(aec_headers:prev_hash(MicroHeader)) of
+        {value, PrevHeader} ->
+            case aec_headers:height(PrevHeader) == aec_headers:height(MicroHeader) of
+                false -> {error, wrong_prev_key_block_height};
```
lucafavatella: I see PrevHeader may be a microblock header - as opposed to key header - though as it is in DB it is assumed to be already checked to have the same height as its preceding key block. So the error atom is not misleading. Ok.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -954,22 +954,33 @@ handle_new_micro_block(S, Msg) ->
...
                 epoch_sync:debug("Got new block: ~s", [pp(HH)]),
+                %% in the full micro block case - conductor will do the necessary checks
                 aec_conductor:post_block(MicroBlock);
             true ->
                 {ok, #{ header := Header, tx_hashes := TxHashes }} =
                     deserialize_light_micro_block(maps:get(micro_block, Msg)),
-                case get_micro_block_txs(TxHashes) of
-                    {all, Txs} ->
-                        MicroBlock = aec_blocks:new_micro_from_header(Header, Txs),
+                %% Before assembling the block, check if it is known, and valid
```
lucafavatella: I see the original description of https://www.pivotaltracker.com/n/projects/2124891/stories/160093084 was:
> The process representing a connection with a peer inserts a micro-block expansion internal message keyed by header hash {expand, Hash}. This request is queued following (unsolicited) request (not response) by peer after only structural parsing (not sanitized).

I see this PR addresses this by performing stricter checks for gossiped micro blocks: only microblocks at recent enough height trigger this process, and even those actors able to do so (for signature) are bound to timestamp. Ok.

apps/aecore/src/aec_peer_connection.erl
hanssv: As it is now the timestamp will be checked also in `aec_chain_state:insert_block` (`assert_...`) - so no risk... But I have discussed with Tobias to later remove some redundant checks and then we should be careful not missing this check!

PR #1542, opened 2018-09-10
PT-159224345 Add POST /key-blocks endpoint

jur0:
[PT-159224345](https://www.pivotaltracker.com/n/projects/2124891/stories/159224345)

### DIFF
--- a/apps/aecore/src/aec_headers.erl
+++ b/apps/aecore/src/aec_headers.erl
@@ -12,6 +12,7 @@
          assert_micro_header/1,
          beneficiary/1,
          deserialize_from_binary/1,
+         deserialize_from_client/2,
          deserialize_from_map/1,
          deserialize_key_from_binary/1,
          deserialize_micro_from_binary/1,
@@ -368,6 +369,24 @@ encode_block_hash(key, Hash) ->
 encode_block_hash(micro, Hash) ->
     aec_base58c:encode(micro_block_hash, Hash).
 
+-spec deserialize_from_client(key, map()) -> {ok, header()} | {error, term()}.
+deserialize_from_client(key, KeyBlock) ->
+    try
+        {ok, #key_header{height       = maps:get(<<"height">>, KeyBlock),
+                         prev_hash    = decode_block_hash(maps:get(<<"prev_hash">>, KeyBlock)),
+                         prev_key     = decode(key_block_hash, maps:get(<<"prev_key_hash">>, KeyBlock)),
+                         root_hash    = decode(block_state_hash, maps:get(<<"state_hash">>, KeyBlock)),
+                         miner        = decode(account_pubkey, maps:get(<<"miner">>, KeyBlock)),
+                         beneficiary  = decode(account_pubkey, maps:get(<<"beneficiary">>, KeyBlock)),
+                         target       = maps:get(<<"target">>, KeyBlock),
+                         pow_evidence = deserialize_pow_evidence(maps:get(<<"pow">>, KeyBlock)),
+                         nonce        = maps:get(<<"nonce">>, KeyBlock),
+                         time         = maps:get(<<"time">>, KeyBlock),
+                         version      = maps:get(<<"version">>, KeyBlock)}}
+    catch
+        _:_ -> {error, invalid_header}
+    end.
+
 -spec deserialize_from_map(map()) -> {'ok', header()} | {'error', term()}.
 deserialize_from_map(#{<<"height">> := Height,
                        <<"prev_hash">> := PrevHash,
@@ -648,3 +667,22 @@ validate_max_time({Header, _}) ->
         false ->
             {error, block_from_the_future}
     end.
+
+%% TODO: this is taken from ws_int_dispatch.erl - needs refactoring after
+%% separation of key/micro block records.
+decode_block_hash(Hash) ->
+    case aec_base58c:safe_decode(key_block_hash, Hash) of
+        {ok, KeyBlockHash} ->
+            KeyBlockHash;
+        {error, _Rsn} ->
+            case aec_base58c:safe_decode(micro_block_hash, Hash) of
+                {ok, MicroBlockHash} ->
+                    MicroBlockHash;
+                {error, _Rsn} = Err ->
+                    Err
+            end
+    end.
+
+decode(Type, Enc) ->
+    {ok, Val} = aec_base58c:safe_decode(Type, Enc),
+    Val.
--- a/apps/aecore/src/aec_headers.erl
+++ b/apps/aecore/src/aec_headers.erl
@@ -373,7 +373,7 @@ encode_block_hash(micro, Hash) ->
 deserialize_from_client(key, KeyBlock) ->
     try
         {ok, #key_header{height       = maps:get(<<"height">>, KeyBlock),
-                         prev_hash    = decode_block_hash(maps:get(<<"prev_hash">>, KeyBlock)),
+                         prev_hash    = decode(block_hash, maps:get(<<"prev_hash">>, KeyBlock)),
                          prev_key     = decode(key_block_hash, maps:get(<<"prev_key_hash">>, KeyBlock)),
                          root_hash    = decode(block_state_hash, maps:get(<<"state_hash">>, KeyBlock)),
                          miner        = decode(account_pubkey, maps:get(<<"miner">>, KeyBlock)),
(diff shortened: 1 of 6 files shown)


apps/aecore/src/aec_headers.erl
```diff
@@ -648,3 +667,22 @@ validate_max_time({Header, _}) ->
         false ->
             {error, block_from_the_future}
     end.
+
+%% TODO: this is taken from ws_int_dispatch.erl - needs refactoring after
+%% separation of key/micro block records.
```
hanssv: They are separated already?!

apps/aecore/src/aec_headers.erl
```diff
@@ -368,6 +369,24 @@ encode_block_hash(key, Hash) ->
 encode_block_hash(micro, Hash) ->
     aec_base58c:encode(micro_block_hash, Hash).
 
+-spec deserialize_from_client(key, map()) -> {ok, header()} | {error, term()}.
+deserialize_from_client(key, KeyBlock) ->
+    try
+        {ok, #key_header{height       = maps:get(<<"height">>, KeyBlock),
+                         prev_hash    = decode_block_hash(maps:get(<<"prev_hash">>, KeyBlock)),
```
ThomasArts: The function decode_block_hash is a nested base58c safe_decode on first key and then micro blocks. It would be better to add a safe_decode(block, Binary) to base58c that does exactly this. Here one could then use that function. NOTE that you should not (as the comment suggests) look at the first two digits to decide which it is, because safe_decode enforces the thing to be correct.

apps/aecore/src/aec_headers.erl
```diff
@@ -648,3 +667,22 @@ validate_max_time({Header, _}) ->
         false ->
             {error, block_from_the_future}
     end.
+
+%% TODO: this is taken from ws_int_dispatch.erl - needs refactoring after
+%% separation of key/micro block records.
+decode_block_hash(Hash) ->
```
ThomasArts: Move to base58c as special safe_decode and let it return {ok, ....}. Then the local decode can take away the ok.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -163,6 +163,20 @@ handle_request('GetKeyBlockByHeight', Params, _Context) ->
             {404, [], #{reason => <<"Block not found">>}}
     end;
 
+handle_request('PostKeyBlock', #{'KeyBlock' := Data}, _Context) ->
```
ThomasArts: To me this seems like an attack vector. We have tried to avoid HTTP spamming by using Noise as protocol to exchange blocks. It cost to, probably little, but something to validate these blocks. In particular invalid blocks take a lot of CPU cycles (due to throwing exceptions deep down in the code being rather expensive).

We need a discussion on whether this really is a good idea.

apps/aecore/src/aec_headers.erl
jur0: There seems to be more work done to separate them. Both block types use the same db table. When reading a block, we find out what type it is from the [returned value](https://github.com/aeternity/epoch/blob/master/apps/aecore/src/aec_db.erl#L237).

apps/aehttp/src/aehttp_dispatch_ext.erl
jur0: There used to be the same endpoint which was removed. Now, it's added back as it will allow creation of mining pools. The mining pool manager creates a key block (but doesn't mine) and the miners can request those via `/key-blocks/pending` (`pow` and `nonce` are not set). Miners then mine the key block, once a block is mined, they can submit the block by posting it to `/key-blocks`. Reward distribution and different mining strategies are to be developed by the community.

config/swagger.yaml
```diff
@@ -204,6 +204,32 @@ paths:
           schema:
             $ref: '#/definitions/Error'
       security:
+  /key-blocks:
+    post:
+      tags:
+        - external
```
lucafavatella: I see this is external, not internal.

jur0: Do you prefer `internal`? I think it should be external so the miners can submit the blocks. I'd probably add a new tag `mining` or similar a let the node operator decide if the endpoint should be allowed.

apps/aehttp/src/aehttp_dispatch_ext.erl
z1617: Miners only solve the puzzle delivered by the pool manger,  if the hash meets the pool difficulty, they get a share,  if the hash also meets the AE network difficulty, the block has been successfully mined, in this case the pool submits the block to the AE network and everyone with shares gets paid accordingly. This article is great: [Bitcoin mining the hard way](http://www.righto.com/2014/02/bitcoin-mining-hard-way-algorithms.html).

If I'm right, only the pool manager request `/key-blocks[/pending]`, there is no need for miners to request this API?

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -1108,6 +1104,19 @@ get_key_block_by_height(micro_block, Config) ->
...
+post_key_block(Config) ->
+    post_key_block(?config(current_block_type, Config), Config).
+
+post_key_block(_CurrentBlockType, Config) ->
+    {ok, 200, PendingKeyBlock} = get_key_blocks_pending_sut(),
+    PowEvidence = lists:duplicate(42, 1),
+    Nonce = 1,
+    KeyBlock = PendingKeyBlock#{<<"pow">> => PowEvidence, <<"nonce">> => Nonce},
+    {ok, 400, Error} = post_key_blocks_sut(KeyBlock),
+    %% Block is always rejected - pow and nonce are not correct.
```
lucafavatella: A test for the positive case (by using the user HTTP API - i.e. fetch pending key block header, find PoW, post mined key block header) would be beneficial for ensuring use case not broken in future changes. May be addressed in following PR/ticket.

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -32,6 +32,20 @@
         Context :: #{}
                    ) -> {Status :: cowboy:http_status(), Headers :: list(), Body :: map()}.
 
+handle_request('PostKeyBlock', #{'KeyBlock' := Data}, _Context) ->
+    case aec_headers:deserialize_from_client(key, Data) of
+        {ok, Header} ->
+            KeyBlock = aec_blocks:new_key_from_header(Header),
+            case aec_conductor:post_block(KeyBlock) of
```
lucafavatella: I see that the key block does not have more info than the key block header, and - based on code inspection - the DB insertion of the posted block is ok (including recomputation of state trees with pruned items). Ok.

lucafavatella: I see - based on code inspection - that the conductor clears and updates the key block candidate (pending block) after the mined block is posted. Ok.

lucafavatella:
I see Python UAT tests [CI job failure](https://circleci.com/gh/aeternity/epoch/30686) is unrelated known intermittent test failure (not ideal) - restarting.
```
integration.test_spend_tx.test_not_enough_tokens ... 
Node dev1 starting
Starting node with config prefix EPOCH_CONFIG="/tmp/tmp5lXL4r/epoch.yaml" 
Bob initial balance is 30000000000000000000 and he will try to spend 1000000000000000000020
Bob's spend_tx is {'amount': 1000000000000000000000L,
 'fee': 20,
 'nonce': None,
 'payload': 'foo',
 'recipient_id': 'ak$juZ2L9zGwFnBAzTWjaix65QifTRjKa5uxpuffRN3Zn2BVyddB',
 'sender_id': 'ak$2rYVpohWXGCMAR9ugaBNQ5F8xuUXRCrkSvnThArPUB7sVNrRhN',
 'ttl': 100}
Transaction sent
Waiting for a few blocks to be mined
Alice's balance is now 0
Coinbase reward is 10000000000000000000, had mined 3 blocks
Bob's balance (with coinbase rewards) is now 70000000000000000000
FAIL

======================================================================
FAIL: integration.test_spend_tx.test_not_enough_tokens
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/builder/epoch/py/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/builder/epoch/py/tests/integration/test_spend_tx.py", line 81, in test_not_enough_tokens
    assert_equals(bob_new_balance.balance, expected_balance)
AssertionError: 70000000000000000000L != 60000000000000000000L
```

apps/aehttp/test/aehttp_integration_SUITE.erl
lucafavatella: I see tracked as https://www.pivotaltracker.com/n/projects/2124891/stories/160445250

PR #1544, opened 2018-09-11
Replace '$' with '_' in addresses

hanssv:
Protocol changes: https://github.com/aeternity/protocol/pull/195
Solves: https://www.pivotaltracker.com/story/show/160254986


hanssv:
Found one more place in system tests...

PR #1455, opened 2018-08-17
[RFC] CUDA PoW docs - recommended hex config parameter

lucafavatella:

PR #1545, opened 2018-09-11
Fix CUDA docs configuration examples

dincho:
https://www.pivotaltracker.com/story/show/159854654


lucafavatella:
@dincho Did you check that the mined block got included in the chain?

(I see that this is the same as #1455 so it is ok for the rest.)

dincho:
> @dincho Did you check that the mined block got included in the chain?

yes, @acidsennui validated it

PR #1547, opened 2018-09-11
Make outbound peer connection initialization more robust

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/160424086

### DIFF
--- a/apps/aecore/src/aec_peer_connection.erl
+++ b/apps/aecore/src/aec_peer_connection.erl
@@ -215,7 +215,9 @@ init([Opts]) ->
                  , kind => permanent
                  , version => Version
                  , genesis => Genesis},
-    {ok, Opts1, 0}.
+    %% NOTE: The init continues at handle_info(init_continue, State).
+    self() ! init_continue,
+    {ok, Opts1}.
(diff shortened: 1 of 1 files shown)


lucafavatella:
> This seems a no-op change, the code is refactored in something completely equivalent. What problem did you try to solve and how did it solve it (the ticket is not very clear about the problem).

@ThomasArts 

I should have detailed - this PR is a by-product of review of `aec_tx_pool_sync` changes in https://github.com/aeternity/epoch/pull/1541/files#diff-0c07dd015037eb446bbd6e7298a61532R223 : while checking the code I realized there was this usage of timeout, that I remember in the past in another place had caused intermittently missed initialization. You are correct - it practically has no impact apart from potential corner corner cases listed in the commit message (e.g. pid reuse).

lucafavatella:
Not necessary for this release anyway.

hanssv:
We had an issue earlier where the conductor was sent a message before the timeout message arrived. This is due to the fact that name registration happens early in the internals of `gen_server:start` - here the process is not named so it should not be a problem...

PR #1550, opened 2018-09-13
PT-160445250 Add positive test for POST /key-blocks endpoint

jur0:


apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -1108,15 +1098,37 @@ post_key_block(Config) ->
...
+    {ok, 200, #{}} = post_key_blocks_sut(PendingKeyBlock#{<<"pow">> => PowEvidence, <<"nonce">> => Nonce1}),
+    ok = aecore_suite_utils:wait_for_height(?config(node, Config), Height),
+    {ok, 200, CurrentBlock} = get_key_blocks_current_sut(),
+    ?assertEqual(Height, maps:get(<<"height">>, CurrentBlock)),
+    ?assertEqual(PowEvidence, maps:get(<<"pow">>, CurrentBlock)),
+    ?assertEqual(Nonce1, maps:get(<<"nonce">>, CurrentBlock)),
     ok.
 
+mine_key_block(HeaderBin, Target, Nonce, Attempts) when Attempts > 0 ->
+    case rpc(aec_mining, mine, [HeaderBin, Target, Nonce]) of
```
lucafavatella: I consider fine to mine the test block from the test on the system under test (I expect just the erlexec process(es) and the RPC worker for this request).

lucafavatella: This uses `rpc:call/5` with timeout 5000 and is not handling the `{badrpc, timeout}`.
* An explicit timeout would help reader.
* A comment mentioning why such timeout is ok (non-production PoW?) would help reader.

lucafavatella:
> Why 5s is assumed to be enough a timeout for RPC call mining.

I see setup of node under test (dev1) relies on `{aec_pow_cuckoo, {"mean16s-generic", "-t 5", 16, false}}` in sys.config. So 5s is reasonable (I would prefer an explicit and larger timeout though debatable). Ok.

lucafavatella:
> Why test assumes that posted block will become new top.

I see test sets user config with test default mining setting `<<"autostart">> => false` so looks reasonable. Though I see that some tests start&stop mining, so reader of tests may not be sure that at the beginning of a certain test the node is not mining. Shall an `aec_conductor:get_mining_state/0` RPC call be present in the `init_per_testcase` of each test assuming this? Though not fully in scope of this PR so ok.

PR #1553, opened 2018-09-13
PT-160237116 SC Chain watching

uwiger:
See [PT #160237116](https://www.pivotaltracker.com/story/show/160237116)

The chain watcher detects if the channel object enters `closing` status, and informs the fsm, which terminates.

Also includes refactoring of the fsm's handling of general events, which was inconsistently implemented.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -693,7 +693,7 @@ accepted(cast, {?FND_CREATED, Msg}, #data{role = responder} = D) ->
         %%     close(Error, D)
     end;
 accepted(Type, Msg, D) ->
-    handle_common_event(Type, Msg, error, D).
+    handle_common_event(Type, Msg, error_all, D).
 
 half_signed(enter, _OldSt, _D) -> keep_state_and_data;
 half_signed(cast, {?FND_SIGNED, Msg}, #data{role = initiator} = D) ->
@@ -710,7 +710,7 @@ half_signed(cast, {?FND_SIGNED, Msg}, #data{role = initiator} = D) ->
         %%     close(Error, D)
     end;
 half_signed(Type, Msg, D) ->
-    handle_common_event(Type, Msg, error, D).
+    handle_common_event(Type, Msg, error_all, D).
 
 dep_half_signed(enter, _OldSt, _D) -> keep_state_and_data;
 dep_half_signed(cast, {Req, _} = Msg, D) when ?UPDATE_CAST(Req) ->
@@ -935,7 +935,7 @@ signed(cast, {?FND_LOCKED, Msg}, D) ->
 signed(cast, {?SHUTDOWN, _Msg}, D) ->
     next_state(closing, D);
 signed(Type, Msg, D) ->
-    handle_common_event(Type, Msg, error, D).
+    handle_common_event(Type, Msg, error_all, D).
 
 funding_locked_complete(D) ->
     D1   = D#data{state = aesc_offchain_state:add_signed_tx(D#data.create_tx,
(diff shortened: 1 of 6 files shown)


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -91,6 +92,30 @@
...
+                         ; T =:= ?UPDATE_ERR
+                         ; T =:= ?DEP_CREATED
+                         ; T =:= ?DEP_SIGNED
+                         ; T =:= ?DEP_LOCKED
+                         ; T =:= ?DEP_ERR
+                         ; T =:= ?WDRAW_CREATED
+                         ; T =:= ?WDRAW_SIGNED
+                         ; T =:= ?WDRAW_LOCKED
+                         ; T =:= ?WDRAW_ERR
+                         ; T =:= ?INBAND_MSG
```
hanssv: Why not ?DISCONNECT

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -714,13 +692,8 @@ accepted(cast, {?FND_CREATED, Msg}, #data{role = responder} = D) ->
...
     end;
-accepted(cast, {?DISCONNECT, _Msg}, D) ->
-    close(disconnect, D);
-accepted(timeout, accepted = T, D) ->
-    close({timeout, T}, D);
-accepted({call, From}, Req, D) ->
-    handle_call(accepted, Req, From, D).
-
+accepted(Type, Msg, D) ->
+    handle_common_event(Type, Msg, error, D).
```
hanssv: What is the difference between `error` and `error_all`?

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1301,7 +1231,46 @@ handle_call_(_St, _Req, From, D) ->
...
+            handle_call(St, Req, From, D)
+    end;
+handle_common_event_(_Type, {_, _} = _Msg, _St, P, D) when P == postpone ->
+    postpone(D);
+handle_common_event_(info, Msg, _St, _P, D) ->
+    handle_info(Msg, D);
+handle_common_event_(Type, Msg, St, discard, D) ->
+    lager:error("Discarding ~p in '~p' state: ~p", [Type, St, Msg]),
+    keep_state(log(drop, msg_type(Msg), Msg, D));
+handle_common_event_(Type, Msg, St, error, D) ->
```
hanssv: What about `error_all`?

apps/aechannel/src/aesc_fsm.erl
velzevur: What I got is `error_all` has some special meaning (still opening)

uwiger: In the `error` mode, calls are still handled (but could potentially be postponed by `handle_call()` code). In `error_all` all calls and casts not explicitly handled lead to protocol error. This is used during the initial start/reestablish handshake.

apps/aechannel/src/aesc_fsm.erl
uwiger: Will fix.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -89,87 +100,179 @@ terminate(_Reason, _St) ->
...
+check_status_(#{mode := tx_hash, tx_hash := TxHash, parent := Parent,
+                type := Type, min_depth := MinDepth} = R,
+              #st{chan_id = ChanId}, C) ->
+    lager:debug("check_status(tx_hash = ~p, parent = ~p)", [TxHash, Parent]),
+    case min_depth_achieved(TxHash, MinDepth, C) of
+        {true, C1} ->
             lager:debug("min_depth achieved", []),
-            case aesc_fsm:minimum_depth_achieved(
+            #{ callback_mod := Mod } = R,
+            case Mod:minimum_depth_achieved(
```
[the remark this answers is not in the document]
uwiger: A remnant from an idea that I think is now obsolete, and the fact that dialyzer is unhappy if a case clause exists but isn't reached, because no callback makes use of it.

PR #1555, opened 2018-09-17
PT-159427123 gas limit per microblock

jur0:
[PT-159427123](https://www.pivotaltracker.com/n/projects/2124891/stories/159427123)

This PR:
* limits the number of txs per micro block by gas
* adds fixed block gas limit into governance - this might be changed in future PRs
* adds fixed gas per tx (which is the same for each tx) - this will be changed in future PRs

TODO:
- [x] add tests

### DIFF
--- a/apps/aecore/src/aec_governance.erl
+++ b/apps/aecore/src/aec_governance.erl
@@ -8,6 +8,7 @@
          expected_block_mine_rate/0,
          block_mine_reward/0,
          block_gas_limit/0,
+         tx_gas/0,
          max_txs_in_block/0,
          beneficiary_reward_delay/0,
          minimum_tx_fee/0,
@@ -37,8 +38,8 @@
 %% Ethereum's gas limit is 8 000 000 and block time ~15s.
 %% For 3s block time it's 1 600 000 (5x less).
 -define(BLOCK_GAS_LIMIT, 1600000).
-
--define(TX_GAS, 21000). %% Taken from Ethereum - a simple tx to send Eth is about 21000 gas.
+%% Taken from Ethereum - a simple tx to send Eth is about 21000 gas.
+-define(TX_GAS, 21000).
 -define(BENEFICIARY_REWARD_DELAY, 180). %% in key blocks / generations
 -define(MICRO_BLOCK_CYCLE, 3000). %% in msecs
 
@@ -80,6 +81,9 @@ block_mine_reward() ->
 
 block_gas_limit() ->
     ?BLOCK_GAS_LIMIT.
+
+tx_gas() ->
+    ?TX_GAS.
 
 max_txs_in_block() ->
     %% TODO: Consider trade-offs sync latency vs pow time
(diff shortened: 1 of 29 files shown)


apps/aecore/src/aec_governance.erl
```diff
@@ -33,6 +34,8 @@
 -define(TIMESTAMP_MEDIAN_BLOCKS, 11).
 -define(EXPECTED_BLOCK_MINE_RATE, 3 * 60 * 1000). %% 60secs * 1000ms * 3 = 180000msecs
 -define(BLOCK_MINE_REWARD, 10000000000000000000).
+-define(BLOCK_GAS_LIMIT, 8000000). %% Taken from Ethereum.
```
hanssv: Since we have 3 seconds between blocks rather than ~15 I guess a smaller value would make sense...

apps/aecore/src/aec_tx_pool.erl
```diff
@@ -236,37 +236,42 @@ int_get_max_nonce(Mempool, Sender) ->
...
+            case Height < TxTTL andalso ok =:= int_check_nonce(Tx, AccountsTree) of
                 true ->
-                    NewAcc = gb_trees:insert({Account, Nonce}, Tx, Acc),
-                    int_get_candidate(N - 1, Dbs, Next, AccountsTree, Height, NewAcc);
+                    case Gas - TxGas of
+                        RemGas when RemGas >= 0 ->
+                            Acc1 = gb_trees:insert({Account, Nonce}, Tx, Acc),
+                            int_get_candidate(RemGas, Dbs, Next, AccountsTree, Height, Acc1);
+                        _ ->
+                            gb_trees:values(Acc)
```
hanssv: Just because this transaction didn't fit does not necessarily mean that no other transaction would?

apps/aecore/src/aec_governance.erl
jur0: Ok, I will use smaller number.

apps/aecore/src/aec_tx_pool.erl
jur0: Yes, others could fit. It'd mean going through the whole pool until the one that fits is found or `$end_of_table` is reached. I'll fix it.

apps/aecore/src/aec_governance.erl
```diff
@@ -33,6 +34,11 @@
 -define(TIMESTAMP_MEDIAN_BLOCKS, 11).
 -define(EXPECTED_BLOCK_MINE_RATE, 3 * 60 * 1000). %% 60secs * 1000ms * 3 = 180000msecs
 -define(BLOCK_MINE_REWARD, 10000000000000000000).
+%% Ethereum's gas limit is 8 000 000 and block time ~15s.
+%% For 3s block time it's 1 600 000 (5x less).
+-define(BLOCK_GAS_LIMIT, 1600000).
+%% Taken from Ethereum - a simple tx to send Eth is about 21000 gas.
+-define(TX_GAS, 21000).
```
ThomasArts: Should we not have a gas price depending on what it is we do? Now every tx seems equally expensive. I would have expected the gas for a contract call to be different from the gas of a name registration? Probably I miss something, but the comment: "Taken from Etherium" does not explain this constant well enough.

jur0: This PR is just to enforce gas limit per micro block. It's a preparation for the following PRs that will implement variable gas for different txs.

PR #1556, opened 2018-09-17
Make difficulty an integer and simplify test_target

hanssv:
In order to avoid floating point computations we change difficulty to be
an integer. To have the same precision as before we introduce a
multiplying factor (1 << 24).

Open questions:
* Do we need a human friendly alternative?

TODO:
* ~Fix the protocol description of the `PING` P2P message.~


hanssv:
https://github.com/aeternity/protocol/pull/200

PR #1562, opened 2018-09-18
Pt 160548220 expired ttl crashes cowboy

ThomasArts:
Sending an expired TTL or any other malformed info in a transaction could cause a 500 status code to be returned due to a crash in aehttp_dispatch_ext.

### DIFF
--- a/apps/aehttp/src/aehttp_dispatch_ext.erl
+++ b/apps/aehttp/src/aehttp_dispatch_ext.erl
@@ -320,9 +320,13 @@ handle_request('PostTransaction', #{'Tx' := Tx}, _Context) ->
         {ok, TxDec} ->
             case deserialize_transaction(TxDec) of
                 {ok, SignedTx} ->
-                    ok = aec_tx_pool:push(SignedTx), %% TODO Add proper error handling
-                    Hash = aetx_sign:hash(SignedTx),
-                    {200, [], #{<<"tx_hash">> => aec_base58c:encode(tx_hash, Hash)}};
+                    case aec_tx_pool:push(SignedTx) of
+                        ok ->
+                            Hash = aetx_sign:hash(SignedTx),
+                            {200, [], #{<<"tx_hash">> => aec_base58c:encode(tx_hash, Hash)}};
+                        {error, _} ->
+                            {400, [], #{reason => <<"Invalid tx">>}}
+                    end; 
                 {error, broken_tx} ->
                     {400, [], #{reason => <<"Invalid tx">>}}
             end;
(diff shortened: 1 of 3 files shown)


apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -320,9 +320,13 @@ handle_request('PostTransaction', #{'Tx' := Tx}, _Context) ->
...
                 {ok, SignedTx} ->
-                    ok = aec_tx_pool:push(SignedTx), %% TODO Add proper error handling
-                    Hash = aetx_sign:hash(SignedTx),
-                    {200, [], #{<<"tx_hash">> => aec_base58c:encode(tx_hash, Hash)}};
+                    case aec_tx_pool:push(SignedTx) of
+                        ok ->
+                            Hash = aetx_sign:hash(SignedTx),
+                            {200, [], #{<<"tx_hash">> => aec_base58c:encode(tx_hash, Hash)}};
+                        {error, _} ->
+                            {400, [], #{reason => <<"Invalid tx">>}}
```
lucafavatella: This may be misleading for the user in the case of the [`{error, already_accepted}`](https://github.com/aeternity/epoch/blob/8d9f5d44e380ad8772ea27cea19404a9813cff13/apps/aecore/src/aec_tx_pool.erl#L454-L457) though it is an improvement. Ok.

ThomasArts: Good point Luca, we don't propagate the reason why things fail.... Bascially because the human is not reading this, a wallet is and I guess that the more error messages we come up with, the harder for the wallet to decide what to do. Invalid means: something went wrong, recompute your transaction.

Whether a status code 500 has changed into 400 is not something I think we should bother the release notes with. Error fixed, that's it.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -2953,13 +2953,28 @@ post_broken_tx(_Config) ->
...
     SignedTxBin = aetx_sign:serialize_to_binary(SignedTx),
+
+    {ok, SpendTTLTx} = 
+        aec_spend_tx:new(
+          #{sender_id => aec_id:create(account, PubKey),
+            recipient_id => aec_id:create(account, random_hash()),
+            amount => Amount,
+            fee => Fee,
+            nonce => Nonce,
+            ttl => 2,
```
lucafavatella: This TTL is absolute hence the test case considering this invalid. Ok.

PR #1563, opened 2018-09-18
Add release note for PT-157317238

hanssv:


lucafavatella:
No way this is going to break CI. merging.

PR #1564, opened 2018-09-18
Add release note for PT-159427123

jur0:


docs/release-notes/next/PT-159427123.md
```diff
@@ -0,0 +1,3 @@
+* Adds fixed gas limit per micro block
+* Adds fixed gas to all transactions (contract create and contract call transactions use the fixed gas plus gas needed for their execution)
```
lucafavatella: > contract create and contract call transactions use the fixed gas plus gas needed for their execution

I understand that the gas charged to the tx sender is only the "gas needed for their execution". This mismatch may need to be reviewed in the following.

lucafavatella:
Thanks.

docs/release-notes/next/PT-159427123.md
jur0: In case of contract txs I believe this is correct: `total_fee = fee + gas * gas_price`. Sender needs to pay `total_fee` to send and execute the tx.

jur0: The fixed gas is not paid by the sender, it's just some hardcoded value to make it possible to use gas as a limit for micro blocks. The sender pays the `fee` and in case of contract txs also `gas` (without the fixed gas).

PR #1565, opened 2018-09-18
Pt 160592568 fix candidate generation

gorillainduction:
https://www.pivotaltracker.com/story/show/160592568


apps/aecore/test/aec_chain_tests.erl
```diff
@@ -835,6 +836,31 @@ fork_on_old_fork_point() ->
                                     dir => backward})),
     ok.
 
+fork_gen_key_candidate() ->
```
lucafavatella: By reverting fix this test fails. That is good.

PR #1566, opened 2018-09-19
Pt 160591333 fix sync with old fork

gorillainduction:
Fixes the second failure of https://www.pivotaltracker.com/story/show/160591333

Also, adapted a unit test to provoke the bad behavior.

### DIFF
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -131,12 +131,12 @@ median_timestamp(Header) ->
 
 -spec insert_block(aec_blocks:block() | map()) -> 'ok' | {'pof', aec_pof:pof()} | {'error', any()}.
 insert_block(#{ key_block := KeyBlock, micro_blocks := MicroBlocks, dir := forward }) ->
-    FoldFun = fun(MB, ok) -> insert_block(MB);
-                 (MB, {pof, _}) -> insert_block(MB);
+    FoldFun = fun(MB, ok) -> do_insert_block(MB, sync);
+                 (MB, {pof, _}) -> do_insert_block(MB, sync);
                  (_MB, Err = {error, _}) -> Err
               end,
     %% First insert key_block
-    case insert_block(KeyBlock, sync) of
+    case do_insert_block(KeyBlock, sync) of
         ok ->
             lists:foldl(FoldFun, ok, MicroBlocks);
         {pof,_Pof} ->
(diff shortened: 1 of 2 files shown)


apps/aecore/test/aec_chain_tests.erl
```diff
@@ -832,8 +845,8 @@ fork_on_old_fork_point() ->
 
     %% But if we add it through sync, it is allowed
     ?assertEqual(ok, insert_block(#{key_block => ForkBlock,
```
lucafavatella: Disabling code fix:
```
  1) aec_chain_tests:forking_test_/0: Fork on old fork point
     Failure/Error: ?assertEqual(ok, insert_block ( # { key_block => ForkBlock , micro_blocks => MicroBlocks , dir => forward } ))
       expected: ok
            got: {error,{too_far_below_top,3,10}}
     %% /Users/luca/dev/aeternity/epoch/_build/test/lib/aecore/test/aec_chain_tests.erl:847:in `aec_chain_tests:fork_on_old_fork_point/0`
```

That is good.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -146,21 +146,21 @@ insert_block(#{ key_block := KeyBlock, micro_blocks := MicroBlocks, dir := forwa
...
         ok ->
-            insert_block(KeyBlock, sync);
+            do_insert_block(KeyBlock, sync);
         {pof,_PoF} ->
-            insert_block(KeyBlock, sync);
+            do_insert_block(KeyBlock, sync);
         Err = {error, _} ->
             Err
     end;
 insert_block(Block) ->
```
lucafavatella: This keeps the check for:
* Key blocks received by gossip
* Micro blocks received by gossip
* Locally mined key block
* Locally signed micro block
* Block posted by user HTTP API (this is debatable though out of scope for this PR)

Ok.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -131,12 +131,12 @@ median_timestamp(Header) ->
 
 -spec insert_block(aec_blocks:block() | map()) -> 'ok' | {'pof', aec_pof:pof()} | {'error', any()}.
 insert_block(#{ key_block := KeyBlock, micro_blocks := MicroBlocks, dir := forward }) ->
```
lucafavatella: This implies that all and only the places where this envelope `#{ key_block := _, micro_blocks := _, dir := forward | backward }` is in sync. I am finding this a little difficult to check (jumps in aec_sync module).

Would it make sense to check this envelope in aec_conductor:add_synced_block (for readability)?

lucafavatella: I quickly checked for `micro_blocks` and I found no points where this envelope is passed that is not sync-related so looks good.

PR #1568, opened 2018-09-19
PT-160619048 remove duplicate redundant properties swagger

jur0:
[PT-160619048](https://www.pivotaltracker.com/n/projects/2124891/stories/160619048)

This PR fixes swagger spec, it:
* removes `txs_hash` from key block
* removes duplicate `poi` from channel force progress tx

### DIFF
--- a/config/swagger.yaml
+++ b/config/swagger.yaml
@@ -2368,8 +2368,6 @@ definitions:
         minItems: 1
         items:
           $ref: '#/definitions/EncodedHash'
-      poi:
-        type: string
       ttl:
         type: integer
         format: int64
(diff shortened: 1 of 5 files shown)


ThomasArts:
I am confused here on why to remove the txs_hash from the KeyBlock definition. With GetTopBlock one gets either a KeyBlock or A MicroBlock in return, at least according to the comment "description: 'Get the top block (either key or micro block)'". 

In fact I just checked and txs_hash is there if there are transactions in the generation.
So, NO, don't remove it!

config/swagger.yaml
```diff
@@ -54,8 +54,9 @@ paths:
       responses:
         '200':
           description: 'Successful operation'
-          schema:
-            $ref: '#/definitions/KeyBlock'
+          oneOf:
```
ThomasArts: ok! Got it.

PR #1569, opened 2018-09-19
PT-160095029 Add ERC-20 token contract and test

rvirding:
The contract had to be modified to handle empty first field in nested
map access.

Implements [PT-160095029](https://www.pivotaltracker.com/story/show/160095029)

### DIFF
--- a/apps/aesophia/test/contracts/erc20_token.aes
+++ b/apps/aesophia/test/contracts/erc20_token.aes
@@ -1,0 +1,97 @@
+contract ERC20Token =
+  record state = {
+    totalSupply : int,
+    decimals : int,
+    name : string,
+    symbol : string,
+    balances : map(address, int),
+    allowed : map(address, map(address,int)),
+    // Logs, remove when native Events are there
+    transfer_log : list((address,address,int)),
+    approval_log : list((address,address,int))}
+
+  // init(100000000, 10, "Token Name", "TKN")
+  public stateful function init(_totalSupply : int, _decimals : int, _name : string, _symbol : string ) = {
+    totalSupply = _totalSupply,
+    decimals = _decimals,
+    name = _name,
+    symbol = _symbol,
+    balances = {[Call.caller] = _totalSupply }, // creator gets all Tokens
+    allowed = {},
+    // Logs, remove when native Events are there
+    transfer_log = [],
+    approval_log = []}
+
+  public stateful function totalSupply() : int = state.totalSupply
+  public stateful function decimals() : int = state.decimals
+  public stateful function name() : string = state.name
+  public stateful function symbol() : string = state.symbol
+
+  public stateful function balanceOf(tokenOwner : address ) : int = lookupDefault(tokenOwner, state.balances, 0)
+
+  public stateful function transfer(to : address, tokens : int) =
+    put( state{balances[Call.caller] = sub(state.balances[Call.caller], tokens) })
+    put( state{balances[to] = add(lookupDefault(to, state.balances, 0), tokens) })
+    transferEvent(Call.caller, to, tokens)
+    true
+
+  public stateful function approve(spender : address, tokens : int) =
+    // allowed[Call.caller] field must have a value!
+    ensure_allowed(Call.caller)
+    put( state{allowed[Call.caller][spender] = tokens} )
+    approvalEvent(Call.caller, spender, tokens)
+    true
+
+  public stateful function transferFrom(from : address, to : address, tokens : int) =
+    put( state{ balances[from] = sub(state.balances[from], tokens) })
+    put( state{ allowed[from][Call.caller] = sub(state.allowed[from][Call.caller], tokens) })
+    put( state{ balances[to] = add(balanceOf(to), tokens) })
+    transferEvent(from, to, tokens)
+    true
+
+  public function allowance(_owner : address, _spender : address) : int =
+    state.allowed[_owner][_spender]
+
+  public stateful function getTransferLog() : list((address,address,int)) =
+    state.transfer_log
+  public stateful function getApprovalLog() : list((address,address,int)) =
+    state.approval_log
+
+  //
+  // Private Functions
+  //
+
+  private function ensure_allowed(key : address) =
+    switch(Map.lookup(key, state.allowed))
+      None => put(state{allowed[key] = {}})
+      Some(_) => ()
+
+  private function transferEvent(from : address, to : address, tokens : int) =
+    let e = (from, to, tokens)
+    put( state{transfer_log = e :: state.transfer_log })
+    e
+
+  private function approvalEvent(from : address, to : address, tokens : int) =
+    let e = (from, to, tokens)
+    put( state{approval_log = e :: state.approval_log })
+    e
+
+  private function abort(err : string) =
+    switch(0) 1 => ()
+
+  private function require(b : bool, err : string) =
+    if(!b) abort(err)
+
+  private function lookupDefault(k : address, m, v) =
+    switch(Map.lookup(k, m))
+      None    => v
+      Some(x) => x
+
+  private function sub(_a : int, _b : int) : int =
+    require(_b =< _a, "Error")
+    _a - _b
+
+  private function add(_a : int, _b : int) : int =
+    let c : int = _a + _b
+    require(c >= _a, "Error")
+    c
--- a/apps/aehttp/test/aehttp_contracts_SUITE.erl
+++ b/apps/aehttp/test/aehttp_contracts_SUITE.erl
@@ -29,6 +29,7 @@
          counter_contract/1,
          dutch_auction_contract/1,
          environment_contract/1,
+         erc20_token_contract/1,
          factorial_contract/1,
          fundme_contract/1,
          identity_contract/1,
@@ -67,6 +68,7 @@ groups() ->
        spend_test_contract,
        dutch_auction_contract,
        fundme_contract,
+       erc20_token_contract,
        null                                     %This allows to end with ,
       ]}
     ].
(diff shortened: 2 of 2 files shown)


velzevur:
May I suggest making this more trustless? Current approach is:
Alice creates contract and Alice has full control of giving tokens to people. If Bob wants to get some sweet ERC20 tokens, he has to ask Alice for it. Also there is no way of getting AE coins out of the contract.

My suggestions are:
1) Selling tokens (instead of giving them away)
Upon creation contract has `state.not_sold_tokens` or something like that - representing the amount of tokens left to sell. It also has a `rate` member of the state, representing the exchange rate AE coins -> ERC tokens. If `rate=2` and the caller spends 100 AE coins (as [`amount`](https://github.com/aeternity/epoch/blob/master/apps/aecontract/src/aect_call_tx.erl#L55) of the contract call) - then one gets 200 ERC tokens in the channel. This amount is subtracted from `state.not_sold_tokens`. You can get the `amount` using [`Call.value`](https://github.com/aeternity/epoch/blob/master/apps/aecontract/src/aect_call_tx.erl#L55). Then only thing we need is a `public function buy()`. Since you get the caller's address from `Call.caller` and the amount being spent with `Call.value` - we might even not need params there at all.
The good thing about this would be that it is trustless (no one has to ask Alice for anything).
2) Owner and a withdraw function
Since people are buying ERC20 tokens for AE coins, the contract aggregates some AE coins. After the sale is over, the creator of the contract shall be able to withdraw those AE coins to an account of one's own choosing. For this we shall keep a contract's `owner` and expose a public function `withdraw(to: address)` (that has a check that the `owner` is calling it)
3) _this one is just nice to have_ Sale open and close height
Contract sales tokens only in a certain amount of time, defined by block's height. So when Alice posts the contract states that the sale will begin at block height=200 and will end at block height=350. You can get the current height using `Chain.block_height` so all has to be done is actually a small check that this `Chain.block_height` is between two variables from contract's state. After the sale is over, people can still transfer tokens between themselves, it is just that the contract no longer sells tokens.

dadaphl:
I think we should add a disclaimer, that this is not a production ready contract. This contract should use `unit` instead of `int` for example.

rvirding:
This code is based on the solidity code in the Ethereum [ERC20 Token Standard](https://theethereum.wiki/w/index.php/ERC20_Token_Standard) with a few small changes, mainly we don't have events yet sot we just log them and allow access tot he logs.

apps/aesophia/test/contracts/erc20_token.aes
```diff
@@ -0,0 +1,97 @@
...
+
+  public stateful function balanceOf(tokenOwner : address ) : int = lookupDefault(tokenOwner, state.balances, 0)
+
+  public stateful function transfer(to : address, tokens : int) =
+    put( state{balances[Call.caller] = sub(state.balances[Call.caller], tokens) })
+    put( state{balances[to] = add(lookupDefault(to, state.balances, 0), tokens) })
+    transferEvent(Call.caller, to, tokens)
+    true
+
+  public stateful function approve(spender : address, tokens : int) =
```
velzevur: I see one can allow another to use more tokens than one have. The check for spending is [done in `transferFrom`](https://github.com/aeternity/epoch/pull/1569/files#diff-5030b9c4077e8e4e533d582c9f5e2595R46)
Ok.

apps/aesophia/test/contracts/erc20_token.aes
```diff
@@ -0,0 +1,97 @@
...
+    name : string,
+    symbol : string,
+    balances : map(address, int),
+    allowed : map(address, map(address,int)),
+    // Logs, remove when native Events are there
+    transfer_log : list((address,address,int)),
+    approval_log : list((address,address,int))}
+
+  // init(100000000, 10, "Token Name", "TKN")
+  public stateful function init(_totalSupply : int, _decimals : int, _name : string, _symbol : string ) = {
```
lucafavatella: I would expect [no stateful keyword for init](https://github.com/aeternity/protocol/blame/99e569126aa9a2b2be9953e6a0461c6b0258cdbd/contracts/sophia.md#L102-L104).

apps/aesophia/test/contracts/erc20_token.aes
```diff
@@ -0,0 +1,97 @@
...
+    totalSupply = _totalSupply,
+    decimals = _decimals,
+    name = _name,
+    symbol = _symbol,
+    balances = {[Call.caller] = _totalSupply }, // creator gets all Tokens
+    allowed = {},
+    // Logs, remove when native Events are there
```
lucafavatella: Could you please confirm which Pivotal Tracker ticket tracks this? If there is none, could you please add one and post the link here?

apps/aesophia/test/contracts/erc20_token.aes
lucafavatella: And I expect the state record definition can be used as a type. So I understand the most explicit definition would be:
```
public function init(...) : state = {
```

lucafavatella: And it looks like [`: state` may be implied for init](https://github.com/aeternity/protocol/blame/99e569126aa9a2b2be9953e6a0461c6b0258cdbd/contracts/sophia.md#L101).

apps/aesophia/test/contracts/erc20_token.aes
```diff
@@ -0,0 +1,97 @@
...
+
+  public stateful function transfer(to : address, tokens : int) =
+    put( state{balances[Call.caller] = sub(state.balances[Call.caller], tokens) })
+    put( state{balances[to] = add(lookupDefault(to, state.balances, 0), tokens) })
+    transferEvent(Call.caller, to, tokens)
+    true
+
+  public stateful function approve(spender : address, tokens : int) =
+    // allowed[Call.caller] field must have a value!
+    ensure_allowed(Call.caller)
```
lucafavatella: I see this contains a put state, then next line is a put state too. I see other functions contain multiple puts. Any reasons for not having only one put, preparing state as a variable before that?

* Having only one put state at the end would seem more readable. Though debatable.
* The intended comparative gas cost between multiple [put](https://github.com/aeternity/epoch/blob/85f5050fbdf4369385ba33a8b51ff8dfed044017/apps/aesophia/src/aeso_icode_to_asm.erl#L367-L370)s and using variable may need to be considered.

apps/aesophia/test/contracts/erc20_token.aes
```diff
@@ -0,0 +1,97 @@
...
+
+  public stateful function totalSupply() : int = state.totalSupply
+  public stateful function decimals() : int = state.decimals
+  public stateful function name() : string = state.name
+  public stateful function symbol() : string = state.symbol
+
+  public stateful function balanceOf(tokenOwner : address ) : int = lookupDefault(tokenOwner, state.balances, 0)
+
+  public stateful function transfer(to : address, tokens : int) =
+    put( state{balances[Call.caller] = sub(state.balances[Call.caller], tokens) })
```
lucafavatella: It seems that the checks here are not trivial, and I see there are some checks in `sub` and `add`. Could you please ensure there is ticket in Pivotal Tracker re updating this contract once `uint` is available in Sophia?

apps/aesophia/test/contracts/erc20_token.aes
UlfNorell: `put` is essentially free. The only thing it does is write a single word to address 0.

apps/aesophia/test/contracts/erc20_token.aes
```diff
@@ -0,0 +1,97 @@
...
+  public stateful function approve(spender : address, tokens : int) =
+    // allowed[Call.caller] field must have a value!
+    ensure_allowed(Call.caller)
+    put( state{allowed[Call.caller][spender] = tokens} )
+    approvalEvent(Call.caller, spender, tokens)
+    true
+
+  public stateful function transferFrom(from : address, to : address, tokens : int) =
+    put( state{ balances[from] = sub(state.balances[from], tokens) })
+    put( state{ allowed[from][Call.caller] = sub(state.allowed[from][Call.caller], tokens) })
```
lucafavatella: I see this function mixes checks and effects. I.e. balance subtracted already though allowance checked now.

If execution stops in between because out of gas, I see that [state trees are kept to the initial version](https://github.com/aeternity/epoch/blob/85f5050fbdf4369385ba33a8b51ff8dfed044017/apps/aecontract/src/aect_dispatch.erl#L163) so no impact of partial effect. Ok. This contract in `sub` definition exhausts gas (`require/abort` compiled to `switch(0) 1 => ()`) if check fails. Ok.

Though if execution stopped in between because of REVERT then [the state trees would be not kept to the initial version](https://github.com/aeternity/epoch/blob/85f5050fbdf4369385ba33a8b51ff8dfed044017/apps/aecontract/src/aect_dispatch.erl#L155) (I [tracked this REVERT behaviour](https://www.pivotaltracker.com/story/show/160721302)). When [`abort` is implemented in Sophia as REVERT](https://www.pivotaltracker.com/story/show/159522821) [probably this contract will be updated](https://www.pivotaltracker.com/story/show/159522821/comments/194537013) potentially exposing issue that from gets tokens removed (destroyed?) by user not allowed.

apps/aesophia/test/contracts/erc20_token.aes
rvirding: The reason for the ensure_allowed is that the next put accesses maps at two levels `allowed[from][Call.caller] = ...` and sophia requires the outermost map to exist. This is what ensured_allowed does.

lucafavatella: Ok.

apps/aesophia/test/contracts/erc20_token.aes
rvirding: I have added a link the ticket in the first comment.

apps/aesophia/test/contracts/erc20_token.aes
rvirding: Done, PT-160728635

apps/aesophia/test/contracts/erc20_token.aes
lucafavatella: Nono - I meant which ticket tracks this (implicitly) TODO comment "Logs, remove when native Events are there". They are currently kept in memory. Once we have events in Sophia we need to remember to change this contract. A ticket in Pivotal specifically for this would enable us to remember.

lucafavatella: Tracked https://www.pivotaltracker.com/story/show/160750737

apps/aesophia/test/contracts/erc20_token.aes
lucafavatella: Tracked https://www.pivotaltracker.com/story/show/160750737

apps/aesophia/test/contracts/erc20_token.aes
lucafavatella: Tracked with https://www.pivotaltracker.com/story/show/160751003 - for when there is Sophia abort()

PR #1570, opened 2018-09-20
Option pattern compiler bug

UlfNorell:
Compilation of option type patterns were buggy (h/t @velzevur). Fixed by removing special handling of option types in the compiler. They are now defined as (a builtin) `datatype option('a) = None | Some('a)`.

### DIFF
--- a/apps/aesophia/test/contracts/complex_types.aes
+++ b/apps/aesophia/test/contracts/complex_types.aes
@@ -78,8 +78,8 @@ contract ComplexTypes =
       None :: ys => None
       Some(x) :: ys =>
         switch(all_some(ys))
-          None     => None
           Some(xs) => Some(x :: xs)
+          None     => None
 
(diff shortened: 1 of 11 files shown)


apps/aesophia/test/contracts/complex_types.aes
```diff
@@ -78,8 +78,8 @@ contract ComplexTypes =
       None :: ys => None
       Some(x) :: ys =>
         switch(all_some(ys))
-          None     => None
           Some(xs) => Some(x :: xs)
+          None     => None
```
lucafavatella: I understand this was the symptom.

It may be worth a release notes line so that users experiencing same symptom realize that they shall retry on latest release (once this fix is included in a release).

UlfNorell: Did we change the way we handle release notes again? Should I add a file to `docs/release-notes/next/`?

PR #1573, opened 2018-09-21
Refactor aetx behaviour

gorillainduction:
In context of https://www.pivotaltracker.com/story/show/160281213

Make an process/check environment for the transaction behavior to facilitate further adds of necessary environment information (e.g., for contracts: Beneficiary, timestamp).

As a side effect, avoid calculating the tx hash for all transactions, since it is only needed for forcing progress.


velzevur:
So much cleaner!

PR #1574, opened 2018-09-21
Use python version in virtualenv command

jur0:


py/Makefile
```diff
@@ -7,13 +7,13 @@ PIP = $(BIN)/pip
 
 .PHONY: venv-present
 venv-present:
-	@virtualenv -q .
+	@virtualenv --python=python2.7 -q .
```
lucafavatella: `make python-single-uat TEST_NAME=test_spend_tx.py` passed for me locally (I used to locally apply `-p python2` - I see that CI passed in this PR).

PR #1576, opened 2018-09-21
Check value passed to VM primops

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/160684022 - please refer to ticket for context and to commit messages for details.

TODO:
* [x] ~Confirm with reviewer that considering `?CALL` successful if it returns non-zero is sub-optimal hence making it more strict as `== 1` shall be tracked as separate ticket.~
  ```
    , i(?CALL)                                %%  Result OOffset
    , jump_if(CallOk) %% Jump if non-zero. CALL places on stack 1 if ok and 0 if failure. I.e. jump if non-failure. TODO: stricter check on ok (1) rather than non-failure.
    , exhaust_gas()
    , jumpdest(CallOk) %%  OOffset
  ```
  * Dropped.
* [ ] ~Squash and merge.~ Merge (preferably without squashing or rebasing - so not to require rebase of #1580 ).


apps/aesophia/src/aeso_icode_to_asm.erl
```diff
@@ -426,7 +427,9 @@ assemble_expr(Funs, Stack, Tail,
     , dup(3)                                  %%  OOffset OSize ISize OOffset IOffset To Gas Value
     , shuffle_stack([1, 2, 4, 3, 5, 7, 8, 6]) %%  Gas To Value IOffset ISize OOffset OSize OOffset
     , i(?CALL)                                %%  Result OOffset
-    , pop(1)                                  %%  OOffset         -- ignoring result for now
+    , jump_if(CallOk) %% Jump if non-zero. CALL places on stack 1 if ok and 0 if failure. I.e. jump if non-failure. TODO: stricter check on ok (1) rather than non-failure.
+    , exhaust_gas()
```
UlfNorell: It seems to me that a better solution would be to have `CALL` exhaust the gas if not successful. Is there a use case where you'd want to recover from a non-successful call (what would the return data be?)?

Also note that I'm changing the semantics of `CALL` in [PT-160377912-cheap-encode-decode](https://github.com/aeternity/epoch/tree/PT-160377912-cheap-encode-decode) for Sophia contracts to return the data on the stack instead of a simple 0/1.

apps/aesophia/src/aeso_icode_to_asm.erl
```diff
@@ -945,6 +943,14 @@ push_label(Label) -> {push_label, Label}.
 jump(Label)    -> [push_label(Label), i(?JUMP)].
 jump_if(Label) -> [push_label(Label), i(?JUMPI)].
 
+exhaust_gas() ->
```
happi: There is actually one instruction that is dedicated as "the" invalid instruction 0xfe INVALID.
Perhaps not in scope of this ticket but perhaps we should use that here.

lucafavatella: > There is actually one instruction that is dedicated as "the" invalid instruction 0xfe INVALID.

I had started coding that (i.e. using "the" invalid instruction) though it was going to impact more the semantics of the Sophia compiled bytecode and - consequently - the tests (match on `out_of_gas`). Hence I settled on exhausting gas at the Sophia bytecode level - as I saw it was used in Sophia compiled bytecode already for switch-case not matching any case.

> Perhaps not in scope of this ticket but perhaps we should use that here.

If we decide to go that way, I am happy to take that forward though I would prefer to keep it as distinct ticket/task/PR.

apps/aesophia/src/aeso_icode_to_asm.erl
lucafavatella: > It seems to me that a better solution would be to have `CALL` exhaust the gas if not successful.

I agree, it should keep these changes re check on value smaller. I will revert&amend.

(I see in https://github.com/aeternity/epoch/tree/PT-160377912-cheap-encode-decode returning 0 **and** exhausting gas is done (`{0, aevm_eeevm_state:set_gas(0, ReturnState2)}`).)

> Is there a use case where you'd want to recover from a non-successful call (what would the return data be?)?

I would expect the return data to be dummy hence not meant to be used. I expect this to be more a question of level of compatibility among different VM ABIs. I consider these cases tracked as https://www.pivotaltracker.com/story/show/159686915

(I agree to your proposal above, will amend.)

> Also note that I'm changing the semantics of CALL ...

I see some changes in CALL semantics e.g. "In Sophia the ISize is the type of the arguments and IOffset is a pointer into the heap." . Thanks for the heads up.

lucafavatella: Addressed. CI is running.

PR #1579, opened 2018-09-21
Add system tests for oracles

sylane:
It is work-in-progress because of some intermittent error.
Sometimes, after the oracle responded and the response transaction got included in the chain, requesting the open transaction still return the query.

### DIFF
--- a/system_test/aest_oracles_SUITE.erl
+++ b/system_test/aest_oracles_SUITE.erl
@@ -13,7 +13,9 @@
 -export([
     test_simple_same_node_query/1,
     test_simple_two_nodes_query/1,
-    test_oracle_ttl_extension/1
+    test_oracle_ttl_extension/1,
+    test_pipelined_same_node_query/1,
+    test_pipelined_two_nodes_query/1
 ]).
 
 -import(aest_nodes, [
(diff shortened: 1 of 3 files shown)


hanssv:
The intermittent failures are expected. The test only uses `GetTransactionByHash` and does not check that it isn't a micro fork. (This is the same problem we had many times in the normal test suites - we solve the Cuckoo puzzle right after signing a micro block which leads to the Tx being on chain, then off chain again, before being finally included on the chain.) The slower (but deterministic) solution is to wait until the Tx is included in a _closed_ generation.

system_test/aest_oracles_SUITE.erl
```diff
@@ -0,0 +1,249 @@
...
+    EncOPubKey = aec_base58c:encode(oracle_pubkey, OPubKey),
+    EncQPubKey = aec_base58c:encode(account_pubkey, QPubKey),
+
+    %% Setup nodes
+    NodeConfig = #{ beneficiary => EncMPubKey },
+    setup([?NODE1, ?NODE2], NodeConfig, Cfg),
+    NodeNames = [node1, node2],
+    start_node(node1, Cfg),
+    start_node(node2, Cfg),
+    wait_for_startup([node1, node2], 4, Cfg),
```
ThomasArts: Note that this requires: 
system_test/aest_channels_SUITE_data/epoch.yaml.mustache:    beneficiary_reward_delay: 2

ThomasArts: which is indeed set in the file!

PR #1582, opened 2018-09-24
PT-160721422 Fragmentation issue again

hanssv:
You need brackets in some kinds of arithmetic expressions 🤦‍♂️ 

Now, there is a proper test at least.


hanssv:
https://www.pivotaltracker.com/story/show/160721422

PR #1585, opened 2018-09-25
PT-160090121 Use integers to compute beneficiary reward

jur0:
[PT-160090121](https://www.pivotaltracker.com/n/projects/2124891/stories/160090121)

### DIFF
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -875,7 +875,7 @@ grant_fees(Node, Trees, Delay, FraudStatus, State) ->
     Beneficiary2 = node_beneficiary(KeyNode2),
     Reward = aec_governance:block_mine_reward(),
     FraudReward = aec_governance:fraud_report_reward(),
-    BeneficiaryReward1 = round(KeyFees * 0.4),
+    BeneficiaryReward1 = KeyFees * 4 div 10,
     BeneficiaryReward2 = KeyFees - BeneficiaryReward1 + Reward,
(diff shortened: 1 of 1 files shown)


apps/aecore/src/aec_chain_state.erl
```diff
@@ -875,7 +875,7 @@ grant_fees(Node, Trees, Delay, FraudStatus, State) ->
     Beneficiary2 = node_beneficiary(KeyNode2),
     Reward = aec_governance:block_mine_reward(),
     FraudReward = aec_governance:fraud_report_reward(),
-    BeneficiaryReward1 = round(KeyFees * 0.4),
+    BeneficiaryReward1 = KeyFees * 4 div 10,
```
ThomasArts: round(22 *0,4) = 9  whereas 22 * 4 div 10 = 8
Is that intentional?

jur0: The intention is to get rid of the computations with floating point numbers as it can be hardware dependent and break the consensus. Now we can get slightly different values than before, but they should be the same on any hardware.

fabiankrol: I'm fine with that as long as long as we're not loosing anything, i.e., we're not using  `KeyFees * 6 div 10` for `BeneficiaryReward2` 🤖

lucafavatella: I see:
```
1> (22 * 4) div 10.
8
2> 22 * (4 div 10).
0
3> 22 * 4 div 10.
8
```
so no parentheses needed. Ok.

PR #1587, opened 2018-09-25
Add backend for external nodes to system test

eproxus:
This is a new backend used to add externally running nodes to the system
test node manager. Some actions are ignored by default (like start and
stop for example) and some actions are not possible (e.g. getting logs
and extracting archives). Those will throw errors.


lucafavatella:
Labelling as WIP as not planned to merge in this release.

PR #1591, opened 2018-09-26
Pt 160732539 implement blockhash

gorillainduction:
https://www.pivotaltracker.com/story/show/160732539


gorillainduction:
Protocol PR https://github.com/aeternity/protocol/pull/211

gorillainduction:
> Does that not test the right thing?

Actually, it is correct, since BLOCKHEIGHT should return 0 for the current height. I will look at adapting the test to test some more.

gorillainduction:
Rebased to resolve conflicts, and modified test that was pointed out by @UlfNorell

PR #1592, opened 2018-09-26
State channels' contracts refer on-chain objects

velzevur:
[SC addressing on-chain objects](https://www.pivotaltracker.com/story/show/160280811)


velzevur:
This will have to be based on top of #1591

velzevur:
Freshly based on top of latest master. Protocol changes are coming

apps/aecore/src/aec_vm_chain.erl
```diff
@@ -37,15 +42,20 @@
 
 -include_lib("apps/aecore/include/blocks.hrl").
 
--record(state, { trees   :: aec_trees:trees()
-               , tx_env  :: aetx_env:env()
-               , account :: aec_keys:pubkey()            %% the contract account
+-define(NO_INNER_STATE, nej).
```
hanssv: Maybe there are better named atoms?

velzevur:
A little bit of refactoring: removed the off-chain transaction enviroment. It is replaced with the on-chain one.
TODO: pin the call enviroment (both `tx_env:env()` and `trees`) to a certain block hash that both participants will share. WRT to force progress - the miner's top will be used. This is not required for main net and will be tracked in a separate PT.

velzevur:
In order to have the call deterministic, the `beneficiary` must be set to `0` in all channel calls. Adding a `WIP` label again

apps/aecontract/src/aect_dispatch.erl
```diff
@@ -160,3 +161,20 @@ error_to_binary(out_of_stack) -> <<"out_of_stack">>;
...
+            % the on-chain state uses the off-chain aetx_env:env()
+            OnChainState = aec_vm_chain:new_state(OnChainTrees, TxEnv, ContractPubKey),
+            OffChainEnv = off_chain_env(TxEnv),
+            OffChainState = aec_vm_chain:new_offchain_state(Trees, OffChainEnv, ContractPubKey),
+            {OffChainEnv, aec_vm_chain:push_state_context(OffChainState, OnChainState)};
+        false ->
+            {TxEnv, aec_vm_chain:new_state(Trees, TxEnv, ContractPubKey)}
+    end.
+
+off_chain_env(OnChainEnv) ->
```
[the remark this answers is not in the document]
velzevur: Isn't the idea behind off-chain calls to be pinned to a block? So different on-chain objects can be resoned about? Then why not use the `Chain` API as well there...

velzevur:
According to the latest discussions: the beneficiary (`coinbase` op) is the one of the last block for off-chain transactions and the current leader when forcing of progress.

velzevur:
Adding remote contract calls exposed a flaw in the `aesc_txs_SUITE`. This is fixed in the last commit

apps/aecontract/src/aect_dispatch.erl
```diff
@@ -160,3 +161,17 @@ error_to_binary(out_of_stack) -> <<"out_of_stack">>;
 error_to_binary(E) ->
     ?DEBUG_LOG("Unknown error: ~p\n", [E]),
     <<"unknown_error">>.
+
+chain_state(#{ contract    := ContractPubKey
+             , trees       := Trees} = CallDef, TxEnv) ->
+    case maps:get(off_chain, CallDef, false) of
```
lucafavatella: Avoiding default would make the code more readable, by making explicit the point of the decision about off/on chain.

apps/aevm/src/aevm_eeevm_state.erl
```diff
@@ -84,6 +84,11 @@ init(#{ env  := Env
 
     ChainState = maps:get(chainState, Env),
     ChainAPI =  maps:get(chainAPI, Env),
+    {IsOffChain, OnChainTrees} =
+        case maps:get(off_chain, Env, false) of
```
lucafavatella: Avoiding default would make the code more readable, by making explicit the point of the decision about off/on chain.

apps/aetx/src/aetx_env.erl
```diff
@@ -71,22 +72,6 @@
...
-        , difficulty = Difficulty
-        , height  = Height
-        , key_hash = KeyHash
-        , signed_tx = none
-        , time = Time
-     }.
-
 -spec tx_env_from_key_header(aec_headers:key_header(),
                              aec_blocks:block_header_hash(),
                              non_neg_integer(),
```
lucafavatella: Not sure why `_PrevHash` passed as argument at all, as unused. Though out of scope for this PR.

gorillainduction: I'm guilty of this. I had the wrong idea of the semantics of the BLOCKHASH instruction was, and it is a leftover from that implementation. No use removing it in this PR as noted.

apps/aetx/src/aetx_env.erl
```diff
@@ -103,13 +88,39 @@ tx_env_from_key_header(KeyHeader, KeyHash, Time,_PrevHash) ->
...
+                                     Type == aetx_transaction ->
+    {TopHeader, TopHash, Trees} = aec_chain:top_header_hash_and_state(),
+    {KeyHeader, KeyHash, Time} =
+        case aec_headers:type(TopHeader) of
+            micro ->
+                KHash = aec_headers:prev_key_hash(TopHeader),
+                {ok, KH} = aec_chain:get_header(KHash),
+                {KH, KHash, aec_headers:time_in_msecs(TopHeader)};
+            key ->
+                {TopHeader, TopHash, aec_headers:time_in_msecs(TopHeader) + 1}
```
lucafavatella: I see the intention on time here is to simulate the first legal timestamp for the first microblock. I see same snippet `aec_blocks:time_in_msecs(LastKeyBlock) + 1` for [micro block candidate](https://github.com/aeternity/epoch/blob/fce48d3542f7cace2d0ccba8c9f84f2f303dbe97/apps/aecore/src/aec_block_micro_candidate.erl#L144). Factoring this in common would help readability. Though totally optional of course.

apps/aechannel/src/aesc_utils.erl
```diff
@@ -479,8 +479,8 @@ process_force_progress(Tx, Addresses,
     PoITrees0 = trees_from_poi(Addresses, PoI),
     Reserve = aesc_channels:channel_reserve(Channel),
     PoITrees =
-        try aesc_offchain_update:apply_on_trees(Update, PoITrees0, NextRound,
-                                                      Reserve)
+        try aesc_offchain_update:apply_on_trees(Update, PoITrees0, Trees, Env,
```
lucafavatella: `Fee` must be charged upfront. Please amend, or track ticket analogous to https://www.pivotaltracker.com/story/show/160669401 , or at least add comment in same mentioned ticket.

lucafavatella: Gas limit must be charged upfront, and the unused gas reimboursed. Please amend, or track ticket analogous to https://www.pivotaltracker.com/story/show/160671577 , or at least add comment in same mentioned ticket.

lucafavatella: Is `Amount` (as in `{?OP_CALL_CONTRACT, _, _, _, Amount, _, _, GasPrice, Gas}`) used at all? When is it charged?

apps/aecontract/src/aect_channel_contract.erl
```diff
@@ -21,28 +21,21 @@ new(Owner, Round, VmVersion, Code, Deposit, Trees0) ->
...
-               , amount      => 0
-               , call_stack  => CallStack
-               , code        => Code
-               , call        => Call
-               , trees       => Trees0
-               , tx_env     => tx_env(Round)
-               },
+    CallDef = make_call_def(OwnerPubKey, ContractPubKey,
+                            _Gas = 1000000, _GasPrice = 1,
+                            _Amount = 0, %TODO: make this configurable
```
lucafavatella: I guess the TODO includes Gas and GasPrice.

lucafavatella: I understand this is never used on-chain but only offchain, right? So gas price may even be 0?

lucafavatella: (While `run` is used on-chain for force progress tx.)

apps/aecore/src/aec_vm_chain.erl
```diff
@@ -468,20 +507,72 @@ do_set_store(Store, PubKey, Trees) ->
         end,
     aect_state_tree:enter_contract(NewContract, ContractsTree).
 
-apply_transaction(Tx, #state{ trees = Trees, tx_env = Env } = State) ->
+apply_transaction(Tx, #state{tx_env = Env } = State) ->
+    Trees = get_top_trees(State),
```
lucafavatella: So execution of state-modifying callbacks (spend, oracle register/query/respond/extend, aens preclaim/claim/transfer/revoke) in force progress tx (contract call update) read&write only off chain trees, right? It means it will not [throw exception](https://github.com/aeternity/protocol/pull/215/files#r221468884)?

lucafavatella: It is to be checked if those can be used for removing tokens from other account - that then get in channel state on chain - though that applies to all contract calls anyway.

apps/aecore/src/aec_vm_chain.erl
```diff
@@ -423,7 +459,8 @@ call_contract(Target, Gas, Value, CallData, CallStack,
                                     call_data   => CallData,
                                     call_stack  => CallStack }),
             case apply_transaction(CallTx, State) of
```
lucafavatella: Does this lose the `off_chain` from the environment? (See also https://github.com/aeternity/epoch/pull/1592/files?utf8=%E2%9C%93&diff=split&w=1#r222327210 ) What is the impact?

apps/aecontract/src/aect_channel_contract.erl
velzevur: Yes, TODO includes amount, gas limit and gas price althogether. All three are part of the in-chain protocol (`aesc_offchain_udpate`'s call `udpate`) but they are not exposed to WebSocket API. Adding the API after mainnet launch will not impact protocol.
Current values are dummy ones and yes, since gas is not consumed off-chain, it could as well be 0. Those values are important for on-chain execution (where gas is consumed and `gas_price` is important)

apps/aechannel/src/aesc_offchain_update.erl
```diff
@@ -102,8 +103,8 @@ apply_on_trees(Update, Trees0, Round, Reserve) ->
             Trees3 = create_account(ContractPubKey, Trees2),
             Trees4 = add_tokens(ContractPubKey, Deposit, Trees3),
             Call = aect_call:new(OwnerId, Round, ContractId, Round, 0),
-            _Trees = aect_channel_contract:run_new(ContractPubKey, Call, CallData,
-                                                  Round, Trees4);
+            _Trees = aect_channel_contract:run_new(ContractPubKey, Call, CallData, Trees4,
+                                                  OnChainTrees, OnChainEnv);
         {?OP_CALL_CONTRACT, CallerId, ContractId, VmVersion, Amount, CallData,
```
lucafavatella: As the semantics of this (call contract) off-chain update is part of consensus, shall this not call back into `aect_call_tx` rather than re-implementing most of the steps? See [dual doc comment](https://github.com/aeternity/protocol/pull/215#discussion_r222593537).

(I am ok with any answer, e.g. "no"  or "yes but later", though I am wondering how maintenance of two distinct code paths meant to have same semantics is going to be.)

Strictly speaking this is not part of this ticket/PR so no changes needed for this PR.

apps/aechannel/src/aesc_utils.erl
lucafavatella: Checking this better, this is two questions:
* (A) Is state hash generated by off-chain update call contract, and meant to be compared with specified state hash, meant to include charging of fee?
* (B) Is fee in force progress transaction meant to be checked as available in the balance (alongside gaslimit * gasprice) before processing force progress transactions?

B seems done correctly - in the `check` callback.

So the question is A.

lucafavatella: Same as https://github.com/aeternity/epoch/pull/1592/files#r222612328 and https://github.com/aeternity/epoch/pull/1592/files#r222614101

So if confirmed that that "the off-chain update can use only the balance of the initiator and responder that is the off-chain trees i.e. balance that has already been taken out of the on-chain trees" then no changes needed.

lucafavatella: I see the amount that is in the off-chain update:
* In happy path, affects the state hash meant to be compared. It makes sense.
* In happy path, affects the balance of initiator / responder that ends up in updated state channel. It makes sense.
* In case of exceptional handling / revert in VM call (not in the `catch _:_` in this file), it keeps affecting the balance of initiator / responder that ends up in updated state channel. This is in line with [contract call transaction](https://github.com/aeternity/protocol/blame/1c5b17b44da6ecd661bac0fe86c2b1c935f99e8d/contracts/contract_transactions.md#L111). (I added [comment in related ticket](https://www.pivotaltracker.com/story/show/160980601/comments/194950662).)

When is the `try ... catch _:_ ...` - with the cath-all - meant to kick in? In case of code implementation bug i.e. unexpected situation? I would expect in that case this function to terminate abnormally so to prevent bug to silently slip under consensus. Comments?

velzevur: `Fee` is the amount being charged from the `from` (the one that posts the force progress on-chain) and awared to the miner for accepting the transaction. Amount is being charged from the `from`'s on-chain account. This does not impact off-chain trees in any way and thus - has no impact on their root hash.

velzevur: > Do I understand correctly that the off-chain update can use only the balance of the initiator and responder that is the off-chain trees i.e. balance that has already been taken out of the on-chain trees? If so, it seems correct that the fee does not impact the state hash of the off-chain update. I.e. correct as implemented already so no changes needed.

So this is a little bit tricky. At channel creation both participants can dedicate some tokens. From then on - those are locked inside the channel. This is the `channel_amount` and can be only modified by deposits and withdrawals.

Inside the channel this `channel_amount` can be devided between participants or accounts of their desire - those could be some further accounts that only live off-chain, for example off-chain contracts.

Channel off-chain contract calls can modify all off-chain balances. They can NOT interact with on-chain balances neither in off-chain calls, nor via a forced progress call.
So yes, off-chain calls can only update off-chain balances but those could be `responder`, `initator` or some other off-chain accounts as well.

apps/aevm/test/aevm_test_utils.erl
```diff
@@ -300,6 +300,7 @@ build_data_array([], Acc) ->
                                 , {post, leave_out}
                                 , {sourceHash, default, <<>>}
                                 , {lllcversion, default, <<>>}
+                                , {off_chain, default, false}
```
lucafavatella: I think you rather wanted to amend DefaultEnv in get_config/1 ? This seems the format of the external test JSON data.

apps/aecore/src/aec_vm_chain.erl
velzevur: The `off_chain` is kept in the `State`

apps/aechannel/src/aesc_utils.erl
velzevur: The try-catch is for force-specific crashes. For example [when a participant is trying to spend more than one has](https://github.com/aeternity/epoch/blob/master/apps/aechannel/src/aesc_offchain_update.erl#L26). Currently - the only example, so probably a specific catch is better

apps/aevm/src/aevm_eeevm_state.erl
```diff
@@ -84,6 +84,11 @@ init(#{ env  := Env
 
     ChainState = maps:get(chainState, Env),
     ChainAPI =  maps:get(chainAPI, Env),
+    {IsOffChain, OnChainTrees} =
+        case maps:get(off_chain, Env) of
```
lucafavatella: Based on code inspection, this function is always called with this `false`. Can this field be removed altogether from this representation? (In the code, the this env that is passed to this function is always created from scratch - not passed over from arguments - so it shall be same to remove this field altogether.) The code is a bit complex to follow, and removing this field from this point would ease that.

apps/aevm/src/aevm_eeevm_state.erl
```diff
@@ -119,6 +124,8 @@ init(#{ env  := Env
          , environment =>
                #{ spec => Spec
                 , options => Opts }
+         , off_chain  => IsOffChain
```
lucafavatella: Based on code inspection, this is not used. Would it be possible to remove it? The code is a bit complex, this would ease following it.

apps/aevm/src/aevm_eeevm_state.erl
```diff
@@ -119,6 +124,8 @@ init(#{ env  := Env
...
+         , on_chain_trees  => OnChainTrees
```
lucafavatella: Based on code inspection, this is not used. Would it be possible to remove it? The code is a bit complex, this would ease following it.

apps/aevm/src/aevm_eeevm_state.erl
lucafavatella: This would affect changes in apps/aebytecode/test/asm_code/identity.aesm too.

apps/aechannel/src/aesc_utils.erl
velzevur: Made it thighter in [63cc334](https://github.com/aeternity/epoch/pull/1592/commits/63cc334feecb7241a237d33ffde8fb558b79aa79)

apps/aechannel/src/aesc_offchain_update.erl
velzevur: I guess it would be `yes but later`. Tracked in PT [SC contracts: refactor aect_channel_contract](https://www.pivotaltracker.com/story/show/160992661)

apps/aecore/src/aec_vm_chain.erl
velzevur: Modifying of trees that are not expected to be modified is tracked as PT [SC contract: register a name (negative test)](https://www.pivotaltracker.com/story/show/160992836)

apps/aechannel/src/aesc_utils.erl
lucafavatella: > For example when a participant is trying to spend more than one has.

I see only in `check_min_amt`. And I see that check on the channel reserve is not enforced in the force progress tx check callback, that means that tx that violates reserve would be considered valid hence included in block and all gas limit taken from sender. This seems to make sense as a way of punishing user that produces attempt at illegal retrieval from channel reserve knowingly (as sender can specifies next round) though I recommend this to be clearly documented.

PR #1593, opened 2018-09-26
Fix and test serialize/deserialize log entry in call.

happi:


lucafavatella:
Waiting for CI though seems good to merge.

PR #1595, opened 2018-09-26
Don't change the timestamp of micro candidates

gorillainduction:
Since contract calls in the transactions might already have used the
timestamp of the block, the timestamp cannot be adjusted after
transactions have been applied.

Part of (but not completely solving)
https://www.pivotaltracker.com/story/show/160681645

EDIT:
New ticket for this PR https://www.pivotaltracker.com/story/show/160790442

### DIFF
--- a/apps/aecore/src/aec_block_micro_candidate.erl
+++ b/apps/aecore/src/aec_block_micro_candidate.erl
@@ -117,10 +117,7 @@ int_create_block(PrevBlockHash, PrevBlock, KeyBlock, Trees, Txs) ->
     Height = aec_blocks:height(KeyBlock),
     Version = aec_hard_forks:protocol_effective_at_height(Height),
 
-    MicroBlockCycle = aec_governance:micro_block_cycle(),
-    LastTime = aec_blocks:time_in_msecs(PrevBlock),
-    Time = max(aeu_time:now_in_msecs(), MicroBlockCycle + LastTime),
-
+    Time = determine_new_time(PrevBlock),
     KeyHeader = aec_blocks:to_header(KeyBlock),
     Env = aetx_env:tx_env_from_key_header(KeyHeader, PrevKeyHash,
                                           Time, PrevBlockHash),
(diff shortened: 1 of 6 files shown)


apps/aecore/src/aec_block_micro_candidate.erl
```diff
@@ -117,7 +117,9 @@ int_create_block(PrevBlockHash, PrevBlock, KeyBlock, Trees, Txs) ->
     Height = aec_blocks:height(KeyBlock),
     Version = aec_hard_forks:protocol_effective_at_height(Height),
 
-    Time = aeu_time:now_in_msecs(),
+    MicroBlockCycle = aec_governance:micro_block_cycle(),
+    LastTime = aec_blocks:time_in_msecs(PrevBlock),
+    Time = max(aeu_time:now_in_msecs(), MicroBlockCycle + LastTime),
```
lucafavatella: This seems correct only if [previous block was microblock - not if previous block was key block](https://github.com/aeternity/protocol/blame/216819c3629d333ae455c0fb51d13009cfed961d/consensus/consensus.md#L161-L164).

lucafavatella:
Also could you please this in separate smaller ticket for release?

apps/aecore/src/aec_block_micro_candidate.erl
gorillainduction: Agreed. I will base the decision on the previous block type.

apps/aecore/src/aec_block_micro_candidate.erl
```diff
@@ -139,6 +136,18 @@ int_create_block(PrevBlockHash, PrevBlock, KeyBlock, Trees, Txs) ->
     BlockInfo = #{ trees => Trees2, txs_tree => TxsTree, tx_env => Env},
     {ok, NewBlock, BlockInfo}.
 
+determine_new_time(PrevBlock) ->
+    LastTime = aec_blocks:time_in_msecs(PrevBlock),
+    case aec_blocks:type(PrevBlock) of
+        key ->
+            %% Just make sure we are progressing time.
+            max(aeu_time:now_in_msecs(), LastTime);
```
lucafavatella: /me checkng if [`+1`](https://github.com/aeternity/protocol/blame/216819c3629d333ae455c0fb51d13009cfed961d/consensus/consensus.md#L164) needed...

lucafavatella: Seems +1 needed https://github.com/aeternity/epoch/blob/1128ea24bbf39c4dd71b528c8bac2621ca6edee3/apps/aecore/src/aec_headers.erl#L715

Not surely an issue, though blocking for being sure.

PR #1596, opened 2018-09-26
PT-160789685 Improve /status endpoint test

jur0:
[PT-160789685](https://www.pivotaltracker.com/n/projects/2124891/stories/160789685)

This PR adds stricter `/status` endpoint checks so there is no confusion what type, for example, `difficulty` is (int vs. float).


lucafavatella:
And I see that swagger.yaml in master is already detailed, that is good.

PR #1597, opened 2018-09-26
Pt fix aest oracles system test

ThomasArts:
Test may fail in clean build, because binaries are used instead of atoms to match again. This PR fixes the use of atoms as well as makes mining faster.


lucafavatella:
I see this affects oracles and channels system tests.

lucafavatella:
> I see this affects oracles and channels system tests.

And neither of these are run by [system smoke tests by CI](https://github.com/aeternity/epoch/blob/9fb4a0496068e465f27a0201fe28ad45c4b363e1/Makefile#L192).

lucafavatella:
It means this PR has zero impact on CI run on PR. Merging without waiting for CI then.

PR #1600, opened 2018-09-27
PT-158023244 3rd party Oracle and AENS operations

hanssv:
To allow a contract to perform Oracle and AENS operations on behalf of a third
    party we need to provide and check a Signature. 

Protocol changes: https://github.com/aeternity/protocol/pull/213


hanssv:
Having signatures being passed directly to the primitive operations (`Oracle.register`, ..., `AENS.revoke`) is simple, and intuitive - but it puts a lot of burden on the contract writer (and also on the signer, who really should audit the contract before signing!). 

One alternative would be (this is what Tobias suggested?) is to add signatures to contract call transactions - effectively signing the transaction, and by doing so temporarily delegating the power to the contract. From an implementation perspective it would be rather different than the current setup, not very hard to change, but a bit of work. From a usability perspective it would easier to review what the particular action/call is doing. On the other hand, having to sign each operation makes certain use cases harder to implement (for example writing a contract that sells a name would require some serious juggling...).

happi:
I think the current solution is good as long as we can clearly document the powers you give away and have a couple of example contracts of "safe" usage.

hanssv:
Track the documentary/educational task in https://www.pivotaltracker.com/story/show/160824696

apps/aevm/src/aevm_ae_primops.erl
```diff
@@ -350,3 +352,6 @@ state_gas_cost(Tag, {delta, TTL}) ->
     aec_governance_utils:state_gas_cost(
       aec_governance:state_gas_cost_per_block(Tag),
       TTL).
+
+to_sign({W1, W2}) ->
+    <<W1:256, W2:256>>.
```
lucafavatella: I see that, because sign_t is `{tuple, [word, word]}` (in this PR) and [`aest_data:from_binary` decodes `word` as 256 bits unsigned](https://github.com/aeternity/epoch/blob/cbb26b2f8588014efc719bb8f9337f52dba62175/apps/aesophia/src/aeso_data.erl#L175) so not loss of data. Ok.

apps/aecore/src/aec_vm_chain.erl
```diff
@@ -138,26 +138,18 @@ oracle_register(AccountKey,_Sign, QueryFee, TTL, QueryFormat, ResponseFormat,
...
-    %%       we need a safe way to sign the register call.
-    %%       It should probably do with sign(PubKey+Nonce)
-    %%       Then we need to check that signature here.
-    %% Registering an oracle on the contract is ok.
     Result =
-        if AccountKey =:= ContractKey -> apply_transaction(Tx, State);
-           true ->
-                %% TODO: Check that Sign is correct for external accounts.
-                {error, signature_check_failed}
+        case check_account_signature(AccountKey, ContractKey, Signature) of
```
lucafavatella: * (A) I see `AccountKey` is the first formal parameter of the `Oracle.register` (I checked `aeso_ast_to_icode` and `aevm_ae_primops`).
  * This is published in either contract bytecode on chain, or in contract bytecode in contract create tx, or call arguments in create-or-call tx. Ok.
* (B) I see `ContractKey` is taken from the chain state that is initialized in [`aect_dispatch`](https://github.com/aeternity/epoch/blob/0118903d4d1ddbfe52bec553d4c26cd21d388ab3/apps/aecontract/src/aect_dispatch.erl#L87) from `contract` map field from [contract call tx](https://github.com/aeternity/epoch/blob/0118903d4d1ddbfe52bec553d4c26cd21d388ab3/apps/aecontract/src/aect_call_tx.erl#L230) (I did not check create tx) that is [recipient i.e. the contract whose code is being executed](https://github.com/aeternity/epoch/blob/0118903d4d1ddbfe52bec553d4c26cd21d388ab3/apps/aecontract/src/aect_call_tx.erl#L224).

In the following I focus only on contract call tx - not contract create tx. (Though similar observations may apply there too.)

I see that (B) is the address of the contract that has among its code the `CALL` instruction with arguments for calling the primop. The contract is on the chain so it is public. Ok.

I see this line does "if AccountKey equal to ContractKey then return ok otherwise check that Signature is valid signature by private key of AccountKey of binary `AccountKey++ContractKey`". Ok. So this is checking that private key of `AccountKey` is needed. Ok.

Observations: this seems to me prone to replay attacks though I may be wrong - I shall formulate some concrete examples. (I may be unable to do this before Mon.)

hanssv:
I did look for existing HTTP tests, but found none... Started writing one, but got stuck trying to represent a variant type (for Oracles) - the input format needs an overhaul and while doing that we should add more tests cases showing usage of the HTTP API.

PR #1604, opened 2018-09-27
PT-160256699 Add gas_price/1 to all txs

jur0:
This is a preparatory PR for [PT-160256699](https://www.pivotaltracker.com/n/projects/2124891/stories/160256699).

It adds `gas_price/1` to all txs and a fixed gas price to governance (this will change in future PRs). All txs have a fixed gas price except for contract create and contract call txs.

### DIFF
--- a/apps/aechannel/src/aesc_close_mutual_tx.erl
+++ b/apps/aechannel/src/aesc_close_mutual_tx.erl
@@ -13,6 +13,7 @@
          type/0,
          fee/1,
          gas/1,
+         gas_price/1,
          ttl/1,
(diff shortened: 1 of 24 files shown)


velzevur:
@jur0 I see the `for_client/1` callback does not show the `gas` and `gas_price` and thus - no changes are required in the `swagger.yaml`
Because the `gas` and `gas_price` are not exposed in the API for creation of transactions but yet transactions with a lower `gas` or `gas_price` can be created by a modified `epoch` code so the check in the mempool shall catch those.

Ok.

PR #1607, opened 2018-09-28
Add basic system tests for the naming service

sylane:


system_test/aest_names_SUITE_data/epoch.yaml.mustache
```diff
@@ -0,0 +1,45 @@
...
+{{/peers}}
+
+sync:
+    port: {{services.sync.port}}
+    single_outbound_per_group: false
+
+http:
+    external:
+        port: {{services.ext_http.port}}
+    internal:
```
ThomasArts: Since there is no API call to any of the debug API, please remove line 15 to 20.
I checked that the tests still pass when doing so.

(It's more a system test if we only use the external API)

PR #1609, opened 2018-10-01
Stabilize sync system test for unexpected forks

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/160892233

Supersedes https://github.com/aeternity/epoch/pull/1608

Depends on https://github.com/aeternity/epoch/pull/1612 - that is preparatory PR for this so that this can be squashed-and-merged without squashing relevant mostly-independent patches.

Depends on #1615 - that is preparatory PR for this so that name resolution times can be configured.

TODO:
* [x] ~Once PR #1612 merged, edit target branch to master.~
* [x] Once PR #1615 merged, edit target branch to master.
* [ ] Squash-and-merge.

PR #1612, opened 2018-10-01
Preparation for stabilizing sync system test for unexpected forks

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/160892233

Preparatory PR for #1609


lucafavatella:
> There is a function aest_nodes:get_node_config that could be used to get the ping_interval.

Addressed. By force-pushing in this branch.

PR #1614, opened 2018-10-02
Refactoring/optimization of mempool and jobs queues

uwiger:
PT-160591670

- mempool GC broken out into separate process
- tx push checking done in client process
- some linear-search operations eliminated in mempool GC
- jobs queues added for HTTP API
- jobs queue definitions moved to JSON-schema/epoch config
- queues renamed for consistency


uwiger:
@ThomasArts the change to the close_mutual_tx could possibly have been broken out into a separate PR, but was needed because the `origin/1` function wasn't deterministic (depended on the existence of the channel object), which could cause the mempool GC to fail to delete the tx from the pool.

PR #1615, opened 2018-10-02
Make peer host name resolution backoff times configurable

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/160924468

Needed for [stabilization of sync system test](https://github.com/aeternity/epoch/pull/1609#discussion_r221890089). Split for easing review and merge.

TODO:
* [x] Add release notes line.
* [ ] Merge (preferably without squashing so not to require further rebase of #1609 )

### DIFF
--- a/apps/aeutils/priv/epoch_config_schema.json
+++ b/apps/aeutils/priv/epoch_config_schema.json
@@ -118,7 +118,7 @@
                     "default" : 7
                 },
                 "resolver_backoff_times" : {
-                    "description" : "Waiting time intervals (milliseconds) before each retry at resolving the host name of peers (both trusted and untrusted).",
+                    "description" : "Waiting time intervals (milliseconds) before each retry at resolving the host name of peers (both trusted and untrusted). Retries further to the length of the array reuse the last item of the array.",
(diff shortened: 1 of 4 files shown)


apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -111,6 +111,23 @@
...
                     "default": true,
                 },
+                "resolver_max_retries": {
+                    "description" : "Maximum number of retries at resolving the host name of untrusted peers.",
+                    "type" : "integer",
+                    "minimum" : 0,
+                    "default" : 7
+                },
+                "resolver_backoff_times" : {
+                    "description" : "Waiting time intervals (milliseconds) before each retry at resolving the host name of peers (both trusted and untrusted).",
```
sylane: Maybe we should add that if there is less values than `resolve_max_retries`, the last one is reused...

lucafavatella: You are right. Let me add that.

lucafavatella: Addressed in 77a68efe

PR #1616, opened 2018-10-03
Pt 160895401 chunk synced generation

gorillainduction:
https://www.pivotaltracker.com/story/show/160895401

Also added some refactorings/optimizations while I was in the vicinity.


ThomasArts:
Great refactoring, 4 micro blocks added always within 0.5 second. From sync against UAT in 254 seconds moved to 89 seconds and no risk we timeout on aec_conductor with some more heavy transactions in a block!

PR #1618, opened 2018-10-03
PT-160921064 Rename tx types to swagger definitions

jur0:
[PT-160921064](https://www.pivotaltracker.com/n/projects/2124891/stories/160921064)

This PR renames tx type to swagger definition name (e.g. `spend_tx` is now `SpendTx`). The tx type is used as discriminator in generic tx in swagger spec. The change was requested by the apps team.

### DIFF
--- a/apps/aetx/src/aetx.erl
+++ b/apps/aetx/src/aetx.erl
@@ -24,9 +24,7 @@
         , signers/2
         , specialize_type/1
         , specialize_callback/1
-        , update_tx/2
-        , tx_type/1
-        , tx_types/0]).
+        , update_tx/2]).
 
 -ifdef(TEST).
 -export([tx/1]).
@@ -157,12 +155,6 @@
 new(Callback, Tx) ->
     Type = Callback:type(),
     #aetx{ type = Type, cb = Callback, tx = Tx }.
-
--spec tx_type(TxOrTxType :: tx_type() | tx()) -> binary().
-tx_type(#aetx{ type = TxType }) ->
-    type_to_binary(TxType);
-tx_type(TxType) when is_atom(TxType) ->
-    type_to_binary(TxType).
 
 -spec fee(Tx :: tx()) -> Fee :: integer().
 fee(#aetx{ cb = CB, tx = Tx }) ->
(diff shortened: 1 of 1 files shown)


apps/aetx/src/aetx.erl
```diff
@@ -298,6 +298,30 @@ type_to_cb(channel_settle_tx)         -> aesc_settle_tx;
 type_to_cb(channel_snapshot_solo_tx)  -> aesc_snapshot_solo_tx;
 type_to_cb(channel_offchain_tx)       -> aesc_offchain_tx.
 
+type_to_binary(spend_tx)                  -> <<"SpendTx">>;
```
ThomasArts: This refactoring introduces a non-reversable operation. Before you could to atom_to_binary and binary_to_atom back if need be. Now you rename and have no easy way to get your atom back.

Hard to see whether this has consequences for other parts of the code.

Therefore, I propose to compute the name instead of hard code it and to have type_to_binary and binary_to_type as part of the code.

A simple way to get from _ notation to camel notation is (not super beautiful, but works):
unicode:characters_to_binary([ string:titlecase(W) || W <- string:split(atom_to_binary(TxType, utf8), "_", all)]).

ThomasArts: Alternative (slightly faster in execution) you add the reverse for each keyword. But that is error prone and hard to maintain.

The best thing would be if we generated code for this in swagger_endpoints and use that... but that would introduce new dependencies.

jur0: I don't think the reverse function is needed. We don't specify the tx type when we create a tx e.g. using the `/debug/*` endpoints - each tx type has its own endpoint so the tx type is not needed and so we don't have to convert it to atom in epoch.

velzevur: Ok, then a couple of observations

1. What is see is that the exported function `aetx:tx_type/1` is not used outside of the scope of this module

2. This function `type_to_binary/1` is only used in the `tx_type/1` and nowhere else

3. `tx_type/1` is used locally only in `serialize_for_client/1` that represents the transaction for client cosumption.

Based on these three points, I'd suggest:

1. Removing `tx_type/1`

2. Renaming `tx_type/1` to a more suitable name, that represents that this is only the swagger definition name there (`tx_swagger_type/1` maybe?) and using it in `serialize_for_client/1`. If so - we probably don't need the reverse function, indeed.

velzevur: We're probably safe using just `aetx:specialize_type/1` and `aetx:specialize_callback/1` depending on the specific case and this has no impact on any of them whatsoever.

PR #1619, opened 2018-10-04
Pt 160920700 Improve stability of post tx

fabiankrol:
Introduce `ensure_transaction_posted` helper function. After posting signed transaction it waits for:
- new keyblock to avoid micro forks)
- tx on chain
---
I removed redundant `common.wait_until_height` calls in spend_tx suite, `common.send_tokens_to_unchanging_user_and_wait_balance` is doing the job

### DIFF
--- a/py/tests/integration/common.py
+++ b/py/tests/integration/common.py
@@ -208,6 +208,14 @@ def genesis_hash(api):
 def wait_until_height(api, height):
     wait(lambda: api.get_current_key_block().height >= height, timeout_seconds=120, sleep_seconds=0.25)
 
+def ensure_transaction_posted(ext_api, signed_tx):
+    tx_object = Tx(tx=signed_tx)
+    tx_hash = ext_api.post_transaction(tx_object).tx_hash
+    top = ext_api.get_current_key_block()
+    wait_until_height(ext_api, top.height + 1)
+    wait(lambda: ext_api.get_transaction_by_hash(tx_hash).block_hash != 'none',
+         timeout_seconds=20, sleep_seconds=0.25)
+
 def get_account_balance(api, pub_key):
     return _balance_from_get_account(lambda: api.get_account_by_pubkey(pub_key), pub_key)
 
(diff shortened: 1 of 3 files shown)


py/tests/integration/common.py
```diff
@@ -208,6 +208,14 @@ def genesis_hash(api):
 def wait_until_height(api, height):
     wait(lambda: api.get_current_key_block().height >= height, timeout_seconds=120, sleep_seconds=0.25)
 
+def ensure_transaction_posted(ext_api, signed_tx):
+    tx_object = Tx(tx=signed_tx)
+    tx_hash = ext_api.post_transaction(tx_object).tx_hash
+    top = ext_api.get_current_key_block()
+    wait_until_height(ext_api, top.height + 1)
+    wait(lambda: ext_api.get_transaction_by_hash(tx_hash).block_hash != 'none',
```
ThomasArts: Assume top = 7 with chain [Key . micro1] at the moment you query for top (get_current_key_block)
Now the transaction arrives in micro2 making the chain [Key . micro1 . micro2].
You wait a next key block, but that may have been mined on micro1, making the chain [Key . micro1 . Key], the micro2 block became a micro fork.
Not sure what exactly get_transaction_by_hash does, but it may well return height -1, meaning it is in the mempool again.

Probably this scenario is covered, but I am simply not sure here.

velzevur: @ThomasArts this can happen in a multinode environment where different nodes compete to produce blocks. Although this is not the case for most UAT tests, this still can happen in a few of them.

Indeed function could have an improved stability if we mine a keyblock(s) on top of the microblock that included the tx.

velzevur:
The rests seems great

py/tests/integration/common.py
fabiankrol: @ThomasArts that's exactly the scenario we're covering here. 
Let's assume we have; `[KeyA, micro1, micro2{our_tx_here}]` after mining forked `KeyB` we have `[KeyA, micro1, Key7]` and we're putting back our tx into mempool. `get_transaction_by_hash`shows it. We're suspended on on 2nd wait (line: 216)

@velzevur 👌 I'll create new PT issue fot it.

PR #1620, opened 2018-10-04
fix aeu_env:schema() when started without epoch.yaml

uwiger:
See [PT #160984234](https://www.pivotaltracker.com/story/show/160984234)


velzevur:
I also tested it locally.

PR #1623, opened 2018-10-06
PT 160993470 implement abort in Sophia

rvirding:
Implement the abort function as a built-in in Sophia. The compiler now generates a local abort function which will be called. Any user defined abort functions are removed.

This implements [PT-160993470](https://www.pivotaltracker.com/story/show/160993470).


lucafavatella:
Could you please update https://github.com/aeternity/protocol/blame/6926021540d90425d454938506f730b4a1098f34/contracts/sophia.md#L759-L760

PR #1624, opened 2018-10-08
Force progress test: register name

velzevur:
PT [SC contract: register a name (negative test)](https://www.pivotaltracker.com/story/show/160992836)
Edit: scope of PR changed, please check [comment](https://github.com/aeternity/epoch/pull/1624#issuecomment-428886778)

### DIFF
--- a/apps/aechannel/test/aesc_txs_SUITE.erl
+++ b/apps/aechannel/test/aesc_txs_SUITE.erl
@@ -3012,7 +3012,7 @@ fp_register_name(Cfg) ->
                   
     Test =
         fun(Owner, Forcer) ->
-            ?TEST_LOG("Name preclaimed off-chain, owner is ~p, forcer is ~p",
+            ?TEST_LOG("Name preclaim off-chain, owner is ~p, forcer is ~p",
                       [Owner, Forcer]),
             ContractCreateRound = 10,
             run(#{cfg => Cfg},
@@ -3073,7 +3073,7 @@ fp_register_name(Cfg) ->
                       GasLimit = aect_call:gas_used(Call),
                       GasPrice = aect_call:gas_price(Call),
                       %% the default catch all reason for error
-                      <<>> = aect_call:return_value(Call),
+                      <<"not_allowed_off_chain">> = aect_call:return_value(Call),
                       error = aect_call:return_type(Call),
 
                       %% expected channel states
--- a/apps/aechannel/test/aesc_txs_SUITE.erl
+++ b/apps/aechannel/test/aesc_txs_SUITE.erl
@@ -150,7 +150,10 @@
          fp_missing_contract_address/1,
 
          fp_insufficent_tokens/1,
-         fp_too_soon/1
+         fp_too_soon/1,
+
+         %
+         fp_register_name/1
         ]).
 
 -include_lib("common_test/include/ct.hrl").
@@ -307,7 +310,8 @@ groups() ->
        fp_missing_contract_address,
 
        fp_insufficent_tokens,
-       fp_too_soon
+       fp_too_soon,
+       fp_register_name
       ]}
     ].
 
@@ -2873,6 +2877,90 @@ fp_too_soon(Cfg) ->
 
     % height is one less, different lock_period
     Test(100, 200, 100),
+    ok.
+
+fp_register_name(Cfg) ->
+    Name = <<"lorem.test">>,
+    Salt = 42,
+    {ok, NameAscii} = aens_utils:to_ascii(Name),
+    CHash           = aeu_hex:hexstring_encode(
+                        aens_hash:commitment_hash(NameAscii, Salt)),
+    #{ public := Pubkey, secret := Privkey } = enacl:sign_keypair(),
+    Account = aeu_hex:hexstring_encode(Pubkey),
+    StateHashSize = aec_base58c:byte_size_for_type(state),
+    StateHash = <<42:StateHashSize/unit:8>>,
+    Round = 42,
+    FPRound = Round + 10,
+    Test =
+        fun(Owner, Forcer) ->
+            ContractCreateRound = 10,
+            run(#{cfg => Cfg},
+               [positive(fun create_channel_/2),
+                % store state on-chain via snapshot
+                set_from(initiator),
+                set_prop(round, 42),
+                set_prop(state_hash, StateHash),
+                positive(fun snapshot_solo_/2),
+                fun(#{state := S,
+                      channel_pubkey := ChannelPubKey} = Props) ->
+                    Channel = aesc_test_utils:get_channel(ChannelPubKey, S),
+                    Round = aesc_channels:round(Channel),
+                    StateHash = aesc_channels:state_hash(Channel),
+                    Props
+                end,
+                % create contract off-chain
+                create_trees_if_not_present(),
+                set_from(Owner, owner, owner_privkey),
+                create_contract_in_trees(_Round    = ContractCreateRound,
+                                         _Contract = "aens",
+                                         _InitArgs = <<"()">>,
+                                         _Deposit  = 2),
+                % force progress contract on-chain
+                fun(#{contract_id := ContractId} = Props) ->
+                    BinToSign = <<Pubkey/binary, ContractId/binary>>,
+                    <<Word1:256, Word2:256>> = enacl:sign_detached(BinToSign,
+                                                                   Privkey),
+                    Sig = aeu_hex:hexstring_encode(
+                            aeso_data:to_binary({Word1, Word2-2})),
+                    PreclaimArgs = <<"(", Account/binary, ",",
+                                          CHash/binary, ",",
+                                          Sig/binary,
+                                      ")">>,
+                    (force_call_contract_first(Forcer, <<"preclaim">>,
+                                          PreclaimArgs, FPRound))(Props)
+                end,
+                % ensure all gas is consumed and channel is updated
+                fun(#{state := S,
+                      channel_pubkey := ChannelPubKey,
+                      signed_force_progress := SignedForceProgressTx,
+                      solo_payload := #{update    := Update,
+                                       state_hash := ExpectedStateHash,
+                                       round      := ExpectedRound}} = Props) ->
+                    {_ContractId, Caller} = aesc_offchain_update:extract_call(Update),
+                    TxHashContractPubkey = aesc_utils:tx_hash_to_contract_pubkey(
+                                          aetx_sign:hash(SignedForceProgressTx)),
+                    CallId = aect_call:id(Caller,
+                                          FPRound,
+                                          TxHashContractPubkey),
+                    Call = aect_test_utils:get_call(TxHashContractPubkey, CallId,
+                                                    S),
+                    {_, GasPrice, GasLimit} = aesc_offchain_update:extract_amounts(Update),
+                    %% assert all gas was consumed
+                    GasLimit = aect_call:gas_used(Call),
+                    GasPrice = aect_call:gas_price(Call),
+                    %% the default catch all reason for error
+                    <<>> = aect_call:return_value(Call),
+
+                    %% expected channel states
+                    Channel = aesc_test_utils:get_channel(ChannelPubKey, S),
+                    FPRound = aesc_channels:round(Channel),
+                    FPRound = ExpectedRound,
+                    ExpectedStateHash = aesc_channels:state_hash(Channel),
+                    Props
+                end])
+        end,
+    [Test(Owner, Forcer) || Owner  <- ?ROLES,
+                            Forcer <- ?ROLES],
     ok.
 
 create_contract_poi_and_payload(Round, ContractRound, Owner) ->
--- a/apps/aechannel/test/aesc_txs_SUITE.erl
+++ b/apps/aechannel/test/aesc_txs_SUITE.erl
@@ -152,7 +152,7 @@
          fp_insufficent_tokens/1,
          fp_too_soon/1,
 
-         %
+         % off-chain name registration not allowed
          fp_register_name/1
         ]).
 
@@ -2921,7 +2921,7 @@ fp_register_name(Cfg) ->
                     <<Word1:256, Word2:256>> = enacl:sign_detached(BinToSign,
                                                                    Privkey),
                     Sig = aeu_hex:hexstring_encode(
-                            aeso_data:to_binary({Word1, Word2-2})),
+                            aeso_data:to_binary({Word1, Word2})),
                     PreclaimArgs = <<"(", Account/binary, ",",
                                           CHash/binary, ",",
                                           Sig/binary,
--- a/apps/aechannel/test/aesc_txs_SUITE.erl
+++ b/apps/aechannel/test/aesc_txs_SUITE.erl
@@ -165,6 +165,12 @@
 -define(BOGUS_CHANNEL, <<0:?MINER_PUB_BYTES/unit:8>>).
 -define(ROLES, [initiator, responder]).
 -define(VM_VERSION, ?AEVM_01_Sophia_01).
+-define(TEST_LOG(Format, Data),
+        try ct:log(Format, Data)
+        catch
+            %% Enable setting up node with "test" rebar profile.
+            error:undef -> ok
+        end).
 %%%===================================================================
 %%% Common test framework
 %%%===================================================================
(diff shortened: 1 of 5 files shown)


velzevur:
`test` failed with unrelated error which is tracked as [Investigate IF with rewards](https://www.pivotaltracker.com/story/show/161059429)

apps/aechannel/test/aesc_txs_SUITE.erl
```diff
@@ -150,7 +150,10 @@
          fp_missing_contract_address/1,
 
          fp_insufficent_tokens/1,
-         fp_too_soon/1
+         fp_too_soon/1,
+
+         % off-chain name registration not allowed
```
lucafavatella: I see the related spec is

https://github.com/aeternity/protocol/blame/d081b9bacd10a3eeaf680902f9258c76e6e162e7/channels/OFF-CHAIN.md#L1041-L1042

apps/aechannel/test/aesc_txs_SUITE.erl
```diff
@@ -2875,6 +2879,90 @@ fp_too_soon(Cfg) ->
...
+                                          FPRound,
+                                          TxHashContractPubkey),
+                    Call = aect_test_utils:get_call(TxHashContractPubkey, CallId,
+                                                    S),
+                    {_, GasPrice, GasLimit} = aesc_offchain_update:extract_amounts(Update),
+                    %% assert all gas was consumed
+                    GasLimit = aect_call:gas_used(Call),
+                    GasPrice = aect_call:gas_price(Call),
+                    %% the default catch all reason for error
+                    <<>> = aect_call:return_value(Call),
```
lucafavatella: Would it be possible to assert the return type too - for readability of the test?

apps/aechannel/test/aesc_txs_SUITE.erl
```diff
@@ -2875,6 +2879,90 @@ fp_too_soon(Cfg) ->
...
+                    BinToSign = <<Pubkey/binary, ContractId/binary>>,
+                    <<Word1:256, Word2:256>> = enacl:sign_detached(BinToSign,
+                                                                   Privkey),
+                    Sig = aeu_hex:hexstring_encode(
+                            aeso_data:to_binary({Word1, Word2})),
+                    PreclaimArgs = <<"(", Account/binary, ",",
+                                          CHash/binary, ",",
+                                          Sig/binary,
+                                      ")">>,
+                    (force_call_contract_first(Forcer, <<"preclaim">>,
```
lucafavatella: It seems (I am not 100% sure, have not followed fully) that this calls `create_contract_call_payload` that calls `aesc_offchain_update:apply_on_trees`. That calls `aect_channel_contract:run` that calls the VM that [errs](https://35097-99802036-gh.circle-artifacts.com/1/home/builder/epoch/_build/test/logs/ct_run.epoch_ct@localhost.2018-10-08_16.08.20/lib.aechannel.aesc_txs_SUITE.logs/run.2018-10-08_16.11.29/aesc_txs_suite.fp_register_name.html) (not saying it is bad, just getting my head around it)
```
Return error error:{badmatch,
                       {error,
                           {badmatch,
                               <<0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
...
```
that, by enabling more stacktraces locally, I see it is on `aeso_data:heap_word(Heap,Addr=9039081679025068027825624833915204708939891774825472053301238029925278094596)` called by `aevm_eeevm_state:get_contract_call_input` at `{ok, Arg}  = aeso_data:from_heap(Type, Heap, Ptr)`. Is this the intended failure? (Getting my head around this.)

velzevur: It shouldn't be. _investigating_

apps/aechannel/test/aesc_txs_SUITE.erl
```diff
@@ -2875,6 +2879,90 @@ fp_too_soon(Cfg) ->
...
+                      channel_pubkey := ChannelPubKey} = Props) ->
+                    Channel = aesc_test_utils:get_channel(ChannelPubKey, S),
+                    Round = aesc_channels:round(Channel),
+                    StateHash = aesc_channels:state_hash(Channel),
+                    Props
+                end,
+                % create contract off-chain
+                create_trees_if_not_present(),
+                set_from(Owner, owner, owner_privkey),
+                create_contract_in_trees(_Round    = ContractCreateRound,
```
lucafavatella: I see from this point on the test is preparing the contract object on the state tree in order to then call the preclaim primop and assert outcome. As often the case in negative test cases, it may happen that assertion succeeds but because of test defect (e.g. because indirectly used code, e.g. preclaim primop, changes while test not updated) rather than expected reason. Would it be feasible to codify in one code path both the assertion that the preclaim would work if called with correct preconditions, and the call of the *same* preclaim with the preconditions of the test case? The precondition seems the chain state to have on-chain trees - rather than off-and-on-chain trees.

Though this may require too much effort so optional.

apps/aechannel/test/aesc_txs_SUITE.erl
lucafavatella: This calls `aesc_offchain_update:op_call_contract` with 5th argument being variable `CallData`. Could you please confirm the type spec of `aesc_offchain_update:op_call_contract` in the 5th argument? (I see `aect_call:call()` though following this test I expect `binary()`.)

lucafavatella: See also type of  4th argument of aect_channel_contract:run.

lucafavatella: I see this calls `CallData = aect_sophia:create_call(Code, Fun, Args)` without checking whether return value is `{error, _}` or binary (binary return is not tagged with `{ok, _}`).  If it returned the `{error, _}` then execution would fall at https://github.com/aeternity/epoch/blob/e6940492d22ea11e79df55575f3552e86a47752f/apps/aevm/src/aevm_eeevm_state.erl#L166 that is the first point that checks that it is binary, and sets gas to zero, that - if happenend - I am concerned may make this test test something else than intended, without failing (did not double check). Would you please consider making `CallData = aect_sophia:create_call(Code, Fun, Args)` more strict by at least failing if not a binary? (I would prefer a tagged `{ok, CallData}` return though not sure why not tagged with `{ok, _}` already.) It would at least improve readability of the test (that is already complex).

apps/aechannel/test/aesc_txs_SUITE.erl
```diff
@@ -2875,6 +2879,90 @@ fp_too_soon(Cfg) ->
...
+                fun(#{contract_id := ContractId} = Props) ->
+                    BinToSign = <<Pubkey/binary, ContractId/binary>>,
+                    <<Word1:256, Word2:256>> = enacl:sign_detached(BinToSign,
+                                                                   Privkey),
+                    Sig = aeu_hex:hexstring_encode(
+                            aeso_data:to_binary({Word1, Word2})),
+                    PreclaimArgs = <<"(", Account/binary, ",",
+                                          CHash/binary, ",",
+                                          Sig/binary,
+                                      ")">>,
```
lucafavatella: I see Account, CHash and Sig hex encoded. I see `aeso_parser:string` takes care of decoding. Ok.

velzevur:
Rebased on top of master. Exposed the problem and fixed it. If we decide this is the way to go, will implement it for other primops as well.

velzevur:
This PR consists of:
* formal verification that the `aens.aes` contract `preclaim` function can be called on-chain with the parameters provided
* formal verification that the `aens.aes` contract `preclaim` function can NOT be called off-chain with the parameters provided. Check is done only for forcing of progress as we have control only for on-chain protocol
* added proper checks in `aec_vm_chain` that certain chain objects changing actions are not allowed

What this PR does not consists:
* check that Oracle can not be registered off-chain. Tracked as PT [SC contract: register an oracle (negative test)](https://www.pivotaltracker.com/story/show/161144955)

Some observations regarding Sophia. The `aens.aes` contract has the following function
```
  function preclaim(addr  : address,          // Claim on behalf of this account (can be Contract.address)
                    chash : hash,             // Commitment hash
                    sign  : signature) : () = // Signed by addr (if not Contract.address)
    AENS.preclaim(addr, chash, sign)
```
the `hash` type is described as a [hex encoded](https://github.com/aeternity/protocol/blame/master/contracts/sophia.md#L546) which is good. Question is regarding the signature. In one place it is said that [it is 64 bytes](https://github.com/aeternity/protocol/blame/master/contracts/sophia.md#L137) which is correct. On other place it is said to be [pair of integers](https://github.com/aeternity/protocol/blame/master/contracts/sophia.md#L890). Internally it is indeed represented as a pair of integers.

An example of a Sophia encoded signature will look like `(41036886491346345127962308224290434319858957269112974056384210486397156156853, 29166479995643099920475598104950177374958912009982571784304636843863216677124)` - a tuple of two numbers

Questions:
* Do we want the signature represented as a hex encoded binary instead (as we treat account addresses, hashes and etc)
* Should we add an example of what a singature looks like [here](https://github.com/aeternity/protocol/blame/master/contracts/sophia.md#L137)

velzevur:
Rebased on top of  `master`

hanssv:
> > Should we add an example of what a singature looks like [here](https://github.com/aeternity/protocol/blame/master/contracts/sophia.md#L137)
> 
> From [aeternity/protocol#213 (comment)](https://github.com/aeternity/protocol/pull/213#discussion_r221206812) :
> 
> > You can't write a literal signature in Sophia. You have to pass it to the contract as an argument.

No we should not, since you (still) can't write a literal signature in Sophia - so that is the wrong place.

apps/aevm/src/aevm_eeevm.erl
```diff
@@ -1526,6 +1526,9 @@ recursive_call2(Op, Gascap, To, Value, OSize, OOffset, I, State8, GasAfterSpend)
                     {error, ?AEVM_PRIMOP_ERR_REASON_OOG(_OogResource, _OogCost, State9)} ->
                         ?TEST_LOG("Out of gas spending ~p gas for ~p", [_OogCost, _OogResource]),
                         eval_error(out_of_gas, State9);
+                    {error, not_allowed_off_chain} ->
+                        ?TEST_LOG("Not allowed calling this contract off-chain", []),
+                        eval_error(not_allowed_off_chain, State8);
```
lucafavatella: I see high level documentation in [state channels off chain doc - section on chain enforcement](https://github.com/aeternity/protocol/blame/26ba07353b62d236c3b9724f990479b8b4999e79/channels/OFF-CHAIN.md#L1041-L1042). Ok as high level but it feels a more detailed description would be beneficial. Please find proposal below.

It feels the not allowed primops shall be listed in https://github.com/aeternity/protocol/blame/26ba07353b62d236c3b9724f990479b8b4999e79/contracts/aevm.md#L45 like

> Calling [these VM primops] [in this situation] results in an error with output data `not_allowed_off_chain`.

apps/aechannel/test/aesc_txs_SUITE.erl
```diff
@@ -150,17 +150,27 @@
...
 -include_lib("stdlib/include/assert.hrl").
+-include_lib("apps/aecontract/src/aecontract.hrl").
 
 -define(MINER_PUBKEY, <<12345:?MINER_PUB_BYTES/unit:8>>).
 -define(BOGUS_CHANNEL, <<0:?MINER_PUB_BYTES/unit:8>>).
 -define(ROLES, [initiator, responder]).
--define(VM_VERSION, 1).
+-define(VM_VERSION, ?AEVM_01_Sophia_01).
+-define(TEST_LOG(Format, Data),
+        try ct:log(Format, Data)
```
lucafavatella: This seems used in the test only so the macro and the try-catch seem not needed, and ct:log can be called directly?

apps/aecore/src/aec_vm_chain.erl
```diff
@@ -229,7 +236,15 @@ oracle_query(Oracle, Q, Value, QTTL, RTTL,
         {error, _} = E -> E
     end.
 
-oracle_respond(Oracle, QueryId, Signature, Response, ResponseTTL,
+-spec oracle_respond(aec_keys:pubkey(), aeo_query:id(), binary(), binary(),
+                     aeo_oracles:relative_ttl(), chain_state()) ->
+    prim_op_result().
+oracle_respond(Oracle, QueryId, Signature, Response, ResponseTTL, State) ->
+    on_chain_only(State,
```
lucafavatella: Is oracle query allowed?

lucafavatella: Is spend allowed?

apps/aevm/src/aevm_eeevm.erl
velzevur: Addressed [Describe result of off-chain calls](https://github.com/aeternity/protocol/pull/235)

apps/aecore/src/aec_vm_chain.erl
velzevur: Not allowed, addressing. Thanks.

velzevur: Spend is allowed. If it is performed to an on-chain addres - on-chain changes are discarded

apps/aechannel/test/aesc_txs_SUITE.erl
```diff
@@ -2875,6 +2886,208 @@ fp_too_soon(Cfg) ->
...
+                      Channel = aesc_test_utils:get_channel(ChannelPubKey, S),
+                      FPRound = aesc_channels:round(Channel),
+                      FPRound = ExpectedRound,
+                      ExpectedStateHash = aesc_channels:state_hash(Channel),
+                      Props
+                  end])
+        end,
+    [Test(Owner, Forcer) || Owner  <- ?ROLES,
+                            Forcer <- ?ROLES],
+    ok.
```
lucafavatella: I see on-chain contracts are [not allowed to be called](https://github.com/aeternity/protocol/blame/26ba07353b62d236c3b9724f990479b8b4999e79/channels/OFF-CHAIN.md#L1053) and that implementation [ignores on-chain trees for calls](https://github.com/aeternity/epoch/blob/ee758a13d008134262bf83bd806e25b790840190/apps/aecore/src/aec_vm_chain.erl#L443) that seems ok. Though I see no test for that (out of scope for this PR). Is it tracked anywhere to eventually write one?

apps/aechannel/test/aesc_txs_SUITE.erl
```diff
@@ -2875,6 +2886,208 @@ fp_too_soon(Cfg) ->
...
+    FPRound = Round + 10,
+    SignContractAddress =
+        fun(PubK, PrivK, ConId) ->
+            BinToSign = <<PubK/binary, ConId/binary>>,
+            SigBin = <<Word1:256, Word2:256>> = enacl:sign_detached(BinToSign, PrivK),
+            %_Sig = aeu_hex:hexstring_encode(aeso_data:to_binary({Word1, Word2}, 0))
+            ?TEST_LOG("Signature binary ~p", [SigBin]),
+            Word11 = integer_to_binary(Word1),
+            Word21 = integer_to_binary(Word2),
+            <<"(", Word11/binary, ", ", Word21/binary, ")">>
```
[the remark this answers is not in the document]
lucafavatella: This is a non-issue, as integer_to_binary does the job. Please ignore.

apps/aechannel/test/aesc_txs_SUITE.erl
velzevur: Tracked as [SC contract: Call on-chain test (negative test)](https://www.pivotaltracker.com/story/show/161215899)

apps/aecore/src/aec_vm_chain.erl
lucafavatella: I see in f806355e0f43f812e3112214f51da60e93fc1bbd - thanks

velzevur:
`test` job failed with 
```

%%% aehttp_integration_SUITE ==> all.external_endpoints.pending_transactions: FAILED
%%% aehttp_integration_SUITE ==> {{badmatch,{359999999999999999878,359999999999999999879,
            {329999999999999999878,3,10000000000000000000}}},
 [{aehttp_integration_SUITE,pending_transactions,1,
                            [{file,"/home/builder/epoch/_build/test/lib/aehttp/test/aehttp_integration_SUITE.erl"},
                             {line,2961}]},
  {test_server,run_test_case_eval1,6,[{file,"test_server.erl"},{line,1062}]},
  {test_server,run_test_case_eval,9,[{file,"test_server.erl"},{line,994}]}]}```
this is already tracked as [Investigate IF with rewards](https://www.pivotaltracker.com/story/show/161059429)

PR #1629, opened 2018-10-09
PT-160412391: implement efficient maps

UlfNorell:
[PT-160412391](https://www.pivotaltracker.com/story/show/160412391)

Protocol PR coming later. For now here is an overview:

### Maps in the VM state

Maps are treated specially by the AEVM and are not stored in the VM memory. Instead map values are unique identifiers pointing into a separate component (`maps`) of the VM state associating map identifiers with a map structure `#pmap{}` with the fields

 - `key_t`, `val_t`: type reps for key and value types
 - `size`: size of the map
 - `parent`: `none | map_id()`, for maps that are modified versions of a map in the store (more on store maps below)
 - `data`: `stored | MapData`

`MapData` is `#{aeso_data:binary_value() => tombstone | aeso_data:binary_heap_value()}`. The data for a store map is `stored`. Map keys cannot contain other maps and in this case `binary_value()` and `binary_heap_value()` are the same. For map values, the important difference between `binary_value()` and `binary_heap_value()` is that heap values do not store map data, but represents maps by their identifier.

A `tombstone` value means that the key has been deleted from the map. This is only present if `parent /= none`, i.e. the map is a modification of a store map.

### Store maps

When the contract state contains maps they are stored in the Store. When the state is loaded into the VM state on contract call initialization the map is kept in the store and a `#pmap{}` record with `data = stored` is added to the VM state.

When writing an updated state to the store care is taken to reuse the old map data. Basically if a state map is used linearly (old state maps are not retained in the contract state) all modifications will be cheap inplace updates of store maps.

Maps in contract state are saved in the store using the following format:

```erlang
 ?SOPHIA_STATE_MAPS_KEY => << <<MapId:256>> || MapId <- AllMapIds >>
 <<MapId:256>>          => <<RealId:256, RefCount:256, Size:256, Bin/binary>>
     where
         RealId   % indirection to allow inplace updates
         RefCount % number of times this map occurs in the map value of another map
         Size     % the size of the map (number of keys)
         Bin      % aeso_data:to_binary({KeyType, ValType})
 <<RealId:256, KeyBin/binary>> => <<ValBin/binary>>
     KeyBin :: aeso_data:binary_value()      %% cannot contain other maps
     ValBin :: aeso_data:binary_heap_value() %% crucially: maps as map ids
```

#### Garbage collection

In order to allow inplace update of modified store maps it's important to know when a specific map is no longer used. This is also important to prevent the store from growing indefinitely. When updating the contract state we get given the maps used in the new state except those appearing inside map values. To get the latter we have to traverse the new maps (`update_ref_counts`). Crucially, we don't have to look at maps already in the store except for the keys that get modified.

If a map is updated and the original is no longer in use it will be updated in place.

### Notes

- Currently the split between aevm_eeevm_store and aevm_eeevm_maps is a bit artificial and not entirely satisfactory. Would benefit from a refactoring.

- Since maps are handled outside the VM memory some work is required to properly calibrate the gas costs for transactions, remote calls and primop calls involving maps ([PT-160257131](https://www.pivotaltracker.com/story/show/160257131)).

- Some more robust testing would be prudent. I have some simple tests, but there are a lot of corner cases not covered.


apps/aesophia/src/aeso_data.erl
```diff
@@ -4,13 +4,332 @@
...
+%%      Same as binary_value() except maps are only stored as their identifier.
+%%      These are used as map values in the store and VM state and the contract
+%%      state data.
+%%
+%%  heap_value()
+%%      A binary_heap_value() (but with the pointer and the rest of the binary
+%%      split) together with the maps used by the value. Also allows an
+%%      arbitrary offset for the binary. These are used when copying values
+%%      inside the VM. The full VM state (heap and maps) together with a
+%%      pointer can treated as a heap value.
```
hanssv: s/can treated/can be treated

UlfNorell:
Protocol PR is up: https://github.com/aeternity/protocol/pull/224

I'll await more reviews and merge tonight if nothing comes up.

PR #1630, opened 2018-10-09
PT-161078655 Compute gas of each tx based on its size

jur0:
[PT-161078655](https://www.pivotaltracker.com/n/projects/2124891/stories/161078655)

```
tx_gas = tx_base_gas + byte_size * byte_gas
```
In case of contract txs, also gas for contract execution is added.

### DIFF
--- a/apps/aecore/test/aecore_sync_SUITE.erl
+++ b/apps/aecore/test/aecore_sync_SUITE.erl
@@ -511,6 +511,8 @@ large_msgs(Config) ->
     aecore_suite_utils:start_node(Dev1, Config),
     aecore_suite_utils:connect(aecore_suite_utils:node_name(Dev1)),
 
+    ok = rpc:call(N1, application, set_env, [aecore, block_gas_limit, 10000000]),
+
     %% Insert enough transactions to make a large generation
     Blob = fun(Size) -> << <<171:8>> || _ <- lists:seq(1, Size) >> end,
     {ok, Tx1} = add_spend_tx(N1, 10, 1, 1, 100, Blob(16#ffff)),
@@ -537,6 +539,9 @@ large_msgs(Config) ->
     T0 = os:timestamp(),
     aecore_suite_utils:start_node(Dev2, Config),
     aecore_suite_utils:connect(N2),
+
+    ok = rpc:call(N2, application, set_env, [aecore, block_gas_limit, 10000000]),
+
     true = expect_same(T0, Config).
 
 %% ==================================================
--- a/apps/aecore/test/aecore_sync_SUITE.erl
+++ b/apps/aecore/test/aecore_sync_SUITE.erl
@@ -163,7 +163,7 @@ end_per_group(one_blocked, Config) ->
     Config1 = config({devs, [dev1]}, Config),
     [Dev1 | _] = proplists:get_value(devs, Config1),
     EpochCfg = aecore_suite_utils:epoch_config(Dev1, Config),
-    aecore_suite_utils:create_config(Dev1, Config, 
+    aecore_suite_utils:create_config(Dev1, Config,
                                      maps:without([<<"blocked_peers">>], EpochCfg),
                                      [{add_peers, true}]);
 end_per_group(_, Config) ->
--- a/apps/aetx/src/aetx.erl
+++ b/apps/aetx/src/aetx.erl
@@ -147,7 +147,7 @@
 -callback for_client(Tx :: tx_instance()) ->
     map().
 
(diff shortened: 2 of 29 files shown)


apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -513,26 +513,26 @@ large_msgs(Config) ->
...
 
-    {ok, Tx5} = add_spend_tx(N1, 10, 1, 5, 100, Blob(16#fce3)), %% Should exactly fit in one message
-    aecore_suite_utils:mine_blocks_until_tx_on_chain(N1, Tx5, 10),
-
-    {ok, Tx6} = add_spend_tx(N1, 10, 1, 6, 100, Blob(16#fce4)), %% Wee bit too large
-    aecore_suite_utils:mine_blocks_until_tx_on_chain(N1, Tx6, 10),
-
-    {ok, Tx7} = add_spend_tx(N1, 10, 1, 7, 100, Blob(16#1fcb8)), %% Even multiple of fragment size
-    aecore_suite_utils:mine_blocks_until_tx_on_chain(N1, Tx7, 10),
+%    {ok, Tx5} = add_spend_tx(N1, 10, 1, 5, 100, Blob(16#fce3)), %% Should exactly fit in one message
```
jur0: @hanssv I commented these tests as the payload is too big and it doesn't fit into the gas limit per micro block.

hanssv: Those tests are there for a reason, I think you need to have some workaround to be able to run them in that case...

jur0: The workaround would be increasing the gas limit per micro block. @ThomasArts is working on a PR to make this value configurable so I can fix it after that.

hanssv: The workaround would be to have a higher limit for this test in particular I guess.

apps/aetx/src/aetx.erl
```diff
@@ -154,19 +157,31 @@
...
+%% txs can add additional gas, e.g. contract txs that need the gas for contract
+%% execution.
 -spec gas(Tx :: tx()) -> Gas :: non_neg_integer().
-gas(#aetx{ cb = CB, tx = Tx }) ->
-    CB:gas(Tx).
-
--spec gas_price(Tx :: tx()) -> GasPrice :: non_neg_integer().
-gas_price(#aetx{ cb = CB, tx = Tx }) ->
-    CB:gas_price(Tx).
+gas(#aetx{ type = Type, cb = CB, size = Size, tx = Tx }) when
```
hanssv: Each TX should have a base cost as well?! It is not only the size... An oracle Tx is more complex than a spend so it should be more costly?!

jur0: What base value do you suggest? I made it simple for now, there is no spec on how much more the oracle tx is more complex than the spend tx.

hanssv: No, those numbers don't exist yet... @ThomasArts is measuring... 

But the idea was that the cost of a TX was: TX cost + TX size * factor1 + State Tree Size * TTL * factor2

Though we don't have to fix all of that in this PR :-) But removing the `gas` callback from all TXs looks like a mistake?

hanssv: Just reinstate the 21000 or something a bit smaller and perhaps lower the 100 BYTE_COST a little bit and it is good enough here... And we can calibrate later.

apps/aetx/src/aetx.erl
```diff
@@ -154,19 +157,25 @@
...
 
 -spec fee(Tx :: tx()) -> Fee :: integer().
 fee(#aetx{ cb = CB, tx = Tx }) ->
     CB:fee(Tx).
 
 -spec gas(Tx :: tx()) -> Gas :: non_neg_integer().
-gas(#aetx{ cb = CB, tx = Tx }) ->
-    CB:gas(Tx).
+gas(#aetx{ cb = CB, size = Size, tx = Tx }) ->
+    CB:gas(Tx, Size).
```
hanssv: I would prefer: `CB:gas(Tx) + Size * aec_governance:byte_gas()`

sennui: @hanssv @jur0 I kind of liked it when I run into it. In the future we may distinguish "dummy size" and size affecting state trees. E.g. Spend payload are dead bytes, but all the hashes in tx fields require computation etc. So the only place having context about it, would be tx module.

PR #1631, opened 2018-10-09
Add basic meta data to the sophia bytecode

gorillainduction:
https://www.pivotaltracker.com/story/show/155763919

This PR aims at adding the general structure of meta data to sophia byte code. The actual meta data is not used for much, but it contains the source hash and compiler version in the byte code itself, which in turn is added in a signed transaction, so if a contract writer publishes the source code, another user can verify that the byte code is indeed generated from this source code.

The type information in the meta data is stubbed, but the intention is that it should be used to type check remote calls.


apps/aecontract/src/aect_evm.erl
```diff
@@ -83,11 +85,13 @@ execute_call(#{ code := Code
               , currentTimestamp := TS
               , chainState := ChainState
               , chainAPI := ChainAPI
+              , type_info := TypeInfo
               , vm_version := VmVersion
               }, Trace) ->
     %% TODO: Handle Contract In State.
     Spec =
         #{ exec => #{ code => Code
+                    , type_info => TypeInfo
```
UlfNorell: I'm not sure we need the type info in the VM state. Once we get to running VM code the types should already be checked, no?

gorillainduction: Hmm, that is probably correct. The type info should be checked when the byte code of the next contract is fetched, so we don't need it here. I can amend this.

apps/aesophia/src/aeso_compiler.erl
```diff
@@ -59,6 +85,41 @@ to_bytecode([Op|Rest], Options) ->
 to_bytecode([], _) -> [].
 
 
+serialize(ByteCode, ContractString) ->
+    ContractBin = list_to_binary(ContractString),
+    Version = version(),
+    Fields = [ {source_hash, aec_hash:hash(sophia_source_code, ContractBin)}
+             , {type_info, []} %% TODO: This
```
hanssv: :-)

PR #1632, opened 2018-10-09
Pt 159426861 jsonrpc in sc ws

uwiger:
First step in adapting State Channels WebSocket API to JSON-RPC.
The (legacy) test suite passes, though the code is mostly converted to expect JSON-RPC.
The actual JSON-RPC interaction has not yet been tested. This will be the next step.

### DIFF
--- a/apps/aehttp/src/sc_ws_handler.erl
+++ b/apps/aehttp/src/sc_ws_handler.erl
@@ -77,10 +77,8 @@ websocket_handle({text, MsgBin}, #handler{protocol = P} = H) ->
     try unpack_request(P, jsx:decode(MsgBin, [return_maps]), H) of
         {Msg, H1} ->
             try process_incoming(Msg, H1) of
-                no_reply -> {ok, H};
-                {reply, Resp} ->
-                    {reply, {text, jsx:encode(Resp)}, H};
-                {error, _} -> {ok, H}
+                no_reply      -> {ok, H};
+                {reply, Resp} -> {reply, {text, jsx:encode(Resp)}, H}
             catch
                 error:E1 ->
                     lager:debug("CAUGHT E1=~p / ~p", [E1, erlang:get_stacktrace()]),
@@ -754,18 +752,24 @@ add_payload(#{payload := Payload}, Msg) -> Msg#{<<"payload">> => Payload};
 add_payload(#{action := _}       , Msg) -> Msg;
 add_payload(Reply                , Msg) -> Msg#{<<"payload">> => Reply}.
 
-json_rpc_reply(#{payload := Payload} = Reply, #handler{orig_request = Req}) ->
+json_rpc_reply(Reply, #handler{orig_request = Req}) ->
     case Req of
         #{id := Id} ->
             {reply, #{ <<"jsonrpc">> => <<"2.0">>
                      , <<"id">>      => Id
-                     , <<"result">>  => Payload }
+                     , <<"result">>  => result(Reply) }
             };
         _ ->
             {reply, #{ <<"jsonrpc">> => <<"2.0">>
                      , <<"method">>  => legacy_to_method_out(Reply)
-                     , <<"params">>  => Payload }}
+                     , <<"params">>  => result(Reply) }}
     end.
+
+result(#{payload := Payload}) ->            
+    Payload;
+result(Result) -> 
+    Result.
+
 
 legacy_to_method_in(Action) ->
     <<"channels.", Action/binary>>.
--- a/apps/aehttp/src/sc_ws_handler.erl
+++ b/apps/aehttp/src/sc_ws_handler.erl
@@ -16,38 +16,43 @@
                   channel_id         :: aesc_channels:id() | undefined,
                   enc_channel_id     :: aec_base58c:encoded() | undefined,
                   job_id             :: term(),
+                  protocol           :: legacy | jsonrpc | undefined,
+                  orig_request       :: map() | undefined,
                   role               :: initiator | responder | undefined,
                   host               :: binary() | undefined,
                   port               :: non_neg_integer() | undefined}).
 
 -opaque handler() :: #handler{}.
 -export_type([handler/0]).
--define(ACTION_SIGNED(Action), Action =:= <<"initiator_sign">>;
-                               Action =:= <<"deposit_tx">>;
-                               Action =:= <<"deposit_ack">>;
-                               Action =:= <<"withdraw_tx">>;
-                               Action =:= <<"withdraw_ack">>;
-                               Action =:= <<"responder_sign">>;
-                               Action =:= <<"shutdown_sign">>;
-                               Action =:= <<"shutdown_sign_ack">>;
-                               Action =:= <<"update">>;
-                               Action =:= <<"update_ack">>).
--define(ACTION_TAG(Action), case Action of
-                                <<"initiator_sign">> -> create_tx;
-                                <<"deposit_tx">> -> deposit_tx;
-                                <<"deposit_ack">> -> deposit_created;
-                                <<"withdraw_tx">> -> withdraw_tx;
-                                <<"withdraw_ack">> -> withdraw_created;
-                                <<"responder_sign">> -> funding_created;
-                                <<"update">> -> update;
-                                <<"update_ack">> -> update_ack;
-                                <<"shutdown_sign">> -> shutdown;
-                                <<"shutdown_sign_ack">> -> shutdown_ack;
-                                <<"leave">> -> leave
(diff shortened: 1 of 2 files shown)


apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -16,38 +16,43 @@
...
+                                <<"channels.shutdown_sign">>     -> shutdown;
+                                <<"channels.shutdown_sign_ack">> -> shutdown_ack;
+                                <<"channels.leave">>             -> leave
                             end).
 
 init(Req, _Opts) ->
-    {cowboy_websocket, Req, maps:from_list(cowboy_req:parse_qs(Req))}.
+    lager:debug("init(~p, ~p)", [Req, _Opts]),
+    {cowboy_websocket, Req,
+     maps:merge(#{protocol => <<"legacy">>},
```
velzevur: So the `legacy` is the default protocol

uwiger: It is, until I've added tests for json-rpc. Presumably, the test suite is the only legacy client of note.

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -59,21 +59,28 @@ start_link_channel(Host, Port, RoleA, Opts) when is_atom(RoleA) ->
...
+    unlink(Pid),
     case wait_for_connect(Pid) of
         {ok, Pid} = Res ->
             set_role(Pid, RoleA),
             Res;
         {error, _} = Err -> Err
     end.
 
+extra_headers() ->
+    [{extra_headers, [{<<"Content-Type">>, <<"application/jsonrpc">>}]}].
```
zp-sd: Isn't that `application/json-rpc`?

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -559,7 +578,17 @@ prepare_handler(Params) ->
         fun(_, {error, _} = Err) -> Err;
             (Fun, Accum) -> Fun(Accum)
         end,
-        #handler{}, Validators).
+        #handler{protocol = protocol(Params)}, Validators).
+
+protocol(#{protocol := P}) ->
+    case P of
+        <<"legacy">>  -> legacy;
+        <<"jsonrpc">> -> jsonrpc;
```
zp-sd: No harm, but binary `protocol` should never happen?

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -618,13 +647,24 @@ read_channel_options(Params) ->
...
-                         <<"reason">> => Reason}}.
+error_response(Reason, #handler{protocol = legacy, orig_request = Req}) ->
+    {reply, #{ <<"action">>  => <<"error">>
+             , <<"payload">> => #{ <<"request">> => Req
+                                 , <<"reason">> => Reason} }
+    };
+error_response(Reason, #handler{protocol = jsonrpc, orig_request = Req}) ->
+    case Req of
+        #{<<"id">> := Id} ->
+            {reply, #{ <<"jsonrpc">> => <<"2.0">>
```
zp-sd: 👍

apps/aehttp/src/sc_ws_handler.erl
uwiger: perhaps...

apps/aehttp/test/aehttp_ws_test_utils.erl
uwiger: Not sure where I got that from. [Looking it up](https://www.jsonrpc.org/historical/json-rpc-over-http.html):
> Content-Type SHOULD be 'application/json-rpc' but MAY be 'application/json' or 'application/jsonrequest'

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -16,38 +16,43 @@
                   channel_id         :: aesc_channels:id() | undefined,
                   enc_channel_id     :: aec_base58c:encoded() | undefined,
                   job_id             :: term(),
+                  protocol           :: legacy | jsonrpc | undefined,
```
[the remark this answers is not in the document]
uwiger: That was mainly to avoid a dialyzer warning, but better then to give the attribute a default value. Fixed.

apps/aehttp/test/aehttp_ws_test_utils.erl
```diff
@@ -316,9 +323,9 @@ websocket_handle({text, MsgBin}, _ConnState, #state{regs=Register, role=Role}=St
         _ ->
             ct:log("[~p] Received msg ~p~n", [Role, Msg])
     end,
-    Origin = binary_to_existing_atom(maps:get(<<"origin">>, Msg,
-                                              atom_to_binary(?CHANNEL, utf8)), utf8),
-    Action = binary_to_existing_atom(maps:get(<<"action">>, Msg), utf8),
+    Origin = binary_to_atom(maps:get(<<"origin">>, Msg,
```
lucafavatella: I see this is test only so no concern for dynamic creation of atoms.

uwiger: Correct. The change was made after a test case failure, and I thought it was better to not have to explicitly create atoms in the test suite just to ensure that it didn't crash on `binary_to_existing_atom/2`.

apps/aehttp/test/aehttp_ws_test_utils.erl
lucafavatella: > Not sure where I got that from. [Looking it up](https://www.jsonrpc.org/historical/json-rpc-over-http.html):

I read in a section of the code of this PR `<<"jsonrpc">> => <<"2.0">>` so is the spec followed the 2.0 one at https://www.jsonrpc.org/specification ? I see no content type / media type there. And, more specifically, I see no [media type registered with IANA](https://www.iana.org/assignments/media-types/media-types.xhtml).

zp-sd: @uwiger Yes, I saw that too, but still `-` is missing, and these specs are for JSON RPC 1.2 ? 🤔

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -68,42 +73,66 @@ websocket_init(Params) ->
...
+websocket_handle({text, MsgBin}, #handler{protocol = P} = H) ->
+    try unpack_request(P, jsx:decode(MsgBin, [return_maps]), H) of
+        {Msg, H1} ->
+            HNext = H1#handler{orig_request = undefined},
+            try process_incoming(Msg, H1) of
+                no_reply      -> {ok, HNext};
+                {reply, Resp} -> {reply, {text, jsx:encode(Resp)}, HNext}
+            catch
+                error:E1 ->
+                    lager:debug("CAUGHT E1=~p / ~p", [E1, erlang:get_stacktrace()]),
```
lucafavatella: I see [this stacktrace retrieval not evaluated if debug logs not set](https://github.com/aeternity/lager/blob/3.6.0.ae1/src/lager_transform.erl#L220) so ok.

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -16,38 +16,43 @@
...
+                                <<"channels.shutdown_sign">>     -> shutdown;
+                                <<"channels.shutdown_sign_ack">> -> shutdown_ack;
+                                <<"channels.leave">>             -> leave
                             end).
 
 init(Req, _Opts) ->
-    {cowboy_websocket, Req, maps:from_list(cowboy_req:parse_qs(Req))}.
+    lager:debug("init(~p, ~p)", [Req, _Opts]),
+    {cowboy_websocket, Req,
+     maps:merge(#{protocol => <<"legacy">>},
```
lucafavatella: [cowboy_req:parse_qs returns keys as binary](https://ninenines.eu/docs/en/cowboy/2.2/manual/cowboy_req.parse_qs/), right? How can this protocol atom act as default then? It means that point in code below (only protocol/1 function?) that uses protocol atom as key in map always fetches this item rather than anything from the param.

lucafavatella:
I took the liberty of labelling as WIP for pending clarifications.

uwiger:
@lucafavatella the protocol parsing is not necessary yet, since it defaults to `"legacy"` (weak excuse for possible bugs, I know). :)  But I'll take a look at it.

The `aest_channels` test suite passes.

apps/aehttp/src/sc_ws_handler.erl
lucafavatella: I see answered in https://github.com/aeternity/epoch/pull/1632#issuecomment-428554067

lucafavatella:
I should have squashed - my bad.

PR #1637, opened 2018-10-10
Use pool_peek also for the last peek in txs_gc-test

hanssv:
Fix intermittent test failure (race between GC and mempool peek).

### DIFF
--- a/apps/aecore/test/aecore_txs_SUITE.erl
+++ b/apps/aecore/test/aecore_txs_SUITE.erl
@@ -150,7 +150,7 @@ txs_gc(Config) ->
     aecore_suite_utils:mine_key_blocks(N1, 2),
 
     %% Now there should be no transactions in mempool
-    {ok, Txs6} = rpc:call(N1, aec_tx_pool, peek, [infinity]),
+    {ok, Txs6} = pool_peek(N1),
     {0, _} = {length(Txs6), Txs6},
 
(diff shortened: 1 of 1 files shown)


hanssv:
https://www.pivotaltracker.com/story/show/161084205

apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -150,7 +150,7 @@ txs_gc(Config) ->
     aecore_suite_utils:mine_key_blocks(N1, 2),
 
     %% Now there should be no transactions in mempool
-    {ok, Txs6} = rpc:call(N1, aec_tx_pool, peek, [infinity]),
+    {ok, Txs6} = pool_peek(N1),
```
lucafavatella: I understand there are:
* The RPC call aec_conductor:stop_mining in mine_blocks that casts garbage_collect to aec_tx_pool that spawns (let us call this process X) that sends new_gc_height back to aec_tx_pool that casts garbage_collect to aec_tx_pool_gc.
* The test calling peek here.

This change adds a synchronous call to aec_tx_pool_gc, before peeking aec_tx_pool.

It looks to me that test could still fail if e.g.:
* process X is slow; or
* aec_tx_pool_gc is slow in performing the actual GC.

So it looks like it would be more robust and readable to add a wait-until with a short time and quick polling here?

Though an improvement so ok.

PR #1639, opened 2018-10-11
change sc ws listen addr to 127.0.0.1

uwiger:
Listen address default was previously `0.0.0.0`. For security reasons, it should default to localhost.

See [PT #161144513](https://www.pivotaltracker.com/story/show/161144513)


apps/aehttp/src/aehttp_app.erl
```diff
@@ -18,7 +18,7 @@
 -define(INT_ACCEPTORS_POOLSIZE, 10).
 
 -define(DEFAULT_CHANNEL_WEBSOCKET_PORT, 8044).
--define(DEFAULT_CHANNEL_WEBSOCKET_LISTEN_ADDRESS, <<"0.0.0.0">>).
+-define(DEFAULT_CHANNEL_WEBSOCKET_LISTEN_ADDRESS, <<"127.0.0.1">>).
```
lucafavatella: Do any of the sys.config require update?

lucafavatella: At least this seems needing change:

https://github.com/aeternity/epoch/blob/0204d168dd2ae992bc0b88e0b3a03cdbe4fcaba3/config/sys.config#L20-L23

PR #1643, opened 2018-10-12
Add response TTL to oracle response TX

hanssv:
Fixing https://www.pivotaltracker.com/story/show/161048676


apps/aevm/src/aevm_ae_primops.erl
```diff
@@ -236,7 +236,7 @@ oracle_call_respond(_Value, Data, State) ->
                                     ArgumentTypes = [word, word, sign_t(), RType],
                                     [_, _, Sign0, R] = get_args(ArgumentTypes, Data),
                                     Callback = fun(API, ChainState) ->
-                                        API:oracle_respond(OracleKey, QueryKey, to_sign(Sign0), R, ChainState) end,
+                                        API:oracle_respond(OracleKey, QueryKey, to_sign(Sign0), R, DeltaRTTL, ChainState) end,
```
lucafavatella: I would expect the call to `API:oracle_query_response_tt` to be removed as unnecessary, and this parameter used instead.

hanssv: I can't parse this comment, sorry?! 

We get `DeltaRTTL` from `oracle_query_response_ttl`? What is the suggested alternative?

lucafavatella: Commit 590a3ad4 introduced extra lookup in chain calling `API:oracle_query_response_ttl` - just for retrieving response TTL from query. Now that response TTL is among args I expect that the response TTL from the args is used, and that `API:oracle_query_response_ttl` lookup removed.

lucafavatella: And I expect aeo_response_tx (in check callback) to check that the response TTL in the tx is the same as the response TTL in the query.

lucafavatella: I had missed that the element is not passed as arg of the Sophia builtin. Apologies.

PR #1649, opened 2018-10-14
Pt 160280865 - SC optimize poi

velzevur:
PT [Force progress tx optimization - PoI](https://www.pivotaltracker.com/story/show/160280865)

Remove `addresses` and replace the `poi` with `offchain_trees` being a MPT


velzevur:
Rebased on top of `master`

apps/aecore/src/aec_accounts_trees.erl
```diff
@@ -53,6 +59,16 @@ empty_with_backend() ->
 new_with_backend(Hash) ->
     aeu_mtrees:new_with_backend(Hash, aec_db_backends:accounts_backend()).
 
+-spec total_balances(tree()) -> integer().
+total_balances(Tree) ->
```
gorillainduction: This should use an iterator, and it is probably better to send in a limit to short circuit the traversal. This is an excellent attack vector otherwise.

apps/aens/src/aens_state_tree.erl
```diff
@@ -171,6 +176,32 @@ name_list(#ns_tree{mtree = Tree}) ->
...
+    aec_object_serialization:serialize(
+        nameservice_mtree,
+        ?VSN,
+        serialization_template(?VSN),
+        [ {mtree, MTBin}
+        , {cache, CacheBin}]).
+
+-spec from_binary_without_backend(binary()) -> tree().
+from_binary_without_backend(Bin) ->
+    [ {mtree, MTBin}
```
gorillainduction: The cache should not be in the serialization since it is not under consensus (not reflected in the root hash). The cache is nowhere defined, and it is just an implementation detail to handle future events.

apps/aeoracle/src/aeo_state_tree.erl
```diff
@@ -403,3 +408,30 @@ cache_pop(C) ->
...
+        aec_object_serialization:deserialize(oracles_mtree, ?VSN,
+                                             serialization_template(?VSN), Bin),
+    OTree = aeu_mtrees:deserialize(OTBin),
+    Cache = aeu_mtrees:deserialize(CacheBin),
+    #oracle_tree{otree = OTree,
+                 cache = Cache}.
+
+serialization_template(?VSN) ->
+    [ {otree, binary}
+    , {cache, binary}].
```
gorillainduction: The same thing here. The cache is not under consensus.

apps/aeutils/src/aeu_mtrees.erl
```diff
@@ -252,3 +259,60 @@ proof_db_commit(_Cache,_DB) ->
...
+                    [ {key, Key}
+                    , {val, Value}
+                    ])
+            end,
+            to_list(Tree)),
+    aec_object_serialization:serialize(
+        mtree,
+        ?VSN,
+        serialization_template(?VSN),
+        [ {leaves, LeavesBins}]).
```
gorillainduction: The values are not only leaves. Values can also be in branches.

apps/aeutils/src/aeu_mtrees.erl
```diff
@@ -252,3 +259,60 @@ proof_db_commit(_Cache,_DB) ->
...
+            fun({Key, Value}) ->
+                aec_object_serialization:serialize(
+                    mtree_leaf,
+                    ?VSN,
+                    leaf_serialization_template(?VSN),
+                    [ {key, Key}
+                    , {val, Value}
+                    ])
+            end,
+            to_list(Tree)),
```
gorillainduction: This should be done with an iterator instead. We don't want to construct the intermediate list.

apps/aechannel/src/aesc_force_progress_tx.erl
```diff
@@ -132,93 +122,85 @@ from_pubkey(#channel_force_progress_tx{from_id = FromId}) ->
...
      , {state_hash    , StateHash}
-     , {addresses     , Addresses}
-     , {poi           , aec_trees:serialize_poi(PoI)}
+     , {offchain_trees, aec_trees:serialize_to_binary(OffChainTrees)}
      , {ttl           , TTL}
      , {fee           , Fee}
      , {nonce         , Nonce}
      ]}.
 
 -spec deserialize(vsn(), list()) -> tx().
```
lucafavatella: I expect update to:
* https://github.com/aeternity/protocol/blob/2d3246e9bed457f4399e4dde1b3e29ee3af11914/serializations.md#channel-solo-force-progress-transaction
* https://github.com/aeternity/protocol/blob/785cf1efa4e387c2a82c5ad8cb773d91fffd0106/channels/ON-CHAIN.md#channel_force_progress_tx

velzevur: Yes, protocol changes are coming

velzevur:
Protocol PR aeternity/protocol#241 makes some changes in related files so the protocol PR will be created after 241 is merged. In order not to forget this, I've created a PT [document new Force Progress behavior](https://www.pivotaltracker.com/story/show/161336288)

PR #1653, opened 2018-10-15
Use record tag as kind for tx to derive governed gas cost

ThomasArts:


apps/aecore/src/aec_governance.erl
```diff
@@ -92,9 +92,31 @@ block_mine_reward() ->
 block_gas_limit() ->
     application:get_env(aecore, block_gas_limit, ?BLOCK_GAS_LIMIT).
 
-tx_base_gas() ->
-    ?TX_BASE_GAS.
-
+tx_base_gas(contract_call_tx) -> 5 * ?TX_BASE_GAS;
```
jur0: There is also `primop_base_gas_cost/1`, the base gas could be the same for the primops and transactions. What do you think @lucafavatella?

ThomasArts: The issue is that running a spend_tx inside a contract takes much more time then running it on chain. We need to take out these cost and @happi and I discussed doing that in the cost of the kind of transaction.

So, you pay more for a contract call, but if you do 1000 spend_tx in that call, you don't pay as much as 1000 spend_tx on chain for each primitive spend in the contract.

ThomasArts: In other words, the on chain and in contract operations need to be charged differently

PR #1659, opened 2018-10-16
Keep CP in argument for more efficient execution.

happi:
Tried several different speedups of the VM (module merging with igor, inling, and native code compilation) but just keeping the PC/CP in a register/argument gave most bang for the buck.

### DIFF
--- a/apps/aevm/src/aevm_eeevm.erl
+++ b/apps/aevm/src/aevm_eeevm.erl
@@ -66,7 +66,7 @@ eval_code(State = #{ address := 0 }) ->
     Trace("  PrimCall ~p\n", [TxType]),
     {ok, State};
 eval_code(State) ->
-    try {ok, loop(valid_jumpdests(State))}
+    try {ok, loop(aevm_eeevm_state:cp(State), valid_jumpdests(State))}
     catch
         throw:?aevm_eval_error(What, StateOut) ->
             ?TEST_LOG("Code evaluation error at ~s",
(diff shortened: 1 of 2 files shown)


apps/aevm/src/aevm_eeevm.erl
```diff
@@ -1164,7 +1136,7 @@ loop(StateIn) ->
                     %% almost entirely identical.
                     {Res, State1} = recursive_call(State0, OP),
                     State2 = push(Res, State1),
-                    next_instruction(OP, State, State2);
+                    next_instruction(CP, State, State2);
                 16#f5 -> eval_error({illegal_instruction, OP}, State0);
```
lucafavatella: In eval_error cases, it seems the CP in the state will not be update hence [test log](https://github.com/aeternity/epoch/blob/297b095df5b35548b8bec4135aed2903136a38ce/apps/aevm/src/aevm_eeevm.erl#L73) may be misleading. Though it seems only a test issue.

apps/aevm/src/aevm_eeevm.erl
```diff
@@ -778,202 +782,170 @@ loop(StateIn) ->
...
                     %% significant place in big endian).
-                    State1 = push_n_bytes_from_cp(OP-?PUSH1+1, State0),
-                    next_instruction(OP, State, State1);
+                    push_n_bytes_from_cp(CP, 1, State, State0);
                 ?PUSH2 ->
                     %% 0x61 PUSH1 δ=0 α=1
                     %% Place 2 byte item on stack.
-                    State1 = push_n_bytes_from_cp(OP-?PUSH1+1, State0),
-                    next_instruction(OP, State, State1);
+                    push_n_bytes_from_cp(CP, 2, State, State0);
```
lucafavatella: No problem though just asking: for PUSH5 onwards I see the derived computation (OP - ThisOp + 1) kept. Is it intended to remove derived computation for PUSH1...PUSH4?

happi: Yes. Push 1-4 and 32 are most used, but 32 is expensive anyway so I think it gives a bit of a speedup to do one calculation less for push 1 to 4. Future version will make all of them more efficient.

PR #1664, opened 2018-10-17
Fix channel system tests

sylane:
Could someone else run the system test to confirm it ?
`SUITE=aest_channels make system-test`


lucafavatella:
I see 3 approvals. CI passed.

PR #1666, opened 2018-10-17
Add some test explicitly testing state handling

hanssv:
Test that a called contract can read and update its own state.
Test that you can pass your state as an argument to another contract. (This is possibly very expensive - good implementation might be needed.)
Test that another contract can update the state and return it.
Test that the callers state is not changed by an updated state in a called contract. (This should not be possible with current implementation but we want a test to make sure future implementations doesnt change this.)
Test that the caller can receive an updated version of its state and choose to update its own state

+ testing remote call inside initialization call of contract create transaction

### DIFF
--- a/apps/aecontract/test/aecontract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_SUITE.erl
@@ -73,6 +73,7 @@
         , sophia_savecoinbase/1
         , sophia_fundme/1
         , sophia_aens/1
+        , sophia_state_handling/1
         , create_store/1
         , update_store/1
         , read_store/1
@@ -138,7 +139,8 @@ groups() ->
                                  sophia_chain,
                                  sophia_savecoinbase,
                                  sophia_fundme,
-                                 sophia_aens ]}
+                                 sophia_aens,
+                                 sophia_state_handling]}
     , {sophia_oracles_ttl, [],
           %% Test Oracle TTL handling
         [ sophia_oracles_ttl__extend_after_expiry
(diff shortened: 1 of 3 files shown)


apps/aesophia/test/contracts/remote_state.aes
```diff
@@ -0,0 +1,12 @@
...
+  record rstate = { i : int, s : string, m : map(int, int) }
+
+  function get(s : rstate) = s
+  function get_i(s : rstate) = s.i
+  function get_s(s : rstate) = s.s
+  function get_m(s : rstate) = s.m
+
+  function fun_update_i(s : rstate, ni) = s{ i = ni }
+  function fun_update_s(s : rstate, ns) = s{ s = ns }
+  function fun_update_m(s : rstate, nm) = s{ m = nm }
```
lucafavatella: A potential additional test case is updating an element of the nested map. My concern is that references to map are handled ok but not the map content.

apps/aesophia/test/contracts/remote_state.aes
```diff
@@ -0,0 +1,12 @@
+contract RemoteState =
+  record rstate = { i : int, s : string, m : map(int, int) }
```
lucafavatella: Is it possible to declare the remote state type as a type in the remote contract, i.e. something like:
```
contract StateHandling =
  record state = { i : int, s : string, m : map(int, int) }

contract RemoteState =
  function get(s : StateHandling.state) = s
```

Though totally optional of course.

apps/aesophia/test/contracts/remote_state.aes
lucafavatella: I mean this as a specialized case of:
> Test that another contract can update the state and return it.

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -2974,6 +2976,73 @@ sophia_aens(_Cfg) ->
...
+    ?assertEqual({2, <<"undefined">>, #{}}, ?call(call_contract, Acc, Ct2, read, StateT, {})),
+    ?assertEqual(2,                         ?call(call_contract, Acc, Ct2, read_i, word, {})),
+    ?assertEqual(<<"undefined">>,           ?call(call_contract, Acc, Ct2, read_s, string, {})),
+    ?assertEqual(#{},                       ?call(call_contract, Acc, Ct2, read_m, MapT, {})),
+
+    %% Test that we can update the state
+    ?call(call_contract, Acc, Ct1, update_i, UnitT, {7}),
+    ?call(call_contract, Acc, Ct1, update_s, UnitT, {<<"defined">>}),
+    ?call(call_contract, Acc, Ct1, update_m, UnitT, {#{3 => 4, 1 => 2}}),
+    ?assertMatch({7, <<"defined">>, #{3 := 4, 1 := 2}},
```
lucafavatella: Why match and not equal? Also below. This may not check if any map keys added.

```
$ erl
...
2> #{1 := 2} = #{2 => 3, 1 => 2}.
#{1 => 2,2 => 3}
```

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -138,7 +139,8 @@ groups() ->
                                  sophia_chain,
                                  sophia_savecoinbase,
                                  sophia_fundme,
-                                 sophia_aens ]}
+                                 sophia_aens,
+                                 sophia_state_handling]}
```
lucafavatella: Potential additional test cases:
* Test that remote call passing larger state to function returning constant return value (e.g. empty tuple - for removing one variable to test case) increases gas cost. The idea is preventing malicious user to obtain somehow large contract state (potentially paid by other users for legitimate use) then send cheap call tx calling function performing remote call passing large state without paying for cost of passing large state. This would actually be a group of tests - depending on how the state is "larger":
  * State as record with map with more-or-less keys. This is for testing against ["the size of the map does not contribute to the gas costs of a contract call reading or updating it"](https://github.com/aeternity/protocol/blob/2d3246e9bed457f4399e4dde1b3e29ee3af11914/contracts/sophia.md#map-implementation).
  * State as record with map with a key with larger-or-smaller value.
  * State as record with more fields.
  * State as record with one larger-or-smaller non-map field.
* Test that remote call passing constant state and returning larger updated state increases gas cost. The idea is similar to above. Sample cases:
  * State passed is record with a field that is a map. The called contract updates that field with a larger-or-smaller map in its own state (idea is that that may be implemented as assignment of a reference hence very cheap) and returns such updated (larger-or-smaller) state.

These ^^^ are related to gas cost - mainly aimed at clarifying [this comment re gas cost](https://www.pivotaltracker.com/story/show/160259586/comments/194951641), and marginally this item in the description of the ticket:
> Test that you can pass your state as an argument to another contract. (This is possibly very expensive - good implementation might be needed.)

Not sure if ["To be determined: gas costs of `CALL` based on size of calldata and return value"](https://github.com/aeternity/protocol/blame/2d3246e9bed457f4399e4dde1b3e29ee3af11914/contracts/sophia.md#L1059) is tracked anyware as a ticket.

apps/aecontract/test/aecontract_SUITE.erl
hanssv: No reason, started with Equal, and then my brain changed to Match after the first three or so?!

apps/aecontract/test/aecontract_SUITE.erl
hanssv: The first bullet point is straightforward to test for, I prepared a couple of tests in `contracts_SUITE:sophia_state_gas/1` but disabled the _asserts_ since this does not seem to be implemented yet?! So I think we need to make a ticket and implement this (unless already covered by an existing ticket? @happi ? 

Second part is harder, since the code for making an update to the passed state before returning would also cost gas? But it should be possible to code up something that can be used to compare.

apps/aesophia/test/contracts/remote_state.aes
hanssv: Nope, I can't figure out a way, but that doesn't mean it isn't possible, I didn't write the compiler :-)

apps/aesophia/test/contracts/remote_state.aes
hanssv: Added...

hanssv:
Addressed comments, and prepared for testing gas cost of argument size (although that isn't necessarily tied to the state).

apps/aesophia/test/contracts/remote_state.aes
lucafavatella: Tracked as https://www.pivotaltracker.com/story/show/161309782

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -3001,48 +3003,82 @@ sophia_state_handling(_Cfg) ->
...
                  ?call(call_contract, Acc, Ct1, read, StateT, {})),
     ?call(call_contract, Acc, Ct1, remote_update_s, UnitT, {Ct2, <<"foo">>}),
-    ?assertMatch({9, <<"foo">>, #{3 := 5}},
+    ?assertEqual({9, <<"foo">>, #{3 => 5}},
                  ?call(call_contract, Acc, Ct1, read, StateT, {})),
     ?call(call_contract, Acc, Ct1, remote_update_m, UnitT, {Ct2, #{2 => 5}}),
-    ?assertMatch({9, <<"foo">>, #{2 := 5}},
+    ?assertEqual({9, <<"foo">>, #{2 => 5}},
+                 ?call(call_contract, Acc, Ct1, read, StateT, {})),
+    ?call(call_contract, Acc, Ct1, remote_update_mk, UnitT, {Ct2, 3, 7}),
```
lucafavatella: I see added. Thanks.

PR #1668, opened 2018-10-18
Instrument and optimize re. overload tests

uwiger:
See [PT #161312630](PT-161312630-instrument-for-overload)

### DIFF
--- a/apps/aehttp/priv/swagger.json
+++ b/apps/aehttp/priv/swagger.json
@@ -195,6 +195,12 @@
             "schema" : {
               "$ref" : "#/definitions/Error"
             }
+          },
+          "503" : {
+            "description" : "Resource not available",
+            "schema" : {
+              "$ref" : "#/definitions/Error"
+            }
           }
         }
       }
@@ -225,6 +231,12 @@
             "schema" : {
               "$ref" : "#/definitions/Error"
             }
+          },
+          "503" : {
+            "description" : "Resource not available",
+            "schema" : {
+              "$ref" : "#/definitions/Error"
+            }
           }
         }
       }
@@ -251,6 +263,12 @@
           },
           "400" : {
             "description" : "Invalid block",
+            "schema" : {
+              "$ref" : "#/definitions/Error"
+            }
+          },
+          "503" : {
+            "description" : "Resource not available",
             "schema" : {
               "$ref" : "#/definitions/Error"
             }
@@ -289,6 +307,12 @@
             "schema" : {
               "$ref" : "#/definitions/Error"
             }
+          },
+          "503" : {
+            "description" : "Resource not available",
+            "schema" : {
+              "$ref" : "#/definitions/Error"
+            }
           }
         }
       }
@@ -321,6 +345,12 @@
           },
           "404" : {
             "description" : "Block not found",
+            "schema" : {
+              "$ref" : "#/definitions/Error"
+            }
+          },
+          "503" : {
+            "description" : "Resource not available",
             "schema" : {
               "$ref" : "#/definitions/Error"
             }
@@ -366,6 +396,12 @@
             "schema" : {
               "$ref" : "#/definitions/Error"
             }
+          },
+          "503" : {
+            "description" : "Resource not available",
+            "schema" : {
+              "$ref" : "#/definitions/Error"
+            }
           }
         }
       }
@@ -398,6 +434,12 @@
           },
           "404" : {
             "description" : "Block not found",
+            "schema" : {
+              "$ref" : "#/definitions/Error"
+            }
+          },
+          "503" : {
+            "description" : "Resource not available",
             "schema" : {
               "$ref" : "#/definitions/Error"
             }
@@ -459,6 +501,12 @@
             "schema" : {
               "$ref" : "#/definitions/Error"
             }
+          },
+          "503" : {
+            "description" : "Resource not available",
+            "schema" : {
+              "$ref" : "#/definitions/Error"
+            }
           }
         }
       }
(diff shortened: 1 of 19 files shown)


apps/aehttp/src/aehttp_app.erl
```diff
@@ -15,11 +15,11 @@
 
 -define(DEFAULT_WEBSOCKET_INTERNAL_PORT, 8144).
 -define(DEFAULT_WEBSOCKET_LISTEN_ADDRESS, <<"127.0.0.1">>).
--define(INT_ACCEPTORS_POOLSIZE, 10).
+-define(INT_ACCEPTORS_POOLSIZE, 100).
```
eproxus: This should be renamed to `MAX_CONNECTIONS` instead, since the acceptor pool size is this number divided by ten:
```suggestion
-define(MAX_CONNECTIONS, 100).
```

apps/aehttp/src/aehttp_app.erl
```diff
@@ -15,11 +15,11 @@
...
 
 -define(DEFAULT_CHANNEL_WEBSOCKET_PORT, 8044).
 -define(DEFAULT_CHANNEL_WEBSOCKET_LISTEN_ADDRESS, <<"127.0.0.1">>).
--define(CHANNEL_ACCEPTORS_POOLSIZE, 10).
+-define(CHANNEL_ACCEPTORS_POOLSIZE, 100).
```
eproxus:
```suggestion
-define(CHANNEL_MAX_CONNECTIONS, 100).
```

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -49,7 +49,12 @@ handle_request(OperationID, Req, Context) ->
         end).
 
 %% run(no_queue, F) -> F();
-run(Queue, F) -> jobs:run(Queue, F).
+run(Queue, F) ->
+    try aec_jobs_queues:run(Queue, F)
+    catch
+        error:{rejected, _} ->
```
ThomasArts: What if we fail for a different reason?

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -32,8 +32,12 @@
                    ) -> {Status :: cowboy:http_status(), Headers :: list(), Body :: map()}.
 
 handle_request(OperationID, Req, Context) ->
-    jobs:run(http_update,
-             fun() -> handle_request_(OperationID, Req, Context) end).
+    try aec_jobs_queues:run(http_update,
+                            fun() -> handle_request_(OperationID, Req, Context) end)
+    catch
+        error:{rejected, _} ->
+            {503, [], #{reason => <<"Temporary overload">>}}
```
ThomasArts: validation of this will fail in aechttp_api_handler line 56 making it throw an exception instead of replying the request

ThomasArts: proposal is to not validate when we return a 500 code

uwiger: Do you suggest that I should remove the 503s from swagger?

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -45,11 +47,16 @@
 handle_request(OperationID, Req, Context) ->
     run(queue(OperationID),
         fun() ->
-                handle_request_(OperationID, Req, Context)
+                ?TC(handle_request_(OperationID, Req, Context), Req)
```
ThomasArts: Do we want to keep timer:tc in the production code?

apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -596,6 +620,16 @@
                         "max_time" : {
                             "type" : "integer",
                             "default" : 0
+                        },
+                        "rate_modifiers" : {
+                            "type" : "string",
+                            "default" : "mempool.size: 10",
```
ThomasArts: over engineering, but fine.

config/swagger.yaml
```diff
@@ -177,6 +177,10 @@ paths:
           description: 'Block not found'
           schema:
             $ref: '#/definitions/Error'
+        '503':
```
ThomasArts: Could we get those out of the spec? This is really ugly.

apps/aeutils/priv/epoch_config_schema.json
uwiger: I guess we'll see... ;-)
At least when working under heavy load with mnesia, detecting overload and applying backpressure can be a vital part of load tolerance.

apps/aehttp/src/aehttp_dispatch_ext.erl
uwiger: It might be worthwhile to instead making it a metric (e.g. histogram)

uwiger: btw, `timer:tc/1` doesn't catch exceptions like `timer:tc/3` does, so in that sense, it's much more transparent.

apps/aehttp/priv/swagger.json
```diff
@@ -210,8 +210,7 @@
           "in" : "path",
           "description" : "The height of the block",
           "required" : true,
-          "type" : "integer",
-          "minimum" : 0
+          "type" : "integer"
```
ThomasArts: This is what Juraj had just fixed. The mimimum 0 is important here!

uwiger:
Closing and creating a new PR

PR #1673, opened 2018-10-19
Check argument types in contract calls (RFC)

gorillainduction:
Just a first progress PR. I don't intend to merge this, but I welcome comments on the approach.

### DIFF
--- a/apps/aecontract/src/aect_dispatch.erl
+++ b/apps/aecontract/src/aect_dispatch.erl
@@ -61,8 +61,8 @@ run(?AEVM_01_Sophia_01, #{code := SerializedCode} = CallDef) ->
     #{ byte_code := Code
      , type_info := TypeInfo} = aeso_compiler:deserialize(SerializedCode),
     case check_type_info(maps:get(call_data, CallDef), TypeInfo) of
-        ok ->
-            CallDef1 = CallDef#{code => Code},
+        {ok, CallDataType} ->
+            CallDef1 = CallDef#{code => Code, call_data_type => CallDataType},
             run_common(CallDef1, ?AEVM_01_Sophia_01);
         {error, What} ->
             Gas = maps:get(gas, CallDef),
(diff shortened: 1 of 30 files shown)


hanssv:
I think the approach looks sound :-)

apps/aecontract/src/aect_dispatch.erl
```diff
@@ -179,3 +191,45 @@ chain_state(#{ contract    := ContractPubKey
...
+                        {ok, GivenArgType} -> ok;
+                        {ok, ArgType} ->
+                            %% Check if both type reps give the same result
+                            ?DEBUG_LOG("Checking representation for ~p\n"
+                                       "Given   : ~p\n"
+                                       "Accepted: ~p"
+                                      , [Function, GivenArgType, ArgType]),
+                            ExpectedType = {tuple, [string, ArgType]},
+                            case aeso_data:from_binary({tuple, [typerep, ExpectedType]}, CallData) of
+                                {ok, {_, {_, Args}}} -> ok;
```
UlfNorell: Is this what we want? In particular it allows one to specify incorrect types of the constructors of a datatype that you aren't using. This is harmless, I suppose, but it's not clear why we'd want to allow it.

apps/aesophia/src/aeso_compiler.erl
```diff
@@ -84,12 +93,34 @@ to_bytecode([Op|Rest], Options) ->
     [aeb_opcodes:m_to_op(Op)|to_bytecode(Rest, Options)];
 to_bytecode([], _) -> [].
 
+extract_type_info(#{functions := Functions} =_Icode) ->
+    io:format("Functions: ~p\n", [Functions]),
+    TypeInfo = [{list_to_binary(Name),
+                 aeso_data:to_binary({tuple, [T || {_, T} <- Args]}),
+                 aeso_data:to_binary(TypeRep)}
+                || {Name,_Attrs, Args,_Body, TypeRep} <- Functions,
```
UlfNorell: Should only extract type info for public functions:
```suggestion
                || {Name, Attrs, Args,_Body, TypeRep} <- Functions,
                   not lists:member(private, Attrs),
```

gorillainduction:
I'm dropping this in favor of the rebased PR https://github.com/aeternity/epoch/pull/1696

PR #1679, opened 2018-10-22
PT-161380219: Add check to disallow reentrant calls for Sophia EVM

UlfNorell:
[PT-161380219](https://www.pivotaltracker.com/story/show/161380219)

### DIFF
--- a/apps/aevm/src/aevm_eeevm_state.erl
+++ b/apps/aevm/src/aevm_eeevm_state.erl
@@ -153,26 +153,35 @@ init_vm(State, Code, Mem, Store) ->
         ?AEVM_01_Solidity_01 ->
             aevm_eeevm_store:init(Store, State1);
         ?AEVM_01_Sophia_01 ->
-            %% We need to import the state first, since the map ids in the store are fixed.
-            State2 = import_state_from_store(Store, State1),
-            %% Next we write the calldata on top of the heap and put a pointer
-            %% to it on the stack.
-            Calldata = data(State1),
-            %% Calldata can contain maps, so we can't simply write it
-            %% to memory. The calldata should be a pair of a typerep
-            %% and the actual calldata.
-            HeapSize = aevm_eeevm_memory:size_in_words(State2) * 32,
-            case aeso_data:from_binary({tuple, [typerep]}, Calldata) of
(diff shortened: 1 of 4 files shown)


apps/aevm/src/aevm_eeevm_state.erl
```diff
@@ -153,27 +153,36 @@ init_vm(State, Code, Mem, Store) ->
...
-                    {ok, CalldataHeap} = aeso_data:binary_to_heap({tuple, [typerep, Type]}, Calldata,
-                                                                  aevm_eeevm_maps:next_id(maps(State2)), HeapSize),
-                    {Ptr, State3} = write_heap_value(CalldataHeap, State2),
-                    aevm_eeevm_stack:push(Ptr, State3);
-                {error, Err} ->
-                    io:format("** Error invalid calldata: ~p\n", [Err]),
-                    set_gas(0, State2)
+            case is_reentrant_call(State) of
+                true -> %% Sophia doesn't allow reentrant calls
+                    io:format("** Attempted reentrant call\n"),
```
lucafavatella: Lager preferred.

docs/release-notes/next/PT-161380219-check-for-reentrant-calls.md
```diff
@@ -0,0 +1 @@
+* Disallows reentrant calls (i.e. contract calls into a currently executing contract) in Sophia.
```
lucafavatella: I see documented [not-yet-implemented attach transaction](https://github.com/aeternity/protocol/blame/a8fc667df4b777ebbb7d5edbffe377d6ce3a085a/contracts/contract_transactions.md#L6). Does this PR answer any of the following questions?
* Is contract call transaction with caller equal to recipient allowed?
* Does contract create transaction allow initialization code to call back to sender of contract create transaction?

UlfNorell: At the moment we allow calls to the origin address (i.e. the signer of the top-level call transaction), see `aect_call_tx:new/1` which sets the call stack to empty if not present. Same for `aect_create_tx`. This PR doesn't take an active position on whether we should allow this, however.

PR #1680, opened 2018-10-22
System test that trigger sync of many full micro blocks per generation

ThomasArts:
Show that forward sync can fail
Fix forward sync

### DIFF
--- a/apps/aecore/src/aec_sync.erl
+++ b/apps/aecore/src/aec_sync.erl
@@ -685,17 +685,16 @@ post_blocks(From, To, [{Height, _Hash, {PeerId, Block}} | Blocks]) ->
 %% smaller chuncks, a few micro blocks at the time.
 %% Each micro block has a fixed maximum gas, by limiting the number of micro
 %% blocks we limit the total amount of work the conductor has to perform in
-%% each synchronous call. 
+%% each synchronous call.
+%% Map contains key dir, saying in which direction we sync
 add_generation(#{micro_blocks := MBlocks} = Generation) ->
-    add_generation(maps:without([micro_blocks], Generation), MBlocks).
+    add_micro_blocks_in_generation(Generation#{micro_blocks => []}, MBlocks).
 
--define(MAX_NR_MICROBLOCKS, 4).
-add_generation(Gen, MBlocks) when length(MBlocks) < ?MAX_NR_MICROBLOCKS ->
-    aec_conductor:add_synced_generation_batched(Gen#{micro_blocks => MBlocks}, true);
-add_generation(Gen, MBlocks) ->
-    {Add, AddLater} = lists:split(?MAX_NR_MICROBLOCKS, MBlocks),
-    case aec_conductor:add_synced_generation_batched(Gen#{micro_blocks => Add}, false) of
-        ok -> add_generation(Gen, AddLater);
+add_micro_blocks_in_generation(Gen, []) ->
+    aec_conductor:add_synced_generation_batched(Gen, true);
+add_micro_blocks_in_generation(Gen, [MBlock | MBlocks]) ->
+    case aec_conductor:add_synced_generation_batched(Gen#{micro_blocks => [MBlock]}, false) of
+        ok -> add_micro_blocks_in_generation(Gen, MBlocks);
         Error -> Error
     end.
 
@@ -761,7 +760,7 @@ do_get_generation(PeerId, LastHash) ->
             Generation = #{ key_block => KeyBlock,
                             micro_blocks => MicroBlocks,
                             dir => forward },
-            aec_conductor:add_synced_generation(Generation);
+            add_generation(Generation);
         Err = {error, _} ->
             Err
     end.
--- a/system_test/aest_heavy_sync_SUITE.erl
+++ b/system_test/aest_heavy_sync_SUITE.erl
@@ -173,12 +173,24 @@ find_txs(Node, [TxHash | TxHashes], Found, Deadline) ->
 
 
 add_spend_txs(Node, SenderAcct, N, NonceStart) ->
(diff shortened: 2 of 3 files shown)


system_test/aest_heavy_sync_SUITE.erl
```diff
@@ -0,0 +1,184 @@
+-module(aest_heavy_sync_SUITE).
```
lucafavatella: I ran this test against master version of the node and I got failure (that is good):
```
Node node2's log "epoch.log" contains errors:
2018-10-24 09:29:23.623 [error] <0.1506.0> CRASH REPORT Process <0.1506.0> with 0 neighbours exited with reason: {timeout,{gen_server,call,[aec_conductor,{add_synced_generation,#{add_keyblock => true,dir => forward,key_block => {key_block,{key_header,3...
```

system_test/aest_heavy_sync_SUITE.erl
```diff
@@ -0,0 +1,184 @@
...
+    TxHashes = add_spend_txs(node3, Patron, 2700, 1),
+
+    start_node(node1, Cfg),
+    wait_for_startup([node1], 0, Cfg),
+    LastTxHash = lists:last(TxHashes),
+
+    %% Check that the valid transactions made it to the chain.
+    #{node1 := [{_Hash, Height}]} = 
+        wait_for_value({txs_on_chain, [LastTxHash]},
+                   [node1], 10 * ?MINING_TIMEOUT, [{key_blocks, 0} | Cfg]),
```
lucafavatella: I do not see where `key_blocks` option is used.

hanssv: It is to _not_ wait for the next keyblock, thus we will have a large "current generation" which was what triggers the timeout.

system_test/aest_heavy_sync_SUITE.erl
```diff
@@ -0,0 +1,184 @@
...
+                    backend => aest_docker,
+                    source  => {pull, "aeternity/epoch:local"},
+                    config  => #{persist => true}
+                  },
+                 #{ name    => node3,
+                    peers   => [node1],
+                    backend => aest_docker,
+                    source  => {pull, "aeternity/epoch:local"},
+                    config  => #{persist => true},
+                    mining  => #{autostart => false}
```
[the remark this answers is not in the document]
ThomasArts: aest_docker magic, not pretty, but it works.

system_test/aest_heavy_sync_SUITE.erl
ThomasArts: It is in aest_nodes:wait_for_value({txs_on_chain,... but probably that only made it to master recently (PR1670)

apps/aecore/src/aec_sync.erl
```diff
@@ -761,7 +765,7 @@ do_get_generation(PeerId, LastHash) ->
             Generation = #{ key_block => KeyBlock,
                             micro_blocks => MicroBlocks,
                             dir => forward },
-            aec_conductor:add_synced_generation(Generation);
+            add_generation(Generation);
```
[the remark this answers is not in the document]
ThomasArts: Yes, intentional, but simplified in updated version anyway

system_test/aest_heavy_sync_SUITE_data/epoch.yaml.mustache
```diff
@@ -0,0 +1,36 @@
...
+keys:
+    password: {{key_password}}
+    dir: ./keys
+
+chain:
+    persist: {{config.persist}}
+
+mining:
+    beneficiary: "ak_2VoAhMd7tVJrDYM5vPJwFRjueZyirDJumVJNeBWL9j1eNTHsRx"
+    micro_block_cycle: 3000
```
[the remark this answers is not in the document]
ThomasArts: Let us agree that this is testdebt adn that we fix it when I need to revise this test in the near future

apps/aecore/src/aec_sync.erl
```diff
@@ -685,17 +685,16 @@ post_blocks(From, To, [{Height, _Hash, {PeerId, Block}} | Blocks]) ->
...
-add_generation(Gen, MBlocks) when length(MBlocks) < ?MAX_NR_MICROBLOCKS ->
-    aec_conductor:add_synced_generation_batched(Gen#{micro_blocks => MBlocks}, true);
-add_generation(Gen, MBlocks) ->
-    {Add, AddLater} = lists:split(?MAX_NR_MICROBLOCKS, MBlocks),
-    case aec_conductor:add_synced_generation_batched(Gen#{micro_blocks => Add}, false) of
-        ok -> add_generation(Gen, AddLater);
+    add_micro_blocks_in_generation(Generation#{micro_blocks => []}, MBlocks).
+
+add_micro_blocks_in_generation(Gen, []) ->
+    aec_conductor:add_synced_generation_batched(Gen, true);
```
lucafavatella: So this is adding the key block last for both backward and forward. This is a change of semantics in how forward generation is added. For forward seems either wrong or redundant as it adds microblocks first then it adds preceding keyblock, but if keyblock insertion was useful then then insertion of all microblocks would have failed as not connected to genesis / orphan. Could you please clarify?

ThomasArts: The problem we solve here is that generations can have many micro blocks and we need to put each micro block into the database *separately*. That is, it takes time for each micro block and if we do that in 1 call to conductor, then we get timeouts.

We can get generations via gossip and via sync. One is forward, the other backward, meaning that one starts with a key block and then all micro blocks before it, the other all micro blocks after the key block. It is ok to add the same key block twice and not expensive to do so. The code is considered clearer if we don't distinguish forward and backward at this level.... hence the way of writing it.

ThomasArts: So, for forward it is redundant, but the code is much nicer and it does not cost.

lucafavatella: Ok.

PR #1681, opened 2018-10-23
PT-158864866: throw away keys 

fabiankrol:
PT: https://www.pivotaltracker.com/story/show/158864866


apps/aecore/src/aec_conductor.erl
```diff
@@ -531,6 +531,10 @@ preempt_if_new_top(#state{ top_block_hash = OldHash,
                                 end,
                     State5 = State4#state{ top_key_block_hash = NewTopKey,
                                            key_block_candidate = undefined },
+
+                    KeyOrNewForkMicro == key
+                      andalso aec_keys:promote_candidate(aec_blocks:miner(NewBlock)),
+
```
hanssv:
```suggestion
[ aec_keys:promote_candidate(aec_blocks:miner(NewBlock)) || KeyOrNewForkMicro == key ]
```

apps/aecore/src/aec_keys.erl
```diff
@@ -27,6 +27,8 @@
 -export([peer_pubkey/0,
          peer_privkey/0,
          pubkey/0,
+         candidate_pubkey/0,
+         promote_candidate/1,
          sign_micro_block/1,
          sign_tx/1
```
hanssv: We agreed to remove this function!?

apps/aecore/src/aec_keys.erl
```diff
@@ -75,22 +77,27 @@
...
-          peer_priv_file :: binary(),
           peer_pass      :: password(),
-          keys_dir       :: binary()
+          keys_dir       :: binary(),
+
+          candidate_pub  = <<>> :: binary(),
+          candidate_priv = <<>> :: binary(),
+          sign_pub       = <<>> :: binary(),
+          sign_priv      = <<>> :: binary(),
+          sign_pub_file  = <<>> :: binary(),
```
hanssv: What file is this, the sign keys are throw away?

apps/aecore/src/aec_keys.erl
```diff
@@ -419,31 +448,48 @@ worker_reply_if_test(From,_Reply) ->
...
-            {Pub, Priv};
-        _ ->
-            {error, enoent}
-    end.
-
+s_promote_candidate(#worker_state{candidate_priv = Priv, candidate_pub  = Pub} = State) ->
+    s_setup_candidate_keys( State#worker_state{ sign_priv = Priv, sign_pub = Pub } ).
 
-setup_sign_keys(Pwd, KeysDir) ->
+s_setup_sign_keys(#worker_state{sign_pass = SignPwd, keys_dir = KeysDir} = State0) ->
```
hanssv: Why are we reading from a file? Shouldn't we just generate a new key-pair?

apps/aecore/src/aec_keys.erl
```diff
@@ -492,6 +545,14 @@ p_gen_new_peer(Pwd, PubFile, PrivFile) ->
             error({generated_key_pair_check_failed, [PubKey, PrivKey]})
     end.
 
+ensure_dir(KeysDir) ->
```
hanssv: Will this work if also the parent directory is missing?

apps/aecore/test/aecore_suite_utils.erl
```diff
@@ -344,11 +350,20 @@ spend(Node, FromPub, ToPub, Amount, Fee) ->
...
+
     {ok, Tx} = rpc:call(Node, aec_spend_tx, new, [Params]),
-    {ok, SignedTx} = rpc:call(Node, aec_keys, sign_tx, [Tx]),
+    {ok, SignedTx} = sign_on_node({dev1, Node}, Tx),
     ok = rpc:call(Node, aec_tx_pool, push, [SignedTx]),
     {ok, SignedTx}.
 
+sign_on_node({Id, Node}, Tx) ->
+    {_, {SignPrivKey, _SignPubKey}} = lists:keyfind(Id, 1, sign_keys()),
+    Bin = rpc:call(Node, aetx, serialize_to_binary, [Tx]),
```
hanssv:
```suggestion
    aec_test_utils:sign_tx(Tx, SignPrivKey);
```

hanssv: This is already in another utils file :-)

fabiankrol:
I’m putting WIP label. Going to push updates in the afternoon/evening

apps/aecore/src/aec_keys.erl
fabiankrol: ✅

apps/aecore/test/aecore_suite_utils.erl
fabiankrol: yes...

apps/aecore/src/aec_keys.erl
fabiankrol: Updated to use `filelib:ensure_dir/1` instead of `file:make_dir/1` ✅

apps/aecore/src/aec_conductor.erl
fabiankrol: Fine by me ✅

apps/aecore/src/aec_keys.erl
```diff
@@ -87,16 +87,16 @@
...
-          candidate_priv = <<>> :: binary(),
-          sign_pub       = <<>> :: binary(),
-          sign_priv      = <<>> :: binary(),
-          sign_pub_file  = <<>> :: binary(),
-          sign_priv_file = <<>> :: binary(),
-          peer_pub       = <<>> :: binary(),
-          peer_priv      = <<>> :: binary(),
-          peer_pub_file  = <<>> :: binary(),
-          peer_priv_file = <<>> :: binary()
+          candidate_pub  :: binary(),
```
fabiankrol: Refactoring in progress to make dialyzer happy 🍨

apps/aecore/src/aec_keys.erl
fabiankrol: There is no hard requirements for reading them from file during the start, but here are some possible future improvements that need it:
- as discussed @sennui we might want to immediately return to microblock signing after node's going down (conductor changes needed for this) 
- allow to turn off key rotation in user config

hanssv: But the current way is protected by clear text password and with some lousy weak encryption... At least comment it out until the conductor is ready for it... Key rotation off is not an option, the badness of the current approach was listed in the security review even ;-)

sennui: My point back then was about question: to save or not to save?. 

In case of hypothetical zero-day exploit that can crash a node, without persistence we face following consequences:
1. Miners that solved the puzzle may be shut down and signing of mic blocks may be stopped
2. Launching signing on another node is not straight forward as it is hard to extract keys 
   
I totally agree with @hanssv, tho, that we can't keep it on disc as we do now. 
What's the viable option for the time we have until mainnet? Some sort of injection of keys from external secure source? (it makes us depend on 3rd party services)

fabiankrol: I underestimated the meaning of disk stored keys from a security perspective. I'm going to remove it 👌

fabiankrol:
Marked as `WIP`: I'm going to remove disk stored keys.

fabiankrol:
Infrastructure PR: https://github.com/aeternity/infrastructure/pull/133
Follow-up configuration PR (blocked by above): https://github.com/aeternity/epoch/pull/1706

fabiankrol:
rebased

PR #1687, opened 2018-10-23
Pt 159522821 abort revert in vm

rvirding:
This implements sophia abort with the REVERT aevm instruction.

This completes [PT-159522821](https://www.pivotaltracker.com/story/show/159522821)


apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -1139,13 +1187,54 @@ get_nonce(Pub) ->
             Nonce0 + NonceOff + 1
     end.
 
+%% call_compute_func(NodeName, Pubkey, Privkey, EncodedContractPubkey,
```
hanssv: Doesn't look like a very important addition ;-)

rvirding:
Added release note with a commit which was probably unnecessary.

PR #1676, opened 2018-10-20
SC: fix security issue with forcing of progress

velzevur:
PT [SC: protocol fix - force-solo_close-force2](https://www.pivotaltracker.com/story/show/160613339)
Protocol changes are already in PR aeternity/protocol#241

PR #1688, opened 2018-10-24
Pt 161281705 clear retval in init

happi:


velzevur:
This will conflict in DB_VERSION with #1676 (which currently has no reviews whatsoever...)

lucafavatella:
@happi Proposed test in #1691

PR #1691, opened 2018-10-25
Pt 161281705 clear retval in init -- test

lucafavatella:


apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -294,7 +294,7 @@ create_contract_init_error(_Cfg) ->
     %% Check that the created init call has the correct details not from the contract create tx
     ?assertEqual(ContractKey, aect_call:contract_pubkey(InitCall)), %% Contract not created.
     ?assertEqual(error, aect_call:return_type(InitCall)),
-    _ = aect_call:return_value(InitCall),
+    ?assertEqual(<<>>, aect_call:return_value(InitCall)),
```
lucafavatella: I checked that changing this to `<<"foo">>` test fails.

PR #1696, opened 2018-10-25
Change calling convention in sophia and add type checking

gorillainduction:
* Type representions are no longer in call data
* Function names are no longer in call data
* The function signature (argtypes and outtype) is stored in meta data
* Calls are done based on a function hash that includes the type signature
* Call data needs the byte code to be encoded
* Primop types are fetched from the chain when needed (e.g., for oracles)

This makes:
- the byte code smaller
- the memory usage smaller
- the contract calls more type safe
- the chain primops more type safe

TODO (longer term):
- Check the type signatures on oracle primops so that the compiled
  code is expecting the correct query/response types.
- Check the types on name resolution. I'm not entirely sure it is safe.

TODO (for this PR):

- [x] Bump protocol version and db dir
- [x] Add release notes
- [ ] Consider merging the new way of constructing call data.
- [ ] Consider removing explicit typereps for the `RETURN` op, and use type info instead.


gorillainduction:
Fixed a stray rebase error.

gorillainduction:
Again, dropping this for a new PR on a rebased and squashed branch https://github.com/aeternity/epoch/pull/1702

Sorry for the noise, but we are two persons constantly rebasing.

PR #1698, opened 2018-10-25
Don't revisit txs until the next leader cycle

uwiger:
See [PT #161473896](https://www.pivotaltracker.com/story/show/161473896).

A 'visited' table is added to the tx_pool. Each tx that's selected (or judged invalid) in `get_candidate()` is moved to the 'visited' table until the next key block top change.

This might make it harder to thwart microblock generation by stuffing a large, expensive, but guaranteed-to-fail transation into the pool.

### DIFF
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -507,10 +507,11 @@ preempt_if_new_top(#state{ top_block_hash = OldHash,
     case aec_chain:top_block_hash() of
         OldHash -> no_change;
         NewHash ->
-            ok = aec_tx_pool:top_change(OldHash, NewHash),
+            {ok, NewBlock} = aec_chain:get_block(NewHash),
+            BlockType = aec_blocks:type(NewBlock),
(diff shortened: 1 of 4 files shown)


apps/aecore/src/aec_conductor.erl
```diff
@@ -507,18 +507,19 @@ preempt_if_new_top(#state{ top_block_hash = OldHash,
     case aec_chain:top_block_hash() of
         OldHash -> no_change;
         NewHash ->
-            ok = aec_tx_pool:top_change(OldHash, NewHash),
+            {ok, NewBlock} = aec_chain:get_block(NewHash),
```
hanssv: We had `Block` where we called `preempt_if_new_top` maybe pass it instead?

uwiger: I tried. It wasn't a 'real' block, so I decided to play it safe.

hanssv: If only Thomas had refactored the code the way some reviewer suggested it would have been a real block ;-)

PR #1702, opened 2018-10-26
Change calling convention in sophia and add type checking

gorillainduction:
* Type representions are no longer in call data
* Function names are no longer in call data
* The function signature (argtypes and outtype) is stored in meta data
* Calls are done based on a function hash that includes the type signature
* Call data needs the byte code to be encoded
* Primop types are fetched from the chain when needed (e.g., for oracles)

This makes:
- the byte code smaller
- the memory usage smaller
- the contract calls more type safe
- the chain primops more type safe

TODO:
- Check the type signatures on oracle primops so that the compiled
  code is expecting the correct query/response types.
- Check the types on name resolution. I'm not entirely sure it is safe.

TODO (for this PR):

- [x] Bump protocol version and db dir
- [x] Add release notes
- [x] Consider removing explicit typereps for the `RETURN` op, and use type info instead.
- [x] Merge the new way of constructing call data.

### DIFF
--- a/apps/aechannel/test/aesc_txs_SUITE.erl
+++ b/apps/aechannel/test/aesc_txs_SUITE.erl
@@ -2382,10 +2382,8 @@ fp_use_onchain_oracle(Cfg) ->
 
                 % verify that Oracle.query_fee works
                 fun(#{query_id := QueryId} = Props) ->
-                    EncodedQueryId = aeu_hex:hexstring_encode(QueryId),
                     (force_call_contract(Forcer, <<"query_fee">>,
-                                         <<"(", EncodedQueryId/binary,
-                                           ")">>))(Props)
+                                         <<"()">>))(Props)
                 end,
                 assert_last_channel_result(QueryFee, word)
                ])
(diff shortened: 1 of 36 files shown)


UlfNorell:
Previously you created calldata by giving a function name and a Sophia constant for the arguments. The new way to create calldata for a call to `foo` is to give a complete contract of the form
```
contract SomeName =
  function foo : (t1, t2, t3) => ret_t
  function __call() = foo(arg1, arg2, arg3)
```
This contract is run through the type checker and can include type definitions necessary for the type of `foo`. To make updating the tests not take weeks I kept the interface of passing a function string and an argument string. For the example above you set the function to `""` and pass the contract code in the argument. There's also a shorthand which made updating tests less of a nightmare: setting the function to `"foo : (t1, t2, t3) => ret_t"` and the arguments to `"arg1, arg2, arg3"` will build the contract code above. This isn't super nice but meant I only had to add a type signature to update most of the tests.

I'm not sure what's the best way to handle the `CreateCompute` and `CallCompute` entrypoints though. At the moment `CreateCompute` only takes `argument` and not a function name (since it's always `init`), so you can't use the shorthand with that. `CallCompute` maps directly to `create_calldata` so admits the shorthand. In the long term I think we'd like to change the arguments to these entrypoints to something nicer, but that would mean a lot of time spent on updating tests.

PR #1703, opened 2018-10-26
Add bad transfer test in erc20 contract

rvirding:
This is to test that when an abort occurs in the erc20 token test the state is preserved unchanged.

This completes [PT-160751003](https://www.pivotaltracker.com/story/show/160751003)

### DIFF
--- a/apps/aehttp/test/aehttp_contracts_SUITE.erl
+++ b/apps/aehttp/test/aehttp_contracts_SUITE.erl
@@ -1013,7 +1013,6 @@ erc20_token_contract(Config) ->
     call_func(APub, APriv, EncCPub, <<"transferFrom">>, args_to_binary([BPub,DPub,10000])),
     call_func(APub, APriv, EncCPub, <<"transferFrom">>, args_to_binary([CPub,DPub,15000])),
 
-
     %% Check the balances.
     call_func(APub, APriv, EncCPub, <<"balanceOf">>, args_to_binary([BPub]), {<<"int">>, 10000}),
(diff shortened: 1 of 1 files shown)


apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -1034,6 +1033,22 @@ erc20_token_contract(Config) ->
 
     force_fun_calls(Node),
 
+    %% Request a bad transfer and check that abort restores the state.
+    revert_call_compute_func(Node, APub, APriv, EncCPub,
+                             <<"transferFrom">>,
+                             args_to_binary([BPub,DPub,100000])),
```
lucafavatella: I see B does not have 100000 so this is meant to fail in sub in first line of transferFrom:
```
  public stateful function transferFrom(from : address, to : address, tokens : int) =
    put( state{ balances[from] = sub(state.balances[from], tokens) })
    put( state{ allowed[from][Call.caller] = sub(state.allowed[from][Call.caller], tokens) })
    put( state{ balances[to] = add(balanceOf(to), tokens) })
    transferEvent(from, to, tokens)
    true
```
Ok.

I checked replacing here 5000, and the test correctly failed (as B has 10000 as balance, and is still allowed to transfer 5000). Ok.

I checked replacing here 5001, and the test correctly still passed (as B has 10000 as balance, but is allowed to transfer 5000 only). This triggered case mentioned in [previous review](https://github.com/aeternity/epoch/pull/1569#discussion_r219817368). Ok.

PR #1709, opened 2018-10-29
PT-160256544 calldata creation part

UlfNorell:
See https://github.com/aeternity/epoch/pull/1702 for the other part of the ticket.

Previously you created calldata by giving a function name and a Sophia constant for the arguments. The new way to create calldata for a call to `foo` is to give a complete contract of the form
```
contract SomeName =
  // here can be type defs
  function foo : (t1, t2, t3) => ret_t  // ret_t can be _ in which case it's not checked 
  function __call() = foo(arg1, arg2, arg3)
```
This contract is run through the type checker and can include type definitions necessary for the type of `foo`. To make updating the tests not take weeks I kept the interface of passing a function string and an argument string. For the example above you set the function to `""` and pass the contract code in the argument. There's also a shorthand which made updating tests less of a nightmare: setting the function to `"foo : (t1, t2, t3) => ret_t"` and the arguments to `"arg1, arg2, arg3"` will build the contract code above. This isn't super nice but meant I only had to add a type signature to update most of the tests.

I'm not sure what's the best way to handle the `CreateCompute` and `CallCompute` entrypoints though. At the moment `CreateCompute` only takes `argument` and not a function name (since it's always `init`), so you can't use the shorthand with that. `CallCompute` maps directly to `create_calldata` so admits the shorthand. In the long term I think we'd like to change the arguments to these entrypoints to something nicer, but that would mean a lot of time spent on updating tests.

TODO
  - [x] Release notes
  - [x] Update EncodeCallData entrypoint
  - [x] Document entrypoint changes: https://github.com/aeternity/protocol/pull/257
  - [x] Tests for new entrypoints

### DIFF
--- a/apps/aehttp/priv/swagger.json
+++ b/apps/aehttp/priv/swagger.json
@@ -3847,7 +3847,7 @@
     },
     "ContractCallInput" : {
       "type" : "object",
-      "required" : [ "abi", "arg", "code", "function" ],
+      "required" : [ "abi", "code" ],
       "properties" : {
         "abi" : {
           "type" : "string"
@@ -3860,9 +3860,14 @@
         },
         "arg" : {
           "type" : "string"
+        },
+        "call" : {
+          "type" : "string",
+          "description" : "Source code for a contract with a function __call() = f(args), if calling a function f"
         }
       },
       "example" : {
+        "call" : "call",
         "code" : "code",
         "function" : "function",
         "arg" : "arg",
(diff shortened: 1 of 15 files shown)


UlfNorell:
Calldata stuff in separate PR: https://github.com/aeternity/epoch/pull/1709

UlfNorell:
An alternative is to keep the old code and make completely new entrypoints for the new calldata creation.

UlfNorell:
This should now be ready for review.

UlfNorell:
- Negative tests should be added at some point, but were not due to time constraints.
- The name `__call` is not reserved, so I suppose if you want to call a function named `__call` you're out of luck (or need to build the calldata manually at any rate). I felt the double underscores would make this scenario unlikely, and even if it does happen it's not a serious problem.

PR #1712, opened 2018-10-29
PT-161532247: Introduce debug api for beneficiary pubkey

fabiankrol:
PT: [PT-161532247](https://www.pivotaltracker.com/story/show/161532247)


fabiankrol:
Rebased on top of master with throw away keys -> test updated

PR #1716, opened 2018-10-31
New governance values:

ThomasArts:
- We can fit a few more than 300 spend transactions in a block
    Therefore can reach theoretical throughput of 100 txs per second
- Contract call and create are using more gas, because they take more time
- Maps and names should be measured more accurately in the future

### DIFF
--- a/apps/aecore/src/aec_governance.erl
+++ b/apps/aecore/src/aec_governance.erl
@@ -46,7 +46,9 @@
 -define(BLOCK_MINE_REWARD, 10000000000000000000).
 %% Ethereum's gas limit is 8 000 000 and block time ~15s.
 %% For 3s block time it's 1 600 000 (5x less).
--define(BLOCK_GAS_LIMIT, (4*1600000)).
+%% But we can improve and use 6 million (this allows reasonable
+%% size contracts to fit in a microblock at all)
+-define(BLOCK_GAS_LIMIT, 6000000).
 -define(TX_BASE_GAS, 15000).
 %% Gas for 1 byte of a serialized tx.
 -define(BYTE_GAS, 20).
@@ -95,8 +97,8 @@ block_mine_reward() ->
 block_gas_limit() ->
     application:get_env(aecore, block_gas_limit, ?BLOCK_GAS_LIMIT).
 
-tx_base_gas(contract_call_tx) -> 5 * ?TX_BASE_GAS;
-tx_base_gas(contract_create_tx) -> 4 * ?TX_BASE_GAS;
+tx_base_gas(contract_call_tx) ->  30 * ?TX_BASE_GAS;
+tx_base_gas(contract_create_tx) -> 5 * ?TX_BASE_GAS;
 tx_base_gas(spend_tx) -> ?TX_BASE_GAS;
 tx_base_gas(channel_deposit_tx) -> ?TX_BASE_GAS;
 tx_base_gas(channel_close_mutual_tx) -> ?TX_BASE_GAS;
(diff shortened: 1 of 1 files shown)


apps/aecore/src/aec_governance.erl
```diff
@@ -163,26 +165,26 @@ state_gas_per_block(oracle_response)     -> state_gas_per_block(oracle_registrat
...
+primop_base_gas(?PRIM_CALL_AENS_CLAIM         ) -> 12000;
+primop_base_gas(?PRIM_CALL_AENS_UPDATE        ) -> 12000;
+primop_base_gas(?PRIM_CALL_AENS_TRANSFER      ) -> 12000;
+primop_base_gas(?PRIM_CALL_AENS_REVOKE        ) -> 12000;
+primop_base_gas(?PRIM_CALL_MAP_EMPTY          ) -> 1000;
+primop_base_gas(?PRIM_CALL_MAP_GET            ) -> 1000;
+primop_base_gas(?PRIM_CALL_MAP_PUT            ) -> 1000;
+primop_base_gas(?PRIM_CALL_MAP_DELETE         ) -> 1000;
+primop_base_gas(?PRIM_CALL_MAP_SIZE           ) -> 1000;
+primop_base_gas(?PRIM_CALL_MAP_TOLIST         ) -> 6000.
```
hanssv: Why would `tolist` be 6 times more expensive than the other maps operations. Once converted to a list we will charge for its size/memory?!

apps/aecore/src/aec_governance.erl
```diff
@@ -163,26 +165,26 @@ state_gas_per_block(oracle_response)     -> state_gas_per_block(oracle_registrat
...
+primop_base_gas(?PRIM_CALL_ORACLE_GET_ANSWER  ) -> 12000;
+primop_base_gas(?PRIM_CALL_ORACLE_GET_QUESTION) -> 12000;
+primop_base_gas(?PRIM_CALL_ORACLE_QUERY_FEE   ) -> 12000;
+primop_base_gas(?PRIM_CALL_AENS_RESOLVE       ) -> 12000;
+primop_base_gas(?PRIM_CALL_AENS_PRECLAIM      ) -> 12000;
+primop_base_gas(?PRIM_CALL_AENS_CLAIM         ) -> 12000;
+primop_base_gas(?PRIM_CALL_AENS_UPDATE        ) -> 12000;
+primop_base_gas(?PRIM_CALL_AENS_TRANSFER      ) -> 12000;
+primop_base_gas(?PRIM_CALL_AENS_REVOKE        ) -> 12000;
+primop_base_gas(?PRIM_CALL_MAP_EMPTY          ) -> 1000;
```
hanssv: I'd like to see some measurements before setting map operations higher than 0...

apps/aecore/src/aec_governance.erl
```diff
@@ -95,8 +97,8 @@ block_mine_reward() ->
 block_gas_limit() ->
     application:get_env(aecore, block_gas_limit, ?BLOCK_GAS_LIMIT).
 
-tx_base_gas(contract_call_tx) -> 5 * ?TX_BASE_GAS;
-tx_base_gas(contract_create_tx) -> 4 * ?TX_BASE_GAS;
+tx_base_gas(contract_call_tx) ->  30 * ?TX_BASE_GAS;
```
hanssv: Wow, that is steep... But if that is how long it takes to run a contract...

hanssv:
NOTE: The values for MAP_XXX has no effect...

lucafavatella:
Map base costs may need to go back to zero.

apps/aecore/src/aec_governance.erl
ThomasArts: Yes, steep, but for a call to idle function compared to normal spend... a reasonable factor

ThomasArts: Based on a normal spend compared to a contract call to the idle function

hanssv: OK, we are working on improving this ;-)

lucafavatella:
I see CI [failed](https://circleci.com/gh/aeternity/epoch/39456) for `cannot_get_gc_height` that shall be fixed in master. I am taking the liberty of restarting CI.

PR #1717, opened 2018-10-31
Pt 153819317 check oracle types

gorillainduction:
https://www.pivotaltracker.com/story/show/153819317

This is a protocol breaking change since it adds the field `vm_version` to the oracles.

- [x] Add test for ?AEVM_NO_VM interaction from contracts
- [x] Bump protocol and deploy db dir


gorillainduction:
Rebased on master

PR #1719, opened 2018-10-31
Fixes PT-#161619322, always check Visited after checking the main pool

uwiger:
Previously, the Visited set was only checked if the main pool check
didn't yield any candidates at all. This could lead to (temporary)
starvation in the Visited set.

See [PT #161619322](https://www.pivotaltracker.com/story/show/161619322)

### DIFF
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -337,18 +337,17 @@ int_get_candidate(MaxGas, BlockHash, #dbs{db = Db} = DBs) ->
     {ok, Trees} = aec_chain:get_block_state(BlockHash),
     {ok, Header} = aec_chain:get_header(BlockHash),
     lager:debug("size(Db) = ~p", [ets:info(Db, size)]),
-    case int_get_candidate(Db, MaxGas, Trees, Header, DBs) of
-        {ok, []} ->
-            int_get_candidate(DBs#dbs.visited_db, MaxGas, Trees, Header, DBs);
-        Other ->
-            Other
-    end.
+    {ok, RemGas, Acc} = int_get_candidate(Db, MaxGas, Trees, Header, DBs,
+                                          gb_trees:empty()),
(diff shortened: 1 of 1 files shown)


apps/aecore/src/aec_tx_pool.erl
```diff
@@ -337,18 +337,17 @@ int_get_candidate(MaxGas, BlockHash, #dbs{db = Db} = DBs) ->
...
     lager:debug("size(Db) = ~p", [ets:info(Db, size)]),
-    case int_get_candidate(Db, MaxGas, Trees, Header, DBs) of
-        {ok, []} ->
-            int_get_candidate(DBs#dbs.visited_db, MaxGas, Trees, Header, DBs);
-        Other ->
-            Other
-    end.
+    {ok, RemGas, Acc} = int_get_candidate(Db, MaxGas, Trees, Header, DBs,
+                                          gb_trees:empty()),
+    {ok, _, Acc1} = int_get_candidate(
```
hanssv: This means we traverse the whole visited every time?

hanssv: Perhaps we should abort if RemGas is smaller than something...

apps/aecore/src/aec_tx_pool.erl
```diff
@@ -337,32 +337,49 @@ int_get_candidate(MaxGas, BlockHash, #dbs{db = Db} = DBs) ->
...
     lager:debug("size(Db) = ~p", [ets:info(Db, size)]),
-    case int_get_candidate(Db, MaxGas, Trees, Header, DBs) of
-        {ok, []} ->
-            int_get_candidate(DBs#dbs.visited_db, MaxGas, Trees, Header, DBs);
-        Other ->
-            Other
-    end.
-
-int_get_candidate(Db, MaxGas, Trees, Header, DBs) ->
+    GasLimit = aec_governance:tx_base_gas(spend_tx),
```
ThomasArts: So this is a lower limit? A spend_tx is kind of cheapest thing to do...

PR #1724, opened 2018-11-01
Improve contract execution performance

UlfNorell:
Improve contract execution performance by not looking up the store all the time

Getting the store from the state trees is expensive, so we shouldn't do it more than we have to. Two fixes:
- Save the store in the VM state during contract execution
- Add a version of looking up a contract in the state trees that does not return the store
  and use this when appropriate

### DIFF
--- a/apps/aevm/src/aevm_eeevm_state.erl
+++ b/apps/aevm/src/aevm_eeevm_state.erl
@@ -195,10 +195,10 @@ is_reentrant_call(State) ->
 
 %% TODO: Currently the Store is saved both in the chain state and the VM state.
 %%       Refactor?
-get_store(#{chain_api := ChainAPI, chain_state := ChainState}) ->
-    ChainAPI:get_store(ChainState).
+get_store(#{storage := Store}) -> Store.
 
-import_state_from_store(Store, State) ->
+import_state_from_store(Store, State0) ->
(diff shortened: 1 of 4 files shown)


lucafavatella:
I see CI [failed](https://circleci.com/gh/aeternity/epoch/39492) for `{timeout,{gen_server,call,[aec_tx_pool,{top_change` that shall be fixed in master.

lucafavatella:
I took the liberty of restarting CI.

apps/aevm/src/aevm_eeevm_state.erl
```diff
@@ -195,10 +195,10 @@ is_reentrant_call(State) ->
 
 %% TODO: Currently the Store is saved both in the chain state and the VM state.
 %%       Refactor?
-get_store(#{chain_api := ChainAPI, chain_state := ChainState}) ->
-    ChainAPI:get_store(ChainState).
+get_store(#{storage := Store}) -> Store.
```
lucafavatella: I need to understan the changes in this module. The rest seems ok.

UlfNorell: The change is to remember the store in the VM state, rather than to compute it from state trees every time we need it.

This is ok, since we don't write to the store during contract execution (no reentrant calls in Sophia).

PR #1731, opened 2018-11-05
Re-use address in aehttp_client

hanssv:


hanssv:
This is necessary for load tests (in repo `aeload`) to work well.

PR #1733, opened 2018-11-05
Pt 161616592 base64 encoding

gorillainduction:
https://www.pivotaltracker.com/story/show/161616592

The base58 encoding/decoding is hopelessly slow for large byte arrays.

* Switch to base64 encoding for all binary data but identifiers
* Use a checked base64 encoding analogous to the base58c checksum (i.e., add the first 4 bytes of the twice sha256 hashed binary to the end of the byte array before base64 encoding the binary)
* Move the aec_base58c module to the aehttp application and rename it to aehttp_api_encoder


gorillainduction:
Protocol PR: https://github.com/aeternity/protocol/pull/261

dschoeni:
I just wanted to comment that base64 is not URL-safe per se, sth we got hinted to by @davidyuk while implementing a sync-scheme for AirGap.

See: https://en.wikipedia.org/wiki/Base64#URL_applications

Not sure how this would impact you, but when supplying TXs for example as a parameter in URLs, developers would need to be aware that they have to properly URL-Encode them.

PR #1735, opened 2018-11-05
Making accounts.json amounts strings

velzevur:
PT [New accounts.json format](https://www.pivotaltracker.com/story/show/161710085)
This is complimented by:
* aeternity/infrastructure#145
* aeternity/testnet-keys#6


ThomasArts:
I have a bit of a hard time to understand why the script could not simply post_process and leave the quotes away for the amounts?

velzevur:
@ThomasAerts  hat's because we're expected to use the `accounts.json` as it is provided by the scraper script. Everyone must be able to run it locally and verify it, possibly via a hash.
Probably this could be a subject of the post script so we can use its hash for verification instead

velzevur:
As discussed - this is to be handled by the script itself. Closing.

PR #1736, opened 2018-11-05
PT-161507006 Roll back state changes on contract call failure

UlfNorell:
In particular, roll back the value transfer.

Pivotal: [PT-161507006](https://www.pivotaltracker.com/story/show/161507006)
Protocol: https://github.com/aeternity/protocol/pull/262

@lucafavatella: I updated your oracle tests without thinking too hard about whether they are behaving correctly. Please verify.


gorillainduction:
I expect that you will need to change [this](https://github.com/aeternity/epoch/blob/0eac0e766c1c5b2fdc621a1b96e83ceaf4997702/py/tests/integration/test_unsigned_tx.py#L151) as well

apps/aecontract/src/aect_call_tx.erl
```diff
@@ -182,19 +182,32 @@ process(#contract_call_tx{caller_id   = CallerId,
...
-    %% Refund unused gas.
+    %% Roll back state changes if call failed
     Trees4 =
+        case aect_call:return_type(Call) of
+            ok -> Trees3;
+            E  ->
+                lager:debug("Contract call error ~w ~w~n",[E, Call]),
+                %% We still need to charge gas and fee
+                case Charges of
+                    0 -> Trees1;
```
lucafavatella: This seems reverting nonce change in case of `CALL` instruction. Could you please confirm?

I expect impacts of this are:
* `CALL` caller nonce not bumped if call errs or reverts.
* Contract call object in state tree overwritten (hence initial ones lost) in case of multiple `CALL`s by same caller in same generation (or same microblock, or same contract create/call tx) if without further nonce bumps in between.
  * Question that follows from this point is whether contract call object is/shall be written in contract call tree only for outermost create/call tx.

Related doc:
* https://github.com/aeternity/protocol/blame/14db60b7db4d51618ebad539aaf6a0b40b72e633/contracts/contract_state_tree.md#L40-L42
* https://github.com/aeternity/protocol/blame/14db60b7db4d51618ebad539aaf6a0b40b72e633/contracts/contract_state_tree.md#L62-L68

gorillainduction: I agree that the nonce still needs to be bumped. The nonce bump was hidden in the spend transaction previously.

UlfNorell: Note that the nonce is bumped for the outermost call (`Charges > 0`). If the call fails we (with this PR) roll back all state changes, including storing the call objects for inner calls, so we should either bump nonces *and* keep call objects, or do neither (which is what this PR implements).

I think it would be really weird to keep call objects for inner calls of contract executions that ultimately fail. Semantically it shouldn't matter if a call fails before or after an inner call.

PR #1737, opened 2018-11-05
Don't push the type hash on the stack for remote calls

gorillainduction:
https://www.pivotaltracker.com/story/show/161718149

This breaks breaks consensus since earlier versions of the vm cannot run code generated with this compiler. Earlier versions of code can be run by this vm, though.

- [ ] Update deploy dir and protocol version. (wait for rebase since other PRs already does this.)


gorillainduction:
Rebased and made the bumping.

PR #1738, opened 2018-11-05
Fix system tests

ThomasArts:
local system test aest_heavy_sync needed updated config:

 keys:
-    password: {{key_password}}
+    peer_password: {{key_password}}

Fixed error in mining autostart at the same time
Added eunit test for tx_pool emerging from intrernal discussions

### DIFF
--- a/system_test/only_local/aest_heavy_sync_SUITE.erl
+++ b/system_test/only_local/aest_heavy_sync_SUITE.erl
@@ -106,8 +106,18 @@ many_spend_txs(Cfg) ->
     %% That is a sligtly unrealistic situation, but if we first mine a few key blocks,
     %% then this test takes ages.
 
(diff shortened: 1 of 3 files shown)


system_test/only_local/aest_heavy_sync_SUITE.erl
```diff
@@ -138,7 +148,7 @@ many_spend_txs(Cfg) ->
     ct:log("MicroBlock distribution ~p", [FoundNode3]),
 
     %% It's to be expected that at least one micro block contains max nr of transactions
-    ?assertEqual(lists:max(maps:values(FoundNode3)), TxsPerMB),
+    ?assertEqual(TxsPerMB, lists:max(maps:values(FoundNode3))),
```
hanssv: Wasn't this deemed very brittle in the internal discussion ?!

ThomasArts: Well, I picked the nonce carefully, but right so, I changed it to be max 2 transactions of.

Issue #1730, opened 2018-11-05
Mined key block posted from http api won't trigger epoch start generation


z1617:
```erlang
handle_request_('PostKeyBlock', #{'KeyBlock' := Data}, _Context) ->
    case aec_headers:deserialize_from_client(key, Data) of
        {ok, Header} ->
            KeyBlock = aec_blocks:new_key_from_header(Header),
            case aec_conductor:post_block(KeyBlock) of
                ok ->
                    {200, [], #{}};
                {error, _Rsn} ->
                    {400, [], #{reason => <<"Block rejected">>}}
            end;
        {error, _} ->
            {400, [], #{reason => <<"Invalid block">>}}
    end;


handle_post_block(Block, State) ->
    case aec_blocks:is_key_block(Block) of
        true ->
            epoch_mining:info("post_block: ~p", [Block]),
            handle_add_block(Block, State, block_received);
        false ->
            epoch_mining:info("post_micro_block: ~p", [Block]),
            handle_add_block(Block, State, micro_block_received)
    end.

setup_loop(State = #state{ consensus = Cons }, RestartMining, IsLeader, Origin) ->
    State1 = State#state{ consensus = Cons#consensus{ leader = IsLeader } },
    State2 =
        case Origin of
            block_created when IsLeader ->
                aec_block_generator:start_generation(),
                start_micro_signing(State1);
            block_received when not IsLeader ->
                aec_block_generator:stop_generation(),
                State1;
            micro_block_created when IsLeader ->
                start_micro_sleep(State1);
            Origin when Origin =:= block_created; Origin =:= micro_block_created;
                        Origin =:= block_received; Origin =:= micro_block_received;
                        Origin =:= block_synced ->
                State1
        end,
    case RestartMining of
        true  -> start_mining(State2);
        false -> State2
    end.
```

When mining with external mining client and post the key block by using the http endpoint `/key-blocks` won't trigger epoch start generation, no micro block will be signed?

@jur0 @ThomasArts

fabiankrol:
@z1617 I created [ticket for it](https://www.pivotaltracker.com/n/projects/2124891/stories/161715282) in our pivotal tracker.

fabiankrol:
@z1617 fix is merged into master

z1617:
@fabiankrol 👍

PR #1742, opened 2018-11-07
Fix microblock signing for `block_received` 

fabiankrol:
[#161715282](https://www.pivotaltracker.com/n/projects/2124891/stories/161715282)

Github issue: #1730

_...rebased..._

### DIFF
--- a/apps/aecore/src/aec_block_generator.erl
+++ b/apps/aecore/src/aec_block_generator.erl
@@ -11,7 +11,10 @@
 
 -export([get_candidate/0]).
 
--export([start_generation/0, stop_generation/0]).
+-export([ get_generation_state/0
+        , start_generation/0
+        , stop_generation/0
+        ]).
 
 -export([prep_stop/0]).
 
@@ -49,6 +52,10 @@ start_generation() ->
 stop_generation() ->
     gen_server:cast(?MODULE, stop_generation).
 
+-spec get_generation_state() -> 'running' | 'stopped'.
+get_generation_state() ->
+    gen_server:call(?MODULE, get_generation_state).
+
 prep_stop() ->
     gen_server:call(?MODULE, prep_stop).
 
@@ -64,6 +71,12 @@ handle_call(get_candidate, _From, State = #state{ candidate = undefined }) ->
     {reply, {error, no_candidate}, State};
 handle_call(get_candidate, _From, State = #state{ candidate = Candidate }) ->
     {reply, {ok, Candidate}, State};
+handle_call(get_generation_state, _From, State = #state{ generating = IsGenerating }) ->
+   Reply = case IsGenerating of
+        true -> running;
+        false -> stopped
+    end,
+    {reply, Reply, State};
 handle_call(prep_stop, _From, State) ->
     {reply, ok, do_stop_generation(State)};
 handle_call(Req, _From, State) ->
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -49,7 +49,7 @@
         , get_mining_workers/0
         , start_mining/0
         , stop_mining/0
-        , handoff_leader/0
+        , is_leader/0
         , get_beneficiary/0
         ]).
 
@@ -122,9 +122,9 @@ get_mining_state() ->
 get_mining_workers() ->
     gen_server:call(?SERVER, get_mining_workers).
 
--spec handoff_leader() -> 'ok'.
-handoff_leader() ->
-    gen_server:call(?SERVER, handoff_leader).
+-spec is_leader() -> boolean().
+is_leader() ->
+    gen_server:call(?SERVER, is_leader).
 
 %%%===================================================================
 %%% Chain API
@@ -244,6 +244,8 @@ handle_call(get_mining_state,_From, State) ->
     {reply, State#state.mining_state, State};
 handle_call(get_mining_workers, _From, State) ->
     {reply, worker_pids_by_tag(mining, State), State};
+handle_call(is_leader, _From, State = #state{ consensus = Cons }) ->
+    {reply, Cons#consensus.leader, State};
 handle_call(reinit_chain, _From, State1 = #state{ consensus = Cons }) ->
     %% NOTE: ONLY FOR TEST
     ok = reinit_chain_state(),
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -947,7 +947,8 @@ setup_loop(State = #state{ consensus = Cons }, RestartMining, IsLeader, Origin) 
     State1 = State#state{ consensus = Cons#consensus{ leader = IsLeader } },
     State2 =
         case Origin of
-            block_created when IsLeader ->
+            Origin when IsLeader, Origin =:= block_created
+                           orelse Origin =:= block_received ->
                 aec_block_generator:start_generation(),
                 start_micro_signing(State1);
             block_received when not IsLeader ->
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -228,6 +228,7 @@ handle_call({post_block, Block},_From, State) ->
 handle_call(stop_mining,_From, State = #state{ consensus = Cons }) ->
     epoch_mining:info("Mining stopped"),
     [ aec_tx_pool:garbage_collect() || is_record(Cons, consensus) andalso Cons#consensus.leader ],
+    aec_block_generator:stop_generation(),
     State1 = kill_all_workers(State),
     State2 = State1#state{mining_state = 'stopped',
                           key_block_candidate = undefined},
--- a/apps/aecore/test/aec_conductor_tests.erl
+++ b/apps/aecore/test/aec_conductor_tests.erl
@@ -491,7 +491,7 @@ beneficiary_keys() ->
     {Pub, Priv}.
 
 preset_accounts({Pub, _}) -> preset_accounts(Pub);
-preset_accounts(Pub) -> [{Pub, 100}].
+preset_accounts(Pub) -> [{Pub, 50000}].
 
 tx({Pub, Priv}) ->
     #{ public := RPub } = enacl:sign_keypair(),
@@ -499,7 +499,7 @@ tx({Pub, Priv}) ->
                                   recipient_id => aec_id:create(account, RPub),
                                   amount => 1,
                                   nonce => 1,
-                                  fee => 1,
+                                  fee => 20000,
                                   ttl => 0,
                                   payload => <<"">>}),
     aec_test_utils:sign_tx(Tx, Priv).
--- a/apps/aecore/test/aec_conductor_tests.erl
+++ b/apps/aecore/test/aec_conductor_tests.erl
@@ -79,7 +79,8 @@ miner_test_() ->
              teardown_common(TmpKeysDir)
      end,
      [{"Stop and restart miner", fun test_stop_restart/0},
-      {"Test consecutive start/stop ", fun test_stop_restart_seq/0}
+      {"Test consecutive start/stop ", fun test_stop_restart_seq/0},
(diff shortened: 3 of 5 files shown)


hanssv:
Needs another rebase, tests were broken for a while on master :-(

apps/aecore/src/aec_block_generator.erl
```diff
@@ -11,7 +11,10 @@
 
 -export([get_candidate/0]).
 
--export([start_generation/0, stop_generation/0]).
+-export([ get_generation_state/0
```
hanssv: Should this be a test only exported function?

apps/aecore/src/aec_conductor.erl
```diff
@@ -228,6 +228,7 @@ handle_call({post_block, Block},_From, State) ->
 handle_call(stop_mining,_From, State = #state{ consensus = Cons }) ->
     epoch_mining:info("Mining stopped"),
     [ aec_tx_pool:garbage_collect() || is_record(Cons, consensus) andalso Cons#consensus.leader ],
+    aec_block_generator:stop_generation(),
```
hanssv: Yay, I noticed while analyzing some test logs the other day that we kept generating blocks after shutting down mining - good fix!

fabiankrol: 👌

apps/aecore/src/aec_block_generator.erl
fabiankrol: Yes, it might be. i didn't do it because it corresponds to `aec_conductor:get_mining_status/0` which is not hidden.

I don't have strong opinion on that, I can hide it if requested

fabiankrol:
[Rebased](https://youtu.be/FGBhQbmPwH8)

apps/aecore/test/aec_conductor_tests.erl
```diff
@@ -373,17 +379,147 @@ test_block_publishing() ->
...
+    MicroBlock = wait_for_micro_block_created(),
+    {ok, KeyHash1} = aec_blocks:hash_internal_representation(KeyBlock1),
+    {ok, KeyHash2} = aec_blocks:hash_internal_representation(KeyBlock2),
+    ?assertEqual(KeyHash1, aec_blocks:prev_key_hash(KeyBlock2)),
+    ?assertEqual(KeyHash2, aec_blocks:prev_key_hash(MicroBlock)),
+    ok.
+
+test_received_block_signing() ->
+    Keys = beneficiary_keys(),
+    meck:new(aec_mining, [passthrough]),
```
lucafavatella: If test fails after this line but before the corresponding unload, does any teardown remove this mock? if not, this would affect validity of results of following tests.

apps/aecore/test/aec_conductor_tests.erl
```diff
@@ -373,17 +379,147 @@ test_block_publishing() ->
...
+    assert_generation_state(stopped),
+    KeyBlock = wait_for_block_created(),
+    assert_leader(true),
+    assert_generation_state(running),
+
+    ok = aec_tx_pool:push(tx(Keys)),
+
+    MicroBlock = wait_for_micro_block_created(),
+    {ok, KeyHash} = aec_blocks:hash_internal_representation(KeyBlock),
+    ?assertEqual(KeyHash, aec_blocks:prev_key_hash(MicroBlock)),
```
lucafavatella: Who guarantees that no new keyblock after `KeyBlock` but before `MicroBlock` was mined? This seems a test potentially failing intermittently.

apps/aecore/test/aec_conductor_tests.erl
```diff
@@ -373,17 +379,147 @@ test_block_publishing() ->
...
+    assert_leader(true),
+    assert_generation_state(running),
+
+    ok = aec_tx_pool:push(tx(Keys)),
+
+    MicroBlock = wait_for_micro_block_created(),
+    {ok, KeyHash1} = aec_blocks:hash_internal_representation(KeyBlock1),
+    {ok, KeyHash2} = aec_blocks:hash_internal_representation(KeyBlock2),
+    ?assertEqual(KeyHash1, aec_blocks:prev_key_hash(KeyBlock2)),
+    ?assertEqual(KeyHash2, aec_blocks:prev_key_hash(MicroBlock)),
```
lucafavatella: Who guarantees that no new keyblock after `KeyBlock2` but before `MicroBlock` was mined? This seems a test potentially failing intermittently.

apps/aecore/src/aec_conductor.erl
```diff
@@ -947,7 +947,8 @@ setup_loop(State = #state{ consensus = Cons }, RestartMining, IsLeader, Origin)
     State1 = State#state{ consensus = Cons#consensus{ leader = IsLeader } },
     State2 =
         case Origin of
-            block_created when IsLeader ->
+            Origin when IsLeader, Origin =:= block_created
+                           orelse Origin =:= block_received ->
```
lucafavatella: Strictly speaking, out of scope of this PR though still re semantics of this block received from HTTP internal API:
Is this block published to network peers (`block_to_publish`)? Is it meant to? In any case, on the assumption that this key block is a key block that starts a generation of microblocks signed by this node, how are peers meant to get this keyblock when a successor microblock is published (`micro_block_created`)? Do peers get it?

apps/aecore/src/aec_conductor.erl
```diff
@@ -947,7 +947,8 @@ setup_loop(State = #state{ consensus = Cons }, RestartMining, IsLeader, Origin)
...
                 aec_block_generator:start_generation(),
```
lucafavatella: I am checking two things:
* What is effect of calling `aec_block_generator:start_generation()` and what is expected state. (I need to read code.)
* Where code branches on whether this block received from HTTP internal API is the mined version of the block candidate. Specifically:
  * Whether signing pubkey (miner) in received key block is same for which priv key is known.
  * Whether rest of the received block fits with internal state e.g. state (for building microblock on correct state trees), timestamp (for correct timing in emission of microblocks). It feels simply the received key block shall be equal to the known block candidate, or (in a more sophisticated manner) in a pool of known block candidates (so to have the priv key).

I have the impression that this `block_received` tag shall be split in two subtags for the following code paths depending on whether this received key block is meant to start a generation of microblocks.

(I may be missing a check that is already in place, checking.)

apps/aecore/test/aec_conductor_tests.erl
fabiankrol: 🔬👍  I'll move it teardown

apps/aecore/test/aec_conductor_tests.erl
fabiankrol: Correct, i'll update assertions

apps/aecore/test/aec_conductor_tests.erl
fabiankrol: Correct, i'll update assertions

apps/aecore/src/aec_conductor.erl
fabiankrol: * `aec_block_generator:start_generation()` 👍 
* There is no such check and I'm not really sure we need one. Should end user be able to post valid through the api? How different it is from receiving one from sync? Anyway I thing the decision possible changes should be mede in another PR. 
For key blocks `IsLeader` flag is set up based based comparing signing keys and block miner. We're not discarding block with different miner

Re splitting: maybe instead of using `post_block/1` in the api implementation we should go through mining code path: [`handle_mining_reply({{ok, {Nonce, Evd}}, HeaderBin}, #state{} = State) ->`](https://github.com/aeternity/epoch/blob/PT-161715282-post-key_block/apps/aecore/src/aec_conductor.erl#L688)

fabiankrol: > There is no such check and I'm not really sure we need one

I'm going to confirm it with original PT story for this endpoint

cc @jur0

fabiankrol:
Putting `WIP` while resolving test issues pointed out by @lucafavatella

apps/aecore/src/aec_conductor.erl
lucafavatella: `aec_keys` keeps (alongside peer keypair):
* (A) The keypair currently used for signing keyblocks (`#worker_state.sign_priv` - see `aec_keys:sign_micro_block/1`).
* (B) The keypair meant to be used in key block candidate, for miner field of key block candidate (`#worker_state.candidate_pub` - see `aec_keys:candidate_pubkey/`).

[**If** key block candidate mined **and** it becomes top](https://github.com/aeternity/epoch/blob/8d69cfd2/apps/aecore/src/aec_conductor.erl#L536), candidate keypair (B) is promoted to sign keypair (A) (hence following microblocks signed with promoted keypair). Beware that conditional is combined effect of:
* [`aec_keys:promote_candidate/1` matching specified pubkey against known candidate pubkey (notice two function clauses), and returning value (as opposed to terminating)](https://github.com/aeternity/epoch/blob/8d69cfd2/apps/aecore/src/aec_keys.erl#L386-L392).
* [Call not checking return value](https://github.com/aeternity/epoch/blob/8d69cfd2/apps/aecore/src/aec_conductor.erl#L536).

So there it looks to me that there is a check - I need to understand effect. (I am considering whether only test enhancement needed.)

apps/aecore/test/aec_conductor_tests.erl
```diff
@@ -373,17 +379,147 @@ test_block_publishing() ->
...
+
+    %% Single tx should trigger micro block
+    ok = aec_tx_pool:push(tx(Keys)),
+
+    meck:unload(aec_mining),
+
+    NewBlock = wait_for_block_to_publish(),
+    ?assertEqual(false, aec_blocks:is_key_block(NewBlock)),
+    {ok, KB1Hash} = aec_blocks:hash_internal_representation(KB1),
+    ?assertEqual(KB1Hash, aec_blocks:prev_key_hash(NewBlock)),
```
lucafavatella: It seems to me that the test shall be integrated with assertions that:
* The signed microblock `NewBlock` is inserted in the chain.
* The preceding posted keyblock `KB1` is inserted in the chain.

It may be constrained by how the code is written, though from a test perspective it is an independent check that would ease readability of test (unless test is explicitly not meant to store blocks in chain for some reason?). The benefit that I have in mind is the following checks having passed on these blocks:
* [Microblock signature verifiable according to pubkey in prev key block](https://github.com/aeternity/epoch/blob/8d69cfd2/apps/aecore/src/aec_chain_state.erl#L429).
* [Connection to genesis](https://github.com/aeternity/epoch/blob/8d69cfd2/apps/aecore/src/aec_chain_state.erl#L419).

apps/aecore/src/aec_conductor.erl
lucafavatella: > **If** key block candidate mined (EDIT: regardless of the origin, i.e. regardless of whether it mined by node or posted to node from HTTP internal API) **and** it becomes top

In same case, [conductor triggers computation of new key block candidate](https://github.com/aeternity/epoch/blob/8d69cfd2/apps/aecore/src/aec_conductor.erl#L538) after [having updated `top_block_hash` in state](https://github.com/aeternity/epoch/blob/8d69cfd2/apps/aecore/src/aec_conductor.erl#L518) that later [passes it to `aec_block_key_candidate:create`
https://github.com/aeternity/epoch/blob/8d69cfd2/apps/aecore/src/aec_conductor.erl#L804) (that fetches block hash from DB).

So key block candidate generation seems fine. I need to check the microblock generation.

fabiankrol:
@lucafavatella @velzevur I'm going to merge this PR to avoid neverending conflicts. However i'll open a fallow up PR to address comments.

apps/aecore/test/aec_conductor_tests.erl
lucafavatella: I see addressed in #1765

apps/aecore/test/aec_conductor_tests.erl
lucafavatella: I see addressed in #1765

PR #1744, opened 2018-11-07
Use infrastructure Docker image.

valkov-ivaylo:
PT-161478102

Requires aeternity/infrastructure#147

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,6 @@ references:
     environment:
       # CircleCI supported versions: https://circleci.com/docs/2.0/building-docker-images/#docker-version
       # Official docker versions: https://download.docker.com/linux/static/stable/x86_64/
-      DOCKER_CLIENT_VERSION: "17.09.0-ce"
       DOCKERHUB_REPO: aeternity/epoch
       S3_BUILDS_BUCKET: aeternity-epoch-builds
       S3_RELEASES_BUCKET: aeternity-epoch-releases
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -38,15 +38,6 @@ references:
         sudo usermod -aG docker ubuntu
         sudo chown -R ubuntu:ubuntu ~/epoch
 
-  install_docker_client: &install_docker_client
-    run:
-      name: Install Docker CLI tools
-      command: |
-        curl -L -o /tmp/docker-${DOCKER_CLIENT_VERSION:?}.tgz \
```
lucafavatella: Is `DOCKER_CLIENT_VERSION: "17.09.0-ce"` in this file now unused and to be removed?

.circleci/config.yml
```diff
@@ -504,18 +495,17 @@ jobs:
       - *fail_notification
 
   docker_push_tag:
-    <<: *container_config
+    <<: *infrastructure_config
```
lucafavatella: Requires new Docker image published. Ok.

.circleci/config.yml
valkov-ivaylo: It might be removed. I did not find any other references it in in the .circleci/config.yml. I was not sure. @dincho ?

dincho: I don't see other references either, let's remove it

valkov-ivaylo: Fixed in 45e3b44

PR #1746, opened 2018-11-07
Pt 161747758 nested abort

gorillainduction:
https://www.pivotaltracker.com/story/show/161747758

### DIFF
--- a/apps/aehttp/test/aehttp_contracts_SUITE.erl
+++ b/apps/aehttp/test/aehttp_contracts_SUITE.erl
@@ -207,7 +207,7 @@ abort_test_contract(Config) ->
     ICode = compile_test_contract("abort_test_int"),
 
     %% Create chain of int contracts to test contract.
-    {_EncodedTestPub,DecodedTestPub,_} =
+    {EncodedTestPub,DecodedTestPub,_} =
         create_compute_contract(Node, APub, APriv, TCode, <<"(42)">>),
     {_EncodedInt1Pub,DecodedInt1Pub,_} =
         create_compute_contract(Node, APub, APriv, ICode,
@@ -232,19 +232,30 @@ abort_test_contract(Config) ->
     %% Do abort where we have put values 42, 142, 242, 342,
     %% then check value in abort_test contract, should still be 3017!
     %% Check that we get the abort string back.
-
+    %% First check when we call the full contract call stack.
     AbortString = <<"yogi bear">>,
     GivenGas = 100000,
-    {RevertValue, Return} =
+    {RevertValue1, Return1} =
         revert_call_compute_func(Node, APub, APriv, EncodedInt3Pub,
                                  <<"do_abort">>,
                                  args_to_binary([42,{string,AbortString}]),
                                  #{gas => GivenGas}
                                 ),
-    #{<<"value">> := AbortString} = decode_data(<<"string">>, RevertValue),
-    #{<<"gas_used">> := GasUsed} = Return,
-    %% Make sure we actually saved some gas by doing the revert
-    ?assert(GasUsed < GivenGas),
+    #{<<"value">> := AbortString} = decode_data(<<"string">>, RevertValue1),
+    #{<<"gas_used">> := GasUsed1} = Return1,
+    ?assert(GasUsed1 < GivenGas), %% Make sure we actually saved some gas
+
+    %% Also check that we get the same behaviour when calling directly to the
+    %% leaf contract.
+    {RevertValue2, Return2} =
+        revert_call_compute_func(Node, APub, APriv, EncodedTestPub,
+                                 <<"do_abort">>,
+                                 args_to_binary([42,{string,AbortString}]),
+                                 #{gas => GivenGas}
+                                ),
+    #{<<"value">> := AbortString} = decode_data(<<"string">>, RevertValue2),
+    #{<<"gas_used">> := GasUsed2} = Return2,
+    ?assert(GasUsed2 < GivenGas), %% Make sure we actually saved some gas
 
     ABal1 = get_balance(APub),
 
--- a/apps/aehttp/test/aehttp_contracts_SUITE.erl
+++ b/apps/aehttp/test/aehttp_contracts_SUITE.erl
@@ -207,7 +207,7 @@ abort_test_contract(Config) ->
     ICode = compile_test_contract("abort_test_int"),
 
     %% Create chain of int contracts to test contract.
-    {EncodedTestPub,DecodedTestPub,_} =
+    {_EncodedTestPub,DecodedTestPub,_} =
         create_compute_contract(Node, APub, APriv, TCode, <<"(42)">>),
     {_EncodedInt1Pub,DecodedInt1Pub,_} =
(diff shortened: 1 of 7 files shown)


apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -234,11 +234,17 @@ abort_test_contract(Config) ->
     %% Check that we get the abort string back.
 
     AbortString = <<"yogi bear">>,
-    {RevertValue,_} =
-        revert_call_compute_func(Node, APub, APriv, EncodedTestPub,
+    GivenGas = 100000,
+    {RevertValue, Return} =
+        revert_call_compute_func(Node, APub, APriv, EncodedInt3Pub,
```
lucafavatella: I see this is calling abort_test_int that does not abort directly though in nested call. Ok.

apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -234,11 +234,17 @@ abort_test_contract(Config) ->
     %% Check that we get the abort string back.
 
     AbortString = <<"yogi bear">>,
-    {RevertValue,_} =
-        revert_call_compute_func(Node, APub, APriv, EncodedTestPub,
```
lucafavatella: I see this replaces the old test on outer revert with nested revert - rather than adding it. Is the outer revert from HTTP API tested in other place?

apps/aevm/src/aevm_eeevm_state.erl
```diff
@@ -237,10 +237,11 @@ do_revert(Us0, Us1, State0) ->
     case vm_version(State0) of
         ?AEVM_01_Sophia_01 ->
             try
-		%% Get the string then convert it to output binary format.
-                {String, State1} = aevm_eeevm_memory:get_area(Us0, Us1, State0),
```
lucafavatella: Reviewing this line before the change:
* This line considers unallocated words in memory as well defined as 0. That is good.
* This expands memory as needed, and accounts that cost on the REVERT instruction. That is good.

apps/aevm/src/aevm_eeevm_state.erl
```diff
@@ -237,10 +237,11 @@ do_revert(Us0, Us1, State0) ->
...
-		Out = aeso_data:to_binary(String, 0),
```
lucafavatella: Reviewing this line before the change:
* This line cannot fail, right?

apps/aevm/src/aevm_eeevm_state.erl
```diff
@@ -237,10 +237,11 @@ do_revert(Us0, Us1, State0) ->
...
-                set_out(Out, State1)
+                %% In Sophia Us1 is a pointer to the actual value.
+                %% The type of the value is always string.
+                HeapValue = aeso_data:heap_value(maps(State0), Us1, mem(State0)),
```
lucafavatella: Reviewing this line after the change:
* This line cannot fail, because it fetching mem and run a constructor. Ok.

apps/aevm/src/aevm_eeevm_state.erl
```diff
@@ -237,10 +237,11 @@ do_revert(Us0, Us1, State0) ->
...
+                {ok, Out} = aeso_data:heap_to_binary(string, get_store(State0), HeapValue),
```
lucafavatella: I am not convinced by this change:
* Can [`aeso_data:heap_to_binary`](https://github.com/aeternity/epoch/blob/21a4be0096fc87bc7d2f0f49e9918ec085a84629/apps/aevm/src/aevm_eeevm_state.erl#L260) return something different from `{ok, _}` or fail?
* How is pointer `Us1` looked up in memory? Is the word at that address considered as 0 if unallocated? (Not sure what to expect.)
* Does this modified semantics allow charging cost of REVERT before performing operation? It seems no to me.

apps/aesophia/src/aeso_ast_to_icode.erl
```diff
@@ -703,11 +703,8 @@ builtin_function(abort) ->
     A = fun(X) -> aeb_opcodes:mnemonic(X) end,
     {{builtin, abort}, [private],
      [{"s", string}],
-     %% length, address, REVERT
-     {inline_asm, [A(?DUP1),
-                   A(?MLOAD),A(?SWAP1),         %Load size and swap
-                   A(?PUSH1),32,A(?ADD),        %Load address of string binary
```
lucafavatella: Trying to understand whether change is improvement:
Maybe this line (before change) was not handling [zero-padded strings](https://github.com/aeternity/protocol/blame/d3126f7ec91067bd10a35c8c02004645b51c95c0/contracts/sophia.md#L925)?

apps/aevm/src/aevm_eeevm_state.erl
UlfNorell: You could look up the length of the string before converting to binary and check/charge gas based on the length.

gorillainduction: The new implementation has exactly the same semantics as before, with the difference that the steps are not expressed in the byte code.

Before, the byte code followed the pointer, found out the size, and took the proper offset/size from memory. Now, it follows the pointer and grabs the memory in the vm instead.

For Sophia, this difference in approach is already used for getting arguments and return values for contract calls/primops.

I just wanted to align the way stuff is happening for Sophia, and as a side effect remove a few byte code instructions. When reading the code, I thought there was a mistake when I first read the code, since few things operate this way in Sophia, so I wanted future readers to not have to do this. This apparently failed ;-)

No big deal, so I can drop the optimization if it feels insecure.

apps/aesophia/src/aeso_ast_to_icode.erl
gorillainduction: The improvement is that it uses fewer byte code instructions. Nothing else.

apps/aehttp/test/aehttp_contracts_SUITE.erl
gorillainduction: I can add it back. It was a mistake in the old code, but it makes sense to test both entrypoints.

gorillainduction:
Rebased and added test for a top level call to abort.

apps/aehttp/test/aehttp_contracts_SUITE.erl
lucafavatella: I see addressed in 53dc639ea27e891bc2261d081826bf2303ae4720

apps/aevm/src/aevm_eeevm_state.erl
lucafavatella: > ...
> For Sophia, this difference in approach is already used for getting arguments and return values for contract calls/primops.
> 
> I just wanted to align the way stuff is happening for Sophia ...

Thanks for providing more context. I did not check the other usages (I see related changes in recent master).

lucafavatella: > You could look up the length of the string before converting to binary and check/charge gas based on the length.

@UlfNorell Good point. Affecting consensus though I expect.

PR #1749, opened 2018-11-07
PT-160254577 Store messages from contributors in the genesis block

zp-sd:
* Store messages file at data/aecore/.genesis/messages.json
* Add messages_hash script, that computes hash of the messages from messages.json file
* Put computed hash in aec_governance:contributors_messages_hash/0, that is used in prev_hash field of a genesis block


zp-sd:
Not a WIP anymore, apart from intermittent failure of the escript (Segfault 11 on `enacl:generichash_final(HashState)`), which can be taken care of separately.

PR #1748, opened 2018-11-07
PT-161750304 upstream sync cuckoo

hanssv:
Do an upstream sync with tromp/cuckoo.git
    
    Change naming of miners to lean<edge_bits> instead of lean<node_bits>
    to align better with upstream (and cause less confusion).

PR #1752, opened 2018-11-08
Add cuckoo section to example epoch configuration file to use lean30 …

baltar:
…to make it work with default docker container memory limit of 2GB. Also added a note to increase the memory limit to 4GB if one wants to use the default miner.

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -39,7 +39,7 @@ mining:
         miner:
             executable: lean29-generic
             extra_args: ""
-            node_bits: 30
+            edge_bits: 29
 
 ```
 
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -37,7 +37,7 @@ mining:
     beneficiary: "encoded_beneficiary_pubkey_to_be_replaced"
     cuckoo:
         miner:
-            executable: lean30
+            executable: lean29-generic
             extra_args: ""
             node_bits: 30
 
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -43,7 +43,7 @@ mining:
 
 ```
 
-Note that if want to use the default miner, remove the `mining.cuckoo` section and increase the docker container memory at least 4GB. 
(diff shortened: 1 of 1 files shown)


dincho:
Yes, let's delay a bit until we sync with upstream, but we should definitely changes the docker defaults or add docker configuration requirements.

dincho:
@lucafavatella what do you think ?

lucafavatella:
> what do you think ?

@dincho 
* This PR makes sense to me.
* I agree on waiting #1748 to be merged first.
* Re "we should definitely change the docker defaults or add docker configuration requirements.": I did not find further places where the Docker config is specified (apart from the `docs/docker.md` amended in this PR) so I do not see further changes needed (apart from rebase after #1748 merged). Please let me know if I am missing anything or misunderstanding your sentence.

dincho:
> * Re "we should definitely change the docker defaults or add docker configuration requirements.": I did not find further places where the Docker config is specified (apart from the `docs/docker.md` amended in this PR) so I do not see further changes needed (apart from rebase after #1748 merged). Please let me know if I am missing anything or misunderstanding your sentence.

It means: yes, I agree with this PR :)

docs/docker.md
```diff
@@ -35,8 +35,16 @@ chain:
 mining:
     autostart: true
     beneficiary: "encoded_beneficiary_pubkey_to_be_replaced"
+    cuckoo:
+        miner:
+            executable: lean30
```
hanssv:
```suggestion
            executable: lean29-generic
```

docs/docker.md
```diff
@@ -35,8 +35,16 @@ chain:
...
+            extra_args: ""
+            node_bits: 30
```
hanssv:
```suggestion
            edge_bits: 29
```

docs/docker.md
hanssv: If avx2 is available use `lean29-avx2`

baltar: Since I don't really know about those miners, I'll just apply the suggestion.

PR #1763, opened 2018-11-12
PT-161440511 mempool nonce check

fabiankrol:
[PT-161440511](https://www.pivotaltracker.com/story/show/161440511)

Two configuration options are present: 
`mempool.nonce_offset (default 5)` - when account is present mempool accepts on chain nonce + nonce_offset
`mempool.nonce_baseline (defualt 1)` - value is used when account is not present 

~~Sync uses `aec_tx_pool:push_sync/1` where upper limit is not defined by offset~~

### DIFF
--- a/apps/aeutils/test/data/epoch_full.yaml
+++ b/apps/aeutils/test/data/epoch_full.yaml
@@ -54,6 +54,8 @@ mempool:
     invalid_tx_ttl: 5
     # Mempool (re-)synchronization interval (in ms)
     sync_interval: 1800000
+    # Maximum nonce offset accepted
+    nonce_offset: 1
     # Maximum nonce accepted when pubkey is not present in state
     nonce_offset_no_account: 1
 
--- a/apps/aeutils/test/data/epoch_full.yaml
+++ b/apps/aeutils/test/data/epoch_full.yaml
@@ -54,6 +54,8 @@ mempool:
     invalid_tx_ttl: 5
     # Mempool (re-)synchronization interval (in ms)
     sync_interval: 1800000
+    # Maximum nonce accepted when pubkey is not present in state
+    nonce_offset_no_account: 1
 
 http:
     cors:
--- a/apps/aeutils/test/data/epoch_full.yaml
+++ b/apps/aeutils/test/data/epoch_full.yaml
@@ -55,9 +55,9 @@ mempool:
     # Mempool (re-)synchronization interval (in ms)
     sync_interval: 1800000
     # Maximum nonce offset accepted
-    nonce_offset: 1
+    nonce_offset: 5
     # Maximum nonce accepted when pubkey is not present in state
-    nonce_offset_no_account: 1
+    nonce_baseline: 1
 
 http:
     cors:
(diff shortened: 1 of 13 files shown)


apps/aecore/src/aec_tx_pool_sync.erl
```diff
@@ -332,7 +332,7 @@ do_local_action(Sync = #sync{ peer_id = PeerId }, {get, TxHashes}) ->
     Fun = fun() ->
               case aec_peer_connection:tx_pool_sync_get(PeerId, TxHashes) of
                   {ok, Txs} ->
-                      [ aec_tx_pool:push(Tx) || Tx <- Txs ],
+                      [ aec_tx_pool:push_sync(Tx) || Tx <- Txs ],
```
hanssv: Is this an attack vector? A malicious node could pollute the mempool of another node via sync?

apps/aeutils/test/data/epoch_full.yaml
```diff
@@ -54,6 +54,10 @@ mempool:
     invalid_tx_ttl: 5
     # Mempool (re-)synchronization interval (in ms)
     sync_interval: 1800000
+    # Maximum nonce offset accepted
+    nonce_offset: 5
```
dincho: Just curious, if this is maximum why the option is not `max_nonce_offset` ?

fabiankrol: no real reason here... imho `nonce_offset` is precise enough 🤷‍♂️

uwiger:
One thing I've been thinking about is that e.g. state channels could benefit from an `auto` nonce parameter, i.e. it's the state channel fsm that picks a nonce, and always tries to pick the next one. Currently, this requires multiple calls to the tx_pool: first to get the latest nonce, and next to push a transaction with the next nonce. It seems like this could be done safer and more efficiently in one call, if the tx_pool server is told to simply pick whatever is the next nonce.

apps/aecore/src/aec_tx_pool_sync.erl
fabiankrol: @hanssv i removed `push_sync/1` function in favour of `push(Tx, tx_received)`. I'm checking nonce offset only for `tx_created` event but jobs is controlling queue for it.

fabiankrol: We're starting to sync mempool without synced chain. Because of that we can end up with huge nonce deltas when account is not yet present in state. I created a ticket for it: [#162147646: Wait for mempool sync until chain sync is done](https://www.pivotaltracker.com/story/show/162147646)

fabiankrol:
@uwiger tracked in PT: [#162147732: Allow mempool to autofill next nonce ](https://www.pivotaltracker.com/story/show/162147732)

fabiankrol:
I'm going to rebase after #1812 is merged

fabiankrol:
As discussed offline: checking nonce is not enough to stop mempool flooding attack. We should address this issue in a separate ticket.

PR #1722, opened 2018-11-01
PT-161183355 Use tx gas to check minimal tx fee

jur0:
[PT-161183355](https://www.pivotaltracker.com/n/projects/2124891/stories/161183355)

The tx gas is now used to compute the minimal tx fee (min_fee = gas * min_gas_price)

TODO:
- [x] add minimal tx fee check into micro block validation
- [x] python UATs
- [x] integration tests 
- [x] system tests

PR #1768, opened 2018-11-13
Fix minimal tx fee validation

jur0:
Related to #1722 and [PT-161183355](https://www.pivotaltracker.com/n/projects/2124891/stories/161183355).

The check of minimal tx fee is moved from `aec_blocks.erl` to `aetx.erl`.

### DIFF
--- a/apps/aecore/src/aec_blocks.erl
+++ b/apps/aecore/src/aec_blocks.erl
@@ -420,7 +420,6 @@ validate_key_block(#key_block{} = Block) ->
 validate_micro_block(#mic_block{} = Block) ->
     Validators = [fun validate_txs_hash/1,
                   fun validate_gas_limit/1,
-                  fun validate_txs_fee/1,
                   fun validate_pof/1
                  ],
     case aec_headers:validate_micro_block_header(to_micro_header(Block)) of
@@ -448,16 +447,6 @@ validate_gas_limit(#mic_block{} = Block) ->
     case gas(Block) =< aec_governance:block_gas_limit() of
         true  -> ok;
         false -> {error, gas_limit_exceeded}
-    end.
-
--spec validate_txs_fee(block()) -> ok | {error, invalid_minimal_tx_fee}.
-validate_txs_fee(#mic_block{txs = STxs}) ->
-    case lists:all(fun(STx) ->
-                           Tx = aetx_sign:tx(STx),
-                           aetx:fee(Tx) >= (aetx:min_gas(Tx) * aec_governance:minimum_gas_price())
-                   end, STxs) of
-        true -> ok;
-        false -> {error, invalid_minimal_tx_fee}
     end.
 
 validate_pof(#mic_block{pof = no_fraud}) -> ok;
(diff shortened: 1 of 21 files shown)


apps/aetx/src/aetx.erl
```diff
@@ -249,7 +249,7 @@ check_tx(#aetx{ cb = CB, tx = Tx } = AeTx, Trees, Env) ->
     end.
 
 check_minimum_fee(AeTx) ->
-    case fee(AeTx) >= aec_governance:minimum_tx_fee() of
+    case fee(AeTx) >= (min_gas(AeTx) * aec_governance:minimum_gas_price()) of
```
lucafavatella: I see this brings this in line with `aec_tx_pool:check_minimum_fee/2` (i.e. `aetx:fee(Tx) >= (aetx:min_gas(Tx) * aec_governance:minimum_gas_price())`). Ok.

apps/aecore/src/aec_blocks.erl
```diff
@@ -420,7 +420,6 @@ validate_key_block(#key_block{} = Block) ->
 validate_micro_block(#mic_block{} = Block) ->
     Validators = [fun validate_txs_hash/1,
                   fun validate_gas_limit/1,
```
lucafavatella: Out of scope for this PR, looks in scope of ["Hook gas limit per microblock to gas for all transactions"](https://www.pivotaltracker.com/story/show/160392862):
* The check on microblock gas limit may need review - at the moment for e.g. call tx the gas limit is considered so a call tx with a large gas limit though little gas used seems exhausting space for txs in microblock while spending only the gas used.
* The conductor, in the DB transaction performing both block insertion and checks, [does not perform checks against block gas limit](https://github.com/aeternity/epoch/blob/e0bfd9e4901a0c1a98c89c172f54f9d726475e35/apps/aecore/src/aec_trees.erl#L417-L418). That check seems performed before calling transaction though (1) I would have expected check inside DB transaction and (2) I have not checked all code paths for that.

apps/aecore/src/aec_blocks.erl
```diff
@@ -420,7 +420,6 @@ validate_key_block(#key_block{} = Block) ->
...
-                  fun validate_txs_fee/1,
```
lucafavatella: I see that the `aec_conductor` inside the DB transaction attempting to insert microblock in DB [checks this by relying on `aetx:check`](https://github.com/aeternity/epoch/blob/e0bfd9e4901a0c1a98c89c172f54f9d726475e35/apps/aecore/src/aec_chain_state.erl#L816). So block invalid for low fee is not inserted in DB anyway, that is good.

I see that this check function used to be run in the process performing the sync network operation (see `aec_conductor:add_synced_generation_batched`) while now the first time this check is run on the microblock is inside the `aec_conductor` process that means this may lead to network peer potential giving more load to `aec_conductor` process. (Similar code path applies for user HTTP API `aec_conductor:post_block` though that is meant to be internal so less of a concern.)

apps/aeoracle/src/aeo_register_tx.erl
```diff
@@ -145,7 +145,7 @@ check(#oracle_register_tx{nonce = Nonce, oracle_ttl = OTTL, fee = Fee,
                %% Contract is paying tx fee as gas.
                aetx_contract -> [];
                aetx_transaction ->
-                   [fun() -> aeo_utils:check_ttl_fee(Height, OTTL, Fee - aec_governance:minimum_tx_fee()) end
+                   [fun() -> aeo_utils:check_ttl_fee(Height, OTTL, Fee) end
```
lucafavatella: This, and other check_ttl_fee invocations, will need review in ["Handle minimum (coin) fee of transactions"](https://www.pivotaltracker.com/story/show/161093573).

PR #1769, opened 2018-11-13
S3 secrets Vault integration

dincho:
Run the scripts on infrastructure docker image.

- [Dummy CircleCI run](https://circleci.com/workflow-run/1598192c-f26c-4d5e-b564-89e39a1d259d)
- [Dummy release](https://github.com/aeternity/epoch/releases/tag/assets-test-master)

https://www.pivotaltracker.com/story/show/160452961

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,13 +11,14 @@ references:
       # Official docker versions: https://download.docker.com/linux/static/stable/x86_64/
       DOCKER_CLIENT_VERSION: "17.09.0-ce"
       DOCKERHUB_REPO: aeternity/epoch
-      S3_BUILDS_BUCKET: aeternity-epoch-builds
-      S3_RELEASES_BUCKET: aeternity-epoch-releases
 
   infrastructure_config: &infrastructure_config
     docker:
       - image: aeternity/infrastructure
     working_directory: ~/epoch
+    environment:
+      S3_BUILDS_BUCKET: aeternity-epoch-builds
+      S3_RELEASES_BUCKET: aeternity-epoch-releases
 
   machine_config: &machine_config
     machine:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -758,6 +774,13 @@ workflows:
             tags:
               only: /^v.*$/
 
+      - upload_build_artifacts:
+          requires:
+            - linux_package
+          filters:
```
lucafavatella: Could you please remind me what this filter means?

Before this change it was only on master. Now it looks like it is on:
* Tags matching `/^v.*$/`
* **All** branches? (I am not sure what default for branches is, and whether default branches is documented in Circle CI doc - [I did not find it](https://circleci.com/docs/2.0/configuration-reference/#filters-1).)

dincho: Yes, all branches. If there is no branch filters, it runs for all branches.
it's documented in the docs you refer:
> If neither only nor ignore are specified then all branches will run the job.

> Before this change it was only on master. Now it looks like it is on:

This PR should not change the behaviour. However, the build artefact are uploaded for each branch build and on all release tags. Also the job itself have an extra check that uploads the "latest" build artefact on master only.

lucafavatella: I see, I was reading incorrectly, it does not change semantics, correct.

----

I see this snippet is kept:
```
            aws s3 cp --acl public-read ${PACKAGE_TARBALL:?} s3://${S3_BUILDS_BUCKET:?}/
            if [ -n "$CIRCLE_BRANCH" -a "$CIRCLE_BRANCH" = "master" ]; then
              aws s3 cp --acl public-read ${PACKAGE_TARBALL:?} s3://${S3_BUILDS_BUCKET:?}/epoch-latest-${PKG_SUFFIX:?}.tar.gz
            fi
```

PR #1774, opened 2018-11-14
PT-161915745 fix serialize exploits

UlfNorell:
In the process of fixing this I also fixed a serious problem where errors saving the store were silently ignored. Now the contract call fails if the store can't be saved. As a consequence the `/debug/contracts/code/call` endpoint no longer works with the `sophia` abi (where you submit the bytecode of the contract to call). The reason is that the contract state isn't available: it's not supplied by the user and you're not calling a contract on the chain. This previously worked (purely accidentally) on functions that didn't look at the state, since the crash trying to write the bogus state was ignored.

Properly fixing this endpoint is out of scope of this ticket, but it's something I expect the apps guys might want to have. Some design work is needed to figure out how to handle state for this though.

### DIFF
--- a/apps/aecontract/test/aecontract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_SUITE.erl
@@ -36,6 +36,7 @@
         , sophia_spend/1
         , sophia_typed_calls/1
         , sophia_no_reentrant/1
+        , sophia_exploits/1
         , sophia_oracles/1
         , sophia_oracles_interact_with_no_vm_oracle/1
         , sophia_oracles_ttl__extend_after_expiry/1
@@ -153,6 +154,7 @@ groups() ->
                                  sophia_match_bug,
                                  sophia_spend,
                                  sophia_typed_calls,
+                                 sophia_exploits,
                                  sophia_oracles,
                                  {group, sophia_oracles_ttl},
                                  {group, sophia_oracles_type_error},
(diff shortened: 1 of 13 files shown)


apps/aevm/src/aevm_eeevm_state.erl
```diff
@@ -239,9 +235,8 @@ do_revert(Us0, Us1, State0) ->
             try
                 %% In Sophia Us1 is a pointer to the actual value.
                 %% The type of the value is always string.
-                HeapValue = aeso_data:heap_value(maps(State0), Us1, mem(State0)),
-                {ok, Out} = aeso_data:heap_to_binary(string, get_store(State0), HeapValue),
-                set_out(Out, State0)
+                {ok, Out, GasUsed} = heap_to_binary(string, Us1, State0),
```
gorillainduction: Why not get rid of the `try...catch` here as well, and simply use a `case` switch?

UlfNorell: I wanted to leave something for the "review use of `try...catch`"-ticket..? We should though.

apps/aevm/src/aevm_ae_primops.erl
```diff
@@ -536,32 +536,50 @@ map_call_get(Data, State) ->
...
+            Res = case aevm_eeevm_maps:get(MapId, KeyBin, State) of
+                    false -> aeso_data:to_binary(none);
+                    <<ValPtr:256, ValBin/binary>> ->
+                        %% Some hacky juggling to build an option value.
+                        NewPtr = 32 + byte_size(ValBin),
+                        <<NewPtr:256, ValBin/binary, 1:256, ValPtr:256>>
+                  end,
+            {ok, {ok, Res}, GasUsed, State};
+        {error, _} = Err ->
+            {ok, Err, aevm_eeevm_state:gas(State), State}
```
lucafavatella: I am not convinced about accessing gas in VM state here. `Gas` first argument of `aevm_ae_primops:call/4` is not the same as VM gas - it is capped and with stipend (and may be different depending on value operand of call). Checking.

lucafavatella: I see [even if computation led to negative number no error is caused](https://github.com/aeternity/epoch/blob/c4742b1697ab7d1ec18a85ed4ab03a1f795605aa/apps/aevm/src/aevm_eeevm.erl#L1508) and [gas is set to zero anyway because of `{error, _}`](https://github.com/aeternity/epoch/blob/c4742b1697ab7d1ec18a85ed4ab03a1f795605aa/apps/aevm/src/aevm_eeevm.erl#L1531), so this has no functional impact - only decreased readability. Ok.

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -852,6 +863,40 @@ sophia_identity(_Cfg) ->
     77    = ?call(call_contract,   Acc1, RemC2, staged_call, word, {IdC, RemC, 77}),
     ok.
 
+sophia_exploits(_Cfg) ->
```
lucafavatella: I tested locally that this commit with test without the related code changes makes the test fail (took veeeeery long time). Ok.

PR #1779, opened 2018-11-15
Compute coinbase at height according to inflation target

gorillainduction:
The module aec_coinbase.erl is generated, but checked in to avoid rounding errors in generation

https://www.pivotaltracker.com/story/show/158105175

- [x] Confirm that the inflation curve and slow start is approved by management
- [x] Confirm the initial token supply


apps/aecore/src/aec_governance.erl
```diff
@@ -90,11 +90,14 @@ expected_block_mine_rate() ->
     aeu_env:user_config_or_env([<<"mining">>, <<"expected_mine_rate">>],
                                aecore, expected_mine_rate, ?EXPECTED_BLOCK_MINE_RATE).
 
+block_mine_reward(0) ->
```
lucafavatella: `0` -> `?GENESIS_HEIGHT`?

apps/aecore/src/aec_governance.erl
```diff
@@ -90,11 +90,14 @@ expected_block_mine_rate() ->
...
+    %% No mining reward for the genesis block
+    0;
+block_mine_reward(Height) when is_integer(Height), Height > 0 ->
```
lucafavatella: `0` -> `?GENESIS_HEIGHT`?

gorillainduction:
Updated to new curve, rebased, and squashed.

apps/aecore/src/aec_governance.erl
gorillainduction: I guess this would make sense, but this code is pretty much hard wired to having genesis at 0, so I think it is better to keep it explicit.

sennui:
It doesn't support constant token supply, right?

gorillainduction:
> It doesn't support constant token supply, right?

No, that comes later.

PR #1792, opened 2018-11-16
Align oracle response tx fee check to other txs

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/162010753

Depends on #1789

Please refer to ticket and commit messages for details.

TODO:
* [x] Ensure CI passes
* [x] Once PR #1789 merged, edit base branch to master.
* [ ] ~Once PR #1796 merged, rebase on master.~ Merging this first as discussed cc @jur0

PR #1796, opened 2018-11-19
PT-161093573 Add ttl state gas to oracle txs

jur0:
[PT-161093573](https://www.pivotaltracker.com/n/projects/2124891/stories/161093573)

Oracle primops had TTL state gas, but it was missing in oracle transactions. This PR fixes it.

### DIFF
--- a/apps/aecore/src/aec_block_micro_candidate.erl
+++ b/apps/aecore/src/aec_block_micro_candidate.erl
@@ -210,7 +210,7 @@ add_txs_to_trees(MaxGas, Trees, Txs, Env) ->
 add_txs_to_trees(_MaxGas, Trees, [], Acc,_Env) ->
     {lists:reverse(Acc), Trees};
 add_txs_to_trees(MaxGas, Trees, [Tx | Txs], Acc, Env) ->
-    TxGas = aetx:gas(aetx_sign:tx(Tx)),
+    TxGas = aetx:gas(aetx_sign:tx(Tx), aetx_env:height(Env)),
     case TxGas =< MaxGas of
         true ->
             case aec_trees:apply_txs_on_state_trees([Tx], Trees, Env) of
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -408,7 +408,7 @@ check_candidate(Db, #dbs{gc_db = GCDb} = Dbs,
     Tx1 = aetx_sign:tx(Tx),
     TxTTL = aetx:ttl(Tx1),
     TxGas = aetx:gas(Tx1, Height),
-    case Height < TxTTL andalso ok =:= int_check_nonce(Tx, AccountsTree) of
+    case Height < TxTTL andalso TxGas > 0 andalso ok =:= int_check_nonce(Tx, AccountsTree) of
         true ->
             case Gas - TxGas of
                 RemGas when RemGas >= 0 ->
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -407,7 +407,7 @@ check_candidate(Db, #dbs{gc_db = GCDb} = Dbs,
                 AccountsTree, Height, Gas, Acc) ->
     Tx1 = aetx_sign:tx(Tx),
     TxTTL = aetx:ttl(Tx1),
-    TxGas = aetx:gas(Tx1),
+    TxGas = aetx:gas(Tx1, Height),
     case Height < TxTTL andalso ok =:= int_check_nonce(Tx, AccountsTree) of
         true ->
             case Gas - TxGas of
@@ -739,7 +739,8 @@ get_account(AccountKey, {block_hash, BlockHash}) ->
 
 check_minimum_fee(Tx0, _Hash) ->
     Tx = aetx_sign:tx(Tx0),
-    case aetx:fee(Tx) >= aetx:min_fee(Tx) of
+    Height = top_height(),
+    case aetx:fee(Tx) >= aetx:min_fee(Tx, Height) of
         true  -> ok;
         false -> {error, too_low_fee}
     end.
--- a/apps/aeoracle/src/aeo_response_tx.erl
+++ b/apps/aeoracle/src/aeo_response_tx.erl
@@ -133,7 +133,7 @@ check(#oracle_response_tx{nonce = Nonce, query_id = QueryId,
                      %% Contract is paying tx fee as gas.
                      aetx_contract -> [];
                      aetx_transaction ->
-                         [fun() -> aetx_utils:check_account(OraclePubKey, Trees, Nonce, Fee) end, %% Sender must be able to pay transaction fee before receiving query fee.
+                         [fun() -> aetx_utils:check_account(OraclePubKey, Trees, Nonce, Fee) end] %% Sender must be able to pay transaction fee before receiving query fee.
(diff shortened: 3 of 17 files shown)


apps/aeoracle/src/aeo_response_tx.erl
```diff
@@ -134,9 +134,9 @@ check(#oracle_response_tx{nonce = Nonce, query_id = QueryId,
                      aetx_contract -> []; %% TODO: Handle fees from contracts right.
                      aetx_transaction ->
                          [fun() -> aetx_utils:check_account(OraclePubKey, Trees,
-                                                            Nonce, Fee - QueryFee) end,
-                          fun() -> aeo_utils:check_ttl_fee(Height, ResponseTTL, Fee) end]
+                                                            Nonce, Fee - QueryFee) end]
```
zp-sd: This will be impacted by https://github.com/aeternity/epoch/pull/1792.

apps/aeoracle/src/aeo_utils.erl
```diff
@@ -8,50 +8,28 @@
...
+        ]).
 
 -include_lib("apps/aecontract/src/aecontract.hrl").
 
--define(ORACLE_TTL_FEE, {1, 1000}). %% One part over a thousand.
+-spec ttl_delta(aec_blocks:height(), aeo_oracles:ttl()) ->
+    non_neg_integer() | {error, too_low_height}.
+ttl_delta(_CurrHeight, {delta, D}) -> D;
+ttl_delta(CurrHeight, {block, H}) when H > CurrHeight -> H - CurrHeight;
+ttl_delta(_CurrHeight, {block, _H}) -> {error, too_low_height}.
```
lucafavatella: I see semantics changed from exception `error({too_low_height, H, CurrHeight})` to value `{error, too_low_height}` - I need to check code around this.

apps/aecore/src/aec_tx_pool.erl
```diff
@@ -407,7 +407,7 @@ check_candidate(Db, #dbs{gc_db = GCDb} = Dbs,
                 AccountsTree, Height, Gas, Acc) ->
     Tx1 = aetx_sign:tx(Tx),
     TxTTL = aetx:ttl(Tx1),
-    TxGas = aetx:gas(Tx1),
+    TxGas = aetx:gas(Tx1, Height),
```
lucafavatella: I see this can fail with badarith exception.
I see this is run in the worker process spawned `aec_block_generator:create_block_candidate`.
What prevents the worker process to terminate abnormally for oracle transaction with absolute TTL of state tree object not high enough for current chain?

apps/aecore/src/aec_block_micro_candidate.erl
```diff
@@ -210,7 +210,7 @@ add_txs_to_trees(MaxGas, Trees, Txs, Env) ->
 add_txs_to_trees(_MaxGas, Trees, [], Acc,_Env) ->
     {lists:reverse(Acc), Trees};
 add_txs_to_trees(MaxGas, Trees, [Tx | Txs], Acc, Env) ->
-    TxGas = aetx:gas(aetx_sign:tx(Tx)),
+    TxGas = aetx:gas(aetx_sign:tx(Tx), aetx_env:height(Env)),
     case TxGas =< MaxGas of
```
lucafavatella: I see that `TxGas` may be zero though in that case transaction will be grecefully skipped with reason `{error, too_low_abs_ttl}` by non-strict call `aec_trees:apply_txs_on_state_trees` below. Ok.

apps/aecore/src/aec_tx_pool.erl
lucafavatella: I see this fixed in 4f65906b (maybe before). Ok.

apps/aetx/src/aetx.erl
```diff
@@ -171,10 +172,21 @@ new(Callback, Tx) ->
 fee(#aetx{ cb = CB, tx = Tx }) ->
     CB:fee(Tx).
 
--spec gas(Tx :: tx()) -> Gas :: non_neg_integer().
-gas(#aetx{ type = Type, cb = CB, size = Size, tx = Tx }) when Type =/= channel_offchain_tx ->
-    aec_governance:tx_base_gas(Type) + Size * aec_governance:byte_gas() + CB:gas(Tx);
-gas(#aetx{ type = channel_offchain_tx }) ->
+-spec gas(Tx :: tx(), Height :: aec_blocks:height()) -> Gas :: non_neg_integer().
```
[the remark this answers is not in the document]
lucafavatella: As of 4f65906b I see this fixed. Ok.

apps/aeoracle/src/aeo_response_tx.erl
lucafavatella: I see this is now rebased on 1792. Ok.

apps/aeoracle/src/aeo_utils.erl
lucafavatella: I see caller `aec_query_tx:check_oracle_ttl` calls this then performs sum:
```
        Delta  = aeo_utils:ttl_delta(Height, QTTL),
        MaxTTL = aeo_utils:ttl_expiry(Height + Delta, RTTL),
```
that means that this would raise badarith though it is inside `try ... catch _:_ ` so this would be caught and return `{error, invalid_ttl}` that sounds reasonable. Ok.

I see test code in `aeo_txs_eqc` uses `+` on result that shall result in runtime error anyway, and I see no logic on type of runtime error (error exception too_low_height vs. badarith exception) so look ok cc @hanssv .

apps/aetx/src/aetx.erl
```diff
@@ -171,10 +172,35 @@ new(Callback, Tx) ->
...
     CB:fee(Tx).
 
--spec gas(Tx :: tx()) -> Gas :: non_neg_integer().
-gas(#aetx{ type = Type, cb = CB, size = Size, tx = Tx }) when Type =/= channel_offchain_tx ->
-    aec_governance:tx_base_gas(Type) + Size * aec_governance:byte_gas() + CB:gas(Tx);
-gas(#aetx{ type = channel_offchain_tx }) ->
+%% In case 0 is returned, the tx will not be included in the micro block
+%% candidate by the mempool.
+-spec gas(Tx :: tx(), Height :: aec_blocks:height()) -> Gas :: non_neg_integer().
+gas(#aetx{type = Type, cb = CB, size = Size, tx = Tx }, Height) when
```
lucafavatella: This function matches on transaction type for oracle transactions, accounting for state TTL, then has catch-all clause (`Type ... when Type =/= channel_offchain_tx`). It means that if code has typo in atom of transaction type (or they get misaligned) then code does not do what it is intended to do. I see there is test coverage in `aeoracle_SUITE` - checking that.

jur0: One thing I noticed is that I could also add `CB:gas(Tx)` to oracle txs (even if all of them return 0)

apps/aeoracle/test/aeoracle_SUITE.erl
```diff
@@ -133,15 +135,33 @@ register_oracle_negative_dynamic_fee(_Cfg) ->
 
     %% Test minimum fee for increasing TTL.
```
lucafavatella: I see these tests may be made more maintainable and precise by retrieving the fraction returned by `aec_governance:state_gas_per_block`, choosing numbers for preventing serialization to cross boundary hence changing required fee. Though ok for now.

PR #1798, opened 2018-11-19
Add dry-run HTTP API

hanssv:
At /debug/transactions/dry-run a list of transactions can be posted to be evaluated off-chain at the
state of a certain block hash. The API supports SpendTx, ContractCreateTx and
ContractCallTx.

https://www.pivotaltracker.com/story/show/161980624

### DIFF
--- a/apps/aehttp/src/aehttp_helpers.erl
+++ b/apps/aehttp/src/aehttp_helpers.erl
@@ -26,8 +26,7 @@
         , get_block_from_chain/1
         , get_block_hash_optionally_by_hash_or_height/1
         , safe_get_txs/1
-        , dry_run_top/1
-        , dry_run_txs/1
+        , do_dry_run/0
         , dry_run_results/1
         ]).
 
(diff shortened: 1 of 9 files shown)


hanssv:
Working on extending the API with an `accounts` option such that it is possible to add test accounts for the dry-run.

hanssv:
Added support for additional accounts to be used in dry-run. With this you can assure there is enough funds while doing the dry-run.

apps/aehttp/src/aehttp_helpers.erl
```diff
@@ -494,34 +493,76 @@ contract_call_input_funargs(CallInput) ->
...
-dry_run_txs([ETx | Txs], Acc) ->
+dry_run_txs_([ETx | Txs], Acc) ->
     case aehttp_api_encoder:safe_decode(transaction, ETx) of
         {ok, DTx} ->
             Tx = aetx:deserialize_from_binary(DTx),
             {Type, _} = aetx:specialize_type(Tx),
-            case lists:member(Type, [spend_tx, contract_crete_tx, contract_call_tx]) of
-                true  -> dry_run_txs(Txs, [Tx | Acc]);
-                false -> {error, list_to_binary(lists:concat(["Unsupported transaction type ", Type]))}
+            case lists:member(Type, [spend_tx, contract_create_tx, contract_call_tx]) of
```
velzevur: Nice catch

hanssv: I wrote a test ;-) It failed :-P

PR #1799, opened 2018-11-19
update osx version for build package

acidsennui:


.circleci/config.yml
```diff
@@ -426,7 +426,7 @@ jobs:
 
   osx_package:
     macos:
-      xcode: "9.0"
+      xcode: "10.0.0"
```
lucafavatella: From https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions : "10.0.0: Xcode 10.0 (Build 10A255)" then in [link](https://circle-macos-docs.s3.amazonaws.com/image-manifest/build-456/index.html) "macOS 10.13.6 (17G65)". Ok.

PR #1802, opened 2018-11-20
Disable Sophia code execution in code/call API

hanssv:
It is broken at the moment so no point in exposing it. The new dry-run
API is superior.


hanssv:
https://www.pivotaltracker.com/story/show/162082369

PR #1806, opened 2018-11-20
Disable Solidity EVM

hanssv:
It was of great use while developing, but it is not well tested enough
to be part of consensus, thus it is disabled.

See also https://github.com/aeternity/protocol/pull/271

https://www.pivotaltracker.com/story/show/162082161


apps/aecontract/src/aect_channel_contract.erl
```diff
@@ -33,6 +33,11 @@ run_new(ContractPubKey, Call, CallData, Trees0, OnChainTrees,
     CallStack = [], %% TODO: should we have a call stack for create_tx also
                     %% when creating a contract in a contract.
     VmVersion = aect_contracts:vm_version(Contract),
+    %% Assert VmVersion before running!
+    case VmVersion of
+        ?AEVM_01_Sophia_01 -> ok
```
lucafavatella: What is under consensus seems to me to be `:run`, and the corresponding check tx. /me checking.

lucafavatella: I am drafting VM check for state channel force progress in [temporary branch](https://github.com/aeternity/epoch/compare/PT-162082161-disable_solidity_evm...PT-162082161-disable_solidity_evm--sc-fp).

lucafavatella: https://github.com/aeternity/epoch/pull/1807

hanssv: Well spotted! I was slightly fooled by Dimitar! I've merged Luca's branch.

PR #1810, opened 2018-11-21
Disable opcodes unsupported in Sophia VM

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/162082307

Depends on PR #1815

TODO:
* [x] Review list of disabled opcodes
  * GASPRICE review is [tracked](https://www.pivotaltracker.com/story/show/160984121)
* [x] Review handling of `illegal_instruction` error
* [x] Release notes line
* [x] Bump DB folder in environments
* [ ] Once PR #1815 is merged, edit base branch back to master

Update of protocol repository tracked as separate ticket https://www.pivotaltracker.com/story/show/162131599


apps/aevm/src/aevm_eeevm.erl
```diff
@@ -131,17 +131,18 @@ loop(CP, StateIn) ->
         false ->
             OP     = code_get_op(CP, Code),
             State  = trace(CP, StateIn),
-            State0 = spend_op_gas(OP, State),
```
hanssv: Would it make sense to do something like:
```
check_invalid_instruction(OP, State),
State0 = spend_op_gas(OP, State),
```
to avoid repeating the code for invalid instructions and the `spend_op_gas` many times?

lucafavatella: I considered that, though that seems to me a worse solution:
* `check_invalid_instruction` would need to repeat the valid clauses of the case-of statement. So this is kicking the repetition somewhere else.
* `check_invalid_instruction` would make the code less clear to follow. I opted for the explicit solution.

UlfNorell: I agree with Hans here. By having `check_invalid_instruction` contain the list of invalid instructions you make the code a lot more readable. It wouldn't actually repeat anything, just move the specification of what instructions are valid to a single place instead of spread out over 1000+ lines of code.

Calling `spend_op_gas` explicitly in every case (is it every case? with the current code it's impossible to tell) is brittle a prone to forgetting it when changing the code.

lucafavatella:
Closing in favour of https://github.com/aeternity/epoch/pull/1818

PR #1812, opened 2018-11-21
Pt 162084238 gossip transactions restriction

ThomasArts:
Some minor improvements in making sync more load resistant

### DIFF
--- a/apps/aecore/src/aec_sync.erl
+++ b/apps/aecore/src/aec_sync.erl
@@ -203,16 +203,20 @@ handle_cast(_, State) ->
 handle_info({gproc_ps_event, Event, #{info := Info}},
             State = #state{ gossip_txs = GossipTxs }) ->
     %% FUTURE: Forward blocks only to outbound connections.
-    PeerIds = [ aec_peers:peer_id(P) || P <- aec_peers:connected_peers() ],
+    %% Take a random subset (possibly empty) of peers that agree with us 
+    %% on chain height to forward blocks and transactions to.
+    PeerIds = [ aec_peers:peer_id(P) || P <- aec_peers:get_random(10) ],
     NonSyncingPeerIds = [ P || P <- PeerIds, not peer_in_sync(State, P) ],
     case Event of
         block_to_publish ->
             Block = Info,
             enqueue(block, Block, NonSyncingPeerIds);
         tx_created when GossipTxs ->
+            %% If we allow http requests creating transactions when we are catching up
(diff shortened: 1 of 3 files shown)


apps/aecore/src/aec_sync.erl
```diff
@@ -203,16 +203,20 @@ handle_cast(_, State) ->
...
+    %% Take a random subset (possibly empty) of peers that agree with us 
+    %% on chain height to forward blocks and transactions to.
+    PeerIds = [ aec_peers:peer_id(P) || P <- aec_peers:get_random(10) ],
     NonSyncingPeerIds = [ P || P <- PeerIds, not peer_in_sync(State, P) ],
     case Event of
         block_to_publish ->
             Block = Info,
             enqueue(block, Block, NonSyncingPeerIds);
         tx_created when GossipTxs ->
+            %% If we allow http requests creating transactions when we are catching up
```
hanssv: This comment is weird, the "If" suggests this is configurable and the "just keep them in mempool" that we might not gossip them, which we will do?!

apps/aecore/src/aec_sync.erl
```diff
@@ -203,16 +203,20 @@ handle_cast(_, State) ->
 handle_info({gproc_ps_event, Event, #{info := Info}},
             State = #state{ gossip_txs = GossipTxs }) ->
     %% FUTURE: Forward blocks only to outbound connections.
-    PeerIds = [ aec_peers:peer_id(P) || P <- aec_peers:connected_peers() ],
+    %% Take a random subset (possibly empty) of peers that agree with us 
+    %% on chain height to forward blocks and transactions to.
+    PeerIds = [ aec_peers:peer_id(P) || P <- aec_peers:get_random(10) ],
     NonSyncingPeerIds = [ P || P <- PeerIds, not peer_in_sync(State, P) ],
```
sennui: we did gossip before to peers that we are in sync with - so commit message is not accurate?
He we limit it to 10 random nodes - maybe we should make it configurable?

sennui: Ah, yes, sorry, blindness doesn't hurt. And it makes sense of course to sync with nodes that have different view of the chain.

Not really convinced with no-config way. We don't protect it with protocol, so there will be number of implementations trying out another values. We can expose it and make it easier for testing.

hanssv: Otherwise the variable would have been named `NotSyncedPeerIds` obviously 🤷‍♂️

PR #1814, opened 2018-11-21
PT-162048442 optimize store access

UlfNorell:
Pivotal: [PT-162048442](https://www.pivotaltracker.com/story/show/162048442)

I'm not super confident of the POI stuff, and in particular some of the changes made to the tests, so I would appreciate careful review of this.


hanssv:
@UlfNorell unless you changed the gas computation and thereby consensus you don't have to bump anything?!

hanssv:
Fixed a failing test case... Wrong initial value of the store...

apps/aecontract/src/aect_state_tree.erl
```diff
@@ -226,21 +210,18 @@ verify_poi(Pubkey, Contract, Poi) ->
...
-                        false -> {error, bad_proof}
-                    end;
-                {ok, _} -> {error, bad_proof};
-                {error, _} = E -> E
-            end
+    StoreMap = aect_contracts_store:subtree(<<>>, Store),
+    try lookup_store_poi(Id, Poi) of
+        {ok, Store1} ->
+            StoreMap1 = aect_contracts_store:subtree(<<>>, Store1),
+            case StoreMap =:= StoreMap1 andalso IsLegalStoreFun(Store) of
```
gorillainduction: I think this is not good enough. The test is supposed to check that the store is "complete", i.e., that the MPT is complete for the store. Calling `aect_contracts_store:subtree(Store)` should be good enough. Or even better, make another function `aect_contracs_store:verify/1` that just traverses the store and throws away the result.

Unless there are some invariants on the values as well, in which case you could check them as well.

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -3409,23 +3426,23 @@ sophia_state_gas(_Cfg) ->
...
+
 create_store(_Cfg) ->
     state(aect_test_utils:new_state()),
     Acc1  = ?call(new_account, 100),
     Ct1   = ?call(insert_contract, Acc1, <<"Code for C1">>),
     Ct1   = ?call(get_contract, Ct1),
-    #{}   = aect_contracts:state(Ct1),
-    ok.
-
-update_store(_Cfg) ->
```
gorillainduction: Why is this test removed?

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -3466,42 +3483,19 @@ merge_new_zero_value(_Cfg) ->
...
+    Ct4    = set_ct_store(Store2, Ct3),
     Ct4    = ?call(enter_contract, Ct4),
     Ct5    = ?call(get_contract, Ct4),
     Store3 = #{ <<0>> => <<0>>
               , <<2>> => <<42>>},
-    Store3 = aect_contracts:state(Ct5),
-    ok.
-
-
-merge_missing_keys(_Cfg) ->
```
gorillainduction: Why is this test removed?

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -1977,8 +1977,14 @@ contract_transactions(_Config) ->    % miner has an account
     PoI = aec_trees:deserialize_poi(PoIBin),
     {ok, ContractInPoI} = aec_trees:lookup_poi(contracts, ContractPubKey, PoI),
     {ok, Trees} = rpc(aec_chain, get_top_state, []),
-    ContractInPoI = rpc(aect_state_tree, get_contract, [ContractPubKey,
+    ContractInPoI1 = rpc(aect_state_tree, get_contract, [ContractPubKey,
                                                          aec_trees:contracts(Trees)]),
+    %% Don't require the store mp trees to be identical, but check that the
```
gorillainduction: MPT is identical if the content is the same. That is the whole point of them. If that test fails, something is wrong.

apps/aecontract/src/aect_state_tree.erl
```diff
@@ -226,21 +210,18 @@ verify_poi(Pubkey, Contract, Poi) ->
     end.
 
 verify_store_poi(Id, Store, IsLegalStoreFun, Poi) ->
-    %% Check separately absence of (invalid) empty key because Merkle
-    %% tree iterator-from-key, used in PoI lookup, does not return
-    %% initial key if present.
```
UlfNorell: I dropped this check. With the new representation empty keys are ignored, but I'm not completely sure what the check was for in the first place, so I don't know if dropping it has unintended consequences.

apps/aevm/src/aevm_eeevm_store.erl
```diff
@@ -196,31 +195,33 @@ store_maps(Maps0, Store) ->
 perform_update({new_inplace, NewId, OldId}, Store) ->
     OldKey   = <<OldId:256>>,
     NewKey   = <<NewId:256>>,
-    OldEntry = maps:get(OldKey, Store),
-    Store1   = maps:remove(OldKey, Store),
-    Store1#{ NewKey => OldEntry };
+    OldEntry = store_get(OldKey, Store),
+    %% Subtle: Don't remove the old entry because the mp trees requires there
```
gorillainduction: This is problematic. And I guess this is the kicker for why you test the content rather than the MPT above.

If you leave old values lying around, the root hash of the store changes, as will the root hash of the state. This is not good enough since it is under consensus.

apps/aevm/src/aevm_eeevm_store.erl
```diff
@@ -196,31 +195,33 @@ store_maps(Maps0, Store) ->
...
+    Data = [ {<<Id:256, Key/binary>>, Val} || {Key, Val} <- maps:to_list(Map#pmap.data) ],
+    lists:foldl(fun({K, V}, S) -> store_put(K, V, S) end, Store, Info ++ Data);
 perform_update({gc, Id}, Store) ->
     RealId = real_id(Id, Store),
-    %% Remove map info entry
-    Store1 = maps:remove(<<Id:256>>, Store),
-    %% Remove all the data
-    maps:filter(fun(<<Id1:256, Key/binary>>, _) when Key /= <<>> -> Id1 /= RealId;
-                   (_, _) -> true end, Store1).
+    %% Remove map info entry. Also remove RealId which we kept around for mp
```
gorillainduction: Ah, ok. So this is supposed to fix the problem above. Is it guaranteed to be performed before setting the store on exit?

apps/aecontract/test/aecontract_SUITE.erl
UlfNorell: As far as I could tell it was the same as `read_store`, but with fewer checks.

apps/aecontract/src/aect_state_tree.erl
UlfNorell: I don't understand. Before, the store in the poi was a `binary() => binary()` map, and we checked that you got the same map back when looking up the poi. Now, due to the write cache, you might not have exactly the same representation so I convert to map and then check.

Or are you talking about the `IsLegalStoreFun` part?

apps/aecontract/test/aecontract_SUITE.erl
UlfNorell: In the new representation you remove things by writing `<<>>` so there is no merging of missing keys. If a key is not in the write cache it means it doesn't change.

apps/aehttp/test/aehttp_integration_SUITE.erl
UlfNorell: Yes, this one had me a bit worried.

In a lot of places we compare stores with a non-empty write cache, to a store obtained from state trees (with values stored in the MPT). I think this test failed with different MPTs though.

I will do some investigation.

apps/aevm/src/aevm_eeevm_store.erl
UlfNorell: I don't think there's a problem leaving the old entry around (but, yes this does affect consensus). This is just an internal detail in how contracts manage the store and shouldn't affect things outside.

apps/aevm/src/aevm_eeevm_store.erl
UlfNorell: No. See above.

UlfNorell:
Hans says I don't have to bump protocol... isn't it already bumped several times? I guess the protocol-relevant thing that has changed is serialisation format for PoIs?

PR #1820, opened 2018-11-22
Increase REQUEST_TIMEOUT to 30s

hanssv:


hanssv:
https://www.pivotaltracker.com/story/show/162156292

PR #1825, opened 2018-11-22
Pt 162089378 cleanup compiler errors

rvirding:
The compiler returns errors in many different ways, and this is first step to clean it up. It is now easier to which errors are those generated by the compiler and which are generated by bugs. The compiler generates 3 types of errors `parse_errors`, `type_errors` and `core_errors` coming from different phases. We now generate better text as well. It is backwards compatible in the HTTP interface through the `aect_sophia:compile` generating better text.

This is just a first stage but more cleanup would entail making some decisions about how would we like it to work which should be discussed after the release.

This fixes [PT-162089378](https://www.pivotaltracker.com/story/show/162089378).

### DIFF
--- a/apps/aecontract/src/aect_sophia.erl
+++ b/apps/aecontract/src/aect_sophia.erl
@@ -24,10 +24,19 @@ compile(ContractAsBinString, OptionsAsBinString) ->
     Options = parse_options(OptionsAsBinString),
     try {ok, aeso_compiler:from_string(ContractText, Options)}
     catch
+        %% The compiler errors.
         error:{type_errors, Errors} ->
             {error, list_to_binary(string:join(["** Type errors\n" | Errors], "\n"))};
+        error:{parse_errors, Errors} ->
+            {error, list_to_binary(string:join(["** Parse errors\n" | Errors], "\n"))};
+        error:{code_errors, Errors} ->
+            ErrorStrings = [ io_lib:format("~p", [E]) || E <- Errors ],
+            {error, list_to_binary(string:join(["** Code errors\n" | ErrorStrings], "\n"))};
+        %% General programming errors in the compiler.
         error:Error ->
-            {error, list_to_binary(io_lib:format("~p", [Error]))}
+            Where = hd(erlang:get_stacktrace()),
+            ErrorString = io_lib:format("Error: ~p in\n   ~p", [Error,Where]),
+            {error, list_to_binary(ErrorString)}
     end.
 
 parse_options(OptionsBinString) ->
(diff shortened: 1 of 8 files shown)


lucafavatella:
@rvirding @sennui @hanssv @happi I took the liberty of amending base branch to `develop` so we consider twice before where to merge.

apps/aecontract/src/aect_sophia.erl
```diff
@@ -24,10 +24,19 @@ compile(ContractAsBinString, OptionsAsBinString) ->
     Options = parse_options(OptionsAsBinString),
     try {ok, aeso_compiler:from_string(ContractText, Options)}
     catch
+        %% The compiler errors.
         error:{type_errors, Errors} ->
             {error, list_to_binary(string:join(["** Type errors\n" | Errors], "\n"))};
+        error:{parse_errors, Errors} ->
```
lucafavatella: I see more specific errors. I broadly checked these tuples are used consistently.

apps/aecontract/src/aect_sophia.erl
```diff
@@ -24,10 +24,19 @@ compile(ContractAsBinString, OptionsAsBinString) ->
...
+            {error, list_to_binary(string:join(["** Parse errors\n" | Errors], "\n"))};
+        error:{code_errors, Errors} ->
+            ErrorStrings = [ io_lib:format("~p", [E]) || E <- Errors ],
+            {error, list_to_binary(string:join(["** Code errors\n" | ErrorStrings], "\n"))};
+        %% General programming errors in the compiler.
         error:Error ->
-            {error, list_to_binary(io_lib:format("~p", [Error]))}
+            Where = hd(erlang:get_stacktrace()),
```
lucafavatella: I see this called only in internal HTTP endpoint. Ok.

apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -1050,18 +1050,37 @@ erc20_token_contract(Config) ->
...
-type_error_contract(_Cfg) ->
-    {ok, 403, #{<<"reason">> := Error}} =
+%% compiler_error_contract(Config)
+%%  Check HTTP output from compiler errors.
+
+compiler_error_contract(_Config) ->
+    %% Test syntax error.
+    {ok,403,#{<<"reason">> := SyntaxError}} =
+        get_contract_bytecode(<<"contract Fail = function fail(x : string) == x + 1">>),
+    <<"** Parse errors\n\n",_/binary>> = SyntaxError,
```
lucafavatella: I see errors propagated to HTTP response. Nice.

PR #1826, opened 2018-11-23
Misc cleanup while running a local node

hanssv:
* Compiler warnings
* Don't re-insert key block when adding forward generations (it will fail for the genesis block and generate a warning!)
* Remove excessive (and confusing/incorrect) debug logging in aec_peers
* Don't busy loop sync workers so agressively
* Fix intermittent eunit test failure (running out of time occasionally)
* Fix intermittent test failure (micro-forks left mempool non-empty)


hanssv:
Solves this one https://www.pivotaltracker.com/story/show/162094456

PR #1831, opened 2018-11-23
Fix erronous spec in aec_pof

velzevur:
Spotted by @d-velev , no PT

### DIFF
--- a/apps/aecore/src/aec_pof.erl
+++ b/apps/aecore/src/aec_pof.erl
@@ -34,7 +34,7 @@
               hash/0
              ]).
 
--spec new(aec_headers:header(), aec_headers:header(), aec_blocks:block_header_hash()) -> pof().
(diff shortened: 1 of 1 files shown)


velzevur:
Updated `aec_keys:pubkey()` -> `miner_pubkey()`
Will squash commits when merging.

apps/aecore/src/aec_pof.erl
```diff
@@ -34,7 +34,7 @@
               hash/0
              ]).
 
--spec new(aec_headers:header(), aec_headers:header(), aec_blocks:block_header_hash()) -> pof().
+-spec new(aec_headers:header(), aec_headers:header(), miner_pubkey()) -> pof().
```
lucafavatella: I see this is called passing key taken from aec_headers:miner so type ok.

PR #1837, opened 2018-11-26
Add option to use Mnesia disc_copies

tolbrino:
This is useful as an alternative to using `rocksdb` on Windows.


apps/aecore/src/aec_db.erl
```diff
@@ -612,6 +613,7 @@ wait_for_tables(Tabs, Sofar, _, _) ->
 check_db() ->
     try
         Mode = case application:get_env(aecore, persist, false) of
```
uwiger: Actually, this value is usually set (via `aecore_app:check_env()` using the epoch config, and the value `chain:persist` is defined in the schema as a boolean. The schema definition should be updated accordingly (and perhaps `true|false|legacy` isn't ideal)

PR #1838, opened 2018-11-26
Parameterize default max gossip

ThomasArts:
Make gossip parameter configurable
```
sync:
      max_gossip: 10
```

### DIFF
--- a/apps/aecore/src/aec_sync.erl
+++ b/apps/aecore/src/aec_sync.erl
@@ -33,6 +33,7 @@
 -endif.
 
 -define(MAX_HEADERS_PER_CHUNK, 100).
+-define(DEFAULT_MAX_GOSSIP, 10).
 
 %%%=============================================================================
 %%% API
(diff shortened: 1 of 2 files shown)


fabiankrol:
We might want to add parameter to `epoch/apps/aeutils/test/data/epoch_full.yaml`

ThomasArts:
I did not add it to epoch_full.yaml because I don't believe this is something the user should fiddle with. As I posted before, we try to be friendly to other nodes using this variable. It is very likely that some users will see this value and say: let's send it to everyone, then my transactions get everywhere quickly.

We don't want users to put 1000 there. This would potentially break the system as a whole. Therefore, try to minimize use of this variable outside what we believe is a good value.

hanssv:
Single users setting it to 1000 wont make much difference... As long as we don't mention it in the main setup documents we should be fairly safe.

fabiankrol:
@ThomasAerts 👍

PR #1851, opened 2018-11-26
Gossip to everyone when a block is created by the local node

gorillainduction:
https://www.pivotaltracker.com/story/show/162221020


apps/aecore/src/aec_sync.erl
```diff
@@ -211,8 +211,13 @@ handle_info({gproc_ps_event, Event, #{info := Info}},
     NonSyncingPeerIds = [ P || P <- PeerIds, not peer_in_sync(State, P) ],
     case Event of
         block_to_publish ->
-            Block = Info,
-            enqueue(block, Block, NonSyncingPeerIds);
+            case Info of
+                {created, Block} ->
+                    PeerIds1 = [ aec_peers:peer_id(P) || P <- aec_peers:connected_peers(all) ],
```
zp-sd: Ok, I understand we will gossip the blocks that we found out ourselves to all connected peers.

PR #1856, opened 2018-11-27
change default network id name

sennui:


sennui:
closing, as it affects aepps to much

PR #1860, opened 2018-11-27
Use modified rebar3 to auto-generate erlang cookie if needed

uwiger:
See https://github.com/aeternity/rebar3, branch uw-auto-set-cookies

The `relx` extended script template has been modified to run a named erlang node when determining the root_dir (it used to run a non-distributed node), and also to simply start and stop a named node even if the ERTS dir is included in the build. This will automatically create a `~/.erlang.cookie` file, if it doesn't already exist.


tolbrino:
@uwiger Can you merge this? I will need to update rebar again later.

lucafavatella:
Marked as WIP as release scope to be clarified.

tolbrino:
@lucafavatella @uwiger Can this be merged now? It blocks the Windows work, since that relies on fixes in recent versions of rebar.

lucafavatella:
The problem was solved by https://github.com/aeternity/aeternity/pull/1859
This feature is under consideration upstream https://github.com/erlware/relx/pull/678#

PR #1869, opened 2018-11-29
Doc example tool for generating key-pair

lucafavatella:


zp-sd:
@lucafavatella We may also want to put additional note in https://github.com/aeternity/epoch/blob/master/docs/docker.md#generating-beneficiary-account-for-the-first-time? This is testnet section, but maybe worth mentioning anyway?

PR #1877, opened 2018-11-30
Gossip to a subset of connected peers

sylane:
Add an intent of fixing one of the timeout from https://www.pivotaltracker.com/n/projects/2124891/stories/162340213 too.

### DIFF
--- a/apps/aecore/src/aec_peers.erl
+++ b/apps/aecore/src/aec_peers.erl
@@ -38,6 +38,7 @@
 -export([available_peers/0, available_peers/1]).
 -export([blocked_peers/0]).
 -export([get_random/1, get_random/2]).
+-export([get_random_connected/1]).
 -export([get_connection/1]).
 
 %% Utility functions.
@@ -263,6 +264,12 @@ get_random(N, Exclude)
        (N =:= all) orelse (is_integer(N) andalso N >= 0) ->
     gen_server:call(?MODULE, {get_random, N, Exclude}).
 
+
+%% @doc Gets up to N random connected peers.
+-spec get_random_connected(pos_integer()) -> [peer_info()].
+get_random_connected(N) when (is_integer(N) andalso N > 0) ->
+    gen_server:call(?MODULE, {get_random_connected, N}).
+
 %% @doc Gets a connection PID from a peer identifier.
 -spec get_connection(peer_id()) -> {ok, pid()} | {error, term()}.
 get_connection(PeerId) ->
@@ -411,6 +418,8 @@ handle_call({get_random, N, Exclude}, _From, State0) ->
     {Subset, State} = pool_random_subset(N, Exclude, State0),
     Result = [ peer_info(P) || {_, P} <- Subset ],
     {reply, Result, State};
+handle_call({get_random_connected, N}, _From, State) ->
+    {reply, random_connected_peers(N, all, State), State};
 handle_call({connection_failed, PeerId, PeerCon}, _From, State0) ->
     State = on_connection_failed(PeerId, PeerCon, State0),
     {reply, ok, update_peer_metrics(schedule_connect(State))};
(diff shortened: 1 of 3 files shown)


apps/aecore/src/aec_peers.erl
```diff
@@ -658,6 +667,13 @@ connected_peers(outbound, #state{ conns = Conns }) ->
     [ peer_info(P) || #conn{ peer = P, type = outbound, state = connected }
                       <- maps:values(Conns) ].
 
+-spec random_connected_peers(N :: pos_integer(), all | inbound | outbound, state())
+    -> [peer_info()].
+random_connected_peers(N, Tag, State) ->
```
sennui: this is introduced for the the usage below in aec_sync. Don't we need in this very case only Outbound peers?

sylane: From what @hanssv says, It seems there is some cases where nodes don't have outbound connection.

sennui: So then we can't do anything. The point of change is to limit initial dataset which we use for picking up random nodes to the ones that are outbound?

sylane: The first point of this PR is to send gossip to connected peers. Then the idea is to send them only to outbound. Theoretically, all the nodes should reach close to the maximum number of outbound connections (10), but @hanssv said he saw some nodes without outbound connection. We should look at these nodes and investigate why they don't have outbound connections and then it will be safe to only gossip to outbound connections.

hanssv: The cases without outbound connections are artificial I think, but for local testing networks it will definitely happen. Gossiping to only 10 nodes is low though...

sylane: The limit could be raised or we could do a smarter connection selection that takes all the outbound connections plus top up to a configured number with random inbound connections...

apps/aecore/src/aec_peers.erl
```diff
@@ -658,6 +667,13 @@ connected_peers(outbound, #state{ conns = Conns }) ->
...
+    Keyed = [{rand:uniform(), P} || P <- connected_peers(Tag, State)],
+    Sorted = lists:keysort(1, Keyed),
```
ThomasArts: Why do you sort them? I would expect that sorting indirectly implies that certain nodes will get more gossipo than others.... not sure if that has any negative effect for the network as a whole.

We might need to experiment a bit with this.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -549,7 +549,7 @@ handle_request_('GetStatus', _Params, _Context) ->
                  end, [], aec_governance:protocols()),
     NodeVersion = aeu_info:get_version(),
     NodeRevision = aeu_info:get_revision(),
-    PeerCount = length(aec_peers:get_random(all)),
+    PeerCount = aec_peers:count(peers),
```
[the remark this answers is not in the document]
hanssv: I think we should...

apps/aecore/src/aec_peers.erl
sylane: This is to shuffle them, they are sorted on the random key.

apps/aehttp/src/aehttp_dispatch_ext.erl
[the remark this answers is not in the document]
sylane: Maybe I am missing something, what will it change ? `aec_peers:count(peers)` gives the total number of peers we know, isn't what you want to show ?

apps/aecore/src/aec_peers.erl
ThomasArts: Good!

apps/aehttp/src/aehttp_dispatch_ext.erl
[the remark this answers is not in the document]
velzevur: I believe we should expose both of those to the endpoint so we can monitor it
```
(epoch@localhost)19> aec_peers:count(peers).
1330
(epoch@localhost)20> aec_peers:count(connections).
3
```
@hanssv ?

PR #1879, opened 2018-12-03
Copy all docs in rebar.config

jur0:


rebar.config
```diff
@@ -63,9 +63,13 @@
 
         {overlay, [{copy, "REVISION", "REVISION"},
                    {copy, "VERSION" , "VERSION"},
+                   {copy, "docs/build.md", "docs/build.md"},
```
lucafavatella: Including build instructions in built package seems unnecessary.

PR #1880, opened 2018-12-03
PT-161759909 Rename `aetx:gas/2` to `aetx:gas_limit/2`

jur0:
[PT-161759909](https://www.pivotaltracker.com/n/projects/2124891/stories/161759909)

### DIFF
--- a/apps/aetx/src/aetx.erl
+++ b/apps/aetx/src/aetx.erl
@@ -12,7 +12,7 @@
         , check/3
         , deserialize_from_binary/1
         , fee/1
-        , gas/2
+        , gas_limit/2
         , min_gas/2
         , gas_price/1
         , ttl/1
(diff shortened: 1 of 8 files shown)


lucafavatella:
Pure refactoring - I prefer to defer this. I am taking the liberty of labelling as WIP.

apps/aetx/src/aetx.erl
```diff
@@ -174,8 +174,8 @@ fee(#aetx{ cb = CB, tx = Tx }) ->
 
 %% In case 0 is returned, the tx will not be included in the micro block
 %% candidate by the mempool.
--spec gas(Tx :: tx(), Height :: aec_blocks:height()) -> Gas :: non_neg_integer().
-gas(#aetx{type = Type, cb = CB, size = Size, tx = Tx }, Height) when
+-spec gas_limit(Tx :: tx(), Height :: aec_blocks:height()) -> Gas :: non_neg_integer().
+gas_limit(#aetx{type = Type, cb = CB, size = Size, tx = Tx }, Height) when
```
lucafavatella: I suggest option A below as way forward. Though only nice to have.

I understand the only semantics of this function is the contribution of the tx towards the block gas limit. Correct?
If yes, I would suggest either of the following two:
* (A) Add comment on top of this function mentioning semantics.
* (B) Rename function in order to have more explicit name (as `gas_limit` as name is used in contract txs as name for max gas specified by user, while this is derived quantity). E.g. `gas_for_micro_block_limit` though this is very debatable - and internal - so not strong opinion.

jur0: @lucafavatella correct, the function is used for calculation of tx's contribution towards the block gas limit.

PR #1883, opened 2018-12-03
PT-161722828 Add network_id to /status endpoint

zp-sd:


zp-sd:
It seems unrelated, but I restarted 2 times already?! I will have last try before I investigate this docker stuff...

zp-sd:
...superseded by https://github.com/aeternity/epoch/pull/1891, branch name with `/` does not work...

PR #1884, opened 2018-12-03
PT-162156314 Remove ulimit check

jur0:
[PT-162156314](https://www.pivotaltracker.com/n/projects/2124891/stories/162156314)

### DIFF
--- a/hooks/pre_start.sh
+++ b/hooks/pre_start.sh
@@ -1,9 +1,0 @@
-#!/bin/bash
-
-OPEN_FILES_RECOMMENDED=24576
-OPEN_FILES_LIMIT=$(ulimit -n)
-
-if [ "$OPEN_FILES_RECOMMENDED" -gt "$OPEN_FILES_LIMIT" ]; then
-    echo "WARNING: ulimit -n is $OPEN_FILES_LIMIT; $OPEN_FILES_RECOMMENDED is the recommended minimum."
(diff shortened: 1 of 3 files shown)


hooks/pre_start.sh
```diff
@@ -1,9 +0,0 @@
-#!/bin/bash
-
-OPEN_FILES_RECOMMENDED=24576
```
lucafavatella: I would expect that if there is not any longer a recommended limit then [our infrastructure shall align](https://github.com/aeternity/infrastructure/blob/a759f5b1a12f49458146fb144b831882eb8fb57e/ansible/setup.yml#L150-L155)? If so, we may consider:
* Amending infrastructure repo removing ulimit limit.
* Provisioning fresh env (or at least fresh boxes in env) with such scripts (so that ulimit is left to default of OS/image).
* Smoke test such env.
* Load test such env e.g.: max peer number reached, sync handshake/ping happens when it should, rotating peers (hence causing potential TCP connections in TIME_WAIT state or similar).

Also we shall smoke test on supported OS/archs + Docker.

hanssv:
The default on my Mac seems to be 256 and that is too low at times, I've had crashes...

dincho:
> The default on my Mac seems to be 256 and that is too low at times, I've had crashes...

@thepiwo also reported crashes without ulimit on OSX

lucafavatella:
Marked as WIP as pending comments.

jur0:
Closing as we prefer warning rather than node crashing.

PR #1893, opened 2018-12-04
Uw leveled backend

uwiger:
Tested two ways:
* Running `make test` with default (`rocksdb`) backend
* Running `make test`, `make docker`, `make system-test` with `epoch_config_schema.json` locally modified to use `"leveled"` for `db_backend` rather than `"rocksdb"`


lucafavatella:
Marked as WIP as release scope to be clarified, and conflicts to be resolved.

tolbrino:
The integration has been merged as part of #2283 , so this can be closed.

PR #1901, opened 2018-12-06
PT-162429019 Disable internal debug api by default

fabiankrol:
[PT-162429019: Disable internal debug endpoints by default](https://www.pivotaltracker.com/story/show/162429019)

By default all internal `/debug/*` APIs are disabled. ~~`404 Not Found`~~ `403 Forbidden` is returned for them.

~~I'm introducing `dev0` node with config based on `dev1` without `debug_endpoints` flag.~~

Debug api can be enabled by setting `debug_endpoints` flag to `true`
##### `epoch.yaml`
```
http:
    internal:
        debug_endpoints: true
```

##### `sys.config`
```  
{aehttp, [
      {internal, [
          {debug_endpoints, true}
          ]}]}
```


fabiankrol:
@ThomasAerts @sylane @eproxus FYI performance test might need configuration update, I didn't check it. ref: https://github.com/aeternity/epoch/pull/1901/commits/57bdef878ac1ae53704e50d36a72b7ab9babb08b#diff-78c2f17afb25bb1b45480d5aa13b53deR20

dincho:
What's the rationale behind this change (can't find details in the PT)?

I see it targets master, I want to stress again that changing configuration defaults IS backward compatibility break.

fabiankrol:
@dincho the `1.0.0` should have already disabled debug endpoints by default. cc @evdoks 

I agree it's a API breaking change, but since it's a debug api (and users should be aware of it) we should be fine with it. cc @dincho @lucafavatella 

I can set the default value to `true`. This way we'll allow users to change exposed endpoints without braking api. We can update the default value to `false` when the first major api braking change is needed.

velzevur:
I find this more of a fine-tuning than to be a backwards compatibilty breaking change. Especially the transaction producing endpoints shall never ever be used in a real life scenario.

apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -241,6 +241,11 @@
                             "description" : "Number of acceptors in internal pool",
                             "type" : "integer",
                             "default" : 10
+                        },
+                        "debug_endpoints" : {
+                            "description" : "Enable (true) debug api. Disabled (false) by default",
```
lucafavatella: This seems overlapping with:
* "http > endpoints > debug"
* "http > debug"

Could you please detail the schema description for those two config parameters?

lucafavatella: What constitutes part of the "debug api"? The paths tagged as `debug` in swagger.yaml, or the paths starting with `/debug`? It may not be immediate for the reader, and may lead to inconsistencies in the future.

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -473,21 +473,15 @@ handle_request_('GetPendingTransactions', _Req, _Context) ->
     {200, [], #{transactions => [aetx_sign:serialize_for_client_pending(T) || T <- Txs]}};
 
 handle_request_('GetPeers', _Req, _Context) ->
-    case aeu_env:user_config_or_env([<<"http">>, <<"debug">>],
```
lucafavatella: Is "http > debug" in user config schema a leftover?

apps/aeutils/priv/epoch_config_schema.json
dincho: I just faced that issue, there are the two options mentioned by @lucafavatella above, and now this one (new) ?

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -473,21 +473,15 @@ handle_request_('GetPendingTransactions', _Req, _Context) ->
...
-                                    aehttp, enable_debug_endpoints, false) of
```
lucafavatella: Leftover usages?
```
$ git grep enable_debug_endpoints origin/PT-162429019-disable-internal-debug-api-by-default
origin/PT-162429019-disable-internal-debug-api-by-default:docker-compose.yml:12:      -aehttp enable_debug_endpoints true
origin/PT-162429019-disable-internal-debug-api-by-default:docker-compose.yml:25:      -aehttp enable_debug_endpoints true
origin/PT-162429019-disable-internal-debug-api-by-default:docker-compose.yml:38:      -aehttp enable_debug_endpoints true
origin/PT-162429019-disable-internal-debug-api-by-default:docs/docker.md:85:    aeternity/epoch -aehttp enable_debug_endpoints true
```

lucafavatella:
If using dev0 in tests, then Circle CI config needs to be adapted (in some jobs we collect the logs of dev1 dev2 dev3 - dev0 shall be added).

apps/aeutils/priv/epoch_config_schema.json
fabiankrol: I updated description to "deprecated" for both. I'm not sure if we should remove them right now or wait for next release.

fabiankrol: Description updated. I'm checking `debug` tag in swagger: https://github.com/aeternity/epoch/pull/1901/files#diff-744f44aa1c55e60ba2c122893c02ebe0R37

apps/aehttp/src/aehttp_dispatch_int.erl
fabiankrol: Marked as deprecated, see https://github.com/aeternity/epoch/pull/1901#discussion_r243129204

apps/aehttp/src/aehttp_dispatch_int.erl
fabiankrol: ✅

fabiankrol:
> lucafavatella https://github.com/aeternity/epoch/pull/1901#issuecomment-446651947
> If using dev0 in tests, then Circle CI config needs to be adapted (in some jobs we collect the logs of dev1 dev2 dev3 - dev0 shall be added).

I removed dev0 instance

fabiankrol:
API change: `403 Forbidden` is returned for disabled debug endpoints

fabiankrol:
@lucafavatella @velzevur rebased
unless we decide to change default behaviour it's ready to review

apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -241,6 +241,11 @@
                             "description" : "Number of acceptors in internal pool",
                             "type" : "integer",
                             "default" : 10
+                        },
+                        "debug_endpoints" : {
+                            "description" : "Enable (true) debug api. Disabled (false) by default. Debug endpoints are defined by a tag in swagger.yaml config",
```
lucafavatella: A minor readability improvement.

```suggestion
                            "description" : "Enable (true) debug api. Disabled (false) by default. Debug endpoints are defined by a tag in the Swagger API schema",
```

apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -270,7 +275,7 @@
                         "type" : "boolean"
                     },
                     "debug" : {
-                        "description" : "Debug endpoints",
+                        "description" : "deprecated",
```
lucafavatella: Could you please add a test config file (`apps/aeutils/test/data`) hooked to unit tests showing that the presence of this config does not make the validation of the config fail? (I did not find one.)

apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -279,7 +284,7 @@
                     }
                 },
                 "debug" : {
-                    "description" : "Whether to enable the debug endpoints",
+                    "description" : "deprecated",
```
lucafavatella: Could you please add a test config file (`apps/aeutils/test/data`) hooked to unit tests showing that the presence of this config does not make the validation of the config fail? (I did not find one.)

apps/aeutils/priv/epoch_config_schema.json
lucafavatella:
```suggestion
                        "description" : "Deprecated. See also 'http > internal > debug_endpoints'",
```

apps/aeutils/priv/epoch_config_schema.json
lucafavatella:
```suggestion
                        "description" : "Deprecated. See also 'http > internal > debug_endpoints'",
```

docs/release-notes/next/PT-162429019-disable-internal-debug-api-by-default.md
```diff
@@ -0,0 +1,2 @@
+* Disable internal debug API endpoints by default. To enable setup epoch.yaml `http > internal > debug_endpoints` to `true`
+* Mark `http > endpoints > debug` and `http > debug` configuration params as deprecated
```
lucafavatella: @fabiankrol Just ensuring we are on the same page - removal of these is planned for Minerva, right? Cc @dincho

lucafavatella: [Tracked](https://www.pivotaltracker.com/story/show/163387507).

apps/aehttp/src/aehttp_app.erl
```diff
@@ -164,3 +165,7 @@ get_channel_websockets_acceptors() ->
     aeu_env:config_value([<<"websocket">>, <<"channel">>, <<"acceptors">>],
                          aehttp, [channel, websocket, handlers],
                          ?DEFAULT_CHANNEL_ACCEPTORS).
+
+enable_internal_debug_endpoints() ->
+    aeu_env:config_value([<<"http">>, <<"internal">>, <<"debug_endpoints">>],
+                         aehttp, [internal, debug_endpoints], false).
```
lucafavatella: The default is duplicated from the schema though minor and debatable so ok.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -5070,7 +5078,27 @@ peers(_Config) ->
...
 
     {ok, 200, #{<<"blocked">> := [], <<"peers">> := []}} = get_peers(),
+    ok.
 
+disabled_debug_endpoints(_Config) ->
+    ?assertMatch(true, rpc(?NODE, aehttp_app, enable_internal_debug_endpoints, [])),
+    {ok, Internal} = rpc(?NODE, application, get_env, [aehttp, internal]),
+    NewInternal = proplists:delete(debug_endpoints, Internal),
+    ok = rpc(?NODE, application, set_env, [aehttp, internal, NewInternal]),
+    ?assertMatch(false, rpc(?NODE, aehttp_app, enable_internal_debug_endpoints, [])),
```
lucafavatella: I see this test is changing state of node that is carried over to other test case. So tests become order dependent. And I see `enabled_debug_endpoints` depends on true, and I did not find documentation on order of test cases (I did not find in doc how `sequence` be inherited somehow). I propose:
* A. Add option sequence to group `debug_endpoints` - mainly for readability. Or
* B. Move at least the initial `?assertMatch(true, rpc(?NODE, aehttp_app, enable_internal_debug_endpoints, []))` in the init_per_testcase of `disabled_debug_endpoints`, and in end_per_testcase of `disabled_debug_endpoints` restore the config (may require carrying state from the init).

A is quicker, B is cleaner (also for case of test failure).

I suggest A, or at least clarifying in a comment how the order is guaranteed (so to avoid reader to consider this as test intermittently failing).

docs/release-notes/next/PT-162429019-disable-internal-debug-api-by-default.md
fabiankrol: Correct 👌

apps/aehttp/src/aehttp_app.erl
fabiankrol: 👍

apps/aeutils/priv/epoch_config_schema.json
fabiankrol: 👍

apps/aeutils/priv/epoch_config_schema.json
fabiankrol: 👍

PR #1904, opened 2018-12-07
PT-162445499 improve mining efficiency

hanssv:
This is all about doing more attempts in one CPU/GPU context. There are possibly a few CPUs fast enough to also gain from this. The caveat is to not produce too long running mining processes since it will behave badly from a network perspective and create micro-forks.

The PR does change how we interface the miner, instead of packaging key_header+nonce and send to miner we send only key_header and then pass the nonce as a separate (numeric) parameter. By doing this the miner can use the -r parameter to do multiple attempts in the same context. The PR also changes the parser slightly to pick up also the nonce from the miner output.

### DIFF
--- a/py/tests/integration/common.py
+++ b/py/tests/integration/common.py
@@ -83,19 +83,21 @@ def setup_node_with_tokens(node, beneficiary, blocks_to_mine):
     ext_api = external_api(node)
     int_api = internal_api(node)
 
+    top0 = ext_api.get_current_key_block()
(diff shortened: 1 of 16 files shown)


py/tests/integration/common.py
```diff
@@ -83,19 +83,21 @@ def setup_node_with_tokens(node, beneficiary, blocks_to_mine):
     ext_api = external_api(node)
     int_api = internal_api(node)
 
+    top0 = ext_api.get_current_key_block()
```
fabiankrol: 🎉

PR #1906, opened 2018-12-08
WARNING! OSX! 

Zwilla:
WARNING IF YOU USE /tmp/node and reboot your mac the folder is deleted, so you will lost your keys if you didn't backup them.

After first testing by the team I will close this pr

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -68,6 +68,8 @@ It is recommended that the partition where the node directory is has at least 10
 Open a Terminal window or get to the command line.
 
 Create a directory and unpack the downloaded package (you may need to amend the directory and/or file name of the package):
+WARNING IF YOU USE /tmp/node and reboot your mac the folder is deleted, so you will lost your keys if you didn't backup them.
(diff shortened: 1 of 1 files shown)


sennui:
Let's try to compose documentation in a coherent way, with one line of thinking without exclamation marks and side thoughts

Re the /tmp. I think it's not, so urgent - we mention it already and also - compilation is probably for users who are aware of the nature of /tmp directory

dincho:
Actually if you CAREFULLY! [read the documentation](https://github.com/aeternity/epoch/blob/master/docs/installation.md#deploy-node) you'll find out that it's documented:

> In the instructions below, the node is deployed in directory /tmp/node: **you may prefer to deploy the node in an alternative (and less ephemeral) location** - e.g. a node directory inside your home directory - by amending the instructions accordingly.

However, we decided to change the example path used in the documentation to minimize the confusion by non-tech savvy users.

dincho:
https://github.com/aeternity/epoch/pull/1908 should improve the UX

PR #1911, opened 2018-12-11
Multi GPU mining - in conductor

hanssv:
Support multiple miner processes in the conductor rather than hacking it in the miner.

### DIFF
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -173,19 +173,22 @@ init(Options) ->
     TopKeyBlockHash = aec_chain:top_key_block_hash(),
     Consensus = #consensus{micro_block_cycle = aec_governance:micro_block_cycle(),
                            leader = false},
-    State1 = #state{ top_block_hash = TopBlockHash,
+    {ok, Beneficiary} = get_beneficiary(),
+    State1 = #state{ top_block_hash     = TopBlockHash,
                      top_key_block_hash = TopKeyBlockHash,
-                     consensus = Consensus},
+                     consensus          = Consensus,
+                     beneficiary        = Beneficiary},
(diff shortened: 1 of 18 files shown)


apps/aecore/src/aec_conductor.erl
```diff
@@ -820,22 +871,32 @@ handle_micro_sleep_reply(ok, State) ->
...
+    start_mining(State);
 create_key_block_candidate(#state{top_block_hash = TopHash,
                                   beneficiary    = Beneficiary} = State) ->
     epoch_mining:info("Creating key block candidate on the top"),
     Fun = fun() ->
-              {aec_block_key_candidate:create(TopHash, Beneficiary), TopHash}
+                  {aec_block_key_candidate:create(TopHash, Beneficiary), TopHash}
           end,
-    dispatch_worker(create_key_block_candidate, Fun, State).
+    {State1, _Pid} = dispatch_worker(create_key_block_candidate, Fun, State),
```
sennui: I see you are matching now on `{State, Pid}`, but top branch of case in dispatch_worker returns State?!

hanssv: Well, we should never hit that one, but fixing it anyway... :-P

hanssv:
Just pressed the ones GitHub suggested 😂 It was late in the afternoon...

docs/cuda-miner.md
```diff
@@ -78,7 +78,19 @@ mining:
             hex_encoded_header: true
 ```
 
-After configuration could be started (or restarted if it's already running):
+If you have multiple GPU cards they can be addressed using the `-d <n>` flag with the cuda-miner. So if you want to use multiple cards
```
lucafavatella: Shall this addition rather be made at end of section "Multiple GPU devices"?

PR #1912, opened 2018-12-11
Make dependencies win32 compatible

tolbrino:
Also replaced erlexec with plain Erlang/OTP variant, to reduce reliance
on specific UNIX features.


tolbrino:
I've rebased this on latest master. All tests are fixed and ready to review. I've verified that it works on Windows as well.

apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -269,9 +266,6 @@ get_executable(#miner_config{executable = Executable}) ->
 get_extra_args(#miner_config{extra_args = ExtraArgs}) ->
     ExtraArgs.
 
-get_nice(#miner_config{nice = Nice}) ->
-    Nice.
```
zp-sd: So nice is no longer supported?

tolbrino: Not for now, I didn't bother re-implementing it since its usefulness kind of diminishes with the increased usage of pool mining.

zp-sd: So maybe we should either remove `nice` property from the config schema, or at least mark it as deprecated there. I mean `epoch_config_schema.json`, and removal from all sample configs from all tests?

tolbrino: Good point. Removing it would make the release incompatible with previous configurations, so I'd prefer marking it deprecated. Will do that.

tolbrino: I've made the requested changes.

rebar.lock
```diff
@@ -22,22 +22,19 @@
...
-  0},
  {<<"exometer_core">>,{pkg,<<"exometer_core">>,<<"1.5.7">>},0},
  {<<"folsom">>,{pkg,<<"folsom">>,<<"0.8.7">>},1},
  {<<"goldrush">>,{pkg,<<"goldrush">>,<<"0.1.9">>},1},
  {<<"gproc">>,{pkg,<<"gproc">>,<<"0.6.1">>},0},
+ {<<"hex2bin">>,{pkg,<<"hex2bin">>,<<"1.0.0">>},1},
  {<<"hut">>,{pkg,<<"hut">>,<<"1.2.1">>},1},
  {<<"idna">>,
-  {git,"https://github.com/aeternity/erlang-idna",
+  {git,"https://github.com/benoitc/erlang-idna",
```
zp-sd: What is our policy, do we want to have everything forked to our org? /cc @sennui

tolbrino: At some point I just stopped harassing @sennui for repo permissions :-)

Generally I think it would be beneficial to have a org mirror of all dependencies and a process for managing those. But that is a bigger and separate endeavour.

tolbrino:
@zp-sd The changes build on Windows, as can be seen [here](https://ci.appveyor.com/project/aeternity/epoch/builds/21257796), which is the WIP CI for Windows.

PR #1914, opened 2018-12-12
Log the miner beneficiary of received micro-blocks to allow analysis

ThomasArts:
When a node receives orphan light micro blocks it logs:
 ```
14:31:22.851 [info] Got invalid light micro_block (<<47,27,152,47,...,50,137,13,90>>): {error,orphan_blocks_not_allowed}
```

Most likely a new key block has arrived and these blocks are from the leader "just before" that is still pushing out micro-blocks. It would be nice to know that, though. Therefore the log is extended with trying to find out which miner beneficiary it belongs to. This may not be possible, because it could be a block of which even the corresponding key-block is missing, but if possible, we log it.

### DIFF
--- a/apps/aecore/src/aec_peer_connection.erl
+++ b/apps/aecore/src/aec_peer_connection.erl
@@ -972,7 +972,14 @@ handle_new_micro_block(S, Msg) ->
                     E = {error, _} ->
                         {ok, HH} = aec_headers:hash_header(Header),
                         epoch_sync:info("Got invalid light micro_block (~s): ~p", [pp(HH), E]),
-                        ok;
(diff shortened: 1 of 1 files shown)


ThomasArts:
Note that we on purpose only log the Erlang binary key and not the base58 encoded key. On one hand it would mean more computation in a probably already overloaded system, on the other hand it would create a dependency between modules that we rather not have depending on each other.

gorillainduction:
I don't agree. An orphan is not connected to genesis. So, there must be a missing block (micro or key). It is not related to having a new leader.

We do accept micro blocks on forks if it is connected to genesis since the fork might be a contender for being the new main chain.

gorillainduction:
@ThomasArts I see. Then it might make sense to include the key hash as well.

PR #1916, opened 2018-12-12
Pt 161990908 fate data serialization

happi:
Initial experiment with serialization.


apps/aefate/src/aefa_encoding.erl
```diff
@@ -0,0 +1,356 @@
+%% First draft of FATE serialization encoding/decoding.
+%% Goal is to experiment with the encoding.
+%% The FATE side of data is not set at all yet and will
+%% probably change.
+-module(aefa_encoding).
+
+-export([ deserialize/1
+        , serialize/1]).
+
+%% Very ecperimental:
```
dincho: The comment confirms that 🤣

apps/aefate/src/aefa_data.erl
```diff
@@ -0,0 +1,59 @@
...
+-export([format/1]).
+
+
+-spec format(fate_type()) -> iolist().
+format(I) when is_integer(I) ->
+    integer_to_list(I);
+format(true) -> "true";
+format(false) -> "false";
+format([]) -> "[]";
+format([E]) -> "[ " ++ format(E) ++ " ]";
```
ThomasArts: format(E) can return iolist and hence "++" won't work.

PR #1918, opened 2018-12-12
Release notes + documentation for Multi GPU

hanssv:


docs/cuda-miner.md
```diff
@@ -149,6 +137,22 @@ mining:
 
 The address of the device can be obtained by running `nvidia-smi`
 
+However, if you want to use multiple cards for GPU mining you can add another
+configuration option, `instances`, for example if you have two (2) GPU-cards:
+
+```
+mining:
```
dincho: I would use a full configuration as above, otherwise users might get confused

docs/cuda-miner.md
```diff
@@ -149,6 +137,22 @@ mining:
...
+    cuckoo:
+        miner:
+            instances: 2
+```
+
+*Note:* You should not have `-d` in `extra_args` if you are using the `instances` configuration option, it will be added automatically
+by the node.
+
+*Note:* If you are combining `repeats` and `instances` it is the *number of repeats per instance* that is configured! I.e. with 2
+instances and repeats 5 each GPU will run 5 attempts per run.
```
dincho: some basic punctuation would help to understand this sentence on first read (at least for me :) )

hanssv: I've use three full stops and one exclamation mark, this is a though crowd :-)

hanssv:
Tried to improve the commented sections...

PR #1931, opened 2018-12-17
Pt 162635989 internal data representation

happi:
Initial version of Fate data representation and serialization and deserilalization.


apps/aefate/src/aefa_data.hrl
```diff
@@ -0,0 +1,91 @@
...
+-define(FATE_ADDRESS(A), {address, A}).
+
+-ifdef(CHECKFATEDATA).
+
+%% Types
+-define(FATE_INTEGER_T, {integer,  integer()}).
+-define(FATE_BYTE_T,    {integer,  0..255}).
+-define(FATE_BOOLEAN_T, {boolean, true | false}).
+-define(FATE_NIL_T,     {nil}).
+-define(FATE_LIST_T(T), {list,    list(A::T) | ?FATE_NiL_T}).
```
UlfNorell: Misspelled `FATE_NIL_T`.

apps/aefate/src/aefa_data.hrl
```diff
@@ -0,0 +1,91 @@
...
+-define(FATE_UNIT_T,    {tuple,   {}}).
+-define(FATE_MAP_T,     {map,     #{ A::fate_none_map_type() => B::fate_type() }}).
```
UlfNorell: Won't these type variables (`A` and `B`) produce weird effects if you use multiple `?FATE_MAP_T` in a type spec?

apps/aefate/src/aefa_data.hrl
```diff
@@ -0,0 +1,91 @@
...
+-define(FATE_STRING_T,s  {string,  binary()}).
```
UlfNorell: Spurious `s`.

apps/aefate/src/aefa_data.hrl
```diff
@@ -0,0 +1,91 @@
...
+-define(MAKE_FATE_STRING(X),  {string, X}).
+
+
+-else. %% No extra tagging on Erlang values.
+
+-define(FATE_INTEGER_T, integer()).
+-define(FATE_BYTE_T,    0..255).
+-define(FATE_BOOLEAN_T, true | false).
+-define(FATE_NIL_T,     []).
+-define(FATE_LIST_T(T), list(A::T) | fate_nil()).
```
UlfNorell: `list(T)` includes nil.

apps/aefate/src/aefa_data.hrl
```diff
@@ -0,0 +1,91 @@
...
+-define(FATE_UNIT_T,    {tuple, {}}).
+-define(FATE_MAP_T,     #{ A::fate_none_map_type() => B::fate_type() }).
+-define(FATE_STRING_T,  binary()).
+-define(FATE_ADDRESS_T, {address, <<_:256>>}).
+-define(FATE_VARIANT_T(T), {variant, ?FATE_BYTE_T, fate_tuple(T)}).
```
UlfNorell: Why does this take a type argument? Typically constructor arguments will have different types. Also having an extra `tuple` tag seems unnecessary.

On a different level it seems strange to have a type for each variant constructor, rather than a type for the full type. I would have expected variant types to be represented by a union type in Erlang.

apps/aefate/src/aefa_data.hrl
```diff
@@ -0,0 +1,91 @@
...
+
+-else. %% No extra tagging on Erlang values.
+
+-define(FATE_INTEGER_T, integer()).
+-define(FATE_BYTE_T,    0..255).
+-define(FATE_BOOLEAN_T, true | false).
+-define(FATE_NIL_T,     []).
+-define(FATE_LIST_T(T), list(A::T) | fate_nil()).
+-define(FATE_UNIT_T,    {tuple, {}}).
+-define(FATE_MAP_T,     #{ A::fate_none_map_type() => B::fate_type() }).
```
UlfNorell: Shouldn't `A` and `B` be type arguments here?

apps/aefate/src/aefa_data.hrl
```diff
@@ -0,0 +1,91 @@
...
+-define(FATE_INTEGER_VALUE(X), (X)).
+-define(FATE_LIST_VALUE(X), (X)).
+-define(FATE_STRING_VALUE(X), (X)).
+-define(FATE_MAP_VALUE(X), (X)).
+-define(FATE_MAP_SIZE(X), (map_size(X))).
+-define(FATE_STRING_SIZE(X), (byte_size(X))).
+-define(FATE_TRUE,  true).
+-define(FATE_FALSE, false).
+-define(FATE_NIL,   []).
+-define(FATE_VOID,  void).
```
UlfNorell: Isn't `void` the empty type? If so, it shouldn't have any values. If not, how is it different from `unit`?

apps/aefate/src/aefa_data.erl
```diff
@@ -0,0 +1,117 @@
...
+    case ?FATE_LIST_VALUE(L) of
+        [E] -> "[ " ++ format(E) ++ " ]";
+        _ -> "[ " ++ format_list(L)
+    end;
+format(?FATE_UNIT) -> "unit";
+format(?FATE_TUPLE(T)) -> "{ " ++ [format(E) ++ " " || E <- tuple_to_list(T)] ++ "}";
+format(S) when ?IS_FATE_STRING(S) -> [S];
+format(?FATE_VARIANT(Tag, T)) ->
+    "( " ++ format(Tag) ++ ", "
+        ++ [format(E) || E <- tuple_to_list(T)]
```
ThomasArts: May want to add an extra space (or comma) between elements

apps/aefate/src/aefa_data.erl
```diff
@@ -0,0 +1,117 @@
...
+                     | ?FATE_TUPLE_T({T,T,T})
+                     | ?FATE_TUPLE_T({T,T,T,T,T})
+                     | ?FATE_TUPLE_T({T,T,T,T,T,T})
+                     | ?FATE_TUPLE_T({T,T,T,T,T,T,T})
+                     | ?FATE_TUPLE_T({T,T,T,T,T,T,T,T})
+                     | ?FATE_TUPLE_T({T,T,T,T,T,T,T,T,T})
+                     | ?FATE_TUPLE_T({T,T,T,T,T,T,T,T,T,T})
+                     | ?FATE_TUPLE_T({T,T,T,T,T,T,T,T,T,T,T})
+                     | ?FATE_TUPLE_T({T,T,T,T,T,T,T,T,T,T,T,T})
+                     | ?FATE_TUPLE_T({T,T,T,T,T,T,T,T,T,T,T,T,T}).
```
UlfNorell: This doesn't look very nice, and I'm not sure I see what you are trying to achieve here. Why do we need Erlang types for the VM types in the first place? Won't all the code handle values of arbitrary VM types?

apps/aefate/src/aefa_data.erl
```diff
@@ -0,0 +1,117 @@
...
+-spec format(fate_type()) -> iolist().
+format(I) when ?IS_FATE_INTEGER(I) -> integer_to_list(?MAKE_FATE_INTEGER(I));
+format(?FATE_VOID) -> "void";
+format(?FATE_TRUE) -> "true";
+format(?FATE_FALSE) -> "false";
+format(?FATE_NIL) -> "[]";
+format(L) when ?IS_FATE_LIST(L) ->
+    case ?FATE_LIST_VALUE(L) of
+        [E] -> "[ " ++ format(E) ++ " ]";
+        _ -> "[ " ++ format_list(L)
```
UlfNorell: `L` should be `?FATE_LIST_VALUE(L)`. And you don't need the special case for the singleton list.

apps/aefate/src/aefa_data.erl
```diff
@@ -0,0 +1,117 @@
...
+format(?FATE_ADDRESS(Address)) -> base58:binary_to_base58(Address);
+format(V) -> exit({not_a_fate_type, V}).
+
+format_list([]) -> " ]";
+format_list([E]) -> format(E) ++ " ]";
+format_list([H|T]) -> format(H) ++ ", " ++ format_list(T).
+
+format_kvs([]) -> "";
+format_kvs([{K,V}]) -> "( " ++ format(K) ++ format(V) ++ " )";
+format_kvs([{K,V} | Rest]) -> "( " ++ format(K) ++ format(V) ++ " ), " ++ format_kvs(Rest).
```
UlfNorell: No separator between `K` and `V`?

apps/aefate/src/aefa_data.erl
happi: Fixed, but should really be commas.
Better formatting will come in next ticket and PR.

apps/aefate/src/aefa_encoding.erl
```diff
@@ -0,0 +1,349 @@
...
+      (<< << (value_to_typerep(Type))/binary >> || Type <- L >>)/binary
+    >>.
+
+
+type_to_typerep(unit)        -> <<0>>;
+type_to_typerep(integer)     -> <<1>>;
+type_to_typerep(boolean)     -> <<2>>;
+type_to_typerep(address)     -> <<3>>;
+type_to_typerep(string)      -> <<4>>;
+type_to_typerep(empty_tuple) -> <<5, 0>>;
```
UlfNorell: Why is there a special constructor for the empty tuple? The code below works fine for `{tuple, {}}`. Also these typereps aren't defined anywhere that I can see.

apps/aefate/src/aefa_encoding.erl
```diff
@@ -0,0 +1,349 @@
...
+type_to_typerep(empty_map)   -> <<10>>;
+type_to_typerep({map, A, B}) -> <<8, (type_to_typerep(A))/binary, (type_to_typerep(B))/binary >>;
+type_to_typerep({variant, Cases}) ->
+    <<9, (length(Cases)),
+      <<
+        << (length(Types)),
+           << << (type_to_typerep(Type))/binary >> || Type <- Types >>
+        >>
+        || Types <- Cases >>
+    >>.
```
UlfNorell: We're going to need typereps to be first class values, so I don't think it makes sense to have this function (and `deserialize_typerep` below) separate from `serialize` and `deserialize`.

apps/aefate/src/aefa_data.hrl
happi: Fixed.

apps/aefate/src/aefa_data.hrl
happi: Maybe. They should not be needed so I removed them,

apps/aefate/src/aefa_data.hrl
happi: Fixed.

apps/aefate/src/aefa_data.hrl
happi: But not really in the other implementation where we wrap everything in an extra layer of tagging.

apps/aefate/src/aefa_data.hrl
happi: The type argument is to separate fate types with maps from fate types without maps to have a type for map keys witouth maps.

I dont really get the second paragraph. There is not a type for each variant, it is just either a variant including a map or not.

apps/aefate/src/aefa_data.erl
happi: I know. It does not look nice at all.
It is so that we can separeate fate types with maps from fate types without maps.

The code will handle arbitrary VM types (at least in most cases). The goal is to limit what a valid Fate type is and to get the limitation that map keys can not have maps in the type description.

The types are also used by Thomas quick check model to generate data.

apps/aefate/src/aefa_data.erl
happi: Fixed.

apps/aefate/src/aefa_encoding.erl
happi: Typreps will be completely reworked.

apps/aefate/src/aefa_data.erl
happi: Not for now. Will probably change. Not settled on format yet, just used for some debugging.

apps/aefate/src/aefa_encoding.erl
happi: Agree. I will rework the type reps and see how to worke them in as first class values and if we should have typereps for variant types always together with their values.

apps/aefate/src/aefa_data.hrl
happi: But not really needed here. Neither is the type variable.

apps/aefate/src/aefa_encoding.erl
happi: Removed typereps for now.

apps/aefate/src/aefa_encoding.erl
happi: Removed typreps for now.

happi:
Starting a new PR

PR #1940, opened 2018-12-18
Remove dependency on aec_hash by performing hash computation directly

ThomasArts:


apps/aesophia/src/aeso_abi.erl
```diff
@@ -163,15 +164,15 @@ function_type_info(Name, Args, OutType) ->
...
-                                   , aeso_data:to_binary(ArgType)
-                                   , aeso_data:to_binary(OutType)
-                                   ])).
-
+    Bin =  iolist_to_binary([ Name
+                            , aeso_data:to_binary(ArgType)
+                            , aeso_data:to_binary(OutType)
+                            ]),
+    %% Calculate a 256 bit digest BLAKE2b hash value of a binary
+    {ok, Hash} = enacl:generichash(?HASH_SIZE, Bin),
```
lucafavatella: I see same as https://github.com/aeternity/epoch/blob/ebb6ff3f8dca719ac02944763a81184cbbbc9440/apps/aecore/src/aec_hash.erl#L33
Ok.

PR #1942, opened 2018-12-18
Pt 162445499 allow multiple simultaneous miners

zp-sd:
TODO:
- [x] Update release notes
- [x] Run all system tests - `Skipped 2 (2, 0) tests. Passed 29 tests.`
- [x] Update docs
- [x] Update infrastructure/envs configs - https://github.com/aeternity/infrastructure/pull/203

### DIFF
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -767,16 +767,16 @@ start_mining(#state{key_block_candidates = [{HeaderBin, Candidate} | Candidates]
         Instance ->
             epoch_mining:info("Starting miner on top of ~p", [State#state.top_block_hash]),
             Target            = aec_blocks:target(Candidate#candidate.block),
-            Nonce             = Candidate#candidate.nonce,
             MinerConfig       = Instance#miner_instance.config,
             AddressedInstance = Instance#miner_instance.instance,
+            Nonce             = aec_pow:trim_nonce(Candidate#candidate.nonce, MinerConfig),
             Info              = [{top_block_hash, State#state.top_block_hash}],
             aec_events:publish(start_mining, Info),
             Fun = fun() ->
                           {aec_mining:mine(HeaderBin, Target, Nonce, MinerConfig, AddressedInstance)
                           , HeaderBin}
                   end,
-            Candidate1 = register_miner(Candidate, MinerConfig),
+            Candidate1 = register_miner(Candidate, Nonce, MinerConfig),
             State1 = State#state{key_block_candidates = [{HeaderBin, Candidate1} | Candidates]},
             {State2, Pid} = dispatch_worker(mining, Fun, State1),
             State3 = register_miner_instance(Instance, Pid, State2),
@@ -784,7 +784,7 @@ start_mining(#state{key_block_candidates = [{HeaderBin, Candidate} | Candidates]
             start_mining(State3)
     end.
 
-register_miner(Candidate = #candidate{nonce = Nonce, refs  = Refs}, MinerConfig) ->
+register_miner(Candidate = #candidate{refs  = Refs}, Nonce, MinerConfig) ->
     NextNonce = aec_pow:next_nonce(Nonce, MinerConfig),
     Candidate#candidate{refs  = Refs + 1,
                         nonce = NextNonce}.
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -335,7 +335,9 @@ try_fetch_and_make_candidate() ->
 
 make_key_candidate(Block) ->
     HeaderBin = aec_headers:serialize_to_binary(aec_blocks:to_header(Block)),
+    Nonce     = aec_pow:pick_nonce(),
     {HeaderBin, #candidate{ block    = Block,
+                            nonce    = Nonce,
                             top_hash = aec_blocks:prev_hash(Block) }}.
 
 make_micro_candidate(Block) ->
@@ -606,7 +608,6 @@ create_miner_instances(MinerConfig, FirstId) ->
 
 create_miner_instance(Id, Instance, Config) ->
     #miner_instance{id       = Id,
-                    nonce    = aec_pow:pick_nonce(),
                     instance = Instance,
                     config   = Config,
                     state    = available}.
@@ -621,8 +622,8 @@ get_first_available_instance([#miner_instance{state = available} = Instance | _I
 get_first_available_instance([_Instance | Instances]) ->
     get_first_available_instance(Instances).
 
-register_miner_instance(Instance, Pid, Nonce, #state{miner_instances = MinerInstances0} = State) ->
-    UpdatedInstance = Instance#miner_instance{state = Pid, nonce = Nonce},
+register_miner_instance(Instance, Pid, #state{miner_instances = MinerInstances0} = State) ->
+    UpdatedInstance = Instance#miner_instance{state = Pid},
     MinerInstances  = lists:keyreplace(Instance#miner_instance.id, #miner_instance.id, MinerInstances0, UpdatedInstance),
     State#state{miner_instances = MinerInstances}.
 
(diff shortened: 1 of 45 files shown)


docs/release-notes/next/PT-162445499-multiple-simultaneous-miners.md
```diff
@@ -0,0 +1,2 @@
+- Changes cuckoo miner config and processing to allow multiple simultaneous miners.
+  This impacts mining configuration (changes `miner` param to `miners`).
```
lucafavatella: I see:
* `edge_bits` moved in common to miners
* `extra_args` made optional
* `extra_args` default changed
* `miner` changed to `miners`

May I suggest:
* Considering to enable old config to still work, though marking relevant sections as deprecated in the schema.
* Add user config test with old config e.g. `apps/aeutils/test/data/epoch_deprecated_miner_config.yaml`
* Add functional test e.g. new system test suite with old config that mines a couple of blocks?

zp-sd: It is doable, but will make the code a bit more complex to maintain both old and new configs. For how long would you want to maintain the deprecated config then? 🤔

lucafavatella: I would expect that 1.2.0 recommends in all docs the new was of specifying miner config and has the deprecation in the user config schema. Then 1.3.0 drops the deprecated config. Cc @dincho @sennui

zp-sd: Another option would be to set `"additionalProperties" : true` in `mining` > `cuckoo` section of the config. Then if user deploys the code as it is in the PR, and keeps the old config, she should notice that the default miner is working instead of what she wants, and will adjust config accordingly? So in this aspect it is backwards compatible (node will not crash with old config), but maybe is even more confusing?! I see @lucafavatella 's point, but am not 100% convinced that this is how we should approach it...

@hanssv @dincho thoughts?

apps/aecore/src/aec_conductor.erl
```diff
@@ -576,26 +580,56 @@ erase_worker(Pid, #state{workers = Workers} = State) ->
...
+        AddressedInstances when is_list(AddressedInstances) ->
+            lists:foldl(
+              fun(AddressedInstance, {Acc, Id}) ->
+                      {[create_miner_instance(Id, AddressedInstance, MinerConfig) | Acc], Id + 1}
+              end, {[], FirstId}, AddressedInstances)
+    end.
+
+create_miner_instance(Id, Instance, Config) ->
+    #miner_instance{id       = Id,
+                    nonce    = aec_pow:pick_nonce(),
```
gorillainduction: We are just trusting randomness for not rechecking nonces more than once, since we bind the nonce to the instance.

Wouldn't it make sense to keep the nonces in the candidate?

zp-sd: I moved nonces to the miner as in my initial idea the whole miner config wasn't kept in the conductor. There was only some reference of the miner, and the `repeats` param of the particular miners was unknown (in the end this "miner references" idea didn't play well with parsing array of elements from yaml config..).

But now, since we have the whole miner config (together with `repeats` param) in the conductor, it is feasible, and I can change that 👍 There should not be much difference though, randomness should be good enough, given how ofter we should receive new candidates on top of key/micro blocks?!

gorillainduction: I agree that it probably wouldn't make a practical difference, but it would make sense to eliminate the possibility of getting the same nonce in different miners, since it is not a costly operation.

I could imagine a later situation where the randomness could be compromised since we are not explicitly seeding the random process, which in turn could lead to all miners working on the same nonce range. It seems safe now, though.

zp-sd:
@hanssv @gorillainduction @lucafavatella Please revisit!

docs/cuda-miner.md
```diff
@@ -213,6 +213,29 @@ by the node.
...
 instances and repeats 5 each GPU will run 5 attempts per run.
 
+### Multiple GPU miners
+
+If you want to address different GPU-cards with different miners' configurations, you can set up multiple miners in the config.
+For example, if you have 4 GPU-cards, and you want to address 0 and 1 with different config than 2 and 3, you can use the following config:
+```yaml
+mining:
+    cuckoo:
+        edge_bits: 29
```
sennui: this line is breaking check_config

docs/configuration.md
```diff
@@ -165,12 +165,31 @@ If you have 16 cores, you could (loosely spoken) assign 14 of them to mining usi
...
+For example, to combine CPU miner with CUDA miner the following configuration can be used:
+```yaml
+mining:
+    beneficiary: "beneficiary_pubkey_to_be_replaced"
+    cuckoo:
+        edge_bits: 29
+        miners:
+            - executable: mean29-generic
+              extra_args: -t 2
+            - executable: cuda29
```
lucafavatella: Without `executable_group: aecuckooprebuilt` this will not work unless user built miner, and it seems a misleading assumption in an example (that) that in a doc (this) separate from the main cuda-miner.md

More generally: Can this section be deleted altogether, or reduced to a single sentence redirecting the user to the cuda-miner.md (that - I double checked - is included in the release binary as per rebar.config)? I am concerned that too many example configurations are too costly to maintain as correct examples (unless we automate that, that is a out of scope).

docs/cuda-miner.md
zp-sd: I don't know how this would be possible, let's chat offline..

sennui: it's OK now

scripts/install-ubuntu1804.sh
```diff
@@ -80,10 +80,10 @@ mining:
     beneficiary: "${BENEFICIARY}"
     autostart: true
     cuckoo:
```
lucafavatella: @tolbrino I am not sure if this is CI tested anywhere. Cc @zp-sd

apps/aecore/src/aecore.app.src
```diff
@@ -72,6 +72,7 @@
                      {110, {aecore_app, check_env, []}},
                      {110, {aehttp_app, check_env, []}},
                      {110, {aec_hard_forks, check_env, []}},
+                     {110, {aec_pow_cuckoo, check_env, []}},
```
lucafavatella: I see this hooks the checks tested in `aeu_env_tests`. Ok.

apps/aeutils/priv/epoch_config_schema.json
```diff
@@ -417,42 +476,45 @@
...
                                     "description" : "Executable binary of the miner. If the executable group is 'aecuckoo', the options are: \"mean29-generic\" (memory-intensive), \"mean29-avx2\" (memory-intensive, benefits from faster CPU supporting AVX2 instructions), \"lean29-generic\" (CPU-intensive, useful if memory-constrained), \"lean29-avx2\" (CPU-intensive, useful if memory-constrained, benefits from faster CPU supporting AVX2 instructions). For the other executable groups, please refer to the user documentation.",
                                     "type" : "string",
-                                    "default": "mean29-generic"
+                                    "default" : "mean29-generic"
                                 },
                                 "extra_args" : {
                                     "description" : "Extra arguments to pass to the miner executable binary. The safest choice is specifying no arguments i.e. empty string.",
                                     "type" : "string",
-                                    "default": "-t 1"
+                                    "default" : ""
```
lucafavatella: This is the only change in previous behaviour that I found based on code inspection.

docker/epoch_node1_mean15.yaml
```diff
@@ -21,10 +21,9 @@ mining:
     beneficiary_reward_delay: 2
     autostart: true
     cuckoo:
-        miner:
```
dincho: this changes was not very good idea, because now docker-compose (master) cannot be used with stable (i.e. 1.1.0) docker image.

PR #1943, opened 2018-12-18
Pt 162731282 more deps unraffled

ThomasArts:
Small steps. This PR moves some code out of aevm into aebytecode and reduces therewith some dependencies of the compiler on aevm.

The other calls to aevm are in aeso_data and aeso_data has to be moved into aevm somehow. This is too large for this PR and will follow later.

### DIFF
--- a/apps/aebytecode/src/aeb_primops.erl
+++ b/apps/aebytecode/src/aeb_primops.erl
@@ -1,0 +1,31 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2018, Aeternity Anstalt
+%%% @doc
+%%%     Handle interaction with the aeternity chain
+%%%     through calls to AEternity primitive operations at address 0.
+%%% @end
+%%% Created : 18 Dec 2018
+%%%-------------------------------------------------------------------
+
+-module(aeb_primops).
+-export([ is_local_primop_op/1
+        , op_needs_type_check/1
+        ]).
+
+-include("aeb_opcodes.hrl").
+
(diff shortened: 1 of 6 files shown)


apps/aesophia/src/aeso_ast_infer_types.erl
```diff
@@ -1525,7 +1525,8 @@ pp({qid, _, Name}) ->
 pp({con, _, Name}) ->
     Name;
 pp({uvar, _, Ref}) ->
-    ["?" | lists:sublist(base58:binary_to_base58(term_to_binary(Ref)), 43, 3)];
+    %% Show some unique representation
+    ["?u" | integer_to_list(erlang:phash2(Ref, 16384)) ];
```
lucafavatella: I expect this changes slightly the error reason in response body of `/debug/contracts/code/compile`. Though this is backward compatible - ok for 1.2.0 cc @noandrea .
May be worth a line in release notes.

apps/aebytecode/src/aeb_primops.erl
```diff
@@ -0,0 +1,31 @@
...
+%%%-------------------------------------------------------------------
+
+-module(aeb_primops).
+-export([ is_local_primop_op/1
+        , op_needs_type_check/1
+        ]).
+
+-include("aeb_opcodes.hrl").
+
+is_local_primop_op(Op) when ?PRIM_CALL_IN_MAP_RANGE(Op) -> true;
```
lucafavatella: I see simply moved - ok.

PR #1944, opened 2018-12-18
Pt 162474517 plant hard fork

gorillainduction:
* Clean up usage of macros for genesis/protocol versions.
* Introduce an unconditional hard fork at height 1 (will need to change before release 

https://www.pivotaltracker.com/story/show/162474517


gorillainduction:
@lucafavatella I can spend some time on this today. It seems like a reasonable approach.

PR #1946, opened 2018-12-19
Restore building on running `make` without args

lucafavatella:
Follows https://github.com/aeternity/epoch/commit/0b52ae5edead34f4f06b9f42ab8cf4416a77d002#diff-b67911656ef5d18c4ae36cb6741b7965R19


lucafavatella:
> One of those situations where `.DEFAULT` would have been useful.

@tolbrino Not sure if that or `.DEFAULT_GOAL`.

PR #1954, opened 2018-12-19
Pt 162578740 fate implementation spike

happi:
Experiment with Fate - transaction engine.
Simple operations, control flow, basic blocks, call/return implemented.


apps/aefate/src/aefa_fate.erl
```diff
@@ -0,0 +1,475 @@
...
+-define(trace(I,S), S).
+-endif.
+
+run(What, Chain) ->
+    EngineState = setup_engine(What, Chain),
+    execute(EngineState).
+
+execute(EngineState) ->
+    Instructions = current_bb(EngineState),
+    loop(Instructions, EngineState).
```
ThomasArts: It seems a bit error prone to have 2 copies of current_bb, one update in the loop in EngineState and one updated in the recurser...

happi: Yes. Not so happy with that. Ill try to fix it.

apps/aefate/src/aefa_fate.erl
```diff
@@ -0,0 +1,475 @@
...
+    case maps:get(Name, Functions, void) of
+        void ->  throw({error, {trying_to_call_function, Name}});
+        {Signature, _Code} -> Signature
+    end.
+
+check_signature({ArgTypes, _RetSignature},
+                #{ accumulator := Acc
+                 , accumulator_stack := Stack}
+               =_EngineState) ->
+    case check_arg_types(ArgTypes, [Acc, Stack]) of
```
UlfNorell: `[Acc | Stack]`

apps/aefate/src/aefa_fate.erl
```diff
@@ -0,0 +1,475 @@
...
+
+push_arguments(Args, #{ accumulator := X
+                      , accumulator_stack := XS} = ES) ->
+    push_arguments(lists:reverse(Args), X, XS, ES).
+
+push_arguments([], Acc, Stack, ES) ->
+    ES#{ accumulator := Acc
+       , accumulator_stack := Stack};
+push_arguments([A|As], Acc, Stack, ES ) ->
+    push_arguments(As, A, [Acc, Stack], ES).
```
UlfNorell: `[Acc | Stack]` ?

apps/aefate/src/aefa_fate.erl
happi: Indeed.

apps/aefate/src/aefa_fate.erl
```diff
@@ -0,0 +1,475 @@
...
+loop(Instructions, EngineState) ->
+    case step(Instructions, EngineState) of
+        {stop, FinalState} ->
+            FinalState;
+        {jump, BB, NewState} ->
+            {NewInstructions, State2} = jump(BB, NewState),
+            loop(NewInstructions, State2)
+    end.
+
+step([], EngineState) ->
```
gorillainduction: Do we really want fall-through? I could imagine always having a jump instruction at the end of BBs.

lucafavatella:
I took the liberty of labelling as WIP as I understand not for merging yet.

PR #1957, opened 2018-12-20
Pt 162761916 unraffle aeso abi

ThomasArts:
Refactorings to remove dependencies from aesophia to other parts of aeternity


rvirding:
Why move the parsing of sophia types to the top-level, I'm just curious.

ThomasArts:
I had to move it out (aeso_data will become part of aevm) and this seemed a good module to put it in. I could have created a completely new module... but for one function that seemed weird

PR #1960, opened 2018-12-20
Add single command install script (bash)

dincho:
in scope of https://www.pivotaltracker.com/story/show/162607658

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -45,7 +45,7 @@ install_deps_ubuntu() {
 
     if [[ "$OS_RELEASE" = "16.04" ]]; then
         sudo apt-get install -y build-essential
-        $LIB_VERSION=1.0.16
+        LIB_VERSION=1.0.16
         wget https://download.libsodium.org/libsodium/releases/libsodium-${LIB_VERSION}.tar.gz
         tar -xf libsodium-${LIB_VERSION}.tar.gz && cd libsodium-${LIB_VERSION} &&
         ./configure && make && sudo make install && sudo ldconfig


docs/installation-windows.md
```diff
@@ -28,7 +28,7 @@ Alternatively, you can also download the individual scripts:
 
 - `scripts/windows/setup-wsl.ps1`
 - `scripts/windows/wsl-run.ps1`
-- `scripts/install-ubuntu1804.sh`
```
lucafavatella: I remember changes in concurrent https://github.com/aeternity/epoch/pull/1942 though shall not be major.

docs/installation.md
```diff
@@ -1,3 +1,13 @@
+# Quick install
+
+Run below command to install latest version of aeternity node.
+
+```bash
+wget https://raw.githubusercontent.com/aeternity/epoch/master/scripts/install.sh && chmod +x install.sh && ./install.sh 1.1.0
```
lucafavatella: Do I understand correctly that this is meant to be the same as the content of VERSION in the tree? If yes, please update it to 1.2.0.

Also, I see 1.1.0 below in this file while master has 1.2.0. Could you please rebase on current master?

docs/installation-windows.md
```diff
@@ -28,7 +28,7 @@ Alternatively, you can also download the individual scripts:
...
+- `scripts/install.sh`
```
lucafavatella: Is this hooked to CI in any way?

docs/installation.md
lucafavatella: I am a bit concerned of this line containing `master` in a doc meant to be tagged. Is this rather meant to be reformulated as a section in the /README.md pointing to a wiki page?

scripts/install.sh
```diff
@@ -0,0 +1,98 @@
...
+    echo -e "\nCleanup...\n"
+    rm "${TEMP_RELEASE_FILE}"
+}
+
+if [[ "$OSTYPE" = "linux-gnu" && $(lsb_release -i -s) = "Ubuntu" ]]; then
+    install_deps_ubuntu
+    install_node "https://github.com/aeternity/epoch/releases/download/v${RELEASE_VERSION}/epoch-${RELEASE_VERSION}-ubuntu-x86_64.tar.gz"
+elif [[ "$OSTYPE" = "darwin"* ]]; then
+    install_deps_osx
+    install_node "https://github.com/aeternity/epoch/releases/download/v${RELEASE_VERSION}/epoch-${RELEASE_VERSION}-osx-10.13.6.tar.gz"
```
lucafavatella: This 10.13.6 seems fragile. Shall this rather be passed as arg, and doc shall advise user on value to pass?

docs/installation.md
lucafavatella: Can two CI jobs (one on Linux and one on OSX) be created (e.g. on master nightly) that:
* Check that the one line doc that user sees (either in docs/installation.md or README.md or wiki page) is the one that the CI plans to run (e.g. with `grep '^CONTENTOFTHISLINE$'`);
* Run script.
* Perform minimal successive steps for getting running node.
* Run smoke test on release.
  * For this we used to have smoke test of release package in CI - not sure if still present, may need refreshing and reenabling it.

scripts/install.sh
```diff
@@ -0,0 +1,98 @@
...
+    sudo apt-get update
+    sudo apt-get upgrade -y
+    sudo apt-get install -y curl libssl1.0.0
+
+    if [[ "$OS_RELEASE" = "16.04" ]]; then
+        sudo apt-get install -y build-essential
+        wget https://download.libsodium.org/libsodium/releases/libsodium-1.0.16.tar.gz
+        tar -xf libsodium-1.0.16.tar.gz && cd libsodium-1.0.16 &&
+        ./configure && make && sudo make install && sudo ldconfig
+        cd .. && rm -rf libsodium-1.0.16 && rm libsodium-1.0.16.tar.gz
```
lucafavatella: Lots of hardcoded `1.0.16` - would be nice to group them in a variable though may be debatable so ok.

docs/installation-windows.md
dincho: no

scripts/install.sh
dincho: I'd rather vote to change the package name, we don't add "16.04.4" in the Ubuntu package name for example.

docs/installation.md
dincho: CircleCI is using 14.04 ..

scripts/install.sh
dincho: Still, I don't see a problem, if for some reason we update the OSX build env, this file must be updated as well. Why we should bother users with our package naming?

lucafavatella: My point is that the script is parametrized by release version, though infrastructure used for building new release version may have bumped macOS version too. So why parametrizing one but not the other?

I am still trying to understand how user will retrieve and experience this script, so please feel free to ignore this specific comment.

docs/installation.md
dincho: Well, maybe it's time to start maintaining `latest` (or `stable` or whatever name)  branch that links to latest release so it can be used in the documentation/tutorials.

dincho: I tested with 1.1.0 and just forgot it .. fixed in 8dc3f4d5

docs/installation-windows.md
dincho: it's ok after the rebase.

scripts/install.sh
dincho: it was copy-pasta :) see 99b5a776

docs/installation.md
dincho: After https://www.pivotaltracker.com/story/show/162785706 I'll remove the version by default

dincho: I don't want to hide this "on-liner" in the wiki. Not sure about master reference tho, maybe it's better to reference on tag, until we have "latest" ?

I don't like having commands even in the master docs that don't actually work, like `install 1.2.0` ..

lucafavatella: > I don't like having commands even in the master docs that don't actually work (yet), like `install 1.2.0` ..

This means there is a small risk that these required changes may go unnoticed before tagging. It would be great if you added specific notes in https://github.com/aeternity/epoch/wiki/Release-process

jur0:
@tolbrino yes, I worked on debian packages some time ago, but then started with a different task. @valkov-ivaylo took over the debian packages.

dincho:
> Another thing: Since there is only a small set of supported OS's, why not add epoch packages to those package managers. `brew` for macOS would be easy, Ubuntu only marginally harder. @jur0 started with this some weeks back.

It's work in-progress, this PR is "temporary solution"

dincho:
> There are plenty of such scripts out there already which work reasonably well, maybe we could also piggy-back on one of those and re-use them?

Could you please post an example ?

docs/installation.md
```diff
@@ -1,3 +1,13 @@
+# Quick install
+
+Run below command to install latest version of aeternity node.
+
+```bash
+wget https://raw.githubusercontent.com/aeternity/epoch/master/scripts/install.sh && chmod +x install.sh && ./install.sh 1.2.0
```
lucafavatella: s/master/v1.2.0

docs/installation.md
```diff
@@ -1,3 +1,13 @@
+# Quick install
+
+Run below command to install latest version of aeternity node.
+
+```bash
+wget https://raw.githubusercontent.com/aeternity/epoch/v1.2.0/scripts/install.sh && chmod +x install.sh && ./install.sh 1.2.0
```
lucafavatella: Mentioned for smoke testing in release notes - @dincho please review.
https://github.com/aeternity/epoch/wiki/Release-process/_compare/9e4071527559570bb5d852e01f0c9ad0c2a04072...a779f4fa113135666621ec1d015b2464e9eeef33

tolbrino:
> > There are plenty of such scripts out there already which work reasonably well, maybe we could also piggy-back on one of those and re-use them?
> 
> Could you please post an example ?

- https://github.com/DataDog/datadog-agent/blob/master/cmd/agent/install_script.sh
- https://github.com/teddysun/shadowsocks_install/blob/master/shadowsocks-all.sh
- https://github.com/creationix/nvm/blob/v0.33.11/install.sh

dincho:
> > > There are plenty of such scripts out there already which work reasonably well, maybe we could also piggy-back on one of those and re-use them?
> > 
> > 
> > Could you please post an example ?
> 
> * https://github.com/DataDog/datadog-agent/blob/master/cmd/agent/install_script.sh
> * https://github.com/teddysun/shadowsocks_install/blob/master/shadowsocks-all.sh
> * https://github.com/creationix/nvm/blob/v0.33.11/install.sh

I don't see anything general and/or re-usable in that scripts.

tolbrino:
@dincho Ok, fine by me. I was just reminded that I've seen this kind of script often enough.

scripts/install.sh
```diff
@@ -0,0 +1,99 @@
...
+install_deps_ubuntu() {
+    OS_RELEASE=$(lsb_release -r -s)
+    echo -e "\nPrepare host system and install dependencies ...\n"
+    sudo apt-get update
+    sudo apt-get upgrade -y
+    sudo apt-get install -y curl libssl1.0.0
+
+    if [[ "$OS_RELEASE" = "16.04" ]]; then
+        sudo apt-get install -y build-essential
+        $LIB_VERSION=1.0.16
```
dincho: `$` typo!?

dincho:
Manually tested latest changes on osx, 1604 and 1804.

PR #1963, opened 2018-12-20
Plant version and tag of first Minerva unstable release

lucafavatella:

PR #1966, opened 2018-12-20
Discriminate CI jobs meant for stable tags only

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/162475176

This PR is for `master`.
It depends on https://github.com/aeternity/epoch/pull/1963 to be merged in `minerva` (so to confirm tag scheme).

TODO:
* [ ] Test that on stable tag `/^v1\..*$/` release workflow works as expected.
  * @dincho How am I meant to test this? Shall we merge it, and at next stable tag&release if we spot any issues we fix CI quickly?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,6 +71,8 @@ references:
           && cd libsodium-src \
           && ./configure && make -j$(nproc) && sudo make install && sudo ldconfig
 
+  tag_regex: &tag_regex /^v.*$/
+
   rebar_cache_key: &rebar_cache_key rebar-cache-{{ checksum "rebar.lock" }}-{{ checksum "rebar.config" }}
   restore_rebar_cache: &restore_rebar_cache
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -72,6 +72,7 @@ references:
           && ./configure && make -j$(nproc) && sudo make install && sudo ldconfig
 
   tag_regex: &tag_regex /^v.*$/
+  stable_tag_regex: &stable_tag_regex /^v1\..*$/
```
lucafavatella: See https://github.com/aeternity/epoch/pull/1967 for companion regex.

lucafavatella:
> I see that API docs, packages (to github and s3), docker image (tag) are not uploaded for minerva but only for stable - OK

@dincho Please see https://www.pivotaltracker.com/story/show/162793073 for re-enabling. (I did not detail S3.)

PR #1969, opened 2018-12-20
Add events to Sophia compiler

hanssv:
Events can now be defined and emitted:

```
contract Events =
  datatype event =
      Event1(indexed int, indexed int, string)
    | Event2(string, indexed address)

  function f1(x : int, y : string) =
    Chain.event(Event1(x, x+1, y))

  function f2(s : string) =
    Chain.event(Event2(s, Call.caller))
```

Also adding a `Int.to_str` to be able to add int/word to event payload.

### DIFF
--- a/apps/aesophia/src/aeso_ast_infer_types.erl
+++ b/apps/aesophia/src/aeso_ast_infer_types.erl
@@ -70,6 +70,7 @@ global_env() ->
     Bool    = {id, Ann, "bool"},
     String  = {id, Ann, "string"},
     Address = {id, Ann, "address"},
+    Event   = {id, Ann, "event"},
     State   = {id, Ann, "state"},
     Hash    = {id, Ann, "hash"},
     Oracle  = fun(Q, R) -> {app_t, Ann, {id, Ann, "oracle"}, [Q, R]} end,
@@ -113,6 +114,7 @@ global_env() ->
      {["Chain",    "block_height"], Int},
      {["Chain",    "difficulty"],   Int},
      {["Chain",    "gas_limit"],    Int},
+     {["Chain",    "event"],        Fun1(Event, Unit)},
      %% State
      {"state", State},
      {"put",   Fun1(State, Unit)},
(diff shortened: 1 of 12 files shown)


apps/aesophia/src/aeso_ast_infer_types.erl
```diff
@@ -283,6 +289,19 @@ check_typedef_sccs(Env, TypeMap, [{cyclic, Names} | SCCs]) ->
     type_error({recursive_types_not_implemented, lists:map(Id, Names)}),
     check_typedef_sccs(Env, TypeMap, SCCs).
 
+check_event(Cons) ->
+    [ check_event(Name, Types) || {constr_t, _, {con, _, Name}, Types} <- Cons ].
+
+%% Initially we limit the type of an event, it can have 0-3 topics/indexed "words"
+%% and 0-1 strings as payload.
+check_event(Name, Types) ->
+    IsIndexed  = fun(T) -> proplists:get_value(indexed, aeso_syntax:get_ann(T), false) end,
```
UlfNorell: `aeso_syntax:get_ann(indexed, T, false)`

apps/aesophia/src/aeso_builtins.erl
```diff
@@ -0,0 +1,481 @@
...
+str_to_icode(String) when is_list(String) ->
+    str_to_icode(list_to_binary(String));
+str_to_icode(BinStr) ->
+    Cpts = [size(BinStr) | aeso_data:binary_to_words(BinStr)],
+    #tuple{ cpts = [ #integer{value = X} || X <- Cpts ] }.
+
+%% Event primitive (dependent on Event type)
+%%
+%% We need to switch on the event and prepare the correct #event for icode_to_asm
+%% NOTE: we assume all errors are already checked!
```
UlfNorell: Are they though? Somewhere we should check that indexed arguments are words and that the payload is a string. This is a nice place to do it since we can translate Sophia types to VM types here.

hanssv: Does Robert agree this is a nice place? 😉 

Any pointers on how to translate Sophia types to VM types would be helpful?! How to report an error from here?

PR #1970, opened 2018-12-21
Add release notes for new AEVM version

gorillainduction:
https://www.pivotaltracker.com/story/show/162792344

### DIFF
--- a/docs/release-notes/next-minerva/PT-162792344-new-aevm-version.md
+++ b/docs/release-notes/next-minerva/PT-162792344-new-aevm-version.md
@@ -1,0 +1,2 @@
+* Introduces a new AEVM version to contain consensus breaking changes and optimizations.
+* Removes references to old planned VM versions that will not be implemented.


docs/release-notes/next-minerva/PT-162792344-new-aevm-version.md
```diff
@@ -0,0 +1,2 @@
+* Introduces a new AEVM version to contain consensus breaking changes and optimizations.
```
lucafavatella: I expect this follows https://github.com/aeternity/epoch/pull/1964#pullrequestreview-187131056
Thanks.

docs/release-notes/next-minerva/PT-162792344-new-aevm-version.md
```diff
@@ -0,0 +1,2 @@
...
+* Removes references to old planned VM versions that will not be implemented.
```
lucafavatella: This seems not visible to the user so unnecessary.

gorillainduction: Since this is visible in the Protocol (where I will make a PR as well), I thought I'd mention it here for completeness.

PR #1978, opened 2018-12-21
Fix channels FSM IF

velzevur:
PT [Test throughput.many_chs_msg_loop failed failed on CI - timeout receive_from_fsm](https://www.pivotaltracker.com/story/show/162808119)

When mining for transactions inclusion, we aim at mining keyblocks in hope that all expected transactions will be included in microblocks before the timeout of keyblocks runs out. Since same genesis microblocks have a minimal time interval between them 3 seconds, if the expected key block interval is set to be lower than 3s, there will be at most 1 micro block per generation. If there are a lot of transactions in the mempool, they might not be included (so we are epxected to bump the maximum key block height to be mined). This is quite non-deterministic. This PR aims at improving the throughput in test suites.

### DIFF
--- a/apps/aechannel/test/aesc_fsm_SUITE.erl
+++ b/apps/aechannel/test/aesc_fsm_SUITE.erl
@@ -764,7 +764,7 @@ multiple_channels_t(NumCs, FromPort, Msg, Slogan, Cfg) ->
             end,
             Txs),
     aecore_suite_utils:mine_blocks_until_txs_on_chain(
-        aecore_suite_utils:node_name(dev1), TxHashes, 20),
+        aecore_suite_utils:node_name(dev1), TxHashes, 10),
     mine_blocks(dev1, ?MINIMUM_DEPTH),
     Cs = collect_acks(Cs, channel_ack, NumCs),
     ct:log("channel pids collected: ~p", [Cs]),
--- a/apps/aecore/test/aecore_suite_utils.erl
+++ b/apps/aecore/test/aecore_suite_utils.erl
@@ -235,6 +235,10 @@ mine_blocks_until_txs_on_chain(Node, TxHashes, MaxBlocks) ->
     mine_blocks_until_txs_on_chain(Node, TxHashes, 100, MaxBlocks).
 
 mine_blocks_until_txs_on_chain(Node, TxHashes, MiningRate, Max) ->
+    OldRate
+        = rpc:call(
+           Node, application, get_env, [aecore, expected_mine_rate, MiningRate],
+           5000),
     ok = rpc:call(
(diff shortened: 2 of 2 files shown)


apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -1,4 +1,4 @@
-%%%=============================================================================
+%%%=====================================================================~/projects/aeternity/epoch/apps/aechannel/test/aesc_fsm_SUITE.erl
```
lucafavatella: ?

lucafavatella:
```suggestion
%%%=============================================================================
```

apps/aecore/test/aecore_suite_utils.erl
```diff
@@ -257,6 +264,10 @@ mine_blocks_until_txs_on_chain_loop(_Node, _TxHashes, 0, _Acc) ->
 mine_blocks_until_txs_on_chain_loop(Node, TxHashes, Max, Acc) ->
     case mine_blocks_loop(1, key) of
         {ok, [Block]} -> %% We are only observing key blocks
+          % set a properly huge key block mining interval
+          ok = rpc:call(
+              Node, application, set_env, [aecore, expected_mine_rate,
+                                           1000 * 60 * 3], 5000),
```
lucafavatella: 180s vs. default 300s? Do you mean huge compared to rate used normally in tests?

apps/aecore/test/aecore_suite_utils.erl
```diff
@@ -257,6 +264,10 @@ mine_blocks_until_txs_on_chain_loop(_Node, _TxHashes, 0, _Acc) ->
 mine_blocks_until_txs_on_chain_loop(Node, TxHashes, Max, Acc) ->
     case mine_blocks_loop(1, key) of
         {ok, [Block]} -> %% We are only observing key blocks
+          % set a properly huge key block mining interval
+          ok = rpc:call(
+              Node, application, set_env, [aecore, expected_mine_rate,
```
lucafavatella: Is this mine_blocks_until_txs_on_chain_loop changing expected_mine_rate during same test? Would this potentially lead to inconsistent DB?

apps/aecore/test/aecore_suite_utils.erl
velzevur: Although not required to be the same as real life, default is [180 s]*(https://github.com/aeternity/epoch/blob/master/apps/aecore/src/aec_governance.erl#L41-L42)

velzevur:
This will impact block validation, so it is not suitable approach

PR #1979, opened 2018-12-21
Fix FSM IF

velzevur:
PT [Test throughput.many_chs_msg_loop failed failed on CI - timeout receive_from_fsm](https://www.pivotaltracker.com/story/show/162808119)
When mining keyblocks fast, we only got 1 microblock per generation (default microblock intervals is 3s, keyblock is set to be 0.1s); if mempool has a little bit too much txs those might or might not be mined in time (for when the last keyblock is mined). This introduces some non-determinism.

This PR setups the node being tested to have microblock interval 1ms.


lucafavatella:
I see this PR is only changing aesc_fsm_SUITE.erl so only integration tests job is affected.

PR #1986, opened 2018-12-29
Fix error in aesc_force_progress_tx:for_client/1

velzevur:
PT [SC: fix broken encoding of payload in `aesc_force_progress_tx:for_client:for_client/1`](https://www.pivotaltracker.com/story/show/162881066)


velzevur:
Not sure how to address this in release notes. cc @lucafavatella

PR #1995, opened 2019-01-08
Pt 162577473 stand alone sophia

ThomasArts:
Remove aesophia from epoch and point to its own repository.


ThomasArts:
Waitign for a little update in aesophia before we can merge

PR #1999, opened 2019-01-08
Handle missing os_pid gracefully

tolbrino:
Also fixed some formatting.

### DIFF
--- a/apps/aecore/src/aec_pow_cuckoo.erl
+++ b/apps/aecore/src/aec_pow_cuckoo.erl
@@ -34,15 +34,18 @@
 
 -define(debug(F, A), epoch_pow_cuckoo:debug(F, A)).
 -define(info(F, A),  epoch_pow_cuckoo:info(F, A)).
+-define(warning(F, A), epoch_pow_cuckoo:warning(F, A)).
 -define(error(F, A), epoch_pow_cuckoo:error(F, A)).
 
--record(state, {os_pid :: integer() | undefined,
+-type os_pid() :: integer() | undefined.
+-type pow_cuckoo_solution() :: [integer()].
+
+-record(state, {os_pid :: os_pid(),
                 port :: port() | undefined,
                 buffer = [] :: string(),
                 target :: aec_pow:sci_int() | undefined,
                 parser :: output_parser_fun()}).
 
--type pow_cuckoo_solution() :: [integer()].
 -type output_parser_fun() :: fun((list(string()), #state{}) ->
                                         {'ok', term(), term()} | {'error', term()}).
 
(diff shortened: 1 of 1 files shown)


apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -620,15 +621,22 @@ exec_run(Cmd, Dir, Args) ->
...
         Port = erlang:open_port(PortName, PortSettings),
-        {os_pid, OsPid} = erlang:port_info(Port, os_pid),
-        ?debug("External mining process started with OS pid ~p", [OsPid]),
-        {ok, Port, OsPid}
+        case erlang:port_info(Port, os_pid) of
+            {os_pid, OsPid} ->
+                ?debug("External mining process started with OS pid ~p", [OsPid]),
+                {ok, Port, OsPid};
+            undefined ->
+                ?error("External mining process finished before ~p could acquire the OS pid", [?MODULE]),
```
lucafavatella: It feels like this shall be a warning rather than an error?
If this case happens, I expect no OS process leaked and the output of the OS process (e.g. if exited because extremely quickly to start and find solution) parsed in the main loop `wait_for_result`.

Also, for testing this error would still trigger test failure for [this check on logs](https://github.com/aeternity/epoch/blob/a9e8f8ae8f7dc3e6dcf2da30342684ae37218661/system_test/helpers/aest_nodes.erl#L94).

apps/aecore/src/aec_pow_cuckoo.erl
```diff
@@ -620,15 +623,22 @@ exec_run(Cmd, Dir, Args) ->
...
         Port = erlang:open_port(PortName, PortSettings),
-        {os_pid, OsPid} = erlang:port_info(Port, os_pid),
-        ?debug("External mining process started with OS pid ~p", [OsPid]),
-        {ok, Port, OsPid}
+        case erlang:port_info(Port, os_pid) of
+            {os_pid, OsPid} ->
+                ?debug("External mining process started with OS pid ~p", [OsPid]),
+                {ok, Port, OsPid};
+            undefined ->
+                ?warning("External mining process finished before ~p could acquire the OS pid", [?MODULE]),
```
hanssv: I suspect this will mostly happen with the test-miner (16-bit graphs)?!

tolbrino: Yes. It actually only happened in system testing. Any real miner would take longer to start.

apps/aecore/src/aec_pow_cuckoo.erl
tolbrino: Changed to warning.

PR #2002, opened 2019-01-09
Syntax for assembler.

happi:
Suggestion for assembler syntax.

At the face to face meeting we talked about having basic blocks as just list of list of instructions.
In order to make the assembler code more readable I added labels (integers only for now).
A jump can only go to a label.

Comments are welcome.

There is no real code yet.
In the base branch (another PR) I have an implementation with tests that can execute the code example.

### DIFF
--- a/apps/aefate/src/aefa_code.erl
+++ b/apps/aefate/src/aefa_code.erl
@@ -1,0 +1,164 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2019, aeternity Anstalt
+%%% @doc
+%%%     Handling FATE code.
+%%% @end
+%%% Created : 9 Jan 2019
+%%%-------------------------------------------------------------------
+-module(aefa_code).
+-export([parse/1
+        ]).
+
+
+
+%%% Assembler syntax:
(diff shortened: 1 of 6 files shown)


rvirding:
So `add_a_a_a` pops the top two elements off the stack adds them to together and pushes the result onto the stack? What is the logic behind calling the instructions `add_a_a_a` and not just `add` as they operate on the stack?

Could probably write a yecc parser for it. 😄

happi:
The accumulator is the same as the top of the stack in this machine.
It is not really a stack machine and there will be an unlimited number of registers (or variables) I have not implemented that part yet but there would be _r_ (or _v_ for variable or _n_ for name) suffixes for referring to these.
So we will have:
_a accumulator/stack
_i immediate
_r register/variable

So regarding @rvirding question, perhaps it should be _a_a_s for acc := acc OP (pop stack one level).
The idea is that the stack will not be used that much since we have variables, but it is practical for some tree like syntax constructs like: foo(X) + foo(Y) - bar(Z) which would be
call_r bar Z
call_r foo Y
call_r foo X
add_a_a_a
sub_a_a_a

Thoughts?

PR #2011, opened 2019-01-10
Use correct Word size - 32

hanssv:
The check in heap_to_heap was done on a word size = 1, this gives a
pessimistic bound and will cause problems. Change where not consensus
breaking.

### DIFF
--- a/apps/aevm/src/aevm_eeevm_state.erl
+++ b/apps/aevm/src/aevm_eeevm_state.erl
@@ -281,12 +281,17 @@ spend_gas(Gas, State) ->
         GasLeft when GasLeft < 0 -> aevm_eeevm:eval_error(out_of_gas)
     end.
 
+%% Of course WordSize should be 32 - but we have to be buggy until the
+%% first scheduled hard-fork.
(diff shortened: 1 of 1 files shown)


apps/aevm/src/aevm_eeevm_state.erl
```diff
@@ -281,12 +281,17 @@ spend_gas(Gas, State) ->
         GasLeft when GasLeft < 0 -> aevm_eeevm:eval_error(out_of_gas)
     end.
 
+%% Of course WordSize should be 32 - but we have to be buggy until the
+%% first scheduled hard-fork.
 heap_to_heap(Type, Ptr, State) ->
+    heap_to_heap(Type, Ptr, State, 1).
```
UlfNorell: Maybe nicer to turn the default around, so that omitted word size means the correct behaviour and put the comment about the hard-fork at the buggy call instead.

hanssv: Yes, true... 

Did it this way since the buggy call is in another module and didn't want to change more than necessary... We still need to make changes here regardless.

PR #2016, opened 2019-01-11
Remove obsolete Mnesia patches

tolbrino:


rebar.config
```diff
@@ -114,9 +114,7 @@
                    {copy, "apps/aeutils/priv/extensions/keys_gen.sh", "bin/extensions/keys_gen"},
                    {copy, "apps/aeutils/priv/extensions/get_peer_key.sh", "bin/extensions/peer_key"},
                    {copy, "apps/aeutils/priv/extensions/export_chain.sh", "bin/extensions/export_chain"},
-                   {copy, "apps/aeutils/priv/extensions/messages_hash.sh", "bin/extensions/messages_hash"},
-                   {copy, "otp_patches/ebin/mnesia.beam"       , "patches/ebin/mnesia.beam"       },
```
lucafavatella: Could you please delete che comment at top of this file?

lucafavatella: I see `config/vm.args:4:-pa patches/ebin` i.e. resulting folder included in code path (similarly symling in integration test). Just checking - is that left on purpose in case any other patches needed? I would be all for deleting that `-pa`.

lucafavatella:
In `config/eunit.vm.args` I read `-pa otp_patches/ebin patches/ebin` - I expect a review of those (at least `otp_patches/ebin` seems a leftover, maybe both).

tolbrino:
@lucafavatella I wanted to keep the otp patches functionality, because it might become useful in the future.

tolbrino:
@uwiger Can you approve this too?

PR #2021, opened 2019-01-15
OTP release rename with backward compatible bin/epoch - preparation

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/163191726

Without non-mainnet network id, mining the first block would take too long for a smoke test. [Proof](https://circleci.com/gh/aeternity/epoch/54369).

### DIFF
--- a/py/tests/release.py
+++ b/py/tests/release.py
@@ -50,6 +50,7 @@ websocket:
         acceptors: 100
 
 mining:
+    autostart: true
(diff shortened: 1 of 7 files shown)


py/tests/release.py
```diff
@@ -60,6 +61,9 @@
 chain:
     persist: true
     db_path: ./my_db
+
+fork_management:
+    network_id: ae_uat
```
dincho: Why not using some other id? I guess we'll add the peers for this ID at some point and the test will take longer because it will connect to the real testnet.

PR #2022, opened 2019-01-15
OTP release rename with backward compatible bin/epoch

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/163191726

Depends on https://github.com/aeternity/epoch/pull/2021
Depends on https://github.com/aeternity/epoch/pull/2024
Depends on https://github.com/aeternity/protocol/pull/291

TODO:
* [x] Ensure https://github.com/aeternity/epoch/pull/2021 merged
* [x] Edit base branch of this PR to master
* [x] Ensure CI passes
* [x] Ensure OSX package job passes - see [auxiliary CI job](https://circleci.com/gh/aeternity/epoch/54411)
* [x] Check how multiple arguments are passed
* [x] Ensure https://github.com/aeternity/epoch/pull/2024 merged
* [ ] Check Windows

PR #2024, opened 2019-01-16
Factor integration tests init code

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/163191726

This is a preparatory PR for https://github.com/aeternity/epoch/pull/2022
While aiming at adding test suite I was not sure how to do so, and the clarification resulted in this refactoring.


apps/aecore/test/aecore_suite_utils.erl
```diff
@@ -102,20 +101,32 @@ patron() ->
 %%% API
 %%%=============================================================================
 
+init_per_suite(NodesList, CTConfig) ->
+    init_per_suite(NodesList, #{}, CTConfig).
+
+init_per_suite(NodesList, CustomNodeConfig, CTConfig) ->
+    init_per_suite(NodesList, CustomNodeConfig, [], CTConfig).
+
+init_per_suite(NodesList, CustomNodeCfg, NodeCfgOpts, CTConfig) ->
```
lucafavatella: I re-assigned the PR to myself while I re-consider whether `init_per_suite` is the appropriate name for this function.

lucafavatella: Seems good enough.

PR #2027, opened 2019-01-17
PT-162811053 split vm- and abi-version

hanssv:
Lots of changed lines, but this does not change the functionality (much!). Slight changes to the HTTP API - but the serialized data format is *unchanged*.

*NOTE:* We don't yet have a _MINERVA_ branch of `aesophia` so some `aeso_...` function calls are yet to be updated - they should get (and in the long run use) the VM and ABI version.


gorillainduction:
> How does one know if a specific combination of ct and abi versions is valid?

In the code, it is controlled by `aect_contracts:is_legal_vm_version_at_height`. It should probably be documented in protocol.

hanssv:
I agree with the API changes... I'll make a follow up PR!

PR #2035, opened 2019-01-18
remove catch-all base_gas clause

uwiger:
See [PT #163250430](https://www.pivotaltracker.com/story/show/163250430)

Main motivation: Make dialyzer able to detect whether a clause is missing after adding a tx type, rather than relying on testing to find out.

### DIFF
--- a/apps/aecore/src/aec_governance.erl
+++ b/apps/aecore/src/aec_governance.erl
@@ -127,10 +127,7 @@ tx_base_gas(name_update_tx) -> ?TX_BASE_GAS;
 tx_base_gas(oracle_extend_tx) -> ?TX_BASE_GAS;
 tx_base_gas(oracle_query_tx) -> ?TX_BASE_GAS;
 tx_base_gas(oracle_register_tx) -> ?TX_BASE_GAS;
-tx_base_gas(oracle_response_tx) -> ?TX_BASE_GAS;
-tx_base_gas(_) ->
-    %% never accept unknown transaction types
-    block_gas_limit().
+tx_base_gas(oracle_response_tx) -> ?TX_BASE_GAS.
 
 byte_gas() ->
     ?BYTE_GAS.


velzevur:
I see test job failed with
```

%%% aecore_sync_SUITE ==> all_nodes.large_msgs.large_msgs: FAILED
%%% aecore_sync_SUITE ==> {retry_exhausted,
    [{aecore_suite_utils,collect_sync_events,2,
         [{file,
              "/home/builder/epoch/_build/test/lib/aecore/test/aecore_suite_utils.erl"},
          {line,476}]},
     {aecore_sync_SUITE,expect_same,2,
         [{file,
              "/home/builder/epoch/_build/test/lib/aecore/test/aecore_sync_SUITE.erl"},
          {line,509}]},
     {aecore_sync_SUITE,large_msgs,1,
         [{file,
              "/home/builder/epoch/_build/test/lib/aecore/test/aecore_sync_SUITE.erl"},
          {line,592}]},
     {test_server,run_test_case_eval1,6,
         [{file,"test_server.erl"},{line,1062}]},
     {test_server,run_test_case_eval,9,
         [{file,"test_server.erl"},{line,994}]}]}
```
which is an unrelated to the changes introduced by this PR. Failure is already tracked as [Intermittent test failure: aecore_sync_SUITE retry_exhausted](https://www.pivotaltracker.com/story/show/163116525)

velzevur:
@ThomasArts great analysis! Here are my thoughts on the matter.

I believe crashing hard is much better than silently putting an enourmous amount of gas to unhandled cases. Even if dialyzer does not rise a warning, a simple test would catch the former and will fail at catching the latter. I don't think we have tests that check specific gas required by transaction types, nor we test it for prim ops.

> This therefore does break compatibility. If we have a transaction on chain that happened to be have failed, then it will have taken all gas...

Indeed it does in theory but only if there is already a unknown transaction included in the main chain. There is a simple test to check it: start a normal vanilla node (not necessarily with this branch applied) and let it sync with main net. A new transaction type will require some new handling that is not present on a vanilla node so the node will fail to sync and will discard any blocks containing such transaction. This will happen not only with our nodes but with all nodes running non-modified code. Basically a new transaction type does require a hard fork and if something like this had already happened - we would have noticed it by now.

Regarding the unknown transaction type being an attack vector - we stop uknown types at deserialization time and we don't let them propagate any further. If we did, it would be a bug.

>Another observation. IF we add a tx_type, then this should be valid from a certain height. This must be done with care.

I agree with you on 100% on this one, as it will require a hard fork. One must always be careful with those.

ThomasAerts:
I agree Dimitar.  The thing that I had liked to see for this pull request is:
1) Argumentation why this itself is not a hard fork and therefore fits in the Roma release
```
Regarding the unknown transaction type being an attack vector - we stop uknown types at deserialization time and we don't let them propagate any further. If we did, it would be a bug.
```
If indeed we don't let unknown transactions through, then this is not a breaking change. But we need to be surer of that. Moreover, the Elixir folks could have done so, thus there are no guarantees.
2) Checking whether this already has happened, i.e. loading the chain and look at transactions we don't recognise is in general not enough for hard-fork detection. Note that we may conic use today that we're save, prepare a releasee, gets it out and at the moment some run this release and some do not, we get such a transaction on chain. Then we have a hard fork.
3) Anyone could write its own implementation and start posting transactions to our chain containing weird things (such as wrong transaction type). We have to be absolutely sure that not only we reject them as incoming transaction, but also that if a malicious miner accepts them and have them in a block, we reject that block.

uwiger:
I should have noted that it can be seen as an incompatibility, but the PR _is_ against Minerva, not Roma...

PR #2040, opened 2019-01-21
Pt 163083413 sc signatures check

velzevur:
PTs:
* [SC FSM tests: reusable ports](https://www.pivotaltracker.com/story/show/159456827)
* [Add SC off-chain signatures validation](https://www.pivotaltracker.com/story/show/163083413)

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -17,9 +17,12 @@
          close_solo/1,
          shutdown/1,              %% (fsm())
          client_died/1,           %% (fsm())
+         connection_died/1,       %% (fsm())
          inband_msg/3,
          get_state/1,
          get_poi/2]).
+
+-export([strict_sig_checks/2]). %% tests only
 
 %% Inspection and configuration functions
 -export([ get_history/1     %% (fsm()) -> [Event]
@@ -141,6 +144,7 @@
               , ongoing_update = false :: boolean()
               , last_reported_update   :: undefined | non_neg_integer()
               , log = #w{}             :: #w{}
+              , check_signatures = true:: boolean()
               }).
 
 -define(TRANSITION_STATE(S),  S=:=awaiting_signature
@@ -358,6 +362,14 @@ get_state(Fsm) ->
 get_poi(Fsm, Filter) ->
     gen_statem:call(Fsm, {get_poi, Filter}).
 
+%% This is supposed to be used in tests only. If Alice stops checking
+%% her own signatures, Bob will still check hers and will reject any wrongly
+%% signed txs. There is a risk for Alice corrupting her own state if using
+%% this function
+-spec strict_sig_checks(pid(), boolean()) -> ok.
+strict_sig_checks(Fsm, Strict) when is_boolean(Strict) ->
+    gen_statem:call(Fsm, {strict_sig_checks, Strict}).
+
 leave(Fsm) ->
     lager:debug("leave(~p)", [Fsm]),
     gen_statem:call(Fsm, leave).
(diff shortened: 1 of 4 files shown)


velzevur:
Added a `WIP` as it must be based on @uwiger 's renaming changes

velzevur:
It must be based on #2048 and accomodate the solo closing sequence

velzevur:
Squashed and rebased on latest `master`. The solo closing sequence is not in place yet, so no accomodation regarding it.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1264,6 +1340,8 @@ handle_call_(St, close_solo, From, D) ->
             D1 = request_signing(close_solo_tx, CloseSoloTx, D),
             next_state(awaiting_signature, D1, [{reply, From, ok}])
     end;
+handle_call_(_, {strict_sig_checks, Strict}, From, #data{} = D) ->
```
uwiger: It might be good to check the type of `Strict` even here.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1306,6 +1384,8 @@ handle_common_event_(cast, {?CHANNEL_CLOSING, ChanId} = Msg, _St, _,
             lager:debug("Channel no longer active: ~p", [Error]),
             close(channel_no_longer_active, D)
     end;
+handle_common_event_({call, From}, {strict_sig_checks, Strict}, _, _, #data{} = D) ->
```
uwiger: Perhaps a comment about why a specific clause is needed here, since the next clause seems to handle all calls (except during `Mode == error_all`)?

Also, check that `Strict` is a boolean.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1345,7 +1425,7 @@ error_binary(E) when is_atom(E) ->
     atom_to_binary(E, latin1).
 
 
-terminate(Reason, _State, Data) ->
+terminate(Reason, _State, #data{session = _Session} = Data) ->
```
uwiger: Why this change? `_Session` doesn't seem to be referred to.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1915,7 +2012,14 @@ check_deposit_signed_msg(#{ channel_id := ChanId
                           , data       := TxBin} = Msg,
                           #data{on_chain_id = ChanId} = Data) ->
     SignedTx = aetx_sign:deserialize_from_binary(TxBin),
-    {ok, SignedTx, log(rcv, ?DEP_SIGNED, Msg, Data)}.
+    case check_type_and_verify_signatures(SignedTx, channel_deposit_tx,
+                                          [other_account(Data),
+                                           my_account(Data)],
```
uwiger: Not terribly important, but one could imagine a `both_accounts(Data)` helper function. ;-)

apps/aechannel/src/aesc_fsm.erl
velzevur: Refactoring side effect

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -17,10 +17,13 @@
          close_solo/1,
          shutdown/1,              %% (fsm())
          client_died/1,           %% (fsm())
+         connection_died/1,       %% (fsm())
          inband_msg/3,
          get_state/1,
          get_poi/2]).
 
+-export([strict_sig_checks/2]). %% tests only
```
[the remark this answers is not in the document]
velzevur: I'd actually like to play with this in a non-test environment :)

PR #2041, opened 2019-01-21
Update installation.md

mintymac:
Version 1.3.0 does not seem to be available.   Currently the url returns a 404.   Propose link to latest stable release **Roma maintenance release** 1.2.0

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -3,7 +3,7 @@
 Run below command to install latest version of aeternity node.
 
 ```bash
-curl -O https://raw.githubusercontent.com/aeternity/epoch/v1.3.0/scripts/install.sh && chmod +x install.sh && ./install.sh 1.3.0
+curl -O https://raw.githubusercontent.com/aeternity/epoch/v1.2.0/scripts/install.sh && chmod +x install.sh && ./install.sh 1.2.0
 ```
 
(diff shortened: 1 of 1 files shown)


dincho:
The `master` branch represents the next release, which is as for now planned to be 1.3.0, thus the changes you are mentioning.

You should look at latest stable release branch at: https://github.com/aeternity/aeternity/blob/v1.2.0/docs/installation.md

However, we understand that concrete issue that users are browsing the master documentation and there are planned improvements in that direction.


Related tickets:
- https://www.pivotaltracker.com/story/show/162785706
- https://www.pivotaltracker.com/story/show/162813285
- https://www.pivotaltracker.com/story/show/163222137

lucafavatella:
> maybe it's better to keep the latest release versions until a new release (postpone the version bumps) ?

@dincho As the master branch represents the next release, the doc may be not applicable to the already released software for other reasons (e.g. additional configuration parameter). Plus we may forget to update all the references to the version when actually releasing. Unless we automate the checks that there are no leftover versions in files, I suggest we keep master with the version of the next release.

mintymac:
@lucafavatella That makes sense somewhat, but from a community point of view, you have a 404 as the first experience for a developer trying to understand the project.

PR #2048, opened 2019-01-22
Pt 159406302 solo closing

uwiger:
See [PT #159406302](https://www.pivotaltracker.com/story/show/159406302)

W.I.P.
Initial push mainly refactors tx creation code in the fsm in order to make it easier to maintain and to add more txs (e.g. force_progress)

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -1751,12 +1751,18 @@ tx_defaults(channel_withdraw_tx = Tx, Opts, D) ->
     maps:merge(
       tx_defaults(channel_deposit_tx, Opts, D),
       #{ fee => default_fee(Tx) });  % same as deposit defaults
+tx_defaults(channel_slash_tx, _Opts, _D) ->
+    #{};
 tx_defaults(channel_close_solo_tx = Tx, _Opts, _D) ->
     #{ fee => default_fee(Tx) }.
 
-default_fee(Tx) ->
-    %% this will most likely be to small and have to be adjusted upwards
-    aec_governance:tx_base_gas(Tx).
+default_fee(_Tx) ->
+    %% TODO: For now, revert back to hard-coding the fee (sc_ws_api needs to be extended not to
+    %% ignore options like 'fee', and the API docs need to be amended accordingly)
+    20000.
+%% default_fee(Tx) ->
+%%     %% this will most likely be to small and have to be adjusted upwards
+%%     aec_governance:tx_base_gas(Tx).
 
 adjust_ttl(undefined) ->
     CurHeight = aec_headers:height(aec_chain:top_header()),
@@ -2502,7 +2508,7 @@ process_update_error(Reason, From, D) ->
 check_closing_event(#data{on_chain_id = ChanId} = D) ->
     case aec_chain:get_channel(ChanId) of
         {ok, Channel} ->
-            case aesc_channel:is_active(Channel) of
+            case aesc_channels:is_active(Channel) of
                 true ->
                     check_closing_event_(Channel, D);
                 false ->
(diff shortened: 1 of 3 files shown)


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1267,7 +1293,19 @@ handle_common_event_(cast, {?DISCONNECT, _}, _St, _, D) ->
 handle_common_event_(cast, {?CHANNEL_CLOSING, ChanId} = Msg, _St, _,
                      #data{on_chain_id = ChanId} = D) ->
     lager:debug("got ~p", [Msg]),
-    close(channel_closing_on_chain, D);
+    D1 = log(rcv, ?CHANNEL_CLOSING, Msg, D),
+    case check_closing_event(D1) of
```
uwiger: The `check_closing_event/1` function is actually buggy, and will (as far as I can tell) always return an error. This happens to result in the old behavior (fsm terminating if it detects a change on-chain). Fixing it in this PR would open a can of worms, since it would require further changes in the watcher and fsm.

One could argue that `check_closing_event/1` should be short-circuited for now rather than accidentally failing in a fortuitous way.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -181,12 +183,12 @@ callback_mode() -> [state_functions, state_enter].
...
 %% |   |        update_* | |  | | |              |                            |   |
 %% |   |                 +-+  +-+ | shutdown     v                            |   |
-%% |   |                          |    +-------------+ [disconnect]           |   |
-%% |   |                          +--> |   closing   | ------------+          |   |
-%% |   |                               +-------------+             |          |   |
-%% |   |                                |  ^                       |          |   |
-%% |   |                                |  | channel_reestablish   |          |   |
-%% |   | <------------------------------+  |                       v          |   |
+%% |   |                          |    +------------------+   [disconnect]    |   |
+%% |   |                          +--> | mutual_closing   | ------+           |   |
```
lucafavatella: Update required in https://github.com/aeternity/protocol/tree/362d566af481214a5f650385b10d6cb3cfd0e831/channels#overview ?

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -14,6 +14,7 @@
          upd_call_contract/2,     %%
          get_contract_call/4,     %% (fsm(), contract_id(), caller(), round())
          leave/1,
+         close_solo/1,
```
lucafavatella: I understand that this is not exposed to the WebSocket endpoint yet, hence:
* It has no doc update in https://github.com/aeternity/protocol/blob/362d566af481214a5f650385b10d6cb3cfd0e831/node/api/channel_ws_api_json-rpc.md yet
* It has no line in release notes

apps/aechannel/src/aesc_fsm.erl
uwiger: Yes, captured in [PT #163554702](https://www.pivotaltracker.com/story/show/163554702).

apps/aechannel/src/aesc_fsm.erl
uwiger: This is correct

apps/aechannel/test/aesc_txs_SUITE.erl
```diff
@@ -4817,6 +4817,9 @@ test_both_old_round(Cfg, Fun) ->
     test_both_old_round(Cfg, Fun, #{}).
```
lucafavatella: Unused clause - may be deleted.

apps/aechannel/test/aesc_txs_SUITE.erl
```diff
@@ -4817,6 +4817,9 @@ test_both_old_round(Cfg, Fun) ->
...
 
 test_both_old_round(Cfg, Fun, Props) ->
```
lucafavatella: Unused clause - may be deleted.

apps/aechannel/src/aesc_utils.erl
```diff
@@ -105,9 +105,12 @@ check_round_greater_than_last(Channel, Round, Type) ->
                 ChannelRound
         end,
     ?TEST_LOG("MinRound ~p, Round ~p", [MinRound, Round]),
-    case MinRound < Round of
-        true  -> ok;
-        false -> {error, old_round}
+    if MinRound < Round ->
+            ok;
+       MinRound == Round ->
+            {error, same_round};
```
lucafavatella: I do not see this under consensus so ok.

PR #2052, opened 2019-01-23
PT-163360420 epoch bin script

tolbrino:
Also add filtering of rocksdb dependencies on win32 platforms.

### DIFF
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -1,12 +1,9 @@
 %% -*- erlang -*-
 
-IsWindows = fun() ->
-                Arch = erlang:system_info(system_architecture),
-                string:find(Arch, "win32") =/= nomatch
-            end,
+IsWindows = case os:type() of {win32, _} -> true; {_, _} -> false end,
 
 FilterRelxRocksDb = fun(Apps) ->
-                        case IsWindows() of
+                        case IsWindows of
                             false ->
                                 Apps;
                             true ->
@@ -15,7 +12,7 @@ FilterRelxRocksDb = fun(Apps) ->
                     end,
 
 FilterDepsRocksDb = fun(Deps) ->
-                        case IsWindows() of
+                        case IsWindows of
                             false ->
                                 Deps;
                             true ->
@@ -24,7 +21,7 @@ FilterDepsRocksDb = fun(Deps) ->
                     end,
 
 AdaptOverlayBin = fun(Overlay) ->
-                      case IsWindows() of
+                      case IsWindows of
                           false ->
                               Overlay;
                           true ->
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -1,8 +1,49 @@
+%% -*- erlang -*-
+
+IsWindows = fun() ->
+                Arch = erlang:system_info(system_architecture),
+                string:find(Arch, "win32") =/= nomatch
+            end,
+
+FilterRelxRocksDb = fun(Apps) ->
+                        case IsWindows() of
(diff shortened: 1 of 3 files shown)


rebar.config.script
```diff
@@ -1,8 +1,49 @@
+%% -*- erlang -*-
+
+IsWindows = fun() ->
+                Arch = erlang:system_info(system_architecture),
+                string:find(Arch, "win32") =/= nomatch
```
lucafavatella:
```
case os:type() of {win32, _} -> true; {_, _} -> false end
```
?

Ref http://erlang.org/documentation/doc-9.1/lib/kernel-5.4/doc/html/os.html#type-0

scripts/epoch.bat
```diff
@@ -0,0 +1,5 @@
+SETLOCAL ENABLEEXTENSIONS
+
+SET REL_NAME=aeternity
+
+call %~dp0/%REL_NAME% %*
```
lucafavatella: Could you please confirm that [this test](https://github.com/aeternity/aeternity/pull/2022#issuecomment-454965501) passes.

rebar.config.script
tolbrino: Ok, changed.

scripts/epoch.bat
tolbrino: I will report back until tomorrow on the result.

rebar.config.script
```diff
@@ -1,8 +1,46 @@
...
+                        end
+                    end,
+
+AdaptOverlayBin = fun(Overlay) ->
+                      case IsWindows of
+                          false ->
+                              Overlay;
+                          true ->
+                              NewEntry = {copy, "scripts/epoch.bat", "bin/epoch.bat"},
+                              lists:keyreplace("bin/epoch", 3, Overlay, NewEntry)
```
lucafavatella: I see avoids shipping `bin/epoch` on win. Ok.

scripts/epoch.bat
lucafavatella: One discussion at https://github.com/aeternity/aeternity/pull/2062 is settled, can you please add deprecation warning matching that one. Cc @dincho

tolbrino: I added the warning.

scripts/epoch.bat
```diff
@@ -2,4 +2,6 @@ SETLOCAL ENABLEEXTENSIONS
 
 SET REL_NAME=aeternity
 
-call %~dp0/%REL_NAME% %*
+@echo WARNING: This script named "%~nx0" is deprecated, please use the script named "%REL_NAME%.bat" instead.
```
lucafavatella: Not sure if stdout or stderr. stderr preferred for operator.

rebar.config.script
```diff
@@ -1,8 +1,46 @@
...
+                                Apps -- [rocksdb, mnesia_rocksdb]
+                        end
+                    end,
+
+FilterDepsRocksDb = fun(Deps) ->
+                        case IsWindows of
+                            false ->
+                                Deps;
+                            true ->
+                                [D || {K, _} = D <- Deps, K =/= rocksdb, K =/= mnesia_rocksdb]
```
lucafavatella: Could this be - in the future - be replaced by explicit matching on deps `{_, _}` rather than list comprehension with filter (if any D in Deps did not match `{K, _}` that would be silently skipped).
This is Win only so no changes in existing archs anyway.

PR #2054, opened 2019-01-23
Pt 163388694 Add Bit shift operations to AEVM

hanssv:


apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -523,12 +523,12 @@ sign_and_apply_transaction(Tx, PrivKey, S1, Height) ->
     Trees    = aect_test_utils:trees(S1),
     Env0     = aetx_env:tx_env(Height),
     Env      = aetx_env:set_beneficiary(Env0, ?BENEFICIARY_PUBKEY),
-    {ok, AcceptedTxs, Trees1} =
-        aec_block_micro_candidate:apply_block_txs([SignedTx], Trees, Env),
-    S2       = aect_test_utils:set_trees(Trees1, S1),
-    case AcceptedTxs of
-        [SignedTx] -> {ok, S2};
-        []         -> {error, S2}
+    case aec_block_micro_candidate:apply_block_txs_strict([SignedTx], Trees, Env) of
```
gorillainduction: Thank you :1st_place_medal: !

apps/aevm/test/VMTests/test_generator.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
gorillainduction: Do we need the generator?

hanssv: Probably not, but we might want to repeat the stunt (for some other tests) at some point?

PR #2055, opened 2019-01-23
Upload macos packages to s3 during release

dincho:
https://www.pivotaltracker.com/story/show/162785706

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -651,6 +651,7 @@ jobs:
           at: *packages_workspace
       - *setup_secrets
       - *upload_github_release_asset
+      - *upload_release_artifacts
 
   upload_packages_linux:
     <<: *infrastructure_config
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -643,16 +643,7 @@ jobs:
...
-    <<: *infrastructure_config
-    steps:
-      - checkout
-      - attach_workspace:
-          at: *packages_workspace
-      - *setup_secrets
-      - *upload_github_release_asset
-
-  upload_packages_linux:
+  upload_release_packages:
```
lucafavatella: I see step `upload_github_release_asset` has `ASSETS=${PACKAGES_DIR:?}/*` so that is ok.

lucafavatella: How does `upload_release_artifacts` work with one `aws s3 cp --acl public-read ${PACKAGE_TARBALL:?} s3://${S3_RELEASES_BUCKET:?}/aeternity-latest-${PKG_SUFFIX:?}.tar.gz`? It feels like it uploads only one package to S3? Is that on purpose?

dincho: Good catch!

dincho: fixed

PR #2058, opened 2019-01-23
Fix IF in sync SUITE

velzevur:
PT [Intermittent test failure: aecore_sync_SUITE retry_exhausted](https://www.pivotaltracker.com/story/show/163116525)

I was not able to reproduce the issue locally. My assumption is that there was a race in the test:
1. `dev1`  starts and has its max gas limit raised. This breaks backwards compatibility with an unmodified node
2. `dev1` mines some transactions, presumably exceeds the original maximum gas limit so there are microblocks that are to be rejected by a non-modified node
3. We start `dev2` and it
3.1 Is updated with the correct value for maximum gas
3.2 Starts syncing with `dev1`

My assumption is that there is a race if steps. `3.1` and `3.2` progress in a reversed order.

This PR spawns both `dev1` and `dev2` without any peers, `dev1` proceeds with the same approach as it did, then we spawn `dev2`, set its max gas limit and only then add `dev1`'s peer address as a trusted peer so they can sync.

In order to test the hypothesis, I will restart the CI a couple of times. Adding `WIP` till then.


velzevur:
I have restarted the tests a couple of times and the issue did not appear

PR #2057, opened 2019-01-23
Fix backward incompatibility in user config discovery and DB

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/163422334

PR #2060, opened 2019-01-23
Change node binary path in configurations pt 163189592

valerifilipov:
https://www.pivotaltracker.com/story/show/163189592

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,8 +5,6 @@ services:
     build: .
     image: aeternity/aeternity:${IMAGE_TAG-latest}
     hostname: node1
-    environment:
-      EPOCH_CONFIG: /home/aeternity/aeternity.yaml
     command: >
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -5,8 +5,6 @@ services:
     build: .
     image: aeternity/aeternity:${IMAGE_TAG-latest}
     hostname: node1
-    environment:
-      EPOCH_CONFIG: /home/aeternity/aeternity.yaml
```
lucafavatella: Please rebase on recent master.
I am concerned that now-merged paraller PR #2057 may affect result in master.

lucafavatella: How is user config discovery meant to work without this var?
I see in line 14 that config is mapped to `/home/aeternity/aeternity.yaml` though not considered in https://github.com/aeternity/aeternity/wiki/User-provided-configuration

PR #2066, opened 2019-01-24
PT-162541397 Improve cleanup of UPnP/NAT-PMP mappings and assert ther…

zp-sd:
…e is maximum one process maintainting the mapping.

* Assert processes adding UPnP/NAT-PMP mappings do not accumulate - there as at maximum one such process.

* Move cleanup of UPnP/NAT-PMP mapping to `prep-stop` phase. This goes together with https://github.com/aeternity/erlang-nat/commit/e8d8eeb9909f83daae7215934b58c6bf6efd34dc, which does not start `inets` application in `erlang-nat` lib. This was not working well during application shutdown phase (when the mapping is to be removed, but `erlang-nat` was trying to do `_ = application:start(inets)` anyways).

### DIFF
--- a/apps/aecore/src/aec_upnp.erl
+++ b/apps/aecore/src/aec_upnp.erl
@@ -54,7 +54,7 @@ prep_stop() ->
             case catch gen_server:call(?MODULE, prep_stop, 10000) of
                 ok -> ok;
                 Error ->
-                    epoch_sync:info("UPnP/NAT-PMP port cleanup failed: ~p", [Error])
+                    epoch_sync:warning("UPnP/NAT-PMP port cleanup failed: ~p", [Error])
             end;
         false ->
             ok
@@ -72,7 +72,6 @@ start_link() ->
 
 init(_Args) ->
     epoch_sync:info("Starting UPnP/NAT-PMP service"),
-    process_flag(trap_exit, true),
     erlang:send_after(rand:uniform(1000), self(), add_port_mapping),
     {ok, #state{}}.
 
@@ -95,9 +94,6 @@ handle_info(add_port_mapping, State0) ->
     %% Give additional 10 secs for UPnP/NAT-PMP discovery and setup, to
     %% make sure there is continuity in port mapping.
     erlang:send_after(1000 * (?MAPPING_LIFETIME - 10), self(), add_port_mapping),
-    {noreply, State};
-handle_info({'EXIT', _Pid, normal}, State) ->
-    %% Received when calling delete_port_mapping/0 during prep_stop
     {noreply, State};
 handle_info({'DOWN', Ref, process, Pid, normal}, #state{pid = Pid, mon = Ref}) ->
     {noreply, #state{}};
--- a/apps/aecore/src/aec_upnp.erl
+++ b/apps/aecore/src/aec_upnp.erl
@@ -13,7 +13,8 @@
 -behaviour(gen_server).
 
 %% API
--export([is_enabled/0]).
+-export([is_enabled/0,
+         prep_stop/0]).
 
 %% gen_server API
 -export([start_link/0]).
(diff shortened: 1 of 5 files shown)


apps/aecore/src/aec_upnp.erl
```diff
@@ -31,17 +32,34 @@
...
                              [user_config, schema_default]) of
         {ok, IsEnabled} -> IsEnabled;
         undefined       -> false
     end.
 
+-spec prep_stop() -> ok.
+prep_stop() ->
+    case is_enabled() of
+        true ->
+            case catch gen_server:call(?MODULE, prep_stop, 10000) of
```
lucafavatella: It seems you rely on `aecore_app:prep_stop` rather than in `terminate` of `aec_upnp` in order to increase the 5000 ms the `aecore_sup` gives this child? (It is tricky as this worker process is now acting partially as a supervisor, and the timeout for a supervisor is recommended to be infinity in order not to kill the supervisor before it had killed the unliked children.)
Ok.

apps/aecore/src/aec_upnp.erl
```diff
@@ -87,6 +118,17 @@ code_change(_OldVsn, State, _Extra) ->
 %%% Internal functions
 %%%===================================================================
 
+kill_mapping_worker(#state{pid = undefined, mon = undefined}) ->
+    ok;
+kill_mapping_worker(#state{pid = Pid, mon = Ref}) ->
+    demonitor(Ref, [flush]),
+    true = exit(Pid, shutdown),
```
lucafavatella: I do not see the spawned process as trapping exits (I did not check the internals of the nat library though) so I understand this is intended to kill the process. (`true =` always succeeds, even if the `Pid` is not there any longer). Ok.

apps/aecore/src/aec_upnp.erl
```diff
@@ -66,18 +88,27 @@ handle_cast(Other, State) ->
...
+handle_info(add_port_mapping, State0) ->
+    ok = kill_mapping_worker(State0),
+    State = spawn_mapping_worker(),
+
     %% Give additional 10 secs for UPnP/NAT-PMP discovery and setup, to
     %% make sure there is continuity in port mapping.
     erlang:send_after(1000 * (?MAPPING_LIFETIME - 10), self(), add_port_mapping),
     {noreply, State};
+handle_info({'EXIT', _Pid, normal}, State) ->
+    %% Received when calling delete_port_mapping/0 during prep_stop
```
lucafavatella: So there are some spawn with link inside the `nat:*` calls in `delete_port_mapping`. (Did not check.) Ok.

apps/aecore/src/aec_upnp.erl
```diff
@@ -31,17 +32,34 @@
...
     end.
 
+-spec prep_stop() -> ok.
+prep_stop() ->
+    case is_enabled() of
+        true ->
+            case catch gen_server:call(?MODULE, prep_stop, 10000) of
+                ok -> ok;
+                Error ->
+                    epoch_sync:info("UPnP/NAT-PMP port cleanup failed: ~p", [Error])
```
lucafavatella: This seems more a warning.

apps/aecore/src/aec_upnp.erl
```diff
@@ -66,18 +88,27 @@ handle_cast(Other, State) ->
...
+    {noreply, #state{}};
+handle_info({'DOWN', Ref, process, Pid, Why}, #state{pid = Pid, mon = Ref}) when Why =/= normal ->
+    epoch_sync:warning("UPnP/NAT-PMP worker finished: ~p", [Why]),
+    {noreply, #state{}};
 handle_info(Other, State) ->
     epoch_sync:warning("Received unknown info message: ~p", [Other]),
     {noreply, State}.
 
 terminate(_Reason, _State) ->
-    delete_port_mapping(),
```
lucafavatella: So why keeping the trapping of exits - with consequent `handle_info({'EXIT', _Pid, normal}, State)` clause?

apps/aecore/src/aec_upnp.erl
zp-sd: Prep_stop is called before application shutting down process begin. Then the entire supervision tree of the application, including included applications, is terminated in reversed start order, so we should be safe to keep the mapping cleanup in `terminate/2`. But it seems to be safer/clearer anyway to clean up mapping before shutting down procedure begins (cleanup procedure rely on other applications, e.g. `inets`, to be started).

Infinity is not used in case some nat discovery/mapping removal hangs in the lib, but should never happen (I used some artificial sleeps in nat lib to provoke very unlikely scenarios).

apps/aecore/src/aec_upnp.erl
zp-sd: True, I will update.

apps/aecore/src/aec_upnp.erl
zp-sd: Good point, this is not needed anymore. I will update.

zp-sd:
@lucafavatella Review remarks applied. Please revisit.

PR #2071, opened 2019-01-24
Pt 162578724 refactor tx primops

gorillainduction:
https://www.pivotaltracker.com/story/show/162578724

- [ ] Run QuickCheck tests to ensure no consensus breaking changes.
- [ ] Build a release and verify that the Roma net chain can be verified.

### DIFF
--- a/apps/aecore/src/aec_spend_tx.erl
+++ b/apps/aecore/src/aec_spend_tx.erl
@@ -137,7 +137,7 @@ check(#spend_tx{}, Trees,_Env) ->
 -spec signers(tx(), aec_trees:trees()) -> {ok, [aec_keys:pubkey()]}.
 signers(#spend_tx{} = Tx, _) -> {ok, [sender_pubkey(Tx)]}.
 
--spec process(tx(), aec_trees:trees(), aetx_env:env()) -> {ok, aec_trees:trees()}.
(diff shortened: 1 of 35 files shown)


sennui:
I really like that we removed process step. It is handled now with exceptions. The only "but" I have is about everything landing in gigantic tx processor file. Hopefully tests will be chopped, but I always lean towards smaller files.

apps/aecore/src/aec_spend_tx.erl
```diff
@@ -125,79 +125,27 @@ recipient_id(#spend_tx{recipient_id = RecipientId}) ->
...
-            case resolve_recipient_pubkey(SpendTx, Trees) of
-                {ok, RecipientPubkey} ->
-                    {ok, aec_trees:ensure_account(RecipientPubkey, Trees)};
-                {error, _} = E ->
-                    E
-            end;
-        {error, _Reason} = Error ->
-            Error
-    end.
+check(#spend_tx{}, Trees,_Env) ->
```
sennui: I see you leave check check functions - where they are used?

gorillainduction: They are used from the `aetx.erl` still. There are a few channel transactions that are not migrated, but I left them out since they are not yet used from contracts.

PR #2073, opened 2019-01-24
Remove misleading 1.3.0 prefix from Win CI

lucafavatella:
Addresses https://github.com/aeternity/aeternity/pull/2072#issuecomment-457257855


lucafavatella:
> I'd suggest adding a commit sha if supported

[Tracked](https://www.pivotaltracker.com/story/show/163476835).

PR #2089, opened 2019-01-29
Additional CI check re types

lucafavatella:
I noticed recent improvements in `minerva` branch following warnings emitted by type checking tool in more recent OTP. This PR proposes to include such check as new job part of CI - though not blocking any other job in case it fails.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ references:
 
   container2_config: &container2_config
     docker:
-      - image: aeternity/builder:ci-build-otp-21
+      - image: aeternity/builder:otp21
         user: builder
     working_directory: ~/aeternity
     environment:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -480,6 +480,7 @@ jobs:
           name: Check OTP version (for effective CI caching)
           # How to retrieve OTP version in installed OTP development system: http://erlang.org/doc/system_principles/versions.html
           command: test "${OTP_VERSION:?}" = $(cat $(erl -noinput -eval 'io:format("~s", [filename:join([code:root_dir(), "releases", erlang:system_info(otp_release), "OTP_VERSION"])]), init:stop().'))
+      - *fail_notification
 
   rebar_lock_check:
     <<: *container_config
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,15 @@ references:
       OTP_VERSION: 20.3.8
   container_otp_plt: &container_otp_plt _build/default/rebar3_20.3.8_plt
 
+  container2_config: &container2_config
+    docker:
+      - image: aeternity/builder:ci-build-otp-21
+        user: builder
+    working_directory: ~/aeternity
+    environment:
+      OTP_VERSION: 21.2.3
+  container2_otp_plt: &container2_otp_plt _build/default/rebar3_21.2.3_plt
+
   infrastructure_config_environment: &infrastructure_config_environment
     resource_class: large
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -7,6 +7,15 @@ references:
         user: builder
     working_directory: ~/aeternity
 
+  container2_config: &container2_config
```
lucafavatella: This approach follows https://github.com/aeternity/aeternity/pull/2088 so this may need update once 2088 commented/rejected/merged.

lucafavatella:
> I'm personally against the idea to have allowed failing tests, that compromises the trustworthiness of the CI system in general.
> 
> I'd prefer this OTP 21 support to be fixed at once for all and go forward.

Fixed in https://github.com/aeternity/aeternity/pull/2119 (master).

lucafavatella:
I cherry-picked a commit from #2119 for aiming at getting green - this would need to be rebased and cannot be merged as is.

lucafavatella:
On recent minerva branch i.e. 51c553da:
```
===> Analyzing 172 files with "/home/builder/aeternity/_build/default/rebar3_21.2.3_plt"...

apps/aecore/src/aec_block_genesis.erl
  56: Function genesis_header/0 has no local return
  97: Function genesis_block_with_state/0 has no local return
 100: Function genesis_block_with_state/1 has no local return

apps/aecore/src/aec_block_key_candidate.erl
  36: Function adjust_target/2 will never be called
  76: Function int_create/5 has no local return
  77: The pattern {'ok', Trees} can never match the type 'error'
  84: Function int_create_block/5 will never be called

apps/aecore/src/aec_blocks.erl
 160: Function new_key/10 has no local return
 162: The call aec_headers:new_key_header(Height::any(), PrevHash::any(), PrevKeyHash::any(), RootHash::any(), Miner::any(), Beneficiary::any(), Target::any(), 'no_value', Nonce::any(), Time::any(), Version::any()) breaks the contract (height(), block_header_hash(), block_header_hash(), state_hash(), miner_pubkey(), beneficiary_pubkey(), aeminer_pow:sci_target(), aeminer_pow_cuckoo:solution(), non_neg_integer(), non_neg_integer(), non_neg_integer()) -> header()

apps/aecore/src/aec_chain_state.erl
 309: Function fake_key_node/4 has no local return

apps/aecore/src/aec_conductor.erl
1081: The pattern {'ok', Block} can never match the type {'error','block_not_found' | 'headers_for_target_adjustment_not_found' | 'key_not_found'}

apps/aecore/src/aecore_app.erl
  23: The pattern 'false' can never match the type 'true'
```

lucafavatella:
Force pushed with recent minerva branch and inmerged type fix in https://github.com/aeternity/aeternity/pull/2185

lucafavatella:
> I'm personally against the idea to have allowed failing tests, that compromises the trustworthiness of the CI system in general.
> 
> I'd prefer this OTP 21 support to be fixed at once for all and go forward.

It is fixed in master now (up to yesterday). I rebased this PR for proving this still stands.

> Also `aeternity/builder:ci-build-otp-21` is a temporary build artifact, I'd prefer additional image without that prefix, otherwise it can be deleted during cleanup.

I am open to suggestions here. I see this image was not deleted in a while so this approach seems good enough to me.

> I'm also not sure why this target minerva, is it supported to have BC breaks with OTP 21 support?

This targets master now.

> However, the configuration looks working, thus the approve.

@dincho Could you please re-consider?

lucafavatella:
Depends on https://github.com/aeternity/docker-builder/pull/14

PR #2100, opened 2019-01-31
PT-163553535 Testing old vm version(s) (and compiler(s))

hanssv:
This is a first step towards better testing. It will run `aecontract_SUITE` with different combinations of VM and compiler. For most of the functionality the VM and the compiler behaves the same for ROMA and MINERVA so most test cases are fairly boring/unnecessary to re-run though.

The macro `?assertMatchVM` is used where the test is behaving differently between VM:s (look for these, they mark _interesting_ tests). Similarily `skipRest` is used when a test can't be compiled with the old compiler...

To achieve this there is a bash script that fetches a standalone Sophia compiler for ROMA and uses it when appropriate. @tolbrino promised to check the script for Windows feasibility...

Thoughts on the approach, and also whether we should remove some tests from the duplicated running?!

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -170,7 +170,7 @@
                             {ref, "a4fb3db"}}},
                             {ephemeral, "2.0.4"}
                            ]},
-                    {pre_hooks, [{ct, "scripts/standalone_compiler.sh"}]}
+                    {pre_hooks, [{ct, "sh ./scripts/standalone_compiler.sh"}]}
                    ]},
             {prod, [{relx, [{dev_mode, false},
                             {include_erts, true},
--- a/rebar.config
+++ b/rebar.config
@@ -167,8 +167,10 @@
                     {deps, [{meck, "0.8.12"},
                             {websocket_client, ".*", {git,
                             "git://github.com/aeternity/websocket_client",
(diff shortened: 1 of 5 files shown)


hanssv:
CI had a couple of valid complaints... Fixing those.

hanssv:
Discussing with @UlfNorell - cross testing the compiler does not really have a value functionality-wise... Removing that to avoid exponential growth...

rebar.config
```diff
@@ -167,8 +167,13 @@
...
                             {websocket_client, ".*", {git,
                             "git://github.com/aeternity/websocket_client",
-                            {ref, "a4fb3db"}}}
-                           ]}
+                            {ref, "a4fb3db"}}},
+                            {ephemeral, "2.0.4"}
+                           ]},
+                    {pre_hooks, [{"(linux|darwin|solaris|freebsd|netbsd|openbsd)",
+                                  ct, "scripts/standalone_compiler.sh"},
+                                 {"win32", ct, "sh ./scripts/standalone_compiler.sh"}
```
tolbrino: If you want to have less trouble in the future for this kind of hook, you could move the shell script execution into a Makefile target, and just call `make mytarget` here. That should work on all platforms.

PR #2106, opened 2019-02-01
Fix script for running node on Win

lucafavatella:
I did not test this though I do not see how this script can work otherwise.

PR #2107, opened 2019-02-01
Amend scripts re software renaming

lucafavatella:
See also #2106

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -11,7 +11,7 @@ if [[ -z "${RELEASE_VERSION}" ]]; then
     echo -e "Usage:\n"
     echo -e "  $0 release_version\n"
     echo "Release version format is X.Y.Z where X, Y, and Z are non-negative integers"
-    echo "You can find a list of aeternity releases at https://github.com/aeternity/epoch/releases"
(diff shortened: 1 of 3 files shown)


lucafavatella:
> let's also use the new binary in https://github.com/aeternity/aeternity/pull/2107/files#diff-80423f13d45b53af35790ffc6009be4dR120 ?

@dincho  You mean in install.sh. Though how would the script work with OLD_RELEASE_VERSIONS then? Shall that handling be deleted?

tolbrino:
> > let's also use the new binary in https://github.com/aeternity/aeternity/pull/2107/files#diff-80423f13d45b53af35790ffc6009be4dR120 ?
> 
> @dincho You mean in install.sh. Though how would the script work with OLD_RELEASE_VERSIONS then? Shall that handling be deleted?

How about adding a note that this script only support v1.3.0 onwards. For older version one should use the script pre-v1.3.0?

lucafavatella:
> How about adding a note that this script only support v1.3.0 onwards. For older version one should use the script pre-v1.3.0?

I suggest to keept this out of scope of this PR.
Tracked as https://www.pivotaltracker.com/story/show/163649762

PR #2108, opened 2019-02-01
Amend comments re software name

lucafavatella:


py/tests/swagger-version-check.py
```diff
@@ -14,7 +14,7 @@
 yaml_version = yaml.safe_load(args.yaml_file)['info']['version']
 json_version = json.load(args.json_file)['info']['version']
 
-print("Epoch version: %s" % epoch_version)
```
dincho: maybe also change the `epoch_version` variable

PR #2111, opened 2019-02-04
Pt 163645385 sum token supply

gorillainduction:
https://www.pivotaltracker.com/story/show/163645385

### DIFF
--- a/apps/aecore/src/aec_chain.erl
+++ b/apps/aecore/src/aec_chain.erl
@@ -599,6 +599,18 @@ get_key_header_by_height(Height) when is_integer(Height), Height >= 0 ->
 %%% Statistics
 %%%===================================================================
 
+-spec sum_tokens_at_height(aec_blocks:height()) ->
+                                  #{ 'accounts'         => non_neg_integer()
+                                   , 'contracts'        => non_neg_integer()
+                                   , 'contract_oracles' => non_neg_integer()
+                                   , 'locked'           => non_neg_integer()
+                                   , 'oracles'          => non_neg_integer()
+                                   , 'oracle_queries'   => non_neg_integer()
+                                   , 'pending_rewards'  => non_neg_integer()
+                                   , 'total'            => non_neg_integer()
+                                   }.
+
+
 sum_tokens_at_height(Height) ->
     %% Wrap in transaction for speed.
     %% TODO: This could be done dirty
--- a/apps/aecore/src/aec_chain.erl
+++ b/apps/aecore/src/aec_chain.erl
@@ -30,6 +30,7 @@
         , hash_is_in_main_chain/1
         , hash_is_connected_to_genesis/1
         , prev_hash_from_hash/1
+        , sum_tokens_at_height/1
         , top_block/0
         , top_block_hash/0
         , top_key_block/0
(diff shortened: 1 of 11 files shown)


apps/aecore/src/aec_chain.erl
```diff
@@ -593,3 +594,60 @@ get_key_header_by_height(Height) when is_integer(Height), Height >= 0 ->
...
+int_sum_tokens_at_height(Height) ->
+    case aec_chain_state:get_key_block_hash_at_height(Height) of
+        error -> {error, chain_too_short};
+        {ok, Hash} ->
+            {ok, Trees} = get_block_state(Hash),
+            Sum = aec_trees:sum_total_coin(Trees),
+            Sum1 = Sum#{pending_rewards => sum_pending_rewards(Height, Hash)},
+            Total = maps:fold(fun(_, X, Acc) -> Acc + X end, 0, Sum1),
+            {ok, Sum1#{ total => Total
+                      , height => Height
```
lucafavatella: Is this height in return value intended?

gorillainduction: Good catch! It was intended, but not documented. I will either remove this or document it.

apps/aecore/src/aec_chain.erl
```diff
@@ -593,3 +594,60 @@ get_key_header_by_height(Height) when is_integer(Height), Height >= 0 ->
...
+                                        , 'oracle_queries'   => non_neg_integer()
+                                        , 'pending_rewards'  => non_neg_integer()
+                                        , 'total'            => non_neg_integer()
+                                        }}.
+
+
+sum_tokens_at_height(Height) ->
+    %% Wrap in transaction for speed.
+    %% TODO: This could be done dirty
+    aec_db:ensure_transaction(fun() -> int_sum_tokens_at_height(Height)end).
```
hanssv: Whitespace error

apps/aecore/src/aec_chain.erl
hanssv: If intended the typespec should be updated...

config/swagger.yaml
```diff
@@ -3488,6 +3512,33 @@ definitions:
           - call_data
           - gas
           - gas_price
+  TokenSupply:
```
hanssv: Height?

apps/aecore/src/aec_chain.erl
gorillainduction: I have now removed it

apps/aecore/src/aec_chain.erl
```diff
@@ -593,3 +594,58 @@ get_key_header_by_height(Height) when is_integer(Height), Height >= 0 ->
...
+    %% so we need stop at the closing key block (hence the +1).
+    StopHeight = max(Height - aec_governance:beneficiary_reward_delay() + 1, 1),
+    sum_pending_rewards(Hash, StopHeight, 0).
+
+sum_pending_rewards(Hash, StopHeight, Acc) ->
+    {ok, Header}  = get_header(Hash),
+    Height        = aec_headers:height(Header),
+    {value, Fees} = aec_db:find_block_fees(Hash),
+    Coinbase      = aec_coinbase:coinbase_at_height(Height),
+    Acc1 = Acc + Fees + Coinbase,
```
lucafavatella: This seems not to cater for fraud already-reported though pending for delay. Could be considered as optional enhancement. (I need to review this logic.)

apps/aecore/src/aec_trees.erl
```diff
@@ -249,6 +250,94 @@ contracts(Trees) ->
...
+
+sum_accounts('$end_of_table',_LockAccount,_Contracts,_Oracles, Acc) ->
+    Acc;
+sum_accounts({LockAccount, SerAccount, Iter}, LockAccount, Contracts, Oracles, Acc) ->
+    Account = aec_accounts:deserialize(LockAccount, SerAccount),
+    Balance = aec_accounts:balance(Account),
+    Acc1 = maps:update_with(locked, fun(X) -> X + Balance end, Acc),
+    Next = aeu_mtrees:iterator_next(Iter),
+    sum_accounts(Next, LockAccount, Contracts, Oracles, Acc1);
+sum_accounts({Pubkey, SerAccount, Iter}, LockAccount, Contracts, Oracles, Acc) ->
```
lucafavatella: I see double accounting for coins in `aec_governance:locked_coins_holder_account()` is prevented by previous clause. Ok.

apps/aecore/src/aec_trees.erl
```diff
@@ -249,6 +250,94 @@ contracts(Trees) ->
...
+    maps:update_with(oracle_queries, fun(X) -> X + Balance end, Acc).
+
+sum_implicit_oracle_query_fees(Pubkey, Oracles, Last, Acc) ->
+    case aeo_state_tree:get_oracle_queries(Pubkey, Last, open, 10, Oracles) of
+        [] -> Acc;
+        List ->
+            Acc1 = lists:foldl(fun(Query, X) ->
+                                       X + aeo_query:fee(Query)
+                               end, Acc, List),
+            Last1 = aeo_query:id(lists:last(List)),
```
lucafavatella: I see this does not cause query to be counted twice because of semantics of iterator that exclude initial key when calling next. Ok.

apps/aecore/src/aec_chain.erl
gorillainduction: Agreed, but since this should be a pretty rare case, I don' think we should spend effort in this.

fabiankrol:
🚀

apps/aecore/src/aec_chain.erl
tolbrino: In that case it would be good to mention that assumption in a comment.

gorillainduction: The pending rewards is always pending until the beneficiary reward delay has passed. Then the decision is made whether to pay out the reward, or to lock the reward and pay a smaller reward to the reporter of the fraud. Given this, the semantics of the summation made here is intact.

It would be, as Luca says, an improvement to specify this as a separate field, and there are other things that would be enhancing the information as well (e.g., differing between fees and coinbase, differing between deposits and available funds in contract). I just wanted to bite off at some level.

The fraud reporting is there to deter fraud, and we are mostly not expecting this to happen. That is why I don't want to spend more CPU cycles to find this information in this code.

tolbrino: Thanks for the explanation. I agree with your reasoning.

PR #2114, opened 2019-02-05
add deploy next_btranch to next and unstable envs

valerifilipov:


.circleci/config.yml
```diff
@@ -652,6 +652,18 @@ jobs:
       - DEPLOY_DOWNTIME: 1800 #30M
     steps: *deploy_steps
 
+  deploy_next:
+    <<: *infrastructure_config
+    environment:
+      - DEPLOY_ENV: next
```
lucafavatella: Not sure if any downtime needed.

.circleci/config.yml
```diff
@@ -652,6 +652,18 @@ jobs:
...
+    steps: *deploy_steps
+
+  deploy_unstable:
+    <<: *infrastructure_config
+    environment:
+      - DEPLOY_ENV: unstable
```
lucafavatella: Not sure if any downtime needed.

valerifilipov: They are just 3 nodes, so i'll give them 15 min

PR #2118, opened 2019-02-06
Failing mac os package uploads to s3 PT-163619978

valerifilipov:
Fix proposal for https://www.pivotaltracker.com/story/show/163619978

* add env var PKG_TARGET_OS for package upload
* extract infrastucture_config environment in variable so it can be appended instead of rewritten
A successful test run here  - https://circleci.com/gh/aeternity/aeternity/4442

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,14 +10,17 @@ references:
       OTP_VERSION: 20.3.8
   container_otp_plt: &container_otp_plt _build/default/rebar3_20.3.8_plt
 
+  infrastructure_config_enviroment: &infrastructure_config_environment
+      DOCKERHUB_REPO: aeternity/aeternity
+      S3_BUILDS_BUCKET: aeternity-node-builds
+      S3_RELEASES_BUCKET: aeternity-node-releases
+
   infrastructure_config: &infrastructure_config
     docker:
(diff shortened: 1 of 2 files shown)


.circleci/set_package_path.sh
```diff
@@ -2,12 +2,17 @@
 
 set -e
 
-if [ "$(uname -s)" == "Darwin" ]; then
-    PKG_SUFFIX="macos-"`uname -m`
-elif [ "$(uname -s)" == "Linux" ]; then
-    PKG_SUFFIX="ubuntu-"`uname -m`
+if [ "${PKG_TARGET_OS}x" == "x" ]; then
```
dincho: I don't like this hackish check, what's wrong with `-z` ?

.circleci/config.yml
```diff
@@ -655,25 +669,15 @@ jobs:
...
     <<: *infrastructure_config
-    steps:
-      - checkout
-      - *set_package_path
-      - attach_workspace:
-          at: *packages_workspace
-      - *setup_secrets
-      - *upload_github_release_asset
-      - *upload_release_artifacts
+    <<: *upload_package
```
dincho: let's add explicit `PKG_TARGET_OS` here as well

PR #2122, opened 2019-02-06
SC FSM: check channel IDs in updates

velzevur:
PT [SC: reject off-chain txs from another channel](https://www.pivotaltracker.com/story/show/163176382)
Also refactored fsm SUITE a bit to reuse abstracted test logic


velzevur:
Since #2129 is opened against `minerva`, I close this one

PR #2129, opened 2019-02-07
SC FSM: check channel IDs in updates

velzevur:
PT [SC: reject off-chain txs from another channel](https://www.pivotaltracker.com/story/show/163176382)
Those are the changes introduced in #2122 but opened against `minerva`


apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -962,6 +985,50 @@ wrong_sig_action(ChannelStuff, Poster, Malicious,
...
+                                                             aec_id:create(channel,
+                                                                          ?BOGUS_PUBKEY)),
+                              log(Debug1, "modified ~p", [Tx]),
+                              SignedTx = aec_test_utils:sign_tx(Tx, [Priv]),
+                              Action(Tag, Signer, SignedTx)
+                      after ?TIMEOUT ->
+                              error(timeout)
+                      end
+                  end,
+                  different_channel_id).
```
lucafavatella: I tried changing this, that is new error reason introduced by this PR, and test failed as expected. Ok.

```
%%% aesc_fsm_SUITE ==> all_tests.channel_ids.check_incorrect_deposit: FAILED
%%% aesc_fsm_SUITE ==> {timeout,[{aesc_fsm_SUITE,receive_from_fsm,6,
```

apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -962,6 +985,50 @@ wrong_sig_action(ChannelStuff, Poster, Malicious,
...
+                {FsmFun, FsmFunArg, FsmNewAction, FsmCreatedAction},
+                 DetectConflictFun) ->
+    wrong_action({I, R, _Spec, _Port, Debug}, Poster, Malicious,
+                 {FsmFun, FsmFunArg, FsmNewAction, FsmCreatedAction},
+                  DetectConflictFun,
+                  fun(Tag, #{priv := Priv} = Signer, Debug1) ->
+                      Action = fun sign_signing_request/3,
+                      receive {aesc_fsm, Fsm, #{type := sign, tag := Tag, info := Tx0} = Msg} ->
+                              log(Debug1, "await_signing(~p, ~p) <- ~p", [Tag, Fsm, Msg]),
+                              Tx = aesc_test_utils:update_tx(Tx0, channel_id,
```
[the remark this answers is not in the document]
velzevur: This is fixed in the subsequent PR.

velzevur:
Rebased on latest `minerva`

velzevur:
Rebased on latest `minerva`

velzevur:
Rebased on latest `master`

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2740,20 +2748,27 @@ verify_signatures(Pubkeys, SignedTx) ->
...
-            verify_signatures(Pubkeys, SignedTx);
+check_tx_and_verify_signatures(SignedTx, Type, ChannelPubkey, Pubkeys, ErrTypeMsg) ->
+    Aetx = aetx_sign:tx(SignedTx),
+    case aetx:specialize_type(Aetx) of
+        {Type, Tx} -> % check type
+            {Mod, Tx} = aetx:specialize_callback(Aetx),
+            case Mod:channel_pubkey(Tx) of
+                ChannelPubkey -> % expected pubkey
+                    verify_signatures(Pubkeys, SignedTx);
+                _ ->
```
lucafavatella: I see all `:channel_pubkey`s return pubkey without any tuple wrapper so correct. For readability, I would have preferred here a guard on the discarded value being a binary though it may be overkill. Ok.

I also see there is a test checking error `different_channel_id`, that makes reviewer more confident. Though I see many txs that expose `channel_pubkey` function, but it is not immediate to see how many of those are meant to be tested, and how many are tested (I checked #2132 too). May I suggest, in the list of test cases in `channel_ids`, to add comment mentioning which txs are knowingly not tested. (I am concerned that pull requests making the state machine aware of more transactions forget to integrate the tests in this test group.)

apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -880,6 +903,7 @@ wrong_sig_create({I, R, #{initiator_amount := IAmt0, responder_amount := RAmt0,
 
 check_incorrect_deposit(Cfg) ->
     Debug = true,
+    Fun = proplists:get_value(wrong_action, Cfg),
```
uwiger: In cases such as these, I personally prefer `{_, Fun} = lists:keyfind(wrong_action, 1, Cfg)`, although that also fails to verify that `Fun` is, in fact, a `fun/4`.

apps/aechannel/src/aesc_fsm.erl
lucafavatella: I see comment added in following commit.

PR #2131, opened 2019-02-08
Deploy from diffrent infrastructure versions 163419653

michalbagrowski:
pt: https://www.pivotaltracker.com/story/show/163419653

Basic review please, still waiting for operational dev1 to test it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,9 @@ executors:
   unstable:
     docker:
       - image: aeternity/infrastructure:deploy_from_diffrent_infrastructure_versions_163419653
+  stable:
+    docker:
+      - image: aeternity/infrastructure:1.7.0
 
 references:
   container_config: &container_config
@@ -478,6 +481,7 @@ jobs:
       - *fail_notification
 
   upload_build_artifacts:
+    executor: stable
     <<: *infrastructure_config
     steps:
       - checkout
@@ -598,6 +602,7 @@ jobs:
       - *fail_notification
 
   docker_push_tag:
+    executor: stable
     <<: *infrastructure_config
     steps:
       - checkout
@@ -612,6 +617,7 @@ jobs:
             docker push ${DOCKERHUB_REPO:?}:latest
 
   uat_chain_snapshot:
+    executor: stable
     <<: *infrastructure_config
     environment:
       - BACKUP_ENV: uat
@@ -622,6 +628,7 @@ jobs:
           path: *chain_snapshots
 
   main_chain_snapshot:
+    executor: stable
     <<: *infrastructure_config
     environment:
       - BACKUP_ENV: main
@@ -686,6 +693,7 @@ jobs:
           path: *system_test_host_logs
 
   deploy_integration:
+    executor: unstable
     <<: *infrastructure_config
     environment:
       - DEPLOY_ENV: integration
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,10 +1,18 @@
-version: 2
-
-references:
-  container_config: &container_config
+version: 2.1
+executors:
+  infrastructure_container_unstable:
+    docker:
+      - image: aeternity/infrastructure:master # should be changed to v1.8.0 after releasing
```
dincho: Not sure what this comment it for, we should keep it master here

michalbagrowski: right wrong line ;]

.circleci/config.yml
```diff
@@ -1,10 +1,18 @@
...
-references:
-  container_config: &container_config
+version: 2.1
+executors:
+  infrastructure_container_unstable:
+    docker:
+      - image: aeternity/infrastructure:master
+  infrastructure_container_stable:
+    docker:
+      - image: aeternity/infrastructure:stable
```
lucafavatella: Is a note anywhere re semantics of stable vs. master branches of infrastructure repo?

lucafavatella: Should https://github.com/aeternity/infrastructure/pull/266 not be present in stable branch also?

lucafavatella: Should https://github.com/aeternity/infrastructure/pull/271 not be present in stable branch also?

.circleci/config.yml
```diff
@@ -1,10 +1,18 @@
-version: 2
-
-references:
-  container_config: &container_config
+version: 2.1
+executors:
```
lucafavatella: Before, `aeternity/infrastructure` was used. I assume it meant tag `latest` though I did not find this documented in Circle CI doc. Now tags are `master` and `stable`. `latest` tag depends on [integration tests in that repo to have passed](https://github.com/aeternity/infrastructure/blob/8fcad03dd84b0ae6cee6615be099d83637330b1f/.circleci/config.yml#L266-L268), `master` and `stable` do not . So tags used now are less stable than before. Is this intended?

lucafavatella: In description of ticket I read:
> For example integration should be deployed with aeternity/infrastructure:latest while UAT and main with aeternity/infrastructure:v1.7.0 (latest working for it)
>
> This will allow implementing BC breaking changes to deploy scripts in aeternity/infrastructure:master

That `latest` vs. tag name makes sense. Not sure about `master` vs. `stable`.

.circleci/config.yml
dincho: no, it's about terraform config, it's not used in this repo. However, it will end-up in stable soon anyway

dincho: yes

dincho: Nope, but stable is meant to replace the tagging

.circleci/config.yml
dincho: `latest` is actually equal to `master`. However, it's true that `latest` is pushed after integration tests pass, while the `master` is pushed before that. This is actually downside of the workflow, but not bug. Otherwise we have to run integration tests on each branch which is expensive both time/cost.

So we definitely should use `latest` instead of `master` @michalbagrowski

The `stable` branch and docker tag are intended to be merged only from master with corresponding PR, thus its explicit that integration tests should have been passed for latest master.

lucafavatella: > The `stable` branch and docker tag are intended to be merged only from master with corresponding PR, thus its explicit that integration tests should have been passed for latest master.

Maybe worth placing a note in https://github.com/aeternity/infrastructure/wiki

dincho: tracking in https://www.pivotaltracker.com/story/show/164457613

PR #2133, opened 2019-02-08
Enable build artefacts upload on next

dincho:


.circleci/config.yml
```diff
@@ -836,7 +836,6 @@ workflows:
             branches:
               ignore:
                 - system-tests
-                - *next_branch
```
lucafavatella: I see upload as file name that is distinct from master. Ok.

PR #2134, opened 2019-02-11
Prefer miner with lower memory usage

lucafavatella:


lucafavatella:
Release package smoke test still failed. So discarding this.

PR #2135, opened 2019-02-11
Fix wrong UAT seed node peer addresses

dincho:
More context and prove in https://github.com/aeternity/infrastructure/pull/248


dincho:
Docs/Unrelated test changes only, thus merging with failing CI test jobs.

PR #2143, opened 2019-02-12
Re-factor aehttp to use standalone compiler as backend

hanssv:
Since we can't include multiple modules with the same name we need to
use a standalone compiler to provide ROMA compilation.

The compiler escript file goes into apps/aehttp/priv/bin/

https://www.pivotaltracker.com/story/show/163858683


config/swagger.yaml
```diff
@@ -966,7 +966,8 @@ paths:
         - contract
         - debug
       operationId: PostContractCreateCompute
-      description: 'Compute the call_data for SOPHIA and get a contract_create transaction object'
+      description: '(DEPRECATED to-be-removed in 3.0) - Compute the call_data for SOPHIA and get a contract_create transaction object'
```
lucafavatella: These deprecation may be worth a line in release notes.

config/swagger.yaml
```diff
@@ -966,7 +966,8 @@ paths:
...
+      deprecated: true
```
lucafavatella: I see [nicely represented in Swagger UI](http://aeternity.com/api-docs/?config=https://raw.githubusercontent.com/aeternity/aeternity/PT-163858683-deprecate_compiler_api/config/swagger.yaml#/internal/PostContractCreateCompute). 👍

hanssv:
Yes, exact design for future setup is yet to be seen... I'd like to wait a bit with updating docs etc in order not having to do it multiple times 🤔

rebar.lock
```diff
@@ -17,6 +17,7 @@
       {ref,"60a335668a60328a29f9731b67c4a0e9e3d50ab6"}},
   0},
  {<<"bear">>,{pkg,<<"bear">>,<<"0.8.7">>},2},
+ {<<"bucs">>,{pkg,<<"bucs">>,<<"1.0.16">>},1},
```
lucafavatella: I see bucrandom process starts though seems ok.

hanssv: If anyone knows of a more lightweight way than `ephemeral` to create temporary files it can be changed!

apps/aehttp/priv/README_aesophia.md
```diff
@@ -0,0 +1,6 @@
+Instructions for building bin/aesophia:
+
+git clone git@github.com:aeternity/aesophia.git
+git checkout a1d33e93abf19d0f99a11a2dcd21ddd1d46c127a
```
lucafavatella: I see this is the merge commit of https://github.com/aeternity/aesophia/pull/26

PR #2144, opened 2019-02-12
Pt 163788523 test roma protocol

gorillainduction:
https://www.pivotaltracker.com/story/show/163788523

Two new make targets are created:
`make ct-roma`
`make eunit-roma`

The new targets are not yet hooked into CI.


apps/aecore/test/aec_mining_tests.erl
```diff
@@ -27,14 +28,18 @@ mine_block_test_() ->
...
+                 TopBlock = aec_blocks:set_height(RawBlock, Height),
                  % if there is a change in the structure of the block
                  % this will result in a change in the hash of the header
                  % and will invalidate the nonce value below
                  % in order to find a proper nonce for your
                  % block uncomment the line below
                  % let_it_crash = generate_valid_test_data(TopBlock, 100000000000000),
-                 Nonce = 391854272740078490,
+                 Nonce = case aec_hard_forks:protocol_effective_at_height(Height + 1) of
+                             ?ROMA_PROTOCOL_VSN    -> io:format("ROMA\n", []), 1566115190779737391;
```
hanssv: `io:format` in the test? Intentional?

apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -303,20 +306,14 @@ create_contract_tx(Node, Name, Args, Fee, Nonce, TTL) ->
     ContractKey = aect_contracts:compute_contract_pubkey(OwnerKey, Nonce),
     {Res, aehttp_api_encoder:encode(tx_hash, aetx_sign:hash(CTx)), ContractKey, Code}.
 
-compile_contract(File) ->
```
hanssv: 👍

apps/aecore/test/aec_mining_tests.erl
gorillainduction: Nope

apps/aecore/src/aec_hard_forks.erl
```diff
@@ -1,79 +1,135 @@
...
+assert_protocols(M) ->
     GenesisVersion = aec_block_genesis:version(),
-    Vs = [GenesisVersion | _] = aec_governance:sorted_protocol_versions(),
+    Vs = [GenesisVersion | _] = sorted_protocol_versions(),
     %% Assert versions are sorted.
     Vs = lists:sort(Vs),
     %% Assert versions are distinct.
     [] = lists:sort(Vs) -- lists:usort(Vs),
-    Vs.
+    {[], _} = {maps:keys(M) -- Vs, check_no_extra_protocol_versions},
```
lucafavatella: Considering how assert_protocols is called (one place), this is a tautology.

apps/aecore/src/aec_hard_forks.erl
```diff
@@ -1,79 +1,135 @@
...
+    {[], _} = {Vs -- maps:keys(M), check_no_missing_protocol_versions},
```
lucafavatella: Considering how assert_protocols is called (one place), this is a tautology.

apps/aecore/src/aec_hard_forks.erl
lucafavatella: This seems to be the case from this change https://github.com/aeternity/aeternity/commit/84e1f517a753d10f62a3cbff9023925e099ac1c5#diff-312e5e881622101a1720cef694345ee9R71

apps/aecore/src/aec_hard_forks.erl
lucafavatella: This seems to be the case from this change https://github.com/aeternity/aeternity/commit/84e1f517a753d10f62a3cbff9023925e099ac1c5#diff-312e5e881622101a1720cef694345ee9R71

apps/aecore/src/aec_hard_forks.erl
```diff
@@ -1,79 +1,135 @@
...
+    assert_heights_strictly_increasing(M),
```
lucafavatella: What prevents user to specify protocol version unknown to the node? e.g. config with versions 1 2 3

PR #2146, opened 2019-02-12
Pt 162725461 compiler version serialized

ThomasArts:
Adding the compiler version to the serialization format

### DIFF
--- a/apps/aechannel/test/aesc_txs_SUITE.erl
+++ b/apps/aechannel/test/aesc_txs_SUITE.erl
@@ -194,7 +194,7 @@
 -include_lib("apps/aecore/include/blocks.hrl").
 -include_lib("stdlib/include/assert.hrl").
 -include_lib("apps/aecontract/src/aecontract.hrl").
--include_lib("aecore/include/hard_forks.hrl").
+-include_lib("aecontract/include/hard_forks.hrl").
 
 -define(MINER_PUBKEY, <<12345:?MINER_PUB_BYTES/unit:8>>).
 -define(BOGUS_CHANNEL, <<1:?MINER_PUB_BYTES/unit:8>>).
--- a/apps/aechannel/test/aesc_txs_SUITE.erl
+++ b/apps/aechannel/test/aesc_txs_SUITE.erl
@@ -7,9 +7,17 @@
 
 -module(aesc_txs_SUITE).
 
-%% common_test exports
--export([all/0,
-         groups/0]).
+-export([
+          all/0
+        , groups/0
+        , suite/0
+        , init_per_suite/1
+        , end_per_suite/1
+        , init_per_group/2
+        , end_per_group/2
+        , init_per_testcase/2
+        , end_per_testcase/2
+        ]).
 
 %% test case exports
 -export([create/1,
@@ -178,13 +186,15 @@
          fp_register_oracle/1,
          fp_oracle_query/1,
          fp_oracle_extend/1,
-         fp_oracle_respond/1
+         fp_oracle_respond/1,
+         fp_fork_awareness/1
         ]).
 
 -include_lib("common_test/include/ct.hrl").
 -include_lib("apps/aecore/include/blocks.hrl").
 -include_lib("stdlib/include/assert.hrl").
 -include_lib("apps/aecontract/src/aecontract.hrl").
+-include_lib("aecore/include/hard_forks.hrl").
 
 -define(MINER_PUBKEY, <<12345:?MINER_PUB_BYTES/unit:8>>).
 -define(BOGUS_CHANNEL, <<1:?MINER_PUB_BYTES/unit:8>>).
@@ -192,6 +202,7 @@
 -define(VM_VERSION, aect_test_utils:latest_sophia_vm_version()).
 -define(ABI_VERSION, aect_test_utils:latest_sophia_abi_version()).
 -define(TEST_LOG(Format, Data), ct:log(Format, Data)).
+-define(MINERVA_FORK_HEIGHT, 1000).
 %%%===================================================================
 %%% Common test framework
 %%%===================================================================
@@ -362,9 +373,42 @@ groups() ->
        fp_register_oracle,
        fp_oracle_query,
        fp_oracle_extend,
-       fp_oracle_respond
+       fp_oracle_respond,
+
+       fp_fork_awareness
       ]}
     ].
+
+suite() ->
+    [].
+
+init_per_suite(Config) ->
+    Config.
+
+end_per_suite(_Config) ->
+    ok.
+
+init_per_group(_Group, Config) ->
+    Config.
+
+end_per_group(_Group, _Config) ->
+    ok.
+
+init_per_testcase(fp_fork_awareness, Config) ->
+    meck:expect(aec_hard_forks, protocol_effective_at_height,
+                fun(V) when V < ?MINERVA_FORK_HEIGHT -> ?ROMA_PROTOCOL_VSN;
+                   (_)                               -> ?MINERVA_PROTOCOL_VSN end),
+    Config;
+init_per_testcase(_, Config) ->
+    Config.
+
+end_per_testcase(fp_fork_awareness, _Config) ->
+    meck:unload(aec_hard_forks),
+    ok;
+end_per_testcase(_Case, _Config) ->
+    ok.
+
+
 
 %%%===================================================================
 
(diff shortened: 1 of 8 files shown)


apps/aecontract/src/aect_sophia.erl
```diff
@@ -61,40 +73,69 @@ parse_options(<<_:8, Rest/binary>>, Acc) ->
     parse_options(Rest, Acc);
 parse_options(<<>>, Acc) -> Acc.
 
+is_legal_serialization_at_height(?SOPHIA_CONTRACT_VSN_1, Height) ->
+    true;
```
lucafavatella: So this means old serialization legal always. Ok.

test/sophia_bytecode_aevm_SUITE.erl
```diff
@@ -21,14 +21,20 @@ all() ->
       execute_identity_fun_from_sophia_file ].
 
 execute_identity_fun_from_sophia_file(_Cfg) ->
+    %% Should work for both versions of serialized contracts
+    execute_identity_fun_from_sophia_file(_Cfg, 1),
+    execute_identity_fun_from_sophia_file(_Cfg, 2).
```
lucafavatella: Is there a test checking that new format is invalid in old consensus? (For ensuring sanity of assumption in test.)

ThomasArts: Yes, I manually ran it against the contract_SUITE and basically all tests failed. I have considered adding a specific test for it, but thought it would polute the contract SUITE.

apps/aecore/src/aec_tx_processor.erl
```diff
@@ -857,7 +857,7 @@ contract_create({OwnerPubkey, Amount, Deposit, GasLimit, GasPrice,
     {Account, S1}  = get_account(OwnerPubkey, S),
     assert_account_balance(Account, TotalAmount),
     assert_contract_create_version(ABIVersion, VMVersion, S),
-    assert_contract_byte_code(ABIVersion, SerializedCode, CallData),
+    assert_contract_byte_code(ABIVersion, SerializedCode, CallData, S),
```
lucafavatella: So this is check in contract create tx. How does this interacts with state channels? Does this mean that contracts with new format created in state channels can be called in force progress tx, and also when tx on chain is under Roma protocol?

apps/aecontract/src/aect_sophia.erl
```diff
@@ -61,40 +73,69 @@ parse_options(<<_:8, Rest/binary>>, Acc) ->
...
+is_legal_serialization_at_height(?SOPHIA_CONTRACT_VSN_2, Height) ->
+    aec_hard_forks:protocol_effective_at_height(Height) >= ?MINERVA_PROTOCOL_VSN.
+
+serialize(Map) ->
+    serialize(Map, ?SOPHIA_CONTRACT_VSN).
```
lucafavatella: Do I understand correctly this is only called in debug endpoint for compilation CompileContract now deprecated?

Is caller of that API meant to be able to control serialization? In any case, I would expect swagger.yaml integrated with how CompileContract behaves in interaction with serialization.

test/sophia_bytecode_aevm_SUITE.erl
gorillainduction: I suggest you make a unit test in `aect_contract_create_tx_tests.erl` where you vary the code serialization in create transactions, and use
 `meck:expect(aec_hard_forks, protocol_effective_at_height, fun(_) -> ?MINERVA_PROTOCOL_VERSION end)`
and
`meck:expect(aec_hard_forks, protocol_effective_at_height, fun(_) -> ?ROMA_PROTOCOL_VERSION end)`
to see that the correct versions are accepted/rejected when calling
`aect_contract_create:process/3`

lucafavatella:
> There should also be a corresponding PR in protocol for the serialization format.

@gorillainduction https://github.com/aeternity/protocol/pull/310

ThomasArts:
> There should also be a corresponding PR in protocol for the serialization format.
https://github.com/aeternity/protocol/pull/310

apps/aecontract/src/aect_sophia.erl
ThomasArts: This has nothing to do with swagger. The serialize/2 function could later potentially be called from the outside if people know what they are doing. Yes, a user should be able to create a serialization of an older version, but Minerva will produce the new version by default once we're there.

test/sophia_bytecode_aevm_SUITE.erl
ThomasArts: Added a test in aecontract_SUITE now

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -450,6 +454,33 @@ create_contract_init_error_illegal_instruction_(OP, ErrReason) when is_binary(Er
     ?assertEqual(Gas, GasUsed),
     ok.
 
+create_version_too_high(Cfg) ->
+    S  = aect_test_utils:new_state(),
+    S0 = aect_test_utils:setup_miner_account(?MINER_PUBKEY, S),
+    {PubKey, S1} = aect_test_utils:setup_new_account(S0),
+    PrivKey      = aect_test_utils:priv_key(PubKey, S1),
+
+    {ok, IdContract} = compile_contract_vsn(identity, 2),
```
lucafavatella: I placed 1 here and I got:
```
%%% aecontract_SUITE ==> aevm_1.transactions.create_version_too_high: FAILED
%%% aecontract_SUITE ==> {{badmatch,
     {ok,#{key_pairs =>
...
 [{aecontract_SUITE,create_version_too_high,1,
      [{file,
           ".../_build/test/lib/aecontract/test/aecontract_SUITE.erl"},
       {line,479}]},
...
```

lucafavatella: That is ok.

apps/aecontract/src/aect_contracts.erl
```diff
@@ -141,7 +141,8 @@ is_legal_version_in_protocol(call, #{vm := VMVersion}, ProtocolVersion) ->
         ?ROMA_PROTOCOL_VSN    when VMVersion =:= ?VM_AEVM_SOPHIA_1 -> true;
         ?MINERVA_PROTOCOL_VSN when VMVersion =:= ?VM_AEVM_SOPHIA_1;
                                    VMVersion =:= ?VM_AEVM_SOPHIA_2 -> true;
-        _                     when VMVersion =:= ?VM_AEVM_SOLIDITY_1 -> ?VM_AEVM_SOLIDITY_1_enabled
+        _                     when VMVersion =:= ?VM_AEVM_SOLIDITY_1 -> ?VM_AEVM_SOLIDITY_1_enabled;
+        _ -> false
```
velzevur: Without this clause - when an unexpexted protocol & version are received - function crashes bad

apps/aechannel/test/aesc_txs_SUITE.erl
```diff
@@ -178,20 +186,23 @@
...
-         fp_oracle_respond/1
+         fp_oracle_respond/1,
+         fp_wrong_encoded_code/1
         ]).
 
 -include_lib("common_test/include/ct.hrl").
 -include_lib("apps/aecore/include/blocks.hrl").
 -include_lib("stdlib/include/assert.hrl").
 -include_lib("apps/aecontract/src/aecontract.hrl").
+-include_lib("aecore/include/hard_forks.hrl").
```
lucafavatella: Will need rebase on [recent minerva branch](https://github.com/aeternity/aeternity/commit/fdb574342a3b9ea73bae87b61b82c246427b9aa9#diff-40b66f66505fd87e31a4c12d0b88bea2).

apps/aechannel/test/aesc_txs_SUITE.erl
```diff
@@ -5230,3 +5280,46 @@ compile_contract(ContractName) ->
     aect_test_utils:compile_contract(
       filename:join(["contracts",
                      filename:basename(ContractName, ".aes") ++ ".aes"])).
+
+%% test that a force progress transaction can NOT produce an on-chain
+%% contract with a code with the wrong serialization
+fp_wrong_encoded_code(Cfg) ->
```
lucafavatella: Could you please confirm that running this test having reverted [`check_code_serialization` change in `check_force_progress_`](https://github.com/aeternity/aeternity/pull/2146/commits/5630cc259b21dc91f47d607050098ee66c6ce4e9) fails.

apps/aechannel/test/aesc_txs_SUITE.erl
```diff
@@ -362,10 +373,47 @@ groups() ->
...
+end_per_group(_Group, _Config) ->
+    ok.
+
+init_per_testcase(fp_wrong_encoded_code, Config) ->
+    meck:expect(aec_hard_forks, protocol_effective_at_height,
+                fun(V) when V < ?MINERVA_FORK_HEIGHT -> ?ROMA_PROTOCOL_VSN;
+                   (_)                               -> ?MINERVA_PROTOCOL_VSN end),
+    meck:new(aect_sophia, [passthrough]),
+    meck:expect(aect_sophia, serialize,
+                fun(Map) -> aect_sophia:serialize(Map, _Vsn = 2) end),
```
lucafavatella: I see this is where the compilation is set.

apps/aechannel/test/aesc_txs_SUITE.erl
ThomasArts: Confirmed. It fails if the code is not fixed.

PR #2148, opened 2019-02-12
Plant test suite re DB

lucafavatella:
As part of https://www.pivotaltracker.com/story/show/163423623 I am planting tests for handling DB migration following change of node name inside DB. This PR is a **RFC for where to place such tests**. It contains a basic test that is meant to pass without code changes, so it can be considered an increase in test coverage for the codebase and Dockerfile.

[Here](https://circleci.com/workflow-run/34c835fa-1296-4550-b242-5d334010474b) is CI workflow with the new test suite (temporarily) enabled in the system smoke test job.

### DIFF
--- a/system_test/aest_db_SUITE.erl
+++ b/system_test/aest_db_SUITE.erl
@@ -1,0 +1,74 @@
+-module(aest_db_SUITE).
+
+%=== EXPORTS ===================================================================
+
+% Common Test exports
+-export([all/0]).
+-export([init_per_suite/1]).
+-export([init_per_testcase/2]).
+-export([end_per_testcase/2]).
+-export([end_per_suite/1]).
+
+% Test cases
+-export([
+    node_can_reuse_db_of_other_node/1
+]).
+
+%=== INCLUDES ==================================================================
+
+-include_lib("stdlib/include/assert.hrl").
+
+%=== MACROS ====================================================================
+
+-define(STARTUP_TIMEOUT, 20000).
+-define(MINING_TIMEOUT,   3000).
+-define(GRACEFUL_STOP_TIMEOUT, 60000).
+
+%=== COMMON TEST FUNCTIONS =====================================================
+
+all() -> [
+    node_can_reuse_db_of_other_node
+].
+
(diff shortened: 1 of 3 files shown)


system_test/aest_db_SUITE.erl
```diff
@@ -0,0 +1,74 @@
...
+    aest_nodes:ct_cleanup(Config).
+
+end_per_suite(_Config) -> ok.
+
+%=== TEST CASES ================================================================
+
+node_can_reuse_db_of_other_node(Cfg) ->
+    DbHostPath = node_db_host_path(node1, Cfg),
+    N1 = node_spec(node1, DbHostPath),
+    N2 = node_spec(node2, DbHostPath),
```
ThomasArts: node1 and node2 have the same database path. On creation node1 puts its node name in the DB. When we then start node2, with a different name, and it works to read that DB, we have successfully replaced the name. OK.

lucafavatella: The name is still [`epoch@localhost`](https://github.com/aeternity/aeternity/blob/fdb574342a3b9ea73bae87b61b82c246427b9aa9/config/vm.args#L1) - for both node1 and node2.

In scope of https://www.pivotaltracker.com/story/show/163423623 , I am aiming at renaming it to `aeternity@localhost` though enabling reuse of database (potentially via operational script invocation) so to avoid users to need to re-download chain. This PR is planting the test that DB can be reused across nodes - still without any renaming done.

----

> As part of https://www.pivotaltracker.com/story/show/163423623 I am planting tests for handling DB migration following change of node name inside DB. This PR is a **RFC for where to place such tests**. It contains a basic test that is meant to pass without code changes, so it can be considered an increase in test coverage for the codebase and Dockerfile.

lucafavatella:
Can be merged to minerva branch once minerva branch stable. This PR increases test coverage.

lucafavatella:
Rebased on latest stable version of minerva branch.

lucafavatella:
Please merge rather than shashing or rebasing.

PR #2160, opened 2019-02-15
Correct template file name for local-only test

lucafavatella:


lucafavatella:
No way this can affect CI.

PR #2163, opened 2019-02-15
Clarify sync state type

lucafavatella:
Prepared while investigating https://github.com/aeternity/aeternity/pull/2159#issuecomment-463729997


lucafavatella:
Waiting for clarification on #2089 before removing WIP label, so to save effort on checking if it introduces warnings with new OTP.

PR #2164, opened 2019-02-15
Token migration for minerva hard fork

velzevur:
[Prepare token migration procedure](https://www.pivotaltracker.com/story/show/163858535)

Currently for each fork there is a subdir in `data/aecore` - for `roma` release it is `.genesis` (as it seeds the genesis block) and for `minerva` there is a `.minerva` dir.
* `accounts.json` is the file for main net
* `accounts_test.json` is for tests
* `accounts_uat.json` is for UAT (or rather - for all the rest cases)

The actual token migration file `data/aecore/.minerva/accounts.json` is to be added in a subsequent PR.


data/aecore/.minerva/accounts_uat.json
```diff
@@ -0,0 +1,4 @@
+{
+   "ak_2a1j2Mk9YSmC1gioUq4PWRm3bsv887MbuRVwyv4KaUGoR1eiKi": 100000000000002
```
dincho: I see you changed the amount, so probably possible

Can you please check if we still own the private key and set a huge amount?

velzevur: Dincho, this is just a dummy address to prove fork works as expected. Provided with the correct address - we can seed it

velzevur:
I've started a node and connected it to UAT to ensure that somehow the PR did not change the genesis hash of UAT
```
peers:
  - aenode://pp_QU9CvhAQH56a2kA15tCnWPRJ2srMJW8ZmfbbFTAy7eG4o16Bf@52.10.46.160:3015
  - aenode://pp_2vhFb3HtHd1S7ynbpbFnEdph1tnDXFSfu4NGtq46S2eM5HCdbC@18.195.109.60:3015
  - aenode://pp_27xmgQ4N1E3QwHyoutLtZsHW5DSW4zneQJ3CxT5JbUejxtFuAu@13.250.162.250:3015
  - aenode://pp_nt5N7fwae3DW8Mqk4kxkGAnbykRDpEZq9dzzianiMMPo4fJV7@18.130.148.7:3015

fork_management:
  network_id: ae_uat

chain:
  persist: false # or true :)
  hard_forks:
    "1": 0
    "2": 10000000000
```

velzevur:
Ran a similar test so the genesis of main net is unchanged as well
aeternity.yaml:
```
chain:
  persist: false # or true :)
  hard_forks:
    "1": 0
    "2": 10000000000
```

velzevur:
Also did a manual test for continuing a persisted database
`aeternity.yaml`:
```yaml
chain:
    # Write chain data to disk.
    persist: true
    db_path: ../../../../../mainnet_db/ #
    hard_forks:
        "1": 0
        "2": 10000000000
```

apps/aecore/src/aec_fork_block_settings.erl
```diff
@@ -0,0 +1,76 @@
...
+                end,
+            Accounts =
+                lists:map(
+                    fun({EncodedPubKey, Amt}) ->
+                        {ok, PubKey} = aehttp_api_encoder:safe_decode(account_pubkey, EncodedPubKey),
+                        {PubKey, Amt}
+                    end,
+                    DecodedData),
+            % ensure deterministic ordering of accounts
+            lists:keysort(1, Accounts)
```
gorillainduction: Should we also check for duplicates?

apps/aecore/src/aec_fork_block_settings.erl
```diff
@@ -0,0 +1,76 @@
...
+
+
+-ifdef(TEST).
+accounts_json_file() ->
+    "accounts_test.json".
+-else.
+accounts_json_file() ->
+    case aec_governance:get_network_id() of
+        <<"ae_mainnet">> -> "accounts.json";
+        _                -> "accounts_uat.json"
```
gorillainduction: I would prefer to match on the uat network id explicitly.

And then use `"accounts_test.json"` for all other network_ids

apps/aecore/src/aec_trees.erl
```diff
@@ -232,7 +233,17 @@ gc_cache(Trees, TreesToGC) ->
 perform_pre_transformations(Trees, Height) ->
     Trees0 = aect_call_state_tree:prune(Height, Trees),
     Trees1 = aeo_state_tree:prune(Height, Trees0),
-    set_ns(Trees1, aens_state_tree:prune(Height, ns(Trees1))).
+    Trees2 = set_ns(Trees1, aens_state_tree:prune(Height, ns(Trees1))),
+    case Height =:= 0 of
```
gorillainduction: Compare with `aec_block_genesis:height()` instead?

apps/aecore/src/aec_fork_block_settings.erl
```diff
@@ -0,0 +1,76 @@
...
+-spec preset_accounts(aec_hard_forks:protocol_vsn(), atom()) -> list().
+preset_accounts(Release, ErrorMsg) ->
+    case read_presets(Release) of
+        {error, {_Err, PresetAccountsFile}} ->
+            % no setup, no preset accounts
+            erlang:error({ErrorMsg, PresetAccountsFile});
+        {ok, JSONData} ->
+            DecodedData =
+                try jsx:decode(JSONData) of
+                    %% JSX decodes an empty json object as [{}]:
```
hanssv: Well it returns a `json_term()` and that is always a list of something...

apps/aecore/src/aec_fork_block_settings.erl
hanssv: Depends on how the scraping script is written?! I would guess some people to have migrated in several batches (I think you were even encouraged to try with a smaller amount first?!).

apps/aecore/src/aec_fork_block_settings.erl
hanssv: What he just said (^^)

apps/aecore/src/aec_trees.erl
hanssv: I would consider treating 0 as a fork_height and just have one case statement?

apps/aecore/src/aec_hard_forks.erl
```diff
@@ -54,6 +55,14 @@ check_protocol_version_validity(Version, Height) ->
 protocol_effective_at_height(H) ->
     protocol_effective_at_height(H, protocols()).
 
+-spec is_fork_height(aec_blocks:height()) -> false | {true, protocol_vsn()}.
+is_fork_height(Height) ->
```
hanssv: I wonder if it might be clearer to define this (reverse map) statically, after all it will be true _very_ rarely...

apps/aecore/test/aec_block_fork_tests.erl
```diff
@@ -0,0 +1,103 @@
...
+            assert_balance(T1, Carol, BalC), % Carol is inserted
+            ok
+        end}
+     ]}.
+
+meck_minerva_accounts(AccountsList) ->
+    meck:expect(aec_fork_block_settings, minerva_accounts,
+                fun() -> AccountsList end).
+
+assert_only_accounts_tree_changed(Trees0, Trees1) ->
```
hanssv: I would be tempted to add this check in `aec_trees:perform_pre_transformation` (when at hard_fork height), but maybe that is overkill...

data/aecore/.minerva/accounts_uat.json
hanssv: Just generate (a couple of) new keypair(s) and give the private key(s) to Dincho ;-)

lucafavatella: Required at least 2 accounts:
* One already with account on testnet.
* One without account on testnet.

lucafavatella: I see this in testnet:
http://52.10.46.160:3013/v2/accounts/ak_2a1j2Mk9YSmC1gioUq4PWRm3bsv887MbuRVwyv4KaUGoR1eiKi
balance 89845016266259

lucafavatella: Enhancement proposed in https://github.com/aeternity/aeternity/pull/2171

apps/aecore/src/aec_fork_block_settings.erl
velzevur: Scrapping script does accumulate balances if a certain account has migrated more than once.
Besides that duplicating accounts in the JSON:
* for genesis migration it could crash while creating an account twice. This did not happen :)
* for `minerva` if an account is not present it is created and if it is present - it is updated, so duplicating accounts even if were present, would not impose a problem

apps/aecore/src/aec_block_fork.erl
```diff
@@ -0,0 +1,25 @@
...
+
+-export([apply_minerva/1]).
+
+-spec apply_minerva(aec_trees:trees()) -> aec_trees:trees().
+apply_minerva(Trees) ->
+    Accounts = aec_fork_block_settings:minerva_accounts(), 
+    AccTrees =
+        lists:foldl(
+            fun({Pubkey, Balance}, AccumAccTrees) when
+                  is_integer(Balance) andalso Balance > 0
```
lucafavatella: In genesis a [balance was 0](https://github.com/aeternity/aeternity/blob/feba9b6811df5b910360c669967466d3045e96c5/data/aecore/.genesis/accounts.json#L124). So no balance 0 for minerva hard fork accounts file?

apps/aecore/src/aec_block_fork.erl
```diff
@@ -0,0 +1,25 @@
...
+                  andalso byte_size(Pubkey) =:= 32 ->
+                Account =
+                    case aec_accounts_trees:lookup(Pubkey, AccumAccTrees) of
+                        none -> aec_accounts:new(Pubkey, Balance);
+                        {value, OldAcc} ->
+                            {ok, NewAcc} = aec_accounts:earn(OldAcc, Balance),
```
lucafavatella: No nonce bump. Ok.

apps/aecore/src/aec_block_fork.erl
hanssv: Lets try to remember this...

apps/aecore/src/aec_trees.erl
```diff
@@ -232,7 +233,17 @@ gc_cache(Trees, TreesToGC) ->
 perform_pre_transformations(Trees, Height) ->
     Trees0 = aect_call_state_tree:prune(Height, Trees),
     Trees1 = aeo_state_tree:prune(Height, Trees0),
-    set_ns(Trees1, aens_state_tree:prune(Height, ns(Trees1))).
+    Trees2 = set_ns(Trees1, aens_state_tree:prune(Height, ns(Trees1))),
+    case Height =:= aec_block_genesis:height() of
```
hanssv: I still don't like this code... But it isn't broken.

apps/aecore/src/aec_hard_forks.erl
gorillainduction: It could be done for the static forks (i.e., for mainnet and uat) at least. But on the other hand, it is still a very small map, so it is ok for now.

apps/aecore/test/aec_block_fork_tests.erl
gorillainduction: It could be nice to make util functions for this code at least. To encourage checking properties based on in which tree changes occur.

gorillainduction: But not needed for this PR.

PR #2165, opened 2019-02-15
Pt 162578740 fate implementation spike

happi:


lucafavatella:
I am taking the liberty of tracking as WIP for easing filtering of PRs re minerva.

lucafavatella:
Temporarily labelling as WIP for easing filtering PRs while releasing.

apps/aefate/rebar.config
```diff
@@ -0,0 +1,33 @@
+{minimum_otp_vsn, "20.1"}.
+
+{erl_opts, [debug_info]}.
+
+{escript_incl_apps, [aefate, aebytecode, lager, getopt]}.
+{escript_main_app, aefate}.
+{escript_name, aefate}.
```
lucafavatella: Is there a smoke test for this, hooked to CI?

happi: No. The stand alone engine is only used for debugging now. We might use it for some tests in the future and then this build will be tested.

PR #2172, opened 2019-02-18
Address leftover log names after concurrent merges

lucafavatella:

### DIFF
--- a/config/dev.config
+++ b/config/dev.config
@@ -89,7 +89,7 @@
            ]}
       ]},
       {traces, [
-        {{lager_file_backend, "log/epoch_pow_cuckoo.log"},
+        {{lager_file_backend, "log/aeternity_pow_cuckoo.log"},
          [{module, aeminer_pow_cuckoo}],
          debug}
       ]}
--- a/config/dev.config
+++ b/config/dev.config
@@ -58,15 +58,6 @@
                ]}
              ]}
            ]},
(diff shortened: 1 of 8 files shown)


tolbrino:
This is good once it is rebased on #2173.

lucafavatella:
It is uncler whether setting log level from user config works for the pow log file. [Tracked](https://www.pivotaltracker.com/story/show/164056695).

config/dev.config
```diff
@@ -58,15 +58,6 @@
                ]}
              ]}
            ]},
-           {epoch_pow_cuckoo_lager_event, [
```
jur0: @lucafavatella have you actually tried the logs are still there? My understanding was that this handler should stay and the traces are redirected into it based on the file `log/aeternity_pow_cuckoo.log` in the handler and the same must be set for `lager_file_backend` in the trace.

tolbrino: Traces are essentially redirections of log output. However, the sink is defined as part of the trace definition and must not exist before.

lucafavatella: > have you actually tried the logs are still there?

@jur0 I did not check manually. This was based on code inspection.

I am giving priority to other PRs.

hanssv: I checkout the branch, made a prod-package, started it with mining and got the expected log file (and only the expected log file) including content.

PR #2174, opened 2019-02-18
PT-164050506 Bump minimum gasprice

hanssv:

PR #2177, opened 2019-02-18
PT-164050660 miner minimum gas price

hanssv:


hanssv:
This is built on top of #2174

lucafavatella:
Labelling as WIP for easing filtering.

apps/aecore/src/aec_tx_pool.erl
```diff
@@ -348,67 +354,70 @@ int_get_candidate(MaxGas, BlockHash, #dbs{db = Db} = DBs) ->
     {ok, Trees} = aec_chain:get_block_state(BlockHash),
     {ok, Header} = aec_chain:get_header(BlockHash),
     lager:debug("size(Db) = ~p", [ets:info(Db, size)]),
+    MinMinerGasPrice = aec_tx_pool:minimum_miner_gas_price(),
```
lucafavatella: I see external call for mocking in tests. Ok.

apps/aecore/src/aec_tx_pool.erl
```diff
@@ -348,67 +354,70 @@ int_get_candidate(MaxGas, BlockHash, #dbs{db = Db} = DBs) ->
...
                 ?KEY(_, _, Account, Nonce, TxHash) = Key, Tx,
-                AccountsTree, Height, Gas, Acc) ->
+                AccountsTree, Height, Gas, MinMinerGasPrice, Acc) ->
     Tx1 = aetx_sign:tx(Tx),
     TxTTL = aetx:ttl(Tx1),
     TxGas = aetx:gas_limit(Tx1, Height),
-    case Height < TxTTL andalso TxGas > 0 andalso ok =:= int_check_nonce(Tx, AccountsTree, check_candidate) of
+    case Height < TxTTL andalso TxGas > 0
+         andalso ok =:= int_check_nonce(Tx, AccountsTree, check_candidate)
+         andalso MinMinerGasPrice =< aetx:min_gas_price(Tx1, Height) of
```
lucafavatella: I see aetx:min_gas_price uses aetx:gas_limit as denominator, and aetx:gas_limit returns zero as error, though this is safe here because `TxGas > 0` present in andalso close preceding. Ok.

apps/aetx/src/aetx.erl
```diff
@@ -209,6 +210,12 @@ gas_price(#aetx{ type = Type, cb = CB, tx = Tx }) when ?IS_CONTRACT_TX(Type) ->
 gas_price(#aetx{}) ->
     undefined.
 
+-spec min_gas_price(Tx :: tx(), Height :: aec_blocks:height()) -> MinGasPrice :: non_neg_integer().
+min_gas_price(AETx = #aetx{ type = Type, cb = CB, tx = Tx }, Height) when ?IS_CONTRACT_TX(Type) ->
+    min(CB:gas_price(Tx), 1 + (CB:fee(Tx) div gas_limit(AETx, Height)));
+min_gas_price(AETx = #aetx{ cb = CB, tx = Tx }, Height) ->
+    1 + (CB:fee(Tx) div gas_limit(AETx, Height)).
```
lucafavatella: So if fee comes up as 1000 and gas_limit as 100, then this function returns 11 rather than 10?

apps/aetx/src/aetx.erl
```diff
@@ -209,6 +210,12 @@ gas_price(#aetx{ type = Type, cb = CB, tx = Tx }) when ?IS_CONTRACT_TX(Type) ->
 gas_price(#aetx{}) ->
     undefined.
 
+-spec min_gas_price(Tx :: tx(), Height :: aec_blocks:height()) -> MinGasPrice :: non_neg_integer().
+min_gas_price(AETx = #aetx{ type = Type, cb = CB, tx = Tx }, Height) when ?IS_CONTRACT_TX(Type) ->
+    min(CB:gas_price(Tx), 1 + (CB:fee(Tx) div gas_limit(AETx, Height)));
```
lucafavatella:
```suggestion
    %% There are two limit cases: zero gas used for contract execution; max gas used for contract execution.
    min( CB:fee(Tx) div min_gas(AETx)
       , (CB:fee(Tx) + (CB:gas(Tx) * CB:gas_price(Tx))) div gas_limit(AETx, Height)
       );
```

lucafavatella: So if fee comes up as 1000 and gas_limit as 100 and gas price as 99999999, then this function clause returns 11 rather than 10?

apps/aetx/src/aetx.erl
hanssv: Good catch, it should be
```
GasLimit = gas_limit(AETx, Height),
(CB:fee(Tx) + GasLimit - 1) div GasLimit.
```

apps/aetx/src/aetx.erl
hanssv: Good catch, it should be

```
GasLimit = gas_limit(AETx, Height),
(CB:fee(Tx) + GasLimit - 1) div GasLimit.
```

hanssv: No, I don't agree... I viewed the implicit gas price of the Fee as one gas price, and the dynamic gas price as a second gas price, simply returning the smallest of the two... One could try to compute some sort of average gas price like you suggest but that is probably harder to reason about for the miner?

lucafavatella: Agreed, easier to reason about.

Also the minimum computatio may be more complex:
```
   %% Both the miner reward and the tx gas increase linearly with the gas used:
    %% * Miner reward: `CB:fee(Tx) + (<gas used> * CB:gas_price(Tx))`.
    %% * Tx gas: `min_gas(AETx) + <gas used>`.
    %%
    %% What is the minimum of the ratio between the miner reward and the tx gas?
```

PR #2182, opened 2019-02-19
Speed up osx package builds pt 163453734

valerifilipov:
https://www.pivotaltracker.com/story/show/163453734
Add circleci cache for homebrew install folders:
- restore cache takes ~ 30 secs to download and unarchive
- restored brew installs require to be relinked, which takes another ~20 secs
overall this decreases macos package build with ~ 3 mins
Other than that I don't know how to speed up this job more
https://circleci.com/gh/aeternity/aeternity/8645

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -517,7 +517,10 @@ jobs:
           command: |
             brew update
             brew install libsodium
+            brew link libsoidium
             brew install file://`pwd`/deployment/homebrew/erlang.rb
+            brew link erlang
+            for erlang_dep in $(brew deps erlang); do brew link $erlang_dep; done
       - *save_macos_package_cache
       - *set_package_path
       - *build_package
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,12 +107,10 @@ references:
       paths:
         - "_build"
 
-  macos_package_cache_key: &macos_package_cache_key macos-package-cache-v1-{{ .Branch }}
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -469,12 +482,16 @@ jobs:
     working_directory: ~/aeternity
     steps:
       - checkout
+      - *restore_macos_package_cache
       - run:
           name: Install required tools
           command: |
-            brew update
```
lucafavatella: I see this removes [`update`](https://docs.brew.sh/Manpage#update-options) while it is recommended in [the documentation](https://github.com/aeternity/aeternity/blame/86430a4cfe0ff773f709a5b24056bbd64b1d09a0/docs/installation.md#L73-L74). This may cause users to spot issues before our CI.

lucafavatella: I see this removes [`update`](https://docs.brew.sh/Manpage#update-options) while not pinning libsodium to a specific version. Could this cause old (insecure?) version to unexpectedly be used in package? I suggest to either keep the update, or pin the version.

valerifilipov: I saw that every `brew install` command invokes update, that's why i removed it

lucafavatella: Could you please add your rationale in the CI config file, or at least in the PR.
I see myself having the same doubt when seeing that CI code path again...

valerifilipov: I brought it back, as it is more clear what it's doing

PR #2183, opened 2019-02-19
Make system test consider old user API withouth info in block

lucafavatella:
Follows https://github.com/aeternity/aeternity/pull/2178#issuecomment-465173509


lucafavatella:
I checked the changed test was run and passed.

I see one failing integration test on CI though this PR only affects system tests so unrelated.

PR #2191, opened 2019-02-20
Convert contract transactions from old persisted format

gorillainduction:
https://www.pivotaltracker.com/story/show/164113577

### DIFF
--- a/apps/aecontract/src/aect_call_tx.erl
+++ b/apps/aecontract/src/aect_call_tx.erl
@@ -43,6 +43,10 @@
          call_data/1,
          call_stack/1]).
 
+%% For converting old db format
+-export([from_db_format/1
+        ]).
+
 -define(CONTRACT_CALL_TX_VSN, 1).
 -define(CONTRACT_CALL_TX_TYPE, contract_call_tx).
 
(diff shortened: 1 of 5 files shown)


apps/aecontract/src/aect_create_tx.erl
```diff
@@ -72,6 +77,15 @@
 
 -export_type([tx/0]).
 
+%%%===================================================================
+%%% Conversion of old db format
+
+-spec from_db_format(tx()) -> tx().
+from_db_format(#contract_create_tx{ct_version = #{}} = Tx) ->
+    Tx;
+from_db_format(#contract_create_tx{ct_version = VMVersion} = Tx) ->
```
lucafavatella: Guard on type of VMVersion would have made this more readable though minor and debatable. Ok.

apps/aecontract/src/aect_call_tx.erl
```diff
@@ -104,6 +123,46 @@ new(#{caller_id   := CallerId,
...
+           contract_id      = ContractId,
+           abi_version      = ABI,
+           fee              = Fee,
+           ttl              = TTL,
+           amount           = Amount,
+           gas              = Gas,
+           gas_price        = GasPrice,
+           call_data        = CallData,
+           call_stack       = CallStack
+          } ->
```
lucafavatella: So the absence of field `call_origin` in record `db_contract_call_tx` is used as signal for detecting old version. Ok.
I see not int/map different representation for any version field so no transformation required for any field. Ok.

apps/aecontract/src/aect_call_tx.erl
```diff
@@ -104,6 +123,46 @@ new(#{caller_id   := CallerId,
...
+               fee              = Fee,
+               ttl              = TTL,
+               amount           = Amount,
+               gas              = Gas,
+               gas_price        = GasPrice,
+               call_data        = CallData,
+               call_stack       = CallStack,
+               call_origin      = CallOrigin
+              };
+        _ ->
```
ThomasArts: is there a risk that one would call this function once more after the first conversion and then fail with an error?

gorillainduction: It will then be caught by the previous function clause.

PR #2193, opened 2019-02-21
Plant release 2.0.0 skeleton

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/163511725


lucafavatella:
This is the same as https://github.com/aeternity/aeternity/pull/2192 though with less ambiguous branch name.

PR #2195, opened 2019-02-21
Fix rebar3 escriptize hook handling

tolbrino:
The rebar3 binary is based on v3.9.0. Therefore, you'll need to rebase this branch onto the latest master next, to get the whole compilation working.

I'll get the fix into rebar3 upstream separately.


happi:
Cherry picked.

PR #2200, opened 2019-02-25
Fix target rename leftover

lucafavatella:
Follows https://github.com/aeternity/aeternity/commit/0b52ae5e#diff-b67911656ef5d18c4ae36cb6741b7965R178


lucafavatella:
I see target all-tests never used so it cannot fail CI. Merging without waiting for remaining CI jobs.

PR #2201, opened 2019-02-25
PT-163485144 Expose log/events in ContractCallObject in HTTP API

hanssv:


hanssv:
This should be rebased on `fortuna` once master has been merged to `fortuna` I guess... Don't review just yet...

apps/aecontract/src/aect_call.erl
```diff
@@ -198,9 +198,9 @@ serialize_for_client(#call{caller_id    = CallerId,
      }.
 
 serialize_log_entry_for_client({Address, Topics, Data}) ->
-    #{ <<"address">> => Address
-     , <<"topics">>  => Topics
-     , <<"data">>    => Data
+    #{ <<"address">> => aehttp_api_encoder:encode(contract_pubkey, Address)
+     , <<"topics">>  => [ Topic || <<Topic:256>> <- Topics ]
```
lucafavatella: This syntax is filtering out Topics that do not match `<<_:256>>`: who guarantees that this constraint holds for all topics? Or does code here intend to filter?

lucafavatella: Could you please track future improvement of restricting module to check that stored log (assert_field?) matches this expectation.

hanssv: All topics are of VM type `word`

lucafavatella: https://www.pivotaltracker.com/story/show/164431952

PR #2211, opened 2019-03-01
Upgrade compiler and fix calldata creation

hanssv:
Bumping to a more recent version of `aesophia` that contains changes to simplify the upcoming standalone (HTTP) compiler. This changes how calldata is created (now from source rather than bytecode since it is compiler-side) and this PR makes the necessary updates to the main node. This is mainly tests that need fixing, but also relying only on bytecode and the bundled old compiler in the main node.


rebar.lock
```diff
@@ -18,7 +18,7 @@
   0},
  {<<"aesophia">>,
   {git,"https://github.com/aeternity/aesophia.git",
-      {ref,"2edafe0adcfea7dcfc52cbe8fe28bf7cace04bfa"}},
+      {ref,"4d9d3077ada60baa4ced35a793555888c9684cac"}},
```
lucafavatella: Based in quick read of aect_sophia, I see that compiler version is not yet included in transactions, so I expect this not applicable now. Though I wonder - I see [same version returned by compiler](https://github.com/aeternity/aesophia/compare/2edafe0adcfea7dcfc52cbe8fe28bf7cace04bfa...4d9d3077ada60baa4ced35a793555888c9684cac#diff-89d02cf039adcda2996076326bc5b76fR79): should it have been bumped? What are criteria and process for ensuring that compiler with not-bumped-version-compared-to-last-release-when-it-should-have-been does not enter master?

hanssv: Yes, we need a process for version of the compiler. There are a few things missing in `aesophia` that we need for the "Minerva" standalone compiler/HTTP API so I expect a new version early next week... And then we should probably tag it and reference that tag here!!

PR #2216, opened 2019-03-05
PT-163476661 Add sync_progress field to /status endpoint

zp-sd:
Superseded by https://github.com/aeternity/aeternity/pull/2219.

### DIFF
--- a/apps/aecore/src/aec_sync.erl
+++ b/apps/aecore/src/aec_sync.erl
@@ -24,6 +24,8 @@
         , has_generation/1
         , set_last_generation_in_sync/0 ]).
 
+-export([sync_progress/0]).
+
 %% gen_server callbacks
 -export([init/1, handle_call/3, handle_cast/2,
          handle_info/2, terminate/2, code_change/3]).
@@ -63,6 +65,9 @@ gossip_txs(GossipTxs) ->
 sync_in_progress(PeerId) ->
     gen_server:call(?MODULE, {sync_in_progress, PeerId}).
 
+sync_progress() ->
+    gen_server:call(?MODULE, sync_progress).
+
 known_chain(Chain, ExtraInfo) ->
     gen_server:call(?MODULE, {known_chain, Chain, ExtraInfo}).
 
@@ -90,7 +95,7 @@ handle_worker(Task, Action) ->
 %% 1. It has worst top hash than our node, do not include in sync
 %% 2. It has better top hash than our node
 %%    We binary search for a block that we agree upon (could be genesis)
-%% We add new node to sync pool to sync agreed block upto top of new
+%% We add new node to sync pool to sync agreed block up to top of new
 %% 1. If we are already synchronizing, we ignore it,
 %%    the ongoing sync will pick that up later
 %% 2. If we are not already synchronizing, we start doing so.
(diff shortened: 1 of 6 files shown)


apps/aecore/src/aec_sync.erl
```diff
@@ -158,6 +163,23 @@ handle_call({next_work_item, STId, PeerId, LastResult}, _From, State) ->
     {reply, Reply, State2};
 handle_call({gossip_txs, GossipTxs}, _From, State) ->
     {reply, ok, State#state{ gossip_txs = GossipTxs }};
+handle_call(sync_progress, _, State) ->
+    State#state.sync_tasks,
+    SyncProgress =
+        case State#state.sync_tasks of
+            [] -> 100.0;
```
hanssv: This means that the node will report 100% for a short while after starting, but I think that is ok.

zp-sd: @hanssv Yes, this is rather approximate number... There are several issues:
* 100% when there are no peers in the config may be misleading, or just no sync tasks (not sure exactly when this could be the case, I haven't grasped all code paths around sync yet),
* relying on `aec_headers:height(aec_chain:top_header())` - we don't know if we're syncing with the chain that is based on the current top header in our DB?! On the other hand if we use height from `#sync_task.agreed` instead, that's not the height that has been already added to the database, but what will be added in the current sync worker step.... So that's not perfect as well... (And relying on `#sync_task.agreed` would save one db call for top header, shouldn't be an argument really 🙂 ).

I'd say this is good enough for now?...

dincho: Why not reporting 0 on start, sounds much more reasonable to me?

dincho: Ah, we can't differentiate between 100% and 0% .. I see.

zp-sd: Yes, technically, we just don't have sync task (maybe yet... maybe already...). We could try to code some bigger machinery, but I'm not sure if that's worth doing.

apps/aecore/src/aec_sync.erl
```diff
@@ -158,6 +163,23 @@ handle_call({next_work_item, STId, PeerId, LastResult}, _From, State) ->
...
+            SyncTasks ->
+                TargetHeight =
+                    lists:foldl(
+                      fun(SyncTask, MaxHeight) ->
+                              #{chain := Chain} = SyncTask#sync_task.chain,
+                              [#{height := Height} | _] = Chain,
+                              max(Height, MaxHeight)
+                      end, 0, SyncTasks),
+                TopHeight = aec_headers:height(aec_chain:top_header()),
+                round(10000000 * TopHeight / TargetHeight) / 100000
```
dincho: Why such hight precision here, I'd not expect more than double number precision to have any meaning

zp-sd: It can be double precision if you prefer, we're very well prepared for long chains now (5 digits after `.`) 😉 . I don't care much 🙂

zp-sd: No, I just recalled why double in not enough already (I remembered changing that, didn't remember why though 🙂 ):

```
➜  aeternity git:(pt-163476661-add-sync_progress-field-to-status-endpoint) curl 127.0.0.1:3013/v2/status
{"difficulty":46614936315406,"genesis_key_block_hash":"kh_pbtwgLrNu23k9PA6XCZnUbtsvEFeQGgavY4FS2do3QP8kcp2z","listening":true,"network_id":"ae_mainnet","node_revision":"0ea51ca9cfdd972506a656cd390f156d145f4bd0","node_version":"2.0.0","peer_count":390,"pending_transactions_count":222,"protocols":[{"effective_at_height":47800,"version":2},{"effective_at_height":0,"version":1}],"solutions":0,"sync_progress":95.68505,"syncing":true}%                                                                                                                                            ➜  aeternity git:(pt-163476661-add-sync_progress-field-to-status-endpoint) curl 127.0.0.1:3013/v2/status
{"difficulty":46614936315406,"genesis_key_block_hash":"kh_pbtwgLrNu23k9PA6XCZnUbtsvEFeQGgavY4FS2do3QP8kcp2z","listening":true,"network_id":"ae_mainnet","node_revision":"0ea51ca9cfdd972506a656cd390f156d145f4bd0","node_version":"2.0.0","peer_count":390,"pending_transactions_count":222,"protocols":[{"effective_at_height":47800,"version":2},{"effective_at_height":0,"version":1}],"solutions":0,"sync_progress":95.68718,"syncing":true}
```

Note `95.68505` vs `95.68718`.

zp-sd:
This PR will be superseded by https://github.com/aeternity/aeternity/pull/2219. Cc @hanssv @dincho

zp-sd:
Superseded by https://github.com/aeternity/aeternity/pull/2219.

PR #2222, opened 2019-03-06
Plant release 2.0.1 skeleton

tolbrino:


docs/release-notes/RELEASE-NOTES-2.0.1.md
```diff
@@ -0,0 +1,106 @@
+# About this release
+
+[This][this-release] is a bugfix release.
+It contains:
```
dincho: Maybe mention it will have *only* windows binaries as there are no related changes to the node but only to windows scripts.

PR #2226, opened 2019-03-07
Move the install.sh to it s own repo 164011629

michalbagrowski:


docs/installation.md
```diff
@@ -3,7 +3,7 @@
 Run below command to install latest version of aeternity node.
 
 ```bash
-curl -O https://raw.githubusercontent.com/aeternity/aeternity/v2.0.0/scripts/install.sh && chmod +x install.sh && ./install.sh 2.0.0
+curl -O https://raw.githubusercontent.com/aeternity/installer/master/scripts/install.sh && chmod +x install.sh && ./install.sh 2.0.0
```
dincho: let's keep using tags, once https://github.com/aeternity/installer/pull/2/ is merged we should tag `v1.0.0`

docs/installation-windows.md
```diff
@@ -16,24 +16,24 @@ NOTE: These steps describe the setup for a basic CPU mining configuration.
 
 ## Step 1: Get helper scripts
 
-The helper scripts are located in the `scripts/` folder in the `aeternity` Git repository.
+The helper scripts are located in the `intaller` Git repository.
```
dincho: typo

dincho:
> It would be nice to have the supported paths of the installer tested on CI though I understand out of scope of this PR so ok.

Installer should be considered as separate component after this is merged, and there are couple of tests already in it's repo.

What I mean is that the node development should not try to keep the installer compatible, but the opposite. The installer should follow the node changes.

Keeping bi-directional dependancy would be hard.

dincho:
> I am concerned both sets of scripts will accumulate independent changes.

I don't see a reason to keep windows build scripts in the installer repo anyway: https://github.com/aeternity/installer/pull/6

docs/installation-windows.md
```diff
@@ -16,24 +16,24 @@ NOTE: These steps describe the setup for a basic CPU mining configuration.
...
 
-The helper scripts are located in the `scripts/` folder in the `aeternity` Git repository.
+The helper scripts are located in the `installer` Git repository.
 You may download the whole repository by
 
-1. Downloading the [Git repository](https://github.com/aeternity/aeternity) itself;
+1. Downloading the [Git repository](https://github.com/aeternity/installer) itself;
 2. Or downloading and unpacking a
-   [Zip archive](https://github.com/aeternity/aeternity/archive/master.zip)
+   [Zip archive](https://github.com/aeternity/installer/archive/v1.0.0.zip)
```
lucafavatella: So the tag of the installer must be available before the tag of the node. Ok.

tolbrino:
> I see the three scripts under scripts/windows are duplicated of https://github.com/aeternity/installer/tree/v1.0.0/windows (I did not check content/checksum). Could you please identify which set of scripts is the canonical one, and add a comment/link to the top of non-canonical file mentioning where the canonical one can be found?
> 
> I am concerned both sets of scripts will accumulate independent changes.

The reference files are the ones in this repository, because these are used by the CI as well. I believe the installer should actually combine these into a single file just like it is done for Unix systems.

dincho:
> This might be a good opportunity to sunset the WSL support in the installer repo. Apart from that there should be a unified installer for Windows which downloads and sets up the latest node package just like `install.sh` doesn.

good idea

PR #2227, opened 2019-03-07
Move ownership of gc db to aec_tx_pool for state consistency

gorillainduction:
https://www.pivotaltracker.com/story/show/164430876

### DIFF
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -52,8 +52,10 @@
         , raw_delete/2
         ]).
 
+-include("aec_tx_pool.hrl").
+
 -ifdef(TEST).
--export([garbage_collect/1]). %% Only for (Unit-)test
+-export([sync_garbage_collect/1]). %% Only for (Unit-)test
 -export([restore_mempool/0]).
 -export([peek_db/0]).
 -export([peek_visited/0]).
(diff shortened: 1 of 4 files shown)


apps/aecore/test/aec_tx_pool_tests.erl
```diff
@@ -734,9 +734,7 @@ tx_pool_test_() ->
      ]}.
 
 tx_pool_gc(Height) ->
-    aec_tx_pool_gc:gc(Height),
-    sys:get_status(aec_tx_pool_gc), %% sync point (gc is asynchronous)
```
lucafavatella: This may leave a harmless redundant call in test - see also https://github.com/aeternity/aeternity/pull/2229

apps/aecore/src/aec_tx_pool.erl
```diff
@@ -260,6 +264,9 @@ init([]) ->
     {ok, _MempoolDb} = pool_db_open(pool_db()),
     {ok, _VisitedDb} = pool_db_open(pool_db_visited()),
     {ok, _NonceDb} = pool_db_open(pool_db_nonce()),
+    %% The gc db should be owned by this process to ensure that the gc state
+    %% is consistent with the actual mempool if any of the servers restart.
+    {ok, _GCDb} = pool_db_open(pool_db_gc(), [{keypos, #tx.hash}]),
```
lucafavatella: I see same set of options as before. Ok.

lucafavatella: I see this includes `public` so no read-write issues for distinct process. Ok.

PR #2228, opened 2019-03-07
Log (debug) reason for invalid tx on external API

lucafavatella:
I am used this locally for debugging tests amended in https://github.com/aeternity/aeternity/pull/2214 and it seems a reasonable addition to logs at debug level. The error reasons are not long.

### DIFF
--- a/apps/aehttp/src/aehttp_dispatch_ext.erl
+++ b/apps/aehttp/src/aehttp_dispatch_ext.erl
@@ -432,10 +432,12 @@ handle_request_('PostTransaction', #{'Tx' := Tx}, _Context) ->
                         ok ->
                             Hash = aetx_sign:hash(SignedTx),
                             {200, [], #{<<"tx_hash">> => aehttp_api_encoder:encode(tx_hash, Hash)}};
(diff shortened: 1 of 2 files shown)


apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -432,10 +432,12 @@ handle_request_('PostTransaction', #{'Tx' := Tx}, _Context) ->
                         ok ->
                             Hash = aetx_sign:hash(SignedTx),
                             {200, [], #{<<"tx_hash">> => aehttp_api_encoder:encode(tx_hash, Hash)}};
-                        {error, _} ->
+                        {error, Reason} ->
+                            lager:debug("Invalid tx: ~p", [Reason]),
```
lucafavatella: Actually I see similar message already logged in [`aec_tx_pool:check_pool_db_put`](https://github.com/aeternity/aeternity/blob/831e1ee30e4092058ca35573bdfd8dd5f392c01d/apps/aecore/src/aec_tx_pool.erl#L665-L667).

PR #2233, opened 2019-03-11
Debian/Ubuntu packaging. PT-162439771

valkov-ivaylo:
The changes are focused on Ubuntu (*.deb) package generation. Added:

 - required Debian/Ubuntu files for package generation
 - added documentation for the internals of a package and build procedure.
 - added Makefile rules to generate/build package on Ubuntu 18.04 
 - building tested against [builder container](../../docker-builder)
 - package creates system user that runs the node
 - package adds systemd service file that runs the node as a service/daemon
 - logging is relocated entirely to /var/log/aeternity/node
 - node files are located in /opt/aeternity/node
 - package operations tested on VM and partly in container.

Missing:
 - procedure does not work on Ubuntu 16.04. Special requirements and hacks are required. On hold.
 - log file rotation rules (in OS context; /etc/logrotate.d/)
 
Documentation requires proof-reading - is everything clear and usable.


debian/changelog
```diff
@@ -0,0 +1,5 @@
+aeternity-node (1.1.0) unstable; urgency=medium
```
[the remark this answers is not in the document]
valkov-ivaylo: In Debian/Ubuntu packaging context, upstream should be us. :) This is the initial version hard-coded in the changelog file when packaging was started. A changelog file must be present before using other packaging tools. On package building `dch` generates proper entries with the correct version from the VERSION file. The latest version in this file is always the one in the VERSION file in the current source tree. Check the Makefile rules.

There might be a way to avoid this initial dummy version, but I had no time to play with `dch` with non-existing changelog file.

debian/aeternity-node.links
```diff
@@ -0,0 +1,2 @@
+var/log/aeternity/node opt/aeternity/node/log
+opt/aeternity/node/docs usr/share/doc/aeternity-node/docs
```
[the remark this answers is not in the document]
valkov-ivaylo: Work-arounds.  

Packaging tools will create the path `usr/share/doc/aeternity-node/` to populate changelog and similar files. We cannot link to it, because it exists. This requires the additional sub-directory.

The path is the standard place for documentation. From user's point of view it is a good idea to have a way to reach the docs from there.

We cannot copy the docs directly there without duplicating files on disk with the `docs`(Ubutnu/Debian context) file. The `install` file does not provide a way to exclude `docs/`. The files in the `_build` directory seem to vary and cause issues, so a wildcard is required instead of listing desired files.

debian/rules
```diff
@@ -0,0 +1,35 @@
...
+export DEB_BUILD_MAINT_OPTIONS=hardening=-all
+
+%:
+	dh $@
+#--with rebar
+
+override_dh_auto_build:
+	$(MAKE) prod-build
+
+override_dh_auto_clean:
```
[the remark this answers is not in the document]
valkov-ivaylo: Left overs from debugging. Probably should uncomment the lines bellow. Right now it blocks the general debian-helper (dh) rules from running which I think gave errors.

Makefile
```diff
@@ -19,6 +19,38 @@ SWAGGER_ENDPOINTS_SPEC = apps/aeutils/src/endpoints.erl
 
 PACKAGE_SPEC_WIN32 ?= ../ci/appveyor/package.cfg
 
+AE_DEB_VERSION=`cat VERSION`
+AE_DEB_DCH_REL_NOTE= \
+"Release notes are available in /usr/share/doc/aeternity-node/docs/release-notes/RELEASENOTES-`cat VERSION`.md"
```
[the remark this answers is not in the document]
valkov-ivaylo: Typo. Fill fix it.

debian/changelog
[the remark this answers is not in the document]
dincho: What about empty file, this information is useless anyway as we're not going to track the changelog. Users should be point to the docs/release-notes of a specific commit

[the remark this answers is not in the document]
valkov-ivaylo: An empty file will not work, already tested. The changelog file is mandatory for package creation. This is the place, from where the version of the software is determined in packaging context. The file is already a hack. When generated, it will just contain a note with the path to the release notes file. 


Example:

```
aeternity-node (2.2.0) bionic; urgency=medium

  * Release notes are available in /usr/share/doc/aeternity-
    node/docs/release-notes/RELEASE-NOTES-2.2.0.md

 -- Aeternity Team <info@aeternity.com>  Thu, 14 Mar 2019 08:38:12 +0000

aeternity-node (1.1.0) unstable; urgency=medium

  * Initial Debianization

 -- Valkov Ivaylo <ivaylo.valkov@quanterall.com>  Tue, 18 Dec 2018 09:43:00 +0200

```

I am aware it will not be tracked, that is why the approach taken is to generate it automatically. The situation about it is documented. However there is no work-around without it. 

The only other option is generating binary only packages with a lot of custom scripts/Makefiles. It is hard to maintain and proven ineffective. I am not in favor of this approach.

Makefile
```diff
@@ -19,6 +19,42 @@ SWAGGER_ENDPOINTS_SPEC = apps/aeutils/src/endpoints.erl
...
```
lucafavatella:
```suggestion
"Release notes are available in /usr/share/doc/aeternity-node/docs/release-notes/RELEASE-NOTES-`cat VERSION`.md"
```

debian/control
```diff
@@ -0,0 +1,14 @@
+Source: aeternity-node
+Section: misc
+Maintainer: Aeternity Team <info@aeternity.com>
+Build-Depends: debhelper, erlang ( >=1:20.2-1),  erlang-base ( >=1:20.2-1), erlang-dev ( >=1:20.2-1),
```
lucafavatella: Could you prevent OTP 21? We are not really supporting it for now.

debian/aeternity-node.service
```diff
@@ -0,0 +1,17 @@
...
+Requires=network-online.target
+After=network-online.target
+
+[Service]
+Type=forking
+User=aeternity
+Group=aeternity
+ExecStart=/opt/aeternity/node/bin/aeternity start
+ExecStop=/opt/aeternity/node/bin/aeternity stop
+Restart=on-failure
```
lucafavatella: Could you please clarify your exploration and design of integration of `bin/aeternity start` with this init system?

Makefile
```diff
@@ -375,6 +411,13 @@ internal-clean: $$(KIND)
 internal-distclean: $$(KIND)
 	@rm -rf ./_build/$(KIND)
 
+prod-deb-package:
+	export DEBEMAIL=$(AE_DEB_EMAIL); \
+	export DEBFULLNAME=$(AE_DEB_NAME) ;\
```
lucafavatella: How does quoting / splitting work here? Could you please review other vars in this recipe too please?

Makefile
```diff
@@ -19,6 +19,42 @@ SWAGGER_ENDPOINTS_SPEC = apps/aeutils/src/endpoints.erl
...
+AE_DEB_VERSION=`cat VERSION`
+AE_DEB_DCH_REL_NOTE= \
+"Release notes are available in /usr/share/doc/aeternity-node/docs/release-notes/RELEASENOTES-`cat VERSION`.md"
+
+AE_DEB_EMAIL="info@aeternity.com"
+AE_DEB_NAME="Aeternity Team"
+
+
+# Packages from master MUST be pre-releases. Git master version
+# usually is always higher then the last stable release. However
```
lucafavatella: > Git master version usually is always higher then the last stable release.

I.e. usually when tagging&releasing v2.1.0 we shortly after (in a couple of days) bump VERSION in master as 2.2.0. Correct. Ok.

lucafavatella: Usually and always seem conflicting.
```suggestion
# usually is higher then the last stable release. However
```

Makefile
```diff
@@ -19,6 +19,42 @@ SWAGGER_ENDPOINTS_SPEC = apps/aeutils/src/endpoints.erl
...
+	AE_DEB_GIT_MASTER=true
+  endif
+else
+  ifneq (, $(shell git rev-parse --abbrev-ref HEAD|grep master))
+	AE_DEB_GIT_MASTER=true
+  endif
+endif
+
+ifeq ($(AE_DEB_GIT_MASTER), true)
+        AE_DEB_VERSION=`cat VERSION`"-1-git-master~"`date +%Y%m%d%H%M%S`
```
lucafavatella: Thanks for the comment above, though I do not manage to follow fully so I feel unable in short time to advise on correctness of version scheme.

valkov-ivaylo:
> How is the build of the package hooked to CI?
> 
It is not yet. It will be covered in PT-162439780

> How is the smoke test of the built package hooked to CI?

It has not been discussed yet with @dincho. Probably will also be covered by PT-162439780.

The current state just allows to build a Debian/Ubuntu package manually.

Makefile
```diff
@@ -19,6 +19,42 @@ SWAGGER_ENDPOINTS_SPEC = apps/aeutils/src/endpoints.erl
...
+
+# Additionally the same (as in name) package from master for the same
+# unreleased/not stable version (i.e. builds) MUST always have higher
+# version (i.e. pre-release). Otherwise package managers and repository management
+# software complain.
+
+# Test source for master branch. Needs special treatment for Debian
+# packages.
+ifeq (, $(shell which git))
+  ifneq (,$(shell cat .git/HEAD | grep master))
```
lucafavatella: This is extremely fragile. What are use cases?

In local copy I never checkout master branch. I am usually in detached HEAD mode (`origin/master`, or other) or branch with other name. Also, if checked out on a branch, the branch name may sometimes contain master (e.g. `master--check?`). Is this only for CI?

Would it be possible to rather reverse the logic, by setting `AE_DEB_VERSION` to the content of the VERSION file only if a special var (`AE_DEB_STABLE`?) is passed by the caller, and always use this pre-release (complex) version otherwise?

Makefile
```diff
@@ -19,6 +19,42 @@ SWAGGER_ENDPOINTS_SPEC = apps/aeutils/src/endpoints.erl
...
+	AE_DEB_GIT_MASTER=true
+  endif
+else
+  ifneq (, $(shell git rev-parse --abbrev-ref HEAD|grep master))
```
lucafavatella: Fragile too. See above.

Makefile
valkov-ivaylo: This approach is tested in real world in a previous company. I really do not have a better approach. It is strictly Ubuntu/Debian package versioning context (`man deb-version`). If there were only stable releases it is not needed. The plan is to have packages from master as well according to @dincho

debian/aeternity-node.service
valkov-ivaylo: I guess your question is, is it working?  Tested with current code-base on a VM and no issues were observed. If you have suggestions for tests that we can run against the code running in systemd, they are welcome. 

I've used systemd with Erlang software before in production - the same approach `start`, `stop`. Generally there are no issues.  It is always a good idea to run a service from the init system than as a normal process.

Makefile
valkov-ivaylo: Your suggestion is to quote AE_DEB_NAME and the rest?

```
export DEBEMAIL="$(AE_DEB_EMAIL)"
```
I am not sure if this will affect something. Working on CI integration in a separate branch shows it behaves as expected - no errors were produced because of spaces or new lines.

debian/aeternity-node.service
lucafavatella: I see [ref](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=). May main question is whether this `on-failure` detects the failure. I.e. if you manually brutally kill the node process, does this service restart kick in?

Makefile
lucafavatella: I see AE_DEB_NAME has quotes already `"Aeternity Team"` so maybe those are kept. Ok.

debian/control
valkov-ivaylo: Hm, I am not sure if this could be achieved. The issue will be with [building on 16.04](
https://github.com/aeternity/aeternity/blob/ubuntu_pacakging_162439771/docs/debian_ubuntu_packaging.md#ubuntu-1604-fixes)

The issues is that for Ubuntu/Debian package every dependency should come as a package. Otherwise a package cannot be created. So in 16.04, an  ErlangSolutions Debian repository is used to get Erlang with sufficient version as a package. 

However 18.04 provides this

```
apt-cache policy erlang
erlang:
  Installed: 1:20.2.2+dfsg-1ubuntu2
  Candidate: 1:20.2.2+dfsg-1ubuntu2
  Version table:
 *** 1:20.2.2+dfsg-1ubuntu2 500
        500 http://archive.ubuntu.com/ubuntu bionic/universe amd64 Packages
```

While the usable version as a package for 16.04 from the mentioned repository would be

```
apt-cache policy erlang | grep 20.2
  Candidate: 1:20.2-1
     1:20.2-1 501
```

There seems to be no way to have exact version in `Build-Depends`. I'll double-check the documentation, requirements and regular expressions supported in `Build-Depends`, but it either stays this way, or we drop 16.04.

Makefile
dincho: I like the idea. I'm also not OK "gitting" here. What if `AE_DEB_VERSION` defaults to contents of `VERSION` file and can be overwritten by the caller ?

Also I don't like the variable name as it means "debian version" to me. `AE_DEB_PKG_VERSION` sounds better and more specific

valkov-ivaylo: I agree it is not the cleanest solution. It is targeted mainly at CI. The idea is that it should not be related to specific technology (CircleCI; variables etc).  One should be able to run `make prod-deb-package` and build a package on ~any shell. Imagine users checking out the code and trying to generate a package for themselves. 

The reverse logic will solve the master issue, but will require passing variables. I'm not a big fan of that. This will run in CI when a tag is created, and I am not aware how we will check for it and pass this variable then. I'll think about it tomorrow.

debian/control
dincho: We can't ship a package with OTP 21 either

debian/aeternity-node.service
valkov-ivaylo: From what I remember by experience, I think it will. I think I also tested it with our code-base. I'll test it again and confirm.

debian/control
valkov-ivaylo: Fixed in e737fba. Works as expected on 18.04 and 16.04 (with external repository)

debian/rules
```diff
@@ -0,0 +1,35 @@
...
+#CXXFLAGS:= $(shell dpkg-buildflags --get CXXFLAGS) $(CPPFLAGS)
+#CXXFLAGS += -Wno-error=format-security
+
+#CFLAGS := ""
+#CXXFLAGS := ""
+#CPPFLAGS := ""
+#DEB_BUILD_OPTIONS=noopt
+
+# Uncomment this to turn on verbose mode.
+export DH_VERBOSE=1
```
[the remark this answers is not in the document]
valkov-ivaylo: Fixed in 4a0461b

debian/rules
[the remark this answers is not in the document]
valkov-ivaylo: Fixed in b9fb06a

Makefile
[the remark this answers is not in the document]
valkov-ivaylo: Fixed in 332a55b

dincho:
The build process should not assume it's run with privileged user or have access to any FS paths than current dir and /tmp.

```
Error in tempfile() using template /aeternity-node_2.2.0.tar.gz.new.XXXXXX: Could not create temp file /aeternity-node_2.2.0.tar.gz.new.w28cQg: Permission denied at /usr/share/perl5/Dpkg/Source/Package/V3/Native.pm line 98.
```

Makefile
```diff
@@ -375,6 +411,13 @@ internal-clean: $$(KIND)
 internal-distclean: $$(KIND)
 	@rm -rf ./_build/$(KIND)
 
+prod-deb-package:
+	export DEBEMAIL=$(AE_DEB_EMAIL); \
```
dincho: Could you please silent the exports?

Example output (too noisy):
```
export DEBEMAIL="info@aeternity.com"; \
export DEBFULLNAME="Aeternity Team" ;\
dch -v `cat VERSION` "Release notes are available in /usr/share/doc/aeternity-node/docs/release-notes/RELEASE-NOTES-`cat VERSION`.md"; \
dch -r "Release notes are available in /usr/share/doc/aeternity-node/docs/release-notes/RELEASE-NOTES-`cat VERSION`.md"; \
debuild -uc -us

```

Makefile
dincho: I can confirm it works.

Makefile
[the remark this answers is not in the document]
valkov-ivaylo: Fixed in 332a55b

> The build process should not assume it's run with privileged user or have access to any FS paths than current dir and /tmp.
> 
> ```
> Error in tempfile() using template /aeternity-node_2.2.0.tar.gz.new.XXXXXX: Could not create temp file /aeternity-node_2.2.0.tar.gz.new.w28cQg: Permission denied at /usr/share/perl5/Dpkg/Source/Package/V3/Native.pm line 98.
> ```

Please specify the environment it was run in:
 - user
 - path 
 - VM or container

The Debian packaging system generates most files (*.deb, *tar.gz, *.buildinfo, *.build etc.) one level above the source directory it was run from. It generally does not require root permission. I assume the issue you are having is related to upper level paths/directories not having write permission for the user used.

Makefile
valkov-ivaylo: Just to clarify, you want me to suppress the lines from the example, not the output of the build?

Example:
```
	@export DEBEMAIL=$(AE_DEB_EMAIL); \
...
```

This will only the lines in question, not the build output.

valkov-ivaylo:
> Perm

valkov-ivaylo:
> The build process should not assume it's run with privileged user or have access to any FS paths than current dir and /tmp.
> 
> ```
> Error in tempfile() using template /aeternity-node_2.2.0.tar.gz.new.XXXXXX: Could not create temp file /aeternity-node_2.2.0.tar.gz.new.w28cQg: Permission denied at /usr/share/perl5/Dpkg/Source/Package/V3/Native.pm line 98.
> ```

Please specify the environment it was run in:
 - user
 - path 
 - VM or container

The Debian packaging system generates most files (*.deb, *tar.gz, *.buildinfo, *.build etc.) one level above the source directory it was run from. It generally does not require root permission. I assume the issue you are having is related to upper level paths/directories not having write permission for the user used.

Makefile
valkov-ivaylo: My proposition is to rework this to check if the version in the VERSION file has a related tag. A stable package should be created. If it doesn't, a pre-release one (the code is from a branch). 

Example:

```
 git tag | grep `cat VERSION`
```
```
ifneq (, $(shell  git tag | grep `cat VERSION`))
   AE_DEB_VERSION=`cat VERSION`"-1-git~"`date +%Y%m%d%H%M%S`
endif
```
Potentially a git hash could also be added, if it does not break Debian versioning scheme and pre-releases incremental state.

The standard situation is to produce a package for a stable version. The pre-releases and builds from master (or whatever) are a deviation. In my understanding the approach  should be to create a stable release and add work-arounds on top for the deviation. I am not in favor of adding work-around (passing variables) to produce a stable version package. 

If we are two to one for passing a variable for a stable build, I'll do it, but I do not think it is the right approach.

dincho:
```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   debian/changelog

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	debian/aeternity-node.debhelper.log

```
Unneeded artefacts

Makefile
dincho: I don't see what's the problem and misunderstanding here.

`AE_DEB_PKG_VERSION ` should default to `cat VERSION`, without any gits mits dates or whatever.

External scripts must handle more exotic versioning models if they needs to by passing the new version:
e.g.
`make deb-prod-package AE_DEB_PKG_VERSION=3.0.0-bla-bla-bla`

dincho: And more concrete the CI master job will run:

`
make deb-prod-package AE_DEB_PKG_VERSION=$(cat VERSION)-1-git-master~$(date +%Y%m%d%H%M%S)
`
or whatever the versioning scheme is

Makefile
dincho: The makefile commands "echo" for export commands.

dincho: And maybe also for `dch` commands

dincho:
Running `make prod-deb-package` multiple times produce invalid (non-sense?) changelog:

```
diff --git a/debian/changelog b/debian/changelog
index 4a90b8d9..b1c248bc 100644
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,24 @@
+aeternity-node (2.2.0) bionic; urgency=medium
+
+  * Release notes are available in /usr/share/doc/aeternity-
+    node/docs/release-notes/RELEASE-NOTES-2.2.0.md
+
+ -- Aeternity Team <info@aeternity.com>  Tue, 19 Mar 2019 09:24:18 +0000
+
+aeternity-node (2.2.0) bionic; urgency=medium
+
+  * Release notes are available in /usr/share/doc/aeternity-
+    node/docs/release-notes/RELEASE-NOTES-2.2.0.md
+
+ -- Aeternity Team <info@aeternity.com>  Mon, 18 Mar 2019 16:10:42 +0000
+
+aeternity-node (2.2.0) bionic; urgency=medium
+
+  * Release notes are available in /usr/share/doc/aeternity-
+    node/docs/release-notes/RELEASE-NOTES-2.2.0.md
+
+ -- Aeternity Team <info@aeternity.com>  Mon, 18 Mar 2019 16:01:35 +0000
+
 aeternity-node (1.1.0) unstable; urgency=medium
 
   * Initial Debianization
```

valkov-ivaylo:
> Running `make prod-deb-package` multiple times produce invalid (non-sense?) changelog:
> 

Addressed in 22c1d48

valkov-ivaylo:
> Unneeded artefacts

> Untracked files:
>   (use "git add <file>..." to include in what will be committed)
 

> 	debian/aeternity-node.debhelper.log

This one is product of the package building process. What exactly is the issue?

> 	modified:   debian/changelog

This one is no longer kept in Git after 22c1d48. Will have the same effect as above.

Makefile
valkov-ivaylo: Addressed in e4a9f0a. After 22c1d48 addressed there.

Makefile
valkov-ivaylo: > Also I don't like the variable name as it means "debian version" to me. `AE_DEB_PKG_VERSION` sounds better and more specific

Addressed in bce08c4

valkov-ivaylo: > I don't see what's the problem and misunderstanding here.
> 
> `AE_DEB_PKG_VERSION ` should default to `cat VERSION`, without any gits mits dates or whatever.
> 
> External scripts must handle more exotic versioning models if they needs to by passing the new version:

Fine. I'll do it the wrong way. This leaves room for error and the possibility to generate stable package version (no pre-release) out of master (or whatever branch). My desire was for it to be self-sustained.

valkov-ivaylo: Addressed in 93b75a5

Makefile
valkov-ivaylo: Addressed in 93b75a5

debian/aeternity-node.service
valkov-ivaylo: Investigations were made related to systemd and aeternity-node. What definitely is working is `service start`, `service stop`, `service restart`. Automatic restart on failure is not working. All possible variants for improvement were tested:
  * timeouts (systemd service file context)
 * running epmd as `ExecStartPre`
  * running epmd as the same user that runs aeternity-node; as a stand alone process
  * running epmd as a service (from erlang-base on Ubuntu 18.04)
  * running epmd as a service (custom service file)
  * disabling pre-start hooks (`--relx-disable-hooks` ; rule-out issues)

It all results in no automatic restart, or the node, not being able to start. All related to the epmd process. Sysytemd does not automatically restarts the service unless the epmd process is killed. If an aeternity sub-process is killed, only epmd is left in the state of the service and it is considered active. 

It should be working anyway compared with similar start scripts using Erlang. 

A separate issue will be open related to this matter.

dincho:
I still have issues building this in the 1804 builder container.

```
docker run -it -v ${PWD}:/home/builder/src -w /home/builder/src --user builder aeternity/builder:1804
```

```
make prod-deb-package
---SNIP---
dpkg-deb: building package 'aeternity-node' in '../aeternity-node_2.3.0_amd64.deb'.
dpkg-deb: building package 'aeternity-node-dbgsym' in 'debian/.debhelper/scratch-space/build-aeternity-node/aeternity-node-dbgsym_2.3.0_amd64.deb'.
dh_builddeb: mv debian/.debhelper/scratch-space/build-aeternity-node/aeternity-node-dbgsym_2.3.0_amd64.deb ../aeternity-node-dbgsym_2.3.0_amd64.ddeb: Invalid cross-device link
	Renaming aeternity-node-dbgsym_2.3.0_amd64.deb to aeternity-node-dbgsym_2.3.0_amd64.ddeb
dh_builddeb: Aborting due to earlier error
debian/rules:8: recipe for target 'binary' failed
make: *** [binary] Error 38
dpkg-buildpackage: error: fakeroot debian/rules binary subprocess returned exit status 2
debuild: fatal error at line 1152:
dpkg-buildpackage -rfakeroot -us -uc -ui failed
Makefile:415: recipe for target 'prod-deb-package' failed
make: *** [prod-deb-package] Error 29
builder@7096841eec70:~/src$ pwd
/home/builder/src
```

valkov-ivaylo:
> Invalid cross-device link

This is caused as a side-effect of the mapping. One level above the build directory is not the same filesystem. It seems that the issues is a cause of hard linking (ln).

```
docker run -it -v ${PWD}:/tmp/src -w /tmp/src --user builder aeternity/builder:1804
```

```
$ pwd
/tmp/src 
```
```
$ touch one
$ ln one ../two
ln: failed to create hard link '../two' => 'one': Invalid cross-device link
```

The work-around for this is to map one level above the source directory.

```
$ ls -lt
drwxrwxr-x 20 master master     4096 Mar 29 12:10 aeternity # git source
```
```
$ docker run -it -v ${PWD}:/home/builder/src/ -w /home/builder/src --user builder aeternity/builder:1804
```
```
builder@f09e7ad1f894:~/src$ cd aeternity
builder@f09e7ad1f894:~/src/aeternity$ make prod-deb-package
```

```
 dh_builddeb
dpkg-deb: building package 'aeternity-node' in '../aeternity-node_2.3.0_amd64.deb'.
dpkg-deb: building package 'aeternity-node-dbgsym' in 'debian/.debhelper/scratch-space/build-aeternity-node/aeternity-node-dbgsym_2.3.0_amd64.deb'.
	Renaming aeternity-node-dbgsym_2.3.0_amd64.deb to aeternity-node-dbgsym_2.3.0_amd64.ddeb
 dpkg-genbuildinfo

```

Makefile
```diff
@@ -391,6 +411,16 @@ internal-clean: $$(KIND)
 internal-distclean: $$(KIND)
 	@rm -rf ./_build/$(KIND)
 
+prod-deb-package:
```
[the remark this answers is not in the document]
valkov-ivaylo: Addressed in 0ca3e96

Makefile
```diff
@@ -391,6 +411,16 @@ internal-clean: $$(KIND)
...
+	@if [ ! -f debian/changelog ] ; \
+	then \
+		export DEBEMAIL=$(AE_DEB_EMAIL); \
+		export DEBFULLNAME=$(AE_DEB_NAME) ; \
+		dch --create --package=aeternity-node -v $(AE_DEB_PKG_VERSION) $(AE_DEB_DCH_REL_NOTE); \
```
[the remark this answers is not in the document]
valkov-ivaylo: Addressed in 7c475f3

dincho:
I got those artefacts after deb build:

```
	debian/.debhelper/
	debian/aeternity-node.debhelper.log
	debian/aeternity-node.substvars
	debian/aeternity-node/
	debian/changelog
	debian/files
```

e.g.:
```
debian/.debhelper/
debian/aeternity-node*
debian/changelog
debian/files
```
in .gitignore

valkov-ivaylo:
> I got those artefacts after deb build:
 
> in .gitignore

Addressed in 1b4d0be

Makefile
```diff
@@ -391,6 +411,16 @@ internal-clean: $$(KIND)
 internal-distclean: $$(KIND)
 	@rm -rf ./_build/$(KIND)
 
+prod-deb-package:
+	@if [ ! -f debian/changelog ] ; \
```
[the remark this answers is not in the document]
valkov-ivaylo: Addressed in 263c744

lucafavatella:
Conflicts on Makefile.

debian/aeternity-node.service
lucafavatella: Thanks for checking.

> A separate issue will be open related to this matter.

Could you please confirm ticket tracking this.

docs/debian_ubuntu_packaging.md
```diff
@@ -0,0 +1,410 @@
...
+the Debian changelog file policy. This is the approach used.
+
+### Using dch
+
+```
+cd ae_node_source/
+DEBEMAIL=info@aeternity.com
+DEBFULLNAME="Aeternity Team"
+AE_VERSION=$(cat VERSION)
+AE_VERSION_NOTE="Release notes are available in /usr/share/doc/aeternity-node/docs/release-notes/RELEASENOTES-$AE_VERSION.md"
```
lucafavatella:
```suggestion
AE_VERSION_NOTE="Release notes are available in /usr/share/doc/aeternity-node/docs/release-notes/RELEASE-NOTES-$AE_VERSION.md"
```

valkov-ivaylo:
> Conflicts on Makefile.

Fixed after rebased to master (2163712; [entire file in commit](https://github.com/aeternity/aeternity/blob/216371295fabc1ded47795c1db7f8d7faabd365e/Makefile))

docs/debian_ubuntu_packaging.md
valkov-ivaylo: Fixed in 5142342. The whole example is improved in c409e5e. In sync with the commands used in the main Makefile (`$(DEB_PKG_CHANGELOG_FILE):`). Required after changes in the logic for the changelog file generation.

debian/aeternity-node.service
valkov-ivaylo: This one. https://www.pivotaltracker.com/n/projects/2152594/stories/165141235

PR #2244, opened 2019-03-14
PT161642500: Fix decode-data error message

hanssv:


apps/aehttp/src/aehttp_logic.erl
```diff
@@ -228,7 +228,7 @@ contract_call(ABI, EncodedCode, Function, Argument) ->
 contract_decode_data(Type, Data) ->
     case aehttp_api_encoder:safe_decode(contract_bytearray, Data) of
         {error, _} ->
-            {error, <<"Data must be hex encoded">>};
+            {error, <<"Data must be a contract bytearray (\"cb_...\")">>};
```
lucafavatella: I proposed release note in #2250

PR #2245, opened 2019-03-14
We need to exit the contract run with the same context we entered it in

ThomasArts:
When joining different transaction instructions, the joined sequence of instructions should behave the same as the consecutive execution of the parts. However, QuickCheck found with txs_glue_eqc.erl that this is not the case.

This patch restores the context to what it was before running the contract. That means that when we run a contract inside a transaction, we're back in transaction context after running.


hanssv:
Rebased on top of working rebar3

lucafavatella:
I am taking the liberty of labelling as WIP as not planned for this release - so to ease filtering of open PRs.

PR #2224, opened 2019-03-06
Pt 164238122 sc protocol vsn -- system tests

lucafavatella:
This PR is:
* Master branch
* With #2204 merged
* With additional system tests covering state channels versions compatibility
* With Makefile target `smoke-test-run` and CI job step "Install deps for system smoke tests" amended for easing CI - **these need to be reverted**

PR #2261, opened 2019-03-21
Pt 164799015 sc backwards compatibility tests

velzevur:
Rebase #2224 and introduce compatibility between State Channel running on nodes with different versions

### DIFF
--- a/system_test/common/aest_channels_SUITE.erl
+++ b/system_test/common/aest_channels_SUITE.erl
@@ -13,10 +13,13 @@
 -export([
     test_simple_same_node_channel/1,
     test_simple_different_nodes_channel/1,
+    test_compat_with_initiator_node_using_minerva_initial_channel_version/1,
+    test_compat_with_responder_node_using_minerva_initial_channel_version/1,
     on_chain_channel/1
 ]).
 
 -import(aest_nodes, [
+    spec/3,
     setup_nodes/2,
     start_node/2,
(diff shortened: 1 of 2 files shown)


velzevur:
Backwards compatibility tested with node versions in tests:
* `local` and `v2.0.0`
* `local` and `latest`

This implies some relying on transitivity on backwards compatibility between all versions in between, so [3bbcfcd](https://github.com/aeternity/aeternity/pull/2261/commits/3bbcfcd70f783022837f79fd1a5ca7e8a12af2d0) checks `v2.1.0` and `v2.2.0` but it will not be ran as part of the system tests.

Tests do not try backwards compatibility with `roma` versions as those are not supported anymore.

system_test/common/aest_channels_SUITE.erl
```diff
@@ -225,3 +253,16 @@ on_chain_channel(Cfg) ->
 
     wait_for_value({balance, maps:get(pubkey, ?BOB), 100}, NodeNames, 5000, []),
     wait_for_value({balance, maps:get(pubkey, ?ALICE), 100}, NodeNames, 5000, []).
+
+node_base_spec_with_minerva_initial_channel_version() ->
+    #{source => {pull, "aeternity/aeternity:v2.0.0"}}.
```
lucafavatella: I see 2.0.0 had already the config at renamed location. Ok.

PR #2264, opened 2019-03-25
Drop next branch

dincho:


.circleci/config.yml
```diff
@@ -760,22 +752,6 @@ jobs:
...
 
-  deploy_next:
-    executor: infrastructure_container_stable
-    <<: *infrastructure_config
-    environment:
-      - DEPLOY_ENV: next
-      - DEPLOY_DOWNTIME: 900 #15M
-    steps: *deploy_steps
-
-  deploy_unstable:
```
lucafavatella: This could be kept if amended config.
See proposal at https://www.pivotaltracker.com/story/show/164861501/comments/200893373

.circleci/config.yml
```diff
@@ -1047,20 +1020,6 @@ workflows:
...
               only: env/dev2
 
-      - deploy_next:
-          requires:
-            - linux_package
-          filters:
-            branches:
-              only: *next_branch
-
-      - deploy_unstable:
```
lucafavatella: This could be kept. See other comment.

.circleci/config.yml
dincho: I prefer to update the config first in scope of the other ticket and then re-add this.

lucafavatella:
Some updates in infrastructure repo may be needed too.

PR #2265, opened 2019-03-25
Update build.md

nikita-fuchs:
add missing dependencies to MacOS build instruction and adjust formatting

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -5,8 +5,8 @@ The commands below assume you are logged in with `sudo` user.
 
 ## Dependencies installation
 
-## Ubuntu
-### Common tools and libraries
+### Ubuntu
+#### Common tools and libraries
 
 Make sure your Ubuntu version and it's packages are up to date, then install required tools and libraries:
 ```bash
@@ -15,17 +15,17 @@ sudo apt-get -qq update \
 && sudo apt-get -qq -y install git curl autoconf build-essential ncurses-dev libssl-dev
 ```
 
-### OTP install
+#### OTP install
 
 Required Erlang OTP version is `20.1`.
 
-#### Ubuntu 18.04
+##### Ubuntu 18.04
 
 ```bash
 sudo apt-get install erlang
 ```
 
-#### Ubuntu 16.04
+##### Ubuntu 16.04
 
 Ubuntu 16.04 ships with outdated erlang version. Version 20 can be installed from source:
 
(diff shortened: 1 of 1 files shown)


dincho:
This would generate the following ToC:
```
- [Build from source](#build-from-source)
  - [Dependencies installation](#dependencies-installation)
  - [Ubuntu](#ubuntu)
    - [Common tools and libraries](#common-tools-and-libraries)
    - [OTP install](#otp-install)
      - [Ubuntu 18.04](#ubuntu-1804)
      - [Ubuntu 16.04](#ubuntu-1604)
    - [Libsodium install](#libsodium-install)
      - [Ubuntu 18.04](#ubuntu-1804-1)
      - [Ubuntu 16.04](#ubuntu-1604-1)
  - [MacOS](#macos)
  - [Builds](#builds)
    - [Production build](#production-build)
    - [Production package](#production-package)
```

While the old one was:
```
- [Build from source](#build-from-source)
 - [Dependencies installation](#dependencies-installation)
  - [Common tools and libraries](#common-tools-and-libraries)
  - [OTP install](#otp-install)
   - [Ubuntu 18.04](#ubuntu-1804)
   - [Ubuntu 16.04](#ubuntu-1604)
  - [Libsodium install](#libsodium-install)
   - [Ubuntu 18.04](#ubuntu-1804-1)
   - [Ubuntu 16.04](#ubuntu-1604-1)
 - [Builds](#builds)
  - [Production build](#production-build)
  - [Production package](#production-package)
```

The new section should be H3 and all children of Ubuntu should be H+1

docs/build.md
```diff
@@ -69,6 +70,20 @@ curl -fsSL -o libsodium-src.tar.gz "$LIBSODIUM_DOWNLOAD_URL" \
...
+## MacOS
+
+Install `brew` if not done yet:
+```
+/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+```
+Also, install dependencies:
+```
+brew update
+brew install erlang
```
dincho: I think this will install 21.3.2 which is not supported by the node yet. @lucafavatella Do we support 21 for builds or we don't support it runtime only ?

nikita-fuchs: Building with 21 succeeded on my side.

dincho: Let's double check with @lucafavatella

lucafavatella: Our CI checks that [building with OTP 21 works for the purpose of running some static checks](https://github.com/aeternity/aeternity/blob/27f5df09a7bb70dc2a43c29106acc1b56ff1c303/.circleci/config.yml#L474). We do not support at runtime only e.g. re logging behaviour.

dincho: So we're OK to publish build docs for OTP 21 ? (default by brew install)

lucafavatella: I am considering about support for such documented instructions. Considering that our CI tests build on macOS using our pinned formula (see [CI config](https://github.com/aeternity/aeternity/blob/27f5df09a7bb70dc2a43c29106acc1b56ff1c303/.circleci/config.yml#L571)), may I suggest to replace this with something like:
```suggestion
brew install aeternity_source/deployment/homebrew/erlang.rb
```
Though this ^^^ may benefit from being better integrated in the document (as the cloning of the repo is documented below this).

PR #2276, opened 2019-03-27
PT-164805105 Re-check nonce of transactions in the mempool

zp-sd:
Introduce additional garbage collection of transactions, based on the
`{origin, nonce}` pair cache of transactions that were accepted into the chain.

### DIFF
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -73,6 +73,7 @@
          terminate/2, code_change/3]).
 
 -export_type([ dbs/0
+             , origins_cache/0
              , pool_db/0
              , pool_db_key/0
              , tx_hash/0]).
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -47,7 +47,8 @@
         ]).
 
 %% exports used by GC (should perhaps be in a common lib module)
--export([ origins_cache/0
+-export([ dbs_/0
+        , origins_cache/0
         , origins_cache_max_size/0
         , pool_db/0
         , pool_db_nonce/0
@@ -266,6 +267,10 @@ dbs() ->
 
 raw_delete(#dbs{} = Dbs, Key) ->
     pool_db_raw_delete(Dbs, Key).
+
+-spec dbs_() -> dbs().
+dbs_() ->
+    #dbs{}.
 
 -spec gc_db(dbs()) -> pool_db().
 gc_db(#dbs{gc_db = GcDb}) ->
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -16,6 +16,8 @@
 -define(MEMPOOL_VISITED, mempool_visited).
 -define(MEMPOOL_NONCE, mempool_nonce).
 -define(MEMPOOL_GC, mempool_gc).
+-define(ORIGINS_CACHE, origins_cache).
+-define(ORIGINS_CACHE_MAX_SIZE, 5000).
 -define(KEY(NegFee, NegGasPrice, Origin, Nonce, TxHash),
            {NegFee, NegGasPrice, Origin, Nonce, TxHash}).
 -define(VALUE_POS, 2).
@@ -60,6 +62,7 @@
 -export([peek_db/0]).
 -export([peek_visited/0]).
 -export([peek_nonces/0]).
+-export([origins_cache_gc/0]).
 -endif.
 
 %% gen_server callbacks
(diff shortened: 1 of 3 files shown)


apps/aecore/src/aec_tx_pool.erl
```diff
@@ -628,13 +661,110 @@ update_pool_on_tx_hash(TxHash, {#dbs{gc_db = GCDb} = Dbs, GCHeight}, Handled) ->
...
+                            true = ets:insert(OriginsCache, {Origin, Nonce}),
+                            ok;
+                        false -> ok
+                    end
+            end;
+        false -> ok
+    end.
+
+origins_cache_gc(#state{origins_cache = OriginsCache, dbs = Dbs}) ->
+    EntriesToProcess = 100,
```
hanssv: 100 is a number from a hat? Might be good to indicate that...

apps/aecore/src/aec_tx_pool.erl
```diff
@@ -628,13 +661,110 @@ update_pool_on_tx_hash(TxHash, {#dbs{gc_db = GCDb} = Dbs, GCHeight}, Handled) ->
...
+    case origins_cache_get(OriginsCache, EntriesToProcess) of
```
hanssv: Should there be some randomness in this selection?

apps/aecore/src/aec_tx_pool.erl
```diff
@@ -263,17 +274,28 @@ init([]) ->
...
+    ok = start_origins_cache_gc(),
     lager:debug("init: GCHeight = ~p", [GCHeight]),
     {ok, #state{gc_height = GCHeight}}.
 
+-ifdef(TEST).
+start_origins_cache_gc() ->
+    ok.
+-else.
+start_origins_cache_gc() ->
+    erlang:send_after(rand:uniform(10000), self(), origins_cache_gc),
```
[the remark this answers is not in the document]
gorillainduction: So this will run on average every five seconds. Seems often to me.

Perhaps have it run a bit more seldom, and also add an offset for a minimum wait?

EDIT: Ok, I see, this is only the initial call. It would still make sense to add an offset to make it not interfere with the start of the system.

apps/aecore/src/aec_tx_pool.erl
```diff
@@ -317,6 +342,10 @@ handle_cast_(Msg, State) ->
 handle_info(Msg, St) ->
     ?TC(handle_info_(Msg, St), Msg).
 
+handle_info_(origins_cache_gc, #state{} = State) ->
+    ok = origins_cache_gc(State),
+    erlang:send_after(30000, self(), origins_cache_gc),
```
gorillainduction: Same as above. Add an offset since it doesn't make sense to run this too often.

apps/aecore/src/aec_tx_pool.erl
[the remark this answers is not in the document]
zp-sd: Yes, delay sounds reasonable 👍

apps/aecore/src/aec_tx_pool.erl
zp-sd: This is to be executed every 30 seconds, which looked reasonable to me, but I can increase the delay... I don't have strong opinions here.

apps/aecore/src/aec_tx_pool.erl
zp-sd: Yes, I'll put it into the constant to tune it if needed.

zp-sd:
@gorillainduction In the first implementation I started I had a rationale to keep it in the the core mempool gen server (this was not queue-based cache which did not look good at some point). But with ETS-based solution it indeed seems rationable to move this GC to the `aec_tx_pool_gc` I think.

@hanssv thoughts?

apps/aecore/src/aec_tx_pool.erl
gorillainduction: Sorry, I read in too much from the previous code snippet. Disregard this.

hanssv:
I think it makes sense to have it in the GC context...

zp-sd:
System tests result:
```
%%% aest_channels_SUITE: .......
%%% aest_commands_SUITE: ..
%%% aest_compatibility_SUITE: .
%%% aest_db_SUITE: .....
%%% aest_genesis_SUITE: .
%%% aest_mempool_SUITE: ..
%%% aest_names_SUITE: ..
%%% aest_oracles_SUITE: .....
%%% aest_peers_SUITE: ..
%%% aest_perf_SUITE: ......
%%% aest_sync_SUITE: ......
%%% aest_sync_SUITE ==> abrupt_stop_new_node: SKIPPED
%%% aest_sync_SUITE ==> {tc_user_skip,database_restart_needs_fix}

%%% aest_sync_SUITE ==> abrupt_stop_mining_node: SKIPPED
%%% aest_sync_SUITE ==> {tc_user_skip,database_restart_needs_fix}

EXPERIMENTAL: Writing retry specification at /Users/zp/src/aeternity/aeternity/system_test/logs/retry.spec
              call rebar3 ct with '--retry' to re-run failing cases.
Skipped 2 (2, 0) tests. Passed 39 tests.
```

PR #2278, opened 2019-03-28
PT-164923024 Move varna out of aeternity into separate repo

rvirding:
Move varna out of aeternity into separate repo


rvirding:
I don't know why the test failed and what it means.

PR #2291, opened 2019-04-01
Pt 165017449 fix include libs

ThomasArts:
The include_lib in many files pointed to apps/.... which is not according to standard and makes a specific cover tool break.

Since we have circular dependencies, the best solution was to point with ../.. to other applications.


test/test.app.src
```diff
@@ -0,0 +1,16 @@
+{application, test,
+ [{description, "Ugly rebar3 hack to make include_lib work for Aethernity"},
```
lucafavatella:
```suggestion
 [{description, "Ugly rebar3 hack to make include_lib work for Aeternity"},
```

ThomasArts: Thanks :-).

PR #2297, opened 2019-04-02
PT-164805105 Add release notes for mempool Garbage Collection based o…

zp-sd:
…n {origin,nonce} cache


docs/release-notes/next/PT-164805105-new-mempool-GC.md
```diff
@@ -0,0 +1 @@
+* Introduces additional Garbage Collection of transactions, based on new transaction `{origin/account, nonce}` cache.
```
lucafavatella: I see follows #2276

PR #2298, opened 2019-04-02
Use latest aesophia and aebytecode. Use new representation of variants.

happi:


apps/aefate/src/aefa_fate_op.erl
```diff
@@ -210,11 +210,12 @@ switch(Arg0, Arg1, EngineState) ->
     N = length(Arg1),
     {Value, ES1} = get_op_arg(Arg0, EngineState),
     if ?IS_FATE_VARIANT(Value) ->
-            ?FATE_VARIANT(Size, Tag, _T) = Value,
-            if Size =:= N ->
+            ?FATE_VARIANT(Arities, Tag, _T) = Value,
+            if Tag < length(Arities)
+               , Tag < N ->
```
UlfNorell: Why not require `N == length(Arities)` as before (and as the other `switch`es)?

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -178,30 +178,30 @@ jumpif(Arg0, Arg1, EngineState) ->
...
     if ?IS_FATE_VARIANT(Value) ->
-            ?FATE_VARIANT(Size, Tag, _T) = Value,
-            if Size =:= 2 ->
+            ?FATE_VARIANT(Arities, Tag, _T) = Value,
+            if length(Arities) =:= 2 ->
                     case Tag of
                         0 -> {jump, Arg1, ES1};
                         1 -> {jump, Arg2, ES1};
-                        _ -> aefa_fate:abort({bad_variant_tag, Tag, Size}, ES1)
+                        _ -> aefa_fate:abort({bad_variant_tag, Tag}, ES1)
```
gorillainduction: This change does not seem to be handled in `aefa_fate:abort/2`

happi: Good catch, had it on my to do in the head, obviously not in a test case.

happi: This can not happen unless you manage to create variant data that is inconsistent in it self, which should not be possible. Ill come up with a fix.

lucafavatella:
[Test failure in CI](https://circleci.com/gh/aeternity/aeternity/18964#artifacts/containers/1) tracked as https://www.pivotaltracker.com/story/show/165082649

rebar.lock
```diff
@@ -22,7 +22,7 @@
   0},
  {<<"aesophia">>,
   {git,"https://github.com/aeternity/aesophia.git",
-      {ref,"8b7e4db4909810d23ebfebebaaaa4190d0fac197"}},
+      {ref,"3271d6fba49fd03499b6b93f18130ff822a9f8e3"}},
```
lucafavatella: https://github.com/aeternity/aesophia/compare/8b7e4db...3271d6f

rebar.lock
```diff
@@ -1,7 +1,7 @@
 {"1.1.0",
 [{<<"aebytecode">>,
   {git,"https://github.com/aeternity/aebytecode.git",
-      {ref,"34ae94e3e74fbd2ff36f7e5d601b9fd8318a9832"}},
+      {ref,"2d599df0ea059df597a49e140e7100b0e8585cd9"}},
```
lucafavatella: https://github.com/aeternity/aebytecode/compare/34ae94e...2d599d

rebar.lock
lucafavatella: This includes https://github.com/aeternity/aebytecode/compare/34ae94e...2d599d that is consistent.

lucafavatella: This cleans up blake2b lib used that is ok.

lucafavatella: This seems reasonably large.

PR #2300, opened 2019-04-02
Pt 165013950 hook in fate in contracts

gorillainduction:
Lots of small hacks to work around that FATE isn't done yet.

https://www.pivotaltracker.com/story/show/165013950


gorillainduction:
Labelled as WIP since this should not go into release, but the PR is ready for review.

PR #2304, opened 2019-04-03
Support slash in branch name/164427899

michalbagrowski:
pt: https://www.pivotaltracker.com/story/show/164427899

@dincho is this acceptable solution? replace / with _


dincho:
> @dincho is this acceptable solution? replace / with _

yes, I can't think of anything better

PR #2305, opened 2019-04-03
Support building on macOS with libs in custom locations

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/165087263

This PR reproduces the failure. It does not address it yet.

If we want to support advanced users building the software on macOS - either Homebrew installing in a non-default path or (probably) nix-shell - we need to consider environment variables already set by the users, that are at least CFLAGS and LDFLAGS. If we do not, then we need to make it clear that we do not support building the software in such configuration.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1027,6 +1027,8 @@ workflows:
             tags:
               only: *tag_regex
 
+      - osx_package_with_libs_flags: {}
+
       - deploy_integration:
           requires:
             - test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -261,6 +261,35 @@ references:
         fi
       when: on_fail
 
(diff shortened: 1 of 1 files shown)


tolbrino:
The problem really is in `lz4` not modifying the `CFLAGS` and `LDFLAGS` one they are set (even if an empty string). This can be easily fixed. Better though would be the upgrade to a more recent version using `port_compiler` which handles this case properly.

lucafavatella:
> Better though would be the upgrade to a more recent version using `port_compiler` which handles this case properly.

@tolbrino That fails on Windows https://github.com/aeternity/aeternity/pull/2292#issuecomment-478687612

tolbrino:
I'll look into the issue on Windows so we can switch to the PC version.

PR #2310, opened 2019-04-05
SC small fixes

velzevur:
No PT

### DIFF
--- a/apps/aechannel/src/aesc_channels.erl
+++ b/apps/aechannel/src/aesc_channels.erl
@@ -290,7 +290,7 @@ new(InitiatorPubKey, InitiatorAmount, ResponderPubKey, ResponderAmount, InitAcco
              state_hash           = StateHash,
              round                = Round,
              solo_round           = 0,
-             locked_until         = 0, % zero means "not closing" as well
+             locked_until         = 0, % zero means "active" as well
              lock_period          = LockPeriod,
              version              = Version}.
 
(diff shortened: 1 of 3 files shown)


apps/aechannel/src/aesc_create_tx.erl
```diff
@@ -152,7 +152,8 @@ process(#channel_create_tx{} = Tx, Trees, Env) ->
           state_hash(Tx),
           lock_period(Tx),
           fee(Tx),
-          nonce(Tx)),
+          nonce(Tx),
+          round(Tx)),
```
lucafavatella: I see this is [hardcoded to 1](https://github.com/aeternity/aeternity/blob/8f04de354e9a85141636cee68e2c9415803ec1d3/apps/aechannel/src/aesc_create_tx.erl#L308). Is the plan to include the round in the transaction? Because this syntax suggests that the round is a variable in the transaction serialized by the user though it is not.

velzevur: No, there is no plan on doing it. The `round/1` callback is needed so we don't have special processing for the `channel_create_tx` in the FSM.
This change is just for consisteny.

apps/aechannel/src/aesc_channels.erl
```diff
@@ -233,9 +235,9 @@ new(InitiatorPubKey, InitiatorAmount, ResponderPubKey, ResponderAmount,
              channel_reserve      = ReserveAmount,
              delegate_ids         = [aeser_id:create(account, D) || D <- DelegatePubkeys],
              state_hash           = StateHash,
-             round                = 1,
+             round                = Round,
              solo_round           = 0,
-             locked_until         = 0,
+             locked_until         = 0, % zero means "not closing" as well
```
ThomasArts: zero means "active" as well

velzevur: Addressed

PR #2311, opened 2019-04-05
Test for namespaces

UlfNorell:


apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -3895,6 +3897,17 @@ sophia_heap_to_heap_bug(_Cfg) ->
 
     ok.
 
+sophia_namespaces(_Cfg) ->
+    ?skipRest(vm_version() < ?VM_AEVM_SOPHIA_2, namespaces_not_in_roma),
```
lucafavatella: Is this support only in recent versions documented in https://github.com/aeternity/protocol/blob/7fc569fb8e93c8dcb4d596d7b171095475d06ca6/contracts/sophia.md#namespaces - or anywhere else? Should it?

PR #2320, opened 2019-04-10
Add win32 support to python release test script

tolbrino:
This PR add support for running the python release tests on Windows. Differences to running on Linux/macOS are:

- Use of `.zip` file instead of `.tar.gz`
- Use of Python3 instead of Python2, because a Python2 setup wasn't possible
- Use of OS package manager for some Python dependencies, while the rest is installed via pip outside of any virtualenv.
- Simulating user behaviour on Windows by using node `console` instead of `start`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ ST_CT_FLAGS = --logdir system_test/logs
 ST_CT_DIR = --dir system_test/common
 ST_CT_LOCALDIR = --dir system_test/only_local
 
-SWAGGER_CODEGEN_CLI_V = 2.3.1
+SWAGGER_CODEGEN_CLI_V = 2.4.4
 SWAGGER_CODEGEN_CLI = swagger/swagger-codegen-cli-$(SWAGGER_CODEGEN_CLI_V).jar
 SWAGGER_CODEGEN = java -jar $(SWAGGER_CODEGEN_CLI)
 SWAGGER_ENDPOINTS_SPEC = apps/aeutils/src/endpoints.erl
@@ -306,7 +306,7 @@ python-single-uat: swagger
 	( cd $(PYTHON_DIR) && TEST_NAME=$(TEST_NAME) $(MAKE) single-uat; )
 
 python-release-test: swagger
-	( cd $(PYTHON_DIR) && WORKDIR="$(WORKDIR)" TARBALL=$(TARBALL) VER=$(VER) $(MAKE) release-test; )
+	( cd $(PYTHON_DIR) && WORKDIR="$(WORKDIR)" PACKAGE=$(PACKAGE) VER=$(VER) $(MAKE) release-test; )
 
 python-package-win32-test:
 	( cd $(PYTHON_DIR) && WORKDIR="$(WORKDIR)" PACKAGESPECFILE=$(PACKAGE_SPEC_WIN32) $(MAKE) package-win32-test; )
--- a/apps/aehttp/priv/swagger.json
+++ b/apps/aehttp/priv/swagger.json
@@ -106,7 +106,13 @@
           "200" : {
             "description" : "Successful operation",
             "schema" : {
-              "$ref" : "#/definitions/inline_response_200"
+              "type" : "object",
+              "properties" : {
+                "hash" : {
+                  "type" : "string",
+                  "description" : "Hash"
+                }
+              }
             }
           },
           "404" : {
@@ -129,7 +135,14 @@
           "200" : {
             "description" : "Successful operation",
             "schema" : {
-              "$ref" : "#/definitions/inline_response_200_1"
+              "type" : "object",
+              "properties" : {
+                "height" : {
+                  "type" : "integer",
+                  "format" : "int64",
+                  "description" : "Height"
+                }
+              }
             }
           },
           "404" : {
@@ -393,7 +406,15 @@
           "200" : {
             "description" : "Successful operation",
             "schema" : {
-              "$ref" : "#/definitions/inline_response_200_2"
+              "type" : "object",
+              "properties" : {
+                "count" : {
+                  "type" : "integer",
+                  "format" : "int64",
+                  "description" : "Count",
+                  "minimum" : 1
+                }
+              }
             }
           },
           "400" : {
@@ -1102,7 +1123,12 @@
           "200" : {
             "description" : "Successful operation",
             "schema" : {
-              "$ref" : "#/definitions/inline_response_200_3"
+              "type" : "object",
+              "properties" : {
+                "pubkey" : {
+                  "type" : "string"
+                }
+              }
             }
           }
         }
@@ -2212,7 +2238,9 @@
       "items" : {
         "type" : "integer",
         "format" : "int32"
-      }
+      },
+      "minItems" : 42,
+      "maxItems" : 42
     },
     "Uri" : {
       "type" : "string"
@@ -2267,15 +2295,10 @@
         }
       },
       "example" : {
-        "prev_hash" : null,
         "nonce" : 1,
         "version" : 5,
-        "miner" : null,
         "target" : 6,
-        "beneficiary" : null,
-        "state_hash" : null,
         "pow" : "",
-        "prev_key_hash" : null,
         "time" : 5,
         "hash" : { },
         "height" : 0,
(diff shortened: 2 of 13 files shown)


Makefile
```diff
@@ -13,7 +13,7 @@ ST_CT_FLAGS = --logdir system_test/logs
 ST_CT_DIR = --dir system_test/common
 ST_CT_LOCALDIR = --dir system_test/only_local
 
-SWAGGER_CODEGEN_CLI_V = 2.3.1
+SWAGGER_CODEGEN_CLI_V = 2.4.4
```
lucafavatella: I do not expect changes in code included in node but only in Python HTTP client. Ok.

scripts/windows/msys2_prepare.bat
```diff
@@ -13,13 +16,16 @@ rem Set required vars defaults
 IF "%ERTS_VERSION%"=="" SET "ERTS_VERSION=9.3"
 IF "%OTP_VERSION%"=="" SET "OTP_VERSION=20.3"
 IF "%WIN_OTP_PATH%"=="" SET "WIN_OTP_PATH=C:\Program Files\erl"
+IF "%WIN_JDK_BASEPATH%"=="" SET "WIN_JDK_BASEPATH=C:\Program Files\Java"
```
lucafavatella: Why is JDK needed? This is run by user - see docs/build-windows.md
If not needed by user, it looks like this shall be separate script for CI only?

lucafavatella: Shall these new params documented for user? See docs/build-windows.md

scripts/windows/msys2_prepare.bat
```diff
@@ -30,18 +36,15 @@ make ^
...
 mingw-w64-x86_64-binutils ^
 mingw-w64-x86_64-gcc ^
 mingw-w64-x86_64-libsodium ^
 mingw-w64-x86_64-nsis ^
 mingw-w64-x86_64-ntldd-git  ^
-mingw-w64-x86_64-python3 ^
-mingw-w64-x86_64-python3-pip ^
 mingw-w64-x86_64-yasm ^
 patch ^
+unzip ^
```
lucafavatella: Needed only for CI? Or also for user?

py/tests/release.py
```diff
@@ -180,33 +174,77 @@ def node_is_online(api):
...
+    assert os.path.isfile(binary), "Not a file: " + binary
+    assert os.access(binary, os.X_OK), "Can't access file: " + binary
+    print("Stopping " + binary)
+    os.chdir(temp_dir)
+    if sys.platform == "win32":
+        # We need to find the PID listening on the right port
+        listen_address = SETUP[node]["listen_address"]
+        result_lines = subprocess.check_output("netstat -ano -p TCP").decode("utf-8").split("\r\n")
+        pid = functools.reduce(lambda acc, r: find_pid(r, listen_address, acc), result_lines, "")
+        os.kill(int(pid), signal.SIGTERM)
```
lucafavatella: So this ^^^ is the way the user is meant to stop node on Window? 🤔

py/tests/release.py
```diff
@@ -180,33 +174,77 @@ def node_is_online(api):
...
+    assert os.path.isfile(binary), "Not a file: " + binary
+    assert os.access(binary, os.X_OK), "Can't access file: " + binary
     print("Starting " + binary)
     os.chdir(temp_dir)
-    os.system('ERL_FLAGS="-config `pwd`/p.config" bin/aeternity start')
+    erl_flags = " ".join(SETUP[node]["config"])
+    if sys.platform == "win32":
+        sub_env = os.environ
+        sub_env["ERL_FLAGS"] = erl_flags
+        subprocess.Popen([binary, "console"], env=sub_env)
```
lucafavatella: Is there something closer to how user would pass flags?

py/tests/release.py
tolbrino: Particularly on Windows the user works via the `werl.exe` binary with the node. This emulates a terminal as a proper window. The user can just close that window to kill the node. However, to emulate that the release test has to jump through the hoops and work via the PID.

py/tests/release.py
tolbrino: Using `ERL_FLAGS` is the way any user would do that. The difference to the previous approach is that one can't pass multiple configuration file arguments on Windows, which is why I had to refactor it.

scripts/windows/msys2_prepare.bat
tolbrino: Yes should be.

tolbrino: The files under `scripts/windows` are meant for users wanting to build the node. Therefore, the JDK is required to generate the swagger test client. So far there wasn't a differentiation between such a user and a developer using Windows.

scripts/windows/msys2_prepare.bat
tolbrino: Needed for the release testing.

apps/aehttp/priv/swagger.json
```diff
@@ -106,7 +106,13 @@
           "200" : {
             "description" : "Successful operation",
             "schema" : {
-              "$ref" : "#/definitions/inline_response_200"
```
lucafavatella: No refs. So schema shall be more explicit. Ok.

apps/aehttp/priv/swagger.json
```diff
@@ -2212,7 +2238,9 @@
       "items" : {
         "type" : "integer",
         "format" : "int32"
-      }
+      },
+      "minItems" : 42,
```
lucafavatella: More precise. Ok.

apps/aehttp/priv/swagger.json
```diff
@@ -2267,15 +2295,10 @@
         }
       },
       "example" : {
-        "prev_hash" : null,
```
lucafavatella: Less noise. Ok.

py/requirements.txt
```diff
@@ -1,3 +1,10 @@
+##
+## When updating these dependencies you must ensure that the relevant entries in the following files are kept up-to-date.
+## as well. This is required due to the more complicated python setup on Windows/msys2. Look for the tag WINDOWS_PYTHON.
+##
+## - scripts/windows/msys2_prepare.py
+## - py/requirements-win32.txt
+##
```
lucafavatella: ❤️

appveyor.yml
```diff
@@ -25,12 +25,12 @@ environment:
       OTP_VERSION: 20.1
       ERTS_VERSION: 9.1
       BUILD_STEP: build
-      TEST_STEP: ct
+      TEST_STEPS: ct,release
```
lucafavatella: I am a bit surprised that these are note rendered as 4 jobs:
* OTP 20.1 ct - artefacts being CT logs and node logs
* OTP 20.3 ct - artefacts being CT logs and node logs
* OTP 20.1 release - artefacts packages, only if smoke test passes
* OTP 20.3 release - artefacts packages, only if smoke test passes

If too slow, the 20.1 jobs may be deleted.

If too complex, they can be kept as is.

tolbrino: On AppVeyor CI it is not possible to share build artifacts between jobs. Therefore, splitting the test steps into 2 job would result in an additional build/package cycle which adds quite some time. That is why I condensed it into a single job.

scripts/windows/msys2_prepare.bat
```diff
@@ -93,6 +86,13 @@ rem Remove breaking tools
 
 %BASH_BIN% -lc "%PACMAN_RM% %PACMAN_PACKAGES_REMOVE% || true"
 
+@echo Current time: %time%
+rem Upgrade the MSYS2 platform
+
+%BASH_BIN% -lc "%PACMAN% -y pacman"
```
lucafavatella: Is this installing `pacman`?
If so, how can `packman` be present earlier?

lucafavatella:
I understand WIP so I took the liberty to label it as so.

tolbrino:
@lucafavatella @dincho This PR is finished, ready for final review.

dincho:
@noandrea could you please confirm that there is no BC breaking changes in swagger schemas ?

tolbrino:
> @tolbrino Could you please confirm that if smoke test fails then CI marks job as failed.

Here is an example of the test failing and the CI marking it as such: https://ci.appveyor.com/project/aeternity/aeternity/builds/23804790/job/aj79oj67doq6wkqf

PR #2314, opened 2019-04-09
Generalized Accounts

hanssv:
This is still WIP, but the change is rather big so better give time for people to have a good look at it...

What is not done:

- Crypto primitivies are missing and Sophia additions are not done https://www.pivotaltracker.com/story/show/164629541
- HTTP API for generalized accounts and its transactions https://www.pivotaltracker.com/story/show/165208165
- Testing state channels with two GAs
- Testing Fee distribution to miners (authorization cost!)
- HTTP tests are not yet committed they exist on my machine ;-) 

WIP documentation: https://github.com/aeternity/protocol/pull/336

PR #2321, opened 2019-04-10
Heap and ABI functions moved aesophia -> aebytcode

hanssv:
Need latest aesophia in #2314 - so this does the vital parts pointing aeternity to aeb_xxx instead och aeso_xxx...


rvirding:
This should finish [PT-164597852](https://www.pivotaltracker.com/story/show/164597852)

rebar.config
```diff
@@ -60,10 +60,10 @@
                    {ref, "9ff46af"}}},
 
         {aebytecode, {git, "https://github.com/aeternity/aebytecode.git",
-                     {ref,"2d599d"}}},
+                     {ref,"9041423"}}},
```
ThomasArts: That's not latest master, but I will bump soon as well. You now did the majority of the work already

hanssv: It is exactly the version that the referenced `aesophia` points to, that is correct I think?!

hanssv:
> This should finish [PT-164597852](https://www.pivotaltracker.com/story/show/164597852)

I'm not familiar with the scope of the ticket, but unless it doesn't also deal with making `aesophia` a test-only dependency this might be it?!

PR #2322, opened 2019-04-11
Pt 165246396 remove dependency

ThomasArts:
Finishes PT-165246396 and removes a call to aeb_fate_data that is inconsistent with other calls


apps/aefate/test/aefate_engine_test.erl
```diff
@@ -700,3 +700,53 @@ contracts() ->
        }.
 
 
+%% Encode is a convinience function for testing, encoding an Erlang term
```
gorillainduction: Can you please put this in a test util file, so it is available for other modules as well?

ThomasArts: done and rebased

PR #2323, opened 2019-04-11
Subtract gas in eval dispatch.

happi:
Use some gas for each instruction.
Note: Not tuned in any way yet.

### DIFF
--- a/apps/aecontract/test/aecontract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_SUITE.erl
@@ -642,13 +642,7 @@ create_contract_(ContractCreateTxGasPrice) ->
     %% Check that the created init call has the correct details not from the contract create tx
     ?assertEqual(ContractKey, aect_call:contract_pubkey(InitCall)),
     _ = aect_call:height(InitCall), %% Unclear if this needed.
-    case sophia_version() of
-        ?SOPHIA_FORTUNA_FATE ->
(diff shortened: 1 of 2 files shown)


apps/aefate/src/aefa_fate_eval.erl
```diff
@@ -5,345 +5,457 @@
 -module(aefa_fate_eval).
 -export([eval/2]).
 
-eval('RETURN', EngineState) ->
-    aefa_fate_op:return(EngineState);
+eval('RETURN',  #{ gas := Gas } = EngineState) ->
+    EngineState1 = EngineState#{ gas => Gas - 2},
```
ThomasArts: Would it not be nicer and shorter to write:
`aefa_fate_op:return( EngineState#{ gas => Gas - 2}).`
Saves a few lines here and there.

or a little wrapper decrease_gas(2, EngineState).
A matter of taste, of course. Advantage for such a wrapper is that you can check in  the wrapper that gas does not get negative... but now you check in return, which is fine as well.

happi: Fixed

happi:
There are actually Fate tests that checks the gas now. Fix needed.

lucafavatella:
A past failure in CI in this ticket was independent and [already tracked](https://www.pivotaltracker.com/story/show/165122907/comments/201657052).

PR #2325, opened 2019-04-11
Add decode function and turn on fate test run per default

gorillainduction:
https://www.pivotaltracker.com/story/show/165298157


apps/aefate/test/aefate_test_utils.erl
```diff
@@ -6,7 +6,11 @@
 
 -module(aefate_test_utils).
 
--export([encode/1]).
+-export([encode/1,
+         decode/1
```
ThomasArts: Yes, symmetry :-).

PR #2328, opened 2019-04-12
PT-164804897 GC of aec_signed_tx objects

### LINKED ISSUE #3030: Garbage collect aec_signed_tx objects
Full `aec_signed_tx` objects seem to stay in the database even after transactions are garbage collected. To track what a node already seen and avoid transactions' circulation keep `aec_gc_tx` with just a hash of transactions. This should reduce some DB size.

Consider garbage collecting `aec_gc_tx` objects as well (to be defined).

This issue is a copy of [PT-2124891](https://www.pivotaltracker.com/n/projects/2124891).

zp-sd:
Closes #3030

### DIFF
--- a/apps/aecore/include/aec_db.hrl
+++ b/apps/aecore/include/aec_db.hrl
@@ -17,4 +17,4 @@
 -record(aec_tx_location        , {key :: SignedTxHash :: aec_hash:hash(), value :: BlockHash :: aec_hash:hash()}).
 -record(aec_tx_pool            , {key :: SignedTxHash :: aec_hash:hash(), value :: PlaceHolder :: []}).
 -record(aec_discovered_pof     , {key, value}).
--record(aec_tx_gc              , {key, value}).
+-record(aec_tx_gc              , {key :: SignedTxHash :: aec_hash:hash(), value :: PlaceHolder :: 0}).
--- a/apps/aecore/include/aec_db.hrl
+++ b/apps/aecore/include/aec_db.hrl
@@ -17,4 +17,4 @@
 -record(aec_tx_location        , {key :: SignedTxHash :: aec_hash:hash(), value :: BlockHash :: aec_hash:hash()}).
 -record(aec_tx_pool            , {key :: SignedTxHash :: aec_hash:hash(), value :: PlaceHolder :: []}).
 -record(aec_discovered_pof     , {key, value}).
-
+-record(aec_tx_gc              , {key, value}).
--- a/apps/aecore/include/aec_db.hrl
+++ b/apps/aecore/include/aec_db.hrl
@@ -17,4 +17,4 @@
 -record(aec_tx_location        , {key :: SignedTxHash :: aec_hash:hash(), value :: BlockHash :: aec_hash:hash()}).
 -record(aec_tx_pool            , {key :: SignedTxHash :: aec_hash:hash(), value :: PlaceHolder :: []}).
 -record(aec_discovered_pof     , {key, value}).
--record(aec_tx_gc              , {key :: SignedTxHash :: aec_hash:hash(), value :: PlaceHolder :: 0}).
+-record(aec_tx_gc              , {key :: SignedTxHash :: aec_hash:hash(), value :: PlaceHolder :: []}).
--- a/apps/aecore/src/aec_db.erl
+++ b/apps/aecore/src/aec_db.erl
@@ -59,12 +59,16 @@
         , remove_tx_from_mempool/1
         , remove_tx_location/1
         , gc_tx/1
+        , remove_tx_gc/1
         ]).
 
 %% Only to be used from aec_tx_pool:init/1
 -export([ fold_mempool/2
         ]).
 
+%% Only to be used from aec_tx_gc:init/1
+-export([ foreach_tx_gc/1
+        ]).
 
 %% MP trees backend
 -export([ find_accounts_node/1
@@ -144,6 +148,7 @@ tables(Mode) ->
    , ?TAB(aec_tx_location)
    , ?TAB(aec_tx_pool)
    , ?TAB(aec_discovered_pof)
+   , ?TAB(aec_tx_gc)
     ].
 
 tab(Mode0, Record, Attributes, Extra) ->
@@ -280,6 +285,7 @@ write_block(Block, Hash) ->
                    TxHashes = [begin
                                    STxHash = aetx_sign:hash(STx),
                                    write_signed_tx(STxHash, STx),
+                                   delete(aec_tx_gc, STxHash), %% In case tx from a block was already GCed.
                                    STxHash
                                end
                                || STx <- Txs],
@@ -558,12 +564,17 @@ gc_tx(TxHash) ->
            BlockHash when is_binary(BlockHash) ->
                {error, BlockHash};
            mempool ->
-               delete(aec_tx_pool, TxHash);
+               delete(aec_tx_pool, TxHash),
+               delete(aec_signed_tx, TxHash),
+               write(aec_tx_gc, #aec_tx_gc{key = TxHash, value = 0});
            none ->
                ok;
            not_found ->
                {error, tx_not_found}
        end).
+
+remove_tx_gc(TxHash) ->
+    ?t(delete(aec_tx_gc, TxHash)).
 
 get_signed_tx(Hash) ->
     [#aec_signed_tx{value = DBSTx}] = ?t(read(aec_signed_tx, Hash)),
@@ -588,8 +599,15 @@ find_tx_location(STxHash) ->
            [] ->
                case mnesia:read(aec_tx_pool, STxHash) of
                    [] -> case mnesia:read(aec_signed_tx, STxHash) of
-                             [] -> not_found;
-                             [_] -> none
+                             [] ->
+                                 case mnesia:read(aec_tx_gc, STxHash) of
+                                     []  -> not_found;
+                                     [_] -> none
+                                 end;
+                             [_] ->
+                                 %% For backwards compatibility: before aec_tx_gc was introduced,
+                                 %% aec_signed_tx objects were stored forever, even after GC.
+                                 none
                          end;
                    [_] -> mempool
                end;
@@ -641,6 +659,15 @@ fold_mempool(FunIn, InitAcc) ->
                   FunIn(Hash, Acc)
           end,
     ?t(mnesia:foldl(Fun, InitAcc, aec_tx_pool)).
+
+foreach_tx_gc(Fun) ->
+    foreach_tx_gc(mnesia:first(aec_tx_gc), Fun).
+
+foreach_tx_gc('$end_of_table', _Fun) ->
+    ok;
+foreach_tx_gc(Key, Fun) ->
+    Fun(Key),
+    foreach_tx_gc(mnesia:next(aec_tx_gc, Key), Fun).
 
 %% start phase hook to load the database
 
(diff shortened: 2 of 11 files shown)


apps/aecore/include/aec_db.hrl
```diff
@@ -17,4 +17,4 @@
 -record(aec_tx_location        , {key, value}).
 -record(aec_tx_pool            , {key, value}).
 -record(aec_discovered_pof     , {key, value}).
-
+-record(aec_tx_gc              , {key, value}).
```
lucafavatella: Do I understand correctly this means that DB generated by new node cannot be used for old node? If so, could you please update https://github.com/aeternity/aeternity/blame/49ffb23ada36910dcda5bd87201cdf4072dcda7f/docs/release-notes/RELEASE-NOTES-2.4.0.md#L11

apps/aecore/src/aec_db.erl
```diff
@@ -558,13 +564,18 @@ gc_tx(TxHash) ->
            BlockHash when is_binary(BlockHash) ->
                {error, BlockHash};
            mempool ->
-               delete(aec_tx_pool, TxHash);
+               delete(aec_tx_pool, TxHash),
+               delete(aec_signed_tx, TxHash),
```
lucafavatella: I see this is deleting actual item. Ok.

apps/aecore/src/aec_db.erl
```diff
@@ -558,13 +564,18 @@ gc_tx(TxHash) ->
...
+               write(aec_tx_gc, #aec_tx_gc{key = TxHash, value = 0});
```
lucafavatella: Is this 0 a placeholder? May I suggest using a macro for readability?

See also https://github.com/aeternity/aeternity/pull/2329/files

apps/aecore/include/aec_db.hrl
lucafavatella:
```suggestion
-record(aec_tx_gc              , {key :: SignedTxHash :: aec_hash:hash(), value :: PlaceHolder :: 0}).
```

apps/aecore/src/aec_db.erl
```diff
@@ -588,8 +599,15 @@ find_tx_location(STxHash) ->
            [] ->
                case mnesia:read(aec_tx_pool, STxHash) of
                    [] -> case mnesia:read(aec_signed_tx, STxHash) of
-                             [] -> not_found;
-                             [_] -> none
+                             [] ->
+                                 case mnesia:read(aec_tx_gc, STxHash) of
+                                     []  -> not_found;
+                                     [_] -> none
```
lucafavatella: This leads to the doubt of how can actor (human or automated) calling an API for pushing to mempool distinguish case tx inserted and case tx not pushed i.e. [this clause](https://github.com/aeternity/aeternity/blob/e9bd96779c585eab037727b008babce242dab2da/apps/aecore/src/aec_tx_pool.erl#L172). Though this is not introduced by this change so ok.

This though leads to doubt of whether there is any actor relying on pool push returning ok, and retrieval of tx succeeding.

apps/aecore/src/aec_db.erl
```diff
@@ -144,6 +148,7 @@ tables(Mode) ->
    , ?TAB(aec_tx_location)
    , ?TAB(aec_tx_pool)
    , ?TAB(aec_discovered_pof)
+   , ?TAB(aec_tx_gc)
```
lucafavatella: Call to `aec_db:add_tx/1` writes to table `aec_signed_tx` without deleting from `aec_tx_gc`. What prevents mempool push, namely [this call](https://github.com/aeternity/aeternity/blob/e9bd96779c585eab037727b008babce242dab2da/apps/aecore/src/aec_tx_pool.erl#L717) (please consider only atomic blocks), to write `aec_signed_tx` entry shortly after `aec_db:gc_tx/1`? This would lead to tx having entry in `aec_signed_tx` (and `aec_tx_pool`) and `aec_tx_gc`. (I may be missing something.) Is this expected? What is user (or state channel FSM?) experience in this case (e.g. event published anywhere for tx included in mempool)? And how does DB/mempool/GC behave in this case?

zp-sd: `aec_db:add_tx/1` can be reach only after https://github.com/aeternity/aeternity/blob/master/apps/aecore/src/aec_tx_pool.erl#L169 returns `ok`. This is not transactional, but returning `ok` there means that there is no either `aec_tx_gc` or `aec_signed_tx` object for a transaction in the system. This means that there is no transaction in the system to be GCed (which prevents transaction from being GCed shortly).

Even is there is a path (as of now - I don't see any) which leads to having both `aec_signed_tx` and `aec_tx_gc` objects in the system - there will be no harm. New GC of transactions, namely `aec_tx_gc` server, removes only `aec_tx_gc` objects from the DB, without any impact on `aec_signed_tx`.

apps/aecore/src/aec_db.erl
zp-sd: Will do. I will also use `[]` as a placeholder, as it is not for `aec_tx_pool` objects.

apps/aecore/include/aec_db.hrl
zp-sd: As discussed offline, this is to be addressed.

lucafavatella: For clarity - new node needs to be able to start from DB of old node.

Re old node being able to start from new DB - not necessary provided documentation (release notes main doc) is fine tuned in order not to be misleading.

lucafavatella:
Could you please add system test in `aest_db_SUITE` that creates DB using latest stable release (`:latest` Docker tag) then tries to start with current codebase (`:local` Docker tag)? Of course such a test may need to be disables in the iteration approaching a major release.

apps/aecore/src/aec_db.erl
lucafavatella: > This though leads to doubt of whether there is any actor relying on pool push returning ok, and retrieval of tx succeeding.

Still applicable.

apps/aecore/src/aec_db.erl
lucafavatella: Shall this be update a metric? It is optional though may help understand what node is doing. Cc @dincho

apps/aecore/src/aec_db.erl
zp-sd: I did not find said code path in the code...

zp-sd:
> Could you please add system test in `aest_db_SUITE` that creates DB using latest stable release (`:latest` Docker tag) then tries to start with current codebase (`:local` Docker tag)? Of course such a test may need to be disables in the iteration approaching a major release.

On it.

apps/aecore/src/aec_db.erl
lucafavatella: > I did not find said code path in the code...

If does not need necessarily to be part of our code. It can be any user of the user API.

Also, it may manifest itself in client (either in node or out of node) using push for ensuring that certain tx is present in mempool - the semantics of ok there is changed.

lucafavatella: > Also, it may manifest itself in client (either in node or out of node) using push for ensuring that certain tx is present in mempool - the semantics of ok there is changed.

This sentence is incorrect. Push returning ok in master already does not guarantee tx in mempool if garbage collected from mempool.

apps/aecore/src/aec_db.erl
zp-sd: No opinion here 🙂

zp-sd:
System tests: `Skipped 2 (2, 0) tests. Passed 40 tests.`

docs/release-notes/next/PT-164804897-garbage-collect-aec_signed_tx.md
```diff
@@ -0,0 +1 @@
+* Introduces new database garbage collection of aec_signed_tx objects.
```
lucafavatella: A bit more user oriented?
```suggestion
* Introduces garbage collection of transactions from the database.
```

system_test/common/aest_db_SUITE.erl
```diff
@@ -32,6 +33,7 @@
 
 all() -> [
           node_can_reuse_db_of_other_node,
+          node_can_reuse_db_of_latest_stable_node,
```
lucafavatella: 👌

apps/aecore/src/aec_db.erl
```diff
@@ -719,28 +756,39 @@ add_index_plugins() ->
...
     case Storage of
         existing_schema ->
-            case check_mnesia_tables(Tables, []) of
-                [] -> ok;
+            case check_mnesia_tables(TablesMandatory, [], []) of
+                [] ->
+                    case check_mnesia_tables(TablesOptional, [], [{create_missing_tables, true}]) of
+                        [] -> ok;
+                        Errors ->
+                            lager:error("Database check failed: ~p", [Errors]),
```
lucafavatella: Is lager guaranteed to be started at this stage? Problem may be present in existing code too.

apps/aecore/src/aec_db.erl
```diff
@@ -146,6 +155,12 @@ tables(Mode) ->
    , ?TAB(aec_discovered_pof)
     ].
 
+tables_created_if_not_present(Mode) when Mode==ram; Mode==disc ->
+    tables_created_if_not_present(expand_mode(Mode));
+tables_created_if_not_present(Mode) ->
+    [?TAB(aec_tx_gc)
```
lucafavatella: Do I understand correctly this is meant to move to `tables_manadatory` for 3.0? Could you please ensure there is ticket in Pivotal Tracker tracking this?

zp-sd: This is not critical, and creation of this table is fine if it is not present. But, for clarity, we may move this to `tables_mandatory` for 3.0. This will block upgrade from `2.3.0` to `3.0.0` though?!

apps/aecore/src/aec_db.erl
zp-sd: It is not, and - yes - this is an issue in the existing code as well. I just coded this this way to keep it consistent.

apps/aecore/src/aec_db.erl
lucafavatella: > This will block upgrade from `2.3.0` to `3.0.0` though?!

Yes.

zp-sd: I have no strong opinion here. I would keep it as is, until we find it conflicting with some introduced changes. But I can create a ticket if you find that appropriate @lucafavatella .

lucafavatella: I am just concerned of keeping complex code path. Ok, no need to.

lucafavatella:
> What is expected result if:
> 
> * Node sees chain at top block X (hence with state computed).
> * Then a new chain comes that has top block Y, and with non-empty microblocks between X and common ancestor of X&Y, with the transactions in these microblocks - let us call them Txs - not included in the chain with top block Y.
> * No miner picks transaction T in Txs - for whatever reason.
> * GC kicks in (are days needed? or less?).
> * Then a continuation of the chain from X appears, with top block X', that makes the DB state up to X reused.
> 
> ? Is transaction T meant to be in DB?
> 
> It would be good to have expected result, test, and handling.

I see you added unit test in https://github.com/aeternity/aeternity/pull/2328/commits/3b8dc6b13e52c9c30aeca85290a7d7a525dcf051#diff-a06858994975631a697cdcd2bf174f84R190
that reproduces this scenario with (for now commented) GC invoked and expecting transactions to be in chain that retakes over.

velzevur:
This shall be revisited and reimplemented

PR #2332, opened 2019-04-15
Build aefa_fate_eval.erl automatically

tolbrino:


apps/aefate/Makefile
```diff
@@ -0,0 +1,7 @@
+.PHONY: all
+all: fate_eval
+
+.PHONY: fate_eval
+fate_eval:
```
lucafavatella: Any reasons why PHONY and not the actual file, with prerequisites?

tolbrino: The prerequisites would be source files in an outside dependency: `aebytecode`. Thus, I made it  `PHONY`.

PR #2335, opened 2019-04-15
PT-159714644 Get channels parameters from chain in reestablish

cytadela8:
Documentation in aeternity/protocol#338
[Pivotal](https://www.pivotaltracker.com/story/show/159714644)

### DIFF
--- a/apps/aehttp/src/sc_ws_handler.erl
+++ b/apps/aehttp/src/sc_ws_handler.erl
@@ -236,6 +236,12 @@ read_channel_options(Params) ->
         fun(K, V) ->
             fun(M) -> maps:put(K, V, M) end
         end,
+    Error =
+        fun({error, _} = Err) ->
+            fun(_M) ->
+                Err
+            end
+        end,
     ReadMap =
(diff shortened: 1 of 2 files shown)


velzevur:
Maybe [test](https://github.com/aeternity/aeternity/blob/master/apps/aehttp/test/aehttp_integration_SUITE.erl#L4035) shall be adapted as well. Not mandatorary, though

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -255,26 +261,48 @@ read_channel_options(Params) ->
                                                      mandatory => false}),
     ReadReport = ReadMap(report, <<"report">>, #{type => boolean,
                                                      mandatory => false}),
+    OnChainOpts =
+        case (read_param(<<"existing_channel_id">>, existing_channel_id,
+                         #{type => {hash, channel}, mandatory => false}))(Params) of
```
velzevur: 🤔 why `mandatory => false`?

PR #2336, opened 2019-04-15
Add quick start section to README.md

dincho:
https://www.pivotaltracker.com/story/show/165168426


dincho:
I don't see how the docs can cause test failures.

PR #2340, opened 2019-04-16
PT165141235: Fix systemd restart on failure issues with aeternity-node (in Ubuntu package)

valkov-ivaylo:
Fixes issues with restart on failure for aeternity-node when ran with systemd (from Debian/Ubuntu package).

* Change aeternity-node start command in systemd service file to foreground.
* Change systemd service file type to simple.
* Remote console is loading
* (Sub)Process killing and restarting of service seems to be working
* Logs are duplicated in node log files and syslog. (Requested not to fix on devops stand-up.)

Tested on Ubuntu 18.04 from package. Tested on 16.04 with released binary from GitHub and service file.

### DIFF
--- a/debian/aeternity-node.service
+++ b/debian/aeternity-node.service
@@ -4,10 +4,10 @@ Requires=network-online.target
 After=network-online.target
 
 [Service]
-Type=forking
+Type=simple
 User=aeternity
 Group=aeternity
-ExecStart=/opt/aeternity/node/bin/aeternity start
+ExecStart=/opt/aeternity/node/bin/aeternity foreground &
 ExecStop=/opt/aeternity/node/bin/aeternity stop
 Restart=on-failure
 # Systemd does not play well with user set limits with ulimit.


debian/aeternity-node.service
```diff
@@ -4,10 +4,10 @@ Requires=network-online.target
 After=network-online.target
 
 [Service]
-Type=forking
+Type=simple
```
lucafavatella: Reading https://www.freedesktop.org/software/systemd/man/systemd.service.html
I see that `exec` similar to `simple` though shall complain on service start if binary absent or user absent.

If e.g. binary or user absent, and service start succeeds (as I understand `simple` does), does a health check report the service as down later?

valkov-ivaylo: Simple check on the `exec` type of the service on Ubuntu 16.04 with release binary from GitHub and the service file shows no difference in behavior with the changes in this pull request. 

My interpretation about `exec` and the suggested checks (user or binary missing) is related only to follow-up / related units and services on systemd level.

* simple:

```
[...] will immediately proceed starting follow-up units, right after creating the main service process, and before executing the service's binary.
```

* exec
```
The service manager will delay starting of follow-up units until that point. (Or in other words: simple proceeds with further jobs right after fork() returns, while exec will not proceed before both fork() and execve() in the service process succeeded.)
```

We currently have no follow-up jobs, units or services that require the `aeternity-node` to be running.

Testing with `simple` and non-existent user or binary (path) does not make the service active/running. I would have been surprised if it was otherwise. 

```
$ grep ExecStart /etc/systemd/system/aeternity-node.service 
ExecStart=/opt/aeternity/node/bin/aeternity_FAKE foreground &

$ sudo service aeternity-node start && for i in `seq 1 5` ; do date ; sudo service aeternity-node status | cat ; sleep 1; done | grep Active 
   Active: deactivating (stop) (Result: exit-code) since Thu 2019-04-25 09:09:46 EEST; 38ms ago
   Active: deactivating (stop) (Result: exit-code) since Thu 2019-04-25 09:09:47 EEST; 364ms ago
   Active: deactivating (stop) (Result: exit-code) since Thu 2019-04-25 09:09:48 EEST; 10ms ago
   Active: deactivating (stop) (Result: exit-code) since Thu 2019-04-25 09:09:49 EEST; 229ms ago
   Active: failed (Result: start-limit-hit) since Thu 2019-04-25 09:09:50 EEST; 543ms ago

```

My interpretation seems valid and the mentioned behavior is only related to systemd follow-up/related units.

PR #2342, opened 2019-04-18
Add new contract and a test case for it

ThomasArts:
According to the documentation, there should be a Contract.creator
https://github.com/aeternity/protocol/blob/master/contracts/sophia.md#account-interface

But since this is missing, this contract is prepared to use it, but not yet using it. The contract will serve as an example and having it in our test suite is good practice to show it actuallly works.


apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -5,6 +5,9 @@
 %% consensus protocol applies hence each test reinitializing the chain should
 %% take care of that at the end of the test.
 %%
+%% A single test can be ran using:
```
hanssv: s/ran/run/

PR #2343, opened 2019-04-19
PT-165440735 reward split 

sennui:


lucafavatella:
This seems WIP - as per title. Labelling as such so to ease PRs filtering.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -875,6 +875,7 @@ grant_fees(Node, Trees, Delay, FraudStatus, State) ->
     KeyFees  = db_get_fees(hash(KeyNode2)),
     Beneficiary1 = node_beneficiary(KeyNode1),
     Beneficiary2 = node_beneficiary(KeyNode2),
+    BeneficiaryProtocolFoundation = aec_governance:protocol_beneficiary(),
```
hanssv: Should we make this a list? It might be that the community decides on multiple "extra" beneficiaries.

sennui: OK, good idea.

sennui: @hanssv however... now I think it raises follow up issues: what's the ration of split between beneficiaries? We will check how much time we have to deliver it and act accordingly.

hanssv: A list of beneficiaries and ratio?

sennui: yes, actually this is how it is done in stratum

sennui:
squashed and rebased on master (conflicts in aec_chain_tests)

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -370,6 +370,26 @@
...
+                    "description" :
+                    "Public key of beneficiary account that belongs to protocol maintainers.",
+                    "type" : "string",
+                    "example" : "ak_DummyPubKeyDoNotEverUse999999999999999999999999999",
+                    "pattern": "^ak_[1-9A-HJ-NP-Za-km-z]*$"
+                },
+                "protocol_beneficiary_enabled" : {
+                    "description" :
+                    "If true, the node will split rewards and send part to protocol_beneficiary_address",
+                    "type" : "boolean"
```
lucafavatella: Could you please move default here - deleting `-define(PROTOCOL_BENEFICIARY_ENABLED, true).`?

In this way, the default value would be clearly documented to the users, while still keeping only one canonical location for the default.

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -370,6 +370,26 @@
...
+                },
+                "protocol_beneficiary_factor" : {
+                    "description" :
+                    "Part of rewards to split and send to protocol_beneficiary_address. Rounded to avoid fractions. 100 is 10%",
+                    "type" : "integer",
+                    "default": 100,
+                    "minimum" : 1,
+                    "maximum" : 1000
                 }
```
lucafavatella: Could you please add proposal of release notes line in `/docs/release-notes/next-fortuna`?

lucafavatella: Could you please add proposal of description of consensus-related change in protocol repo?

lucafavatella: Could you please add new parameters in `apps/aeutils/test/data/epoch_full.yaml`?

apps/aecore/src/aec_governance.erl
```diff
@@ -479,3 +483,31 @@ vm_gas_table() ->
        %%   }
        %%   ```
      }.
+
+
+-define(PROTOCOL_BENEFICIARY_ADDRESS, <<"ak_2uN1CEVGs4D5QnFudhp8dSqrwJEMFiTfwvbCK3SJzVYRDiusfd">>). %% TODO: random, change for real
```
lucafavatella: I see this todo item referred to in description:
> add new config section for the protocol beneficiary (now in chain)

Does that include placing real proposed one? Placing real proposed one must either be addressed in this PR, or tracked as Pivotal Tracker ticket.

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -370,6 +370,26 @@
...
+                    "If true, the node will split rewards and send part to protocol_beneficiary_address",
+                    "type" : "boolean"
+                },
+                "protocol_beneficiary_factor" : {
+                    "description" :
+                    "Part of rewards to split and send to protocol_beneficiary_address. Rounded to avoid fractions. 100 is 10%",
+                    "type" : "integer",
+                    "default": 100,
+                    "minimum" : 1,
+                    "maximum" : 1000
```
lucafavatella:
```suggestion
                    "maximum" : 999
```
?

apps/aecore/src/aec_chain_state.erl
```diff
@@ -1158,3 +1165,17 @@ calc_rewards(FraudStatus1, FraudStatus2, GenerationFees,
     LockedAmount = TotalBlockAmount - B1Amt - B2Amt,
     {B1Amt, B2Amt, LockedAmount}.
 
+reward_split_for_protocol_foundation(BeneficiaryReward1, BeneficiaryReward2, NewestNodeHeight) ->
+    ActivationHeight = aec_governance:protocol_beneficiary_activation(NewestNodeHeight),
+    ActivationFlag = aec_governance:protocol_beneficiary_enabled(),
+    if ActivationHeight and ActivationFlag ->
+        ContribFactor = aec_governance:protocol_beneficiary_factor(),
+        Contrib1 = BeneficiaryReward1 * ContribFactor div 1000, %% todo: drop zeros, div 100 after merges
```
lucafavatella: How can this TODO be addressed? (Not clear to me what it is meant to track.)

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -370,6 +370,26 @@
...
+                    "pattern": "^ak_[1-9A-HJ-NP-Za-km-z]*$"
+                },
+                "protocol_beneficiary_enabled" : {
+                    "description" :
+                    "If true, the node will split rewards and send part to protocol_beneficiary_address",
+                    "type" : "boolean"
+                },
+                "protocol_beneficiary_factor" : {
+                    "description" :
+                    "Part of rewards to split and send to protocol_beneficiary_address. Rounded to avoid fractions. 100 is 10%",
```
lucafavatella:
```suggestion
                    "Part of rewards to split and send to protocol_beneficiary_address. Truncated to avoid fractions. 100 is 10%",
```

Examples I tried:
```
1> 123 * 100 div 1000.
12
2> 1234 * 100 div 1000.
123
3> 1239 * 100 div 1000.
123
```

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -370,6 +370,26 @@
...
+                    "type" : "integer",
+                    "default": 100,
```
lucafavatella: Can code be amended to fetch default from schema i.e. here, and delete `-define(PROTOCOL_BENEFICIARY_FACTOR, 100). %% 10%`? This would enable having only one location for this value, while keeping doc.

apps/aecore/src/aec_governance.erl
sennui: This is not the the todo. I think we should propose the final one in the PR. Checkbox talks about having new subsection in the config for protocol-beneficiary, but I am not sure if it's good after all. Removing the todo, checkboxes, so it's not suggesting anything.

apps/aeutils/priv/aeternity_config_schema.json
sennui: I propose to keep in both places, similarly to e.g. EXPECTED_BLOCK_MINE_RATE_MINUTES

lucafavatella: Provided we keep them in sync, 👍

apps/aeutils/priv/aeternity_config_schema.json
sennui: I think 1000 is OK. It would be: whole reward send away.

apps/aecore/src/aec_chain_state.erl
sennui: I will remove it.

apps/aeutils/priv/aeternity_config_schema.json
sennui: @lucafavatella my concern is that the config file is very obscure from developer's point of view vs having them in clear view in arc_governance. If we go config, lets do it as refactoring and move all values.

sennui: (by obscure I mean, for example, that IDE is not automatically suggesting value when reading the code)

apps/aeutils/priv/aeternity_config_schema.json
sennui: The same as PROTOCOL_BENEFICIARY_ENABLED (and EXPECTED_BLOCK_MINE_RATE_MINUTES)

apps/aeutils/priv/aeternity_config_schema.json
sennui: done

apps/aecore/src/aec_chain_state.erl
```diff
@@ -883,15 +884,21 @@ grant_fees(Node, Trees, Delay, FraudStatus, State) ->
...
         lists:foldl(
             fun({K, Amt}, TreesAccum) when Amt > 0 ->
                     aec_trees:grant_fee(K, TreesAccum, Amt);
                 (_, TreesAccum) -> TreesAccum
             end,
             Trees,
-            [{Beneficiary1, BeneficiaryReward1},
-             {Beneficiary2, BeneficiaryReward2}]),
+            [{Beneficiary1, AdjustedReward1},
+             {Beneficiary2, AdjustedReward2},
```
lucafavatella: May I suggest we clarify https://github.com/aeternity/protocol/pull/344 before merging this PR.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -1158,3 +1165,17 @@ calc_rewards(FraudStatus1, FraudStatus2, GenerationFees,
...
     {B1Amt, B2Amt, LockedAmount}.
 
+reward_split_for_protocol_foundation(BeneficiaryReward1, BeneficiaryReward2, NewestNodeHeight) ->
+    ActivationHeight = aec_governance:protocol_beneficiary_activation(NewestNodeHeight),
+    ActivationFlag = aec_governance:protocol_beneficiary_enabled(),
+    if ActivationHeight and ActivationFlag ->
+        ContribFactor = aec_governance:protocol_beneficiary_factor(),
+        Contrib1 = BeneficiaryReward1 * ContribFactor div 1000,
+        AdjustedBeneficiaryReward1 = BeneficiaryReward1 - Contrib1,
+        Contrib2 = BeneficiaryReward2 * ContribFactor div 1000,
```
lucafavatella: Explicating:
This is portion of the sum of the following items (or zero if miner of new key block was subsequently reported to be fraudulent):
* 60% of the tx fees (i.e. explicit tx fees plus gas fees) in generations started by old key block.
* Coinbase of new key block.
* Fraud reward of the old key block (i.e. 5% of the coinbase of the old key block) if the miner of the old key block was subsequently reported to be fraudulent.

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,122 @@
...
+         allocated_shares/0,
+         total_shares/0,
+         split/3]).
+
+-include_lib("aecontract/include/hard_forks.hrl").
+
+
+-define(ENABLED, false).
+-define(TOTAL_SHARES, 1000). % 100 shares == 10% of the reward
+%%% for: "ak_2uN1CEVGs4D5QnFudhp8dSqrwJEMFiTfwvbCK3SJzVYRDiusfd:100" TODO - use actual pubkey for real account
```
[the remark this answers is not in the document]
lucafavatella: Please confirm with @noandrea re burning of portion of miner reward in testnet (in case split enabled).

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,122 @@
...
+        Val when is_boolean(Val) ->
+            Val
+    end.
+
+
+unparse_beneficiary({PK, Share}) ->
+    <<(aeser_api_encoder:encode(account_pubkey, PK))/binary, ":", (integer_to_binary(Share))/binary>>.
+
+parse_beneficiary(BeneficiaryShareStr) ->
+    Regex = "^(?'account'ak_[1-9A-HJ-NP-Za-km-z]*):(?'share'[0-9]+)$",
```
lucafavatella: Can share of a beneficiary be configured to be zero?

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -370,6 +370,20 @@
...
+                    "description" :
+                    "If true, the node will split rewards and send part to protocol_beneficiaries",
+                    "type" : "boolean",
+                    "default" : true
+                },
+                "protocol_beneficiaries" : {
+                    "type" : "array",
+                    "items" : {
+                        "description" : "Public keys belonging to protocol maintainers with reward shares (100 is 10%).",
+                        "pattern": "^ak_[1-9A-HJ-NP-Za-km-z]*:[0-9]+$"
```
lucafavatella: Can share of a beneficiary be configured to be zero?

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,122 @@
...
+        {ok, Beneficiaries} ->
+            {_, Shares} = lists:unzip(Beneficiaries),
+            AllocShares = lists:sum(Shares),
+            case AllocShares =< ?TOTAL_SHARES of
+                true ->
+                    application:set_env(aecore, dev_reward_enabled, true),
+                    application:set_env(aecore, dev_reward_allocated_shares, AllocShares),
+                    application:set_env(aecore, dev_reward_beneficiaries, Beneficiaries);
+                false ->
+                    lager:error("sum of shares for protocol beneficiaries is too large (> ~p)",
```
lucafavatella: Who guarantees lager available in setup hook?

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,122 @@
...
+ensure_env() ->
+    Enabled = cfg(<<"protocol_beneficiaries_enabled">>, ?ENABLED),
+    Benefs0 = cfg(<<"protocol_beneficiaries">>, lists:map(fun unparse_beneficiary/1, ?BENEFICIARIES)),
+    case Enabled andalso parse_beneficiaries(Benefs0) of
+        false ->
+            application:set_env(aecore, dev_reward_enabled, false);
+        {ok, Beneficiaries} ->
+            {_, Shares} = lists:unzip(Beneficiaries),
+            AllocShares = lists:sum(Shares),
+            case AllocShares =< ?TOTAL_SHARES of
```
lucafavatella: Any check on AllocShares positive?

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,122 @@
...
+            {_, DevRewards} =
+                lists:foldl(
+                  fun ({PK, PKShares}, {Remaining, Acc}) ->
+                          Reward0 = DevContrib * PKShares div AllocShares,
+                          Reward1 = min(Reward0, Remaining),
+                          {Remaining - Reward1, [{PK, Reward1} | Acc]}
+                  end, {DevContrib, []}, beneficiaries()),
+            {{BeneficiaryReward1 - AbsContrib1, BeneficiaryReward2 - AbsContrib2},
+             DevRewards};
+        _ ->
```
lucafavatella:
```suggestion
        {A, E} when is_boolean(A), is_boolean(E) ->
```

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,122 @@
...
+            {error, Reason}
+    end.
+
+
+split(BeneficiaryReward1, BeneficiaryReward2, NewestNodeHeight) ->
+    case {activated(NewestNodeHeight), enabled()} of
+        {true, true} ->
+            AllocShares = allocated_shares(),
+            TotalShares = total_shares(),
+            AbsContrib1 = BeneficiaryReward1 * AllocShares div TotalShares,
```
lucafavatella:
```suggestion
            AbsContrib1 = (BeneficiaryReward1 * AllocShares) div TotalShares,
```

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,122 @@
...
+            AbsContrib2 = BeneficiaryReward2 * AllocShares div TotalShares,
```
lucafavatella:
```suggestion
            AbsContrib2 = (BeneficiaryReward2 * AllocShares) div TotalShares,
```

Totally optional readability suggestion.

----


```
3> 2 * 3 div 5.                                    
1
4> 2 * (3 div 5).
0
5> (2 * 3) div 5.
1
```

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,122 @@
...
+    env(dev_reward_enabled, ?ENABLED).
+beneficiaries() ->
+    env(dev_reward_beneficiaries, ?BENEFICIARIES).
+total_shares() ->
+    ?TOTAL_SHARES.
+allocated_shares() ->
+    case env(dev_reward_allocated_shares, undefined) of
+        undefined ->
+            AllocShares = lists:foldl(fun ({_, X}, A) -> A + X end, 0, beneficiaries()),
+            application:set_env(aecore, dev_reward_allocated_shares, AllocShares),
```
lucafavatella: May I suggest to avoid lazily setting this environment variable in a getter? It reduces readability and, I expect, makes setup/teardown difficult.

Is deriving this value each time so slow to need lazily setting it?

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,122 @@
...
+
+split(BeneficiaryReward1, BeneficiaryReward2, NewestNodeHeight) ->
+    case {activated(NewestNodeHeight), enabled()} of
+        {true, true} ->
+            AllocShares = allocated_shares(),
+            TotalShares = total_shares(),
+            AbsContrib1 = BeneficiaryReward1 * AllocShares div TotalShares,
+            AbsContrib2 = BeneficiaryReward2 * AllocShares div TotalShares,
+            DevContrib  = AbsContrib1 + AbsContrib2,
+            {_, DevRewards} =
```
lucafavatella: What prevents value to be matched with 0?
```suggestion
            {0, DevRewards} =
```
If value is not zero, it means that either some coins are not-assigned-and-not-locked, or that some tokens are created, right?

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,122 @@
...
+                lists:foldl(
+                  fun ({PK, PKShares}, {Remaining, Acc}) ->
+                          Reward0 = DevContrib * PKShares div AllocShares,
+                          Reward1 = min(Reward0, Remaining),
+                          {Remaining - Reward1, [{PK, Reward1} | Acc]}
```
lucafavatella: Can `Remaining - Reward1` be a negative quantity? `Reward1` is `Remaining` or larger (based on `min` in line above).

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -1117,10 +1157,28 @@ fees_delayed_reward() ->
...
+with_dev_reward_env_instrumentation(KVs, F) ->
+    Prev = lists:foldl(
+             fun ({K, _}, Acc) ->
+                     case application:get_env(aecore, K) of
+                         {ok, PrevVal} -> [{set_env, [K, PrevVal]} | Acc];
+                         undefined -> [{unset_env, [K]} | Acc]
+                     end
+             end, [], KVs),
+    [application:set_env(aecore, K, V) || {K, V} <- KVs],
+    try F()
```
lucafavatella: I plan to come back to tests after other comments clarified - but from a quick look it seems to me that the set_env (and teardown) can be moved to proper setup/teardown.

apps/aeutils/priv/aeternity_config_schema.json
sennui: we work on that here: https://github.com/aeternity/protocol/pull/345

apps/aecore/src/aec_chain_state.erl
sennui: merged

apps/aecore/src/aec_dev_reward.erl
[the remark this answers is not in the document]
sennui: it is here: https://www.pivotaltracker.com/story/show/165891851

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,122 @@
...
+
+-define(ENABLED, false).
+-define(TOTAL_SHARES, 1000). % 100 shares == 10% of the reward
+%%% for: "ak_2uN1CEVGs4D5QnFudhp8dSqrwJEMFiTfwvbCK3SJzVYRDiusfd:100" TODO - use actual pubkey for real account
+-define(BENEFICIARIES, [{<<250,151,56,184,99,123,38,230,217,93,156,146,231,90,209,233,15,203,25,102,32,140,178,207,59,171,81,2,249,197,198,61>>, 100}]).
+
+ensure_env() ->
+    Enabled = cfg(<<"protocol_beneficiaries_enabled">>, ?ENABLED),
+    Benefs0 = cfg(<<"protocol_beneficiaries">>, lists:map(fun unparse_beneficiary/1, ?BENEFICIARIES)),
+    case Enabled andalso parse_beneficiaries(Benefs0) of
```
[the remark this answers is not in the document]
skkw: Looks sometimes sneaky - one of the few benefits of Erlang's dynamic typing. 
It saves one level of branching though.

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,122 @@
...
+                    application:set_env(aecore, dev_reward_enabled, true),
+                    application:set_env(aecore, dev_reward_allocated_shares, AllocShares),
+                    application:set_env(aecore, dev_reward_beneficiaries, Beneficiaries);
+                false ->
+                    lager:error("sum of shares for protocol beneficiaries is too large (> ~p)",
+                                [?TOTAL_SHARES]),
+                    exit({invalid_protocol_beneficiaries, sum_shares_too_large})
+            end;
+        {error, Reason} ->
+            lager:error("invalid protocol beneficiaries: ~p", [Reason]),
```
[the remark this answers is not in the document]
skkw: Aeutil which reads the config has lager as dependency here: https://github.com/aeternity/aeternity/blob/master/apps/aeutils/src/aeutils.app.src#L8

apps/aecore/src/aec_dev_reward.erl
skkw: Yes it can, but if that happens the amount is 0 as well and such balance update is skipped here: https://github.com/aeternity/aeternity/blob/pt-165440735-coinbase-split/apps/aecore/src/aec_chain_state.erl#L894

apps/aeutils/priv/aeternity_config_schema.json
skkw: The regexp allows it, but it's harmless. (same reason as above - no balance update happens).

apps/aecore/src/aec_dev_reward.erl
skkw: Aeutils which reads the config and runs hooks has lager as dependency in app.src file.

apps/aecore/src/aec_dev_reward.erl
skkw: The shares forming the sum of shares must pass the regex in config which prevents negative numbers.

apps/aecore/src/aec_dev_reward.erl
skkw: I think it can't go below 0. If Reward0 is smaller than Remaining than next Remaining is positive, if larger, next Remaining is 0 and stays that way until end of fold.

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -370,6 +370,20 @@
...
                     }
+                },
+                "protocol_beneficiaries_enabled" : {
+                    "description" :
+                    "If true, the node will split rewards and send part to protocol_beneficiaries",
+                    "type" : "boolean",
+                    "default" : true
+                },
+                "protocol_beneficiaries" : {
+                    "type" : "array",
```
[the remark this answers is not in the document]
skkw: Good idea, thanks.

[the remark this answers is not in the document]
sennui: fixed by @skkw

apps/aecore/src/aec_dev_reward.erl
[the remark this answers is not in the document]
lucafavatella: But ensure_env is run as setup hook as part of setup start, right (I did not triple check, going from memory here)? Who guarantees that lager started before setup?

apps/aecore/src/aec_dev_reward.erl
lucafavatella: Same as other comment.

apps/aecore/src/aec_dev_reward.erl
lucafavatella: I see check on non negative. I am asking on check on AllocShares positive (non-zero) though.

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,122 @@
...
+                {ok, PubKey} ->
+                    {PubKey, binary_to_integer(Share)};
+                {error, _} ->
+                    error({invalid_protocol_beneficiary, Account})
+            end;
+        _ ->
+            error({invalid_protocol_beneficiary_share, BeneficiaryShareStr})
+    end.
+
+parse_beneficiaries(BeneficiarySharesStrs) ->
```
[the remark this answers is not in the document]
lucafavatella: Unresolving as question seems still applicable to me. (Question. Maybe there is good reason for not enforcing that.)

apps/aecore/src/aec_dev_reward.erl
lucafavatella: Unresolving as question seems still applicable to me.

apps/aecore/src/aec_dev_reward.erl
lucafavatella: Unresolving as questions seems still applicable to me.

apps/aecore/src/aec_dev_reward.erl
lucafavatella: I do not easily see that Reward0 is not larger than Remaining.

If you are confident that `Remaining - Reward1` is never negative quantity, can you please add hardcoded expectation?

And if Reward0 is not larger than Remaining, why having the `min` call in the first place?

apps/aecore/src/aec_dev_reward.erl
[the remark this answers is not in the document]
skkw: I see, it's not guaranteed. Will remove that logs.

apps/aecore/src/aec_dev_reward.erl
skkw: Will remove these lager logs...

apps/aecore/src/aec_dev_reward.erl
skkw: I am adding a if-branch for this case so it fails if it's 0.

apps/aecore/src/aec_dev_reward.erl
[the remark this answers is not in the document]
skkw: Will add this, with "minItems" in the schema file it's fine.

apps/aecore/src/aec_dev_reward.erl
skkw: This isn't really easy question to answer - it's more in the math area. 
My suspicion is that there *is* a combination of shares which could result in accumulation of not-assigned-and-not-locked amount at the end of fold, as normal float division could end up with result like 13.55, but with DIV this is cut to 13. 
So I think the best idea here is to distribute the amounts as is, but if we end up with some not assigned amount, take it and add add it to the last beneficiary - it would be just small remains (if any) - will implement this.

apps/aecore/src/aec_dev_reward.erl
skkw: Thinking in the line of previous comment (^^), the min can be removed, as DIV will cut off the fractional part, so the intermittently accumulated not-assigned-not-locked would end up in Remaining. The implementation when we take this remaining amount (if non-zero), after fold, and add it to last beneficiary should fix this and then all tokens will be distributed.

lucafavatella:
> Would it be possible to rebase on recent master? I see several changes on master.
> 
> Would it be possible to confirm that system tests have no regressions? Either via local full run (CI on branch/PR does not run all of them) or by pushing commit to [`system-tests` branch](https://github.com/aeternity/aeternity/blob/f2a7e5beef6cc3ac5a6c15897fe980dd4abf1c34/.circleci/config.yml#L1085) and commenting on results.

I took the liberty of fetching this branch and master, locally merging this branch on master, then triggering a few workflows for testing:
* [Full system tests running on recent master with this branch merged in it](https://circleci.com/workflow-run/7ef8af68-4ea9-4ec1-b4e5-368e1f670db2).
* [Full system tests running on recent master with this branch merged in it, and with dev reward enabled](https://circleci.com/workflow-run/aa4988cc-fedb-40f6-8661-c5377a153839).
* [Normal branch CI workflow running on recent master with this branch merged in it, and with dev reward enabled](https://circleci.com/workflow-run/869beaff-58a4-4c9e-bc79-ec00f345768a).

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,113 @@
...
+                    {PubKey, binary_to_integer(Share)};
+                {error, _} ->
+                    error({invalid_protocol_beneficiary, Account})
+            end;
+        _ ->
+            error({invalid_protocol_beneficiary_share, BeneficiaryShareStr})
+    end.
+
+parse_beneficiaries([_|_] = BeneficiarySharesStrs) ->
+    try lists:foldl(fun (Str, Acc) -> [parse_beneficiary(Str) | Acc] end,
```
lucafavatella: I am trying to understand if any order of beneficiaries is relevant for consensus, and if this is deterministic.

This function is reversing the input list. Ok.
```
1> lists:foldl(fun (X, Acc) -> [X | Acc] end, [], [1,2,3])                                     
1> .
[3,2,1]
```

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,113 @@
...
+split(BeneficiaryReward1, BeneficiaryReward2, NewestNodeHeight) ->
+    case enabled() andalso activated(NewestNodeHeight) of
+        true ->
+            AllocShares = allocated_shares(),
+            TotalShares = total_shares(),
+            AbsContrib1 = (BeneficiaryReward1 * AllocShares) div TotalShares,
+            AbsContrib2 = (BeneficiaryReward2 * AllocShares) div TotalShares,
+            DevContrib  = AbsContrib1 + AbsContrib2,
+            {Leftover, [{PK0, Amount0} | RemDevRewards]} =
+                lists:foldl(
```
lucafavatella: This reverses `beneficiaries()` (again) - ok.

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,113 @@
...
+
+split(BeneficiaryReward1, BeneficiaryReward2, NewestNodeHeight) ->
+    case enabled() andalso activated(NewestNodeHeight) of
+        true ->
+            AllocShares = allocated_shares(),
+            TotalShares = total_shares(),
+            AbsContrib1 = (BeneficiaryReward1 * AllocShares) div TotalShares,
+            AbsContrib2 = (BeneficiaryReward2 * AllocShares) div TotalShares,
+            DevContrib  = AbsContrib1 + AbsContrib2,
+            {Leftover, [{PK0, Amount0} | RemDevRewards]} =
```
lucafavatella: So this `PK0` is the first in the `Benefs0` var in `ensure_env()` - ok.

apps/aecore/src/aec_dev_reward.erl
skkw: If the beneficiaries in config file are identical (including order) all is fine. 
But if the order is different, it will break consensus, as the last one in config will also get leftovers (if any) from accumulation of integer division. 
Not sure how serious it is, I expect that if somebody wants to use this feature, they will copy paste the list of beneficiaries.

apps/aecore/src/aec_dev_reward.erl
skkw: Last in config. (foldl)

lucafavatella:
> * [Normal branch CI workflow running on recent master with this branch merged in it, and with dev reward enabled](https://circleci.com/workflow-run/869beaff-58a4-4c9e-bc79-ec00f345768a).

`eunit_fortuna` job [failed](https://circleci.com/gh/aeternity/aeternity/31603):
```
  1) aec_chain_tests:forking_test_/0: Generate candidate from top with high fork
     Failure/Error: {error,
                        {badmatch,undefined},
                        [{aec_dev_reward,allocated_shares,0,
                             [{file,
                                  "/home/builder/aeternity/apps/aecore/src/aec_dev_reward.erl"},
                              {line,55}]},
                         {aec_dev_reward,split,3,
                             [{file,
                                  "/home/builder/aeternity/apps/aecore/src/aec_dev_reward.erl"},
                              {line,98}]},
                         {aec_chain_state,grant_fees,5,
                             [{file,
                                  "/home/builder/aeternity/apps/aecore/src/aec_chain_state.erl"},
                              {line,890}]},
                         {aec_chain_state,apply_node_transactions,4,
                             [{file,
                                  "/home/builder/aeternity/apps/aecore/src/aec_chain_state.erl"},
                              {line,821}]},
                         {aec_chain_state,calculate_state_for_new_keyblock,3,
                             [{file,
                                  "/home/builder/aeternity/apps/aecore/src/aec_chain_state.erl"},
                              {line,190}]},
                         {aec_chain_tests,'-fork_gen_key_candidate/0-fun-1-',
                             2,
                             [{file,
                                  "/home/builder/aeternity/apps/aecore/test/aec_chain_tests.erl"},
                              {line,891}]},
                         {aec_chain_tests,fork_gen_key_candidate,0,
                             [{file,
                                  "/home/builder/aeternity/apps/aecore/test/aec_chain_tests.erl"},
                              {line,890}]}]}
```

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -370,6 +370,21 @@
...
+                    "default" : false
+                },
+                "protocol_beneficiaries" : {
+                    "type" : "array",
+                    "items" : {
+                        "description" : "Public keys belonging to protocol maintainers with reward shares (100 is 10%).",
+                        "pattern": "^ak_[1-9A-HJ-NP-Za-km-z]*:[0-9]+$"
+                    },
+                    "minItems": 1,
+                    "default" : ["ak_2uN1CEVGs4D5QnFudhp8dSqrwJEMFiTfwvbCK3SJzVYRDiusfd:100"]
```
lucafavatella: Needs update.

skkw: Will do

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -0,0 +1,135 @@
...
+-export([split_int/5]).
+-endif.
+
+-include_lib("aecontract/include/hard_forks.hrl").
+
+
+-define(ENABLED, false).
+-define(TOTAL_SHARES, 1000). % 100 shares == 10% of the reward
+%%% for: "ak_2KAcA2Pp1nrR8Wkt3FtCkReGzAi8vJ9Snxa4PcmrthVx8AhPe8:100"
+-define(BENEFICIARIES, [{<<172,241,128,85,116,104,119,143,197,105,4,192,224,207,200,138,230,84,111,38,89,33,239,21,201,183,185,209,19,60,109,136>>, 100}]).
```
thepiwo: 100 will be the percentage of shares? how can I validate the binary corresponds to the given account?

lucafavatella: > how can I validate the binary corresponds to the given account?

```
$ docker run --rm -it --entrypoint bin/aeternity --network=none aeternity/aeternity:v2.5.0 console
...
(aeternity@localhost)1> aeser_api_encoder:encode(account_pubkey, <<172,241,128,85,116,104,119,143,197,105,4,192,224,207,200,138,230,84,111,38,89,33,239,21,201,183,185,209,19,60,109,136>>).
<<"ak_2KAcA2Pp1nrR8Wkt3FtCkReGzAi8vJ9Snxa4PcmrthVx8AhPe8">>
```

PR #2346, opened 2019-04-23
PT-162439780: Automatic CI build of Debian/Ubuntu packages

valkov-ivaylo:
Adds CircleCI configuration for building Ubuntu/Debian packages on tag/release and master.
Ubuntu/Debian (*.deb) package build passes (https://circleci.com/gh/aeternity/aeternity/24377).
Package is not uploaded anywhere yet - it will be part of separate ticket.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,10 @@ executors:
     docker:
       - image: aeternity/builder
         user: builder
+  builder_container_1804:
+    docker:
+      - image: aeternity/builder:1804
+        user: builder
 
 references:
   container_config: &container_config
@@ -160,6 +164,16 @@ references:
         make prod-package
         mkdir ${PACKAGES_DIR:?}
         mv _build/prod/rel/aeternity/aeternity-$(cat VERSION).tar.gz ${PACKAGE_TARBALL:?}
+
+  build_ubuntu_package: &build_ubuntu_package
+    run:
+      name: Build Ubuntu package (*.deb)
+      environment:
+        PACKAGES_DIR: *packages_workspace
+      command: |
+        make prod-deb-package
+        mkdir ${PACKAGES_DIR:?}
+        mv _build/../../*aeternity-node_$(cat VERSION).*.deb ${PACKAGES_DIR:?}
 
   package_tests_workspace: &package_tests_workspace /tmp/package_tests
   test_package: &test_package
@@ -588,6 +602,29 @@ jobs:
           root: *packages_workspace
           paths:
             - "*.tar.gz"
+      - *fail_notification
+
+  ubuntu_package:
+    executor: builder_container_1804
+    <<: *container_config
+    steps:
+      - run:
+          name: Checkout Ubuntu packaging branch
+          command: git clone https://github.com/aeternity/aeternity.git -b ubuntu_pacakging_162439771 --single-branch
+      - *set_package_path
+      - *build_ubuntu_package
+      - *test_package
+      - store_artifacts:
+          path: /tmp/package_tests/node1/log
+      - store_artifacts:
+          path: /tmp/package_tests/node2/log
+      - store_artifacts:
+          path: /tmp/package_tests/node3/log
+      - *store_package_artifacts
+      - persist_to_workspace:
+          root: *packages_workspace
+          paths:
+            - "*.deb"
       - *fail_notification
 
   upload_build_artifacts:
@@ -1084,6 +1121,11 @@ workflows:
                 - system-tests
             tags:
               only: *tag_regex
+
+      - ubuntu_package:
+          filters:
+            branches:
+              only: ci_deb_package_build_162439780
 
       - upload_build_artifacts:
           requires:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -554,6 +587,26 @@ jobs:
...
 
+  ubuntu_package:
+    executor: builder_container_1804
+    <<: *container_config
+    working_directory: ~/aeternity_deb
+    steps:
+      - checkout
+      - *build_ubuntu_package
+      - store_artifacts:
+          path: /tmp/package_tests/node1/log
```
lucafavatella: Who creates these folders? What are they meant to contain?

.circleci/config.yml
```diff
@@ -554,6 +587,26 @@ jobs:
...
+      - store_artifacts:
+          path: /tmp/package_tests/node2/log
```
lucafavatella: Who creates these folders? What are they meant to contain?

.circleci/config.yml
```diff
@@ -554,6 +587,26 @@ jobs:
...
+      - store_artifacts:
+          path: /tmp/package_tests/node3/log
```
lucafavatella: Who creates these folders? What are they meant to contain?

.circleci/config.yml
valkov-ivaylo: Leftovers from the copy of `linux_package` from which the `ubuntu_package` rule was initially started. Probably I should remove them.

.circleci/config.yml
valkov-ivaylo: Leftovers from the copy of `linux_package` from which the `ubuntu_package` rule was initially started. Probably I should remove them.

dincho: yeah, please cleanup

.circleci/config.yml
```diff
@@ -10,6 +10,10 @@ executors:
     docker:
       - image: aeternity/builder
         user: builder
+  builder_container_1804:
+    docker:
+      - image: aeternity/builder:1804
```
lucafavatella: I see there is [CI job pushing it](https://github.com/aeternity/docker-builder/blob/1b919eeac1628ea41eb2f525665148cad9e82862/.circleci/config.yml#L71). Ok.

.circleci/config.yml
```diff
@@ -156,6 +160,35 @@ references:
...
+        # rebar3).
+        #
+        # The debian/rules file has conditional *clean targets based on
+        # DEB_SKIP_DH_AUTO_CLEAN.
+        #
+        # This prevents the presence of broken/dummy debian/rules file without
+        # clean targets. For example when running debuild directly without
+        # `make prod-deb-package`.
+      command: |
+        make prod-clean
```
lucafavatella: It looks like this group of steps assumes clean initial state. If so, why placing commands for clean (that could stain) rather than ensuring in the place where this item is referenced that there is only checkout in a clean container?

.circleci/config.yml
```diff
@@ -156,6 +160,35 @@ references:
...
+        make clean
+        make distclean
+        if [ "$CIRCLE_TAG" ];
```
lucafavatella: So this is actually two jobs: one for master, and one for tag.

I wonder if it would be more readable to have two separate jobs. Though debatable so ok.

lucafavatella:
May benefit from a follow up PR with some leftover.

.circleci/config.yml
valkov-ivaylo: Yes they are both run from here. There were some discussions about that in the devops stand-ups. My initial intention was two jobs, but it will duplicate configuration, duplicate effort to manage it and might be unclear long-term which is what and why.

.circleci/config.yml
valkov-ivaylo: This is unfortunately an ugly hack, because of CircleCI. 

Debian/Ubuntu packaging tools and rules always run clean targets before building. However with our build container only in CircleCI, rebar3 has issues running clean form within the Debian wrappers (frakeroot especially). This is not observed on the same build container on a stand-alone machine with Docker in it.

This just mimics  as much as possible the Debian/Ubuntu packaging tasks when in CircleCI. Just to be safe. The Debian rules file does not run the clean targets when ran in CircleCI. 

I see no other work-around. It is at all not clear why container behaves like this in CircleCI. 

I think I am not confusing and did not forget where and when this happens, since the conclusion was made a while ago.

.circleci/config.yml
valkov-ivaylo: Addressed in 2cff31a

.circleci/config.yml
valkov-ivaylo: Removed in 2cff31a

.circleci/config.yml
valkov-ivaylo: Left-overs from copied job. Removed in 2cff31a

.circleci/config.yml
dincho: What Luca meant that we don't need the extra clean steps here, as the workdir is clean already

.circleci/config.yml
dincho: I don't remember the discussion, I also prefer separate jobs.

PR #2348, opened 2019-04-23
Set default value for info field in key headers ...

gorillainduction:
...to track miner release adaption.

https://www.pivotaltracker.com/story/show/165529000


lucafavatella:
Some tests failed on CI, I checked already known and tracked in Pivotal Tracker, I took liberty of restarting CI.

apps/aecore/src/aec_headers.erl
```diff
@@ -226,9 +227,15 @@ new_key_header(Height, PrevHash, PrevKeyHash, RootHash, Miner, Beneficiary,
                 pow_evidence = Evd,
                 nonce        = Nonce,
                 time         = Time,
+                info         = Info,
```
lucafavatella: I see default in record was `<<>>`. Ok.

PR #2344, opened 2019-04-22
Pt 159406473 sc continuous chain monitoring

uwiger:
Addresses [PT #159406473](https://www.pivotaltracker.com/story/show/159406473) and [#159406302](https://www.pivotaltracker.com/story/show/159406302). Work in progress. Missing e.g. some test case(s) in `aehttp_integration_SUITE`.

PR #2349, opened 2019-04-23
Pt 159406473 sc continuous chain monitoring -- fix test micro fork

lucafavatella:
Auxiliary PR for #2344


lucafavatella:
@uwiger There are some failures, though this seems a step forward from #2344 , could you please consider for inclusion.

PR #2350, opened 2019-04-24
Store rebar3.crashdump in CircleCI

tolbrino:


.circleci/config.yml
```diff
@@ -174,6 +174,9 @@ references:
       path: *packages_workspace
       destination: /
 
+  store_rebar3_crashdump: &store_rebar3_crashdump
+    store_artifacts:
+      path: rebar3.crashdump
```
lucafavatella: If file is absent, does CI fail?

tolbrino: No, it just doesn't store it.

lucafavatella: I see from other usage that relative path works... so it looks ok.

PR #2358, opened 2019-04-25
SC remove pinned env

velzevur:
PT [SC: Remove pinned env from FP and offchain tx](https://www.pivotaltracker.com/story/show/165593706)

This is `aeternity` node's part of the ticket


apps/aechannel/src/aesc_force_progress_tx.erl
```diff
@@ -74,53 +58,13 @@
           offchain_trees:: aec_trees:trees(),
           ttl           :: aetx:tx_ttl(),
           fee           :: non_neg_integer(),
-          nonce         :: non_neg_integer(),
-          block_hash    :: aesc_pinned_block:hash()
```
lucafavatella: Is there a released version of the software with such record? I understand yes. So this breaks DB backward compatibility, that may be ok with 3.* though we need to decide this.

velzevur: Those records are allowed only from Fortuna on, so assumption is no such records on main net

config/swagger.yaml
```diff
@@ -2638,9 +2638,6 @@ definitions:
       state_hash:
         $ref: '#/definitions/EncodedHash'
         description: 'Channel''s next state_hash'
-      block_hash:
```
lucafavatella: As this was re feature introduced for Fortuna though never used, this shall not impact users. Though I suggest adding release note line mentioning removal of this.

apps/aechannel/src/aesc_force_progress_tx.erl
```diff
@@ -391,21 +358,14 @@ state_hash(#channel_force_progress_tx{state_hash = StateHash}) ->
...
 -spec version(tx()) -> non_neg_integer().
-version(#channel_force_progress_tx{block_hash = BlockHash}) ->
-    case BlockHash =:= aesc_pinned_block:no_hash() of
-        true ->
-            ?INITIAL_VSN;
-        false ->
-            ?PINNED_BLOCK_VSN
-    end.
+version(_) ->
+    ?LATEST_VSN.
```
lucafavatella: It seems to me we rather want to keep INITIAL_VSN i.e. 1, and optionally skip 2 in future?

lucafavatella: I need to understand better if this is in serialization or DB though.

apps/aechannel/src/aesc_force_progress_tx.erl
```diff
@@ -74,23 +62,35 @@
...
+         }).
 %%%===================================================================
 %%% Conversion of old db format
 
--spec from_db_format(tx()) -> tx().
+ -spec from_db_format(tx()) -> tx().
 from_db_format(#channel_force_progress_tx{} = Tx) ->
     Tx;
 from_db_format(Tuple) ->
     case setelement(1, Tuple, old_fp_db_record) of
```
lucafavatella: Bad atom?

apps/aechannel/src/aesc_force_progress_tx.erl
lucafavatella: As I understand from https://github.com/aeternity/protocol/pull/341/files
using 3 is unneeded. A simple revert to 1 is ok?

Also - could this version 2 anywhere got to the DB for supported runtime i.e. Minerva? I expect no.

apps/aechannel/src/aesc_force_progress_tx.erl
velzevur: Thanks

config/swagger.yaml
lucafavatella: Still suggested though not strictly needed.

apps/aechannel/src/aesc_offchain_tx.erl
```diff
@@ -56,8 +54,7 @@
           channel_id         :: aeser_id:id(),
           updates            :: [aesc_offchain_update:update()],
           state_hash         :: binary(),
-          round              :: non_neg_integer(),
-          block_hash         :: aesc_pinned_block:hash()
```
lucafavatella: So this was not stored in DB as record.

apps/aechannel/src/aesc_force_progress_tx.erl
```diff
@@ -74,23 +60,35 @@
...
-          nonce         :: non_neg_integer(),
-          block_hash    :: aesc_pinned_block:hash()
+          nonce         :: non_neg_integer()
          }).
 
 -opaque tx() :: #channel_force_progress_tx{}.
 
 -export_type([tx/0]).
 
+-record(v2_db_record, {
```
lucafavatella:
```suggestion
%% Record introduced temporarily during Fortuna development, shipped in releases 2.4.0 and 2.5.0, and potentially stored in DB.
-record(v2_db_record, {	
```

apps/aechannel/src/aesc_offchain_tx.erl
velzevur: It was always stored in a serialized format and never as a record

PR #2365, opened 2019-04-26
Addresses PT #165630532, accidental forks during fast mining

uwiger:
See [#165630532](https://www.pivotaltracker.com/story/show/165630532)

> During testing, blocks are mined much faster than usual. Although an effort is made to terminate all old block candidates when the top changes, it's still possible for a fork to happen. Even though the code should be robust to fork switches, test code should be able to control when they happen and when they don't.
>
> A check could be added before inserting a locally generated block that its parent is actually the current top. This is very likely to always be true when mining at normal speeds, but can occasionally be violated during extra fast mining. We should then discard such a block and mine a new one.

### DIFF
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -49,6 +49,7 @@
         , get_mining_state/0
         , get_mining_workers/0
         , start_mining/0
+        , start_mining/1
         , stop_mining/0
         , is_leader/0
         , get_beneficiary/0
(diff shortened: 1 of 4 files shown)


apps/aecore/src/aec_conductor.erl
```diff
@@ -979,26 +980,46 @@ handle_post_block(Block, State) ->
...
 
 handle_add_block(Header, Block, State, Origin) ->
     {ok, Hash} = aec_headers:hash_header(Header),
+    Prev = aec_headers:prev_hash(Header),
+    case Prev =:= State#state.top_block_hash of
+        false when Origin =:= block_created;
+                   Origin =:= micro_block_created;
+                   Origin =:= block_received;
+                   Origin =:= micro_block_received ->
+            epoch_mining:info("Mined block has old top - discarding (~p)", [Prev]),
```
hanssv: We _only_ want to discard work if it is old because of something _we_ did! Not in other cases. This looks wrong.

If we have a problem in testing this isn't the way to solve it...

uwiger: @hanssv was that objection in reference to including `block_received` and `micro_block_received`?

I agree that those checks have no bearing on the problems observed during testing.

Anyway, I will continue to study the issue and see if I can come up with something better.

PR #2368, opened 2019-04-29
Pt 165593709 pinned env in noise

velzevur:
PT [SC noise: add pinned env hash](https://www.pivotaltracker.com/story/show/165593709)

This PR enriches off-chain communication with pinned env for:
* new off-chain updates so the `state_hash` can be computed based on a certain pinned environment as well as checking the initial upater's signatures
* new off-chain change confirmation ( `_ACK`) for second participant's signatures checks

Signatures checks shall be checked according to the pinned environment esp with regard of GAs

This is based on #2358 so the removal of pinned env must be merged before this one


velzevur:
Needs a rebase once #2358 is merged

velzevur:
Replaced by PR #2369

PR #2369, opened 2019-04-29
Introduce pinned env in noise protocol

velzevur:
PT [SC noise: add pinned env hash](https://www.pivotaltracker.com/story/show/165593709)

This PR enriches off-chain communication with pinned env for:
* new off-chain updates so the `state_hash` can be computed based on a certain pinned environment as well as checking the initial upater's signatures
* new off-chain change confirmation ( `_ACK`) for second participant's signatures checks

Signatures checks shall be checked according to the pinned environment esp with regard of GAs

This PR replaces #2368

### DIFF
--- a/apps/aechannel/src/aesc_codec.erl
+++ b/apps/aechannel/src/aesc_codec.erl
@@ -233,43 +233,53 @@ dec_ch_reest_ack(<< ChainHash:32/binary
 
 
 -type fnd_created_msg() :: #{ temporary_channel_id := chan_id()
+                            , block_hash           := binary()
                             , data                 := binary()}.
 
 -spec enc_fnd_created(fnd_created_msg()) -> binary().
 enc_fnd_created(#{ temporary_channel_id := ChanId
+                 , block_hash           := BlockHash
                  , data                 := Data }) ->
     Length = byte_size(Data),
     << ?ID_FND_CREATED:1 /unit:8
      , ChanId         :32/binary
+     , BlockHash      :32/binary
      , Length         :2 /unit:8
      , Data           :Length/bytes >>.
 
 -spec dec_fnd_created(binary()) -> fnd_created_msg().
 dec_fnd_created(<< ChanId:32/binary
+                 , BlockHash:32/binary
                  , Length:2/unit:8
                  , Data/binary >>) ->
     Length = byte_size(Data),
     #{ temporary_channel_id => ChanId
+     , block_hash           => BlockHash
      , data                 => Data}.
 
 -type fnd_signed_msg() :: #{ temporary_channel_id := chan_id()
+                           , block_hash           := binary()
                            , data                 := binary()}.
 
 -spec enc_fnd_signed(fnd_signed_msg()) -> binary().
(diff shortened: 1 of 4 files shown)


apps/aechannel/src/aesc_codec.erl
```diff
@@ -233,43 +233,53 @@ dec_ch_reest_ack(<< ChainHash:32/binary
...
     Length = byte_size(Data),
     << ?ID_FND_CREATED:1 /unit:8
      , ChanId         :32/binary
+     , BlockHash      :32/binary
      , Length         :2 /unit:8
      , Data           :Length/bytes >>.
 
 -spec dec_fnd_created(binary()) -> fnd_created_msg().
 dec_fnd_created(<< ChanId:32/binary
+                 , BlockHash:32/binary
```
lucafavatella: Doc: https://github.com/aeternity/protocol/blob/56793c15061a1b316ccb7506cea1daf2daeb3bd2/channels/OFF-CHAIN.md#funding_created

Doc needs update.

Versioning for incompatibility?

lucafavatella: In other points too.

lucafavatella:
I took the liberty of labelling as WIP - I understand planned updates.

apps/aechannel/src/aesc_codec.erl
velzevur: Addressed in aeternity/protocol#343

docs/release-notes/next-fortuna/PT-165593709_pinned_env_in_noise
```diff
@@ -0,0 +1,2 @@
+* Introduces a pinned environment in State Channels noise protocol. This is
+  not backwards compatible
```
lucafavatella: If I understand this correctly:
* The noise-based state channel protocol has no versioning. ([Past draft](https://github.com/aeternity/aeternity/pull/2204).)
* New node and old node manage to connect at TCP and noise level, though when exchanging any meaningful message either one of them (depending on who requested whom what first) close the TCP connection.

Right?

docs/release-notes/next-fortuna/PT-165593709_pinned_env_in_noise
```diff
@@ -0,0 +1,2 @@
+* Introduces a pinned environment in State Channels noise protocol. This is
```
lucafavatella: So there is no versioning in the actual channel protocol. Plus there is no versioning [in the doc](https://github.com/aeternity/protocol/pull/343) - apart from the tag name of course. May I suggest using these release notes for detailing what changed in the protocol?
```suggestion
* Introduces a pinned environment in State Channels noise protocol (off-chain), by introducing a `block_hash` field in some of the messages. This is
```

docs/release-notes/next-fortuna/PT-165593709_pinned_env_in_noise
velzevur: Yes

PR #2357, opened 2019-04-25
Pt 165208165 http api for ga

ThomasArts:
Make http endpoints work with Generalized accounts

PR #2370, opened 2019-04-29
Pt 165558593 fix ga meta fees

gorillainduction:
https://www.pivotaltracker.com/story/show/165558593

### DIFF
--- a/apps/aega/src/aega_meta_tx.erl
+++ b/apps/aega/src/aega_meta_tx.erl
@@ -34,6 +34,7 @@
          auth_data/1,
          auth_id/1,
          auth_id/2,
+         call_id/2,
          ga_id/1,
          ga_pubkey/1,
          gas_limit/2,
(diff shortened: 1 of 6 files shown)


apps/aega/src/aega_meta_tx.erl
```diff
@@ -210,6 +212,23 @@ set_meta_result(Err = {error, _}, Tx, Trees, Env) ->
     {ok, Trees1, _Env} = aeprimop:eval(SetInstructions, Trees, Env),
     Trees1.
 
+-spec call_id(tx(), aec_trees:trees()) -> aect_call:id().
```
hanssv: This could be used in #2357 perhaps?

gorillainduction: I was wondering who should rebase, but I agree with you ;-)

hanssv: You and Thomas can fight that out I think ;-) I'll rebase on the end result regardless...

gorillainduction: @ThomasArts ^^

hanssv: One alternative is that we merge both, and then I'll resolve it in my upcoming PR...

hanssv: Looking at the two functions again (with a fresh cup of coffee) I think they are short enough and different enough not to require the extra effort...

PR #2372, opened 2019-04-30
Handle GAMetaTx better in mempool

hanssv:
The mempool is doing a lot of special handling based on `Nonce` - this breaks down for GA MetaTxs... This makes a crude fix for this. In the long run the mempool could need a bigger overhaul.

### DIFF
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -384,12 +384,21 @@ int_get_candidate(MaxGas, BlockHash, #dbs{db = Db} = DBs) ->
     lager:debug("size(Db) = ~p", [ets:info(Db, size)]),
     MinMinerGasPrice = aec_tx_pool:minimum_miner_gas_price(),
     GasLimit = aec_governance:tx_base_gas(spend_tx),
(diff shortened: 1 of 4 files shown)


apps/aecore/src/aec_tx_pool.erl
```diff
@@ -427,10 +436,11 @@ fold_txs([], Gas, _, _, _, _, _, _, Acc) ->
...
 
 int_get_candidate_({?KEY(_, _, Account, Nonce, _) = Key, Tx},
-                   Gas, MinMinerGasPrice, Db, Dbs, AccountsTree, Height, Acc) ->
-    case gb_trees:is_defined({Account, Nonce}, Acc) of
-        true ->
-            %% The earlier must have had higher fee. Skip this tx.
+                   Gas, MinMinerGasPrice, Db, Dbs, AccountsTree, Height,
+                   Acc = #{ tree := AccTree }) ->
+    case gb_trees:is_defined({Account, Nonce}, AccTree) of
+        true when Nonce > 0 ->
```
velzevur: So this does not apply for meta txs (having a [nonce = 0](https://github.com/aeternity/aeternity/blob/master/apps/aega/src/aega_meta_tx.erl#L155-L157)

PR #2374, opened 2019-04-30
Relax requirements for deploying next environment

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/165714971


lucafavatella:
This only changes deployment so no point in waiting for CI.

PR #2383, opened 2019-05-03
PT-164597896 Remove deprecated compiler API + make aesophia a test dependency

hanssv:
This solves https://www.pivotaltracker.com/story/show/164597896, https://www.pivotaltracker.com/story/show/164597852, and https://www.pivotaltracker.com/story/show/164597902

General approach for tests has been to remove tests doing off-chain calls and compiler tests only, and replacing compilation, calldata creation, and data decode with direct calls to the compiler elsewhere.


rebar.lock
```diff
@@ -69,8 +65,8 @@
   0},
  {<<"jobs">>,{pkg,<<"jobs">>,<<"0.9.0">>},0},
  {<<"jsx">>,
-  {git,"https://github.com/talentdeficit/jsx.git",
-      {ref,"3074d4865b3385a050badf7828ad31490d860df5"}},
+  {git,"git://github.com/talentdeficit/jsx.git",
+      {ref,"fc2a001073f2300ba38427c23e83d5673c020542"}},
```
lucafavatella: I see is it tag 2.9.0.

I see https://github.com/talentdeficit/jsx/compare/3074d4865b3385a050badf7828ad31490d860df5...fc2a001073f2300ba38427c23e83d5673c020542

lucafavatella: For understanding context:
* Any specific reasons for upgrade (good in general)?
* Any expected changes in API.
* How does this version of the code behave in encoding/decoding [this transaction in sdk-testnet](https://sdk-testnet.aepps.com/v2/transactions/th_2goCaiAVxSa73A3DVSmUFgFZCW4hc9M1vz425TGgd2zEdGpTRP) (see same tx in [seed testnet node](http://52.10.46.160:3013/v2/transactions/th_2goCaiAVxSa73A3DVSmUFgFZCW4hc9M1vz425TGgd2zEdGpTRP)).

hanssv: I haven't touched this, rebar3 is inherently broken, and since we keep using it this is what you get!?

hanssv:
No, I have no plans on adding anything else for the python mess, the alternative which I did consider is `git rm -rf py/`...

rebar.lock
hanssv: But this could certainly be reverted, the change is not deliberate.

lucafavatella: It used to be taken from https://github.com/aeternity/aesophia/blob/71b97cba6219ebdb021ce02cc813d4c19825b36b/rebar.lock#L18

It probably means that two aesophia and other dep depended on different versions of library.

PR #2388, opened 2019-05-06
PT-165538326 Improve test case stability

hanssv:
Occasionally we get a `bad_match` error in `tx_is_on_chain.init_per_group` and at other times we fail to setup the test case properly ending in a failure.

This PR takes away the `bad_match` and also tries harder to setup the test case (in init_per_group), and if it is unsuccessful it will `skip` the test group rather than `fail`.

### DIFF
--- a/apps/aehttp/test/aehttp_integration_SUITE.erl
+++ b/apps/aehttp/test/aehttp_integration_SUITE.erl
@@ -644,19 +644,23 @@ init_per_group(on_micro_block = Group, Config) ->
     {ok, Tx} = aecore_suite_utils:spend(Node, Pub, Pub, 1, ?SPEND_FEE),
     {ok, [Tx]} = rpc:call(Node, aec_tx_pool, peek, [infinity]),
     ct:log("Spend tx ~p", [Tx]),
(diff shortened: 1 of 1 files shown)


apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -6085,27 +6093,31 @@ sc_ws_broken_call_code_(Owner, GetVolley, _CreateContract, _ConnPid1, _ConnPid2,
...
-                    ct:pal(
-                        "Unexpected transactions in mempool:~n~p~nUnexpected new block events:~n~p",
-                        [MempoolTxs, NewBlocks]),
-                    ct:fail("Unexpected transactions in mempool")
+            case [ aec_blocks:type(B) || B <- NewBlocks ] of
+                [key, micro] -> %% We had a microfork but then added the Tx again!
+                    {ok, NewBlocks};
+                [key]        -> %% We had a microfork and stopped, we can retry
+                    mine_micro_block_emptying_mempool_or_fail(Node, Retries - 1);
+                _Other       ->
```
ThomasArts: Would it make sense to also here continue mining, until for example `length(NewBlocks)`> 5... that is, give it some more change to succeed...

and then really fail if that does not work? Not sure here.

PR #2394, opened 2019-05-08
Use CircleCI variables context

dincho:
In scope of https://www.pivotaltracker.com/story/show/161649989

references:
- https://circleci.com/docs/2.0/contexts/
- https://circleci.com/docs/2.0/configuration-reference/#context
- [ae-node-builds context Config](https://circleci.com/gh/organizations/aeternity/settings#contexts/81741cd0-d6ba-4380-8c8a-49c625fc7ca2)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -951,6 +951,7 @@ workflows:
                 - system-tests
 
       - docker_smoke_tests:
+          context: ae-node-builds
           filters:
             branches:
               ignore:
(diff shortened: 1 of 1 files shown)


dincho:
> * Is it tracked as separate ticket?
I just created https://www.pivotaltracker.com/story/show/165888560
> * Does it mean that if preparing point release from latest stable tag then we need to backport this CI change? (No problem with that - just checking.)
Yes

dincho:
> Could you please check who is the user invoking the job considered to be for:
> 
> * Scheduled job.
> * Merge to master.
> * Tag.

We merged the infrastructure contexts, so I'll test with it.

dincho:
As we don't use secrets in scheduled jobs, this should be good to merge.

PR #2396, opened 2019-05-08
Pt 165593720 sc move updates to noise

velzevur:
PT [SC: move `updates` in off-chain tx to noise](https://www.pivotaltracker.com/story/show/165593720)


velzevur:
Left TODO:
[ ] Adapt off-chain WebSocket API to match off-chain client's serialization
[ ] On-chain updates at creation time

PR #2401, opened 2019-05-09
aecore_txs_SUITE: Skip checks in case of too many micro-forks

hanssv:
https://www.pivotaltracker.com/story/show/165538182

### DIFF
--- a/apps/aecore/test/aecore_txs_SUITE.erl
+++ b/apps/aecore/test/aecore_txs_SUITE.erl
@@ -93,17 +93,14 @@ txs_gc(Config) ->
     Height1 = Height0 + length(Blocks1), %% Very unlikely to be > 6...
 
     %% At Height1 there should be 4 transactions in mempool
-    case Height1 of
-        2                -> pool_check(N1, 4);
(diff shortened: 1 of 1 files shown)


apps/aecore/test/aecore_txs_SUITE.erl
```diff
@@ -93,17 +93,14 @@ txs_gc(Config) ->
...
-    end,
+    pool_check(N1, 4, {Height1, 2}),
 
     %% Mine 1 more key block to ensure one TX is GC:ed.
     aecore_suite_utils:mine_key_blocks(N1, 1),
     Height2 = Height1 + 1,
 
     %% Now at Height2 there should be 3 transactions in mempool - _GC1 is GC:ed
-    pool_check(N1, 3),
+    pool_check(N1, 3, {Height2, 6}),
```
lucafavatella: I understand this is main change.

hanssv: The TX is GC:ed at 7, so 6 is the highest height were we can see this effect.

PR #2405, opened 2019-05-10
Pt 165857097 add gas and value to calls

gorillainduction:
https://www.pivotaltracker.com/story/show/165857097

### DIFF
--- a/apps/aefate/src/aefa_fate.erl
+++ b/apps/aefate/src/aefa_fate.erl
@@ -17,12 +17,12 @@
 
 -export([get_trace/1]).
 
-%% Type handling.
 -export([ bind_args_from_signature/2
         , check_return_type/1
         , check_signature/2
         , check_type/2
         , get_function_signature/2
+        , push_gas_cap/2
         , push_return_address/1
         , set_function/3
         , set_local_function/2
@@ -388,9 +388,27 @@ push(V, ES) ->
 push_return_address(ES) ->
     aefa_engine_state:push_return_address(ES).
 
+%% Push a gas cap on the call stack to limit the available gas, but keep the
+%% remaining gas around.
+push_gas_cap(Gas, ES) when not ?IS_FATE_INTEGER(Gas) ->
+    abort({value_does_not_match_type, Gas, integer}, ES);
+push_gas_cap(Gas, ES) when ?IS_FATE_INTEGER(Gas) ->
+    GasInt = ?FATE_INTEGER_VALUE(Gas),
+    case GasInt > 0 of
+        false ->
+            abort({call_error, bad_gas_cap}, ES);
+        true ->
+            aefa_engine_state:push_gas_cap(GasInt, ES)
+    end.
+
 pop_call_stack(ES) ->
     case aefa_engine_state:call_stack(ES) of
         [] -> {stop, ES};
+        [{gas_store, StoredGas}| Rest] ->
+            Gas = StoredGas + aefa_engine_state:gas(ES),
+            ES1 = aefa_engine_state:set_gas(Gas, ES),
+            ES2 = aefa_engine_state:set_call_stack(Rest, ES1),
+            pop_call_stack(ES2);
         [{Contract, Function, BB, Mem, Value}| Rest] ->
             ES0 = aefa_engine_state:set_call_value(Value, ES),
             ES1 = aefa_engine_state:set_call_stack(Rest, ES0),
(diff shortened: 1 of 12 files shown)


rebar.config
```diff
@@ -62,7 +62,7 @@
                    {ref, "9ff46af"}}},
 
         {aebytecode, {git, "https://github.com/aeternity/aebytecode.git",
-                     {ref,"2555868"}}},
+                     {ref,"2f4e1888"}}},
```
hanssv: Why 8 characters?

apps/aefate/src/aefa_fate.erl
```diff
@@ -374,18 +397,46 @@ push(V, ES) ->
...
+            pop_call_stack(ES2);
+        [{Contract, Function, BB, Mem, Value}| Rest] ->
+            ES0 = aefa_engine_state:set_call_value(Value, ES),
+            ES1 = aefa_engine_state:set_call_stack(Rest, ES0),
             ES2 = aefa_engine_state:set_memory(Mem, ES1),
             ES3 = set_function(Contract, Function, ES2),
             {jump, BB, ES3}
     end.
 
+abort_out_of_gas(ES) ->
```
UlfNorell: Doesn't actually abort, so call it something else?

gorillainduction:
Rebase to resolve conflicts

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -138,34 +141,75 @@ returnr(Arg0, EngineState) ->
...
+    {Fun, ES1} = get_op_arg(Arg0, EngineState),
     Signature = aefa_fate:get_function_signature(Fun, ES1),
-    {ok, ES2} = aefa_fate:check_signature_and_bind_args(Signature, ES1),
+    ok = aefa_fate:check_signature(Signature, ES1),
+    ES2 = aefa_fate:bind_args_from_signature(Signature, ES1),
     {jump, 0, aefa_fate:set_local_function(Fun, ES2)}.
 
-call_r(Arg0, Arg1, EngineState) ->
+call_r(Arg0, Arg1, Arg2, EngineState) ->
+    Sender = aefa_engine_state:current_contract(EngineState),
```
hanssv: Could be done in `call_r_common`?

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -138,34 +141,75 @@ returnr(Arg0, EngineState) ->
...
+    ES5 = transfer_value(Sender, Address, Value, ES4),
+    {jump, 0, ES5}.
 
-call_t(Arg0, EngineState) ->
-    {Fun, ES0} = get_op_arg(Arg0, EngineState),
-    Signature = aefa_fate:get_function_signature(Fun, ES0),
-    {ok, ES2} = aefa_fate:check_signature_and_bind_args(Signature, ES0),
-    {jump, 0, aefa_fate:set_local_function(Fun, ES2)}.
+call_gr(Arg0, Arg1, Arg2, Arg3, EngineState) ->
+    Sender = aefa_engine_state:current_contract(EngineState),
```
hanssv: Could be done in `call_gr_common`?

rebar.config
gorillainduction: Fumbly fingers? No reason whatsoever.

apps/aefate/src/aefa_fate_op.erl
gorillainduction: :thinking: Yes. It just needs to happen before the binding of new arguments.

apps/aefate/test/aefa_sophia_test.erl
```diff
@@ -62,6 +74,7 @@ make_call(Contract, Function, Arguments) ->
     Calldata = {tuple, {Function, {tuple, EncArgs}}},
     #{ contract => pad_contract_name(Contract),
        gas      => 1000000,
+       value    => 10000,
        call     => aeb_fate_encoding:serialize(Calldata) }.
```
UlfNorell: Might consider adding a `Call.value` to the remote test. Perhaps
```
  function remote(x : int) = x * (2 + Call.value)
```

apps/aefate/src/aefa_fate.erl
gorillainduction: Sure. It used to do abort in the end ;-)

apps/aefate/test/aefa_sophia_test.erl
gorillainduction: I would if the compiler could handle it ;-)
I see another PR in the future.

hanssv: 🔮

PR #2411, opened 2019-05-13
Check for reentrant remote calls in FATE

gorillainduction:
https://www.pivotaltracker.com/story/show/165876613

### DIFF
--- a/apps/aecontract/test/aecontract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_SUITE.erl
@@ -164,6 +164,7 @@ groups() ->
                           , sophia_identity
                           , sophia_remote_identity
                           , sophia_spend
+                          , sophia_no_reentrant
                           , fate_environment
                           , fate_call_origin
                           , fate_call_value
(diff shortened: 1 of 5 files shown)


apps/aefate/src/aefa_engine_state.erl
```diff
@@ -147,15 +164,48 @@ push_arguments([], Acc, Stack, ES) ->
...
+            {remote, Contract, Function, BB,
+             ES#es{ call_value = Value
+                  , memory = Mem
+                  , call_stack = Rest
+                  , seen_contracts = Seen
+                  }}
+    end.
+
+pop_seen_contracts(Pubkey, #es{seen_contracts = Seen}) ->
+    [_|Seen1] = lists:dropwhile(fun(X) -> X =/= Pubkey end, Seen),
```
UlfNorell: Note: non-`Pubkey` entries are remote tail calls. Maybe a comment would be appropriate.

gorillainduction: Isn't it obvious? :stuck_out_tongue: 
I'll add a comment on this.

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -1451,9 +1464,10 @@ sophia_no_reentrant(_Cfg) ->
     Acc   = ?call(new_account, 10000000 * aec_test_utils:min_gas_price()),
     IdC   = ?call(create_contract, Acc, identity, {}),
     RemC  = ?call(create_contract, Acc, remote_call, {}, #{amount => 100}),
-    Err   = {error, <<"reentrant_call">>},
-            %% Should fail due to reentrancy
-    Err   = ?call(call_contract, Acc, RemC, staged_call, word, {IdC, RemC, 77}),
+    Res   = ?call(call_contract, Acc, RemC, staged_call, word, {IdC, RemC, 77}),
+    %% Should fail due to reentrancy
+    ?assertMatchFATE({error, <<"Reentrant call">>}, Res),
```
hanssv: This difference might have been avoidable...

PR #2412, opened 2019-05-13
Narrow down the spec for aesc_channels:payload()

velzevur:


uwiger:
Added manually to #2425 in commit [943dcee](https://github.com/aeternity/aeternity/pull/2425/commits/943dcee32890c3fdd9074b94a66f7f8478414652)

PR #2415, opened 2019-05-14
PT-165853843 Properly encode binary JSON-content

hanssv:
Some fields in our HTTP APIs are really generic binary data (byte arrays); in some places (especially fields called `payload`) we use an ordinary JSON string to transfer these. This does not work well, since the strings are interpreted as UTF8, and not all bytes/byte combinations are valid UTF8...

This PR fixes this, and also adds a level of granularity to the Swagger-file distinguishing public keys from hashes, and from other encoded values.

### DIFF
--- a/apps/aecore/src/aec_spend_tx.erl
+++ b/apps/aecore/src/aec_spend_tx.erl
@@ -212,7 +212,7 @@ for_client(#spend_tx{sender_id    = SenderId,
       <<"fee">>          => Fee,
       <<"ttl">>          => TTL,
       <<"nonce">>        => Nonce,
-      <<"payload">>      => Payload}.
+      <<"payload">>      => aeser_api_encoder:encode(bytearray, Payload)}.
 
 -spec version(tx()) -> non_neg_integer().
 version(_) ->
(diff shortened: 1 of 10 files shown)


apps/aecore/src/aec_spend_tx.erl
```diff
@@ -212,7 +212,7 @@ for_client(#spend_tx{sender_id    = SenderId,
       <<"fee">>          => Fee,
       <<"ttl">>          => TTL,
       <<"nonce">>        => Nonce,
-      <<"payload">>      => Payload}.
```
ThomasArts: Indeed, no hard-fork height needed, since this is just towards clients.

rebar.config
```diff
@@ -65,7 +65,7 @@
                      {ref,"08a09b0"}}},
 
         {aeserialization, {git, "https://github.com/aeternity/aeserialization.git",
-                           {ref,"816bf99"}}},
+                          {ref,"3416ff5"}}},
```
lucafavatella: I see https://github.com/aeternity/aeserialization/compare/816bf99...3416ff5
Ok.

PR #2407, opened 2019-05-10
Pt 159406473 sc continuous chain monitoring 3

uwiger:
Rebased on top of latest changes.

Addresses:
* [#159406473 SC Watcher: listen for on-chain txs](https://www.pivotaltracker.com/story/show/159406473)
* [#159406302 SC WS: Solo closing](https://www.pivotaltracker.com/story/show/159406302)

Of particular note, the FSM now stays up during the closing phase, allowing the client to request `close_solo`, `slash` or `settle`. The FSM may also automatically slash if configured to do so (no tests exist for this). Documentation remains to be written. Also, http integration tests for the solo-closing phase.

PR #2418, opened 2019-05-14
PT #165980068 - Use trace_runner wrapper for testing

uwiger:
See [PT #165980068](https://www.pivotaltracker.com/story/show/165980068)

Extracted from #2407

### DIFF
--- a/docs/release-notes/next/PT-165980068-use-trace_runner.md
+++ b/docs/release-notes/next/PT-165980068-use-trace_runner.md
@@ -1,1 +1,0 @@
-* Add support for using the [trace_runner](https://github.com/uwiger/trace_runner) wrapper in test suites for better debugging.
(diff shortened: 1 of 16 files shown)


docs/release-notes/next/PT-165980068-use-trace_runner.md
```diff
@@ -0,0 +1 @@
+* Add support for using the [trace_runner](https://github.com/uwiger/trace_runner) wrapper in test suites for better debugging.
```
lucafavatella: Is this change meant to be user oriented? 🤔 If so there should be some doc.

I looks to me this is developer-only, not visible to users, hence shall not be present in release notes.

uwiger: @lucafavatella Release notes mention removed.

rebar.config
```diff
@@ -89,7 +89,12 @@
 
         % waiting for https://github.com/szktty/erlang-lz4/pull/12 to be merged
         {lz4, {git, "https://github.com/aeternity/erlang-lz4.git",
-              {ref, "1ff9f36"}}}
+              {ref, "1ff9f36"}}},
+
+        % for lightweight trace-based debugging
+        {trace_runner, {git, "git://github.com/uwiger/trace_runner.git",
```
lucafavatella: I see related comment https://github.com/aeternity/aeternity/pull/2407#discussion_r283734107

PR #2419, opened 2019-05-14
PT #165979853 Add log extraction script

uwiger:
See [PT #165979853](https://www.pivotaltracker.com/story/show/165979853)

Extracted from PR #2407

### DIFF
--- a/docs/release-notes/next/PT-165979853-extract-log-info.md
+++ b/docs/release-notes/next/PT-165979853-extract-log-info.md
@@ -1,1 +1,0 @@
-* Add a script for extracting session-specific log info from aeternity.log files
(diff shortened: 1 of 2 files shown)


docs/release-notes/next/PT-165979853-extract-log-info.md
```diff
@@ -0,0 +1 @@
+* Add a script for extracting session-specific log info from aeternity.log files
```
lucafavatella: Is scripts/log_xtract.escript shipped to the users as part of the package? Is it user oriented? It does not look to me so I see no sense in having a release note line for this.

velzevur: Although it could be used by users, I think this is mainly for developer's convenience so probably the release notes line will be misleading

uwiger: Yes, it's mainly for developers, although, as Dimitar says, it could perhaps become useful for advanced users or app developers.

PR #2422, opened 2019-05-14
Satisfy dialyzer

velzevur:


uwiger:
I made this change already in #2418 before taking a look at this.

PR #2425, opened 2019-05-14
#165982215 Refactor and clarify types in channel-related modules

uwiger:
Extracted from #2407 

See [PT #165982215](https://www.pivotaltracker.com/story/show/165982215)

### DIFF
--- a/apps/aechannel/src/aesc_channels.erl
+++ b/apps/aechannel/src/aesc_channels.erl
@@ -76,7 +76,7 @@
 -type pubkey() :: aec_keys:pubkey().
 -type amount() :: non_neg_integer().
 -type seq_number() :: non_neg_integer().
--type payload() :: aesc_offchain_tx:tx() | <<>>.
+-type payload() :: aesc_offchain_tx:tx().
 -type hash32() :: <<_:256>>.
 -type auth() :: basic | {hash32(), aeser_id:id()}.
 -type serialized() :: binary().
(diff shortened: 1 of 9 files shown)


apps/aechannel/src/aesc_channels.erl
```diff
@@ -57,20 +57,41 @@
...
+             , serialized/0
+             , seq_number/0
+             , payload/0
+             ]).
+
 -type id() :: aeser_id:id().
 -type pubkey() :: aec_keys:pubkey().
 -type amount() :: non_neg_integer().
 -type seq_number() :: non_neg_integer().
--type payload() :: aesc_offchain_tx:tx() | <<>>.
```
lucafavatella: This removal of `<<>>` is the only unclear thing to me.

velzevur: In this context the payload must be an off-chain transaction. Although in solo closing sequence we could use an empty payload, there are distinct functions in this module that handle this specific case. For example we have a function for solo close with [an empty payload](https://github.com/aeternity/aeternity/blob/943dcee32890c3fdd9074b94a66f7f8478414652/apps/aechannel/src/aesc_channels.erl#L123-L127) and yet another for [non-empty one](https://github.com/aeternity/aeternity/blob/943dcee32890c3fdd9074b94a66f7f8478414652/apps/aechannel/src/aesc_channels.erl#L129-L133). Note that the latter has a strong expectations for the `payload()` being passed to be explicitly an `aesc_offchain_tx:tx()`.

PR #2426, opened 2019-05-15
Bump FP base price

velzevur:
PT [SC FP minimum base gas bump](https://www.pivotaltracker.com/story/show/166001363)

PR #2428, opened 2019-05-15
Tighter gas limit for ga meta tx

hanssv:
https://www.pivotaltracker.com/story/show/166009375 and https://www.pivotaltracker.com/story/show/166018387

### DIFF
--- a/apps/aega/src/aega_meta_tx.erl
+++ b/apps/aega/src/aega_meta_tx.erl
@@ -92,7 +92,7 @@ gas(#ga_meta_tx{gas = Gas}) ->
 
 -spec gas_limit(tx(), non_neg_integer()) -> amount().
 gas_limit(#ga_meta_tx{gas = Gas, tx = InnerTx}, Height) ->
-    aetx:gas_limit(aetx_sign:tx(InnerTx), Height) + Gas.
+    aetx:inner_gas_limit(aetx_sign:tx(InnerTx), Height) + Gas.
 
 -spec gas_price(tx()) -> amount().
 gas_price(#ga_meta_tx{gas_price = GasPrice}) ->
--- a/apps/aetx/src/aetx.erl
+++ b/apps/aetx/src/aetx.erl
@@ -15,6 +15,7 @@
         , fee/1
         , from_db_format/1
         , gas_limit/2
+        , inner_gas_limit/2
         , min_gas/2
         , gas_price/1
         , min_gas_price/2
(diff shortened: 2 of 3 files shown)


hanssv:
Conflicts with #2426

hanssv:
Rebased on top of #2426

apps/aega/src/aega_meta_tx.erl
```diff
@@ -92,7 +92,7 @@ gas(#ga_meta_tx{gas = Gas}) ->
 
 -spec gas_limit(tx(), non_neg_integer()) -> amount().
 gas_limit(#ga_meta_tx{gas = Gas, tx = InnerTx}, Height) ->
-    aetx:gas_limit(aetx_sign:tx(InnerTx), Height) + Gas.
+    aetx:inner_gas_limit(aetx_sign:tx(InnerTx), Height) + Gas.
```
lucafavatella: Not introduced by this PR, though it seems to me that the `aetx:gas_limit` for `ga_meta_tx`, that [recurses using `CB:gas_limit`](https://github.com/aeternity/aeternity/pull/2428/files#diff-b7fc65d08fbfee442d57ae86b1b2576cR259), violates the semantics assumed by `aext_pool:check_candidate`:
> %% In case 0 is returned, the tx will not be included in the micro block
> %% candidate by the mempool.

in which the inner tx may be invalid though included. Or is that code path never exercised hence that semantics can be made tighter (enforcing never error)?

lucafavatella: I have not followed yet what is impact of not complying with this semantics in micro block candidate building - in case this code path (error) is exercised.

lucafavatella: I see this change is for [not accounting twice size of inner tx for the sake of micro block gas limit](https://www.pivotaltracker.com/n/projects/2124891/stories/166018387). So strictly speaking this is done. Ok.

hanssv: `aetx:gas_limit` will never return 0 for a `ga_meta_tx` there is nothing that can go wrong at this level for this transaction. If the inner transaction has a bad oracle TTL or is an offchain_channel_tx the MetaTx can still be included on the chain with the effects of the authentication...

hanssv: It wouldn't be wrong to deny such a Tx, but accepting it is also OK.

apps/aetx/src/aetx.erl
```diff
@@ -261,19 +262,37 @@ gas_limit(#aetx{ type = Type, cb = CB, size = Size, tx = Tx }, Height) when Type
...
+    FeeGas = if Kind == outer -> min_gas(AETx, Height);
+                Kind == inner -> min_gas(AETx, Height) - size_gas(Size)
+             end,
+    FeeGasPrice = (CB:fee(Tx) + FeeGas - 1) div FeeGas,
+    case Type of
+        ga_meta_tx ->
+            %% Also compute the minimum gas price for the wrapped Tx - make sure
+            %% not to count the size twice!
+            InnerMinGasPrice = min_gas_price(aetx_sign:tx(CB:tx(Tx)), Height, inner),
+            lists:min([CB:gas_price(Tx), FeeGasPrice, InnerMinGasPrice]);
```
lucafavatella: I see the changes in this function `min_gas_price/2` address https://www.pivotaltracker.com/n/projects/2124891/stories/166009375

Behaviour is the same, except that for ga meta tx gas price may be lower than before if inner tx is particularly underincetivized.

Ok.

PR #2431, opened 2019-05-15
Pt 159406473 sc continuous chain monitoring 4

uwiger:
See [PT #159406302](https://www.pivotaltracker.com/story/show/159406302) and [PT #159406473](https://www.pivotaltracker.com/story/show/159406473)

Supersedes #2407 after having extracted functionality into separate PRs and then rebased on top of them.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -47,6 +47,7 @@
 -export([minimum_depth_achieved/4,     %% (Fsm, OnChainId, Type, TxHash)
          channel_changed_on_chain/2,   %% (Fsm, Info)
          channel_closing_on_chain/2,
+         channel_closed_on_chain/2,    %% (Fsm, Info)
          channel_unlocked/2]). %% (Fsm, Info)
 
 -export([init/1,
@@ -76,6 +77,7 @@
          open/3,
          channel_closing/3,   % on-chain closing has been detected
          mutual_closing/3,
+         channel_closed/3,
          disconnected/3]).
 
 -export([timeouts/0,
@@ -112,6 +114,8 @@
 -define(WATCH_FND, funding).
 -define(WATCH_DEP, deposit).
 -define(WATCH_WDRAW, withdraw).
+-define(WATCH_CLOSED, closed).
+-define(MIN_DEPTH, min_depth).
 
 -define(UPDATE_CAST(R), R==?UPDATE; R==?DEP_CREATED; R==?WDRAW_CREATED).
 -define(UPDATE_REQ(R),
@@ -145,6 +149,12 @@
                          ; T =:= ?SHUTDOWN_ACK
                          ; T =:= ?CH_REESTABL
                          ; T =:= ?CH_REEST_ACK).
+
+-define(WATCHER_EVENT(T), T =:= ?CHANNEL_CHANGED
+                        ; T =:= ?CHANNEL_UNLOCKED
+                        ; T =:= ?MIN_DEPTH_ACHIEVED
+                        ; T =:= ?CHANNEL_CLOSING
+                        ; T =:= ?CHANNEL_CLOSED).
 
 -define(KEEP, 10).
 
@@ -277,6 +287,9 @@ channel_changed_on_chain(Fsm, Info) ->
 channel_closing_on_chain(Fsm, Info) ->
     gen_statem:cast(Fsm, {?CHANNEL_CLOSING, Info}).
 
+channel_closed_on_chain(Fsm, Info) ->
+    gen_statem:cast(Fsm, {?CHANNEL_CLOSED, Info}).
+
 channel_unlocked(Fsm, Info) ->
     lager:debug("sending unlocked", []),
     gen_statem:cast(Fsm, {?CHANNEL_UNLOCKED, Info}).
@@ -325,6 +338,7 @@ timer_subst(wdraw_signed            ) -> funding_lock;
 timer_subst(initialized             ) -> accept;
 timer_subst(mutual_closing          ) -> accept;
 timer_subst(channel_closing         ) -> idle;
+timer_subst(channel_closed          ) -> idle;
 timer_subst(channel_changed         ) -> idle.
 
 default_timeouts() ->
@@ -481,6 +495,8 @@ connection_died(Fsm) ->
     stop_ok(catch gen_statem:stop(Fsm)).
 
 stop_ok(ok) ->
+    ok;
+stop_ok({'EXIT', noproc}) ->
     ok;
 stop_ok({'EXIT', {normal,{sys,terminate,_}}}) ->
     ok.
@@ -779,7 +795,7 @@ awaiting_signature(cast, {?SIGNED, ?FND_CREATED, SignedTx} = Msg,
                                                 latest = {ack, ?FND_CREATED,
                                                           HSCTx, Updates}})),
             report_on_chain_tx(?FND_CREATED, NewSignedTx, D1),
-            {ok, D2} = start_min_depth_watcher(?WATCH_FND, NewSignedTx,
+            {ok, D2} = start_min_depth_watcher({?MIN_DEPTH, ?WATCH_FND}, NewSignedTx,
                                                Updates, D1),
             gproc_register_on_chain_id(D2),
             next_state(awaiting_locked, D2)
@@ -795,7 +811,7 @@ awaiting_signature(cast, {?SIGNED, ?DEP_CREATED, SignedTx} = Msg,
                                                            ?DEP_CREATED,
                                                            HSCTx, Updates}}),
             report_on_chain_tx(?DEP_CREATED, NewSignedTx, D1),
-            {ok, D2} = start_min_depth_watcher(?WATCH_DEP, NewSignedTx,
+            {ok, D2} = start_min_depth_watcher({?MIN_DEPTH, ?WATCH_DEP}, NewSignedTx,
                                                Updates, D1),
             next_state(awaiting_locked,
                       log(rcv, ?SIGNED, Msg, D2))
@@ -811,7 +827,7 @@ awaiting_signature(cast, {?SIGNED, ?WDRAW_CREATED, SignedTx} = Msg,
                                                            ?WDRAW_CREATED,
                                                            HSCTx, Updates}}),
             report_on_chain_tx(?WDRAW_CREATED, NewSignedTx, D1),
-            {ok, D2} = start_min_depth_watcher(?WATCH_WDRAW, NewSignedTx,
+            {ok, D2} = start_min_depth_watcher({?MIN_DEPTH, ?WATCH_WDRAW}, NewSignedTx,
                                                Updates, D1),
             next_state(awaiting_locked, log(rcv, ?SIGNED, Msg, D2))
         end, D);
@@ -904,7 +920,7 @@ half_signed(cast, {?FND_SIGNED, Msg}, #data{role = initiator,
             ok = aec_tx_pool:push(SignedTx),
             D2 = D1#data{create_tx = SignedTx},
             {ack, create_tx, _, Updates} = Latest,
-            {ok, D3} = start_min_depth_watcher(?WATCH_FND, SignedTx, Updates, D2),
+            {ok, D3} = start_min_depth_watcher({?MIN_DEPTH, ?WATCH_FND}, SignedTx, Updates, D2),
             next_state(awaiting_locked, D3);
         {error, Error} ->
             close(Error, D)
@@ -924,7 +940,7 @@ dep_half_signed(cast, {?DEP_SIGNED, Msg}, #data{latest = Latest} = D) ->
             report_on_chain_tx(?DEP_SIGNED, SignedTx, D1),
             ok = aec_tx_pool:push(SignedTx),
             {ack, deposit_tx, _, Updates} = Latest,
-            {ok, D2} = start_min_depth_watcher(deposit, SignedTx, Updates, D1),
+            {ok, D2} = start_min_depth_watcher({?MIN_DEPTH, ?WATCH_DEP}, SignedTx, Updates, D1),
             next_state(awaiting_locked, D2);
         {error, _Error} ->
             handle_update_conflict(?DEP_SIGNED, D)
@@ -986,7 +1002,8 @@ wdraw_half_signed(cast, {?WDRAW_SIGNED, Msg}, #data{latest = Latest} = D) ->
             report_on_chain_tx(?WDRAW_SIGNED, SignedTx, D1),
             ok = aec_tx_pool:push(SignedTx),
             {ack, withdraw_tx, _, Updates} = Latest,
-            {ok, D2} = start_min_depth_watcher(withdraw, SignedTx, Updates, D1),
+            {ok, D2} = start_min_depth_watcher(
+                         {?MIN_DEPTH, ?WATCH_WDRAW}, SignedTx, Updates, D1),
             next_state(awaiting_locked, D2);
         {error, _Error} ->
             handle_update_conflict(?WDRAW_SIGNED, D)
@@ -1060,7 +1077,7 @@ is_channel_locked(LockedUntil) ->
 
 awaiting_locked(enter, _OldSt, _D) -> keep_state_and_data;
 awaiting_locked(cast, {?MIN_DEPTH_ACHIEVED, ChainId, ?WATCH_FND, TxHash},
-                #data{latest = {watch, ?WATCH_FND, TxHash, CTx, Updates}} = D) ->
+                #data{latest = {min_depth, ?WATCH_FND, TxHash, CTx, Updates}} = D) ->
     report(info, own_funding_locked, D),
     next_state(
       signed, send_funding_locked_msg(D#data{on_chain_id = ChainId,
@@ -1068,7 +1085,7 @@ awaiting_locked(cast, {?MIN_DEPTH_ACHIEVED, ChainId, ?WATCH_FND, TxHash},
                                                        Updates}}));
 awaiting_locked(cast, {?MIN_DEPTH_ACHIEVED, ChainId, ?WATCH_DEP, TxHash},
                 #data{on_chain_id = ChainId,
-                      latest = {watch, ?WATCH_DEP, TxHash, SignedTx, Updates}} = D) ->
+                      latest = {min_depth, ?WATCH_DEP, TxHash, SignedTx, Updates}} = D) ->
     report(info, own_deposit_locked, D),
     next_state(
       dep_signed, send_deposit_locked_msg(
@@ -1076,7 +1093,7 @@ awaiting_locked(cast, {?MIN_DEPTH_ACHIEVED, ChainId, ?WATCH_DEP, TxHash},
                     D#data{latest = {deposit, SignedTx, Updates}}));
 awaiting_locked(cast, {?MIN_DEPTH_ACHIEVED, ChainId, ?WATCH_WDRAW, TxHash},
                 #data{on_chain_id = ChainId,
-                      latest = {watch, ?WATCH_WDRAW, TxHash, SignedTx, Updates}} = D) ->
+                      latest = {min_depth, ?WATCH_WDRAW, TxHash, SignedTx, Updates}} = D) ->
     report(info, own_withdraw_locked, D),
     next_state(
       wdraw_signed, send_withdraw_locked_msg(
@@ -1094,7 +1111,12 @@ awaiting_locked(cast, {?DEP_ERR, _Msg}, D) ->
 awaiting_locked(cast, {?WDRAW_ERR, _Msg}, D) ->
     %% TODO: Stop min depth watcher!
     handle_update_conflict(?WDRAW_SIGNED, D);
+awaiting_locked(cast, {?CHANNEL_CHANGED, #{tx_hash := TxHash}} = Msg,
+                #data{ latest = {min_depth, ?WATCH_DEP, TxHash, _}} = D) ->
+    %% Expected
+    keep_state(log(rcv, ?CHANNEL_CHANGED, Msg, D));
 awaiting_locked(Type, Msg, D) ->
+    lager:debug("Unexpected. Msg = ~p, Latest = ~p", [Msg, D#data.latest]),
     handle_common_event(Type, Msg, error, D).
 
 awaiting_initial_state(enter, _OldSt, _D) -> keep_state_and_data;
@@ -1331,11 +1353,26 @@ channel_closing(cast, {?CHANNEL_UNLOCKED, #{chan_id := ChId}} = Msg,
             ok
     end,
     keep_state(D1);
-channel_closing(cast, {?MIN_DEPTH_ACHIEVED, _ChanId, close, undefined},
-                #data{latest = {watch, close, undefined, _}} = D) ->
-    report(info, closed, D),
-    close(closed, D);
+%% channel_closing(cast, {?MIN_DEPTH_ACHIEVED, _ChanId, close, undefined},
+%%                 #data{latest = {min_depth, close, undefined, _}} = D) ->
+%%     report(info, closed, D),
+%%     close(closed, D);
 channel_closing(Type, Msg, D) ->
+    handle_common_event(Type, Msg, discard, D).
+
+channel_closed(enter, _OldSt, D) ->
+    keep_state(D#data{ongoing_update = false});
+channel_closed(cast, {?MIN_DEPTH_ACHIEVED, ChainId, ?WATCH_CLOSED, TxHash},
+               #data{ latest = {min_depth, ?WATCH_CLOSED, TxHash, _}
+                    , on_chain_id = ChainId } = D) ->
+    close(closed_confirmed, D);
+channel_closed(cast, {?CHANNEL_CHANGED, _Info} = Msg, D) ->
+    %% This is a weird case. The channel was closed, but now isn't.
+    %% This would indicate a fork switch. TODO: detect channel status
+    %% and respond accordingly. For now. Panic and terminate
+    report(info, zombie_channel, D),
+    close(zombie_channel, log(rcv, msg_type(Msg), Msg, D));
+channel_closed(Type, Msg, D) ->
     handle_common_event(Type, Msg, discard, D).
 
 disconnected(cast, {?CH_REESTABL, _Msg}, D) ->
@@ -1347,6 +1384,9 @@ close(Reason, D) ->
 
 close_(close_mutual, D) ->
     report(info, close_mutual, D),
+    {stop, normal, D};
+close_(closed_confirmed, D) ->
+    report(info, closed_confirmed, D),
     {stop, normal, D};
 close_(leave, D) ->
     {stop, normal, D};
@@ -1663,10 +1703,24 @@ handle_common_event_(cast, {?CHANNEL_CLOSING, Info} = Msg, _St, _, D) ->
             lager:debug("Channel not found, or other: ~p", [Error]),
             close(channel_no_longer_active, D)
     end;
-handle_common_event_(cast, {?CHANNEL_CHANGED, Info} = Msg, _St, _, D) ->
-    lager:debug("Fsm notes channel change ~p", [Info]),
-    report_on_chain_tx(channel_changed, maps:get(tx, Info), D),
-    keep_state(log(rcv, msg_type(Msg), Msg, D));
+handle_common_event_(cast, {?CHANNEL_CHANGED, #{tx_hash := TxHash} = Info} = Msg,
+                     _St, _, D) ->
+    D1 = log(rcv, msg_type(Msg), Msg, D),
+    case D1#data.latest of
+        {min_depth, _, LatestTxHash, _} when TxHash =/= LatestTxHash ->
+            %% This is a problem: channel changed while waiting to confirm
+            %% other tx hash
+            close({error, unexpected_tx_on_chain}, D);
+        _ ->
+            lager:debug("Fsm notes channel change ~p", [Info]),
+            report_on_chain_tx(channel_changed, maps:get(tx, Info), D1),
+            keep_state(D1)
+    end;
+handle_common_event_(cast, {?CHANNEL_CLOSED, #{tx := SignedTx} = _Info} = Msg, _St, _, D) ->
+    %% Start a minimum-depth watch, then (if confirmed) terminate
+    report_on_chain_tx(channel_closed, SignedTx, D),
+    {ok, D1} = start_min_depth_watcher({?MIN_DEPTH, ?WATCH_CLOSED}, SignedTx, D),
+    next_state(channel_closed, log(rcv, msg_type(Msg), Msg, D1));
 handle_common_event_({call, From}, Req, St, Mode, D) ->
     case Mode of
         error_all ->
@@ -1690,6 +1744,7 @@ handle_common_event_(Type, Msg, St, Err, D) when Err==error;
 msg_type(Msg) when is_tuple(Msg) ->
     T = element(1, Msg),
     if ?KNOWN_MSG_TYPE(T) -> T;
+       ?WATCHER_EVENT(T) -> T;
        true -> unknown
     end;
 msg_type(T) -> T.
@@ -1706,10 +1761,11 @@ error_binary(E) when is_atom(E) ->
     atom_to_binary(E, latin1).
 
 
-terminate(Reason, _State, Data) ->
+terminate(Reason, _State, #data{session = Sn} = Data) ->
     lager:debug("terminate(~p, ~p, _)", [Reason, _State]),
     report(info, {died, Reason}, Data),
     report(debug, {log, win_to_list(Data#data.log)}, Data),
+    try aesc_session_noise:close(Sn) catch error:_ -> ok end,
     ok.
 
 code_change(_OldVsn, OldState, OldData, _Extra) ->
@@ -2113,8 +2169,8 @@ settle_tx(Account, Nonce, #data{ opts  = #{initiator := I,
     {ok, RAmt} = aesc_offchain_state:balance(R, State),
     Fee = maps:get(fee, Opts1),
     TTL = adjust_ttl(maps:get(ttl, Opts1, 0)),
-    {ok, _} = Ok = aesc_settle_tx:new(#{ channel_id => aec_id:create(channel, ChanId)
-                                       , from_id    => aec_id:create(account, Account)
+    {ok, _} = Ok = aesc_settle_tx:new(#{ channel_id => aeser_id:create(channel, ChanId)
+                                       , from_id    => aeser_id:create(account, Account)
                                        , initiator_amount_final => IAmt
                                        , responder_amount_final => RAmt
                                        , ttl        => TTL
@@ -2864,13 +2920,17 @@ start_min_depth_watcher(Type, SignedTx, Updates,
     evt({nonce, Nonce}),
     {OnChainId, D1} = on_chain_id(D, Tx),
     case {Type, Watcher0} of
-        {funding, undefined} ->
+        {{?MIN_DEPTH, ?WATCH_FND = Sub}, undefined} ->
             WOpts = maps:get(watcher, Opts, #{}),
             {ok, Watcher1} = aesc_fsm_min_depth_watcher:start_link(
-                               Type, TxHash, OnChainId, MinDepth, ?MODULE, WOpts),
+                               Sub, TxHash, OnChainId, MinDepth, ?MODULE, WOpts),
             evt({watcher, Watcher1}),
             {ok, D1#data{watcher = Watcher1,
-                         latest = {watch, Type, TxHash, SignedTx, Updates}}};
+                         latest = {min_depth, Type, TxHash, SignedTx, Updates}}};
+        {{?MIN_DEPTH, Sub}, Pid} when is_pid(Pid) ->
+            ok = aesc_fsm_min_depth_watcher:watch_for_min_depth(
+                   Pid, TxHash, MinDepth, ?MODULE, Sub),
+            {ok, D1#data{latest = {min_depth, Sub, TxHash, SignedTx, Updates}}};
         {unlock, Pid} when Pid =/= undefined ->
             ok = aesc_fsm_min_depth_watcher:watch_for_unlock(Pid, ?MODULE),
             {ok, D1#data{latest = {watch, unlock, TxHash, SignedTx, Updates}}};
@@ -2878,6 +2938,7 @@ start_min_depth_watcher(Type, SignedTx, Updates,
             ok = aesc_fsm_min_depth_watcher:watch_for_channel_close(Pid, MinDepth, ?MODULE),
             {ok, D1#data{latest = {watch, close, TxHash, SignedTx, Updates}}};
         {_, Pid} when Pid =/= undefined ->  % assertion
+            lager:debug("Unknown Type = ~p, Pid = ~p", [Type, Pid]),
             ok = aesc_fsm_min_depth_watcher:watch(
                    Pid, Type, TxHash, MinDepth, ?MODULE),
             {ok, D1#data{latest = {watch, Type, TxHash, SignedTx, Updates}}}
@@ -3116,6 +3177,9 @@ check_tx_and_verify_signatures(SignedTx, Updates, Type, Data, Pubkeys, ErrTypeMs
     #data{state = State, opts = Opts} = Data,
     Aetx = aetx_sign:tx(SignedTx),
     case aetx:specialize_type(Aetx) of
+        {channel_settle_tx, _Tx} ->
+            %% TODO: conduct more relevant checks
+            verify_signatures(Pubkeys, SignedTx);
         {Type, _Tx} -> % check type
             case call_cb(Aetx, channel_pubkey, []) of
                 ChannelPubkey -> % expected pubkey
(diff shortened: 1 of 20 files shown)


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1257,12 +1330,47 @@ mutual_closing(cast, {closing_signed, _Msg}, D) ->  %% TODO: not using this, rig
...
+channel_closed(cast, {?MIN_DEPTH_ACHIEVED, ChainId, ?WATCH_CLOSED, TxHash},
+               #data{ latest = {?MIN_DEPTH, ?WATCH_CLOSED, TxHash, _, _}
+                    , on_chain_id = ChainId } = D) ->
+    close(closed_confirmed, D);
+channel_closed(cast, {?CHANNEL_CHANGED, _Info} = Msg, D) ->
+    %% This is a weird case. The channel was closed, but now isn't.
+    %% This would indicate a fork switch. TODO: detect channel status
+    %% and respond accordingly. For now. Panic and terminate
+    report(info, zombie_channel, D),
+    close(zombie_channel, log(rcv, msg_type(Msg), Msg, D));
```
velzevur: This is already tracked as [Handle zombie state channels](https://www.pivotaltracker.com/story/show/165438326). Ok

lucafavatella:
From [test_minerva](https://circleci.com/gh/aeternity/aeternity/37632#tests/containers/1) CI job:
```
%%% aesc_fsm_SUITE ==> all_tests.transactions.channel_detects_close_solo: FAILED
%%% aesc_fsm_SUITE ==> {timeout,[{aesc_fsm_SUITE,receive_from_fsm,6,
                          [{file,"/home/builder/aeternity/apps/aechannel/test/aesc_fsm_SUITE.erl"},
                           {line,1613}]},
          {aesc_fsm_SUITE,receive_from_fsm,5,
                          [{file,"/home/builder/aeternity/apps/aechannel/test/aesc_fsm_SUITE.erl"},
                           {line,1601}]},
          {aesc_fsm_SUITE,settle_,5,
                          [{file,"/home/builder/aeternity/apps/aechannel/test/aesc_fsm_SUITE.erl"},
                           {line,1271}]},
          {aesc_fsm_SUITE,channel_detects_close_solo,1,
                          [{file,"/home/builder/aeternity/apps/aechannel/test/aesc_fsm_SUITE.erl"},
                           {line,670}]},
          {test_server,run_test_case_eval1,6,
                       [{file,"test_server.erl"},{line,1062}]},
          {test_server,run_test_case_eval,9,
                       [{file,"test_server.erl"},{line,994}]}]}
```

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -22,6 +22,7 @@
 
 init(Req, _Opts) ->
     lager:debug("init(~p, ~p)", [Req, _Opts]),
+    process_flag(trap_exit, true),
```
lucafavatella: Based on the [cowboy doc](https://ninenines.eu/docs/en/cowboy/2.2/guide/handlers/#_cleaning_up), this is not needed for the terminate to be called. Could you please elaborate on why it addresses?

My concern is unintended behaviour change (in calling process) hence distinct behaviour in error case in tests and user experience.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -3778,6 +3796,80 @@ sc_ws_close_mutual_(Config, Closer) when Closer =:= initiator
...
+                                                                      Config),
+    {IStartB, RStartB} = channel_participants_balances(IPubKey, RPubKey),
+    #{initiator := IConnPid,
+      responder := RConnPid} = proplists:get_value(channel_clients, Config),
+    ok = ?WS:register_test_for_channel_events(IConnPid, [sign, info, on_chain_tx]),
+    ok = ?WS:register_test_for_channel_events(RConnPid, [sign, info, on_chain_tx]),
+
+    CloseSolo =
+        fun(CloserConn, CloserPrivKey, OtherConn, OtherPrivKey) ->
+                ws_send_tagged(CloserConn, <<"channels.close_solo">>, #{}, Config),
```
lucafavatella: Where are these methods meant to be documented? In [intended usage of channels API](https://github.com/aeternity/protocol/blob/8b2724d14d8b4878b6b2c74850ec3f7f7716aeb2/node/api/channels_api_usage.md)? It seems doc requires integration.

apps/aehttp/src/sc_ws_handler.erl
uwiger: This is to ensure that the ws handler is killed immediately through cascading exit before being able to process the termination info message sent by the fsm. That is, the fsm sends an info msg before terminating, but the exit may propagate to the linked ws handler, killing it, before it has time to receive and forward that message. Trapping exits ensures that it will be able to process all messages sent by the fsm before it gets to the `'DOWN'` message.

apps/aehttp/test/aehttp_integration_SUITE.erl
uwiger: Working on the documentation additions.

apps/aehttp/src/sc_ws_api_jsonrpc.erl
```diff
@@ -39,7 +45,13 @@
                                 <<"channels.update_ack">>        -> update_ack;
                                 <<"channels.shutdown_sign">>     -> shutdown;
                                 <<"channels.shutdown_sign_ack">> -> shutdown_ack;
-                                <<"channels.leave">>             -> leave
+                                <<"channels.leave">>             -> leave;
+                                <<"channels.close_solo_tx">>     -> close_solo_tx;
+                                <<"channels.close_solo_sign">>   -> close_solo_tx;
+                                <<"channels.slash_tx">>          -> slash_tx;
+                                <<"channels.shash_sign">>        -> slash_tx;
```
lucafavatella:
```suggestion
                                <<"channels.slash_sign">>        -> slash_tx;
```
?

lucafavatella: I admit I did not understand yet where these different binaries are hooked (I see similar binaries e.g. `<<"channels.slash_tx">>`, `<<"channels.shash_sign">>`, `<<"channels.slash">>`) and which test is meant to fail if any do not match.

uwiger: :)

apps/aehttp/src/sc_ws_api_jsonrpc.erl
```diff
@@ -39,7 +45,13 @@
                                 <<"channels.update_ack">>        -> update_ack;
                                 <<"channels.shutdown_sign">>     -> shutdown;
                                 <<"channels.shutdown_sign_ack">> -> shutdown_ack;
-                                <<"channels.leave">>             -> leave
+                                <<"channels.leave">>             -> leave;
```
lucafavatella: I see this `<<"channels.leave">>` is the only element in `METHOD_TAG` but not in `METHOD_SIGNED`. Is this intended?

uwiger: The `leave` action doesn't require signing.

apps/aecore/src/aec_db.erl
```diff
@@ -244,13 +246,29 @@ backend_mode(<<"mnesia">> , #{persist := true } = M) -> M#{module => mnesia,
...
+dirty(Fun) when is_function(Fun, 0) ->
+    mnesia:activity(async_dirty, Fun).
+
 ensure_transaction(Fun) when is_function(Fun, 0) ->
     %% TODO: actually, some non-transactions also have an activity state
-    case get(mnesia_activity_state) of undefined ->
+    case get(mnesia_activity_state) of
+        undefined ->
+            transaction(Fun);
+        {_, _, non_transaction} ->
```
lucafavatella: I am concerned of this DB transaction handling change in this large PR. Would you please consider splitting DB `ensure_transaction` improvement as separate PR? I am proposing https://github.com/aeternity/aeternity/pull/2472

Any reasons why this point in the code is not using `mnesia:is_transaction()`?

uwiger: Sorry, I didn't read carefully and didn't see that you already started a new PR. But I'd prefer to also break the `aec_dirty/1` and `ensure_activity/2` changes, while I'm at it. See #2477 

The `mnesia:is_transaction()` function lacks sufficient resolution to work when you add dirty ops to the mix.

uwiger: #2477 merged into master

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2894,42 +3150,43 @@ process_update_error(Reason, From, D) ->
...
-                    {error, not_active}
-            end;
-        {error, _} = Error ->
-            Error
-    end.
-
-check_closing_event_(Ch, #data{state = St}) ->
-    Height = cur_height(),
+check_closing_event(Info, D) ->
+    aec_db:dirty(fun() ->
```
lucafavatella: I see the only DB operation inside this dirty context is `aec_chain:get_header` that anyway adds transactional context. If optimized DB read here is really necessary, I wonder: why not removing need for handling transaction inside dirty context by cutting one layer of abstraction and perform read in dirty activity directly? Something like:

```erlang
check_closing_event(Info, D) ->
    {value, Hdr} = aec_db:find_header(async_dirty, maps:get(block_hash, Info)), %% TODO `aec_db:find_header/2`.
    try check_closing_event_(maps:get(tx, Info), maps:get(channel, Info), Hdr, D)
    catch
        error:E ->
            ?LOG_CAUGHT(E),
            error(E)
    end.
```
If need be, read can be made conditional to `aesc_channels:is_solo_closing(Ch)`.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -104,122 +239,707 @@ terminate(_Reason, _St) ->
...
+                  error:E ->
+                      lager:error("CAUGHT ~p / ~p", [E, erlang:get_stacktrace()]),
+                      error(E)
+              end
+      end).
+
+-spec check_requests_([req()], #st{}, cache(), [req()]) -> #st{}.
+check_requests_([#{check_at_height := CheckAt} = R|Reqs], St, C, Acc) ->
+    lager:debug("CheckAt: R = ~p", [R]),
+    {TopHeight, C1} = top_height(C),
```
lucafavatella: Is new cache `C1` here unused on purpose?

apps/aechannel/src/aesc_fsm.erl
lucafavatella: I see that calls to `aec_db:ensure_activity(async_dirty, ...)` are out of scope for this ^^^ comment, as those follow other code path not attempting to nest transaction inside dirty activity.

apps/aecore/src/aec_db.erl
```diff
@@ -244,13 +246,29 @@ backend_mode(<<"mnesia">> , #{persist := true } = M) -> M#{module => mnesia,
...
+            %% Transaction inside a dirty context; rely on mnesia to handle it
```
lucafavatella: It may be worth to rebase on master, to see more clearly changes of this PR.

Also the definition of macro `t` in this PR is misleading, and different of what will be after merge to master.

apps/aechannel/src/aesc_utils.erl
```diff
@@ -151,6 +152,11 @@ unpack_payload(Tx) ->
         {_, _}                            -> {error, bad_offchain_state_type}
     end.
 
+is_offchain_tx_type(channel_offchain_tx) -> true;
+is_offchain_tx_type(ga_meta_tx         ) -> true;
```
lucafavatella: 🤔 ?

lucafavatella: Based on function call, I see that more meaningful name would be `is_legal_payload_of_channel_close_solo_tx`. If so, [doc](https://github.com/aeternity/protocol/blame/8b2724d14d8b4878b6b2c74850ec3f7f7716aeb2/serializations.md#L603) indicates that this shall be false:
> The payload is a serialized signed channel off-chain transaction or it is empty.

apps/aechannel/src/aesc_session_noise.erl
```diff
@@ -193,7 +217,7 @@ handle_info({'DOWN', Ref, process, Pid, _Reason},
             #st{parent_mon_ref=Ref, parent=Pid, econn=EConn}=St) ->
     lager:debug("Got DOWN from parent (~p)", [Pid]),
     close_econn(EConn),
-    {stop, normal, St};
+    {stop, shutdown, St};
```
lucafavatella: I expect this change to change the `terminate` reason though without any functional or logging change. Ok.

apps/aechannel/src/aesc_session_noise.erl
```diff
@@ -139,35 +145,53 @@ establish({accept, Port, Opts}, St) ->
...
+            exit(shutdown)
+    after T ->
+            ok
+    end.
+
+
 get_reconnect_params() ->
     %% {ConnectTimeout, Retries}
-    {10000, 30}.
+    {3000, 40}.  %% total 2 minutes
```
lucafavatella:
```suggestion
    {3000, 40}.  %% max 4 minutes (up to 40 retries, where each retry is up to 3 seconds TCP connection timeout and 3 seconds sleep time).
```

apps/aechannel/src/aesc_session_noise.erl
```diff
@@ -139,35 +145,53 @@ establish({accept, Port, Opts}, St) ->
...
         {error, _} ->
-            timer:sleep(1000),
-            connect_tcp(Retries-1, Host, Port, Opts, Timeout)
+            sleep(Timeout, Ref),
+            connect_tcp(Retries-1, Host, Port, Ref, Opts, Timeout)
     end.
 
+sleep(T, Ref) ->
+    receive
+        {'$gen_call', From, close} = Msg ->
```
lucafavatella: I see that `init` calls `proc_lib:init_ack`, then the function that calls this (delayed initialization) then `gen_server:enter_loop`. So: while process is trying-and-retrying to establish TCP connection, after each attempt it inspects mailbox for checking if it is the case to give up - i.e. either it was instructed to close (someone called `gen_server:call(..., close)`, and this code assumed implementation of messages generated by `gen_server` OTP module) or parent died (and this process not killed because processing flag...).

There may be room for simplification though ok.

apps/aechannel/src/aesc_session_noise.erl
lucafavatella: Timeout and number of retries may be made user configurable though seems ok for now.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -104,122 +239,707 @@ terminate(_Reason, _St) ->
...
+              #st{chan_id = ChanId} = St, C) ->
+    #{info := #{parent := Parent}} = R,
     lager:debug("check_status(tx_hash = ~p, parent = ~p)", [TxHash, Parent]),
-    case min_depth_achieved(TxHash, MinDepth, C) of
-        {true, C1} ->
+    case current_depth(TxHash, C) of
+        {undefined, C1} ->
+            {R, C1};
+        {Depth, C1} when Depth >= MinDepth ->
             lager:debug("min_depth achieved", []),
```
lucafavatella: It feels the computation of the current depth does not enforce the expectation that the block hash in which the tx hash was found is an ancestor of the top block hash. So in case of fork the depth computation may be misleading (shorter fork may have more work on it). Though seems good enough. Ok.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -13,87 +14,221 @@
...
+                                  , aesc_channels:locked_until()
+                                  , aesc_channels:state_hash() }.
+-type ch_status() :: undefined | #{ is_active    := aesc_channels:is_active()
+                                  , vsn          := chan_vsn()
+                                  , changed      := boolean()
+                                  , channel      := aesc_channels:channel()
+                                  , lock_period  := aesc_channels:lock_period()
+                                  , locked_until := aesc_channels:locked_until() }.
+
+%% (Commented out since dialyzer complains, but can't really explain why)
```
lucafavatella: If you paste the error I am happy to help.

I find maps more difficult to follow than records, for the lack of machine-parseable structure. Having dialyzer to perform the basic sanity check on the map would help.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -194,7 +207,7 @@ patterns() ->
                             not lists:member(F, [module_info, patterns])].
 
 record_fields(data ) -> record_info(fields, data);
-record_fields(w    ) -> record_info(fields, w);
+record_fields(w    ) -> aesc_window:record_fields(w);
 record_fields(Other) -> aesc_offchain_state:record_fields(Other).
 %% ==================================================================
```
lucafavatella: Is the state machine at line 222 up-to-date? Shall the copy of the FSM in the protocol repo be made canonical, this deleted, and[ the FSM in the protocol repo enhanced](https://github.com/aeternity/protocol/tree/master/channels#overview)?

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -841,9 +882,29 @@ awaiting_signature(cast, {?SIGNED, ?SHUTDOWN_ACK, SignedTx} = Msg,
             D1 = send_shutdown_ack_msg(NewSignedTx, D),
             D2 = D1#data{latest = undefined,
                         log    = log_msg(rcv, ?SIGNED, Msg, D1#data.log)},
-            report(on_chain_tx, NewSignedTx, D1),
+            report_on_chain_tx(close_mutual, NewSignedTx, D1),
             close(close_mutual, D2)
         end, D);
+awaiting_signature(cast, {?SIGNED, close_solo_tx, SignedTx} = Msg,
```
lucafavatella: I am surprised that this clause does not store `SignedTx` for `aesc_offchain_state:get_latest_signed_tx` in `open(enter, _OldSt, D) ->` to find it.

apps/aechannel/src/aesc_utils.erl
velzevur:
```suggestion
is_offchain_tx_type(ga_meta_tx         ) -> true; %% TODO: inspect the most inner transaction for being off-chain type
```

And I will address it in FSM's Generalized Accounts awareness PR

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
lucafavatella: Uncommenting dialyzer passed for me.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2716,27 +2957,40 @@ default_minimum_depth(responder) -> ?MINIMUM_DEPTH.
...
+                   Pid, TxHash, MinDepth, ?MODULE, Sub),
+            {ok, D1#data{latest = {?MIN_DEPTH, Sub, TxHash, SignedTx, Updates}}};
+        {unlock, Pid} when Pid =/= undefined ->
+            ok = aesc_fsm_min_depth_watcher:watch_for_unlock(Pid, ?MODULE),
+            {ok, D1#data{latest = {watch, unlock, TxHash, SignedTx, Updates}}};
+        {close, Pid} when Pid =/= undefined ->
+            ok = aesc_fsm_min_depth_watcher:watch_for_channel_close(Pid, MinDepth, ?MODULE),
+            {ok, D1#data{latest = {watch, close, TxHash, SignedTx, Updates}}};
         {_, Pid} when Pid =/= undefined ->  % assertion
+            lager:debug("Unknown Type = ~p, Pid = ~p", [Type, Pid]),
```
lucafavatella: Shall this ever happen? Is this controlled by the remote party?

This sounds more a warning than a debug.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1005,26 +1067,74 @@ withdraw_locked_complete(SignedTx, Updates, #data{state = State, opts = Opts} =
...
+                            ok;
+                        {error, Reason} ->
+                            lager:debug("Close_solo tx failed: ~p", [Reason]),
+                            ok
+                    end;
+                true ->
+                    lager:debug("channel is locked: cannot solo-close", []),
+                    ok
+            end,
+            %% Let the watcher pick up the tx once it makes it onto the chain
```
lucafavatella: Where is watcher looking for close? 🤔 

I see the only point where a watcher looks for close `start_min_depth_watcher(close, ...)` is in `settle_signed` that seems not this code path.

Maybe it is implicit in a `aesc_fsm_min_depth_watcher` started previously?

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -104,122 +239,707 @@ terminate(_Reason, _St) ->
...
+      fun() ->
+              case Event of
+                  changed_on_chain ->
+                      Mod:channel_changed_on_chain(Parent, Info);
+                  closing_on_chain ->
+                      Mod:channel_closing_on_chain(Parent, Info)
+              end
+      end, C).
+
+report_closed_on_chain(#{ tx_type    := TxType
```
lucafavatella: Unused var?
```suggestion
report_closed_on_chain(#{ tx_type    := _TxType
```

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
lucafavatella: See #2478

apps/aechannel/src/aesc_fsm.erl
lucafavatella: Can this be made more explicit by:
* Either enhancing the comment to where this watcher was started and where it was tasked with watching in relation to this code path;
* Or an assertion like "assert that watcher is aware of watching what is relevant for this point, fail otherwise"?

apps/aechannel/src/aesc_fsm.erl
uwiger: > I see the only DB operation inside this dirty context is `aec_chain:get_header` that anyway adds transactional context. 

Actually, this didn't use to be true, until PR #2477 changed the `?t(Expr)` macro in `aec_db.erl`. I believe this was a mistake. The `aec_db` read operations shouldn't add a transaction context if there is already a transaction _or_ non-transaction context.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
uwiger: No, it's a bug.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
uwiger: Hmm, when there is a `top_changed` after a fork switch, if `TxHash` was in a block that's no longer on the chain, it will be thrown back into the mempool, and `current_depth/2` will return `{undefined, C1}`. Also, the function does check whether its enclosing block is in the main chain.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -104,122 +239,707 @@ terminate(_Reason, _St) ->
...
+                      error(E)
+              end
+      end).
+
+-spec check_requests_([req()], #st{}, cache(), [req()]) -> #st{}.
+check_requests_([#{check_at_height := CheckAt} = R|Reqs], St, C, Acc) ->
+    lager:debug("CheckAt: R = ~p", [R]),
+    {TopHeight, C1} = top_height(C),
+    if TopHeight >= CheckAt ->
+            check_at_height(R, Reqs, St, C, Acc);
```
uwiger:
```suggestion
            check_at_height(R, Reqs, St, C1, Acc);
```

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -104,122 +239,707 @@ terminate(_Reason, _St) ->
...
+       true ->
+            lager:debug("Not yet (TopHeight = ~p), skip", [TopHeight]),
+            check_requests_(Reqs, St, C, [R|Acc])
```
uwiger:
```suggestion
            check_requests_(Reqs, St, C1, [R|Acc])
```

apps/aecore/src/aec_db.erl
uwiger: Actually, I believe that this macro definition is the right one, and the one currently on master was a mistake.

apps/aechannel/src/aesc_fsm.erl
lucafavatella: As this PR is not based on latest master, when merged this PR will keep the `?t` macro definition in master.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
lucafavatella: Addressed in 6ca8c386d (was https://github.com/aeternity/aeternity/pull/2478/commits/5342ead4f720c5854461a1d7fb875e4252cdc189)

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
lucafavatella: > Hmm, when there is a `top_changed` after a fork switch, if `TxHash` was in a block that's no longer on the chain, it will be thrown back into the mempool, and `current_depth/2` will return `{undefined, C1}`.

The tx may be included in the new fork too.

> Also, the function does check whether its enclosing block is in the main chain.

Correct! I had missed that. (I had probably confused block hash for tx hash.)

Good then.

apps/aecore/src/aec_db.erl
lucafavatella: This shall be now addressed in 6ca8c386d

apps/aechannel/src/aesc_fsm.erl
lucafavatella: The `?t` macro is now aligned to master in 6ca8c386d

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
lucafavatella: I see addressed.

apps/aechannel/src/aesc_window.erl
```diff
@@ -0,0 +1,117 @@
...
+add(Item, #w{n = N, a = A, keep = Keep} = W) when N < Keep ->
+    W#w{n = N+1, a = [Item|A]};
+add(Item, #w{a = A} = W) ->
+    W#w{n = 1, a = [Item], b = A}.
+
+-spec pop(window()) -> {entry(), window()} | error.
+pop(#w{a = [], b = []}) ->
+    error;
+pop(#w{a = [], b = [H|T], n = N} = W) ->
+    {H, W#w{n = N-1, b = T}};
```
lucafavatella: This is popping one element out of `b`, and reducing `n`. So, based on this, it seems that `n` keeps `b` into account.

But in the second clause of `add/2`, the body `W#w{n = 1, a = [Item], b = A}` disregards the length of the new `b` in the computation of `n` (`n` seems to represent only `a` that is a list of length 1).

I still miss the full context/usage though seems suspicious.

A comment in the definition of field `n` in the record would ease understanding of its semantics.

uwiger: Correct, `n` should not be reduced when `pop()` picks from the `b` list, or the number of elements in `b` should be included in the total. Will consider a fix.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -104,122 +233,707 @@ terminate(_Reason, _St) ->
...
+            undefined
+    end.
+
+
+sort_txs(Txs) ->
+    lists:sort(fun compare_txs/2, Txs).
+
+%% See lists:sort/2: The fun Comp(A,B) should return true if A =< B
+compare_txs(SignedA, SignedB) ->
+    aetx:nonce(aetx_sign:tx(SignedA)) =< aetx:nonce(aetx_sign:tx(SignedB)).
```
lucafavatella: Unclear to me how this will work for:
* Txs from different senders (initiator / responder). I do not have full understanding of assumptions of code on this ordering.
* Generic account meta tx' nonce - that I consider out of scope for this PR cc @velzevur 

I see no issues so far though.

apps/aechannel/src/aesc_window.erl
lucafavatella: I see addressed in abb8642

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -176,10 +218,13 @@
               , latest = undefined     :: latest_op()
               , ongoing_update = false :: boolean()
               , last_reported_update   :: undefined | non_neg_integer()
-              , log = #w{}             :: #w{}
+              , log = aesc_window:new(#{}) :: aesc_window:window()
```
lucafavatella:
```suggestion
              , log                    :: aesc_window:window()
```

lucafavatella: Invalid default for channel client FSM field log. Default is unused in init.

lucafavatella: Addressed in b7ac7d0

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
lucafavatella: I see changed in f9afc7b41c8e69f72edd677e26823135461611b4

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -13,46 +14,152 @@
...
+-type tx_req() :: #{ mode      := tx_hash
+                   , tx_hash   := tx_hash()
+                   , min_depth := aec_blocks:height()
+                   , info      := req_info() }.
+
+-type req() :: close_req() | unlock_req() | watch_req() | tx_req().
+
+-type block_hash() :: binary().
+-type tx_hash()    :: binary().
+-type scenario()   :: top | next_block | fork_switch | {has_tx, tx_hash()}.
```
lucafavatella: Type `{has_tx, tx_hash()}` is wrong. Please see #2488

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -231,46 +947,67 @@ tx_location(error, TxHash, C) ->
...
+    Res = aec_chain_state:hash_is_in_main_chain(Hash, TopHash),
+    {Res, C1}.
+
+update_tx_log(TxHash, BlockHash, #{tx_log := TxLog} =C)
+  when is_binary(BlockHash) ->
+    C#{tx_log => aesc_window:add(
+                   { {TxHash, BlockHash},
+                     #{ tx_hash      => TxHash
+                      , block_hash   => BlockHash
+                      , block_origin => chain } }, TxLog)};
```
lucafavatella: The second element in this tuple that is the entry of the `aesc_window` data structure has no `type` key here.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -104,122 +243,699 @@ terminate(_Reason, _St) ->
...
+            TxLog1 =
+                lists:foldl(
+                  fun({BlockHash, Txs}, Acc) ->
+                          lists:foldl(
+                            fun(TxHash, Acc1) ->
+                                    aesc_window:add(
+                                      { {TxHash, BlockHash},
+                                        #{ tx_hash      => TxHash
+                                         , block_hash => BlockHash
+                                         , block_origin => chain } }, Acc1)
```
lucafavatella: The second element in this tuple that is the entry of the `aesc_window` data structure has no `type` key here.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -104,122 +243,699 @@ terminate(_Reason, _St) ->
...
+log_tx(#{ tx_hash      := TxHash
+        , block_hash   := BlockHash
+        , block_origin := _Origin
+        , type         := _Type } = Info, #st{tx_log = TxLog} = St) ->
+    Key = {TxHash, BlockHash},
+    case aesc_window:keymember(Key, 1, TxLog) of
+        true ->
+            St;
+        false ->
+            TxLog1 = aesc_window:add({Key, Info}, TxLog),
```
lucafavatella: The second element in this tuple that is the entry of the `aesc_window` data structure has a `type` key here (see pattern in function clause head).

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -104,122 +243,699 @@ terminate(_Reason, _St) ->
...
+            end;
+       true ->
+            lager:debug("Vsn = undefined", []),
+            {R, C}
+    end;
+check_req(#{mode := watch} = R, #st{chan_id = ChId} = St,
+          #{scenario := Scenario, top_hash := Hash } = C)
+  when ?IS_SCENARIO(Scenario) ->
+    case Scenario of
+        {has_tx, {_, #{ type    := TxType
```
lucafavatella: This assumes that second element in tuple that is the entry of the `aesc_window` (log) data structure has a `type` key.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
lucafavatella: I noticed this via stricter type checks in https://github.com/aeternity/aeternity/pull/2488

From https://circleci.com/gh/aeternity/aeternity/40229
```
apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
 775: Record construction #tx_log_entry{key::{_,_},value::#{'block_hash':=_, 'block_origin':=_, 'tx_hash':=_, 'type':=_, _=>_}} violates the declared type of field value::#{'block_hash':=binary(), 'block_origin':='chain', 'tx_hash':=binary()}
```

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
lucafavatella: I see that only entries that satisfy filter `[{block_hash, BHash}]` end up here. That does not solve problem.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -13,46 +14,152 @@
...
+                                  , locked_until := aesc_channels:locked_until() }.
+
+-type cache() :: #{ mode := mode()
+                  , tx_log := aesc_window:window()
+                  , rpt_log := aesc_window:window()
+                  , last_block := block_hash()
+                  , chan_vsn   := chan_vsn()
+                  , block_hash => block_hash()
+                  , scenario   := scenario()
+                  , ch_status  => ch_status()
```
lucafavatella:
```suggestion
                  , {ch_status, block_hash()} => ch_status()
```
?

See the `channel_status` function.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -13,46 +14,152 @@
...
+
+-type block_hash() :: binary().
+-type tx_hash()    :: binary().
+-type scenario()   :: top | next_block | fork_switch | {has_tx, tx_hash()}.
+-type chan_vsn()   :: undefined | { aesc_channels:round()
+                                  , aesc_channels:solo_round()
+                                  , aesc_channels:is_active()
+                                  , aesc_channels:locked_until()
+                                  , aesc_channels:state_hash() }.
+-type ch_status() :: undefined | #{ is_active    := aesc_channels:is_active()
```
lucafavatella: Can undefined be a value?

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -13,46 +14,152 @@
...
+                                  , vsn          := chan_vsn()
+                                  , changed      := boolean()
+                                  , channel      := aesc_channels:channel()
+                                  , lock_period  := aesc_channels:lock_period()
+                                  , locked_until := aesc_channels:locked_until() }.
```
lucafavatella:
```suggestion
                                  , locked_until := aesc_channels:locked_until() }
                               | closed.
```

I see `{ch_status, BHash} => closed`.

lucafavatella: I see addressed at https://github.com/aeternity/aeternity/pull/2431/commits/caf3de5cd5c3d45c9d75e3fd5d19a4410f6ab737#diff-f609866532cd9838417a435634ee892aR102

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -13,46 +14,202 @@
...
+-type scenario()   :: top
+                    | next_block
+                    | fork_switch
+                    | {has_tx, info_of_scenario_has_tx()}.
+-type chan_vsn()   :: undefined | { aesc_channels:round()
+                                  , aesc_channels:solo_round()
+                                  , aesc_channels:is_active()
+                                  , aesc_channels:locked_until()
+                                  , aesc_channels:state_hash() }.
+-type ch_status() :: undefined
```
lucafavatella: I do not see this `undefined` used.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -13,46 +14,202 @@
...
+
 start_link(Type, TxHash, ChanId, MinDepth, Mod) ->
+    I = #{ parent       => self()
+         , type         => Type
+         , callback_mod => Mod },
+    Reqs = [#{ mode      => tx_hash
+             , tx_hash   => TxHash
+             , min_depth => MinDepth
+             , info      => I},
+            #{ mode  => watch
```
lucafavatella: I see the `init` function sends to itself `check_status` message and places these `Reqs` in `#st.requests`. Ok.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -63,37 +220,69 @@ init(#{parent := Parent, chan_id := ChanId, requests := Reqs}) ->
...
+             #{info := #{ block_hash := _BlockHash
+                        , tx_hash    := _TxHash
+                        , type       := _Type } = I}}, #st{chan_id = ChId} = St) ->
+    lager:debug("tx_event: I = ~p", [I]),
+    {noreply, log_tx(I, St)};
 handle_info({'DOWN', _, process, Parent, _}, #st{parent = Parent} = St) ->
     {stop, normal, St};
 handle_info(check_status, St) ->
-    {noreply, check_status(St)};
+    {noreply, check_status(top, St)};
```
lucafavatella: I see that on reception of `check_status` message the function `check_status/2` is called with `top` as first argument (later called "scenario"). Ok.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -104,173 +293,811 @@ terminate(_Reason, _St) ->
...
+    {TopHeight, C1} = top_height(C),
+    if TopHeight >= CheckAt ->
+            check_at_height(R, Reqs, St, C1, Acc);
+       true ->
+            lager:debug("Not yet (TopHeight = ~p), skip", [TopHeight]),
+            check_requests_(Reqs, St, C1, [R|Acc])
     end;
-check_status_(#{mode := close, parent := Parent} = R, #st{chan_id = ChId}, C) ->
-    lager:debug("check_status(type = close, parent = ~p)", [Parent]),
+check_requests_([Req|Reqs], St, Cache, Acc) ->
```
lucafavatella: I see requests are processed sequentially. So, coming from `init`, request `tx_hash` first then `watch`. Ok.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -104,173 +293,811 @@ terminate(_Reason, _St) ->
...
+                                            , block_hash => Hash
+                                            , height  => H }, R, St, C2);
+                _ ->
+                    {Status, C3} = channel_status(ChId, C2),
+                    lager:debug("Status = ~p", [Status]),
+                    watch_for_change_in_ch_status(Status, H, R, St, C3)
+            end;
+        fork_switch ->
+            watch_for_channel_change(R, St, C);
+        _ ->
```
lucafavatella: I see that scenario `top` gets this request `watch` discarded.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
uwiger: It would be the return value of `get_basic_ch_status_/3` before the channel object has been created. Ideally, this should never happen, since status isn't checked until the channel is active (I think).

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
uwiger: Well, the `watch` request will remain in play, but will not perform a check in a `top` scenario, which triggers only when a new request object is introduced.

lucafavatella: Ah! the watch request is not discarded but placed in `requests` field of state. Thanks.

apps/aechannel/src/aesc_fsm.erl
lucafavatella: @uwiger Thanks for having clarified that the watch request is not necessarily removed from the watcher, and that the watcher reacts to events from chain.

Not necessary to include in PR though I comment here for tracking.
```suggestion
            %% Let the watcher pick up the tx once it makes it onto the chain.
            %% The watcher was started when opening the channel.
            %% The watcher on initialization:
            %% * Subscribed to chain events.
            %% * Stored a never-done "watch" request - processed at each event and calling back to this FSM.
```

PR #2432, opened 2019-05-15
Set final BRI voting reward to 109 of 1000 shares

thepiwo:


thepiwo:
corresponding protocol PR https://github.com/aeternity/protocol/pull/355

PR #2433, opened 2019-05-15
enable reward split by default

sennui:


apps/aecore/src/aec_dev_reward.erl
```diff
@@ -16,10 +16,10 @@
 -include_lib("aecontract/include/hard_forks.hrl").
 
 
--define(ENABLED, false).
+-define(ENABLED, true).
 -define(TOTAL_SHARES, 1000). % 100 shares == 10% of the reward
 %%% for: "ak_2KAcA2Pp1nrR8Wkt3FtCkReGzAi8vJ9Snxa4PcmrthVx8AhPe8:100"
```
lucafavatella: Misleading comment. 100?

sennui:
rebasing on master

apps/aecore/src/aec_dev_reward.erl
sennui: After the rebase on master comments are fine

PR #2435, opened 2019-05-15
Introduce gas cap for GAMetaTx authentication function

hanssv:

PR #2436, opened 2019-05-16
Introduce mempool gas cap for GAMetaTx authentication function

hanssv:
Not part of consensus, instead checking this when pushing to the mempool.


hanssv:
This is the non-consensus breaking version of #2435

PR #2437, opened 2019-05-16
Make sure swagger.yaml is correct w.r.t integers

hanssv:
~~Still needs the proper commit hash from `swagger_endpoints` repo...~~

### DIFF
--- a/apps/aehttp/priv/swagger.json
+++ b/apps/aehttp/priv/swagger.json
@@ -109,8 +109,7 @@
               "type" : "object",
               "properties" : {
                 "hash" : {
-                  "type" : "string",
-                  "description" : "Key block hash"
+                  "$ref" : "#/definitions/EncodedHash"
                 }
               }
             }
@@ -138,9 +137,7 @@
               "type" : "object",
               "properties" : {
                 "height" : {
-                  "type" : "integer",
-                  "format" : "int64",
-                  "description" : "Height"
+                  "$ref" : "#/definitions/UInt64"
                 }
               }
             }
@@ -227,9 +224,10 @@
         "parameters" : [ {
           "name" : "height",
           "in" : "path",
-          "description" : "The height of the block",
+          "description" : "The height",
           "required" : true,
           "type" : "integer",
+          "maximum" : 18446744073709551615,
           "minimum" : 0
         } ],
         "responses" : {
@@ -409,10 +407,8 @@
               "type" : "object",
               "properties" : {
                 "count" : {
-                  "type" : "integer",
-                  "format" : "int64",
                   "description" : "Count",
-                  "minimum" : 1
+                  "$ref" : "#/definitions/UInt32"
                 }
               }
             }
@@ -499,9 +495,10 @@
         "parameters" : [ {
           "name" : "height",
           "in" : "path",
-          "description" : "The height of the generation",
+          "description" : "The height",
           "required" : true,
           "type" : "integer",
+          "maximum" : 18446744073709551615,
           "minimum" : 0
         } ],
         "responses" : {
@@ -570,9 +567,10 @@
         }, {
           "name" : "height",
           "in" : "path",
-          "description" : "The height of the key block",
+          "description" : "The height",
           "required" : true,
           "type" : "integer",
+          "maximum" : 18446744073709551615,
           "minimum" : 0
         } ],
         "responses" : {
@@ -1864,7 +1862,8 @@
           "in" : "query",
           "description" : "Salt",
           "required" : true,
-          "type" : "integer"
+          "type" : "integer",
+          "minimum" : 0
         } ],
         "responses" : {
           "200" : {
@@ -1992,9 +1991,10 @@
         "parameters" : [ {
           "name" : "height",
           "in" : "path",
-          "description" : "The key block height after which the tokens are summed",
+          "description" : "The height",
           "required" : true,
           "type" : "integer",
+          "maximum" : 18446744073709551615,
           "minimum" : 0
         } ],
         "responses" : {
@@ -2015,6 +2015,30 @@
     }
   },
   "definitions" : {
+    "UInt" : {
+      "type" : "integer",
+      "minimum" : 0
+    },
+    "UInt16" : {
+      "type" : "integer",
+      "minimum" : 0,
+      "maximum" : 65535
+    },
+    "UInt32" : {
+      "type" : "integer",
+      "minimum" : 0,
+      "maximum" : 4294967295
+    },
+    "UInt64" : {
+      "type" : "integer",
+      "minimum" : 0,
+      "maximum" : 18446744073709551615
+    },
+    "TxBlockHeight" : {
+      "type" : "integer",
+      "minimum" : -1,
+      "maximum" : 18446744073709551615
+    },
     "EncodedHash" : {
       "type" : "string",
       "description" : "Base58Check encoded tagged hash"
@@ -2034,14 +2058,10 @@
     "Pow" : {
       "type" : "array",
       "items" : {
-        "type" : "integer",
-        "format" : "int32"
+        "$ref" : "#/definitions/UInt32"
       },
       "minItems" : 42,
       "maxItems" : 42
-    },
-    "Uri" : {
-      "type" : "string"
     },
     "KeyBlock" : {
       "type" : "object",
@@ -2051,8 +2071,7 @@
           "$ref" : "#/definitions/EncodedHash"
         },
         "height" : {
-          "type" : "integer",
-          "format" : "int64"
+          "$ref" : "#/definitions/UInt64"
         },
         "prev_hash" : {
           "$ref" : "#/definitions/EncodedHash"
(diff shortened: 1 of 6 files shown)


apps/aehttp/priv/swagger.json
```diff
@@ -409,10 +407,8 @@
               "type" : "object",
               "properties" : {
                 "count" : {
-                  "type" : "integer",
-                  "format" : "int64",
                   "description" : "Count",
-                  "minimum" : 1
+                  "$ref" : "#/definitions/UInt32"
```
velzevur: I see this now allows the value of 0. OK

hanssv: We allow block height to be -1 even though we can't have blocks at height 0 even 😛

velzevur: This is a special block height

hanssv: No, it is a hack...

rebar.config
```diff
@@ -97,7 +97,7 @@
 
        ]}.
 
-{plugins, [{swagger_endpoints, {git, "https://github.com/aeternity/swagger_endpoints", {ref, "ac38525ba55e8eefc00fb4fc0ec697ec3b2c26cf"}}}]}.
+{plugins, [{swagger_endpoints, {git, "https://github.com/aeternity/swagger_endpoints", {ref, "37d3f03"}}}]}.
```
lucafavatella: I see this is https://github.com/aeternity/swagger_endpoints/pull/8

config/swagger.yaml
```diff
@@ -1667,51 +1660,104 @@ paths:
...
+    minimum: 0
+
 ##
 ## Definitions
 ##
 
 definitions:
+  # Swagger doesn't have any proper integer types, we mainly use four
+  # different unsigned integers
+  #  * UInt - unbound unsigned ingegers
```
lucafavatella:
```suggestion
  #  * UInt - unbound unsigned integers
```

config/swagger.yaml
```diff
@@ -1667,51 +1660,104 @@ paths:
...
+  #  * UInt16 - 16-bit unsigned ingegers
```
lucafavatella:
```suggestion
  #  * UInt16 - 16-bit unsigned integers
```

config/swagger.yaml
```diff
@@ -1667,51 +1660,104 @@ paths:
...
+  #  * UInt32 - 32-bit unsigned ingegers
```
lucafavatella:
```suggestion
  #  * UInt32 - 32-bit unsigned integers
```

config/swagger.yaml
```diff
@@ -1667,51 +1660,104 @@ paths:
...
+  #  * UInt64 - 64-bit unsigned ingegers
```
lucafavatella:
```suggestion
  #  * UInt64 - 64-bit unsigned integers
```

config/swagger.yaml
```diff
@@ -1667,51 +1660,104 @@ paths:
...
+  UInt:
+    type: integer
+    minimum: 0
+  UInt16:
+    type: integer
+    minimum: 0
```
lucafavatella: I see exclusiveMinimum defaults to false. Ok.

config/swagger.yaml
```diff
@@ -1667,51 +1660,104 @@ paths:
...
+    maximum: 65535
```
lucafavatella: I see exclusiveMaximum defaults to false. Ok.

config/swagger.yaml
```diff
@@ -3305,37 +3133,21 @@ definitions:
...
+            $ref: '#/definitions/UInt'
           call_data:
             description: Contract call data
             $ref: '#/definitions/EncodedByteArray'
         required:
           - caller
           - contract
-          # should be required once vm_version is removed
-          # - abi_version
+          - abi_version
```
lucafavatella: I see this covered by https://github.com/aeternity/aeternity/blob/1c6a34e0c4f940ce3c9d10ac0f465b9b07ecc57f/docs/release-notes/next-fortuna/PT-164962293-no_more_backwards_compat_abi_and_vm.md
Ok.

config/swagger.yaml
```diff
@@ -92,9 +92,7 @@ paths:
             type: object
```
lucafavatella: I see that `'200': ...` is meant to be a [response object](https://github.com/OAI/OpenAPI-Specification/blob/8a387ac71f7f31aaff5b910f1e0e99ad12ff3ef0/versions/2.0.md#responsesObject). Ok.

lucafavatella: In particular, see [patterned fields in response object](https://github.com/OAI/OpenAPI-Specification/blob/8a387ac71f7f31aaff5b910f1e0e99ad12ff3ef0/versions/2.0.md#patterned-fields-3).

lucafavatella: I see a single response is defined [here](https://github.com/OAI/OpenAPI-Specification/blob/8a387ac71f7f31aaff5b910f1e0e99ad12ff3ef0/versions/2.0.md#responseObject): it defines fields `description`, `schema`, `headers`, `examples`.

Here, description and schema are used. Ok.

config/swagger.yaml
```diff
@@ -92,9 +92,7 @@ paths:
...
             properties:
               hash:
-                description: 'Key block hash'
-                # Encoded key block hash 'kh_...'
-                type: string
+                $ref: '#/definitions/EncodedHash'
```
lucafavatella: The schema object is defined [here](https://github.com/OAI/OpenAPI-Specification/blob/8a387ac71f7f31aaff5b910f1e0e99ad12ff3ef0/versions/2.0.md#schemaObject).

It defines `properties` as object where each value is a schema object.

In particular, this follows [`address` in this example](https://github.com/OAI/OpenAPI-Specification/blob/8a387ac71f7f31aaff5b910f1e0e99ad12ff3ef0/versions/2.0.md#simple-model).

Ok.

config/swagger.yaml
```diff
@@ -183,12 +180,7 @@ paths:
       produces:
         - application/json
       parameters:
-        - in: path
-          name: height
-          description: 'The height of the block'
-          required: true
-          type : integer
-          minimum: 0
+        - $ref: '#/parameters/HeightIn'
```
lucafavatella: The path object is defined [here](https://github.com/OAI/OpenAPI-Specification/blob/8a387ac71f7f31aaff5b910f1e0e99ad12ff3ef0/versions/2.0.md#pathItemObject).

`parameters` is defined as `[Parameter Object | Reference Object]`. So this can be a ref. Ok.

config/swagger.yaml
```diff
@@ -339,10 +335,8 @@ paths:
             type: object
             properties:
               count:
-                type: integer
-                minimum: 1
                 description: 'Count'
-                format: int64
+                $ref: '#/definitions/UInt32'
```
lucafavatella: A path item object is defined [here](https://github.com/OAI/OpenAPI-Specification/blob/8a387ac71f7f31aaff5b910f1e0e99ad12ff3ef0/versions/2.0.md#pathItemObject).

For get operation, the responses object is defined [here](https://github.com/OAI/OpenAPI-Specification/blob/8a387ac71f7f31aaff5b910f1e0e99ad12ff3ef0/versions/2.0.md#responsesObject).

For a specific response code, the value can be `Response Object | Reference Object`. For this 200, it is response object.

The response object defines value of field schema as [schema object](https://github.com/OAI/OpenAPI-Specification/blob/8a387ac71f7f31aaff5b910f1e0e99ad12ff3ef0/versions/2.0.md#schemaObject).

Schema object defines value `properties`. The value of each property is a schema object.

The schema object (namely - the value of object field `count`) can have fields `$ref`, `description`, ... /me thinking of semantics of `$ref`, and how it interacts of other fields 🤔

hanssv: If you have a `$ref` field all other fields are ignored if I remember correctly... So the `description` is really just for documentation...

lucafavatella: In [definition of schema object](https://github.com/OAI/OpenAPI-Specification/blob/8a387ac71f7f31aaff5b910f1e0e99ad12ff3ef0/versions/2.0.md#schemaObject) I read:
> The following properties are taken directly from the JSON Schema definition and follow the same specifications:
> * $ref - As a [JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03)
> * ...
> * description

In [linked JSON reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03#section-3) I read:
> A JSON Reference is a JSON object, which contains a member named "$ref", which has a JSON string value.  Example:
>    { "$ref": "http://example.com/example.json#/foo/bar" }
> If a JSON value does not have these characteristics, then it SHOULD NOT be interpreted as a JSON Reference.
> The "$ref" string value contains a URI [RFC3986], which identifies the location of the JSON value being referenced.  It is an error condition if the string value does not conform to URI syntax rules. Any members other than "$ref" in a JSON Reference object SHALL be ignored.

So, according to the linked JSON referce a `"$ref"` property makes any other members in the same object were it is to-be-ignored.

In the case, the member to-be-ignored is a description. In cases where more structural members are to-be-ignored, I would not be surprised if there were surprises depending on the tool.

velzevur: Also - parsed by swagger JSON builder, as seen [here](https://github.com/aeternity/aeternity/blob/2f1890932cbd3b9225f46a0fe7846875adebc7c4/apps/aehttp/priv/swagger.json#L410)

apps/aehttp/src/aehttp_api_validate.erl
```diff
@@ -46,11 +46,24 @@ response(OperationId, Method0, Code, Response, Validator) ->
...
+            Schema = fix_def_refs(Schema0),
             _ = jesse_schema_validator:validate_with_state(Schema, Response, Validator),
             ok
     end.
 
+fix_def_refs(Map) ->
+    fix_def_refs(unused, Map).
+
+fix_def_refs(_, Map) when is_map(Map) ->
+    %% Recursively fix refs (and drop extra, to be ignored, fields since jesse is broken)
```
lucafavatella:
```suggestion
    %% Recursively fix refs (and [drop extra, to be ignored, fields](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03#section-3) since jesse is broken)
```

lucafavatella:
I see http://aeternity.com/api-docs/?config=https://raw.githubusercontent.com/aeternity/epoch/a9c5b7220357a0a7cef6c936cf76bb7d434dbc70/apps/aehttp/priv/swagger.json not showing warnings/errors. Ok.

hanssv:
https://circleci.com/workflow-run/4bb8b936-d0f7-4398-aa83-f9bdee6d82af succesful tests

PR #2445, opened 2019-05-17
Update doc re Docker testnet config

lucafavatella:


docs/docker.md
```diff
@@ -109,7 +109,7 @@ For more information see [miner configuration documentation](configuration#miner
 
 ## Testnet
 
-The minimal configuration to join the testnet needs a list of seed peers and network identifier:
+The minimal configuration to join the testnet needs a list of seed peers, network identifier and block reward beneficiary:
```
dincho: And number of shares ? It also under consensus right?

lucafavatella: Addressed in new commit.

PR #2448, opened 2019-05-20
Make sync.progress.value metric follow sync_progress API

hanssv:
Solves https://www.pivotaltracker.com/story/show/166029931


hanssv:
Not sure this interesting enough to write a release note, users will access the value through HTTP API and that hasn't changed?

PR #2449, opened 2019-05-20
Merge tag v3.0.0 rc.1+windows to master

lucafavatella:


lucafavatella:
There are lots of jobs that are part of the tag - as this commit is both branch and tag - and these jobs are not meant to complete. So I am merging without green CI.

PR #2452, opened 2019-05-20
Set proper value for not set pinned block hash

velzevur:
No PT, setting a proper empty value for when the block hash has not been pinned off-chain. This is a placeholder to be used later on with enriching the API with proper pinning.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -98,7 +98,7 @@
                   | ?SHUTDOWN
                   | ?SHUTDOWN_ACK.
 
--define(DUMMY_BLOCK_HASH, <<12345:32/unit:8>>).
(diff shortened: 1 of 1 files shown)


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -98,7 +98,7 @@
                   | ?SHUTDOWN
                   | ?SHUTDOWN_ACK.
 
--define(DUMMY_BLOCK_HASH, <<12345:32/unit:8>>).
+-define(NOT_SET_BLOCK_HASH, <<0:32/unit:8>>).
```
lucafavatella: I see this value used for matching.

Is this field persisted from requests produced by node `v2.*`?

lucafavatella: @velzevur Thanks for pointing out this was not released in stable release yet. 🤦‍♂ So good.

PR #2454, opened 2019-05-20
Persist DB by default

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/165168390

TODO:
* [x] Update docs

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -62,14 +62,8 @@ More details about node configuration can be found in [configuration documentati
 
 ### Persisting Data
 
-To persist blockchain data between container runs use [Docker volumes](https://docs.docker.com/engine/admin/volumes/volumes/) and enable chain persistence in your configuration:
-
-```yaml
-# ... SNIP
-chain:
-    persist: true
-# SNIP ...
-```
+The blockchain data is persisted by default, inside the Docker container.
+In order to persist the data in a directory on the host machine, use [Docker volumes](https://docs.docker.com/engine/admin/volumes/volumes/).
 
 Assuming your configuration file path is `~/.aeternity/myaeternity.yaml` on host machine.
 Replace `~/.aeternity/myaedb` with location of your choice where the data will be stored in.
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -62,7 +62,7 @@ More details about node configuration can be found in [configuration documentati
 
 ### Persisting Data
 
-The blockchain data is persisted by default, inside the Docker container.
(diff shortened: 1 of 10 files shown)


docs/docker.md
```diff
@@ -62,14 +62,8 @@ More details about node configuration can be found in [configuration documentati
...
-```yaml
-# ... SNIP
-chain:
-    persist: true
-# SNIP ...
-```
+The blockchain data is persisted by default, inside the Docker container.
+In order to persist the data in a directory on the host machine, use [Docker volumes](https://docs.docker.com/engine/admin/volumes/volumes/).
 
 Assuming your configuration file path is `~/.aeternity/myaeternity.yaml` on host machine.
```
dincho: This line could be removed as well, also the config mount in the command below.

docs/docker.md
```diff
@@ -62,14 +62,8 @@ More details about node configuration can be found in [configuration documentati
...
 
-To persist blockchain data between container runs use [Docker volumes](https://docs.docker.com/engine/admin/volumes/volumes/) and enable chain persistence in your configuration:
-
-```yaml
-# ... SNIP
-chain:
-    persist: true
-# SNIP ...
-```
+The blockchain data is persisted by default, inside the Docker container.
```
dincho: It's worth nothing mentioning that by default the database is persisted at: `/home/aeternity/node/data/mnesia` in the container.

lucafavatella: I expect it benefits Docker image user re which parameters to tune when starting container. User may be surprised to find that the new container fails after a while if container started with large memory but little storage.

docs/docker.md
lucafavatella: You are right. Amending.

lucafavatella: Addressed in 369af4090

docs/docker.md
```diff
@@ -62,22 +62,14 @@ More details about node configuration can be found in [configuration documentati
 
 ### Persisting Data
 
-To persist blockchain data between container runs use [Docker volumes](https://docs.docker.com/engine/admin/volumes/volumes/) and enable chain persistence in your configuration:
+The blockchain data is persisted by default, inside the Docker container.
```
dincho:
```suggestion
The blockchain data is persisted by default at `/home/aeternity/node/data/mnesia`, inside the Docker container.
```

docs/docker.md
lucafavatella: (I had not got initial comment, you suggested change now included. 👍 )

PR #2457, opened 2019-05-22
Consider ping:s successful during sync

hanssv:


apps/aecore/src/aec_sync.erl
```diff
@@ -698,8 +698,7 @@ do_work_on_sync_task(PeerId, Task, LastResult) ->
                 end,
             do_work_on_sync_task(PeerId, Task, Res);
         abort_work ->
-            epoch_sync:info("~p aborting sync work against ~p", [self(), PeerId]),
-            schedule_ping(PeerId)
+            epoch_sync:info("~p aborting sync work against ~p", [self(), PeerId])
```
ThomasArts: So, if I understand it well, when we abort a sync task, we are not automatically going to try again against that node. It may happen as a side effect of something else, but not enforced by our node.

We abort when we cannot find the peer in any of our sync tasks.

Is there a risk that when we're done synchonizing, that we fail to restart a new sync in time and need to be triggered by others to sync with us to get sync started again?

hanssv: My idea was that we will re-try this one anyway, just not immediately...

Since we no longer stop ping:ing when synchronizing due to the other change we will keep ping:ing also the node that fail here?!

PR #2458, opened 2019-05-22
Fix Windows packaging

venimus:
* Fix Windows packaging missplacing of VERSION and REVISION files
* Enhance smoke test with checks to validate version and revision returned by /status endpoint
https://www.pivotaltracker.com/n/projects/2152594/stories/166118490


lucafavatella:
This seems in scope of https://www.pivotaltracker.com/story/show/166118490 ?

dincho:
> This seems in scope of https://www.pivotaltracker.com/story/show/166118490 ?

Yes, but with additional fix included

PR #2459, opened 2019-05-22
Fix loading of FP transactions from old Database format #166184378

velzevur:
PT [Bad function clause match causing process crash](https://www.pivotaltracker.com/story/show/166184378)

### DIFF
--- a/apps/aechannel/src/aesc_force_progress_tx.erl
+++ b/apps/aechannel/src/aesc_force_progress_tx.erl
@@ -69,24 +69,27 @@
 
 %% Record introduced temporarily during Fortuna development, shipped in releases 2.4.0 and 2.5.0, and potentially stored in DB.
 -record(v2_db_record, {	
-          channel_id    :: aeser_id:id(),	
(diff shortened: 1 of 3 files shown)


apps/aechannel/src/aesc_force_progress_tx.erl
```diff
@@ -69,24 +69,27 @@
...
 %%%===================================================================
 %%% Conversion of old db format
 
 -spec from_db_format(tx()) -> tx().
-from_db_format(#channel_force_progress_tx{} = Tx) ->
-    Tx;
+from_db_format(#channel_force_progress_tx{update = Update} = Tx) ->
+    case aesc_offchain_update:from_db_format(Update) of
+        Update -> Tx; % update is already in new serialization
+        NewSUpdate -> Tx#channel_force_progress_tx{update = NewSUpdate}
```
velzevur: Actual fix

lucafavatella:
I see in scope of https://www.pivotaltracker.com/story/show/166184378

PR #2460, opened 2019-05-22
Adapt system tests to JSON-RPC API

velzevur:
[System test aest_channels_SUITE failing](https://www.pivotaltracker.com/story/show/166155618)


lucafavatella:
The enhanced system test re state channels is not hooked to CI for branch so it is not included in outcome of CI of this PR. What it could do is breaking compilation of tests of `docker_system_smoke_tests` but that passed.

PR #2469, opened 2019-05-27
Fix AppVeyor build failures

tolbrino:
`pacman` breaks during AppVeyor preparation, leaving some important deps uninstalled and breaking the build. This change tries to fix the `pacman` upgrade procedure.

### DIFF
--- a/scripts/windows/msys2_prepare.bat
+++ b/scripts/windows/msys2_prepare.bat
@@ -90,9 +90,7 @@ rem Remove breaking tools
 @echo Current time: %time%
 rem Upgrade the MSYS2 platform
 
-%BASH_BIN% -lc "%PACMAN% -y pacman"
-@echo Current time: %time%
-%BASH_BIN% -lc "%PACMAN% -u"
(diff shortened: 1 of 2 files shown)


tolbrino:
@lucafavatella @dincho This change will unblock the Appveyor builds.

scripts/windows/msys2_prepare.bat
```diff
@@ -90,9 +90,7 @@ rem Remove breaking tools
 @echo Current time: %time%
 rem Upgrade the MSYS2 platform
 
-%BASH_BIN% -lc "%PACMAN% -y pacman"
-@echo Current time: %time%
-%BASH_BIN% -lc "%PACMAN% -u"
+%BASH_BIN% -lc "%PACMAN% -yuu"
```
dincho: is the second `u` what's expected to fix the issue ?

hanssv: Magic!

tolbrino: Yes. It performs a complete system upgrade instead of a single package upgrade.

PR #2471, opened 2019-05-27
Integrate new secrets scheme

dincho:


dincho:
Windows builds are failing because of enacl

PR #2476, opened 2019-05-28
Changed TUPLE instruction for FATE

UlfNorell:
~~Doesn't pass testing yet since it generates deep types for variants instead of just constructor arities. Waiting for PR by @gorillainduction to make that work.~~
All good now.


gorillainduction:
Rebased on master to fix CI problems

PR #2481, opened 2019-05-31
Add missing context to docker_system_tests CI job

dincho:
https://www.pivotaltracker.com/story/show/166359026

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1453,4 +1453,5 @@ workflows:
               only:
                 - master
     jobs:
-      - docker_system_tests
+      - docker_system_tests:
+          context: ae-node-builds


.circleci/config.yml
```diff
@@ -1453,4 +1453,5 @@ workflows:
               only:
                 - master
     jobs:
-      - docker_system_tests
+      - docker_system_tests:
```
lucafavatella: This looks the only scheduled workflow. Will this work? Maybe to be moved to other org?

dincho: 🤔

dincho:
@lucafavatella good point, I added the variables to the old non-restricted environment configuration

PR #2484, opened 2019-06-04
Introduce a ping-pong mechanism in WebSocket data messages

velzevur:
[SC WS: ping-pong events](https://www.pivotaltracker.com/story/show/166433992)

WebSocket protocol does not detect abrupt disconnects so this must be implemented on both sides. The node timeouts in 1 minute if no messages had been sent or received and is protected from leaking connections. In order for the client to be safe it will send ping messages and will receive pongs.

For browser compatibility, the ping-pong mechanism is implemented via data frames..

### DIFF
--- a/apps/aehttp/src/sc_ws_api.erl
+++ b/apps/aehttp/src/sc_ws_api.erl
@@ -4,7 +4,8 @@
 
 %TODO type -> opaque
 -opaque protocol() :: jsonrpc.
--type response() :: {reply, map()} | no_reply | stop.
+-type reply() :: {{error, atom()}, map()} | map().
+-type response() :: {reply, reply() | [reply()]} | no_reply | stop.
 
 -export_type([protocol/0,
               response/0]).
@@ -37,8 +38,7 @@
     {reply, map()}.
 
 -callback process_incoming(Msg :: map() | list(map()), FsmPid :: pid()) ->
-    {reply, map()} | no_reply | stop.
-
+    response().
 
 %%%==================================================================
 %%% Trace settings
--- a/apps/aehttp/src/sc_ws_api_jsonrpc.erl
+++ b/apps/aehttp/src/sc_ws_api_jsonrpc.erl
@@ -234,6 +234,9 @@ process_incoming(Msg, FsmPid) ->
     end.
 
 -spec process_request(map(), pid()) -> no_reply | {reply, map()} | {error, term()}.
+process_request(#{<<"method">> := <<"channels.system">>,
+                  <<"params">> := #{<<"action">> := <<"ping">>}}, _FsmPid) ->
+    {reply, #{action => system, tag => pong}};
 process_request(#{<<"method">> := <<"channels.update.new">>,
                    <<"params">> := #{<<"from">>    := FromB,
                                      <<"to">>      := ToB,
--- a/apps/aehttp/test/aehttp_integration_SUITE.erl
+++ b/apps/aehttp/test/aehttp_integration_SUITE.erl
@@ -211,6 +211,7 @@
     sc_ws_update_conflict/1,
     sc_ws_close_mutual/1,
     sc_ws_leave_reestablish/1,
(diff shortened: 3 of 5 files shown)


apps/aehttp/src/sc_ws_api_jsonrpc.erl
```diff
@@ -234,6 +234,9 @@ process_incoming(Msg, FsmPid) ->
     end.
 
 -spec process_request(map(), pid()) -> no_reply | {reply, map()} | {error, term()}.
+process_request(#{<<"method">> := <<"channels.system">>,
```
lucafavatella: It seems requires doc at https://github.com/aeternity/protocol/blob/8b2724d14d8b4878b6b2c74850ec3f7f7716aeb2/node/api/channel_ws_api.md ?

lucafavatella:
Worth a line in release notes?

apps/aehttp/src/sc_ws_api_jsonrpc.erl
```diff
@@ -234,6 +234,9 @@ process_incoming(Msg, FsmPid) ->
...
+                  <<"params">> := #{<<"action">> := <<"ping">>}}, _FsmPid) ->
+    {reply, #{action => system, tag => pong}};
```
lucafavatella: What would be the behaviour from client side if client sent this before this code change?

I see that:
* `process_request /1` would return `{error, unhandled}`.
* `process_incoming/2` would therefore return `{reply, [{{error, unhandled}, #{<<"method">> := <<"channels.system">>, ...}}]}`. Though this callback is defined in behaviour `sc_ws_api` as [returning `{reply, map()} | no_reply | stop`](https://github.com/aeternity/aeternity/blob/de28b27c0fb627aadeaed42711bf4d8abcf7974a/apps/aehttp/src/sc_ws_api.erl#L40).

Could you please clarify behaviour on reception of this message before this change. And if possible also clarify return type of callback.

velzevur: Callback return value is ddressed.
If the ping message is received by a node without this change being applied, the node responds with
```
{  
   'channel_id':'ch_zVDx935M1AogqZrNmn8keST2jH8uvn5kmWwtDqefYXvgcCRAX',
   'error':{  
      'code':-32601,
      'message':'Method not found',
      'request':{  
         'jsonrpc':'2.0',
         'method':'channels.system',
         'params':{  
            'action':'ping'
         }
      }
   },
   'id':null,
   'jsonrpc':'2.0',
   'version':1
}```

Comment (lucafavatella) [reply apps/aehttp/src/sc_ws_api_jsonrpc.erl]: From https://www.jsonrpc.org/specification#error_object
> -32601 | Method not found ...

Ok.

File: apps/aehttp/src/sc_ws_api.erl [hunk 3]
```diff
@@ -4,7 +4,8 @@
 
 %TODO type -> opaque
 -opaque protocol() :: jsonrpc.
--type response() :: {reply, map()} | no_reply | stop.
+-type reply() :: {{error, atom()}, map()} | map().
```
lucafavatella: Is this confusing between `sc_ws_api_jsonrpc:process_request` and `sc_ws_api_jsonrpc:process_incoming`?

lucafavatella: Apologies - I had somehow missed the ` | [reply()]` in the patch. Correct.

uwiger:
What about instead using PING frames?
https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers#Pings_and_Pongs_The_Heartbeat_of_WebSockets

velzevur:
For better traceability, the reasoning behind the not relying on the control frames `ping` and `pong` can be found in [the protocol](https://github.com/aeternity/protocol/blob/master/node/api/channels_api_usage.md#connection-keep-alive)

PR #2491, opened 2019-06-05
Enhance HTTP endpoint for channel creation transaction to accept opti…

velzevur:
…onal list of delegates
[SC HTTP add delegates](https://www.pivotaltracker.com/story/show/166463891)


docs/release-notes/next/PT-166463891.md
```diff
@@ -0,0 +1 @@
+* Enhances the HTTP endpoint for channel creation transaction to accept an optional list of delegates
```
lucafavatella: I see not backward incompatible.

PR #2492, opened 2019-06-05
PT-166407568 polymorphic functions in FATE

UlfNorell:
[PT-166407568](https://www.pivotaltracker.com/story/show/166407568)

### DIFF
--- a/apps/aefate/src/aefa_fate.erl
+++ b/apps/aefate/src/aefa_fate.erl
@@ -313,53 +313,85 @@ check_arg_types([T|Ts], [A|As]) ->
         false -> {error, T, A}
     end.
 
-check_all_types([], []) -> true;
-check_all_types([T|Ts], [A|As]) ->
-    check_type(T, A) andalso  check_all_types(Ts, As).
-
-check_same_type(_, []) -> true;
-check_same_type(T, [A|As]) ->
-    check_type(T, A) andalso check_same_type(T, As).
-
 bind_args(N, _, [], Mem, EngineState) ->
     ES1 = drop(N, EngineState),
     aefa_engine_state:set_memory(Mem, ES1);
 bind_args(N, [Arg|Args], [_Type|Types], Mem, EngineState) ->
     bind_args(N+1, Args, Types, Mem#{{arg, N} => {val, Arg}}, EngineState).
 
-%% TODO: Add types (and tests).
-check_type(any, _) -> true;
-check_type(_, any) -> true;
-check_type(integer, I) when ?IS_FATE_INTEGER(I) -> true;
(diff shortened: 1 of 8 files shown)


apps/aefate/src/aefa_fate.erl
```diff
@@ -306,60 +315,118 @@ bind_args_from_signature({ArgTypes, _RetSignature}, ES) ->
...
-
-check_variant_arities([?FATE_TUPLE(Types)|Left1], [Arity|Left2]) ->
-    Arity =:= length(Types) andalso check_variant_arities(Left1, Left2);
-check_variant_arities([], []) ->
-    true;
-check_variant_arities(_, _) ->
-    false.
+check_type(T, V) ->
+    try
+        match_type(T, infer_type(V))
```
lucafavatella: I see call e.g. `aefa_fate:check_type(integer, Index)` hooked to VM (though not in Fortuna).

Even if this form (i.e. infer type, then match) is more readable than the old one (match, inferring only as much as needed), I wonder if this makes more difficult to price VM operations, and potentially enabling spending little (static) though causing node to spend large computational effort.

apps/aefate/src/aefa_fate.erl
```diff
@@ -306,60 +315,118 @@ bind_args_from_signature({ArgTypes, _RetSignature}, ES) ->
...
+    N    = length(Arities),
+    [ throw({bad_variant_tag, Tag, Arities}) || Tag >= N ],
+    Type = infer_type(?FATE_TUPLE(Value)),
+    %% Partial variant type (only arities for other constructors)
+    {variant, [ if I == Tag -> Type; true -> A end
+                || {A, I} <- lists:zip(Arities, lists:seq(0, N - 1))]};
+infer_type(X) -> throw({not_a_fate_value, X}).
+
+infer_element_type(Xs) ->
+    lists:foldl(fun intersect_types/2, any, [infer_type(X) || X <- Xs]).
```
lucafavatella: I understand intersect, and not union, because union type is expressed as variant type, and value ov variant type needs to be expressed [as constructor](https://github.com/aeternity/protocol/blame/34e8eeee3c9d3c7fafa83883bc7d95273315125d/contracts/sophia.md#L1282-L1287).

apps/aefate/src/aefa_fate.erl
UlfNorell: `check_type` is not exported from the module.

Type inference and type checking have similar cost when successful. Inference will be more expensive to check a large structure against `integer` (say), but you would still have to pay for the size of the structure (and you have the same problem for checking if the type is not `integer` but some structured type)

PR #2493, opened 2019-06-05
code changes due to stricter types

uwiger:


apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -104,9 +105,10 @@
                                   , locked_until := aesc_channels:locked_until() }.
 
 -type tx_log_entry_key() :: {tx_hash(), block_hash()}.
--type tx_log_entry_value() :: #{ tx_hash := tx_hash()
-                               , block_hash := block_hash()
-                               , block_origin := chain }.
+-type tx_log_entry_value() :: #{ tx           => aetx_sign:signed_tx()
```
uwiger: A note here: when we log a tx_event, we don't have the full SignedTx from the start, and we don't necessarily want to fetch it right away. So it's added on-demand when `get_latest_tx()` is called.

uwiger: Also, no need to store the `tx_hash` in the `value` part, since it's already part of the key. We do want the full tx, though, when it's time to report something.

lucafavatella:
I see this PR follows https://github.com/aeternity/aeternity/pull/2431#pullrequestreview-245878572
Ok.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -27,7 +27,8 @@
                     andalso is_binary(element(2, element(1, X)))
                   )
           andalso ( is_map(element(2, X))
-                    andalso (map_size(element(2, X)) =:= 3)
+                    andalso (map_size(element(2, X)) =:= 3
```
lucafavatella: I see tx_hash removed though type added. Ok.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -27,7 +27,8 @@
...
+                             orelse map_size(element(2, X)) =:= 4)
```
lucafavatella: I see tx optional. Ok.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -510,9 +512,9 @@ check_req(#{mode := unlock} = R, #st{chan_id = ChId, chan_vsn = Vsn} = St, C) ->
 check_req(#{mode := watch} = R, #st{chan_id = ChId} = St,
           #{scenario := Scenario, top_hash := Hash } = C)
   when ?IS_SCENARIO(Scenario) ->
+    lager:debug("Scenario = ~p", [Scenario]),
     case Scenario of
-        {has_tx, {_, #{ type    := TxType
-                      , tx_hash := TxHash }}} ->
+        {has_tx, {{TxHash,_}, #{ type    := TxType }}} ->
```
lucafavatella: This makes sense. Ok.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -702,15 +705,16 @@ handle_ch_status_changed(ChId, #{ last_block := Last
     lager:debug("channel status changed, Hash=~p, Last=~p", [Hash,Last]),
     {TxLog, C1} = get_txs_since({any_after_block, Last}, Hash, ChId, C),
     lager:debug("New TxLog: ~p", [TxLog]),
-    #{tx_hash := Tx, block_hash := BlockHash} = get_latest_tx(TxLog),
-    lager:debug("Latest Tx = ~p, Hash = ~p", [Tx, BlockHash]),
-    {Status, C2} = get_basic_ch_status_(ChId, BlockHash, C1),
+    {TxHash, #{tx := Tx, block_hash := BlockHash}, C2} =
```
lucafavatella: I need to understand how this optional `tx` field is guaranteed to be present here 🤔

uwiger: See `ensure_signed_tx_included/2`, called from line 793.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -772,18 +776,31 @@ log_tx(#{ tx_hash      := TxHash
...
+    lager:debug("TxInfo = ~p", [TxInfo]),
+    {TxInfo1, C1} = ensure_signed_tx_included(TxHash, TxInfo, C),
+    {TxHash, TxInfo1, C1}.
+
+ensure_signed_tx_included(TxHash, TxInfo, C) ->
+    case maps:is_key(tx, TxInfo) of
+        true ->
+            {TxInfo, C};
+        false ->
+            {SignedTx, C1} = get_signed_tx(TxHash, C),
```
lucafavatella: I am trying to understand if `undefined` here is accepted by the code using this function.

lucafavatella: If there is assumption on fetching tx not returning undefined, I would expect that expectation hardcoded.

uwiger: That would be a very strange occurrence, but the most reasonable action in such a case might be to simply remove that TxHash from the TxLog and pick the next one. But I'm not really sure what exactly it would mean at this point.

lucafavatella: Would it make sense in that (unlikely? impossible?) case to cause runtime error here, rather than cause invalid value to end up in state and then cause error down the line, maybe in separate event, making it harder to identify cause?

Or at least logging it as warning?

lucafavatella: or a comment...

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
```diff
@@ -978,18 +998,23 @@ tx_location(TxHash, C) ->
...
 
 tx_location_(TxHash, C) ->
-    L = case aec_chain:find_tx_with_location(TxHash) of
+    {L, Type, SignedTx} =
+        case aec_chain:find_tx_with_location(TxHash) of
             none ->
                 lager:debug("couldn't find tx hash", []),
-                undefined;
-            {mempool, _} ->
+                {undefined, undefined, undefined};
```
lucafavatella: So the third element in this tuple means that `undefined` can be stored as `tx` in the map in `#tx_log_entry.value`? Is this supported flow? If yes, type of field `tx` shall reflect this (I currently see `aetx_sign:signed_tx()`).

lucafavatella: So the second element in this tuple means that `undefined` can be stored as value of key `type` in `#tx_log_entry.value`? If yes, this shall be reflected in the type (I currently see `aetx:tx_type()`).

lucafavatella:
@uwiger I am merging as this is improvement so we move forward, though I would prefer we address the pending comments of this PR (probably in other PR).

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
uwiger: Actually, it shouldn't be. I was pondering that while making changes before, but didn't change it then.

uwiger:
I have a change in the pipe to address the comments. Will apply it to the original PR instead.

apps/aechannel/src/aesc_fsm_min_depth_watcher.erl
lucafavatella: I see this addressed at https://github.com/aeternity/aeternity/pull/2488/commits/d04810664b3e91cdba8ee6a94419c9f203e7d812#diff-f609866532cd9838417a435634ee892aR1034

lucafavatella: I see addressed at https://github.com/aeternity/aeternity/pull/2488/commits/d04810664b3e91cdba8ee6a94419c9f203e7d812#diff-f609866532cd9838417a435634ee892aR1034

PR #2483, opened 2019-06-03
Pt 165413555 fsm ga awareness

velzevur:
[SC FSM: GA awareness](https://www.pivotaltracker.com/story/show/165413555)

Must be based upon #2431 and adapted correspondingly.

PR #2502, opened 2019-06-09
Implement GA awareness of State Channel's FSM

velzevur:
[SC FSM: GA awareness](https://www.pivotaltracker.com/story/show/165413555)

### DIFF
--- a/apps/aechannel/src/aesc_create_tx.erl
+++ b/apps/aechannel/src/aesc_create_tx.erl
@@ -141,6 +141,8 @@ check(#channel_create_tx{}, Trees,_Env) ->
 
 -spec process(tx(), aec_trees:trees(), aetx_env:env()) -> {ok, aec_trees:trees(), aetx_env:env()}.
 process(#channel_create_tx{} = Tx, Trees, Env) ->
+    {value, SignedTx} = aetx_env:signed_tx(Env),
+    {ok, ChannelPubkey} = aesc_utils:channel_pubkey(SignedTx),
     Instructions =
         aeprimop:channel_create_tx_instructions(
           initiator_pubkey(Tx),
@@ -153,7 +155,8 @@ process(#channel_create_tx{} = Tx, Trees, Env) ->
           lock_period(Tx),
           fee(Tx),
           nonce(Tx),
-          round(Tx)),
+          round(Tx),
+          ChannelPubkey),
     aeprimop:eval(Instructions, Trees, Env).
 
 -spec signers(tx(), aec_trees:trees()) -> {ok, list(aec_keys:pubkey())}.
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -28,7 +28,9 @@
 
 -export([attach_responder/2]).   %% (fsm(), map())
 
+-ifdef(TEST).
 -export([strict_checks/2]). %% tests only
+-endif.
 
 %% Inspection and configuration functions
 -export([ get_history/1     %% (fsm()) -> [Event]
@@ -329,7 +331,7 @@ message(Fsm, {T, _} = Msg) when ?KNOWN_MSG_TYPE(T) ->
 %%
 -spec signing_response(pid(), sign_tag(), any()) -> ok.
 signing_response(Fsm, Tag, Obj) ->
-    gen_statem:cast(Fsm, {signed, Tag, Obj}).
+    gen_statem:cast(Fsm, {?SIGNED, Tag, Obj}).
 
 minimum_depth_achieved(Fsm, ChanId, Type, TxHash) ->
     gen_statem:cast(Fsm, {?MIN_DEPTH_ACHIEVED, ChanId, Type, TxHash}).
@@ -525,6 +527,7 @@ has_gproc_key(Fsm, #{gproc_key := K}) ->
             false
     end.
 
+-ifdef(TEST).
 %% This is supposed to be used in tests only. If Alice stops checking
 %% her own signatures, Bob will still check hers and will reject any wrongly
 %% signed txs. There is a risk for Alice corrupting her own state if using
@@ -532,6 +535,7 @@ has_gproc_key(Fsm, #{gproc_key := K}) ->
 -spec strict_checks(pid(), boolean()) -> ok.
 strict_checks(Fsm, Strict) when is_boolean(Strict) ->
     gen_statem:call(Fsm, {strict_checks, Strict}).
+-endif.
 
 leave(Fsm) ->
     lager:debug("leave(~p)", [Fsm]),
@@ -915,7 +919,7 @@ awaiting_signature(cast, {?SIGNED, create_tx, Tx} = Msg,
                             end, D);
 awaiting_signature(cast, {?SIGNED, deposit_tx, Tx} = Msg,
                    #data{latest = {sign, deposit_tx, DTx, Updates}} = D) ->
-    maybe_check_sigs(Tx, Updates, channel_deposit_tx, not_deposit_tx, me,
+    maybe_check_sigs(Tx, Updates, aesc_deposit_tx, not_deposit_tx, me,
         fun() ->
             next_state(dep_half_signed,
                   send_deposit_created_msg(Tx, Updates,
@@ -925,7 +929,7 @@ awaiting_signature(cast, {?SIGNED, deposit_tx, Tx} = Msg,
         end, D);
 awaiting_signature(cast, {?SIGNED, withdraw_tx, Tx} = Msg,
                    #data{latest = {sign, withdraw_tx, WTx, Updates}} = D) ->
-    maybe_check_sigs(Tx, Updates, channel_withdraw_tx, not_withdraw_tx, me,
+    maybe_check_sigs(Tx, Updates, aesc_withdraw_tx, not_withdraw_tx, me,
         fun() ->
             next_state(wdraw_half_signed,
                     send_withdraw_created_msg(Tx, Updates,
@@ -936,55 +940,49 @@ awaiting_signature(cast, {?SIGNED, withdraw_tx, Tx} = Msg,
 awaiting_signature(cast, {?SIGNED, ?FND_CREATED, SignedTx} = Msg,
                    #data{role = responder,
                          latest = {sign, ?FND_CREATED, HSCTx, Updates}} = D) ->
-    NewSignedTx = aetx_sign:add_signatures(
-                    HSCTx, aetx_sign:signatures(SignedTx)),
-    maybe_check_sigs_create(NewSignedTx, Updates, both,
+    maybe_check_sigs_create(SignedTx, Updates, both,
         fun() ->
             D1 = send_funding_signed_msg(
-                  NewSignedTx,
-                  log(rcv, ?SIGNED, Msg, D#data{create_tx = NewSignedTx,
+                  SignedTx,
+                  log(rcv, ?SIGNED, Msg, D#data{create_tx = SignedTx,
                                                 latest = {ack, ?FND_CREATED,
                                                           HSCTx, Updates}})),
-            report_on_chain_tx(?FND_CREATED, NewSignedTx, D1),
-            {ok, D2} = start_min_depth_watcher({?MIN_DEPTH, ?WATCH_FND}, NewSignedTx,
+            report_on_chain_tx(?FND_CREATED, SignedTx, D1),
+            {ok, D2} = start_min_depth_watcher({?MIN_DEPTH, ?WATCH_FND}, SignedTx,
                                                Updates, D1),
             gproc_register_on_chain_id(D2),
             next_state(awaiting_locked, D2)
         end, D);
 awaiting_signature(cast, {?SIGNED, ?DEP_CREATED, SignedTx} = Msg,
                    #data{latest = {sign, ?DEP_CREATED, HSCTx, Updates}} = D) ->
-    NewSignedTx = aetx_sign:add_signatures(
-                    HSCTx, aetx_sign:signatures(SignedTx)),
-    maybe_check_sigs(NewSignedTx, Updates, channel_deposit_tx, not_deposit_tx, both,
+    maybe_check_sigs(SignedTx, Updates, aesc_deposit_tx, not_deposit_tx, both,
         fun() ->
-            D1 = send_deposit_signed_msg(NewSignedTx,
+            D1 = send_deposit_signed_msg(SignedTx,
                                           D#data{latest = {ack,
                                                            ?DEP_CREATED,
                                                            HSCTx, Updates}}),
-            report_on_chain_tx(?DEP_CREATED, NewSignedTx, D1),
-            {ok, D2} = start_min_depth_watcher({?MIN_DEPTH, ?WATCH_DEP}, NewSignedTx,
+            report_on_chain_tx(?DEP_CREATED, SignedTx, D1),
+            {ok, D2} = start_min_depth_watcher({?MIN_DEPTH, ?WATCH_DEP}, SignedTx,
                                                Updates, D1),
             next_state(awaiting_locked,
                       log(rcv, ?SIGNED, Msg, D2))
         end, D);
 awaiting_signature(cast, {?SIGNED, ?WDRAW_CREATED, SignedTx} = Msg,
                    #data{latest = {sign, ?WDRAW_CREATED, HSCTx, Updates}} = D) ->
-    NewSignedTx = aetx_sign:add_signatures(
-                    HSCTx, aetx_sign:signatures(SignedTx)),
-    maybe_check_sigs(NewSignedTx, Updates, channel_withdraw_tx, not_withdraw_tx, both,
+    maybe_check_sigs(SignedTx, Updates, aesc_withdraw_tx, not_withdraw_tx, both,
         fun() ->
-            D1 = send_withdraw_signed_msg(NewSignedTx,
+            D1 = send_withdraw_signed_msg(SignedTx,
                                           D#data{latest = {ack,
                                                            ?WDRAW_CREATED,
                                                            HSCTx, Updates}}),
-            report_on_chain_tx(?WDRAW_CREATED, NewSignedTx, D1),
-            {ok, D2} = start_min_depth_watcher({?MIN_DEPTH, ?WATCH_WDRAW}, NewSignedTx,
+            report_on_chain_tx(?WDRAW_CREATED, SignedTx, D1),
+            {ok, D2} = start_min_depth_watcher({?MIN_DEPTH, ?WATCH_WDRAW}, SignedTx,
                                                Updates, D1),
             next_state(awaiting_locked, log(rcv, ?SIGNED, Msg, D2))
         end, D);
 awaiting_signature(cast, {?SIGNED, ?UPDATE, SignedTx} = Msg,
                    #data{latest = {sign, ?UPDATE, OCTx, Updates}} = D) ->
-    maybe_check_sigs(SignedTx, Updates, channel_offchain_tx, not_offchain_tx, me,
+    maybe_check_sigs(SignedTx, Updates, aesc_offchain_tx, not_offchain_tx, me,
         fun() ->
             D1 = send_update_msg(
                   SignedTx, Updates,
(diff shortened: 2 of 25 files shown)


velzevur:
Supersedes #2483.
Complementary protocol changes are to be provided as well.

apps/aetx/src/aetx_env.erl
```diff
@@ -244,7 +244,7 @@ tx_event(Name, #env{events = Events} = Env) ->
         none -> Env;
         {value, SignedTx} ->
             TxHash = aetx_sign:hash(SignedTx),
-            {Type, _} = aetx:specialize_type(aetx_sign:tx(SignedTx)),
+            {Type, _} = aetx:specialize_type(aetx_sign:innermost_tx(SignedTx)),
```
velzevur: This is made GA aware. For transactions wrappend in a meta authentication transaction, the old processing was emitting not channel's type but rather `ga_meta_tx` instead.

apps/aeprimop/src/aeprimop.erl
```diff
@@ -341,9 +343,7 @@ channel_create_tx_instructions(InitiatorPubkey, InitiatorAmount,
                         ResponderPubkey, ResponderAmount,
                         ReserveAmount, DelegatePubkeys,
                         StateHash, LockPeriod, Nonce, Round)
-    , tx_event_op({channel, aesc_channels:pubkey(InitiatorPubkey,
-                                                 Nonce,
-                                                 ResponderPubkey)})
+    , tx_event_op({channel, ChannelPubkey})
```
velzevur: This is made GA aware. For channel create transactions wrappend in a meta authentication transaction(s) the channel pubkey is computed in a different GA-specific way (the nonce is 0, so we use the `auth_id` instead). This was emitting events with the wrong channel ID.

This is not breaking consesus but rather modifying the emitted events.

apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -183,45 +214,101 @@ init_per_group(state_hash, Config0) ->
...
-end_per_group(_Group, Config) ->
-    _Config1 = stop_node(dev1, Config),
+end_per_group(Group, Config) ->
+    case proplists:get_value(ga_group, Config, false) of
+        true ->
+            case lists:member(Group, [initiator_is_ga,
+                                      responder_is_ga, both_are_ga]) of
+                true ->
+                    _Config1 = stop_node(dev1, Config);
+                false -> pass
```
uwiger: When does this happen?

apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -1851,3 +2021,137 @@ mine_key_blocks(Node, KeyBlocks) ->
...
+    AttachTx = aega_test_utils:ga_attach_tx(Owner, Options1),
+
+    SignedTx = aec_test_utils:sign_tx(AttachTx, [OwnerPrivkey]),
+    ok = rpc(dev1, aec_tx_pool, push, [SignedTx]),
+    wait_for_signed_transaction_in_block(dev1, SignedTx),
+
+%    ConKey   = aect_contracts:compute_contract_pubkey(Owner, Nonce),
+%    CallKey  = aect_call:id(Owner, Nonce, ConKey),
+%    CallTree = aect_test_utils:calls(S1),
+%    Call     = aect_call_state_tree:get_call(ConKey, CallKey, CallTree),
```
uwiger: Delete commented code?

apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -1851,3 +2021,137 @@ mine_key_blocks(Node, KeyBlocks) ->
...
+    Options1 = maps:merge(#{nonce => Nonce, code => ByteCode,
+                            auth_fun => AuthFunHash, call_data => Calldata},
+                          Opts),
+    AttachTx = aega_test_utils:ga_attach_tx(Owner, Options1),
+
+    SignedTx = aec_test_utils:sign_tx(AttachTx, [OwnerPrivkey]),
+    ok = rpc(dev1, aec_tx_pool, push, [SignedTx]),
+    wait_for_signed_transaction_in_block(dev1, SignedTx),
+
+%    ConKey   = aect_contracts:compute_contract_pubkey(Owner, Nonce),
```
uwiger:
```suggestion
```

apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -1851,3 +2021,137 @@ mine_key_blocks(Node, KeyBlocks) ->
...
+%    CallKey  = aect_call:id(Owner, Nonce, ConKey),
```
uwiger:
```suggestion
```

apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -1851,3 +2021,137 @@ mine_key_blocks(Node, KeyBlocks) ->
...
+%    CallTree = aect_test_utils:calls(S1),
```
uwiger:
```suggestion
```

apps/aechannel/test/aesc_fsm_SUITE.erl
uwiger:
```suggestion
```

apps/aechannel/test/aesc_fsm_SUITE.erl
velzevur: This might be a bit too implicit: the groups that belong to `generalized_accounts` group - namely - `initator_is_ga`, `responder_is_ga` and `both_are_ga` start with setting up one or two generalized accounts. Usually nodes are being stopped at each group end but since we can save some contracts upgrading: we don't stop the node when finishing any of the following groups:
```
  ga_sequence() ->
      [ {group, transactions}  
      , {group, errors}
      , {group, signatures}
      , {group, channel_ids}                                 
      ].
```

To answer your question: while finishing `transactions` group for example:
* when being ran outside of the scope of `generalized_accounts` group - there is no `ga_group` being set, so we stop the node (to start it with the next group)
* when being ran in the context of any of the  `generalized_accounts` groups - we do not stop the node (so we don't lose the GA setup). Instead we stop it once finishing the group

lucafavatella:
Related doc (so we remember):
* Protocol repo may need update
* [Wiki](https://github.com/aeternity/aeternity/wiki/Troubleshooting) assumes this in in 3.2.0.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -3326,32 +3326,51 @@ naming_system_broken_txs(_Config) ->
...
-    Tx = aetx:deserialize_from_binary(CreateBinTx),
-    SignedCreateTx = aec_test_utils:sign_tx(Tx, Privkey),
-    EncSignedCreateTx = aeser_api_encoder:encode(transaction,
-                                  aetx_sign:serialize_to_binary(SignedCreateTx)),
-    ws_call_async_method(ConnPid, Method, #{tx => EncSignedCreateTx}, Config),
-    #{tx => Tx, signed_tx => SignedCreateTx, updates => Updates}.
+    {ok, TxBin} = aeser_api_encoder:safe_decode(transaction, EncTx),
+    Tx = aetx:deserialize_from_binary(TxBin),
+    SignedTx = 
+        case account_type(Pubkey) of
```
lucafavatella: I see this is changing behaviour of the test depending on the behaviour of the system under test - chain. I see that when test makes account generic it mines with `strictly_follow_top`. Seems good enough. Ok.

apps/aetx/src/aetx_env.erl
lucafavatella: Based on inspection of the code this shall make the `aesc_fsm_min_depth_watcher` work better for the `has_tx` scenario.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -5299,24 +5324,13 @@ sc_ws_timeout_open_(Config) ->
 sc_ws_attach_initiator(Config) ->
     #{initiator := #{pub_key  := Pubkey,
                      priv_key := Privkey}} = proplists:get_value(participants, Config),
-    attach({Pubkey, Privkey}, "simple_auth", "authorize", ["42"]),
+    attach({Pubkey, Privkey}, "simple_auth", "authorize", [?SIMPLE_AUTH_GA_SECRET]),
```
lucafavatella: Could you please consider PR #2510 - it aims at polishing this area of the test.

velzevur:
I am still to provide a PR in the `protocol` repo

apps/aechannel/src/aesc_create_tx.erl
```diff
@@ -153,7 +155,8 @@ process(#channel_create_tx{} = Tx, Trees, Env) ->
           lock_period(Tx),
           fee(Tx),
           nonce(Tx),
-          round(Tx)),
+          round(Tx),
+          ChannelPubkey),
```
lucafavatella: This is independent refactoring and not strictly needed, right?

It is ok to keep here - I am asking simply because I am considering regression testing across tags using this module as interface.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1551,7 +1547,8 @@ handle_call_(open, {upd_create_contract, Opts}, From, #data{} = D) ->
     end,
     new_contract_tx_for_signing(Opts#{owner => FromPub}, From, D);
 handle_call_(open, {upd_call_contract, Opts, ExecType}, From,
-             #data{state=State, opts = ChannelOpts} = D) ->
+             #data{state=State, opts = ChannelOpts,
+                   on_chain_id = ChannelId} = D) ->
```
lucafavatella: The code path seems to expect a binary rather than `undefined` - could this clause be made more specific?
```suggestion
                   on_chain_id = ChannelId} = D) when ChannelId =/= undefined ->
```
Reading it is not immediate that this is not undefined - maybe for a previous operation/check?

There may be other points where this applies.

apps/aechannel/src/aesc_offchain_state.erl
```diff
@@ -335,13 +332,9 @@ fallback_to_stable_state(#state{signed_tx = LastSignedTx}=State)
...
 -spec mutually_signed(aetx_sign:signed_tx()) -> boolean().
 mutually_signed(SignedTx) ->
-    case aetx_sign:signatures(SignedTx) of
-        [_, _] ->
-            %% mutually signed
-            true;
-        _ ->
-            false
-    end.
+    %% This does not really check if authications were valid. TODO: Fix it
```
lucafavatella: Can this TODO be deleted? Code was not checking signatures earlier either.
```suggestion
```

apps/aechannel/src/aesc_create_tx.erl
velzevur: Without this, the wrong `channel_id` event is being sent for GAs - if the `initiator` is a generalized account one's nonce is not being used in `channel_id` computation. This fixes it

apps/aechannel/src/aesc_utils.erl
```diff
@@ -791,3 +851,46 @@ add_call(Call0, TxHash, Trees, Env) ->
...
+            try {ok, Mod:channel_pubkey(Txi)}
+            catch error:_ -> {error, not_channel_tx}
+            end
+    end.
+
+get_nonce_or_auth_id(SignedTx, Signer) ->
+    case aetx:specialize_callback(aetx_sign:tx(SignedTx)) of
+        {aega_meta_tx, MetaTx} ->
+            case aega_meta_tx:ga_pubkey(MetaTx) =:= Signer of
+                true -> aega_meta_tx:auth_id(MetaTx);
```
lucafavatella: This seems to take the auth id of **outermost** ga meta tx related to the account, while an on-chain processing of tx takes the **innermost** one (see [`aega_meta_tx:process`](https://github.com/aeternity/aeternity/blob/996df3264e07c6b6f5b3f9b8f5c3d10ca2dd1ffe/apps/aega/src/aega_meta_tx.erl#L187)). Is this intended?

I am still getting my head around this, and this may have no impact if multiple ga meta tx for same account prevented somewhere else. Though there may be room for opportunity in readability (maybe in other point in code) hence this note.

velzevur: Different transaction types have different `origin` and the `origin` of `channel_create_transaction` is the `initiator`. This function is intended to get the `initiator`'s nonce or auth_id out from `channel_create_transaction` so it can later on be used in `channel_id` computation. In case that the outermost transaction indeed is a meta one, it could be either the `initiator` or the `responder` authentication method, thus the `aega_meta_tx:ga_pubkey(MetaTx) =:= Signer` check - we return only the `Signer`'s `auth_id`.

What botters me now is the other option - if none of the participants is a GA, the funciton returns `Mod:nonce(Tx)`. This could return unexpected result if the `get_nonce_or_auth_id/2` function is called for the `responder` instead.

I am to rework and rename it to express more clearly the intention regarding the transaction type and the participant.

velzevur: Addressed

lucafavatella: Change is improvement thanks.

Doubt re outermost-vs-innermost still stays. For the same account-meant-to-be-authenticated, multiple ga meta txs can be nested within each other.

apps/aechannel/src/aesc_utils.erl
```diff
@@ -524,6 +554,36 @@ is_delegatable_tx_type(Type) ->
 delegatable_tx_types() ->
     [slash].
 
+-spec verify_signatures_onchain(aetx_sign:signed_tx(), aec_trees:trees(),
+                                aetx_env:env()) -> ok | {error, atom()}.
+verify_signatures_onchain(SignedTx, Trees, Env) ->
+    verify_signatures_onchain(SignedTx, Trees, Env, []).
+
+verify_signatures_onchain(SignedTx, Trees, Env, CheckedSigners) ->
```
lucafavatella: I arrived here from clause `awaiting_signature(cast, {?SIGNED, create_tx, Tx} = Msg,` in the FSM.

I have the impression that this function does not prevent not relevant extra ga meta txs - i.e. ga meta tx authenticating accounts not relevant to operation. Though I see that this [is checked on chain](https://github.com/aeternity/aeternity/pull/2354/files#diff-15a4b5418b5f5c5a1f2fb8d8d0d3f619R1330), and FSM waits for tx in chain anyway. So ok.

Though an assertion `aesc_utils:count_authentications(...) =:= 1` would improve readability in code and reduce chance of user(s?) preparing wrong tx. Maybe could be considered as future improvement (outside of this PR).

lucafavatella: Similarly for code path from `half_signed(cast, {?FND_SIGNED, Msg}, #data{role = initiator,`. Though ok.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -3368,8 +3442,38 @@ account_type(Pubkey) ->
...
     case {account_type(Initiator), account_type(Responder)} of
-        {{ok, basic}, {ok, basic}}  -> ok;
-        {{ok, generalized}, _}      -> {error, generalized_account};
-        {_, {ok, generalized}}      -> {error, generalized_account};
-        _                           -> {error, not_found}
+        {{ok, _}, {ok, _}}  -> ok;
+        _                   -> {error, not_found}
     end.
+
+combine_sigs_or_meta(HalfSigned1, HalfSigned2) ->
```
lucafavatella: I see that a correct ga auth fun [should enquire](https://github.com/aeternity/protocol/blob/34e8eeee3c9d3c7fafa83883bc7d95273315125d/generalized_accounts/ga_explained.md#authorization) [the inner tx hash](https://github.com/aeternity/aeternity/blob/996df3264e07c6b6f5b3f9b8f5c3d10ca2dd1ffe/apps/aega/src/aega_meta_tx.erl#L183), otherwise the auth can be used for forging other tx.

(Trying to understand how this combining works.)

velzevur:
The `combine_sigs_or_meta` must be revisited. Added a `WIP`.

apps/aeprimop/src/aeprimop.erl
velzevur: This solves: [Check event emited when GA channel is created](https://www.pivotaltracker.com/story/show/166272180)

apps/aecore/test/aecore_suite_utils.erl
```diff
@@ -1123,3 +1124,16 @@ rpc(Node, Mod, Fun, Args) ->
...
     {Pubkey, Privkey}.
+
+meta_tx(Owner, AuthOpts, AuthData, InnerTx0) ->
+    InnerSTx =
+        try aetx_sign:tx(InnerTx0) of
+            _Tx -> InnerTx0
+        catch _:_ ->
+            aetx_sign:new(InnerTx0, [])
+        end,
+    Options1 = maps:merge(#{%fee => 50000 * aec_test_utils:min_gas_price(),
```
lucafavatella: Shall this comment be deleted?

apps/aehttp/test/aehttp_sc_SUITE.erl
```diff
@@ -3113,3 +3130,23 @@ to_binary(I) when is_integer(I) ->
     integer_to_binary(I);
 to_binary([_|_] = L) ->
     iolist_to_binary([to_binary(X) || X <- L]).
+
+simple_auth_meta(Owner, Secret, GANonce, InnerMostTx, InnerTx) ->
+    AuthTx =
+        case aetx:specialize_type(InnerMostTx) of
+            {channel_offchain_tx, _} -> InnerMostTx;
+            _ -> InnerTx
```
lucafavatella: Totally optional. I find it more readable in this way.
```suggestion
            {_, _} -> InnerTx
```

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -936,55 +940,49 @@ awaiting_signature(cast, {?SIGNED, withdraw_tx, Tx} = Msg,
 awaiting_signature(cast, {?SIGNED, ?FND_CREATED, SignedTx} = Msg,
```
lucafavatella: It feels that the macros like `FND_CREATED` in aesc_codec.hrl (`-define(FND_CREATED  , funding_created).`) shall be shared with `sc_ws_api_jsonrpc` - as part of encoding/decoding in user API.

Though orthogonal to this PR.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -936,55 +940,49 @@ awaiting_signature(cast, {?SIGNED, withdraw_tx, Tx} = Msg,
...
                    #data{role = responder,
                          latest = {sign, ?FND_CREATED, HSCTx, Updates}} = D) ->
-    NewSignedTx = aetx_sign:add_signatures(
-                    HSCTx, aetx_sign:signatures(SignedTx)),
```
lucafavatella: Who does now check that `SignedTx` is related to `HSCTx` rather than being a completely distinct transaction?

lucafavatella: Is the assumption maybe that the user of the user API is trusted?

velzevur: Assumption is that if it is a different:
* new transaction - then authentication method will not match
* old co-signed transaction - the check round will catch it

velzevur:
We used to have the following flow:
* Alice requests a change from her `FSM_A`
* the `FSM_A` prepares a set of updates as well as a new transaction
* the `FSM_A` requests Alice to authenticate the `aetx:tx()` (1)
* Alice returnns to the `FSM_A` the solo-authenticated tx
* `FSM_A` sends the solo-authenticated tx to `FSM_B`
* the `FSM_B` requests Alice to authenticate the `aetx:tx()` (2)
* the `FSM_B` sends the solo-authenticated tx to `FSM_B` (3)

This has changed to (differences are `1`, `2` and `3`):
* Alice requests a change from her `FSM_A`
* the `FSM_A` prepares a set of updates as well as a new transaction
* the `FSM_A` requests Alice to authenticate the `aetx_sign:signed_tx()` (1)
* Alice returnns to the `FSM_A` the solo-authenticated tx
* `FSM_A` sends the solo-authenticated tx to `FSM_B`
* the `FSM_B` requests Alice to sign the solo-authenticated `aetx_sign:signed_tx()` (2)
* the `FSM_B` sends the co-authenticated tx to `FSM_B` (3)

Note that `3` changes the `noise` protocol message and expectations for the message format itself, thus making it non-backwards compatible with older versions.

velzevur:
From the remaining System Tests for State Channels, commit [eac7cb1](https://github.com/aeternity/aeternity/pull/2502/commits/eac7cb19ef7052a89af045f3b966cf3b80dbbb61) allowed runing `aest_channels_SUITE` as smoke tests and commit [6ab8f3b](https://github.com/aeternity/aeternity/pull/2502/commits/6ab8f3b3107fa2b18169e145111d93aec6a6c209) fixed a small issue in it. The latter passed.

The normal smoke tests behaviour is returned in commit [531ae60](https://github.com/aeternity/aeternity/pull/2502/commits/531ae60a384d066de7006a690da0aae2ef7db7fb).

docs/release-notes/next/PT-165413555_FSM_GA_awareness.md
```diff
@@ -0,0 +1,8 @@
+* Introduces State Channel FSM Generalized Accounts awareness. It now fully
+  supports creation of meta transactions and verifies them.
+* Changes WebSocket API regarding signing transactions: now sign requests
```
lucafavatella: I read this line mentions the change in the user API. Ok.

lucafavatella: > * the `FSM_B` sends the co-authenticated tx to `FSM_B` (3)
> 
> Note that `3` changes the `noise` protocol message and expectations for the message format itself, thus making it non-backwards compatible with older versions.

Shall there be somewhere in this file at least a mention of the channels Noise node-to-node protocol breakage/change?

lucafavatella:
Amended [wiki](https://github.com/aeternity/aeternity/wiki/Troubleshooting/_compare/ca589e5186b92c2e5ecff80ff0e46a041b3d9886...140099103743a97a67e0661c255e7d04c1814ee9) with target release.

PR #2504, opened 2019-06-10
Channels system test failure

uwiger:
See [PT #166493847](https://www.pivotaltracker.com/story/show/166493847)

Matching of `channel_changed` events added to the flow. Test still fails locally on amount mismatch after `close_mutual`

**EDIT Jun 12** Tests now pass. Code has been added to consume new "channel_changed" messages sent by the state channel fsm, and hard-coded assumptions about the `close_mutual` fee (confusingly represented as `SplitOpenFee` in the code) were removed; instead, the actual fee is fetched from the generated tx.


uwiger:
The correction has been verified by pushing to [system-tests](https://github.com/aeternity/aeternity/tree/system-tests)

PR #2500, opened 2019-06-07
Upgrade rebar3 build tool to latest stable one

lucafavatella:
TODO:
* [x] Check CI passes, using latest upstream binary. See also See also https://github.com/aeternity/rebar3/pull/9#issuecomment-499881414
* [x] Smoke test running tests locally.
  * Build ran locally (macOS, non-default path) as on master. I [updated the wiki re rocksdb build](https://github.com/aeternity/aeternity/wiki/Troubleshooting/_compare/c6557a7058605108d6ba2d102cfa81d00cc790a0...b11762b9bf57a0b6d7008e6dde99fc8f95a08830).
  * Unit tests ran ok locally.
  * An integration test ran ok locally.
* [ ] Update PR using **released** (tag name `3.11.1-aeternity.1` ?) binary built from branch `integration` of `aeternity/rebar3` fork after https://github.com/aeternity/rebar3/pull/9 merged.

PR #2505, opened 2019-06-11
Upgrade rebar3 build tool to latest stable one

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/166546345

Supersedes #2500


lucafavatella:
@venimus Would you be able to test https://www.pivotaltracker.com/story/show/166600986 on this PR? Or after merge?

venimus:
@lucafavatella The findstr issue is fixed but another one appears. If node is installed in a folder that has a space it does not work. So the default suggested path for installation `C:\Program Files\aeternity-windows-w64` won't work out of the box. The issue is yet again in relx which has messy quotes when executing erlang. 
If the installation path does not include spaces, and is executed as Administrator it works without problems.

PR #2506, opened 2019-06-11
Add crypto operators and improve FATE tests in aecontract_SUITE

hanssv:
- Test addr_to_str and int_to_str in FATE

- Implement hash operations in FATE

- Tests for FATE hash instructions

- Implement ecverify(_secp256k1) for FATE

- introduce `aeu_crypto` as a utility module for (ECDSA) crypto.

- Use Map-result from Sophia-compiler for FATE

- Add contract_to_address to FATE and enable test

- Rename AEVM specific tests in aecontract_SUITE

- Adjust savecoinbase test to work also for FATE

- Adjust sophia_operators test to work also for FATE

- Fix Bits errors and adjust test for FATE

- Fix safe_math test and implement POW for FATE

### DIFF
--- a/apps/aefate/src/aefa_fate_op.erl
+++ b/apps/aefate/src/aefa_fate_op.erl
@@ -114,7 +114,9 @@
         , aend_update/1
         , aens_transfer/1
         , aens_revoke/1
-        , ecverify/1
+        , ecverify/5
+        , ecverify_secp256k1/5
+        , contract_to_address/3
         , sha3/3
         , sha256/3
         , blake2b/3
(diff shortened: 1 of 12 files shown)


apps/aefate/src/aefa_fate_op.erl
```diff
@@ -911,13 +940,11 @@ op('div', A, B)  when ?IS_FATE_INTEGER(A)
...
-    %% TODO: Implement arbitrary precision pow function.
-    try round(math:pow(A, B)) of
-        I -> I
-    catch error:badarith ->
-            aefa_fate:abort(pow_too_large_exp)
+op(pow, A, B)  when ?IS_FATE_INTEGER(A), ?IS_FATE_INTEGER(B) ->
+    if B < 0 ->
+           aefa_fate:abort({arithmetic_error, negative_exponent});
+       true ->
+           pow(A, B)
```
happi: Throws an exception for B > 400.

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -1045,3 +1083,13 @@ binary_reverse(Binary) ->
     Size = erlang:size(Binary)*8,
     <<X:Size/integer-little>> = Binary,
     <<X:Size/integer-big>>.
+
+%% TODO: we should check Gas as we go along here...
```
happi: I think we should charge gas upfront relative to the size of B. (possibly somewhat related to A also).

PR #2508, opened 2019-06-14
Fixed double gas expense on DELEGATECALL

dominicletz:
This case seems to have been missed by accident.

### DIFF
--- a/apps/aevm/src/aevm_eeevm.erl
+++ b/apps/aevm/src/aevm_eeevm.erl
@@ -1542,6 +1542,9 @@ spend_op_gas(?CALL, State) ->
 spend_op_gas(?CALLCODE, State) ->
     %% Delay this until the actual operation
     State;
+spend_op_gas(?DELEGATECALL, State) ->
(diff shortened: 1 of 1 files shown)


apps/aevm/src/aevm_eeevm.erl
```diff
@@ -1542,6 +1542,9 @@ spend_op_gas(?CALL, State) ->
 spend_op_gas(?CALLCODE, State) ->
     %% Delay this until the actual operation
     State;
+spend_op_gas(?DELEGATECALL, State) ->
```
happi: If there would be any contract on a chain for the original aevm/solidity VM (it is not possible to create such contracts any longer) and that contract has this instruction then any execution of that contract would now have a different gas and it would not be possible to validate such contract.  Any change on the protocol level has to be guarded by a hard fork height test, making sure the change only take effekt after that hard fork height,

happi:
This change is not protocol safe and it is in dead code.

PR #2511, opened 2019-06-17
Stabilize test waiting for both nodes to mine

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/166728543

### DIFF
--- a/system_test/common/aest_compatibility_SUITE.erl
+++ b/system_test/common/aest_compatibility_SUITE.erl
@@ -71,6 +71,7 @@ test_mining_algorithms_compatibility(Cfg) ->
 
     #{ public := PubKey1 } = enacl:sign_keypair(),
     #{ public := PubKey2 } = enacl:sign_keypair(),
+    ?assertNotEqual(PubKey1, PubKey2),
     Beneficiary1 = aeser_api_encoder:encode(account_pubkey, PubKey1),
     Beneficiary2 = aeser_api_encoder:encode(account_pubkey, PubKey2),
 
--- a/system_test/common/aest_compatibility_SUITE.erl
+++ b/system_test/common/aest_compatibility_SUITE.erl
@@ -102,11 +102,7 @@ get_block_beneficiary(NodeName, Height) ->
 
 get_chain_beneficiaries(NodeName, Height) when is_integer(Height),
                                                Height >= 1 ->
-    sets:to_list(
-      lists:foldl(
-        fun(H, Bs) ->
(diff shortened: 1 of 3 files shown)


system_test/common/aest_compatibility_SUITE_data/aeternity.yaml.mustache
```diff
@@ -19,7 +19,7 @@ keys:
     dir: ./keys
 
 chain:
-    persist: true
+    persist: false
```
lucafavatella: Eventually I did not rely on this though I suggest to keep disabled.

dincho: Isn't this going to consume more memory then ? Eventually slow down the execution 🤔

lucafavatella: It shall not matter - no txs and short chain - though reverted in 15f6e8b0.

lucafavatella: It shall not matter - no txs and short chain - though reverted in 15f6e8b0.

system_test/common/aest_compatibility_SUITE.erl
```diff
@@ -81,27 +84,51 @@ test_mining_algorithms_compatibility(Cfg) ->
...
+setup(NodeSpecs, Config, Cfg) ->
+    setup_nodes([maps:put(config, Config, N) || N <- NodeSpecs], Cfg).
 
+get_block_beneficiary(NodeName, Height) ->
+    #{ beneficiary := Beneficiary } = get_block(NodeName, Height),
+    Beneficiary.
+
+get_chain_beneficiaries(NodeName, Height) when is_integer(Height),
+                                               Height >= 1 ->
+    sets:to_list(
```
hanssv:
```
lists:usort([ get_block_beneficiary(NodeName, H) || H <- lists:seq(1, Height) ])
```
is better?

lucafavatella: The set rather than full list was mainly for preventing copy-paste of function and reuse in case of many beneficiaries (avoiding creating long list).

Also, the direction of the traversals from top to bottom (still by height) is debatable.

It is good enough?

lucafavatella: Addressed in new commit, [passed CI](https://circleci.com/gh/aeternity/aeternity/43762).

PR #2514, opened 2019-06-17
Enhance status endpoint

dincho:
- public peer key would allow stronger health checks for seed nodes (node with static peer keys)
- top block hash and height would allow easier manual validation during releases for example, also when helping diagnose user issues as they have to provide single endpoint output.

While all of this endpoint response data is available in their corresponding API endpoints, summary endpoint is a must for easy debug and health checks.

PT: https://www.pivotaltracker.com/story/show/166054978


apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -617,15 +621,10 @@ handle_request_('GetStatus', _Params, _Context) ->
...
-
-handle_request_('GetPeerKey', _Req, _Context) ->
-    case aehttp_logic:peer_pubkey() of
-        {ok, PeerKey} ->
-            {200, [], #{pub_key => aeser_api_encoder:encode(peer_pubkey, PeerKey)}};
-        {error, key_not_found} ->
-            {404, [], #{reason => <<"Keys not configured">>}}
-    end;
+       <<"network_id">>                 => aec_governance:get_network_id(),
+       <<"peer_pubkey">>                => aeser_api_encoder:encode(peer_pubkey, PeerPubkey),
```
lucafavatella: Could you please update swagger.yml (and re-generate the json) for these 3 new fields.

dincho: Yeah, I realized I missed it myself. Now updated and the original commits has been updated. (I Plan merge instead of squash + rebase)

dincho:
> Worth a line in the release notes?

Add in 1914c4c4

PR #2517, opened 2019-06-19
SC port multiplexing

uwiger:
See [PT #161143845](https://www.pivotaltracker.com/story/show/161143845)

Existing test suites pass. No test case to test the "server scenario" where the responder sets `initiator => any`. There might be corner cases where responders terminate, leaving lingering acceptor processes; looking into ways to prevent that.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -26,6 +26,8 @@
          get_contract/2,
          get_poi/2]).
 
+-export([attach_responder/2]).   %% (fsm(), map())
+
 -export([strict_checks/2]). %% tests only
 
 %% Inspection and configuration functions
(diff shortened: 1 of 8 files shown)


lucafavatella:
Dialyzer is not happy.

lucafavatella:
Worth a line in the release notes?

uwiger:
> Dialyzer is not happy.

Yes, an annoying feature of dialyzer, that you have to run it every time you've made some substantial change. ;-)
I've addressed the warnings. Basically, the type specs no longer matched the implementation.

lucafavatella:
You may want to run aest_channels_SUITE

uwiger:
> You may want to run aest_channels_SUITE

I have done so locally, and it passes.

uwiger:
Rebased against current master

PR #2520, opened 2019-06-19
Pt 166696064 add decode calldata

ThomasArts:


UlfNorell:
Should go into `FATE-codefreeze-master`?

PR #2521, opened 2019-06-19
Fix badly typed JSON data for abi_version/vm_version

hanssv:


gorillainduction:
I wonder if this is going to break the SDK, though.

PR #2528, opened 2019-06-20
PT-166788837 bytes

UlfNorell:
Proper bytes type in FATE

Depends on:
- [x] https://github.com/aeternity/aebytecode/pull/56 
- [x] https://github.com/aeternity/aesophia/pull/93


ThomasArts:
pointing to specific aebytecode commit, should be update to point to latest master 59af12b

PR #2533, opened 2019-06-25
PT-166233670 FATE events

UlfNorell:
~~TODO~~:
- [x] Merge https://github.com/aeternity/aebytecode/pull/58
- [x] Merge https://github.com/aeternity/aesophia/pull/95
- [x] More thorough testing with different payload and index types

### DIFF
--- a/apps/aefate/src/aefa_fate_op.erl
+++ b/apps/aefate/src/aefa_fate_op.erl
@@ -704,17 +704,46 @@ call_value(Arg0, EngineState) ->
     Value = aefa_engine_state:call_value(EngineState),
     write(Arg0, aeb_fate_data:make_integer(Value), EngineState).
 
-log(_Arg0, _EngineState) -> exit({error, op_not_implemented_yet}).
+log(Arg0, EngineState) ->
+    log_([Arg0], EngineState).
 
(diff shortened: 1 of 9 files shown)


apps/aefate/src/aefa_fate_op.erl
```diff
@@ -692,17 +704,46 @@ call_value(Arg0, EngineState) ->
...
+    log_([Arg0, Arg1, Arg2, Arg3, Arg4, Arg5], EngineState).
+
+log_(Args, EngineState) ->
+    Pubkey = aefa_engine_state:current_contract(EngineState),
+    {[Payload | Indices], ES1} = get_op_args(Args, EngineState),
+    ToBin = fun(?FATE_STRING(Bin)) when ?IS_FATE_STRING(Bin) -> Bin;
+               (?FATE_BYTES(Bin))  -> Bin;
+               (Other)             -> aefa_fate:abort({value_does_not_match_type, Other, string})
+            end,
+    ToWord = fun(N) when is_integer(N) -> <<N:256>>;
```
happi: Do we want to just silently drop bits for large N?

UlfNorell: It can be debated of course, but since events don't affect the operation of a contract I felt this to be preferable to failing.

PR #2537, opened 2019-06-26
Dev stratum -- review

lucafavatella:


lucafavatella:
> We need to build stratum client first before running stratum CT suite:
> `KIND=test make stratum-client-internal-build`

I still need to get familiar with that portion of the makefile though it seems it is run as prerequisite without need to explicitly run it - see the `aestratum_client` section in the dry-run output below:
```
$ make -n ct-fortuna SUITE=apps/aestratum/test/aestratum_integration
git rev-parse HEAD > REVISION
./rebar3 as test compile -d
./rebar3 swagger_endpoints
./rebar3 as test release
(cd ./_build/test/ && \
	  if [ ! -d aestratum_client ]; then \
	      git clone --depth 1 https://github.com/aeternity/aestratum_client; \
	  fi && \
	  cd aestratum_client && \
	  ./rebar3 as test compile -d && \
	  ./rebar3 as test release)
NODE_PROCESSES="$(ps -fea | grep bin/aeternity | grep -v grep)"; \
	if [ $(printf "%b" "${NODE_PROCESSES}" | wc -l) -gt 0 ] ; then \
		(printf "%b\n%b\n" "Failed testing: another node is already running" "${NODE_PROCESSES}" >&2; exit 1);\
	else \
		AETERNITY_TESTCONFIG_DB_BACKEND=mnesia \
			./rebar3 ct  --suite=apps/aestratum/test/aestratum_integration_SUITE --sys_config config/test-fortuna.config; \
	fi
```

skkw:
OK, I missed there already is a dependency:

`internal-ct: internal-build stratum-client-internal-build`

PR #2540, opened 2019-06-26
HTTP API cache

dincho:
Since the blockchain data is immutable (after certain period of time), it could be cached very aggressively at the **clients** and intermediaries (e.g. HTTP proxies) level. This would drastically improve the **clients** performance and implicitly lower the node load.

This PR implements [HTTP caching](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching) headers in the node HTTP API. It utilize  [HTTP Expires header](https://tools.ietf.org/html/rfc7234#section-5.3) to control the resource [freshness](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#Freshness) and [ETag HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#ETags) for effective cache revalidation.

This has been implemented as callbacks for [cowboy rest behaviour](https://ninenines.eu/docs/en/cowboy/2.2/guide/rest_handlers/).

Expires header has been chosen over Cache-Control because of former cowboy support. There shouldn't be any functional differences between the two.

This is not about HTTP API backend cache.

PT: https://www.pivotaltracker.com/story/show/165271539

- [x] Add PR description
- [x] Add release notes

### DIFF
--- a/apps/aehttp/src/aehttp_api_handler.erl
+++ b/apps/aehttp/src/aehttp_api_handler.erl
@@ -20,11 +20,7 @@
     validator :: jesse_state:state()
 }).
 
--ifdef(TEST).
--define(DEFAULT_HTTP_CACHE_ENABLED, true).
--else.
 -define(DEFAULT_HTTP_CACHE_ENABLED, false).
--endif.
 
 init(Req, {OperationId, AllowedMethod, LogicHandler}) ->
     %% TODO: make this validator a proper service.
--- a/apps/aehttp/src/aehttp_api_handler.erl
+++ b/apps/aehttp/src/aehttp_api_handler.erl
@@ -10,6 +10,8 @@
 -export([delete_resource/2]).
 -export([forbidden/2]).
 -export([handle_request_json/2]).
+-export([generate_etag/2]).
+-export([expires/2]).
 
 -record(state, {
     operation_id :: atom(),
@@ -50,6 +52,12 @@ content_types_provided(Req, State) ->
 
 delete_resource(Req, State) ->
     handle_request_json(Req, State).
+
+generate_etag(Req, State = #state{operation_id = OperationId}) ->
+    aehttp_cache_etag:generate(OperationId, Req, State).
+
+expires(Req, State = #state{operation_id = OperationId}) ->
+    aehttp_cache_expires:expires(OperationId, Req, State).
 
 handle_request_json(Req0, State = #state{
         operation_id = OperationId,
--- a/apps/aehttp/src/aehttp_api_handler.erl
+++ b/apps/aehttp/src/aehttp_api_handler.erl
@@ -108,7 +108,7 @@ to_error({Reason, Name, Info}) ->
        parameter => Name,
        info => Info }.
 
--spec cache_enabled() -> non_neg_integer().
+-spec cache_enabled() -> boolean().
 cache_enabled() ->
     aeu_env:user_config_or_env([<<"http">>, <<"cache">>, <<"enabled">>],
                                aehttp, [cache, enabled], ?DEFAULT_HTTP_CACHE_ENABLED).
--- a/apps/aehttp/src/aehttp_api_handler.erl
+++ b/apps/aehttp/src/aehttp_api_handler.erl
@@ -54,10 +54,16 @@ delete_resource(Req, State) ->
     handle_request_json(Req, State).
 
 generate_etag(Req, State = #state{operation_id = OperationId}) ->
-    aehttp_cache_etag:generate(OperationId, Req, State).
+    case cache_enabled() of
+        true -> aehttp_cache_etag:generate(OperationId, Req, State);
+        false -> {undefined, Req, State}
+    end.
 
 expires(Req, State = #state{operation_id = OperationId}) ->
-    aehttp_cache_expires:expires(OperationId, Req, State).
+    case cache_enabled() of
+        true -> aehttp_cache_expires:expires(OperationId, Req, State);
+        false -> {undefined, Req, State}
+    end.
 
 handle_request_json(Req0, State = #state{
         operation_id = OperationId,
@@ -95,3 +101,8 @@ to_error({Reason, Name, Info}) ->
     #{ reason => Reason,
        parameter => Name,
        info => Info }.
+
+-spec cache_enabled() -> non_neg_integer().
+cache_enabled() ->
+    aeu_env:user_config_or_env([<<"http">>, <<"cache">>, <<"enabled">>],
+                               aehttp, [cache, enabled], false).
--- a/apps/aehttp/src/aehttp_api_handler.erl
+++ b/apps/aehttp/src/aehttp_api_handler.erl
@@ -20,6 +20,12 @@
     validator :: jesse_state:state()
 }).
 
+-ifdef(TEST).
+-define(DEFAULT_HTTP_CACHE_ENABLED, true).
+-else.
+-define(DEFAULT_HTTP_CACHE_ENABLED, false).
+-endif.
+
 init(Req, {OperationId, AllowedMethod, LogicHandler}) ->
     %% TODO: make this validator a proper service.
     JsonSpec = aehttp_api_validate:json_spec(),
@@ -105,4 +111,4 @@ to_error({Reason, Name, Info}) ->
 -spec cache_enabled() -> non_neg_integer().
 cache_enabled() ->
     aeu_env:user_config_or_env([<<"http">>, <<"cache">>, <<"enabled">>],
-                               aehttp, [cache, enabled], false).
+                               aehttp, [cache, enabled], ?DEFAULT_HTTP_CACHE_ENABLED).
--- a/apps/aehttp/src/aehttp_cache_etag.erl
+++ b/apps/aehttp/src/aehttp_cache_etag.erl
@@ -5,9 +5,12 @@
 
 -export([generate/3]).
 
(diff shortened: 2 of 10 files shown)


apps/aehttp/src/aehttp_cache_expires.erl
```diff
@@ -0,0 +1,218 @@
...
+%% Get a micro block
+%% Can be cached aggressively, however micro-blocks of last X key blocks can be reorganized by forks,
+%% thus their cache time should not exceed EXPECTED_BLOCK_MINE_RATE_MINUTES (3 min.)
+%% Expires = ExpectedNextBlockTime() -> TopBlockTime + InterBlockTime
+%%
+%% Also micro blocks can be reorganized because of micro-forks,
+%% so micro-blocks of current generation cannot be cached for more than MICRO_BLOCK_CYCLE (3s)
+%% Expires = ExpectedNextMicroBlockTime() -> TopMicroBlockTime + InterMicroBlockTime
+%%
+%% Get a key block
```
sennui: this one is the most controversial. We have key blocks happening after 30sec or after 7 minutes...

apps/aehttp/src/aehttp_cache_expires.erl
```diff
@@ -0,0 +1,218 @@
...
+    case block_header_cache_time(Header) of
+        CacheTime when is_integer(CacheTime) ->
+            unixtime_to_datetime(CacheTime);
+        _ -> error
+    end.
+
+-spec block_header_cache_time(aec_headers:header()) -> non_neg_integer() | error.
+block_header_cache_time(Header) ->
+    Now = aeu_time:now_in_secs(),
+    BlockTimestamp = aec_headers:time_in_secs(Header),
```
sennui: Potential problem here: we compare our local time with block's time synced from the network (which may have different clock time)

apps/aehttp/src/aehttp_cache_expires.erl
hanssv: Wouldn't it be better to listen for new block events and invalidate the cache based on this? Not sure we have exactly the right events, but that could be fixed I think?!

sennui: Exactly the only thing we can make. Price to pay to have a cache

apps/aehttp/src/aehttp_cache_expires.erl
dincho: That's a valid point. However we compare aggregated time period, which statistically given the protocol time constraints should be fine if `aged_blocks_time` is not set too low.

apps/aehttp/src/aehttp_cache_expires.erl
dincho: I think there is misunderstanding here (I updated the PR description). This is about HTTP cache headers that's a spec. for HTTP caching implementation in the clients. There is no caching in the node.

The only consequence here is that a block can be cached for a bit longer than it's real generation time if the block is found earlier than expected. However, that time won't exceed the expected block time. So on average, the clients should have their fresh copy on time, it small exceptions.

I don't think this is too different from the protocol expectations and implementation ? 🤔

sennui: Yes our comments are invalid. The problem is the guidance for clients when to consider resource expired and we can't do anything about it from server side.

sennui: Maybe use the most pessimistic one? (like the shortest keyblock time over the last day?)

sennui:
To sum up - only pain points are:
1. caching /top
2. caching microblocks in case of microfork

Re 1. it's covered with protocol protected 3 sec frequency of microblock dispatch
Re 2. here we need to pick a good expiration which will not hurt clients, so they don't consider a microblock on main chain for too long

dincho:
> Re 2. here we need to pick a good expiration which will not hurt clients, so they don't consider a microblock on main chain for too long

This should be addressed in a1142129df5b76e67b415292aded150168deaa75

apps/aehttp/src/aehttp_cache_expires.erl
hanssv: Ah, client caching 🤦‍♂

apps/aehttp/src/aehttp_api_handler.erl
```diff
@@ -18,6 +20,12 @@
     validator :: jesse_state:state()
 }).
 
+-ifdef(TEST).
+-define(DEFAULT_HTTP_CACHE_ENABLED, true).
```
lucafavatella: Why rather not setting a default in `/config/test-lima.config`?

lucafavatella: (and all other test configs)

lucafavatella: Rather - if needed for dev1 ... instantiated in integration tests, this shall be set in config of that dev1 ... node.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -6040,6 +6046,26 @@ cors_returned_on_get_request(_Config) ->
...
+%% Test HTTP cache headers
+%% ============================================================
+
+cache_headers(_Config) ->
+    Host = external_address(),
+    {ok, {{_, 200, _}, Headers, _Body}} =
+        httpc_request(get, {Host ++ "/v2/blocks/top", []}, [], []),
+
+    true = proplists:is_defined("expires", Headers),
+    true = proplists:is_defined("etag", Headers),
```
lucafavatella: I am not seeing where the actual usage of etag is tested. I expect a conditional get request (is [If-None-Match](https://tools.ietf.org/html/rfc7232#section-3.2) the tag sent by the client?), checking that the response has correct status code (not 200) and that response has not the full blown response body.

lucafavatella: An additional test case, testing this, would be beneficial. So to check that the functionality actually works.

(BTW It may be done at several layers - either system or integration test.)

apps/aehttp/src/aehttp_cache_etag.erl
```diff
@@ -0,0 +1,174 @@
...
+-spec etag_block(aec_blocks:block()) -> binary().
+etag_block(Block) ->
+    Header = aec_blocks:to_header(Block),
+    etag_block_header(Header).
+
+-spec etag_block_header(aec_headers:header()) -> binary().
+etag_block_header(Header) ->
+    {ok, Hash} = aec_headers:hash_header(Header),
+    Hex = aeu_hex:bin_to_hex(Hash),
+    list_to_binary(Hex).
```
lucafavatella: As the etag is about the representation of the resource - not necessarily only the resource - I wonder if best practice shall include in the etag also e.g. the hash of the swagger.json. (Of course this applies to all endpoints.)

Though it seems good enough.

apps/aehttp/src/aehttp_cache_etag.erl
```diff
@@ -0,0 +1,174 @@
...
+
+generate('GetGenerationByHeight', Req, State) ->
+    handle_generation_height(Req, State);
+
+generate('GetTransactionByHash', Req, State) ->
+    Hash = cowboy_req:binding(hash, Req),
+    {{strong, Hash}, Req, State};
+
+generate('GetTransactionInfoByHash', Req, State) ->
+    Hash = cowboy_req:binding(hash, Req),
```
lucafavatella: I see the `height` of the `ContractCallObject` is part of the hash so shall be ok.

apps/aehttp/src/aehttp_cache_etag.erl
```diff
@@ -0,0 +1,174 @@
...
+    {{strong, Hash}, Req, State};
+
+generate('GetGenerationByHash', Req, State) ->
+    handle_generation_hash(Req, State);
+
+generate('GetGenerationByHeight', Req, State) ->
+    handle_generation_height(Req, State);
+
+generate('GetTransactionByHash', Req, State) ->
+    Hash = cowboy_req:binding(hash, Req),
```
lucafavatella: Are `block_height` and `block_hash` in the response represented in the hash?

apps/aehttp/src/aehttp_cache_etag.erl
dincho: No, this has to be addressed by the API versioning. We already have backward compatibility policy set, it must enforce major version (in the URL as of now v2) bumps if there is backward incompatible changes.

In regards to backward compatible changes, that is new fields only, they will be available once the cache expires.

apps/aehttp/src/aehttp_cache_etag.erl
dincho: To be honest I didn't looked into details. I expect the hash to be unique and immutable per transaction.

apps/aehttp/src/aehttp_api_handler.erl
dincho: Yeah, I prefer that as well. However I wasn't sure what config file should be used thus is hardcoded constant check in the source.

Changed in 8f2df327a12efbcbd8dcb1f5fc9c2abbd4f66db1

apps/aehttp/test/aehttp_integration_SUITE.erl
dincho: Etag integration test added in d477b34c693b25d9ae3496333ff37b1133e198a8

PR #2551, opened 2019-07-01
Pt 166927306 names as strings

gorillainduction:
https://www.pivotaltracker.com/story/show/166927306

### DIFF
--- a/apps/aecontract/src/aect_contracts.erl
+++ b/apps/aecontract/src/aect_contracts.erl
@@ -112,13 +112,16 @@
 -spec is_legal_call(FromVersion :: version(), ToVersion :: version()) ->
                           boolean().
 %% NOTE: Keep this up to date if the different sophia vms gets incompatible.
-is_legal_call(X, X) -> true;
-is_legal_call(#{vm := ?VM_AEVM_SOPHIA_2, abi := X},
-              #{vm := ?VM_AEVM_SOPHIA_1, abi := X}) -> true;
-is_legal_call(#{vm := ?VM_AEVM_SOPHIA_3, abi := X},
-              #{vm := ?VM_AEVM_SOPHIA_2, abi := X}) -> true;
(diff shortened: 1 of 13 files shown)


apps/aecontract/src/aect_contracts.erl
```diff
@@ -145,14 +148,21 @@ is_legal_version_in_protocol(create, #{vm := ?VM_AEVM_SOPHIA_2, abi := ?ABI_AEVM
...
         ?FORTUNA_PROTOCOL_VSN -> true;
-        ?LIMA_PROTOCOL_VSN    -> true %% TODO: If you bump to VM_AEVM_SOPHIA_3 please turn this off!
+        ?LIMA_PROTOCOL_VSN    -> false
+    end;
+is_legal_version_in_protocol(create, #{vm := ?VM_AEVM_SOPHIA_4, abi := ?ABI_AEVM_SOPHIA_1}, ProtocolVersion) ->
+    case ProtocolVersion of
+        ?ROMA_PROTOCOL_VSN    -> false;
+        ?MINERVA_PROTOCOL_VSN -> false;
+        ?FORTUNA_PROTOCOL_VSN -> false;
+        ?LIMA_PROTOCOL_VSN    -> true %% TODO: If you bump to VM_AEVM_SOPHIA_4 please turn this off!
```
hanssv: The comment is symmetric to the previous code, but I am not sure what it means?

hanssv:
Needs a rebase...

apps/aecontract/src/aect_contracts.erl
gorillainduction: It i a bit cryptic, I agree. It was originally intended as a reminder to turn off creation of contracts with old VM:s but it doesn't read well.

hanssv: Shouldn't it at least say `SOPHIA_5` then?

PR #2552, opened 2019-07-03
added initial version of stratum operator document

skkw:


docs/stratum.md
```diff
@@ -0,0 +1,123 @@
+AESTRATUM application (alpha)
+==========
+
+Introduction
+==========
+
+By setting of the configuration value `stratum -> enabled` to `true`, the node also starts the aestratum app.
```
sennui: Can you extend initial paragraph? 

Specify that stratum protocol coordinates pool server and pool clients. Pool server doesn't mine and this is how we deliver it as well (so mining and stratum is exclusive in ae)

docs/stratum.md
```diff
@@ -0,0 +1,123 @@
...
+When miner finds the result artefacts of the cuckoo cycle algorithm - nonce and proof of work, the miner is awarded a share in a mining epoch happening between two subsequent key blocks of the chain.
+A share is understood as a smallest unit of eligible right to be paid out if the share was created in the mining epoch to be rewarded.
+If the difficulty of the solution matches the difficulty of the main network, we can combine the block candidate with the computed nonce and proof of work, construct a next block and publish it.
+
+If the network accepts our new block, the node which constructed that next block will receive a mining reward. This reward will be used for paying out of the participants which submitted shares during mining epochs.
+When straum app notices that the latest keyblock changed, stratum looks back 180 epochs behind top keyblock and locates two (configurable) subsequent mining epochs.
+Miners which contributed their shares to these mining epochs will be paid out proportionally to their contribution, taking also miner difficulty into account.
```
sennui: Please provide more details on payment strategy. Current state which is mainnet contract common for all. How often payment happens in real time? 

Also, I think the intro may end with proposal of next changes.

sennui:
Please add a section here linking stratum config: https://github.com/aeternity/aeternity/blob/master/docs/configuration.md

Do you mention that stratum beneficiary overrides node beneficiary if present?

@skkw ^^

dincho:
How this new configuration options correlate with `mining > autostart` ?

dincho:
> Do you mention that stratum beneficiary overrides node beneficiary if present?

If I understand correctly (not much clear to me), the actual block reward beneficiary is an intermediate account? I guess that's what `keys -> dir` option is for, however it's not clear if the stratum server generate that keypair or the operator have to provide it ?

After the reward is actually received in the intermediate account, after that the reward is split to pool operator (stratum -> beneficiaries) and the miners (workers).

I didn't looked into the client yet, but I suppose the client can set their own beneficiary accounts to receive the split rewards and that's communicate over the stratum protocol to the server?

docs/stratum.md
```diff
@@ -0,0 +1,123 @@
...
+The server sends jobs to the clients, they try to solve them and send their results (called shares) back. The server keeps targets for each client in line with its computational power and distributes rewards.
+When miner finds the result artefacts of the cuckoo cycle algorithm - nonce and proof of work, the miner is awarded a share in a mining epoch happening between two subsequent key blocks of the chain.
+A share is understood as a smallest unit of eligible right to be paid out if the share was created in the mining epoch to be rewarded.
+If the difficulty of the solution matches the difficulty of the main network, we can combine the block candidate with the computed nonce and proof of work, construct a next block and publish it.
+
+If the network accepts our new block, the node which constructed that next block will receive a mining reward. This reward will be used for paying out of the participants which submitted shares during mining epochs.
+When straum app notices that the latest keyblock changed, stratum looks back 180 epochs behind top keyblock and locates two (configurable) subsequent mining epochs.
```
dincho: I guess `180` magic number is `BENEFICIARY_REWARD_DELAY` - it's worth mentioning that.

docs/stratum.md
```diff
@@ -0,0 +1,123 @@
...
+(default 2)
+
+`beneficiaries` - a list in "account_address:percent_share" format - denoting the pool operator addresses with their respective shares. The sum of shares must be less than or equal to 100.
+The sum of shares equal to 100 effectively means that all rewards coming from mining are distributed to pool operators, while miners don't receive any reward.
+
+`keys -> dir` - points to a location where to store the stratum account keypair. Since we want to send tokens to miners from this account, we need to have access to its private key.
+The stratum account holds reward(s) from mining which are scheduled to be payed out via a payment contract. While it may sound tempting, stratum account should never have any significant accumulated wealth since it only serves as a intermittent place for receiving of reward, out of which are then the funds distrubuted to miners soon.
+If `dir` represents a relative path, the directory will be in `priv` directory of aestratum app. Providing absolute path allows storing of keypair outside of the aestratum app.
```
dincho: It's worth mentioning that this keys are of intermediate account and should be backed up, otherwise some reward (180 + 2 (default)) can be lost in case of lost keys because of server/disk outage?

docs/stratum.md
skkw: added, although this is not configurable as we want to ensure that mined blocks are not reversed

docs/stratum.md
skkw: added

skkw:
> How this new configuration options correlate with `mining > autostart` ?

Added clarification that `mining > autostart` needs to be set to true

docs/stratum.md
skkw: added

docs/stratum.md
skkw: extended

skkw:
> How this new configuration options correlate with `mining > autostart` ?

Added to the configuration section.

skkw:
> Please add a section here linking stratum config: https://github.com/aeternity/aeternity/blob/master/docs/configuration.md
> 
> Do you mention that stratum beneficiary overrides node beneficiary if present?
> 
> @skkw ^^

It's added now...

PR #2559, opened 2019-07-04
Pt sc ws accept any initiator

uwiger:
See [PT #167089835](https://www.pivotaltracker.com/story/show/167089835)

This PR is based on #2558 (which should be merged first) and adds type checking support in the WS API (`sc_ws_handler.erl`) for setting `"initiator_id"` to `"any"` if `"role": "responder"`. A test case was added in `aehttp_integration_SUITE`.


uwiger:
Rebased against latest master (3.3.0) and added release notes

PR #2560, opened 2019-07-05
Pt 165674014 sc ws handle missing accounts

velzevur:
[SC WS: handle missing accounts](https://www.pivotaltracker.com/story/show/165674014)

This actually grew orthogonally, covering not only missing participants but also invalid numbers as well

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -456,8 +456,8 @@ init_checks(Opts) ->
               end
               ],
     case aeu_validation:run(Checks) of
-        {error, Err} ->
-            {error, Err};
+        {error, _Reason} = Err ->
+            Err;
         ok ->
             ok
     end.
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -420,23 +420,46 @@ set_ongoing(D) ->
 
 initiate(Host, Port, #{} = Opts0) ->
     lager:debug("initiate(~p, ~p, ~p)", [Host, Port, Opts0]),
-    Opts = maps:merge(#{client => self()}, Opts0),
-    case check_amounts(Opts) of
+    Opts = maps:merge(#{client => self(),
+                        role   => initiator}, Opts0),
+    case init_checks(Opts) of
         ok ->
             start_link(#{ host => Host
                         , port => Port
-                        , opts => Opts#{role => initiator} });
+                        , opts => Opts });
         {error, _Reason} = Err -> Err
     end.
 
 respond(Port, #{} = Opts0) ->
     lager:debug("respond(~p, ~p)", [Port, Opts0]),
-    Opts = maps:merge(#{client => self()}, Opts0),
-    case check_amounts(Opts) of
+    Opts = maps:merge(#{client => self(),
+                        role   => responder}, Opts0),
+    case init_checks(Opts) of
         ok ->
             start_link(#{ port => Port
-                        , opts => Opts#{role => responder} });
+                        , opts => Opts });
         {error, _Reason} = Err -> Err
+    end.
+
+init_checks(Opts) ->
+    #{initiator   := Initiator,
+      responder   := Responder,
+      role        := Role,
+      lock_period := LockPeriod} = Opts,
+    Checks = [fun() -> check_amounts(Opts) end,
+              fun() -> check_accounts(Initiator, Responder, Role) end,
+              fun() ->
+                  case LockPeriod < 0 of
+                      true -> {error, lock_period_too_low};
+                      false -> ok
+                  end
+              end
+              ],
+    case aeu_validation:run(Checks) of
+        {error, Err} ->
+            {error, Err};
+        ok ->
+            ok
     end.
 
 upd_transfer(_Fsm, _From, _To, Amount) when Amount < 0 ->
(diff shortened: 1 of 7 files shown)


apps/aehttp/test/aehttp_sc_SUITE.erl
```diff
@@ -2266,7 +2268,11 @@ sc_ws_dont_open_ga(Config) ->
     RAmt = 4,
 
     ChannelOpts = channel_options(IPubkey, RPubkey, IAmt, RAmt, #{}, Config),
-    {error, rejected} = channel_ws_start(initiator, maps:put(host, <<"localhost">>, ChannelOpts), Config),
+    {ok, IPid} = channel_ws_start(initiator, maps:put(host, <<"localhost">>, ChannelOpts), Config),
+    %% although connection is established, it dies instnatly as the
+    %% participant is a GA
+    timer:sleep(100),
+    undefined = process_info(IPid),
```
lucafavatella: Shall this rather be a wait-until with larger timeout so to prevent this to fail intermittently?

Though looks good enough.

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -62,6 +89,12 @@ websocket_handle({text, MsgBin}, #handler{protocol = Protocol,
 websocket_handle(_Data, H) ->
     {ok, H}.
 
+websocket_info(?ERROR_TO_CLIENT(Err), #handler{protocol = Protocol} = H) ->
+    {reply, Resp} = sc_ws_api:error_response(Protocol, Err),
+    timer:sleep(1000),
```
lucafavatella: ?

velzevur: This is when an initial openning settings were erronous - we give some time the client to open the connection on their side so the message is not lost

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -420,25 +420,48 @@ set_ongoing(D) ->
...
 
+init_checks(Opts) ->
+    #{initiator   := Initiator,
+      responder   := Responder,
+      role        := Role,
+      lock_period := LockPeriod} = Opts,
+    Checks = [fun() -> check_amounts(Opts) end,
+              fun() -> check_accounts(Initiator, Responder, Role) end,
+              fun() ->
+                  case LockPeriod < 0 of
```
grzegorz225: Is there any case in which a Lock Period of 0 makes sense?

velzevur: Although not advised to use `LockPeriod = 0` on main net, there is no actual reason to forbid it. We've used it in some examples aready so people get _fast_ channel opening, for example on testnet.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -420,25 +420,48 @@ set_ongoing(D) ->
...
+                      true -> {error, lock_period_too_low};
+                      false -> ok
+                  end
+              end
+              ],
+    case aeu_validation:run(Checks) of
+        {error, Err} ->
```
grzegorz225: Why is the order like this? Shouldn't the first match case be matching with the ok atom like in the rest of the project?

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -420,25 +420,48 @@ set_ongoing(D) ->
...
+            {error, Err};
```
grzegorz225: Why unpack the error reason instead of writing ```{error, _Reason} = Err -> Err```?

apps/aehttp/src/sc_ws_handler.erl
grzegorz225: This case seems rare enough that it should be logged using lager.

apps/aehttp/test/aehttp_sc_SUITE.erl
```diff
@@ -15,6 +15,7 @@
    [sc_ws_timeout_open/1,
     sc_ws_attach_initiator/1,
     sc_ws_attach_responder/1,
+		sc_ws_broken_open_params/1,
```
grzegorz225: Formatting

apps/aechannel/src/aesc_fsm.erl
velzevur: If we have such convention, I am not aware of it :D

apps/aehttp/test/aehttp_sc_SUITE.erl
```diff
@@ -15,7 +15,7 @@
    [sc_ws_timeout_open/1,
     sc_ws_attach_initiator/1,
     sc_ws_attach_responder/1,
-		sc_ws_broken_open_params/1,
+    ws_broken_open_params/1,
```
grzegorz225: Typo, Should be: sc_ws_broken_open_params instead of ws_broken_open_params

apps/aehttp/test/aehttp_sc_SUITE.erl
```diff
@@ -15,6 +15,7 @@
    [sc_ws_timeout_open/1,
     sc_ws_attach_initiator/1,
     sc_ws_attach_responder/1,
+    ws_broken_open_params/1,
```
lucafavatella:
```suggestion
    sc_ws_broken_open_params/1,
```

apps/aehttp/src/sc_ws_handler.erl
velzevur: Addressed

PR #2565, opened 2019-07-09
Improve building scripts for Windows

venimus:
+ Auto detect Visual studio version and paths (allow support of VS2019)
+ Auto detect msys2 and other tools paths
+ Improve docs

Background: Building from windows is very confusing and messy because many of the current tools versions and their installation paths are different. To match the expected environment users need to decode the scripts and scan for paths and version numbers. Older versions are no longer available from the official download links and users need to search for them.

### DIFF
--- a/scripts/windows/msys2_prepare.bat
+++ b/scripts/windows/msys2_prepare.bat
@@ -1,4 +1,4 @@
-@echo on
+@echo off
 @rem Script to prepare msys2 environment for builds.
 @rem Required vars:
 @rem    ERTS_VERSION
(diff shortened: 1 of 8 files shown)


scripts/windows/msys2_prepare.bat
```diff
@@ -1,4 +1,4 @@
-@echo on
+@echo off
```
tolbrino: I found this to be `on` useful when debugging build issues.

venimus: Actually it was `on` but all commands (except few) were implicitly 'muted', so it didn't change the output. I turned it `on` to not change the behavior, but removed `@` so it would be easier to just change this line to toggle the echo.

It could be made to be controlled from the command line though

venimus: I plan to revise those scripts again, so they are more robust and compatible with the newly introduced CircleCI windows builds, but I had hard time to debug those, and are currently postponed. So may be I will  include optional flag to turn echo on with it

PR #2567, opened 2019-07-09
Return gas used in dry run of contract create tx

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/167168077

### DIFF
--- a/apps/aehttp/test/aehttp_dryrun_SUITE.erl
+++ b/apps/aehttp/test/aehttp_dryrun_SUITE.erl
@@ -114,7 +114,9 @@ spend_txs(Config) ->
     Tx1 = create_spend_tx(APub, EPub, 100000 * aec_test_utils:min_gas_price(), 20000 * aec_test_utils:min_gas_price(), 1, 100),
     Tx2 = create_spend_tx(EPub, APub, 100, 20000 * aec_test_utils:min_gas_price(), 1, 100),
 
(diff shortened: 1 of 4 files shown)


apps/aehttp/test/aehttp_dryrun_SUITE.erl
```diff
@@ -114,7 +114,9 @@ spend_txs(Config) ->
     Tx1 = create_spend_tx(APub, EPub, 100000 * aec_test_utils:min_gas_price(), 20000 * aec_test_utils:min_gas_price(), 1, 100),
     Tx2 = create_spend_tx(EPub, APub, 100, 20000 * aec_test_utils:min_gas_price(), 1, 100),
 
-    {ok, 200, #{ <<"results">> := [#{ <<"result">> := <<"ok">> }, #{ <<"result">> := <<"ok">> }] }} =
+    {ok, 200, #{ <<"results">> := [#{ <<"result">> := <<"ok">>,
+                                      <<"type">> := <<"spend">> },
+                                   #{ <<"result">> := <<"ok">> }] }} =
```
uwiger: Does the second map not also contain `<<"type">> := <<"spend">>`?

lucafavatella: Yes. I tested only on the first one.

PR #2570, opened 2019-07-11
Add type checking of tail call return values in FATE

gorillainduction:
This is a crude solution of the problem stated in https://www.pivotaltracker.com/story/show/164302162
In particular, it reduces the tail call constant stack optimization.

We should tackle this in a more sophisticated way, but now at least it works. Creating a new PT ticket to make this properly.

### DIFF
--- a/apps/aecontract/test/aecontract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_SUITE.erl
@@ -5200,9 +5200,18 @@ call_wrong_type(_Cfg) ->
     Contract1 = ?call(create_contract, Acc1, remote_type_check, {}),
     Contract2 = ?call(create_contract, Acc1, remote_type_check, {}),
     42        = ?call(call_contract, Acc1, Contract1, remote_id, word, {?cid(Contract2), 42}),
-    {error, Error} = ?call(call_contract, Acc1, Contract1, remote_wrong_type, word,
-                           {?cid(Contract2), <<"hello">>}),
-    ?assertMatchVM(<<"out_of_gas">>, <<"Type error on call:", _/binary>>, Error),
+    {error, Error1} = ?call(call_contract, Acc1, Contract1, remote_wrong_arg, string,
+                            {?cid(Contract2), <<"hello">>}),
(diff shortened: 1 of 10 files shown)


apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -257,7 +258,7 @@ all() ->
 
 groups() ->
     [ {aevm, [], ?ALL_TESTS ++ ?AEVM_SPECIFIC}
-    , {fate, [], ?ALL_TESTS ++ ?FATE_SPECIFIC}
+    , {fate, [], [{group, sophia}]}%%?ALL_TESTS ++ ?FATE_SPECIFIC}
```
lucafavatella: Is this intended?

gorillainduction: ah, no :man_facepalming:

apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -1146,21 +1146,21 @@ remote_gas_test_contract(Config) ->
     call_get(APub, APriv, EncC2Pub, Contract, 100),
     force_fun_calls(Node),
     Balance1 = get_balance(APub),
-    ?assertMatchVM(1600476, 1600010, (Balance0 - Balance1) div ?DEFAULT_GAS_PRICE),
+    ?assertMatchVM(1600596, 1600010, (Balance0 - Balance1) div ?DEFAULT_GAS_PRICE),
```
lucafavatella: I need to understand why these numbers changed for Sophia if the PR is for FATE only. 🔬

gorillainduction: I added functions to the contract, and the way AEVM works is that the code layout affects the gas cost because of the function dispatch in the beginning of each contract,

While I didn't verify that this corresponds exactly to these changed numbers, the numbers changed are in the correct range.

lucafavatella: Yes, I see contract `remote_gas_test` changed.

PR #2572, opened 2019-07-15
Support multiple SC responder keys per port

uwiger:
See [PT #167194534](https://www.pivotaltracker.com/story/show/167194534)

Note: at initial push, there are no test cases testing multiple responder keys listening on the same port. Nor are there tests for the logic that is supposed to kick in if an accept times out.

### DIFF
--- a/apps/aechannel/src/aesc_listeners.erl
+++ b/apps/aechannel/src/aesc_listeners.erl
@@ -232,5 +232,5 @@ get_lsock_info(Items, I, St) when is_list(Items) ->
               get_lsock_info(Item, Acc, St)
       end, I, Items);
 get_lsock_info(all, I, St) ->
-    get_lsock_info([pids, responder], I, St).
+    get_lsock_info([pids, responders], I, St).
 
--- a/apps/aechannel/src/aesc_listeners.erl
+++ b/apps/aechannel/src/aesc_listeners.erl
@@ -22,13 +22,15 @@
         , record_fields/1]).
 
 -record(st, {
-              ports = new_tab(aesc_listeners_ports)
+              responders = new_tab(aesc_listeners_responders)
+            , ports = new_tab(aesc_listeners_ports)
             , socks = new_tab(aesc_listener_socks)
             , refs  = new_tab(aesc_listeners_refs)
             }).
--record(port, {key, responder, mref, lsock}).
+-record(resp, {key}).
+-record(port, {key, mref, lsock}).
 -record(sock, {key, port}).
--record(ref , {mref, port, lsock, pid}).
+-record(ref , {mref, port, lsock, responder, pid}).
 
 -define(SERVER, ?MODULE).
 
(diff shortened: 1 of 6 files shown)


apps/aechannel/src/aesc_listeners.erl
```diff
@@ -177,63 +191,41 @@ maybe_close_lsock(Port, LSock, #st{ ports = Ports, socks = Socks } = St) ->
...
 get_lsock_info(pids, #{port := Port} = I, #st{ ports = Ports }) ->
     I#{pids => list_pids(Ports, Port)};
-get_lsock_info(responder, #{port := Port} = I, #st{ ports = Ports }) ->
-    case db_lookup(Ports, {Port, 0}) of
-        [#port{responder = R}] ->
-            I#{responder => R};
-        [] ->
-            I
-    end;
+get_lsock_info(responders, #{port := Port} = I, #st{ responders = Resps }) ->
```
lucafavatella: Could you please review `responder` in the `all` clause below?

apps/aechannel/src/aesc_listeners.erl
```diff
@@ -103,18 +109,23 @@ handle_call({listen, Port, Responder, Opts}, {Pid,_Ref}, #st{ ports = Ports
                     {reply, {exception, Reason}, St}
             end
     end;
-handle_call({close, LSock}, {Pid,_Ref}, #st{ socks = Socks
+handle_call({close, LSock}, {Pid,_Ref}, #st{ responders = Resps
```
lucafavatella: Who calls `aesc_listeners:close/1`? Nobody? Shall this piece of API be deleted?

apps/aechannel/src/aesc_listeners.erl
lucafavatella: Shall that `all` piece of functionality be deleted?

apps/aechannel/src/aesc_session_noise.erl
```diff
@@ -211,10 +211,16 @@ accept_tcp(LSock, Port, R, AcceptTimeout) ->
         {ok, _} = Ok ->
```
lucafavatella: I see the line above (`gen_tcp:accept`) and the origin of the listening socket `LSock` mean that concurrent `gen_tcp:accept` may be called on the same listening socket. I read that [that is ok in recent OTP](https://github.com/erlang/otp/blob/OTP-20.1/lib/kernel/doc/src/gen_tcp.xml#L128-L133). So the implementation seem sound.

apps/aechannel/src/aesc_listeners.erl
uwiger: > Who calls `aesc_listeners:close/1`? Nobody? Shall this piece of API be deleted?

I'm undecided. It seems symmetric to have it. Ok if we keep it around for a while? It might come in handy.

apps/aechannel/src/aesc_listeners.erl
uwiger: Let's keen the `all` option available for now, e.g. for debugging purposes. But I've pushed a fix for the `responder -> responders` bug.

lucafavatella:
@uwiger I noticed a CI failure in this PR that though seems unrelated - I took the liberty of [tracking it](https://www.pivotaltracker.com/story/show/167317380).

uwiger:
Rebased on top of Dimitar's latest changes.
The `channel_open` and `channel_accept` protocol messages have been extended to include both pubkeys, to prevent mixups when pairing acceptors and responders. This is an off-chain protocol-breaking change.

Some test failures relating to GA occur when running tests locally. I've been unable to locate the cause tonight.

uwiger:
Superseded by #2578 (merged), so closing.

PR #2576, opened 2019-07-16
SC: allow generic messages in any FSM state

velzevur:
[Handle SC inband messages in any state](https://www.pivotaltracker.com/story/show/165563336)

Protocol PR will follow shortly

### DIFF
--- a/docs/release-notes/next/PT-165563336_SC_generic_messages_in_every_state.md
+++ b/docs/release-notes/next/PT-165563336_SC_generic_messages_in_every_state.md
@@ -1,4 +1,5 @@
 * Allows sending and receiving generic messages in State Channels in any FSM
   state. Until this - generic messages were allowed only in `open` state.
   `channel_id` is part of the message body and if it is unknown - the
-  temporary one is used instead. This is breaking off-chain consesus.
+  temporary one is used instead. This is not backwards compatible for the
+  `noise` protocol. This enhances the WebSocket API accordingly.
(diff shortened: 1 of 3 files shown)


docs/release-notes/next/PT-165563336_SC_generic_messages_in_every_state.md
```diff
@@ -0,0 +1,4 @@
+* Allows sending and receiving generic messages in State Channels in any FSM
+  state. Until this - generic messages were allowed only in `open` state.
+  `channel_id` is part of the message body and if it is unknown - the
+  temporary one is used instead. This is breaking off-chain consesus.
```
lucafavatella: It is unclear to me what it is meant as "off-chain consensus".

Based on what I see in code, the items affected are called "inband messages" and I see them defined as part of the protocol  inside the channel Noise endpoint among two nodes, as [message code 96](https://github.com/aeternity/protocol/blob/3a7b1bdf1b2fa9dee98ac0788ca084b654912c84/channels/OFF-CHAIN.md#inband_message):
> Inband messages are arbitrary messages sent between the channel participants. The payload must be limited to 65,535 bytes. The fsm must deliver an inband message immediately, and the receiver must process it (e.g. conveying it to the client) immediately, preserving the message ordering.
>
> One possible use of inband messages could be to synchronize off-chain state updates, but any application is allowed.

Is what is broken by this change:
- The protocol in the channels Noise endpoint among two nodes; plus
- The user channels WebSocket API

?

velzevur: > The protocol in the channels Noise endpoint among two nodes; plus

This is broken. This also is the off-chain (noise) protocol between the nodes/FSMs

> The user channels WebSocket API

This is not broken but enhanced - consumers of the WebSocket API are now able to send and receive messages while the channel is being updated.

velzevur:
This is already defined in [protocol](https://github.com/aeternity/protocol/blob/master/node/api/channels_api_usage.md#generic-messages)
> There is the functionality to send participants messages .... These roles can be taken by any of the participants, **anytime**.

So no complementary `protocol` PR

PR #2580, opened 2019-07-17
Bump CI to recognize tag for v4

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/167061956

See also #2442


lucafavatella:
No way this can fail CI on non-tag. Merging.

PR #2584, opened 2019-07-18
Update channels system tests after major release

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/167128385

Follows https://github.com/aeternity/aeternity/commit/e164fc4518263db9692c02a9b84e179d69bfcc13#diff-9ccf6fa61d67a8da333df6695d09e1e5

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -308,7 +308,7 @@ system-test-deps:
 	$(MAKE) system-smoke-test-deps
 	docker pull "aeternity/aeternity:v2.1.0"
 	docker pull "aeternity/aeternity:v2.3.0"
-	docker pull "aeternity/aeternity:v3.1.0"
+	docker pull "aeternity/aeternity:v4.0.0"
 	docker pull "aeternity/aeternity:latest"
 
 system-test: KIND=system_test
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -308,7 +308,7 @@ system-test-deps:
 	$(MAKE) system-smoke-test-deps
 	docker pull "aeternity/aeternity:v2.1.0"
 	docker pull "aeternity/aeternity:v2.3.0"
-	docker pull "aeternity/aeternity:v3.1.0"
```
tolbrino: This version is updated because the new major release is not a hard fork and therefore no extra image is needed?

lucafavatella: This is not about hard fork (re consensus protocol) but about channels APIs i.e. the node-to-client-WebSocket and the node-to-node-Noise.

Before this commit, this pull of v3.1.0 was unused as the corresponding tests were commented out.

The tests meant to use this line test that the version of the node built from the codebase passes a smoke test re compatibility of channel APIs (WebSocket and Noise) with the oldest compatible release. That is 4.0.0 hence the change.

tolbrino: Got it. Thanks.

PR #2585, opened 2019-07-19
PT-164691690 Allow mutual close after solo close/slash

grzegorz225:
No changes in aeternity/protocol repo are needed.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -97,6 +97,7 @@
 
 -include("aesc_codec.hrl").
 %% -include_lib("trace_runner/include/trace_runner.hrl").
+-include_lib("aecontract/include/hard_forks.hrl").
 
 -type role() :: initiator | responder.
 -type sign_tag() :: create_tx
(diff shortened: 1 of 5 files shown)


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1061,6 +1062,14 @@ awaiting_signature(cast, {?SIGNED, settle_tx, SignedTx} = Msg,
                 settle_signed(SignedTx, Updates, D1)
         end, D);
 
+% Timeouts
```
uwiger: A minor quibble, but use two %% at the beginning of a line. The Emacs indentation mode will try to indent a single %

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2496,6 +2535,11 @@ adjust_ttl(TTL) when is_integer(TTL), TTL >= 0 ->
 cur_height() ->
     aec_headers:height(aec_chain:top_header()).
 
+is_fork_active(ProtocolVSN) ->
+    CurHeight = max(cur_height(), ?MINIMUM_DEPTH),
+    %% Enable a new fork only after MINIMUM_DEPTH generations in order to avoid fork changes
+    ProtocolVSN == aec_hard_forks:protocol_effective_at_height(CurHeight - ?MINIMUM_DEPTH).
```
uwiger:
```suggestion
    ProtocolVSN == aec_hard_forks:protocol_effective_at_height(max(cur_height() - ?MINIMUM_DEPTH, aec_block_genesis:height())).
```

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2496,6 +2535,11 @@ adjust_ttl(TTL) when is_integer(TTL), TTL >= 0 ->
 cur_height() ->
     aec_headers:height(aec_chain:top_header()).
 
+is_fork_active(ProtocolVSN) ->
+    CurHeight = max(cur_height(), ?MINIMUM_DEPTH),
```
uwiger:
```suggestion
```

uwiger: It feels counter-intuitive to call this `CurHeight` when it possibly isn't.

apps/aechannel/src/aesc_fsm.erl
grzegorz225: Ok, good to know. BTW a quick grep of the codebase ```grep -r -e "^% " apps | wc -l ``` shows that there are 66 occurrences of a single % at the beginning of a line.

PR #2588, opened 2019-07-19
Improve FSM in order to compute fees dynamically

velzevur:
[Integration test plain.sc_ws_slash failed on CI -- max_reached](https://www.pivotaltracker.com/story/show/167254422)
The issue was caused by FSM using hardcoded fee of `30 000 * local gas price` (or 33 000) for all transactions. In certain cases this was insufficant and transaction was rejected.

This PR introduces computing the fee according the transaction instead. The only transaction type that still uses hardcoded fee is `channel_mutual_close_tx` as the fee is important component in participant amounts' computation.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -1285,7 +1285,7 @@ new_onchain_tx_for_signing(Type, Opts, D) ->
 new_onchain_tx_for_signing_(Type, Opts, D) ->
     Defaults = tx_defaults(Type, Opts, D),
     Opts1 = maps:merge(Defaults, Opts),
-    CurrHeight = cur_height(),
+    CurrHeight = curr_height(),
     {ok, Tx, Updates} = new_onchain_tx(Type, Opts1, D, CurrHeight),
     case {aetx:min_fee(Tx, CurrHeight), aetx:fee(Tx)} of
         {MinFee, Fee} when MinFee =< Fee ->
@@ -1580,10 +1580,10 @@ get_nonce(Pubkey) ->
 %% we have an actual transaction record (required by aetx:min_fee/2).
 %% The default should err on the side of being too low.
 default_fee(_Tx) ->
-    CurHeight = aec_headers:height(aec_chain:top_header()),
+    CurrHeight = aec_headers:height(aec_chain:top_header()),
     %% this could be fragile on hard fork height if one participant's node had
     %% already forked and the other had not yet
-    20000 * max(aec_governance:minimum_gas_price(CurHeight),
+    20000 * max(aec_governance:minimum_gas_price(CurrHeight),
                 aec_tx_pool:minimum_miner_gas_price()).
 
 default_ttl(_Type, Opts, #data{opts = DOpts}) ->
@@ -1591,12 +1591,12 @@ default_ttl(_Type, Opts, #data{opts = DOpts}) ->
     Opts#{ttl => adjust_ttl(TTL)}.
 
 adjust_ttl(undefined) ->
-    CurHeight = aec_headers:height(aec_chain:top_header()),
-    CurHeight + 100;
+    CurrHeight = aec_headers:height(aec_chain:top_header()),
+    CurrHeight + 100;
 adjust_ttl(TTL) when is_integer(TTL), TTL >= 0 ->
     TTL.
 
-cur_height() ->
+curr_height() ->
     aec_headers:height(aec_chain:top_header()).
 
 
@@ -2962,7 +2962,7 @@ has_gproc_key(Fsm, #{gproc_key := K}) ->
 
 is_channel_locked(0) -> false;
 is_channel_locked(LockedUntil) ->
-    LockedUntil >= cur_height().
+    LockedUntil >= curr_height().
 
 withdraw_locked_complete(SignedTx, Updates, #data{state = State, opts = Opts} = D) ->
     {OnChainEnv, OnChainTrees} =
@@ -3443,4 +3443,4 @@ init_checks(Opts) ->
 
 is_fork_active(ProtocolVSN) ->
     %% Enable a new fork only after MINIMUM_DEPTH generations in order to avoid fork changes
-    ProtocolVSN == aec_hard_forks:protocol_effective_at_height(max(cur_height() - ?MINIMUM_DEPTH, aec_block_genesis:height())).
+    ProtocolVSN == aec_hard_forks:protocol_effective_at_height(max(curr_height() - ?MINIMUM_DEPTH, aec_block_genesis:height())).
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -1458,15 +1458,15 @@ new_onchain_tx_(Mod, Opts, CurrHeight) ->
         true -> % use preset fee
             apply(Mod, new, [Opts]);
         false ->
-            create_with_minuimum_fee(Mod, Opts#{fee => 0}, CurrHeight)
+            create_with_minimum_fee(Mod, Opts#{fee => 0}, CurrHeight)
     end.
 
-create_with_minuimum_fee(Mod, Opts, CurrHeight) ->
-    create_with_minuimum_fee(Mod, Opts, CurrHeight, 5).
+create_with_minimum_fee(Mod, Opts, CurrHeight) ->
+    create_with_minimum_fee(Mod, Opts, CurrHeight, 5).
 
-create_with_minuimum_fee(_, _, _, Attempts) when Attempts < 1 ->
+create_with_minimum_fee(_, _, _, Attempts) when Attempts < 1 ->
     erlang:error(could_not_compute_fee);
-create_with_minuimum_fee(Mod, Opts, CurrHeight, Attempts) ->
+create_with_minimum_fee(Mod, Opts, CurrHeight, Attempts) ->
     {ok, Tx} = apply(Mod, new, [Opts]),
     MinTxFee = aetx:min_fee(Tx, CurrHeight),
     MinGas = aetx:min_gas(Tx, CurrHeight),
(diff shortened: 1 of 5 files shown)


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2292,23 +2290,113 @@ new_onchain_tx(channel_withdraw_tx, #{acct := ToId,
...
+
+
+
+
+new_onchain_tx_(Mod, Opts, CurrHeight) ->
+    case maps:is_key(fee, Opts) of
+        true -> % use preset fee
+            apply(Mod, new, [Opts]);
+        false ->
+            create_with_minuimum_fee(Mod, Opts#{fee => 0}, CurrHeight)
```
lucafavatella: s/minuimum/minimum/g though not functionally relevant :)

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2292,23 +2290,113 @@ new_onchain_tx(channel_withdraw_tx, #{acct := ToId,
...
+    end.
+
+create_with_minuimum_fee(Mod, Opts, CurrHeight) ->
+    create_with_minuimum_fee(Mod, Opts, CurrHeight, 5).
```
lucafavatella: I see 5 attempts.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2292,23 +2290,113 @@ new_onchain_tx(channel_withdraw_tx, #{acct := ToId,
...
+    MinGas = aetx:min_gas(Tx, CurrHeight),
+    MinMinerGasPrice = aec_tx_pool:minimum_miner_gas_price(),
+    MinGasRequirements = MinGas * MinMinerGasPrice,
+    MinFee = max(MinTxFee, MinGasRequirements),
+    TxFee = aetx:fee(Tx),
+    case MinFee =< TxFee of
+        true ->
+            {ok, Tx};
+        false ->
+            create_with_minuimum_fee(Mod, Opts#{fee => MinFee}, CurrHeight,
```
lucafavatella: I see the only argument that changes in the recursion is the fee. I understand this is because a larger fee implies a larger serialization.

So 5 attempts seems enough.

velzevur: Yes, I don't see how it could possibly go beyond 3 attempts, but put 5 anyway

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2199,37 +2199,36 @@ new_onchain_tx_for_signing(Type, Opts, D) ->
     end.
 
 new_onchain_tx_for_signing_(Type, Opts, D) ->
-    FeeSpecified = maps:is_key(fee, Opts),
     Defaults = tx_defaults(Type, Opts, D),
     Opts1 = maps:merge(Defaults, Opts),
-    {ok, Tx, Updates} = new_onchain_tx(Type, Opts1, D),
-    CurHeight = cur_height(),
-    case {aetx:min_fee(Tx, CurHeight), aetx:fee(Tx)} of
+    CurrHeight = cur_height(),
```
grzegorz225:
```
$ grep -r -o -e "cur_height" apps | wc -l 
4
 $ grep -r -o -e "curr_height" apps | wc -l
0
 $ grep -r -o -e "CurrHeight" apps | wc -l 
44
 $ grep -r -o -e "CurHeight" apps | wc -l 
13
```
There should be a ticket to make the variable names consistent.

velzevur: yes, we are rather inconsistent in that regard

velzevur: Since almost all occurrences were of `CurrHeight`, I've replaced all `CurHeight` accordingly.

velzevur:
As discussed, this must be based on #2591. Adding a `WIP` label.

velzevur:
Rebased on top of recent `master` that includes changes introduced in #2591. I am removing `WIP` label

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1376,23 +1374,113 @@ new_onchain_tx(channel_withdraw_tx, #{acct := ToId,
...
+                                         {account, Responder}], State),
+    Opts2 =
+        Opts1#{ channel_id => aeser_id:create(channel, ChanId)
+              , from_id    => aeser_id:create(account, Account)
+              , poi        => Poi
+              , ttl        => TTL},
+    {ok, Tx} = new_onchain_tx_(aesc_slash_tx, Opts2, CurrHeight),
+    {ok, Tx, []}.
+
+
```
tolbrino: Why so many empty lines?

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1561,23 +1580,23 @@ get_nonce(Pubkey) ->
 %% we have an actual transaction record (required by aetx:min_fee/2).
 %% The default should err on the side of being too low.
 default_fee(_Tx) ->
-    CurHeight = aec_headers:height(aec_chain:top_header()),
+    CurrHeight = aec_headers:height(aec_chain:top_header()),
     %% this could be fragile on hard fork height if one participant's node had
     %% already forked and the other had not yet
-    30000 * max(aec_governance:minimum_gas_price(CurHeight),
+    20000 * max(aec_governance:minimum_gas_price(CurrHeight),
```
tolbrino: Can we move this rather static value into `aesc_fsm.hrl`?

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1561,23 +1580,23 @@ get_nonce(Pubkey) ->
...
 
 default_ttl(_Type, Opts, #data{opts = DOpts}) ->
     TTL = maps:get(ttl, Opts, maps:get(ttl, DOpts, 0)),
     Opts#{ttl => adjust_ttl(TTL)}.
 
 adjust_ttl(undefined) ->
-    CurHeight = aec_headers:height(aec_chain:top_header()),
-    CurHeight + 100;
+    CurrHeight = aec_headers:height(aec_chain:top_header()),
+    CurrHeight + 100;
```
tolbrino: Can we move this rather static value into `aesc_fsm.hrl`?

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1376,23 +1374,113 @@ new_onchain_tx(channel_withdraw_tx, #{acct := ToId,
...
+
+new_onchain_tx_(Mod, Opts, CurrHeight) ->
+    case maps:is_key(fee, Opts) of
+        true -> % use preset fee
+            apply(Mod, new, [Opts]);
+        false ->
+            create_with_minimum_fee(Mod, Opts#{fee => 0}, CurrHeight)
+    end.
+
+create_with_minimum_fee(Mod, Opts, CurrHeight) ->
```
tolbrino: Please document the reasoning for the iterative approach and 5 attempts as limit here.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1376,23 +1374,113 @@ new_onchain_tx(channel_withdraw_tx, #{acct := ToId,
...
+              , from_id    => aeser_id:create(account, Account)
+              , poi        => Poi
+              , ttl        => TTL},
+    {ok, Tx} = new_onchain_tx_(aesc_slash_tx, Opts2, CurrHeight),
+    {ok, Tx, []}.
+
+
+
+
+new_onchain_tx_(Mod, Opts, CurrHeight) ->
```
tolbrino: This should have a guard making sure that `Mod` is a supported value. That way we will find out at compile-time and test-time if we use it wrong.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1376,23 +1374,113 @@ new_onchain_tx(channel_withdraw_tx, #{acct := ToId,
...
-    Opts2 = Opts1#{ state_hash    => StateHash
-                  , initiator_id  => aeser_id:create(account, Initiator)
-                  , responder_id  => aeser_id:create(account, Responder)
-                  },
-    {ok, CreateTx} = aesc_create_tx:new(Opts2),
-    {ok, CreateTx, []}. % no updates
+    {ok, CreateTx} = new_onchain_tx_(aesc_create_tx, Opts1, CurrHeight),
+    {ok, CreateTx, []}; % no updates
+new_onchain_tx(channel_settle_tx, Opts,
+               #data{ opts  = #{initiator := I,
```
tolbrino: We should decide which formatting to use in case of multi-line maps and records. 6 lines above the alternative is used.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1283,37 +1283,36 @@ new_onchain_tx_for_signing(Type, Opts, D) ->
...
 new_onchain_tx(channel_close_mutual_tx, #{ acct := From } = Opts,
-               #data{opts = DOpts, on_chain_id = Chan, state = State}) ->
+               #data{opts = DOpts, on_chain_id = Chan, state = State}, _) ->
     #{initiator := Initiator,
       responder := Responder} = DOpts,
     ChanId = aeser_id:create(channel, Chan),
     FromId = aeser_id:create(account, From),
     {ok, IAmt} = aesc_offchain_state:balance(Initiator, State),
     {ok, RAmt} = aesc_offchain_state:balance(Responder, State),
+    % fee is preset. This is important for participant's amounts calculation
```
uwiger: Use `%%` instead of `%` when comment is expected to line up with existing indentation. Otherwise, Emacs will send it to far-right if asked to fix indentation.

apps/aechannel/src/aesc_fsm.erl
uwiger: Redundant whitespace

apps/aechannel/src/aesc_fsm.erl
velzevur: Addressed

apps/aechannel/src/aesc_fsm.erl
velzevur: Tracked as [Decide on multi-line formatting](https://www.pivotaltracker.com/story/show/167484806)

PR #2591, opened 2019-07-22
[PT-166471712] Add state diagrams for state channel fsm

tolbrino:
This is a draft for updating the FSM state diagram documentation.

The included source code changes are merely cosmetic to make the structure easier to understand. No logic has been changed.


docs/state-channels/fsm_initiator.puml
```diff
@@ -0,0 +1,57 @@
+@startuml
```
lucafavatella: Would it be possible to automate as Makefile the generation of this plantuml file, and hook that to CI?

docs/state-channels/fsm_responder.puml
```diff
@@ -0,0 +1,48 @@
+@startuml
```
lucafavatella: Would it be possible to automate as Makefile the generation of this plantuml file, and hook that to CI?

Makefile
```diff
@@ -512,6 +520,14 @@ $(DEB_PKG_CHANGELOG_FILE):
 prod-deb-package: $(DEB_PKG_CHANGELOG_FILE)
 	debuild --preserve-envvar DEB_SKIP_DH_AUTO_CLEAN -b -uc -us
 
+$(PLANTUML_JAR):
+	curl -fsS --create-dirs -o $@ https://netcologne.dl.sourceforge.net/project/plantuml/${PLANTUML_V}/plantuml.${PLANTUML_V}.jar
+
+build-uml: ${PLANTUML_JAR} ${uml-svg-files}
```
lucafavatella: Would it be possible to hook this (puml -> svg) to CI so to prevent it not to be working in the future and go unnoticed?

tolbrino: Sure, we could do that.

.circleci/config.yml
```diff
@@ -946,6 +945,20 @@ jobs:
...
 
+  build_uml_diagrams:
+    executor: builder_container_stable
+    <<: *container_config
+    steps:
+      - checkout
+      - run:
+          name: Install PlantUML dependencies
+          command: |
+            sudo apt-get -qq -y update && sudo apt-get -qq -y install graphviz librsvg2-bin
```
lucafavatella: @tolbrino @dincho Shall this rather go in https://github.com/aeternity/docker-builder/blob/d15fa8e500d4ec4ee808aac6c3ea89a5b42c8f22/Dockerfile ?

lucafavatella: (Eventually...)

tolbrino: As per @dincho 's suggestion I used another pre-built image, which works fine.

.circleci/config.yml
```diff
@@ -1443,6 +1457,18 @@ workflows:
             tags:
               only: *stable_tag_regex
 
+      - build_uml_diagrams:
+          requires:
+            - build
+          filters:
+            branches:
```
lucafavatella: It feels this should be only master.

PR #2592, opened 2019-07-22
fix wrong name of OracleResponseTx

marc0olo:
when implementing oracle support for the java-sdk I experienced an issue with serialization of the generated `OracleRespondTx`

### DIFF
--- a/config/swagger.yaml
+++ b/config/swagger.yaml
@@ -1119,7 +1119,7 @@ paths:
           name: body
           required: true
           schema:
-            $ref: '#/definitions/OracleRespondTx'
+            $ref: '#/definitions/OracleResponseTx'
       responses:
         '200':
           description: Successful operation
@@ -2042,7 +2042,7 @@ definitions:
       - response_ttl
       - fee
       - sender_id
-  OracleRespondTx:
+  OracleResponseTx:
     type: object
     properties:
       query_id:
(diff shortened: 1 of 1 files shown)


lucafavatella:
@marc0olo Thanks for the patch! Using which version did you experience that? Earlier than 4.0.0?

Version 4.0.0 includes https://github.com/aeternity/aeternity/commit/b3f5e1f0 that was an intended alternative solution for the issue I expect you experienced.

marc0olo:
@lucafavatella yes I experienced this on an earlier version (3.3.0) so this PR can probably be closed. this means that from 4.0.0 on the response-type of the node will be `OracleRespondTx` and no longer `OracleResponseTx` and there is no change in the swagger.yaml required, right?

lucafavatella:
> ... from 4.0.0 on the response-type of the node will be `OracleRespondTx` and no longer `OracleResponseTx` and there is no change in the swagger.yaml required, right?

Yes. We try to mention [this kind of things in release notes](https://github.com/aeternity/aeternity/blame/v4.0.0/docs/release-notes/RELEASE-NOTES-4.0.0.md#L31).

> I personally prefer fixing it in the swagger.yaml. I am wondering how other SDKs were handling this. I guess they all need to adapt regardless where the fix will be done.
> 
> **edit:**
> 
> * at least all SDKs that use the swagger.yaml for code-generation need to adapt regardless

You are totally correct. Another SDK having this problem is how this was noticed - I remember. The fix that was merged changed the discriminator (that is not mentioned in the Swagger file hence that did not change).

> yes I experienced this on an earlier version (3.3.0) so this PR can probably be closed. ...

Closing. Thanks.

PR #2594, opened 2019-07-24
Clarifications re block validations

lucafavatella:
Prepared while designing https://www.pivotaltracker.com/story/show/166642114

### DIFF
--- a/apps/aecore/src/aec_sync.erl
+++ b/apps/aecore/src/aec_sync.erl
@@ -109,7 +109,18 @@ handle_worker(Task, Action) ->
 %% We sync with several nodes at the same time and use as strategy
 %% to pick a random hash from the hashes in the pool.
 
--record(sync_task, {id, chain, pool = [], agreed,
+-type chain_id() :: reference().
+-type chain_block() :: #{ hash := aec_blocks:block_header_hash()
+                        , height := aec_blocks:height()
+                        }.
+-type chain() :: #{ chain_id := chain_id()
+                  , peers := [aec_peers:peer_id()]
+                  , chain := [chain_block(), ...]
+                  }.
+-record(sync_task, {id :: chain_id(),
+                    chain :: chain(),
+                    pool = [],
+                    agreed :: undefined | chain_block(),
                     adding = [], pending = [],
                     workers = [] :: [{aec_peers:peer_id(), pid()}]}).
 -record(state, { sync_tasks = []                 :: [#sync_task{}]
--- a/apps/aecore/src/aec_sync.erl
+++ b/apps/aecore/src/aec_sync.erl
@@ -411,16 +411,18 @@ maybe_end_sync_task(State, ST) ->
             set_sync_task(ST, State)
     end.
 
-maybe_new_top_target(#chain{ chain = [#chain_block{ height = ChainTopTarget } | _] }, State) ->
+maybe_new_top_target(#chain{ chain = [B | _] }, State) ->
+    ChainTopTarget = B#chain_block.height,
     case State#state.top_target < ChainTopTarget of
         true  -> update_top_target(ChainTopTarget, State);
         false -> State
     end.
 
-maybe_update_top_target(State = #state{ top_target = TopTarget, sync_tasks = STs }) ->
+maybe_update_top_target(State = #state{}) ->
+    #state{ top_target = TopTarget, sync_tasks = STs } = State,
     NewTop = lists:foldl(fun(#sync_task{ chain = Chain }, PrevMax) ->
-                             case Chain of
-                                 #chain{ chain = [#chain_block{ height = H } | _] } when H > PrevMax -> H;
+                             case Chain#chain.chain of
+                                 [#chain_block{ height = H } | _] when H > PrevMax -> H;
                                  _ -> PrevMax
                              end
                          end, 0, STs),
@@ -498,7 +500,8 @@ match_tasks(Chain1, [ST = #sync_task{ chain = Chain2 } | STs], Acc) ->
         {snd, N}  -> match_tasks(Chain1, STs, [{N, Chain2} | Acc])
     end.
 
-match_chains([#chain_block{ height = N1 } | C1], [#chain_block{ height = N2, hash = H } | _]) when N1 > N2 ->
+match_chains([#chain_block{ height = N1 } | C1],
+             [#chain_block{ height = N2, hash = H } | _]) when N1 > N2 ->
     case find_hash_at_height(N2, C1) of
         {ok, H}   -> equal;
         {ok, _}   -> different;
@@ -647,7 +650,9 @@ init_chain(PeerId, Header) ->
 init_chain(ChainId, Peers, Header) ->
     {ok, Hash} = aec_headers:hash_header(Header),
     Height     = aec_headers:height(Header),
-    PrevHash   = [ #chain_block{ height = Height - 1, hash = aec_headers:prev_hash(Header)} || Height > 1 ],
+    PrevHash   = [ #chain_block{ height = Height - 1
+                               , hash = aec_headers:prev_hash(Header)
+                               } || Height > 1 ],
     #chain{ id = ChainId, peers = Peers,
             chain = [#chain_block{ hash = Hash, height = Height }] ++ PrevHash }.
 
@@ -826,7 +831,10 @@ fill_pool(PeerId, StartHash, TargetHash, ST) ->
             epoch_sync:info("Sync done (according to ~p)", [ppp(PeerId)]),
             aec_events:publish(chain_sync, {chain_sync_done, PeerId});
         {ok, Hashes} ->
-            HashPool = [ #pool_item{ height = Height, hash = Hash, got = false } || {Height, Hash} <- Hashes ],
+            HashPool = [ #pool_item{ height = Height
+                                   , hash = Hash
+                                   , got = false
+                                   } || {Height, Hash} <- Hashes ],
             do_work_on_sync_task(PeerId, ST, {hash_pool, HashPool});
         {error, _} = Error ->
             epoch_sync:info("Abort sync with ~p (~p) ", [ppp(PeerId), Error]),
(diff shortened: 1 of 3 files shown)


apps/aecore/src/aec_sync.erl
```diff
@@ -110,19 +110,22 @@ handle_worker(Task, Action) ->
...
 
 -type chain_id() :: reference().
--type chain_block() :: #{ hash := aec_blocks:block_header_hash()
-                        , height := aec_blocks:height()
-                        }.
--type chain() :: #{ chain_id := chain_id()
-                  , peers := [aec_peers:peer_id()]
-                  , chain := [chain_block(), ...]
-                  }.
+-record(chain_block, { hash :: aec_blocks:block_header_hash()
```
lucafavatella: Records makes the code much easier to follow statically for process state that does not travel across module boundaries.

----

From https://learnyousomeerlang.com/maps

> ... despite the similarity in syntax between maps and records, dicts and maps are much closer together semantically than maps are to records. Therefore, using maps to replace records would be similar to trying to replace 'structs' in a language like C with a hash map. It's not impossible, but that doesn't mean it's a good idea given they often have different purposes.
>
> Keys being known at compile time brings advantages with fast access to specific values (faster than what is possible dynamically), additional safety (crash early rather than corrupting state), and easier type checking. These make records absolutely appropriate for a process' internal state, despite the occasional burden of writing a more verbose code_change function.
>
> On the other hand, where Erlang users would use records to represent complex nested key/value data structures (oddly similar to objects in object-oriented languages) that would frequently cross module boundaries, maps will help a lot. Records were the wrong tool for that job.
>
> To put it briefly, places where records really felt out of place and cumbersome could be replaced by maps, but most record uses should not be replaced that way.

apps/aecore/src/aec_sync.erl
```diff
@@ -925,3 +928,15 @@ sync_progress(#state{sync_tasks = SyncTasks} = State) ->
...
+    #{hash => Hash, height => Height}.
+
+pretty_print_chain(C = #chain{}) ->
+    #{ chain_id => C#chain.id
+     , peers => C#chain.peers
+     , chain => lists:map(fun pretty_print_chain_block/1, C#chain.chain)
+     }.
+
+pretty_print_sync_task(ST = #sync_task{}) ->
+    setelement(#sync_task.chain, ST, pretty_print_chain(ST#sync_task.chain)).
```
lucafavatella: This needs to pretty print `agreed` and `workers` too.

apps/aecore/src/aec_sync.erl
```diff
@@ -913,3 +933,40 @@ sync_progress(#state{sync_tasks = SyncTasks} = State) ->
                 false -> SyncProgress
             end
     end.
+
+pretty_print_chain_block(#chain_block{hash = Hash, height = Height}) ->
+    #{ hash => Hash
```
tolbrino: You could generalize these pretty print function by taking using the `record_fields` to dynamically build that map.

uwiger: Ahem... here is of course the ultimate record pretty-printing generalization ;-)
https://github.com/uwiger/jobs/blob/master/src/jobs_info.erl

tolbrino: That is pretty much what I had in mind. Someone please submit a PR to OTP :-)

apps/aecore/src/aec_sync.erl
```diff
@@ -109,9 +109,34 @@ handle_worker(Task, Action) ->
 %% We sync with several nodes at the same time and use as strategy
 %% to pick a random hash from the hashes in the pool.
 
--record(sync_task, {id, chain, pool = [], agreed,
-                    adding = [], pending = [],
-                    workers = [] :: [{PeerId::binary(), pid()}]}).
+-type chain_id() :: reference().
+-record(chain_block, { hash :: aec_blocks:block_header_hash()
```
hanssv: Wouldn't `chain_key_block` be more appropriate?!

apps/aecore/src/aec_sync.erl
```diff
@@ -357,28 +377,28 @@ get_next_work_item(ST = #sync_task{ pool = [{_, _, {_, _}} | _] = Pool, adding =
...
+    epoch_sync:info("Nothing to do: ~p", [pretty_print_sync_task(ST)]),
     {take_a_break, ST}.
 
 maybe_end_sync_task(State, ST) ->
     case ST#sync_task.chain of
-        #{ peers := [], chain := [Target | _] } ->
+        #chain{ peers = [], chain = [Target | _] } ->
             epoch_sync:info("Removing/ending sync task ~p target was ~p",
-                            [ST#sync_task.id, Target]),
+                            [ST#sync_task.id, pretty_print_chain_block(Target)]),
```
hanssv:
```suggestion
                            [ST#sync_task.id, pp_chain_block(Target)]),
```

apps/aecore/src/aec_sync.erl
```diff
@@ -357,28 +377,28 @@ get_next_work_item(ST = #sync_task{ pool = [{_, _, {_, _}} | _] = Pool, adding =
...
+get_next_work_item(ST = #sync_task{ pool = [#pool_item{ got = false } | _] = Pool }) ->
+    PickFrom = [ P || P = #pool_item{ got = false } <- Pool ],
     Random = rand:uniform(length(PickFrom)),
-    {PickH, PickHash, false} = lists:nth(Random, PickFrom),
+    #pool_item{ height = PickH, hash = PickHash, got = false} = lists:nth(Random, PickFrom),
     epoch_sync:debug("Get block at height ~p", [PickH]),
     {{get_generation, PickH, PickHash}, ST};
 get_next_work_item(ST) ->
-    epoch_sync:info("Nothing to do: ~p", [ST]),
+    epoch_sync:info("Nothing to do: ~p", [pretty_print_sync_task(ST)]),
```
hanssv:
```suggestion
    epoch_sync:info("Nothing to do: ~p", [pp_sync_task(ST)]),
```

apps/aecore/src/aec_sync.erl
hanssv: Don't make it more complex than necessary, but please don't use function names spanning half a page...

apps/aecore/src/aec_sync.erl
```diff
@@ -913,3 +933,40 @@ sync_progress(#state{sync_tasks = SyncTasks} = State) ->
                 false -> SyncProgress
             end
     end.
+
+pretty_print_chain_block(#chain_block{hash = Hash, height = Height}) ->
```
hanssv:
```suggestion
pp_chain_block(#chain_block{hash = Hash, height = Height}) ->
```

hanssv: Nice 👍

lucafavatella:
I need to process recent comments.

apps/aecore/src/aec_sync.erl
lucafavatella: One such item is build taking `aec_headers:prev_hash(Header)` - not the prev key hash - so it seems no to me. 🤔

lucafavatella:
> Aren't there a risk that the pretty-printed thingies introduce newlines? Not sure Dincho will be happy :-)

@hanssv The refactoring did not change this behaviour though I now pushed a commit that is meant to reduce the likelihood of newline occurrences.

dincho:
@lucafavatella @hanssv thanks for taking care for new lines. It's not strict requirement as it does not chokes the parser too much (if it's rare), however we could miss that lines from the logs collectors otherwise

lucafavatella:
[In CI job re Docker system smoke tests](https://circleci.com/gh/aeternity/aeternity/52819#tests/containers/0) I read:
```
$ 
#!/bin/bash -eo pipefail
# Install OTP package deps
sudo apt-get update && sudo apt-get install libwxbase3.0-dev libwxgtk3.0-dev
# Install OTP binary package
PACKAGE_NAME=esl-erlang_${OTP_VERSION}-1~ubuntu~trusty_amd64.deb
OTP_DOWNLOAD_URL=https://packages.erlang-solutions.com/erlang/debian/pool/${PACKAGE_NAME}
curl -fsSL -o ${PACKAGE_NAME} "$OTP_DOWNLOAD_URL"
sudo dpkg -i ${PACKAGE_NAME}

E: Could not get lock /var/lib/apt/lists/lock - open (11: Resource temporarily unavailable)
E: Unable to lock directory /var/lib/apt/lists/
Selecting previously unselected package esl-erlang.
(Reading database ... 71372 files and directories currently installed.)
Preparing to unpack esl-erlang_20.1-1~ubuntu~trusty_amd64.deb ...
Unpacking esl-erlang (1:20.1) ...
dpkg: dependency problems prevent configuration of esl-erlang:
 esl-erlang depends on libwxbase2.8-0 | libwxbase3.0-0 | libwxbase3.0-0v5; however:
  Package libwxbase2.8-0 is not installed.
  Package libwxbase3.0-0 is not installed.
  Package libwxbase3.0-0v5 is not installed.
 esl-erlang depends on libwxgtk2.8-0 | libwxgtk3.0-0 | libwxgtk3.0-0v5; however:
  Package libwxgtk2.8-0 is not installed.
  Package libwxgtk3.0-0 is not installed.
  Package libwxgtk3.0-0v5 is not installed.

dpkg: error processing package esl-erlang (--install):
 dependency problems - leaving unconfigured
Errors were encountered while processing:
 esl-erlang
Exited with code 1
```

lucafavatella:
Longer lines are broken up.

apps/aecore/src/aec_sync.erl
```diff
@@ -109,9 +109,34 @@ handle_worker(Task, Action) ->
...
--record(sync_task, {id, chain, pool = [], agreed,
-                    adding = [], pending = [],
-                    workers = [] :: [{PeerId::binary(), pid()}]}).
+-type chain_id() :: reference().
+-record(chain_block, { hash :: aec_blocks:block_header_hash()
+                     , height :: aec_blocks:height()
+                     }).
+-record(chain, { id :: chain_id()
+               , peers :: [aec_peers:peer_id()]
+               , chain :: [#chain_block{}, ...]
```
tolbrino: Feels confusing to have `chain` in `chain`. Maybe `blocks` in `chain`?

apps/aecore/src/aec_sync.erl
hanssv: 😱 That sounds like a bug! I'm not within easy reach of a computer so I can't check properly - but I thought sync would work on key-blocks (headers) 🤔

lucafavatella: [Tracked](https://www.pivotaltracker.com/story/show/167589944).

PR #2595, opened 2019-07-24
Fix duplicate line

grzegorz225:


apps/aeprimop/src/aeprimop.erl
```diff
@@ -1407,7 +1407,6 @@ assert_relevant_signature(AccountPK, STx, State) ->
         {_, _} ->
             case aetx:signers(Tx, State#state.trees) of
                 {ok, Signers} ->
-                    aetx:signers(Tx, State#state.trees),
```
lucafavatella: Harmless but more clear now. Thanks!

PR #2599, opened 2019-07-25
Names bidding

sennui:
Implemented according to protocol repository PR: 
https://github.com/aeternity/protocol/pull/379

DONE:
- whole backend adjusted including fate, aevm etc for new argument in claim transaction with bid value
- aens SUITE tests all versions:
   old names & old behaviour (skips auctions),
   old names & new behaviour
   new names & new behaviour
   utf8 on all heights 
- some tests edited to escape auction behaviour at all heights. Done that by extending name to non-auction length (over 31 chars now)

NOT DONE:

- Swagger has only new interface
- HTTP integration test tests only new interface
- aecontract SUITE not finished as top level /test templates are not adjusted for extra argument in claim. Need to revert it and write new test that is skipped in pre-lima
- couple tests around failing due to not adjusted contract templates in /test 

Depends on unmerged aebytecode and sophia PRs (they are pointed in this PR rebar.config). Pre-merge commit will need to fix rebar.config and rebar.lock with rebased aebytecode and sophia PRs.

### DIFF
--- a/apps/aecontract/test/aecontract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_SUITE.erl
@@ -3472,7 +3472,9 @@ sophia_signatures_aens(Cfg) ->
     Acc             = ?call(new_account, 20000000 * aec_test_utils:min_gas_price()),
     NameAcc         = ?call(new_account, 20000000 * aec_test_utils:min_gas_price()),
     Ct              = ?call(create_contract, NameAcc, aens, {}, #{ amount => 100000 }),
-    Name1           = <<"bla.test">>,
+    LongPrefix      = <<"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx">>,
+    Name1           = <<LongPrefix/binary, "bla.test">>,
+    NameFee         = aec_governance:name_claim_fee(aens_commitments:name_length(Name1)),
     Salt1           = rand:uniform(10000),
     {ok, NameAscii} = aens_utils:to_ascii(Name1),
     CHash           = ?hsh(aens_hash:commitment_hash(NameAscii, Salt1)),
@@ -3483,7 +3485,7 @@ sophia_signatures_aens(Cfg) ->
                   VMVersion when ?IS_FATE_SOPHIA(VMVersion) -> Name1
               end,
     NameAccSig      = sign(<<NameAcc/binary, Ct/binary>>, NameAcc),
-    {ok, NameHash}  = aens:get_name_hash(<<"bla.test">>),
+    {ok, NameHash}  = aens:get_name_hash(<<LongPrefix/binary, "bla.test">>),
     NameSig         = sign(<<NameAcc/binary, NameHash/binary, Ct/binary>>, NameAcc),
     AccSig          = sign(<<Acc/binary, NameHash/binary, Ct/binary>>, Acc),
     APubkey  = 1,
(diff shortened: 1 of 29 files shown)


apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -3472,7 +3472,9 @@ sophia_signatures_aens(Cfg) ->
     Acc             = ?call(new_account, 20000000 * aec_test_utils:min_gas_price()),
     NameAcc         = ?call(new_account, 20000000 * aec_test_utils:min_gas_price()),
     Ct              = ?call(create_contract, NameAcc, aens, {}, #{ amount => 100000 }),
-    Name1           = <<"bla.test">>,
+    LongPrefix      = <<"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx">>,
```
ThomasArts: Why are we replacing the existing test with this one?

apps/aecore/src/aec_governance.erl
```diff
@@ -12,10 +12,13 @@
...
          minimum_gas_price/1,
          name_preclaim_expiration/0,
-         name_claim_locked_fee/0,
+         name_claim_bid_increment/0,
+         name_claim_fee/1,
+         name_claim_fee_base/0,
          name_claim_max_expiration/0,
          name_protection_period/0,
-         name_claim_preclaim_delta/0,
+         name_claim_preclaim_timeout/0,
```
ThomasArts: could we not simply keep the name name_claim_preclaim_delta?

ThomasArts: it is a delta, not really a time

sennui: @ThomasArts I change that to correlate with name_claim_bid_timout. In both cases it is not a time range that matters but the action that happens when they expire. For such activities, I found word `timeout` more suitable. Similar to timers, which, those functions represent in some abstract form, in fact, configure.

apps/aecontract/test/aecontract_SUITE.erl
sennui: @ThomasArts In some cases of old test we just want to use registered name. The Long Prefix falls back to non-auction behaviour (it's over 32 characters). Surly, like that it is for state channel test. Here, I think, we might want to have new test with auction and then with short name.

sennui:
> Would it make sense to have a first PR for the transactions... actually working for all heights and then a second PR for the contract stuff?

I don't thinks it is possible as contracts use now the same backend (primop) that I am changing for this functionality. 

Currently remaining bit is support of old contracts running on AEVM (I updated it to new stuff and it is not easy to make it work on both, no version around)

apps/aecore/src/aec_aens_governance.erl
```diff
@@ -0,0 +1,66 @@
+%%%-------------------------------------------------------------------
+%%% @doc Placeholder for functions instrumenting Naming auctions
+%%%
+%%% @end
+%%%-------------------------------------------------------------------
+-module(aec_aens_governance).
```
ThomasArts: I don't like the idea of spreading out governance in multiple modules.... hard to get a feeling what is under governance and what is not.

sennui: I followed up here on what we did with coinbase inflation function. I guess it was for code clarity. Agree with you that holding everything in single governance module may be bigger hint for newcomer. I am not attached to the new module at all, if you think it's clearer to bring it to home governance, I am all up for it. Let me know.

apps/aens/src/aens_claim_tx.erl
```diff
@@ -196,7 +234,7 @@ account_pubkey(#ns_claim_tx{account_id = AccountId}) ->
 
 -spec version(tx()) -> non_neg_integer().
 version(_) ->
```
ThomasArts: This must be depending on  height

ThomasArts: Please make all tests pass and compatible with synchronizing to an existing chain.

sennui: Done

PR #2600, opened 2019-07-26
Pt 166698281 allow sc ws to disconnect

uwiger:
See [PT #166698281](https://www.pivotaltracker.com/story/show/166698281).

WIP: The WS API has not yet been updated to support the reconnect operation. Currently, disconnect/reconnect is simulated in the `aesc_fsm_SUITE`.

The `aeserialization` app has been updated to support the `channel_client_reconnect_tx` type. See PR aeternity/aeserialization#7


apps/aechannel/src/aesc_client_reconnect_tx.erl
```diff
@@ -0,0 +1,185 @@
...
+-export_type([tx/0]).
+
+
+-spec new(map()) -> {ok, aetx:tx()}.
+new(#{ channel_id := ChannelId
+     , role       := Role
+     , pub_key    := Pubkey }) ->
+    channel = aeser_id:specialize_type(ChannelId),
+    account = aeser_id:specialize_type(Pubkey),
+    true = (Role == initiator) orelse (Role == responder),
```
tolbrino: So you want to fail hard here if the role is unsupported? Which other values could be in `role`? And shouldn't that check be common for most of `aechannel`?

apps/aechannel/src/aesc_client_reconnect_tx.erl
```diff
@@ -0,0 +1,185 @@
...
+    aeser_id:specialize(Pubkey, account).
+
+-spec check(tx(), aec_trees:trees(), aetx_env:env()) -> {ok, aec_trees:trees()} | {error, term()}.
+check(#channel_client_reconnect_tx{
+        channel_id = _ChannelId
+      , role       = _Role
+      , pub_key    = _Pubkey }, Trees, _Env) ->
+    {ok, Trees}.
+
+-spec process(tx(), aec_trees:trees(), aetx_env:env()) -> {ok, aec_trees:trees()}.
```
tolbrino: Wrong return spec?

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2197,6 +2219,22 @@ check_inband_msg(#{ channel_id := ChanId
 check_inband_msg(_, _) ->
     {error, chain_id_mismatch}.
 
+check_client_reconnect_tx(Tx, #data{} = D) ->
+    lager:debug("Check reconnect tx", []),
+    case D#data.strict_checks of
```
tolbrino: This check can be moved into a function clause. Improves readability.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2535,10 +2576,22 @@ check_tx_and_verify_signatures(SignedTx, Updates, Mod, Data, Pubkeys, ErrTypeMsg
...
     ExpectedRound = next_round(Data),
     #data{state = State, opts = Opts, on_chain_id = ChannelPubkey} = Data,
+    Aetx = aetx_sign:tx(SignedTx),
     case aetx:specialize_callback(aetx_sign:innermost_tx(SignedTx)) of
         {aesc_settle_tx, _Tx} ->
             %% TODO: conduct more relevant checks
             verify_signatures_onchain_check(Pubkeys, SignedTx);
+        {aesc_client_reconnect_tx = Mod, Tx} ->
+            MyPubkey = my_account(Data),
+            MyRole = Data#data.role,
```
tolbrino: Should be moved  upwards into line 2576.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -3325,6 +3424,18 @@ handle_call_(St, _Req, _From, D) when ?TRANSITION_STATE(St) ->
 handle_call_(_St, _Req, From, D) ->
     keep_state(D, [{reply, From, {error, unknown_request}}]).
 
+handle_info({'DOWN', MRef, process, Client, _} = M,
+            #data{ client = Client, client_mref = MRef } = D) ->
+    lager:debug("Client disconnected ~p", [Client]),
+    case D#data.client_may_disconnect of
```
tolbrino: Moving this case clause into a function clause would improve readability.

apps/aechannel/src/aesc_client_reconnect_tx.erl
```diff
@@ -0,0 +1,185 @@
+-module(aesc_client_reconnect_tx).
+
+-behavior(aetx).
+-behavior(aesc_signable_transaction).
```
grzegorz225: There should be a comment saying that this TX needs to be signed by the person reconnecting in order to prove the identity of the peer.

apps/aechannel/src/aesc_client_reconnect_tx.erl
```diff
@@ -0,0 +1,185 @@
...
+-export([
+          channel_id/1
+        , channel_pubkey/1
+        , role/1
+        , round/1
+        , state_hash/1
+        ]).
+
+-define(CHANNEL_CLIENT_RECONNECT_TX_TYPE, channel_client_reconnect_tx).
+-define(CHANNEL_CLIENT_RECONNECT_TX_FEE, 0).  % off-chain
```
grzegorz225: Please use a double %% instead of a single %.

apps/aechannel/src/aesc_client_reconnect_tx.erl
```diff
@@ -0,0 +1,185 @@
...
+-spec check(tx(), aec_trees:trees(), aetx_env:env()) -> {ok, aec_trees:trees()} | {error, term()}.
+check(#channel_client_reconnect_tx{
+        channel_id = _ChannelId
+      , role       = _Role
+      , pub_key    = _Pubkey }, Trees, _Env) ->
+    {ok, Trees}.
+
+-spec process(tx(), aec_trees:trees(), aetx_env:env()) -> {ok, aec_trees:trees()}.
+process(#channel_client_reconnect_tx{}, _Trees, #{}) ->
+    error(off_chain_tx).
```
grzegorz225: Does this mean that the TX can hit the mem pool but the miners won't include this TX inside a block?

apps/aechannel/src/aesc_codec.hrl
```diff
@@ -52,6 +52,7 @@
 
 %% Non-encoded cast types
 -define(DISCONNECT, disconnect).
+-define(CONNECT_CLIENT, connect_client).
```
grzegorz225: Should be called RECONNECT_CLIENT.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -17,6 +17,8 @@
         , attach_responder/2      %% (fsm(), map())
         , client_died/1           %% (fsm())
         , close_solo/1
+        , connect_client/2        %% (fsm(), signed_tx())
```
grzegorz225: connect -> reconnect

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -17,6 +17,8 @@
...
+        , connect_client/3        %% (fsm(), Client :: pid(), signed_tx())
```
grzegorz225: connect -> reconnect

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -135,6 +137,14 @@ close_solo(Fsm) ->
     lager:debug("close_solo(~p)", [Fsm]),
     gen_statem:call(Fsm, close_solo).
 
+-spec connect_client(pid(), aetx_sign:signed_tx()) -> ok | {error, any()}.
```
grzegorz225: Same as above

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -3071,14 +3128,56 @@ close_(Reason, D) ->
     end,
     {stop, Reason, D}.
 
-handle_call(St, Req, From, #data{} = D) ->
+handle_call(_, {?CONNECT_CLIENT, Pid, Tx} = Msg, From,
+            #data{ client_connected = false
```
grzegorz225: If someone tries to reconnect while the client is connected then this will crash the FSM, someone can use this to DOS the FSM.
Let's say Alice and Mallory uses the same node for state channels. Mallory want's to DOS Alice - all she needs to do is just constantly trying to reconnect to Alice's FSM - this will crash the FSM every time Alice reestablishes the channel.

apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -170,7 +172,10 @@ groups() ->
       ]},
      {initiator_is_ga, [sequence], ga_sequence()},
      {responder_is_ga, [sequence], ga_sequence()},
-     {both_are_ga, [sequence], ga_sequence()}
+     {both_are_ga, [sequence], ga_sequence()},
+     {client_reconnect, [sequence],
+      [ client_reconnect_initiator
```
grzegorz225: This can be one test case.

apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -1545,6 +1550,83 @@ settle_(TTL, MinDepth, #{fsm := FsmI, channel_id := ChannelId} = I, R, Debug,
     ct:log("died_normal detected from both", []),
     ok.
 
+client_reconnect_initiator(Cfg) ->
+    Debug = get_debug(Cfg),
```
grzegorz225: Wrap the test case inside a list comprehension with atoms [initiator, responder] as this is the only difference between the two test cases.

apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -170,7 +172,10 @@ groups() ->
       ]},
      {initiator_is_ga, [sequence], ga_sequence()},
      {responder_is_ga, [sequence], ga_sequence()},
-     {both_are_ga, [sequence], ga_sequence()}
+     {both_are_ga, [sequence], ga_sequence()},
+     {client_reconnect, [sequence],
+      [ client_reconnect_initiator
+      , client_reconnect_responder ]}
```
grzegorz225: There is no test case for an unauthorized user trying to reconnect to the FSM. We don't check what happens when someone tries to reconnect while the FSM has a connected client.

apps/aechannel/src/aesc_client_reconnect_tx.erl
uwiger: The old convention for trailing comments is otherwise a single `%` ...
http://www.erlang.se/doc/programming_rules.shtml#HDR36

apps/aechannel/src/aesc_client_reconnect_tx.erl
uwiger: I figured that it's in line with the assertions regarding channel_id and pubkey type on the lines above. `Role` is usually not part of the channel txs.

apps/aechannel/src/aesc_client_reconnect_tx.erl
uwiger: Seems like it

apps/aechannel/src/aesc_client_reconnect_tx.erl
uwiger: This is the same as for `aesc_offchain_tx`. Also, the fee and ttl are zero, so chances that miners will pick it up even if it would be pushed to the mempool should be 0.

apps/aechannel/src/aesc_fsm.erl
uwiger: It pattern-matches on `client = undefined`, so this clause should not execute if a client is connected. But there should be some test code verifying this.

apps/aechannel/src/aesc_fsm_sup.erl
```diff
@@ -0,0 +1,24 @@
+-module(aesc_fsm_sup).
+-behaviour(supervisor).
+
+-export([
+          start_child/1
```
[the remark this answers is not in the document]
uwiger: Yeah...

apps/aechannel/test/aesc_fsm_SUITE.erl
uwiger: I am partial to reasonably specialized test cases, since debugging failing complex test cases can be quite time-consuming.

rebar.config
```diff
@@ -69,7 +69,7 @@
                       {ref,"17c2a93"}}},
 
         {aeserialization, {git, "https://github.com/aeternity/aeserialization.git",
```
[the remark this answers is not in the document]
uwiger: It was noted in the PR description - my assumption was that the related PR would be merged first.

apps/aechannel/src/aesc_client_reconnect_tx.erl
```diff
@@ -0,0 +1,185 @@
+-module(aesc_client_reconnect_tx).
```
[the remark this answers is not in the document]
uwiger: Done

apps/aechannel/src/aesc_client_reconnect_tx.erl
uwiger: Fixed

apps/aechannel/src/aesc_fsm.erl
uwiger: Done

apps/aechannel/src/aesc_fsm.erl
uwiger: Done

apps/aechannel/src/aesc_client_reconnect_tx.erl
uwiger: Now explained in the top-level comment about the module.

apps/aechannel/src/aesc_codec.hrl
uwiger: Fixed

apps/aechannel/src/aesc_fsm.erl
uwiger: Fixed

apps/aechannel/src/aesc_fsm.erl
uwiger: Code has been added to the `aesc_fsm_SUITE` to check that trying to reconnect while a client is already connected yields an error. Currently, the error will be `unknown_request`.

apps/aechannel/test/aesc_fsm_SUITE.erl
uwiger: I refactored, but not exactly like that.

apps/aechannel/test/aesc_fsm_SUITE.erl
uwiger: This is now checked.

tolbrino:
This PR has been replaced by #2602 . Can be closed.

uwiger:
Superseded by #2602 - Closing.

PR #2601, opened 2019-07-30
Sync clarification

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/167567788

### DIFF
--- a/apps/aecore/src/aec_sync.erl
+++ b/apps/aecore/src/aec_sync.erl
@@ -830,12 +830,33 @@ fill_pool(PeerId, StartHash, TargetHash, ST) ->
             update_sync_task({done, PeerId}, ST),
             epoch_sync:info("Sync done (according to ~p)", [ppp(PeerId)]),
             aec_events:publish(chain_sync, {chain_sync_done, PeerId});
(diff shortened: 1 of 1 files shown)


apps/aecore/src/aec_sync.erl
```diff
@@ -830,12 +830,33 @@ fill_pool(PeerId, StartHash, TargetHash, ST) ->
...
-            HashPool = [ #pool_item{ height = Height
-                                   , hash = Hash
-                                   , got = false
-                                   } || {Height, Hash} <- Hashes ],
-            do_work_on_sync_task(PeerId, ST, {hash_pool, HashPool});
+        {ok, Hashes = [{FirstHeight, _} | _]} when
+              %% Guaranteed by deserialization.
+              is_integer(FirstHeight), FirstHeight >= 0 ->
+            case
+                lists:foldl(fun ({H, _}, {ok, AccH}) when is_integer(H) ->
```
uwiger: Perhaps this could be broken out into a helper function, with an instructive name, like `heights_are_consequtive`?

lucafavatella: Addressed in commit I pushed.

PR #2602, opened 2019-07-31
Pt 166698281 allow sc ws to disconnect 2

uwiger:
Same as #2600 , but rebased on top of #2596 (just merged).
Deps updated to account for the related `aeserialization` PR being merged, changing the commit hash.

PR #2603, opened 2019-07-31
add round in reconnect_tx to guard against replay attacks

uwiger:
Proposed change to #2602 - adding `round` to the reconnect_tx in order to protect against replay attacks.

### DIFF
--- a/apps/aechannel/src/aesc_client_reconnect_tx.erl
+++ b/apps/aechannel/src/aesc_client_reconnect_tx.erl
@@ -51,6 +51,7 @@
 -record(channel_client_reconnect_tx,
         {
           channel_id   :: aeser_id:id()
+        , round        :: non_neg_integer()
         , role         :: initiator | responder
         , pub_key      :: aeser_id:id()
         }).
@@ -62,13 +63,16 @@
 
 -spec new(map()) -> {ok, aetx:tx()}.
 new(#{ channel_id := ChannelId
+     , round      := Round
      , role       := Role
      , pub_key    := Pubkey }) ->
     channel = aeser_id:specialize_type(ChannelId),
     account = aeser_id:specialize_type(Pubkey),
+    true = is_integer(Round) andalso Round > 0,
     true = (Role == initiator) orelse (Role == responder),
     Tx = #channel_client_reconnect_tx{
             channel_id = ChannelId
+          , round      = Round
           , role       = Role
           , pub_key    = Pubkey },
     {ok, aetx:new(?MODULE, Tx)}.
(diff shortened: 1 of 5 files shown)


apps/aechannel/src/aesc_client_reconnect_tx.erl
```diff
@@ -62,13 +63,16 @@
 
 -spec new(map()) -> {ok, aetx:tx()}.
 new(#{ channel_id := ChannelId
+     , round      := Round
      , role       := Role
      , pub_key    := Pubkey }) ->
     channel = aeser_id:specialize_type(ChannelId),
     account = aeser_id:specialize_type(Pubkey),
+    true = is_integer(Round) andalso Round > 0,
```
lucafavatella: Type `non_neg_integer()` but checked for positive? 🤔

apps/aechannel/src/aesc_client_reconnect_tx.erl
```diff
@@ -124,40 +128,48 @@ signers(#channel_client_reconnect_tx{role = Role} = Tx, Trees) ->
...
      ]}.
 
 -spec deserialize(vsn(), list()) -> tx().
 deserialize(?INITIAL_VSN, [ {channel_id, ChannelId}
+                          , {round     , Round}
                           , {role      , Role0}
                           , {pub_key   , Pubkey} ]) ->
     channel = aeser_id:specialize_type(ChannelId),
     account = aeser_id:specialize_type(Pubkey),
+    true = is_integer(Round) andalso Round > 0,
```
lucafavatella: Type `non_neg_integer()` but checked for positive? 🤔

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2556,8 +2558,9 @@ check_tx_and_verify_signatures(SignedTx, Updates, Mod, Data, Pubkeys, ErrTypeMsg
             %% TODO: conduct more relevant checks
             verify_signatures_onchain_check(Pubkeys, SignedTx);
         {aesc_client_reconnect_tx = Mod, Tx} ->
-            try {Mod:channel_pubkey(Tx), Mod:role(Tx), Mod:origin(Tx)} of
-                {ChannelPubkey, MyRole, MyPubkey} ->
+            try {Mod:channel_pubkey(Tx), Mod:role(Tx), Mod:origin(Tx), Mod:round(Tx)} of
+                {ChannelPubkey, MyRole, MyPubkey, Round}
+                  when Round > Data#data.client_reconnect_nonce ->
```
lucafavatella: Is the uncaught `try_clause` intended?
```
If the evaluation of Exprs succeeds without an exception, the patterns Pattern are sequentially matched against the result in the same way as for a case expression, except that if the matching fails, a try_clause run-time error occurs.
```
```
1> try 2 of Round when Round > 5 -> ok catch error:_ -> {error, invalid} end.
** exception error: no try clause matching 2
```

lucafavatella: I see that the only point where this function is called with reconnect is:
```
    try check_client_reconnect_tx(Tx, D) of
    ...
    catch
        error:E ->
            lager:debug("CAUGHT ~p / ~p", [E, erlang:get_stacktrace()]),
            keep_state(D, [{reply, From, E}])
    end
```
so seems ok even if counterintuitive - as same function returns either error tuple of exception.

lucafavatella: It happens to catch class error hence works...
```
1> try begin try 2 of Round when Round > 5 -> ok catch error:_ -> {error, invalid} end end of _ -> ok catch error:_ -> caught end.
caught
```

uwiger: Addressed in new commit

apps/aechannel/src/aesc_client_reconnect_tx.erl
uwiger: Well, the type info is just informative. :)

PR #2606, opened 2019-08-01
API URLs of Localnet compatibility with swagger schema

dincho:
Swagger schema expects /v2/debug prefix, while current configuration is adding additional /internal prefix. While SDKs have support to configure 2 different URLs for internal and external endpoints, the additional compatibility configuration in this PR should make the user lives easier

### DIFF
--- a/docker/nginx-default.conf
+++ b/docker/nginx-default.conf
@@ -1,16 +1,30 @@
 server {
     listen 3001;
 
+    # External API
     location / {
         include cors.conf;
         proxy_pass http://node1:3013/;
     }
 
+    # Internal API
     location /internal/ {
+        # Backward compatibility
         include cors.conf;
         proxy_pass http://node1:3113/;
     }
 
+    location /v2/debug {
+        include cors.conf;
+        proxy_pass http://node1:3113/v2/debug;
+    }
+
+    location /v2/key-blocks {
+        include cors.conf;
+        proxy_pass http://node1:3113/v2/key-blocks;
+    }
+
+    # State Channels Web Sockets
     location /channel {
         include cors.conf;
         include ws.conf;
@@ -21,16 +35,30 @@ server {
 server {
     listen 3002;
 
+    # External API
     location / {
         include cors.conf;
         proxy_pass http://node2:3013/;
     }
 
+    # Internal API
     location /internal/ {
+        # Backward compatibility
         include cors.conf;
         proxy_pass http://node2:3113/;
     }
 
+    location /v2/debug {
+        include cors.conf;
+        proxy_pass http://node2:3113/v2/debug;
+    }
+
+    location /v2/key-blocks {
+        include cors.conf;
+        proxy_pass http://node2:3113/v2/key-blocks;
+    }
+
+    # State Channels Web Sockets
     location /channel {
         include cors.conf;
         include ws.conf;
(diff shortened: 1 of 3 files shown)


dincho:
This is marked as draft, just because of the ongoing release.

docker/nginx-default.conf
```diff
@@ -1,16 +1,30 @@
...
     }
 
+    # Internal API
     location /internal/ {
+        # Backward compatibility
         include cors.conf;
         proxy_pass http://node1:3113/;
     }
 
+    location /v2/debug {
```
lucafavatella: So this is prefix match.

docker/nginx-default.conf
```diff
@@ -1,16 +1,30 @@
...
+        include cors.conf;
+        proxy_pass http://node1:3113/v2/debug;
```
lucafavatella: Why URI specified? Not clear whether it has intention.

docker/nginx-default.conf
```diff
@@ -1,16 +1,30 @@
...
+    }
+
+    location /v2/key-blocks {
```
lucafavatella:
```suggestion
    location = /v2/key-blocks {
```

docker/nginx-default.conf
```diff
@@ -21,16 +35,30 @@ server {
...
         include cors.conf;
         proxy_pass http://node2:3113/;
     }
 
+    location /v2/debug {
+        include cors.conf;
+        proxy_pass http://node2:3113/v2/debug;
+    }
+
+    location /v2/key-blocks {
```
lucafavatella:
```suggestion
    location = /v2/key-blocks {
```

docker/nginx-default.conf
```diff
@@ -41,16 +69,30 @@ server {
...
         include cors.conf;
         proxy_pass http://node3:3113/;
     }
 
+    location /v2/debug {
+        include cors.conf;
+        proxy_pass http://node3:3113/v2/debug;
+    }
+
+    location /v2/key-blocks {
```
lucafavatella:
```suggestion
    location = /v2/key-blocks {
```

docker/nginx-default.conf
```diff
@@ -1,16 +1,30 @@
 server {
```
lucafavatella:
```suggestion
# Assumptions:
# - The set of paths tagged as `internal` and the set of paths tagged as `external` are disjoint.
# - The set of paths tagged as `internal` is equal to the union of: the set of paths with prefix `/v2/debug`; and the exact path `/v2/key-blocks` (beware that if you do not consider "exact" here, additional constraints need to be considered).

server {
```

docker/nginx-default.conf
dincho: Because nginx pass the path without the prefix only.

docker/nginx-default.conf
dincho: Did you tested if the `proxy_pass` below still works with the exact match ?

lucafavatella: No.

docker/nginx-default.conf
lucafavatella: Interesting.

From [the manual](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass):
> If proxy_pass is specified without a URI, the request URI is passed to the server in the same form as sent by a client when the original request is processed, or the full normalized request URI is passed when processing the changed URI:
```
location /some/path/ {
    proxy_pass http://127.0.0.1;
}
```

dincho: Perhaps you're reading latest docs. The version used in this setup is 2 years old

dincho: I tested it locally and it works. Applied in 878de61b

I think the previous proxy_pass had `/` in the argument, thus interpreted as path

dincho:
> I wonder if it were easy to hook to CI a static check on the schema that checks constraints above. Though optional as it may be not worth the effort.

I plan to move out this configuration in its own repo, thus I prefer to delay this suggestion until then

dincho:
Windows build issues are known, and this PR has nothing to do with it. Merging.

PR #2607, opened 2019-08-01
Update misleading VM version comment

lucafavatella:
... meant to future-proof changes to VM enabled for contract creation.

See https://github.com/aeternity/aeternity/pull/2551/files#r299020421

PR #2610, opened 2019-08-01
Make consensus checks more future proof

lucafavatella:
Supersedes #2607


lucafavatella:
> Can be done much cleaner.

@grzegorz225 I [would prefer](https://github.com/aeternity/aeternity/pull/2616#pullrequestreview-270746578) to keep this PR to the minimum set of changes for the intended functional behaviour.

PR #2613, opened 2019-08-02
Draft of design of miner signalled consensus upgrade

lucafavatella:
In scope of https://www.pivotaltracker.com/story/show/166642114

This PR is not meant to be merged - it is mainly for easing review.

### DIFF
--- a/docs/design/PT-166642114-miner-activated-consensus.md
+++ b/docs/design/PT-166642114-miner-activated-consensus.md
@@ -29,6 +29,7 @@ By exploiting such current design, the expected consensus protocol version of a 
   These blocks are not validated.
   This includes the cases of:
   - Locally mined or signed block.
+  - Mined block submitted by Stratum mining pool server (if node configured as Stratum server).
 - A block candidate (eventually leading to a trusted block) is prepared.
   This includes the cases of key block and micro block.
 
@@ -51,7 +52,7 @@ Specifically, the contexts in which the consensus version must be checked before
 - The peer connection / gossip process.
 - The HTTP user API worker.
 
-For blocks mined or signed locally, the conductor must instruct the generation of the block candidate with the correct consensus protocol version.
+For trusted blocks, the conductor must instruct the generation of the block candidate with the correct consensus protocol version.
 
 ###### Asynchronous computation of expected consensus protocol version
 
--- a/docs/design/PT-166642114-miner-activated-consensus.md
+++ b/docs/design/PT-166642114-miner-activated-consensus.md
@@ -71,8 +71,17 @@ immediately after the synchronous write call finishes from the point of view of 
 
 Callers requiring the signalling outcome for a certain block
 shall enquiry the chain storage and, if no outcome yet,
-ensure that there is a worker alive for that
-then yield (e.g. for sync worker process, for locally mined key blocks) or terminate (e.g. for block from user API or gossip).
+ensure that there is a worker alive for that then:
+- For sync worker process,
+  yield (`take_a_break`).
+- For user API worker process,
+  terminate returning 503 with reason "Computing miner signalling outcome".
+- For peer connection (gossip),
+  ignore logging info message.
+- For locally mined block (hence context of `aec_conductor` process),
+  ignore logging error message (as it should not have happened).
+- For mined block submitted by Stratum server (hence context of `aec_conductor` process),
+  ignore logging error message (as it should not have happened).
 
 ##### Blocks not necessarily connected to genesis
 
--- a/docs/design/PT-166642114-miner-activated-consensus.md
+++ b/docs/design/PT-166642114-miner-activated-consensus.md
@@ -59,7 +59,7 @@ For trusted blocks, the conductor must instruct the generation of the block cand
 In order to prevent long and potentially duplicated computational efforts - requiring multiple lookups into the chain storage -
 the computation of the outcome of the miner signalling should be asynchronous and lazy.
 
-Whenever a block at height preceding the signalling block interval end is inserted in the chain storage,
+Whenever a block at the height preceding the signalling block interval end (i.e. [HE][fspecs] minus 1) is inserted in the chain storage,
 the `aec_conductor` spawns - and keeps track of - a worker responsible for
 computing the outcome of the miner signalling - using non-transactional chain storage access -
 and storing such outcome (true / false) in an auxiliary table.
--- a/docs/design/PT-166642114-miner-activated-consensus.md
+++ b/docs/design/PT-166642114-miner-activated-consensus.md
@@ -59,7 +59,7 @@ For trusted blocks, the conductor must instruct the generation of the block cand
 In order to prevent long and potentially duplicated computational efforts - requiring multiple lookups into the chain storage -
 the computation of the outcome of the miner signalling should be asynchronous and lazy.
 
-Whenever a block at height preceding the signalling block interval end is inserter in the chain storage,
+Whenever a block at height preceding the signalling block interval end is inserted in the chain storage,
 the `aec_conductor` spawns - and keeps track of - a worker responsible for
 computing the outcome of the miner signalling - using non-transactional chain storage access -
 and storing such outcome (true / false) in an auxiliary table.
--- a/docs/design/PT-166642114-miner-activated-consensus.md
+++ b/docs/design/PT-166642114-miner-activated-consensus.md
@@ -1,0 +1,146 @@
+# Miner Signalled Consensus Upgrade: Design
+
+This document describes the technical specifications for the miner signalled consensus upgrade whose [functional specifications are available][fspecs].
+
+[fspecs]: https://github.com/aeternity/protocol/blob/1ccb7fb099f7da9ba682b0039d37790050499ed1/consensus/miner_signalled_consensus.md
+
+## Technical specifications
+
+### Behaviour
+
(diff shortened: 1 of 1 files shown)


docs/design/PT-166642114-miner-activated-consensus.md
```diff
@@ -0,0 +1,146 @@
...
+By exploiting such current design, the expected consensus protocol version of a block can be computed whenever:
+- An untrusted block is meant to be able to be inserted in the storage.
+  This includes the cases of:
+  - Sync worker process intending to add one block to the storage.
+  - Gossip.
+  - User API.
+- A trusted block is meant to be able to be inserted in the storage.
+  These blocks are not validated.
+  This includes the cases of:
+  - Locally mined or signed block.
```
jur0: There is also a case when the node runs as a Stratum server and gets a block from a client.

docs/design/PT-166642114-miner-activated-consensus.md
```diff
@@ -0,0 +1,146 @@
...
+
+For blocks mined or signed locally, the conductor must instruct the generation of the block candidate with the correct consensus protocol version.
+
+###### Asynchronous computation of expected consensus protocol version
+
+In order to prevent long and potentially duplicated computational efforts - requiring multiple lookups into the chain storage -
+the computation of the outcome of the miner signalling should be asynchronous and lazy.
+
+Whenever a block at height preceding the signalling block interval end is inserter in the chain storage,
+the `aec_conductor` spawns - and keeps track of - a worker responsible for
```
jur0: As discussed in the chat. A worker can be spawned at height `HE - 1` (signalling end interval - 1) and not necessarily for each inserted block within the signalling interval.

docs/design/PT-166642114-miner-activated-consensus.md
```diff
@@ -0,0 +1,146 @@
...
+### Data structures
+
+#### User configuration
+
+User configuration for accommodating the [defined parameters][fspecs].
+**TODO Detail format and constraints. (See [comment Pivotal Tracker ticket](https://www.pivotaltracker.com/story/show/166642114/comments/205266011) and following comments.)**
+
+#### Auxiliary table for asynchronous computation of outcome of signalling
+
+Table whose name is computed based on the set of [user configured parameters][fspecs] - including current consensus protocol version C1, and excluding the height H1 at which it was first effective.
```
jur0: The table might not be necessary as we could recompute the result and keep it in a worker process.

docs/design/PT-166642114-miner-activated-consensus.md
jur0: Also, the worker can be spawned at hard fork height - 1, so we know the result at fork height. As the signalling computation result could take a significant amount of time, it must be done earlier than hard fork height - 1.

docs/design/PT-166642114-miner-activated-consensus.md
lucafavatella: [Related refactoring PR](https://github.com/aeternity/aeternity/pull/2624).

docs/design/PT-166642114-miner-activated-consensus.md
```diff
@@ -0,0 +1,146 @@
...
+
+###### Asynchronous computation of expected consensus protocol version
+
+In order to prevent long and potentially duplicated computational efforts - requiring multiple lookups into the chain storage -
+the computation of the outcome of the miner signalling should be asynchronous and lazy.
+
+Whenever a block at height preceding the signalling block interval end is inserter in the chain storage,
+the `aec_conductor` spawns - and keeps track of - a worker responsible for
+computing the outcome of the miner signalling - using non-transactional chain storage access -
+and storing such outcome (true / false) in an auxiliary table.
```
jur0: We can use the worker process to query the signalling result. However, it the node is restarted, we will lose the result and need to recompute it. Is it worth having the extra table to keep the result just to cover the case when the node is restarted?

docs/design/PT-166642114-miner-activated-consensus.md
lucafavatella: Addressed in new commit afaa8484d.

docs/design/PT-166642114-miner-activated-consensus.md
lucafavatella: > As discussed in the chat. A worker can be spawned at height `HE - 1` (signalling end interval - 1) and not necessarily for each inserted block within the signalling interval.

Clarified in new commit baae00f12

lucafavatella: > Also, the worker can be spawned at hard fork height - 1, so we know the result at fork height. As the signalling computation result could take a significant amount of time, it must be done earlier than hard fork height - 1.

I believe this is addressed, as the worker is documented as spawned at "HE - 1".

docs/design/PT-166642114-miner-activated-consensus.md
lucafavatella: I like the idea. It simplifies things, at the cost that if on restart the outcome is required then the lazy computation plus yielding/ignoring/503 strategy kicks in.

A mitigation could be that insertion of any blocks in the range from signalling end (HE, included) to fork height (included) checks that there is a result for the outcome or a worker spawned, otherwise spawn. Though this could be listed as future enhancement.

docs/design/PT-166642114-miner-activated-consensus.md
lucafavatella: Or an in-memory table.

lucafavatella: Anyway the lookup table/dict for reaching the process is required.

So it looks like that, once worker finished computing outcome, outcome can be kept in some in-memory state (either in conductor state as dict, or in ETS conditionally created by conductor init, or in *overcomplicated* non-persisted Mnesia table).

jur0: Conductor's state is presumably the simplest option.

lucafavatella: Unresolved so to keep track of this (I have not addressed yet).

lucafavatella: Addressed in 5f917bbc8

docs/design/PT-166642114-miner-activated-consensus.md
lucafavatella: Addressed in 5f917bbc8

lucafavatella:
Rebased on recent master so to ease investigation for design.

docs/design/PT-166642114-miner-activated-consensus.md
```diff
@@ -0,0 +1,158 @@
...
+
+After having configured the node for miner signalling,
+if afterwards the user configures a fixed consensus protocol at a certain height and restarts the node
+the node will not re-check transactions for the now-invalid protocol.
+
+##### State channel client finite state machine
+
+No structural changes needed.
+
+**TODO Handling of transactions read from database after user configuration makes consensus upgrade unconditional.**
```
lucafavatella: This can be removed, as anyway new protocol is meant to be confirmed.

docs/design/PT-166642114-miner-activated-consensus.md
```diff
@@ -0,0 +1,158 @@
...
+
+##### User API
+
+**TODO The user API returns transactions to the user. Mark those APIs as deprecated and remove those? Let user specify consensus protocol?**
+
+##### Other reads
+
+**TODO Handling of transactions read from database after user configuration makes consensus upgrade unconditional.**
```
lucafavatella: No problem. To remove.

docs/design/PT-166642114-miner-activated-consensus.md
```diff
@@ -0,0 +1,158 @@
...
+
+##### State channel client finite state machine
+
+No structural changes needed.
+
+**TODO Handling of transactions read from database after user configuration makes consensus upgrade unconditional.**
+
+##### User API
+
+**TODO The user API returns transactions to the user. Mark those APIs as deprecated and remove those? Let user specify consensus protocol?**
```
lucafavatella: Re serialization of transactions: add note re that if any ambiguity on serialization of transaction depending on protocol, then the user API shall be enhanced with new optional field so that user can control the protocol version.

lucafavatella: Re dry run API: already the API relies on:
- Either the top or a specified block hash.
- And the serialized transactions. The normal aetx:process is called.

So no need to enhance that for miner signalled hard fork.

docs/design/PT-166642114-miner-activated-consensus.md
lucafavatella: Removed in new commit.

docs/design/PT-166642114-miner-activated-consensus.md
lucafavatella: Removed in new commit.

docs/design/PT-166642114-miner-activated-consensus.md
lucafavatella: Addressed in new commit.

lucafavatella: Addressed in new commit.

jur0:
This wasn't meant to be merged...

PR #2614, opened 2019-08-04
Aens subdomains

skkw:


lucafavatella:
I took the liberty to mark as WIP as not ready to merge e.g. conflict.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -2209,6 +2209,7 @@ nameservice_transactions(_Config) ->
     nameservice_transaction_update(MinerAddress, MinerPubkey),
     nameservice_transaction_transfer(MinerAddress, MinerPubkey),
     nameservice_transaction_revoke(MinerAddress, MinerPubkey),
+    nameservice_transaction_subname(MinerAddress, MinerPubkey),
```
lucafavatella: Please update comment at top or delete it.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -2369,6 +2370,54 @@ nameservice_transaction_revoke(MinerAddress, MinerPubkey) ->
     test_missing_address(account_id, Encoded, fun get_name_revoke/1),
     ok.
 
+nameservice_transaction_subname(MinerAddress, MinerPubkey) ->
```
lucafavatella: Should not this test be running only for `ct-lima` CI job (or equivalent from point of view of consensus) and later? Is it conditional somewhere and I am not noticing it?

Or at least what is being tested should be different depending on consensus version e.g. not accepted before Lima, accepted from Lima.

apps/aens/src/aens_utils.erl
```diff
@@ -7,15 +7,16 @@
...
+
 %% API
 -export([check_name_claimed_and_owned/3,
-         to_ascii/1]).
-
-%%%===================================================================
-%%% Types
-%%%===================================================================
-
--define(LABEL_SEPARATOR, <<".">>).
```
lucafavatella: I expect conflicts in master. Leading to multiple PRs rebased and re-reviewed multiple times... May I suggest to split out of this PR all preparation bits (e.g. refactoring)?

apps/aens/test/aens_SUITE.erl
```diff
@@ -560,3 +578,122 @@ do_prune_until(N1, N1, OTree) ->
...
+-define(TOPNAME, <<"姆.test"/utf8>>).
+-define(SNAME1, <<"斯1.姆.test"/utf8>>).
+-define(SNAME2, <<"姆2.姆.test"/utf8>>).
+-define(SNAME21, <<"姆2.斯1.姆.test"/utf8>>).
+-define(SNAME54321, <<"詹5.斯4.詹3.姆2.斯1.姆.test"/utf8>>).
+
+-define(SNAME321, <<"詹3.姆2.斯1.姆.test"/utf8>>).
+-define(SNAME4321, <<"斯4.詹3.姆2.斯1.姆.test"/utf8>>).
+
+subname(_Cfg) ->
```
lucafavatella: Should not this test be running only for `ct-lima` CI job (or equivalent from point of view of consensus) and later? Is it conditional somewhere and I am not noticing it?

apps/aeprimop/src/aeprimop_state.erl
```diff
@@ -169,7 +169,7 @@ get_name(Key, S) ->
     get_x(name, Key, name_does_not_exist, S).
 
 put_name(Object, S) ->
-    cache_put(name, Object, S).
+    cache_put(element(1, Object), Object, S).
```
lucafavatella: Please split as put_subname. Or is there a reason for being dynamic here?

apps/aens/src/aens_subname_tx.erl
```diff
@@ -0,0 +1,227 @@
...
+account_pubkey(#ns_subname_tx{account_id = AccountId}) ->
+    aeser_id:specialize(AccountId, account).
+
+-spec version(tx()) -> non_neg_integer().
+version(_) ->
+    ?SUBNAME_TX_VSN.
+
+-spec valid_at_protocol(aec_hard_forks:protocol_vsn(), tx()) -> boolean().
+valid_at_protocol(_, _) ->
+    true.
```
lucafavatella: ?

apps/aehttp/test/aehttp_integration_SUITE.erl
skkw: updated comment

skkw:
> The new hash computation is incompatible with the old hash computation. This means that there could be a preclaim on chain with a hash valid before this change, which is now an invalid transaction on chain... in orther words, this new code cannot sync with existing chains.

The difference in behavior happens only when we are computing hash of invalid name (like: "name_without_registrar.") 

In older version of node, this is detected and rejected on different place, but the result is the same - TX is not included.

skkw:
We will come up with better - incremental - subdomains implementation. Keeping this branch for archival at the moment.

PR #2617, opened 2019-08-05
Fix SC WebSocket API examples' generation

velzevur:
PT [SC WS API: Fix examples generation](https://www.pivotaltracker.com/story/show/167702900)

For some time now the generation of examples had been broken, this PR fixes it.

Result of execution can be seen in aeternity/protocol#393


apps/aehttp/test/aehttp_sc_SUITE.erl
```diff
@@ -2890,9 +2877,20 @@ log_basename(Config) ->
     GOpts = ?config(tc_group_properties, Config),
     GName = ?config(name, GOpts),
     SubDir =
-        case GName =:= continuous_sc_ws of
```
lucafavatella: I see this expression is main fix.

PR #2620, opened 2019-08-06
PT-166642114 Hard fork signalling (config, system tests)

jur0:
https://www.pivotaltracker.com/n/projects/2124891/stories/166642114

Contains config schema update and system tests - meant to me merged.

### DIFF
--- a/apps/aecore/src/aec_hard_forks.erl
+++ b/apps/aecore/src/aec_hard_forks.erl
@@ -259,7 +259,7 @@ assert_fork_height(SignallingEndHeight, ForkHeight)
 assert_fork_height(_SignallingEndHeight, ForkHeight) ->
     error({illegal_fork_height, ForkHeight}).
 
-assert_fork_version(Version) when Version > ?LIMA_PROTOCOL_VSN ->
+assert_fork_version(Version) when Version > ?MINERVA_PROTOCOL_VSN ->
     ok;
 assert_fork_version(Version) ->
     error({illegal_fork_version, Version}).
--- a/apps/aecore/src/aec_hard_forks.erl
+++ b/apps/aecore/src/aec_hard_forks.erl
@@ -2,6 +2,7 @@
 
 -export([check_env/0]).
 -export([protocols/0,
+         fork/1,
          check_protocol_version_validity/2,
          protocol_effective_at_height/1,
          is_fork_height/1
@@ -30,8 +31,16 @@
                       | ?ROMA_PROTOCOL_VSN.
 -type protocols() :: #{protocol_vsn() => aec_blocks:height()}.
 
+-type fork() :: #{signalling_start_height => aec_blocks:height(),
+                  signalling_end_height   => aec_blocks:height(),
+                  signalling_block_count  => pos_integer(),
+                  fork_height             => aec_blocks:height(),
+                  info_field              => non_neg_integer(),
+                  version                 => version()}.
+
 -export_type([ protocols/0
              , protocol_vsn/0
+             , fork/0
              ]).
 %%%===================================================================
 %%% API
@@ -42,9 +51,17 @@ protocols() ->
     NetworkId = aec_governance:get_network_id(),
     protocols_from_network_id(NetworkId).
 
+%% If there is a fork specified in the config (fork_management > fork) and the
+%% Height parameter is less than or equal the fork height, the info about
+%% the fork is returned.
+-spec fork(aec_block:height()) -> fork() | undefined.
+fork(Height) ->
+    fork_from_height(Height).
+
 -spec check_env() -> ok.
 check_env() ->
-    assert_protocols(protocols()).
+    assert_protocols(protocols()),
+    assert_fork(fork_config()).
 
 -spec check_protocol_version_validity(version(), aec_blocks:height()) ->
                                          ok | {error, Reason} when
@@ -107,10 +124,20 @@ protocols_from_network_id(_ID) ->
              };
         M when is_map(M) ->
             maps:fold(fun(K, V, Acc) ->
-                              Acc#{binary_to_integer(K) => V} 
+                              Acc#{binary_to_integer(K) => V}
                       end, #{}, M)
     end.
 
+fork_from_height(Height) ->
+    case fork_config() of
+        #{fork_height := ForkHeight} = Fork ->
+            case Height =< ForkHeight of
+                true  -> Fork;
+                false -> undefined
+            end;
+        undefined ->
+            undefined
+    end.
 
 %% Exported for tests
 check_protocol_version_validity(Version, Height, Protocols) ->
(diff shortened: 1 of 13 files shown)


apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -934,14 +934,44 @@
...
+                            "description" : "Height at which signalling period starts.",
+                            "type" : "integer"
+                        },
+                        "signalling_end_height" : {
+                            "description" : "Height at which signalling period ends.",
+                            "type" : "integer"
+                        },
+                        "signalling_block_count" : {
+                            "description" : "Count of the key blocks in signalling period necessary for the fork to take effect.",
+                            "type" : "integer"
```
lucafavatella: Positive integer?

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -934,14 +934,44 @@
...
+                        },
+                        "fork_height" : {
+                            "description" : "Height at which the fork should occur.",
+                            "type" : "integer"
+                        },
+                        "info_field" : {
+                            "description" : "Pseudorandom number (part of key block) used by the miners indicating which fork they support.",
+                            "type" : "integer"
```
lucafavatella: 🤔 

Is there any other portion of the user experience where this value is exposed? If so we could use that (`cb_...`?).

Though number seems simpler... I like it. May I suggest limiting it to between 0 and the max that can be expressed in the info field?

jur0: Let's try using the number (at least for now), it looks simpler. Added `maximum` in the schema.

apps/aecore/src/aec_hard_forks.erl
```diff
@@ -30,8 +31,16 @@
                       | ?ROMA_PROTOCOL_VSN.
 -type protocols() :: #{protocol_vsn() => aec_blocks:height()}.
 
+-type fork() :: #{signalling_start_height => aec_blocks:height(),
```
lucafavatella: I was expecting mandatory i.e. `:=` (in all map).

apps/aeutils/test/aeu_env_tests.erl
```diff
@@ -28,7 +28,15 @@ extra_checks_test_() ->
...
       {"User configuration cannot contain both 'mining > cuckoo > miners' and deprecated 'mining > cuckoo > miner'",
-       fun deprecated_miner_section_conflicting_with_miners/0}]
+       fun deprecated_miner_section_conflicting_with_miners/0},
+      {"User configuration cannot contain 'fork_management > fork > signalling_start_height' greater or equal to 'fork_management > fork > signalling_end_height'",
+       fun invalid_fork_signalling_interval/0},
+      {"User configuration cannot contain 'fork_management > fork > signalling_block_count' greater than signalling interval",
+       fun invalid_fork_signalling_block_count/0},
+      {"User configuration cannot contain 'fork_management > fork > fork_height' lower or equal to 'fork_management > fork > signalling_end_height'",
+       fun invalid_fork_height/0},
+      {"User configuration cannot contain 'fork_management > fork > version' lower or equal to Lima protocol version (4)",
```
lucafavatella: I [would rather expect Minerva](https://github.com/aeternity/aeternity/pull/2613/files#diff-119c822db20ee796b91225c284221d8eR15).

lucafavatella: Because of info field.

lucafavatella: Strictly greater than Minerva.

jur0: Fixed.

system_test/common/helpers/aest_docker.erl
```diff
@@ -223,19 +223,46 @@ setup_node(Spec, BackendState) ->
         end,
     HardForkVars =
         case maps:find(hard_forks, Spec) of
-            error -> #{};
             {ok, HardForks} ->
-                #{hard_forks_present => [#{}],
```
lucafavatella: Did this key have a special meaning in the mustache library? I cannot find any use.

jur0: I tried to look it up using grep, it's not used anywhere.

system_test/common/helpers/aest_docker.erl
```diff
@@ -223,19 +223,46 @@ setup_node(Spec, BackendState) ->
...
         end,
-    RootVars = (maps:merge(CuckooMinerVars, HardForkVars))#{
+    RootVars = maps:merge(CuckooMinerVars, HardForkVars),
+    RootVars1 = maps:merge(RootVars, ForkManagementVars),
+    RootVars2 = RootVars1#{
         hostname => Name,
         ext_addr => format("http://~s:~w/", [Hostname, ?EXT_HTTP_PORT]),
         peers => PeerVars,
         key_password => ?PEER_KEYS_PASSWORD,
-        config => ExtraConfig,
```
lucafavatella: ???
So now this variable is bound (from key `config` documented in type spec) but unused.

Not sure why fork-specific handling - rather than re-using this ExtraConfig - though there must be reason so ok...

system_test/common/helpers/aest_docker.erl
lucafavatella: I see it was used in past test. Ok.

system_test/common/aest_community_fork_SUITE.erl
```diff
@@ -0,0 +1,166 @@
...
+          fork_management => ?FORK,
+          mine_rate       => 1000,
+          backend         => aest_docker,
+          source          => {pull, "aeternity/aeternity:local"}}).
+
+%% This node is not configured for fork.
+-define(MINER_NODE3,
+        #{name            => miner_node3,
+          peers           => [miner_node1, miner_node2],
+          mine_rate       => 1000000, %% The node shouldn't mine so it doesn't produce blocks with wrong info field.
```
lucafavatella: This may affect block validation. Is there way to properly disable mining autostart?

lucafavatella: aest_db_SUITE seems to have `autostart: {{mining.autostart}}` in the config. Would that help?

system_test/common/aest_community_fork_SUITE.erl
```diff
@@ -0,0 +1,166 @@
...
+                signalling_end_height => ?SIGNALLING_END_HEIGHT,
+                signalling_block_count => ?SIGNALLING_BLOCK_COUNT,
+                fork_height => ?FORK_HEIGHT,
+                info_field => ?INFO_FIELD,
+                version => ?VERSION}}).
+
+-define(MINER_NODE1,
+        #{name            => miner_node1,
+          peers           => [miner_node2, miner_node3],
+          fork_management => ?FORK,
```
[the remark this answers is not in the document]
jur0: This depends on `fork_management > network_id` (which is `ae_system_tests` for system tests), therefore [this applies](https://github.com/aeternity/aeternity/blob/master/apps/aecore/src/aec_hard_forks.erl#L106-L108).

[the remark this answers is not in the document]
jur0: Also, we can set it in the config in `chain > hard_forks` for this test.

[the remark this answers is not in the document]
jur0: > shall we introduce constraint that user config with community fork must also explicitly set other hard forks? (I would say no need to.)

I think it shouldn't be necessary.

system_test/common/aest_community_fork_SUITE.erl
jur0: Fixed.

jur0:
Closing as there is a new PR: #2640

PR #2621, opened 2019-08-06
Pt 167126818

ThomasArts:
Adding abi_version to compiler result

The changes here are only making serialization format used explicit. Tests do improve a bit when new format isa dded in the future.

### DIFF
--- a/apps/aecontract/src/aect_sophia.erl
+++ b/apps/aecontract/src/aect_sophia.erl
@@ -12,8 +12,7 @@
 -include("../../aecontract/include/hard_forks.hrl").
 -include("../../aecontract/src/aect_sophia.hrl").
 
--export([ serialize/1
-        , serialize/2
+-export([ serialize/2
         , deserialize/1
         , is_legal_serialization_at_height/2
         , prepare_for_json/2
@@ -22,22 +21,16 @@
 -type wrapped_code() :: #{ source_hash := aec_hash:hash()
                          , type_info   := [{binary(), binary(), binary(), binary()}]
                          , byte_code   := binary()
-                         , compiler_version := binary()
+                         , compiler_version => binary()
                          , contract_vsn := integer()
                          }.
 
 -export_type([ wrapped_code/0 ]).
 
-%% After hard fork Minerva, we accept ?SOPHIA_CONTRACT_VSN_2 in serialization
--define(SOPHIA_CONTRACT_VSN, ?SOPHIA_CONTRACT_VSN_2).
-
 is_legal_serialization_at_height(?SOPHIA_CONTRACT_VSN_1, _Height) ->
     true;
 is_legal_serialization_at_height(?SOPHIA_CONTRACT_VSN_2, Height) ->
     aec_hard_forks:protocol_effective_at_height(Height) >= ?MINERVA_PROTOCOL_VSN.
-
-serialize(Map) ->
-    serialize(Map, ?SOPHIA_CONTRACT_VSN).
 
 serialize(#{byte_code := ByteCode, type_info := TypeInfo,
             contract_source := ContractString, compiler_version := Version}, SophiaContractVersion) ->
@@ -49,7 +42,6 @@ serialize(#{byte_code := ByteCode, type_info := TypeInfo,
              , {type_info, TypeInfo}
              , {byte_code, ByteCode} ] ++
              [ {compiler_version, BinVersion} || SophiaContractVersion > ?SOPHIA_CONTRACT_VSN_1 ],
-             %% Add version here in release when Minerva height has been reached
     aeser_chain_objects:serialize(compiler_sophia,
                                        SophiaContractVersion,
                                        serialization_template(SophiaContractVersion),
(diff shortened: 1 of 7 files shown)


apps/aecontract/src/aect_sophia.erl
```diff
@@ -83,6 +79,21 @@ deserialize(Binary) ->
              , compiler_version => CompilerVersion
              , contract_vsn => Vsn
              };
+        {compiler_sophia = Type, ?SOPHIA_CONTRACT_VSN_3 = Vsn, _Rest} ->
```
lucafavatella: I expect corresponding PR in protocol repo.

lucafavatella: Worth a line in release notes in docs/release-notes/next-lima ?

rebar.lock
```diff
@@ -1,7 +1,7 @@
 {"1.1.0",
 [{<<"aebytecode">>,
   {git,"https://github.com/aeternity/aebytecode.git",
-      {ref,"17c2a93e7213387a0fd9f9f9395e12de50355982"}},
+      {ref,"2a9035d5ef72bc65e51e4b9a1a958bb8ba2c6747"}},
```
lucafavatella: https://github.com/aeternity/aebytecode/compare/17c2a93e7213387a0fd9f9f9395e12de50355982...2a9035d5ef72bc65e51e4b9a1a958bb8ba2c6747

apps/aestratum/src/aestratum_deploy.erl
```diff
@@ -16,7 +16,7 @@ deploy_payout_contract(#{public := PubKey, secret := PrivKey}) ->
     {ok, WrappedTx}  =
         aect_create_tx:new(#{owner_id    => aeser_id:create(account, PubKey),
                              nonce       => aec_accounts:nonce(Account) + 1,
-                             code        => aect_sophia:serialize(compiled_contract()),
+                             code        => aect_sophia:serialize(compiled_contract(), 2),
```
lucafavatella: I see this fixes to SOPHIA_CONTRACT_VSN_2 so not to change semantics. Ok.

apps/aecontract/src/aect_sophia.erl
```diff
@@ -22,34 +21,31 @@
...
-
 is_legal_serialization_at_height(?SOPHIA_CONTRACT_VSN_1, _Height) ->
     true;
 is_legal_serialization_at_height(?SOPHIA_CONTRACT_VSN_2, Height) ->
-    aec_hard_forks:protocol_effective_at_height(Height) >= ?MINERVA_PROTOCOL_VSN.
-
-serialize(Map) ->
-    serialize(Map, ?SOPHIA_CONTRACT_VSN).
+    aec_hard_forks:protocol_effective_at_height(Height) >= ?MINERVA_PROTOCOL_VSN;
+is_legal_serialization_at_height(?SOPHIA_CONTRACT_VSN_3, Height) ->
```
lucafavatella: Ideally this shall move away from assuming protocol known at height. Though orthogonal.

apps/aecontract/src/aect_sophia.erl
```diff
@@ -22,34 +21,31 @@
 -type wrapped_code() :: #{ source_hash := aec_hash:hash()
                          , type_info   := [{binary(), binary(), binary(), binary()}]
                          , byte_code   := binary()
-                         , compiler_version := binary()
+                         , compiler_version => binary()
```
lucafavatella: I see - this is type correction preceding the new format.

apps/aecontract/src/aect_sophia.erl
```diff
@@ -12,8 +12,7 @@
 -include("../../aecontract/include/hard_forks.hrl").
 -include("../../aecontract/src/aect_sophia.hrl").
 
--export([ serialize/1
-        , serialize/2
+-export([ serialize/2
         , deserialize/1
```
lucafavatella: This module returns data that is not opaque so it is not immediate to follow usage, though based on the nature of the change and a quick look to DB deserialization it seems there shall be no issues in loading old DB.

apps/aestratum/src/aestratum_deploy.erl
ThomasArts: Actually it does change semantics... it was wrong in aestratum_deploy were the Roma version was used to serialize a contract

apps/aecontract/src/aect_sophia.erl
ThomasArts: And the new contract version turned out not to be necessary, we can do with the old version. So, new update code has only minor changes.

apps/aestratum/src/aestratum_deploy.erl
hanssv: Nope, we were just blind... This is equivalent.

apps/aecontract/src/aect_sophia.erl
```diff
@@ -22,23 +21,17 @@
...
                          , byte_code   := binary()
-                         , compiler_version := binary()
+                         , compiler_version => binary()
                          , contract_vsn := integer()
                          }.
 
 -export_type([ wrapped_code/0 ]).
 
-%% After hard fork Minerva, we accept ?SOPHIA_CONTRACT_VSN_2 in serialization
--define(SOPHIA_CONTRACT_VSN, ?SOPHIA_CONTRACT_VSN_2).
```
hanssv: The default was already pointing to VSN_2 - I think we were temporarily blind or something...

ThomasArts: I blame it to the hiolidays, but now it is at least very explicitely right.

apps/aecontract/src/aect_sophia.erl
lucafavatella: Better!

PR #2623, opened 2019-08-07
Preparation for miner signalled consensus upgrade

lucafavatella:
Prepared while designing https://www.pivotaltracker.com/story/show/166642114

Supersedes #2618 (fixes `aec_block_key_candidate_tests `)

TODO:
* [x] Relax from `aec_headers:deserialize_key_from_binary` the check of the consensus based on the height.
* [x] `aec_headers:validate_key_block_header` has a consensus-height check `validate_version`, such `aec_headers:validate_key_block_header` is called only from `aec_conductor:post_block` and `aec_conductor:add_synced_block`, and in both points the block being added is meant to be connected to genesis. Restructure code so to rely on miner signalling having ended already.
* [x] `aec_headers:validate_micro_block_header` has a consensus-height check `validate_version`, and is called from the following points:
  -  `aec_conductor:post_block` and `aec_conductor:add_synced_block`, and in both points the block being added is meant to be connected to genesis. Restructure code so to rely on miner signalling having ended already.
  - `aec_peer_connection:handle_new_micro_block`: function `pre_assembly_check` already has assumption - in a check - that previous key block is present, and that hence the micro block signature can be checked. Such assumption of knowledge of key block may be exploited, relying on consensus meant to change only at key blocks.
* [x] In `aec_chain_state:calculate_state_for_new_keyblock`, in line `fake_key_node` compute consensus version considering user config and miner signalling. This is possible because this function is called from two points:
  - `create_key_block_candidate` in `aec_conductor`. This bases the candidate on the top block.
  - `get_pending_key_block` in `aec_conductor`. This bases the candidate on the top block.
    - Orthogonal question: is this function needed at all?
* [ ] ...

### DIFF
--- a/apps/aecore/src/aec_trees.erl
+++ b/apps/aecore/src/aec_trees.erl
@@ -45,7 +45,7 @@
          apply_txs_on_state_trees/4,
          apply_txs_on_state_trees_strict/3,
          grant_fee/3,
-         perform_pre_transformations/2
+         perform_pre_transformations/4
         ]).
 
 %% Proof of inclusion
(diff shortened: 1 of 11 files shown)


lucafavatella:
Rebased on current master.

apps/aecore/test/aec_chain_tests.erl
```diff
@@ -1968,13 +1968,6 @@ hard_fork_inserts_new_accounts() ->
...
 
 meck_minerva_fork_height(Height) ->
-    meck:expect(aec_hard_forks, is_fork_height,
-                fun(H) ->
-                    case H =:= Height of
-                        true -> {true, ?MINERVA_PROTOCOL_VSN};
-                        false -> false
-                    end
-                end),
     meck:expect(aec_hard_forks, protocol_effective_at_height,
```
lucafavatella: This mocking needs to be double checked.

apps/aecore/src/aec_trees.erl
```diff
@@ -45,7 +45,7 @@
          apply_txs_on_state_trees/4,
          apply_txs_on_state_trees_strict/3,
          grant_fee/3,
-         perform_pre_transformations/2
```
lucafavatella: I expect this to require update of tests.

PR #2627, opened 2019-08-07
SC reestablish: fix issue

velzevur:
Fix to the issue @AleksandarFilipov found
No PT, no release notes line


apps/aechannel/src/aesc_offchain_state.erl
```diff
@@ -90,7 +90,7 @@ recover_from_offchain_tx(#{ existing_channel_id := ChId
                 false ->
                     {error, latest_state_mismatch}
             end;
-        error ->
+        {error, _} ->
```
velzevur: `aesc_state_cache:reestablish/2` [error return value](https://github.com/aeternity/aeternity/blob/master/apps/aechannel/src/aesc_state_cache.erl#L113-L114)

PR #2634, opened 2019-08-09
change count_authentications() strategy

uwiger:
Testing a different way of counting authentications


uwiger:
> Looks good, but keep in mind that this will still fail in cases in which one of the peers double signs the TX(for instance Alice signs then Bob then Alice again resulting in `A1(A2(B1(B2(A3(A4(Tx))))))`) - fortunately currently there are no transactions where this is allowed.

Yes, so I added a comment to the function, which tries to outline where it would fail.

The fsm should perhaps be a bit more restrictive when checking the signing responses, but OTOH, we must account for the SC peer not using the fsm at all, and doing all sorts of crazy stuff. In this case, if the algo fails, it should fail predictably and gracefully.

PR #2635, opened 2019-08-09
External localnet configuration/repo

dincho:
relates to: https://github.com/aeternity/localnet/pull/1
[PT #165181713](https://www.pivotaltracker.com/story/show/165181713)

Test runs: https://circleci.com/workflow-run/9fbed746-78f8-42ce-ac91-fdbbd53528d5


.circleci/config.yml
```diff
@@ -750,24 +746,18 @@ jobs:
       - *fail_notification
 
   docker_smoke_tests:
-    <<: *machine_config
+    executor: buildpack
```
lucafavatella: This commit is:
- Splitting localnet to other repo.
- Changing machine in which the Docker image is built.
- Reduce smoke testing on built Docker image to healthcheck of single image.

In particular, now machine on which Docker image is built is distinct from machine on which system tests are built. Could system tests be moved to same kind of machine?

Anyway seems good enough.

lucafavatella: Please review https://github.com/aeternity/aeternity/wiki/Release-process on whether you intended new check to be added in release process. It seems good enough to me, though for 4.2.0 you may want to perform some extra testing.

dincho: > In particular, now machine on which Docker image is built is distinct from machine on which system tests are built. Could system tests be moved to same kind of machine?

Unfortunately system tests are using bind mounts which does not work with remote docker machine. If we somehow replace the bind mounts with `docker cp` or other alternative we can probably make it work. I'm not sure I'm qualified enough to change system tests infrastructure tho.

dincho: > Please review https://github.com/aeternity/aeternity/wiki/Release-process on whether you intended new check to be added in release process. It seems good enough to me, though for 4.2.0 you may want to perform some extra testing.

Do you have some specific additional tests in mind ? We already have connectivity tests for the docker image for testnet and mainnet in the release process document

lucafavatella: Seems fine.

After this PR merged, you may want to check that localnet repo still works with master version of this docker image.

lucafavatella:
I see https://github.com/aeternity/localnet/pull/1 is merged.

PR #2636, opened 2019-08-12
Add meta update object to SC offchain updates

uwiger:
See [PT #167808073](https://www.pivotaltracker.com/story/show/167808073)

This PR adds a `{meta, binary()}` update op to the channel_offchain_updates. This could be added to any update request, but is currently only supported by the fsm for `upd_transfer`. The `fsm_SUITE` has been slightly extended to verify that meta objects can be added, but the WS API has not yet been changed to support them.

This PR depends on a PR to aeserialization.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -65,6 +65,7 @@
         , record_fields/1
         , report_tags/0
         , timeouts/0
+        , version_tags/0
         ]).
 
 %% Used by noise session
@@ -300,6 +301,9 @@ record_fields(Other) -> aesc_offchain_state:record_fields(Other).
 
 report_tags() ->
     maps:keys(?DEFAULT_REPORT_FLAGS).
+
+version_tags() ->
+    [offchain_update].
 
 timeouts() ->
     maps:keys(?DEFAULT_TIMEOUTS).
@@ -3070,7 +3074,8 @@ init(#{opts := Opts0} = Arg) ->
               fun(O) -> check_minimum_depth_opt(DefMinDepth, Role, O) end,
               fun check_timeout_opt/1,
               fun check_rpt_opt/1,
-              fun check_log_opt/1
+              fun check_log_opt/1,
+              fun check_version_opts/1
              ], Opts2),
     #{initiator := Initiator} = Opts,
     Session = start_session(Arg, Reestablish, Opts#{role => Role}),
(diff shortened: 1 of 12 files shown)


lucafavatella:
Kept WIP to label only - for avoiding duplication.

apps/aechannel/src/aesc_offchain_update.erl
```diff
@@ -102,6 +112,7 @@ from_db_format(Tuple) ->
         #deposit{} -> Updated;
         #create_contract{} -> Updated;
         #call_contract{} -> Updated;
+        #meta{} -> Updated;
```
velzevur:
```suggestion
```
This is not reachable

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -259,11 +260,14 @@ upd_deposit(Fsm, #{amount := Amt} = Opts) when is_integer(Amt) ->
     lager:debug("upd_deposit(~p)", [Opts]),
     gen_statem:call(Fsm, {upd_deposit, Opts}).
 
-upd_transfer(_Fsm, _From, _To, Amount) when Amount < 0 ->
+upd_transfer(Fsm, From, To, Amount) ->
+    upd_transfer(Fsm, From, To, Amount, #{}).
+
+upd_transfer(_Fsm, _From, _To, Amount, _Opts) when Amount < 0 ->
```
tolbrino: Possibly unrelated: Why would a transfer of 0 make sense?

rebar.config
```diff
@@ -68,7 +68,7 @@
         {aebytecode, {git, "https://github.com/aeternity/aebytecode.git", {ref, "10cc127"}}},
 
         {aeserialization, {git, "https://github.com/aeternity/aeserialization.git",
-                          {ref,"4a07297"}}},
+                          {ref,"d87adfd"}}},
```
tolbrino: Note: Requires https://github.com/aeternity/aeserialization/pull/9 to be merged first.

rebar.lock
```diff
@@ -18,7 +18,7 @@
   0},
  {<<"aeserialization">>,
   {git,"https://github.com/aeternity/aeserialization.git",
-      {ref,"4a072977745571ceb6cf34e6387a96b808ae20b3"}},
+      {ref,"d87adfd58333b70a725e707a2caf8a4604e6a126"}},
```
tolbrino: Note: Requires aeternity/aeserialization#9 to be merged first.

apps/aechannel/src/aesc_fsm.erl
velzevur: This is a funny edge case. I don't see a reason of forbidding it, nor do I see any meaningful use of it.

uwiger: Well, it doesn't break anything, so why would we disallow it? ;-)

apps/aechannel/src/aesc_offchain_update.erl
```diff
@@ -320,62 +339,70 @@ fields2update(call_contract, [ {caller, CallerId},
...
     [ {caller,      id},
       {contract,    id},
       {abi_version, int},
       {amount,      int},
       {gas,         int},
       {gas_price,   int},
       {call_data,   binary},
-      {call_stack,  [int]}].
+      {call_stack,  [int]}];
+update_serialization_template(?UPDATE_VSN, meta) ->
```
velzevur: I see this is allowed to be deserialized only using the new version. OK.

uwiger:
Updated deps in line with @tolbrino's comments.

apps/aechannel/src/aesc_offchain_update.erl
```diff
@@ -1,6 +1,7 @@
 -module(aesc_offchain_update).
 
--define(UPDATE_VSN, 1).
+-define(UPDATE_VSN, 2).
```
velzevur: This makes [serialization](https://github.com/aeternity/aeternity/blob/2db747d7e378d4bf5235e532edc5a75ae150f0ed/apps/aechannel/src/aesc_offchain_update.erl#L265) to use v2 off-chain serialization. This is good but being not backwards compatible regarding the off-chain protocol communication, I expect [some system tests](https://github.com/aeternity/aeternity/blob/master/system_test/common/aest_channels_SUITE.erl#L16-L19) to fail

PR #2637, opened 2019-08-12
WIP ecrecover

johnsnewby:


apps/aeutils/src/aeu_crypto.erl
```diff
@@ -11,6 +11,7 @@
         , ecdsa_to_der_pk/1
         , ecdsa_to_der_sig/1
 
+	, ecrecover/5
```
radrow: Avoid tabs, we use spaces ( :( )
```suggestion
        , ecrecover/5
```

PR #2638, opened 2019-08-12
[PT-167805291] Add opcode for ecrecover

johnsnewby:
Ref [PT-16780529](https://www.pivotaltracker.com/story/show/167805291)

Merge order:

- [x] PR for aebytecode: https://github.com/aeternity/aebytecode/pull/64
- [x] PR for aesophia: https://github.com/aeternity/aesophia/pull/122
- this PR


apps/aeutils/src/aeu_crypto.erl
```diff
@@ -11,6 +11,7 @@
         , ecdsa_to_der_pk/1
         , ecdsa_to_der_sig/1
 
+	, ecrecover/5
```
radrow: Use spaces
```suggestion
        , ecrecover/5
```

lucafavatella:
Moved WIP to label for easing filtering.

rebar.config
```diff
@@ -66,14 +66,17 @@
                    {ref, "0a82f0f"}}},
 
         {aebytecode, {git, "https://github.com/aeternity/aebytecode.git",
-                      {ref, "2a9035d"}}},
+                      {ref, "7c6a80f"}}},
```
radrow: Isn't it going to change anyway after merge?

tolbrino: Yes.

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -4368,6 +4369,27 @@ sophia_crypto(_Cfg) ->
                                 , {test_string_verify_secp256k1, Message, true}
                                 , {test_string_verify_secp256k1, <<"Not the secret message">>, false}] ],
 
+    %% Test ecrecover
+    %%   These static examples are taken from
+    %%   https://github.com/aeternity/parity-ethereum/blob/master/ethcore/builtin/src/lib.rs#L656
```
hanssv: It would be nice to check that we can actually sign things and get it through ecrecover (like we do for ecverify). I.e. start from a message and a keypair?!

apps/aeutils/src/aeu_crypto.erl
```diff
@@ -32,6 +33,19 @@ ecdsa_to_der_sig(<<R0:32/binary, S0:32/binary>>) ->
     {LR, LS} = {byte_size(R1), byte_size(S1)},
     <<16#30, (4 + LR + LS), 16#02, LR, R1/binary, 16#02, LS, S1/binary>>.
 
+%% ECRECOVER
+ecrecover(secp256k1, Hash, Sig) ->
+    % ecrecover expects the signature to be 96 bytes long
```
hanssv: Why? This is Erlang?

apps/aecontract/test/aecontract_SUITE.erl
tolbrino: Ok, then I'll extend that.

apps/aevm/src/aevm_ae_primops.erl
```diff
@@ -643,6 +645,8 @@ crypto_call(Gas, Op, _Value, Data, State) ->
             crypto_call(Gas, Op, Data, State)
     end.
 
+crypto_call(Gas, ?PRIM_CALL_CRYPTO_ECRECOVER_SECP256K1, Data, State) ->
```
hanssv: There needs to be a check (^^) that we are only allowed to run this if VM >= ?VM_AEVM_SOPHIA_4

tolbrino: Thanks. I added a check.

hanssv: Not really where I'd have put it, but it works :-)

apps/aevm/src/aevm_ae_primops.erl
```diff
@@ -666,6 +675,13 @@ crypto_call(_Gas, ?PRIM_CALL_CRYPTO_BLAKE2B_STRING, Data, State) ->
 crypto_call(_, _, _, _) ->
     {error, out_of_gas}.
 
+crypto_call_ecrecover_secp256k1(_Gas, Data, State) ->
+    [MsgHash0, Sig0] = get_args([hash_t(), bytes_t(65)], Data),
+    MsgHash = <<MsgHash0:32/unit:8>>,
+    <<_:31/binary, Sig:65/binary>> = words_to_bin(96, Sig0),
```
UlfNorell: This is wrong. Byte arrays are padded on the right, not on the left. See

https://github.com/aeternity/aebytecode/blob/befa1e3ff9c53c1054917c82e3b30c1c3c743f4f/src/aeb_memory.erl#L13-L18

Suggests that there also need to be tests that goes through the proper calldata creation instead of the hack in `aecontract_SUITE`.

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -1404,6 +1404,7 @@ make_calldata_from_id(Id, Fun, Args, State) ->
 format_aevm_args(?cid(<<N:256>>)) -> N;
 format_aevm_args(?hsh(<<N:256>>)) -> N;
 format_aevm_args(?sig(<<W1:256, W2:256>>)) -> {W1, W2};
+format_aevm_args(?sig(<<W0:8, W1:256, W2:256>>)) -> {W0, W1, W2};
```
UlfNorell: Not right. See comment in `aevm_ae_primops` below.

apps/aevm/src/aevm_ae_primops.erl
hanssv: Yes, it should be `Sig = words_to_bin(65, Sig0)`...

apps/aecontract/test/aecontract_SUITE.erl
hanssv: Don't add this case, use `{bytes, Bin}` instead for the long signatures.

tolbrino: Thanks. Changed.

apps/aevm/src/aevm_ae_primops.erl
tolbrino: Thanks. Fixed.

PR #2640, opened 2019-08-12
PT-166642114 hard fork signalling

jur0:


apps/aecore/src/aec_conductor.erl
```diff
@@ -421,6 +422,14 @@ set_stratum_mode(State) ->
             State#state{stratum_mode = false}
     end.
 
+set_is_fork_configured(State) ->
+    IsForkConfigured =
```
lucafavatella: It may be more readable to call a helper function in `aec_fork_signalling` module here. Though debatable so ok.

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,310 @@
...
+-type header_hash()      :: aec_blocks:block_header_hash().
+
+-type worker()           :: {pid(), reference()}.
+
+-type result()           :: true | false | pending.
+
+-type pointer()          :: header_hash().
+
+-record(state, {fork     :: aec_hard_forks:fork() | undefined,
+                results  :: #{header_hash() := result() | pointer()},
```
lucafavatella: A comment here on which header hashes are meant to be keys in this map would ease readability of intent.

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,350 @@
+%%%============================================================================
+%%% @copyright (C) 2019, Aeternity Anstalt
+%%% @doc
+%%% H1 - height before the the signalling start height;
+%%% HS - signalling start height;
+%%% HE - signalleing end height;
```
lucafavatella:
```suggestion
%%% HE - signalling end height;
```

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,350 @@
...
+%%% (inclusive).
+%%%
+%%% The block at height HE - 1 is the last signalling block, and there must
+%%% be a result computed for it. Once a worker process that computes the
+%%% result is spawned, a 'pending' atom for the given block hash is  stored in
+%%% results map. Once the worker process is done with the computation it will
+%%% send a boolean result and the 'pending' atom is replaced with the boolean
+%%% result.
+%%%
+%%% The blocks with height between HE and FH (inclusive) are treated
```
lucafavatella:
```suggestion
%%% The blocks with height between HE and H2 (inclusive) are treated
```

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,350 @@
...
+%%% differently. Once such block arrives, its previous key hash is looked up
```
lucafavatella: I would expect:
- Identify the block hash of the predecessor key block at height HE - 1.
- If such block hash at height HE - 1 does not have an entry (boolean or pending), spawn worker for it.

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,350 @@
...
+%%% in the results map. If it's present, it means that the previous key hash
+%%% belongs to the block at height HE - 1 (which has result set to 'true',
+%%% 'false' or 'pending'), or the previous key hash has a pointer (key block
+%%% hash) to a block at height HE - 1.
+%%%
+%%% The pointers are used because the result at height HE - 1 can change
+%%% asynchronosly (from 'pending' to a boolean value). Moreover, it's not
+%%% necessary to search a block at HE - 1 height to get the result in case
+%%% a new block arrives. Instead, the results map may have an entry for
+%%% previous key hash pointing to the block with HE - 1 height.
```
lucafavatella: So it is trading space for time 🤔 

May work as anyway key blocks have PoW.

apps/aecore/src/aec_fork_signalling.erl
lucafavatella: I am surprised on FH/H2 included, and I expect such block cannot be inserted unless outcome at HE - 1 known.

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,350 @@
...
+%%%
+%%% The worst case scenario is restarting the node at height H2 - 1. If the
+%%% node has the top key block at height H2 - 1 and the results map is empty,
+%%% it needs to start a worker process the will search for a block at HE - 1,
```
lucafavatella:
```suggestion
%%% it needs to start a worker process that will search for a block at HE - 1,
```

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,350 @@
...
+            State#state{results = Results2};
+        undefined ->
+            %% There is no result reachable via PrevKeyhash. A new process is
+            %% spawned, it will set both a pointer (for BlockHash) and
+            %% a result for a block hash at height HE - 1.
+            spawn_worker(Block, BlockHash, State)
+    end.
+
+worker_process(Parent, Block, BlockHash, Fork) ->
+    case search_last_signalling_block(Block, BlockHash, Fork) of
```
lucafavatella: I see this used that performs several distinct small DB transactions. I expect:
- Not one large DB transaction for preventing locking unnecessarily.
- Not dirty as it would not play well with (async-)transaction.

Seems ok.

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,350 @@
...
+            Err
+    end.
+
+search_last_signalling_block(Block, BlockHash, Fork) ->
+    case is_last_signalling_block(Block, Fork) of
+        true ->
+            {ok, Block, BlockHash};
+        false ->
+            PrevKeyHash = aec_blocks:prev_key_hash(Block),
+            case get_key_block(PrevKeyHash) of
```
lucafavatella: Any reasons why not only the header?

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,350 @@
...
+        {error, _Rsn} = Err -> Err
+    end.
+
+compute_signalling_result(Block, _BlockHash, Fork, Count) ->
+    case is_first_signalling_block(Block, Fork) of
+        true ->
+            {ok, Count + count_inc(is_matching_info_present(Block, Fork))};
+        false ->
+            PrevKeyHash = aec_blocks:prev_key_hash(Block),
+            case get_key_block(PrevKeyHash) of
```
lucafavatella: Any reasons why not only the header?

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,350 @@
...
+%% API
+
+start_link() ->
+    gen_server:start_link({local, ?SERVER}, ?MODULE, [], []).
+
+get_fork_result(Block, BlockHash) ->
+    gen_server:call(?SERVER, {get_fork_result, Block, BlockHash}).
+
+compute_fork_result(Block, BlockHash) ->
+    gen_server:cast(?SERVER, {compute_fork_result, Block, BlockHash}).
```
lucafavatella: Does this mean that on node from an empty storage, the aec_conductor process floods this aec_fork_signalling  process? 🤔

apps/aecore/src/aec_fork_signalling.erl
jur0: Yes, you're right, I'll fix that.

apps/aecore/src/aec_fork_signalling.erl
jur0: Key block is just a header, right? I can use `aec_chain:get_header`, but I assume it will return the same result.

apps/aecore/src/aec_fork_signalling.erl
jur0: If a block at height between HE and H2 - 1 arrives:
* if there is an entry in the results map for prev_key_hash, no new process is spawned

apps/aecore/src/aec_fork_signalling.erl
lucafavatella: You are correct. So it does not matter.

apps/aecore/src/aec_fork_signalling.erl
lucafavatella: I see the problem as being space vs. time, especially in identifying "predecessor key block at height HE - 1". 🤔

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,350 @@
...
+-define(SERVER, ?MODULE).
+
+-type header_hash()      :: aec_blocks:block_header_hash().
+
+-type worker()           :: {pid(), reference()}.
+
+-type result()           :: true | false | pending.
+
+%% The pointer is a key block hash which belongs to the last signalling block.
+-type pointer()          :: header_hash().
```
lucafavatella: So this is at HE - 1. Thanks for the comment.

apps/aecore/src/aec_fork_signalling.erl
jur0: You're probably right. There should be a case in the conductor which checks the origin of the block, if it's database, then skip. I'll add a TODO in the code.

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,350 @@
...
+    {noreply, State}.
+
+terminate(_Rsn, #state{workers = Workers} = State) ->
+    maps:foldl(fun(W, _BH, S1) -> kill_worker(W, S1) end, State, Workers).
+
+%% Internal functions
+
+handle_get_fork_result(Block, BlockHash,
+                       #state{fork = Fork, results = Results} = State) ->
+    case is_fork_height_block(Block, Fork) of
```
lucafavatella: Does this mean that this is meant to be called with argument the block and block hash of the first key block that may have the new consensus protocol? I would rather expect relevant argument to be its predecessor.

If node needs to build key block candidate at fork height, it does not have the block at that height yet...

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,350 @@
...
+        true ->
+            case get_result(BlockHash, Results) of
+                P when ?IS_POINTER(P) ->
+                    %% The block at height H2 is supposed to have a pointer
```
lucafavatella: Same as above: H2 - 1?

apps/aecore/src/aec_fork_signalling.erl
jur0: Ah, I misread the comment, you meant an empty storage. I will have a look at it later.

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,350 @@
...
+                    %% to the block hash at height HE - 1 with the actual
+                    %% result.
+                    case get_result(P, Results) of
+                        R when ?IS_RESULT(R) ->
+                            {{ok, R}, State};
+                        _Other ->
+                            {{error, not_found}, State}
+                    end;
+                _Other ->
```
lucafavatella: If node is stopped when top is H2-1, when node starts which API of this module is meant to be used to make sure that computation of outcome of signalling is started?

jur0: Good catch, will fix.

apps/aecore/src/aec_fork_signalling.erl
lucafavatella: User for the first time starts node. All PoWs are all computed so blocks get retrieved from network and added to storage quickly. This cast is asynchronous so I am concerned for message queue length. It feels some state of the process may need to go in (publicly readable, privately writable?) ETS table so to perform check quickly in context of conductor.

jur0: The agreed solution is to move the fork config from the `aec_fork_signalling` to `aec_conductor` and make block height vs signalling end/fork height check in the context of the conductor and pass fork config as a param from the conductor.

apps/aecore/src/aec_fork_signalling.erl
jur0: This is resolved as not all the pointers between `HE` and `H2 - 1` are stored anymore.

apps/aecore/src/aec_fork_signalling.erl
jur0: Fixed in 4d496b383

apps/aecore/src/aec_fork_signalling.erl
jur0: Fixed in 4d496b383

apps/aecore/src/aec_fork_signalling.erl
jur0: Fixed in 4d496b383

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,359 @@
...
+
+handle_get_fork_result(Block, BlockHash,
+                       #state{fork = Fork, results = Results} = State) ->
+    case is_last_block_before_fork(Block, Fork) of
+        true ->
+            case get_result(BlockHash, Results) of
+                R when ?IS_RESULT(R) ->
+                    %% The block at height H2 - 1 has a result (not a pointer).
+                    %% It means that the block at height HE - 1 is the same as
+                    %% H2 - 1. There are no blocks between, so HE < H2 doesn't
```
lucafavatella: Shall this clause ever apply then?

I would then expect this code to either:
- Rely on constraint having been enforced by user configuration load, and not have this clause at all.
- Return an error that clearly signals misconfiguration, and then caller code shall log error (blocking, though shall never happen).

apps/aecore/src/aec_fork_signalling.erl
jur0: Fixed in 12b7f6d07.

apps/aecore/src/aec_fork_signalling.erl
jur0: Fixed in 88ca0f4c3.

jur0:
Rebased on master.

apps/aecore/src/aec_fork_signalling.erl
```diff
@@ -0,0 +1,355 @@
...
+            %% situation).
+            case get_result(BlockHash, Results) of
+                P when ?IS_POINTER(P) ->
+                    %% The block at height H2 - 1 is supposed to have a pointer
+                    %% (in the best case scenario) to the block hash at height
+                    %% HE - 1 with the actual result.
+                    case get_result(P, Results) of
+                        R when ?IS_RESULT(R) ->
+                            {{ok, R}, State};
+                        _Other ->
```
lucafavatella: Not clear how this can happen.

jur0: Fixed.

apps/aecore/test/aec_fork_signalling_tests.erl
```diff
@@ -0,0 +1,130 @@
...
+-define(INFO_FIELD_SUPPORT, make_info(1)).
+-define(INFO_FIELD_OTHER, make_info(999)).
+
+-define(VERSION_OLD, 4).
+-define(VERSION_NEW, 5).
+
+%% [0] --- [1] --- [2] --- [3] --- [4] --- [5] --- [6] --- [7] ---
+%%         H1      HS            HE - 1    HE    H2 - 1    H2
+
+-define(FORK_CFG,
```
lucafavatella: Shall this value be passed through the same validation function as used for the user config - as a sanity check that the test data is meaningful? Though it may be considered as breaking the boundaries of the unit test, so debatable, so optional.

apps/aecore/test/aec_fork_signalling_tests.erl
```diff
@@ -0,0 +1,130 @@
...
+basic_chain(BlockCfgs, ExpectedForkResult) ->
+    %% FORK_HEIGHT blocks are needed to be generated (including the genesis),
+    %% so the chain ends up at H2 - 1 height where it's possible to find out
+    %% the fork signalling result.
+    Chain = [B0, B1, B2, B3, B4, B5, B6] =
+        aec_test_utils:gen_blocks_only_chain(?FORK_HEIGHT, BlockCfgs),
+    [BH0, BH1, BH2, BH3, BH4, BH5, BH6] = [block_hash(B) || B <- Chain],
+
+    ok = insert_block(B0),
+    ?TEST_MODULE:compute_fork_result(B0, BH0, ?FORK_CFG),
```
lucafavatella: Shall return value be enhanced to indicate if skipped or not - for testability that no unnecessary message is sent? (For case of node bootstrapping from empty storage.)

apps/aecore/test/aec_fork_signalling_tests.erl
```diff
@@ -0,0 +1,130 @@
...
+    ok = insert_block(B2),
+    ?TEST_MODULE:compute_fork_result(B2, BH2, ?FORK_CFG),
+    {error, not_last_block_before_fork} = ?TEST_MODULE:get_fork_result(B2, BH2, ?FORK_CFG),
+
+    ok = insert_block(B3),
+    ?TEST_MODULE:compute_fork_result(B3, BH3, ?FORK_CFG),
+    {error, not_last_block_before_fork} = ?TEST_MODULE:get_fork_result(B3, BH3, ?FORK_CFG),
+
+    ok = insert_block(B4),
+    ?TEST_MODULE:compute_fork_result(B4, BH4, ?FORK_CFG),
```
lucafavatella: Would it be possible to test that computation is asynchronous?

lucafavatella: Can two workers start concurrently? Maybe some workers can end out of order?

I am mindful that test may introduce artificial synchronization points. Would it be possible to add a test that attempts to cause multiple concurrent workers (I mean subsequent blocks) e.g. by creating large chain with large signalling interval, call `compute_fork_result` for all inserted blocks, check that outcome of computation still makes sense?

apps/aecore/test/aec_fork_signalling_tests.erl
```diff
@@ -0,0 +1,130 @@
...
+
+    ok = insert_block(B5),
+    ?TEST_MODULE:compute_fork_result(B5, BH5, ?FORK_CFG),
+    {error, not_last_block_before_fork} = ?TEST_MODULE:get_fork_result(B5, BH5, ?FORK_CFG),
+
+    ok = insert_block(B6),
+    ?TEST_MODULE:compute_fork_result(B6, BH6, ?FORK_CFG),
+    {ok, ExpectedForkResult} = await_result(B6, BH6, ?FORK_CFG),
+
+    ok.
```
lucafavatella: I fail to see the failure case of node restarted before computation completed. Multiple cases - from HE-1 to H2-1.

apps/aecore/test/aec_fork_signalling_tests.erl
jur0: Yes, it can be done.

jur0:
`Fork` config is in the conductor's state, but it is needed for getting a protocol version in `add_synced_block` and `post_block` that are executed in the caller's context. Consider moving the fork config to application environment (`application:get_env`).

PR #2641, opened 2019-08-12
Pt 167837421 race in sc state deletion

velzevur:
[SC state: race condition](https://www.pivotaltracker.com/story/show/167837421)

Fixes a race condition in State Channel's keeping of state.

When a `aesc_chain_watcher` process is spawned for the first time for a new channel, the channel object is not yet present in state trees, thus the watcher's `chan_vsn` is `undefined`. It is properly updated on subsequent microblocks. Later on when the channel's connection is closed a watcher is started in order to check if the channel is being closed on-chain. 

When the channel is being reestablished the `aesc_chain_watcher` is being restarted with a `chan_vsn` still being `undefined`. If now the connection is closed again, the [close waiting check](https://github.com/aeternity/aeternity/blob/master/apps/aechannel/src/aesc_chain_watcher.erl#L487) is being mislead that the channel is already closed on-chain and thus - both cached and persisted states are being deleted in 4 blocks' time.

This PR fixes the issue with optionally initiating the `chan_vsn` with its proper value at watcher's init time.

### DIFF
--- a/apps/aechannel/test/aesc_fsm_SUITE.erl
+++ b/apps/aechannel/test/aesc_fsm_SUITE.erl
@@ -125,27 +125,26 @@ groups() ->
                              ]},
      {transactions, [sequence],
       [
-     %   create_channel
-     % , multiple_responder_keys_per_port
-     % , channel_insufficent_tokens
-     % , inband_msgs
-     % , upd_transfer
-     % , update_with_conflict
-     % , update_with_soft_reject
-     % , deposit_with_conflict
-     % , deposit_with_soft_reject
-     % , withdraw_with_conflict
-     % , withdraw_with_soft_reject
-     % , upd_dep_with_conflict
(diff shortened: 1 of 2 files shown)


apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -896,36 +896,50 @@ t_leave_reestablish_(Cfg) ->
...
+                  fun() ->
+                          Cache2 = cache_status(ChId),
+                          [] = in_ram(Cache2),
+                          true = on_disk(Cache2)
+                  end),
+            %%
+            %% reestablish
+            %%
+            ChId = maps:get(channel_id, RLocal),
+            mine_key_blocks(dev1, 6), % min depth at 4, so more than 4
```
velzevur: instead of waiting for 500ms, now we mine 6 blocks

PR #2642, opened 2019-08-12
Pt 167060729 allow signing of just tx hash 2

hanssv:
Rebased version of TX signing

### DIFF
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -774,7 +774,7 @@ check_valid_at_protocol(STx, _Hash, Height, _Event) ->
 
 check_signature(Tx, Hash, _Height, _Event) ->
     {ok, Trees} = aec_chain:get_top_state(),
-    case aetx_sign:verify(Tx, Trees) of
+    case aetx_sign:verify(Tx, Trees, top_height()) of
         {error, _} = E ->
             lager:info("Failed signature check on tx: ~p, ~p\n", [E, Hash]),
             E;
--- a/apps/aehttp/test/aehttp_sc_SUITE.erl
+++ b/apps/aehttp/test/aehttp_sc_SUITE.erl
@@ -730,7 +730,7 @@ channel_update(#{initiator := IConnPid, responder :=RConnPid},
 
     %% validate it is co-signed
     {ok, Trees} = rpc(aec_chain, get_top_state, []),
-    ok = rpc(aetx_sign, verify, [SignedStateTx, Trees]), % RPC because of DB
+    ok = rpc(aetx_sign, verify, [SignedStateTx, Trees, 1]), % RPC because of DB
     {UnsignedStateTx, _} = % same transaction that was signed
         {aetx_sign:innermost_tx(SignedStateTx), UnsignedStateTx},
 
@@ -3105,7 +3105,7 @@ wait_for_json_rpc_action_event(ConnPid, Action, Event) ->
                                                                  Action,
                                                                  Event)
                               end).
-    
+
 wait_for_json_rpc_action(ConnPid, Action) ->
     wait_for_json_rpc_action_(Action,
                               fun() ->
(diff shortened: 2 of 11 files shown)


apps/aehttp/test/aehttp_sc_SUITE.erl
```diff
@@ -730,7 +730,7 @@ channel_update(#{initiator := IConnPid, responder :=RConnPid},
 
     %% validate it is co-signed
     {ok, Trees} = rpc(aec_chain, get_top_state, []),
-    ok = rpc(aetx_sign, verify, [SignedStateTx, Trees]), % RPC because of DB
+    ok = rpc(aetx_sign, verify, [SignedStateTx, Trees, 1]), % RPC because of DB
```
ThomasArts: is 1 a valid environment? Yes, verify/3 changed to take height as 3rd argument. Ok.

lucafavatella:
I see protocol PR is https://github.com/aeternity/protocol/pull/372

apps/aetx/src/aetx_sign.erl
```diff
@@ -104,71 +105,85 @@ from_db_format(#signed_tx{tx = Tx} = STx) ->
...
+        error          -> {error, signature_check_failed}
+    end;
+verify_signatures(PubKeys,_Bin, Sigs, _Height) ->
     lager:debug("Signature check failed: ~p ~p", [PubKeys, Sigs]),
     {error, signature_check_failed}.
 
-verify_one_pubkey(Sigs, PubKey, Bin) ->
-    verify_one_pubkey(Sigs, PubKey, Bin, []).
+verify_one_pubkey(Sigs, PubKey, Bin, Height) when ?VALID_PUBK(PubKey) ->
+    Protocol = aec_hard_forks:protocol_effective_at_height(Height),
```
lucafavatella: I would have preferred awareness to protocol - rather than height so deep in call - on the basis that support for miner signalling consensus upgrade may remove assumption that consensus is known based on height. Though ok.

apps/aecore/src/aec_tx_pool.erl
```diff
@@ -774,7 +774,7 @@ check_valid_at_protocol(STx, _Hash, Height, _Event) ->
 
 check_signature(Tx, Hash, _Height, _Event) ->
     {ok, Trees} = aec_chain:get_top_state(),
-    case aetx_sign:verify(Tx, Trees) of
+    case aetx_sign:verify(Tx, Trees, top_height()) of
```
lucafavatella:
```suggestion
    case aetx_sign:verify(Tx, Trees, Height) of
```

apps/aecore/src/aec_tx_pool.erl
```diff
@@ -774,7 +774,7 @@ check_valid_at_protocol(STx, _Hash, Height, _Event) ->
 
 check_signature(Tx, Hash, _Height, _Event) ->
```
lucafavatella:
```suggestion
check_signature(Tx, Hash, Height, _Event) ->
```

apps/aetx/src/aetx_sign.erl
hanssv: This was done a bit before we started discussing the signalling, sorry...

apps/aetx/src/aetx_sign.erl
```diff
@@ -104,71 +105,85 @@ from_db_format(#signed_tx{tx = Tx} = STx) ->
...
     case aetx:signers(Tx, Trees) of
         {ok, Signers} ->
-            verify_signatures(Signers, Bin, Sigs);
+            verify_signatures(Signers, Bin, Sigs, Height);
         {error, _Reason} ->
             {error, signature_check_failed}
     end.
 
 %% this function is strict and does not allow having more signatures that the
 %% one being checked
```
lucafavatella: Dangling comment?

hanssv: Probably from one of the rebases yes...

PR #2647, opened 2019-08-13
Pt 167781061 snapshot solo in ws api

uwiger:
See [PT #167781061](https://www.pivotaltracker.com/story/show/167781061), adding support for requesting a `snapshot_solo_tx` via the WS API.

A test case has been added to `aehttp_sc_SUITE`. It verifies that the `"channels.snapshot_solo"` request returns an error if the current state is not an offchain tx or if the latest round is already reflected on-chain.

A min-depth watch is set locally on the reuqester side, but doesn't block the FSM. An info report with `"event": "min_depth_achieved"` is generated when minimum depth has been reached.

Documentation is being worked on.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -28,7 +28,8 @@
 %% API
 -export([ start_link/1
         , attach_responder/2      %% (fsm(), map())
-        , close_solo/1
+        , close_solo/1            %% (fsm())
+        , snapshot_solo/1         %% (fsm())
         , reconnect_client/2      %% (fsm(), signed_tx())
         , reconnect_client/3      %% (fsm(), Client :: pid(), signed_tx())
         , connection_died/1       %% (fsm())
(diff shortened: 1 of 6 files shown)


uwiger:
@grzegorz225 the test updated (which triggered a bunch of refactoring in the sc_SUITE). The conflict scenario is not tested. What can be expected to happen in that case - the worst case - is that both FSMs proceed after an initial check, then one snapshot fails in processing. Thus, it won't make it onto the chain. The fsm doesn't have any supervision on that, so the client would have to detect that it never gets a corresponding `on_chain_tx` event.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1541,6 +1571,30 @@ new_onchain_tx(channel_close_solo_tx, Opts,
...
+                    Payload = aetx_sign:serialize_to_binary(SignedTx),
+                    Opts1 =
+                        Opts#{ channel_id => aeser_id:create(channel, ChanId)
+                             , from_id    => aeser_id:create(account, Account)
+                             , payload    => Payload
+                             , ttl        => TTL },
+                    {ok, Tx} = new_onchain_tx_(aesc_snapshot_solo_tx, Opts1, CurrHeight),
+                    {ok, Tx, []};
+                false ->
+                    {error, already_onchain}
```
grzegorz225: Shouldn't we pass this error to the client?

uwiger: I think we do, although there is no error code attached to this particular error reason.

uwiger: Error code and description added, and test case enhanced to verify error returns.

PR #2653, opened 2019-08-14
Delete unused tag regex

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/167350192

Idea is to use this for tag 5.0.0-rc.1


.circleci/config.yml
```diff
@@ -106,7 +106,6 @@ references:
           && ./configure && make -j$(nproc) && sudo make install && sudo ldconfig
 
   tag_regex: &tag_regex /^v.*$/
-  stable_tag_regex: &stable_tag_regex /^v4\..*$/
```
lucafavatella: This regex will cause issues for v5.

PR #2656, opened 2019-08-14
Sync review

lucafavatella:
Delivers https://www.pivotaltracker.com/story/show/167597656


lucafavatella:
I added pasted key `mainnet_peers` in `config/dev.config`, I started node with `make local-build console`, I observed node progressing in syncing ok until approx. 12500 and no `bad` messages in the sync log.

PR #2661, opened 2019-08-16
Cache tx env for better SC FSM performance

uwiger:
See [PT #167938013](https://www.pivotaltracker.com/story/show/167938013). Caching the tx_env data turns out to have a significant impact on SC FSM performance.

(This PR was based on PR #2659, for better precision evaluating benchmark results.)

The aetx_env data is cached in an env cache ets table introduced in the `aesc_state_cache` process. The state cache subscribes to `top_changed` events and refreshes the top block tx_env, but only if it has cached offchain states, indicating that there are actually running FSMs.

The cost of an individual tx_env lookup went down from `4-6 ms` to `5 us`, and performance in the `multiple_channels` benchmark test increased from ca `77 tps` to `105 tps`.

### DIFF
--- a/apps/aechannel/src/aesc_tx_env_cache.erl
+++ b/apps/aechannel/src/aesc_tx_env_cache.erl
@@ -1,0 +1,63 @@
+-module(aesc_tx_env_cache).
+-behavior(gen_server).
+
+-export([tx_env_and_trees_from_top/1]).
+
+-export([start_link/0]).
+
+%% Behavior exports
+-export([ init/1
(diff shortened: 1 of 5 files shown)


apps/aechannel/src/aesc_tx_env_cache.erl
```diff
@@ -0,0 +1,63 @@
+-module(aesc_tx_env_cache).
```
tolbrino: Some documentation on the purpose of the module would be nice.

PR #2664, opened 2019-08-17
Do not allow #meta{} before Lima hardfork

velzevur:
Current implementation allows both versions of off-chain updates with and without a list of off-chain updates. Adding a new update type allows for it to be smuggled on-chain in a off-chain transaction that is provided for a payload for force progress, solo close or slash transactions.

This PR makes it protocol aware so a `#meta{}` doesn't make it on-chain before Lima.

### DIFF
--- a/apps/aechannel/src/aesc_offchain_tx.erl
+++ b/apps/aechannel/src/aesc_offchain_tx.erl
@@ -242,7 +242,12 @@ version(#channel_offchain_tx{updates = Updates}) ->
 -spec valid_at_protocol(aec_hard_forks:protocol_vsn(), tx()) -> boolean().
 valid_at_protocol(Protocol, Tx) ->
     case version(Tx) of
-        ?INITIAL_VSN -> true;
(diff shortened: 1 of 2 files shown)


apps/aechannel/src/aesc_offchain_tx.erl
```diff
@@ -242,7 +242,13 @@ version(#channel_offchain_tx{updates = Updates}) ->
 -spec valid_at_protocol(aec_hard_forks:protocol_vsn(), tx()) -> boolean().
 valid_at_protocol(Protocol, Tx) ->
     case version(Tx) of
-        ?INITIAL_VSN -> true;
+        ?INITIAL_VSN when Protocol >= ?LIMA_PROTOCOL_VSN -> false;
```
velzevur: This line is not essentially part of this PT but while we are here, we can also forbid old style off-chain transactions making it on-chain. This will allow us not to worry of off-chain updates making it on-chain ever again. Post Lima that is :)

PR #2665, opened 2019-08-17
Fix IF race in tests

velzevur:
Fix IF that happens once in a blue moon. Now even less often.


velzevur:
I see this is a duplicate of #2663

Closing this PR

PR #2667, opened 2019-08-19
disable 'test' registrar and enable 'aet' registrar from LIMA onwards

skkw:


apps/aens/src/aens_utils.erl
```diff
@@ -59,13 +60,13 @@ check_claimed_status(Name) ->
...
 validate_name(Name) ->
-    case binary:split(Name, ?LABEL_SEPARATOR, [global, trim]) of
-        [_Label, RegistrarNS] ->
-            case [RN || RN <- aec_governance:name_registrars(), RN =:= RegistrarNS] of
-                [] -> {error, registrar_unknown};
-                _  -> ok
-            end;
+    case name_parts(Name) of
+        [_Label, _RegistrarNS] ->
+            ok;
```
ThomasArts: Why do we no longer need to validate the name?

apps/aens/test/aens_utils_test.erl
```diff
@@ -13,15 +13,13 @@ to_ascii_test() ->
     ?assertEqual({error, no_registrar},
                  aens_utils:to_ascii(<<"abcdefg">>)),
     ?assertEqual({error, multiple_namespaces},
-                 aens_utils:to_ascii(<<"abcd.efgh.test">>)),
```
ThomasArts: I guess, <<"bacd.efgh-test">> test should also still fail with this error, but is no longer tested for...
I expected tests to be added, not so much to be removed

apps/aens/test/aens_utils_test.erl
```diff
@@ -13,15 +13,13 @@ to_ascii_test() ->
...
-    ?assertEqual({error, registrar_unknown},
-                 aens_utils:to_ascii(<<"abcd.aettt">>)),
+                 aens_utils:to_ascii(<<"abcd.efgh.aet">>)),
     LongName = list_to_binary(string:lowercase(base58:binary_to_base58(crypto:strong_rand_bytes(1000)))),
     ?assertEqual({error, no_label_in_registrar},
-                 aens_utils:to_ascii(<<".test">>)),
+                 aens_utils:to_ascii(<<".aet">>)),
```
ThomasArts: same here... before Lima, this should be failing differently then after Lima, but that seems not to be tested

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -2317,9 +2317,6 @@ nameservice_transaction_claim(MinerAddress, MinerPubkey) ->
     Missing = aeser_api_encoder:encode(name, <<"missing">>),
     {ok, 400, #{<<"reason">> := <<"Name validation failed with a reason: no_registrar">>}} =
         get_name_claim(maps:put(name, Missing, Encoded)),
-    MissingReg = aeser_api_encoder:encode(name, <<"missing.reg">>),
```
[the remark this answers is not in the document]
skkw: The reason is that now the valid registrars are height dependent so we know it only at aeprimop level.
Submitting TX with invalid registrars now succeeds only to be rejected later.

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -3148,10 +3145,6 @@ naming_system_broken_txs(_Config) ->
     {ok, []} = rpc(aec_tx_pool, peek, [infinity]),
 
     %% Try to submit txs with empty account
-    {ok, 400, #{<<"reason">> := <<"Name validation failed with a reason: registrar_unknown">>}} =
```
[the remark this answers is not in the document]
skkw: Was removed because of same reasons as above - we don't have static list of valid registrars anymore, and can't do this validation at TX creation time.

apps/aens/src/aens_utils.erl
skkw: We do, but only validate what is possible to check without additional height parameter.

The registrar check which was removed here is now validated here:  https://github.com/aeternity/aeternity/blob/PT-167889960-disallow-claiming-test-registrar/apps/aeprimop/src/aeprimop.erl#L1619

apps/aens/test/aens_utils_test.erl
skkw: The diff is a bit confusing here, this test is still present on the branch:

```    
?assertEqual({error, multiple_namespaces},
            aens_utils:to_ascii(<<"abcd.efgh.aet">>)),
```

We can't check registrar of `bacd.efgh-test` thought, for the same reason as mention above.

apps/aehttp/test/aehttp_integration_SUITE.erl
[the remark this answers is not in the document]
ThomasArts: Ok. This then means that we cannot even create such a transaction claiming a name with a different registery. But others can, however, these transactions will fail (due to process transaction) on the new node and will fail due to invalid regestry on other nodes. So both old and new cannot make a transaction name claim that ends up on chain with wrong registry. ok.

apps/aens/src/aens_utils.erl
ThomasArts: It would not be a lot of work to actually here check the registrats from all possible protocol heights.... could even make that a aec_governance function. Then it is super easy to be backward compatible and have an extra check in process transaction

apps/aens/test/aens_SUITE.erl
```diff
@@ -560,3 +562,35 @@ do_prune_until(N1, N1, OTree) ->
     aens_state_tree:prune(N1, OTree);
 do_prune_until(N1, N2, OTree) ->
     do_prune_until(N1 + 1, N2, aens_state_tree:prune(N1, OTree)).
+
+%%%===================================================================
+%%% Change of registrar
+%%%===================================================================
+
+registrar_change(_Cfg) ->
```
ThomasArts: good!

ks:
Duplicate of #2704

PR #2668, opened 2019-08-19
PT-167970620 - Add GA Attach/Meta Tx definition in swagger yaml/json

noandrea:


dincho:
I triggered manually the CI build by pushing copy of this branch https://github.com/aeternity/aeternity/tree/167970620_ga_swagger

@noandrea you should use `make swagger` to build the swagger.json otherwise CI checks will fail as mismatch

PR #2670, opened 2019-08-19
[PT-167778774] Decrease test runtime

grzegorz225:
This PR:
* Implements a mocking interface for tests to mock remotely parts of a node.
* Implements mocks for block mining.
* Cache contract compilations/calls - the cache is persisted after a test run and loaded on the next run (Keep in mind that not having the persisted cache increases the test run time only by ~30-60s).
* Decreases some timeouts in the FSM tests.
* Decreases the reconnection interval in the test environment.
* Discards ct:logs about new keyblocks as the mining rate with mocking enabled became insanely high.

This pull request decreases the test time significantly:
* For ct-roma inside CI the tests inside a single container now take 8-9 min instead of 17-18 min.
* For ct-lima inside CI the tests inside a single container now take 11-12 min instead of 18-21 min.

Additionally:
* Eliminates random test failures with for example {error, max_reached} - This was caused by a race condition between block mining and microblock generation which I've noticed when I've decreased the mining interval to 50ms :)
* For the aesc_tx suite I've observed a decrease in runtime from >500s to <70s.
* For the aesc_fsm suite I've observed a significant increase in performance of some operations by changing the noise socket connect pooling interval - pretty often the first connection attempt between SC peers always fails - this means that on most connections we needed to wait 3s for a SC to be established - I've changed the interval in the test environment to 50ms - this was uncovered when this became the main contributor to the runtime of the aesc_fsm suite - somehow decreasing this interval to 10ms also works - apparently only the first connection attempt fails.

Right now I haven't touched aec_* suites but on first look there are some tests like aec_txs suite which could be speed up by using my mocks. I'm including a contract call encode/decode cache as it is movable between computers and is pretty small - keep in mind that this cache must be updated after an aesophia version bump - some work would need to be done for the compilation cache to also be movable as currently the compiler outputs hardcoded paths in the compilation results. I'm unsure whether it is a good idea to have this cache kept in the repo - Ideally it should be cached in CI but not having the cache in the first place only increases the runtime by ~30-60s. Keeping it in github on the other hand could be useful for people who run contract/ga tests pretty often.


apps/aecontract/test/aect_cache_restorer.erl
```diff
@@ -0,0 +1,90 @@
...
+
+dets_cache_path(CacheDir, DepsHash, ETSTable) ->
+    filename:absname_join(CacheDir, dets_cache_filename(DepsHash, ETSTable)).
+
+try_load_cache(CacheDir, DepsHash, ETSTable) ->
+    Path = dets_cache_path(CacheDir, DepsHash, ETSTable),
+    case dets:open_file(Path) of
+        {ok, Handle} ->
+            try
+                ETSTable = dets:to_ets(Handle, ETSTable)
```
uwiger: Personally, I prefer `ets:file2tab/2`, which uses disk_log instead of ets. I believe that this should be more robust, but it's not a major issue.

UlfNorell:
@grzegorz225 @uwiger @velzevur 

The cache as currently implemented does not work properly. It's caching based on dependecies and *file name*, which means that it doesn't notice if you change the code of a test contract. We could change this to memoize on the contract source code, but I would much prefer if we got rid of the persistent cache completely. If it only saves 30-60s per test run it's not worth the trouble.

PR #2672, opened 2019-08-20
Fix some fate bugs

UlfNorell:
Doing some QuickCheck testing I found some bugs in FATE.

### DIFF
--- a/apps/aecontract/src/aect_contracts.erl
+++ b/apps/aecontract/src/aect_contracts.erl
@@ -513,7 +513,6 @@ assert_field_store(store_k = Field, X, #{vm := VM} = Version) when is_binary(X),
     catch _:_ -> error({illegal, Field, X}) end;
 assert_field_store(store_k, X, #{vm := VM}) when is_binary(X),
                                                  byte_size(X) > 0,
-                                                 byte_size(X) =< 32,
                                                  ?IS_FATE_SOPHIA(VM) ->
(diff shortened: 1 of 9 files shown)


apps/aefate/src/aefa_stores.erl
```diff
@@ -115,7 +115,7 @@ find_value_(Pubkey, StorePos, #store{cache = Cache} = S) ->
         {ok, Term} ->
             {ok, Term};
         {ok, Term, Entry} ->
-            {ok, Term, S#store{cache = #{Pubkey => Entry}}};
+            {ok, Term, S#store{cache = Cache#{Pubkey => Entry}}};
```
hanssv: 🙀

apps/aecontract/src/aect_contracts.erl
```diff
@@ -513,7 +513,6 @@ assert_field_store(store_k = Field, X, #{vm := VM} = Version) when is_binary(X),
     catch _:_ -> error({illegal, Field, X}) end;
 assert_field_store(store_k, X, #{vm := VM}) when is_binary(X),
                                                  byte_size(X) > 0,
-                                                 byte_size(X) =< 32,
```
gorillainduction: I just checked https://github.com/aeternity/protocol/blob/master/serializations.md#contract and the upper size is indeed not defined.

PR #2676, opened 2019-08-21
PT-167996886 A proper ecverify

hanssv:
Needs https://github.com/aeternity/aebytecode/pull/70
and https://github.com/aeternity/aesophia/pull/128

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -292,7 +292,7 @@ smoke-test: docker smoke-test-run
 
 smoke-test-run: KIND=system_test
 smoke-test-run: internal-build
-	@$(REBAR) as $(KIND) do ct $(ST_CT_DIR) $(ST_CT_FLAGS) --suite=aest_sync_SUITE,aest_commands_SUITE,aest_peers_SUITE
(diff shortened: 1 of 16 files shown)


hanssv:
Docs is currently being written.

Makefile
```diff
@@ -292,7 +292,7 @@ smoke-test: docker smoke-test-run
 
 smoke-test-run: KIND=system_test
 smoke-test-run: internal-build
-	@$(REBAR) as $(KIND) do ct $(ST_CT_DIR) $(ST_CT_FLAGS) --suite=aest_sync_SUITE,aest_commands_SUITE,aest_peers_SUITE
+	@$(REBAR) as $(KIND) do upgrade, ct $(ST_CT_DIR) $(ST_CT_FLAGS) --suite=aest_sync_SUITE,aest_commands_SUITE,aest_peers_SUITE
```
dincho: Isn't this `upgrade` workaround the lock file?

tolbrino: The lock file is only used by the deps in the `default` profile. All additional deps defined in other profiles are not locked. Therefore, this upgrade ensure that they are updated to the defined reference.

dincho: Interesting tool :)

hanssv: @dincho there is no lock file for `system_test`?

hanssv: And except that the update does not seem to help... Interesting tool indeed :-)

PR #2678, opened 2019-08-21
PT-168025582 Relax protocol version checks in block serialization

jur0:
[PT-168025582](https://www.pivotaltracker.com/n/projects/2124891/stories/168025582)

`deserialize_[key|micro]_from_binary` is called in `aec_blocks:deserialize_from_binary` which is used in:
* ~`aeu_import:disklog_fold/3` - not used anywhere?!~ - module was deleted
* `aec_peer_connection:deserialize_block/2` - when `aec_peer_connection` wants to insert a new block to the db, it calls `aec_conductor:post_block/1` where the block is validated (including the version check).

`serialization_template` called in `aec_blocks:serialize_to_binary` used `aec_hard_forks:protocol_effective_at_height` which is not needed because `aec_blocks:serialize_to_binary` is called for blocks read from the db, so they were validated for the right protocol version when inserted into the db.

### DIFF
--- a/apps/aecore/src/aec_blocks.erl
+++ b/apps/aecore/src/aec_blocks.erl
@@ -333,6 +333,9 @@ update_micro_candidate(#mic_block{} = Block, TxsRootHash, RootHash, Txs) ->
 %%% Serialization
 %%%===================================================================
 
+%% Serialization assumes the protocol version in the header to be valid for the
+%% provided height. This should have been validated before calling this
+%% function.
 -spec serialize_to_binary(block()) -> binary().
 serialize_to_binary(#key_block{} = Block) ->
     aec_headers:serialize_to_binary(to_key_header(Block));
--- a/apps/aecore/src/aec_blocks.erl
+++ b/apps/aecore/src/aec_blocks.erl
@@ -337,24 +337,14 @@ update_micro_candidate(#mic_block{} = Block, TxsRootHash, RootHash, Txs) ->
 serialize_to_binary(#key_block{} = Block) ->
     aec_headers:serialize_to_binary(to_key_header(Block));
 serialize_to_binary(#mic_block{} = Block) ->
-    Hdr    = to_micro_header(Block),
+    Hdr = to_micro_header(Block),
     HdrBin = aec_headers:serialize_to_binary(Hdr),
-    Height = aec_headers:height(Hdr),
-    Vsn    = version(Block),
-    case serialization_template(micro, Height, Vsn) of
-        {error, What} ->
-            error({serialization_error, What});
-        {ok, Template} ->
-            Txs = [ aetx_sign:serialize_to_binary(Tx) || Tx <- txs(Block)],
-            Rest = aeser_chain_objects:serialize(
-                     micro_block,
-                     Vsn,
-                     Template,
-                     [ {txs, Txs}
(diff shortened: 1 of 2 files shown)


apps/aecore/src/aec_headers.erl
```diff
@@ -581,26 +581,21 @@ deserialize_key_from_binary(<<Version:32,
                               Time:64,
                               Info/binary
                             >>) ->
-    case aec_hard_forks:protocol_effective_at_height(Height) =:= Version of
```
ThomasArts: I feel uncomfortable by just removing these version checks... I don't see immediately how this can be exploited, but it needs a lot of reasoning to conclude that nobody in the future will add a block on Lima height with Minerva version...

apps/aecore/src/aec_blocks.erl
```diff
@@ -337,24 +337,14 @@ update_micro_candidate(#mic_block{} = Block, TxsRootHash, RootHash, Txs) ->
 serialize_to_binary(#key_block{} = Block) ->
     aec_headers:serialize_to_binary(to_key_header(Block));
 serialize_to_binary(#mic_block{} = Block) ->
```
ThomasArts: For avoiding future misuse of this function. Please add a comment:
```
%% Serialization assumes the protocol version in the header to be valid for the provided height. This should have been validated before calling this function.
```

apps/aecore/src/aec_headers.erl
```diff
@@ -581,26 +581,21 @@ deserialize_key_from_binary(<<Version:32,
...
-                            target = Target,
-                            pow_evidence = PowEvidence,
-                            nonce = Nonce,
-                            time = Time,
-                            version = Version,
-                            info = Info
-                           },
-            {ok, H}
-    end;
+    PowEvidence = deserialize_pow_evidence_from_binary(PowEvidenceBin),
```
ThomasArts: Same kind of warning in comment here would be in place

apps/aecore/src/aec_headers.erl
```diff
@@ -562,6 +562,10 @@ deserialize_from_binary_partial(<<_Version:32,
             {error, malformed_header}
     end.
 
+%% The function does not check the validity of the protocol version based on
+%% height. It gets the protocol version from the block header. The protocol
+%% version check based on height is performed before inserting it into the
+%% database (aec_conductor).
```
ThomasArts: :thumbsup:

apps/aecore/src/aec_blocks.erl
```diff
@@ -368,27 +361,16 @@ deserialize_from_binary(Bin) ->
...
-            [{txs, Txs0}, {pof, PoF0}] =
-                aeser_chain_objects:deserialize(micro_block, Vsn, Template, Bin),
-            Txs = [aetx_sign:deserialize_from_binary(Tx)
-                   || Tx <- Txs0],
-            PoF = aec_pof:deserialize(PoF0),
-            {ok, #mic_block{header = Header, txs = Txs, pof = PoF}};
-        Err = {error, _} ->
-            Err
-    end.
+    Version = aec_headers:version(Header),
```
gorillainduction: The validation height/Vsn is lost here. Perhaps add the comment here as well.

Are you sure this is validated elsewhere, and that we do not rely on these checks?

I am more worried about the deserialization than the serialization.

PR #2680, opened 2019-08-21
Improve build times when testing

tolbrino:
By reducing the number of invocation of `rebar` when trying to run tests, the overhead of rebar is reduced before tests can start.


tolbrino:
The benefit doesn't seem to be useful anymore so I am closing this PR as wontfix, but will keep the branch around for some time.

PR #2683, opened 2019-08-21
Assert that a private function cannot be called remotely

gorillainduction:
https://www.pivotaltracker.com/story/show/168025996

### DIFF
--- a/apps/aefate/src/aefa_fate.erl
+++ b/apps/aefate/src/aefa_fate.erl
@@ -34,7 +34,7 @@
         , runtime_exit/2
         , runtime_revert/2
         , set_local_function/2
-        , set_remote_function/4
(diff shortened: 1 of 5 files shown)


apps/aefate/src/aefa_fate.erl
```diff
@@ -309,7 +311,7 @@ check_remote(Contract, EngineState) ->
             abort(reentrant_call, EngineState)
     end.
 
-set_remote_function(?FATE_CONTRACT(Pubkey), Function, CheckPayable, ES) ->
+set_remote_function(?FATE_CONTRACT(Pubkey), Function, CheckPayable, CheckPrivate, ES) ->
```
hanssv: The next person that adds a check will have to make a list of them :-)

PR #2684, opened 2019-08-21
restrict name pointers from lima onwards

skkw:


skkw:
We should address the ambiguity by allowing duplicate names of pointers somehow. Keeping this branch for archival, as it has most of the lifting implemented.

PR #2689, opened 2019-08-22
Drop rename db

dincho:
During my work on https://github.com/aeternity/aeternity/pull/2655 I found some magics in the system-tests related to rename_db command.

This migration command has been created in v2.2.0 (Minerva) to migrate the database node name from `epoch` to `aeternity`. I realized that it shouldn't be needed anymore 2 major release later approaching 3th major release (Lima 5.0.0-rc1) so it seems to me the right time to remove it and it's legacy baggage. Supporting Roma databases in Lima releases does not make any sense to me.

During the cleanup I also realized that there a few system tests that are not testing compatibility based on the *local* code but using remote (already published and immutable) docker images (create + reuse), that does not make any sense as once passed this test cannot fail anymore as the docker images won't change.


dincho:
> There are references to the rename script in rebar.config
> Also aec_db references this script:

This has been addressed already, are you sure you're looking all the commits ?!

grzegorz225:
> > There are references to the rename script in rebar.config
> > Also aec_db references this script:
> 
> This has been addressed already, are you sure you're looking all the commits ?!

Oh I overlooked those changes.

dincho:
> A mention that the tool was removed might be necessary as the tool was mentioned in the release notes before:
> https://github.com/aeternity/aeternity/blob/8a43124386bcea4fe08ee10f541529dc062964c0/docs/release-notes/RELEASE-NOTES-2.2.0.md

I forgot about that see fbb74846

PR #2692, opened 2019-08-22
CircleCI windows builds

venimus:
### Changes:
* Migrate from Appveyor to CircleCI
* New install and build scripts (usable locally)
* Removed Java requirements
* Docs for windows builds are updated

### Detailed notes:

Migrate from Appveyor to CircleCI:
- Workflow for windows builds with OTP21.3. Version is configurable. In general it should not differ from Appveyor.
- Cache (layered) of deps folders. Force cache rebuild by bumping `&windows_cache_version`
- Newest windows orbs already contains VS2019.
- Unfortunately CI jobs fail (with a timeout error) when a background process is left from an installer or build. So some stuff was refactored to avoid such leftovers (i.e. manually download and control installers and test script fixes).
- Tagging of new Aeternity versions no longer require to change package.cfg. It is now automatically resolved.
- Windows jobs are controlled by parameters, so they can be reused without much changes to the config. Adding different ENV builds (i.e. different Erlang version) is just by adding another set of aliases for `*windows_build` and `*windows_test` jobs and use them in `jobs:` with the proper parameters. Unfortunately using aliases is the only way (currently) to reuse jobs.
- CI build scripts are now under .circleci/windows (ci/appveyor are kept for BC but will be removed eventually). They rely on the scripts designed for users.

Building scripts:
- Scripts are refactored and fixed to be more user-friendly and keep CI-friendly at the same time. This will increase consistency among build environments. Thus CI `build.bat` script could be used locally to build the artifacts (although it is not meant for that).
- ENV preparation script (`msys2_prepare.bat`) is refactored and split to several standalone scripts that could be used by users (or CI)
- MSYS scripts could be used to tune and handle different installation folders much easier (and faster). It also have help (--help) and debug (-v) mode used by CI.
- Erlang/OTP script now downloads from a much faster mirror (by default). Downloading from the official site is incredibly slow. Script now allows installation of different versions (multiple as well) by providing params or ENV vars. So it is suitable for both users or CI.
- Certain ENV variables could be used to control the build, most of them are optional and used for speeding up the builds: i.e. some of them represent different paths and if undefined, the scripts will try to resolve automatically, which takes some time. Those are set in CircleCI config.
- Docs for windows builds are improved. Although nothing much changed from the user point of view.

Too many changes were needed for this migration, because old scripts were not flexible enough to be fixed to work with new env. That is why I decided to improve the user tools and base the CI builds on them. In fact now CI would mostly mimic how a user could build a local release.
*Note: Windows shell scripting is quite messy and certain basic tasks need some hacks.*
*Note: msys2_prepare seems to be copied from erlang docs, but originally it is designed for building erlang itself. So there is no reason to keep it in sync with it, furthermore the origin is badly written anyway. However I tried to not change it much and I think there are no any breaking changes in this PR.*
Originally branched v4.2.0 and then rebased to master (both work ok)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -412,13 +412,10 @@ references:
     OTP_VERSION: << parameters.otp_version >>
     WIN_OTP_PATH: << parameters.otp_install_path >>
     WIN_MSYS2_ROOT: << parameters.msys_install_path >>
-    JAVA_VERSION: 12.0.2
-    WIN_JDK_BASEPATH: C:\Program Files\OpenJDK
-    JAVA_TOP: /c/Program Files/OpenJDK/jdk-12.0.2
     MSYSCON: defterm
     ANSICON: true
     MSVC_VERSION: 14.22.27905
-    VCVARSALL: C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvarsall.bat
+    VCVARSALL: C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvarsall
 
   windows_container: &windows_container
     executor:
@@ -432,7 +429,7 @@ references:
         default: C:\tools\erl
         type: string
       otp_version:
-        default: "20.1"
+        default: "21.3"
         type: string
     working_directory: *windows_working_directory
     environment: *windows_env
@@ -505,7 +502,7 @@ commands:
       - restore_windows_package_cache
       - run:
           name: Install MSYS2
-          command: .circleci\windows\msys2_prepare.bat
+          command: .circleci\windows\msys2_prepare
       - save_windows_package_cache:
           install_path: << parameters.install_path >>
 
@@ -1096,16 +1093,17 @@ jobs:
   windows_package:
     <<: *windows_container
     steps:
+      - run: systeminfo
       - windows_install_deps:
           otp_install_path: << parameters.otp_install_path >>
           otp_version: << parameters.otp_version >>
           msys_install_path: << parameters.msys_install_path >>
       - run:
-          name: Build
-          command: .circleci\windows\build.bat
+          name: Build packages
+          command: .circleci\windows\build
       - run:
           name: Smoke test
-          command: .circleci\windows\test.bat
+          command: .circleci\windows\test
       - store_artifacts:
           path: *windows_artifacts
           destination: /tmp/packages-<< parameters.otp_version >>
@@ -1118,7 +1116,7 @@ workflows:
   build_test_deploy:
     jobs:
       - windows_package:
-          otp_version: "20.3"
+          otp_version: "21.3"
           context: ae-node-builds
           filters:
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -404,7 +404,7 @@ references:
     - *fail_notification
 
   # Windows
-  windows_cache_version: &windows_cache_version "v11"
+  windows_cache_version: &windows_cache_version "v12"
   windows_artifacts: &windows_artifacts C:\Users\circleci\project\packages
   windows_working_directory: &windows_working_directory C:\Users\circleci\project
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,8 @@
 version: 2.1
+
+orbs:
+  win: circleci/windows@1.0.0
+
 executors:
   infrastructure_container_unstable:
     docker:
@@ -399,6 +403,44 @@ references:
           envdir /secrets aws s3 cp --acl public-read ${PACKAGE_TARBALL:?} s3://${S3_RELEASES_BUCKET:?}/aeternity-latest-${PKG_SUFFIX:?}.tar.gz
     - *fail_notification
 
+  # Windows
+  windows_cache_version: &windows_cache_version "v11"
+  windows_artifacts: &windows_artifacts C:\Users\circleci\project\packages
+  windows_working_directory: &windows_working_directory C:\Users\circleci\project
+
+  windows_env: &windows_env
+    OTP_VERSION: << parameters.otp_version >>
+    ERTS_VERSION: << parameters.erts_version >>
+    WIN_OTP_PATH: C:\tools\erl<< parameters.erts_version >>
+    WIN_MSYS2_ROOT: C:\tools\msys64
+    JAVA_VERSION: 12.0.2
+    WIN_JDK_BASEPATH: C:\Program Files\OpenJDK
+    JAVA_TOP: /c/Program Files/OpenJDK/jdk-12.0.2
+    MSYSCON: defterm
+    ANSICON: true
+    MSVC_VERSION: 14.22.27905
+    VCVARSALL: C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvarsall.bat
+
+  windows_container: &windows_container
+    executor:
+      name: win/vs2019
+      shell: cmd.exe
+    environment: *windows_env
+    parameters:
+      msys_install_path:
+        default: C:\tools\msys64
+        type: string
+      otp_install_path:
+        default: C:\tools\erl9.1
+        type: string
+      otp_version:
+        default: "20.1"
+        type: string
+      erts_version:
+        default: "9.1"
+        type: string
+    working_directory: *windows_working_directory
+
 commands:
   setup_secrets:
     steps:
@@ -432,6 +474,111 @@ commands:
             STATUS=$(docker inspect --format='{{.State.Health.Status}}' ae_${CIRCLE_SHA1:?})
             echo $STATUS
             [ $STATUS == 'healthy' ]
+
+  restore_windows_package_cache:
+    parameters:
+      win_cache_version:
+        default: *windows_cache_version
+        type: string
+    steps:
+     - restore_cache:
+        keys:
+          - windows-package-cache-<< parameters.win_cache_version >>-{{ .Branch }}-
+          - windows-package-cache-<< parameters.win_cache_version >>-master-
+          - windows-package-cache-<< parameters.win_cache_version >>-
+
+  save_windows_package_cache:
+    parameters:
+      win_cache_version:
+        type: string
+        default: *windows_cache_version
+      install_path:
+        type: string
+    steps:
+      - save_cache:
+          key: windows-package-cache-<< parameters.win_cache_version >>-{{ .Branch }}-
+          paths:
+            - << parameters.install_path >>
+
+  install_win_msys2:
+    parameters:
+      install_path:
+        default: C:\tools\msys64
+        type: string
+    steps:
+      - restore_windows_package_cache
+      - run:
+          name: Install MSYS2
+          command: .circleci\windows\msys2_prepare.bat
+      - save_windows_package_cache:
+          install_path: << parameters.install_path >>
+
+  restore_windows_erl_cache:
+    parameters:
+      win_cache_version:
+        default: *windows_cache_version
+        type: string
+      erts_version:
+        type: string
+    steps:
+      - restore_cache:
+          keys:
+            - windows-erl<< parameters.erts_version >>-cache-<< parameters.win_cache_version >>-{{ .Branch }}-
+            - windows-erl<< parameters.erts_version >>-cache-<< parameters.win_cache_version >>-master-
+            - windows-erl<< parameters.erts_version >>-cache-<< parameters.win_cache_version >>-
+
+  save_windows_erl_cache:
+    parameters:
+      win_cache_version:
+        default: *windows_cache_version
+        type: string
+      install_path:
+        type: string
+      erts_version:
+        type: string
+    steps:
+      - save_cache:
+          key: windows-erl<< parameters.erts_version >>-cache-<< parameters.win_cache_version >>-{{ .Branch }}-
+          paths:
+            - << parameters.install_path >>
+
+  install_win_otp:
+    parameters:
+      install_path:
+        type: string
+      otp_version:
+        type: string
+      erts_version:
+        type: string
+    steps:
+      - restore_windows_erl_cache:
+          erts_version: << parameters.erts_version >>
+      - run:
+          name: Install Erlang/OTP << parameters.otp_version >>
+          command: scripts\windows\install_erlang << parameters.otp_version >> << parameters.install_path >>
+      - save_windows_erl_cache:
+          erts_version: << parameters.erts_version >>
+          install_path: << parameters.install_path >>
+
+  windows_install_deps:
+    parameters:
+      msys_install_path:
+        type: string
+      otp_install_path:
+        type: string
+      otp_version:
+        type: string
+      erts_version:
+        type: string
+    steps:
+      - checkout
+      - install_win_otp:
+          install_path: << parameters.otp_install_path >>
+          otp_version: << parameters.otp_version >>
+          erts_version: << parameters.erts_version >>
+      - install_win_msys2:
+          install_path: << parameters.msys_install_path >>
+
 
 jobs:
   build:
@@ -954,6 +1101,49 @@ jobs:
           path: docs/state-channels
       - *fail_notification
 
+  # Windows jobs
+  win_build: &windows_build
+    <<: *windows_container
+    steps:
+      - windows_install_deps:
+          otp_install_path: << parameters.otp_install_path >>
+          otp_version: << parameters.otp_version >>
+          erts_version: << parameters.erts_version >>
+          msys_install_path: << parameters.msys_install_path >>
+      - run:
+          name: Build
+          command: .circleci\windows\build.bat
+      - store_artifacts:
+          path: *windows_artifacts
+          destination: /tmp/packages-<< parameters.otp_version >>
+      - persist_to_workspace:
+          root: *windows_artifacts
+          paths:
+            - "*.zip"
+            - "*.exe"
+
+  win_test: &windows_test
+    <<: *windows_container
+    steps:
+      - windows_install_deps:
+          otp_install_path: << parameters.otp_install_path >>
+          otp_version: << parameters.otp_version >>
+          erts_version: << parameters.erts_version >>
+          msys_install_path: << parameters.msys_install_path >>
+      - attach_workspace:
+          at: *windows_artifacts
+      - run:
+          name: Smoke test
+          command: .circleci\windows\test.bat
+      - store_artifacts:
+          path: *windows_artifacts
+          destination: /tmp/packages-<< parameters.otp_version >>
+
+  win_build_20_1: *windows_build
+  win_build_20_3: *windows_build
+  win_test_20_1: *windows_test
+  win_test_20_3: *windows_test
+
 # CircleCI skips a job for a tag by default.
 # A job must have a filters tags section to run as a part of a tag push
 # and all its transitively dependent jobs must also have a filters tags section.
@@ -961,6 +1151,58 @@ workflows:
   version: 2
   build_test_deploy:
     jobs:
+      - win_build_20_3:
+          otp_install_path: C:\tools\erl9.3
+          otp_version: "20.3"
+          erts_version: "9.3"
+          context: ae-node-builds
+          filters:
+            branches:
+              ignore:
+                - env/dev1
+                - env/dev2
+                - system-tests
+
+      - win_test_20_3:
+          otp_install_path: C:\tools\erl9.3
+          otp_version: "20.3"
+          erts_version: "9.3"
+          context: ae-node-builds
+          requires:
+            - win_build_20_3
+          filters:
+            branches:
+              ignore:
+                - env/dev1
+                - env/dev2
+                - system-tests
+
+      - win_build_20_1:
+          otp_install_path: C:\tools\erl9.1
+          otp_version: "20.1"
+          erts_version: "9.1"
+          context: ae-node-builds
+          filters:
+            branches:
+              ignore:
+                - env/dev1
+                - env/dev2
+                - system-tests
+
+      - win_test_20_1:
+          otp_install_path: C:\tools\erl9.1
+          otp_version: "20.1"
+          erts_version: "9.1"
+          context: ae-node-builds
+          requires:
+            - win_build_20_1
+          filters:
+            branches:
+              ignore:
+                - env/dev1
+                - env/dev2
+                - system-tests
+
       - build:
           filters:
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1093,7 +1093,7 @@ jobs:
       - *fail_notification
 
   # Windows jobs
-  win_build: &windows_build
+  windows_package:
     <<: *windows_container
     steps:
       - windows_install_deps:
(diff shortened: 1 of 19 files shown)


ci/windows/package.cfg
```diff
@@ -0,0 +1,79 @@
...
+
+packages = {pkg_prefix}libsodium
+
+filename_stub = aeternity-windows
+
+display_name = Aeternity for Windows
+
+description = The full Aeternity node for Windows.
+
+version = 4.2.0
```
venimus: This should be changed when new version is introduced

venimus: This is now automatically resolved

.circleci/windows/erlang_install.bat
```diff
@@ -0,0 +1,20 @@
+@echo off
+IF NOT "x%1"=="x" IF NOT "x%2"=="x" GOTO:INSTALL
+@echo Usage: %0 ^<otp_version^> ^<install_path^>
+exit /b 2
+
+:INSTALL
+@echo %time% : Install Erlang/OTP %1
+IF EXIST %2 GOTO:INSTALLED
+
+SET "URL=https://packages.erlang-solutions.com/erlang/erlang/esl-erlang/FLAVOUR_1_general/esl-erlang_%1~windows_amd64.exe"
```
venimus: This is a much faster download mirror, not sure if reliable. Official site download takes some minutes vs few seconds here

.circleci/windows/erlang_install.bat
```diff
@@ -0,0 +1,20 @@
...
+@echo %time% : Download %URL% started
+curl -s -o "%temp%\erl%1.exe" "%URL%"
+IF NOT "%ERRORLEVEL%"=="0" (echo %time% : Download failed && exit /b %ERRORLEVEL%)
+
+@echo %time% : Install started
+powershell start-process -wait '%temp%\erl%1.exe' -argumentlist '/S /D=%2'
```
venimus: since the installer always runs as a separate process we need to hold until it finish

.circleci/windows/erlang_install.bat
```diff
@@ -0,0 +1,20 @@
+@echo off
+IF NOT "x%1"=="x" IF NOT "x%2"=="x" GOTO:INSTALL
+@echo Usage: %0 ^<otp_version^> ^<install_path^>
```
venimus: although this is CI script i made it user friendly so it could be reused locally by users

.circleci/windows/msys2_install.bat
```diff
@@ -0,0 +1,35 @@
...
+ECHO."%PATH%" | FINDSTR /V /C:"%_PATHS%">nul && SET "PATH=%_PATHS%;%PATH%"
+SET "_PATHS="
+
+SETLOCAL ENABLEEXTENSIONS
+
+IF EXIST "%WIN_MSYS2_ROOT%\mingw64" GOTO :MSYS2_UPDATE
+md "%WIN_MSYS2_ROOT%"
+pushd "%WIN_MSYS2_ROOT%\.."
+
+curl -s -O https://tukaani.org/xz/xz-5.2.4-windows.zip
```
venimus: MSYS2 is distributed as this inadequate (for windows) .xz format which requires an additional tool to decompress it.
I didn't want to use 7z because it might not be available

.circleci/windows/msys2_install.bat
```diff
@@ -0,0 +1,35 @@
...
+tar -xf xz-5.2.4-windows.zip
```
venimus: tar is available in Windows 10

.circleci/windows/msys2_install.bat
```diff
@@ -0,0 +1,35 @@
...
+@echo %time%: Msys2 download
+curl -s -O https://netcologne.dl.sourceforge.net/project/msys2/Base/x86_64/msys2-base-x86_64-20190524.tar.xz
+
+@echo %time%: Msys2 extract
+bin_x86-64\xz -d -qq msys2-base-x86_64-20190524.tar.xz
+tar -xf msys2-base-x86_64-20190524.tar
+popd
+
+@echo %time%: Msys2 Initial setup
+bash -lc "echo Done."
```
venimus: First run does some preparations of MSYS so I just dump tracking text

.circleci/windows/msys2_install.bat
```diff
@@ -0,0 +1,35 @@
...
+
+@echo %time%: Msys2 Update 1/2
+bash -lc "pacman --noconfirm -Syuu"
```
venimus: msys requires few runs of pacman to complete the installation.

.circleci/windows/msys2_prepare.bat
```diff
@@ -0,0 +1,4 @@
+@echo on
+git config --global --unset url."ssh://git@github.com".insteadOf "https://github.com"
```
venimus: CircleCI by default has set this in git config, but it prevents to download from public repos (specifically styrene) without a key

.circleci/windows/package.cfg
```diff
@@ -0,0 +1,79 @@
...
+
+packages = {pkg_prefix}libsodium
+
+filename_stub = aeternity-windows
+
+display_name = Aeternity for Windows
+
+description = The full Aeternity node for Windows.
+
+version = VERSION_PLACEHOLDER
```
venimus: This will be replaced by the node version automatically, to avoid changing the script with every release

.circleci/windows/test.bat
```diff
@@ -0,0 +1,66 @@
+@echo on
```
venimus: This script simulates user behavior. It extracts the .zip artifact and runs smoke tests

.circleci/windows/test.bat
```diff
@@ -0,0 +1,66 @@
...
+:TEST_eunit
+@echo %time% : Run test: eunit
+bash -lc "cd ${BUILD_PATH} && epmd -daemon && make eunit"
+exit /b 0
+
+:TEST_release
+@echo %time% : Run test: release
+bash -lc ^"cd ${BUILD_PATH} ^&^& ^
+           make python-env PIP=/mingw64/bin/pip3 ^&^& ^
+           rm -rf ${PACKAGE_TESTS_DIR} ^&^& ^
```
venimus: I did remove `epmd -daemon` as it make CircleCI wait forever after the job finishes (I reported them this, but don't expect much) I guess CircleCI finds that there is a background process that is executing and will wait for it but will fail after 10 minutes with "No output for 10 minutes" message.
Unfortunately using 'epmd -kill' didn't fix the problem.

.circleci/windows/vcvarsall.bat
```diff
@@ -0,0 +1,19 @@
+@echo on
```
venimus: this helper script finds and executes `vcvarsall.bat` which is located on different places depending on the VS version and sets MSCV_VERSION var to the proper value.

scripts/windows/msys2_shell.bat
```diff
@@ -23,21 +23,23 @@ FOR /F "tokens=* USEBACKQ delims=" %%F IN (`type "%ToolsVerFile%"`) DO SET MSVC_
...
+rem Set the paths appropriately avoiding duplicates
+SET "_PATHS=%WIN_MSYS2_ROOT%\mingw64\bin;%WIN_MSYS2_ROOT%\usr\bin;%WIN_MSYS2_ROOT%"
+ECHO."%PATH%" | FINDSTR /V /C:"%_PATHS%">nul && SET "PATH=%_PATHS%;%PATH%"
+SET "_PATHS="
 
 echo Current time: %time%
 rem Open shell
 
-%WIN_MSYS2_ROOT%\mingw64.exe
+"%WIN_MSYS2_ROOT%\mingw64.exe"
```
venimus: path might include spaces and it will fail without `"`

.circleci/config.yml
```diff
@@ -434,6 +476,111 @@ commands:
...
+      - restore_windows_erl_cache:
+          erts_version: << parameters.erts_version >>
+      - run:
+          name: Install Erlang/OTP << parameters.otp_version >>
+          command: scripts\windows\install_erlang << parameters.otp_version >> << parameters.install_path >>
+      - save_windows_erl_cache:
+          erts_version: << parameters.erts_version >>
+          install_path: << parameters.install_path >>
+
+  windows_install_deps:
```
venimus: OTP and MSYS are split to separate steps, so they can use different caches (this saves some build time) and allows different jobs with different Erlang versions to be cached properly

.circleci/config.yml
```diff
@@ -960,13 +1107,108 @@ jobs:
...
+      - attach_workspace:
+          at: *windows_artifacts
+      - run:
+          name: Smoke test
+          command: .circleci\windows\test.bat
+      - store_artifacts:
+          path: *windows_artifacts
+          destination: /tmp/packages-<< parameters.otp_version >>
+
+  win_build_20_1: *windows_build
```
venimus: We need to create something like aliases for the jobs, because config requires every job to have unique key. This is kind of hackish

.circleci/windows/build.bat
```diff
@@ -0,0 +1,34 @@
...
+CALL "%~dp0vcvarsall.bat"
+
+@call:log Build production release
+"%WIN_MSYS2_ROOT%\usr\bin\bash.exe" -lc "${PROJECT_ROOT}/.circleci/windows/build.sh %PACKAGES_PATH%"
+@call:log Build done.
+
+exit /b %ERRORLEVEL%
+
+:log :: Display a log message
+@echo :: [1;33m %time% : %* [0m>con && exit /b
```
venimus: Unfortunately there is no way to send escaped ESC sequences in windows, so we need to put the ESC char as is. It is a valid ASCII char anyway. This adds some color in the build log and makes it much easier to inspect/debug

.circleci/windows/build.sh
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
```
venimus: This script was extracted from build.bat (which was simply running bash commands)

.circleci/windows/build.sh
```diff
@@ -0,0 +1,64 @@
...
+rm -rf ${WORK_ROOT}
+mkdir -p ${WORK_ROOT}
+
+echo -e "${bold}Start build${reset}"
+cd ${PROJECT_ROOT}
+time make prod-build
+
+echo -e "${bold}Create package environment${reset}"
+NODE_VERSION=`cat ${PROJECT_ROOT}/VERSION`
+sed "s/^version =.*$/version = ${NODE_VERSION}/g" .circleci/windows/package.cfg > ${WORK_ROOT}/package.cfg
```
venimus: Resolve node's version and put it in package.cfg

.circleci/windows/build.sh
```diff
@@ -0,0 +1,64 @@
...
+time cp -R ${RELEASE_PATH} ${WORK_PATH}/usr/lib/
+mv -f ${WORK_PATH}/usr/lib/aeternity/REVISION ${WORK_PATH}/
+mv -f ${WORK_PATH}/usr/lib/aeternity/VERSION ${WORK_PATH}/
+
+# Remove erl.ini files from release
+find ${WORK_PATH}/usr/lib -name erl.ini -type f -delete
+
+# Copy genesis and hard-fork account migrations into top-level data folder
+mkdir -p ${WORK_PATH}/data
+cp -R ${RELEASE_PATH}/data/aecore ${WORK_PATH}/data/
```
venimus: I couldn't find why the old script was copying the genesis accounts implicitly one by one.

.circleci/windows/msys2_prepare.bat
```diff
@@ -0,0 +1,6 @@
+@echo on
+:: Unset CircleCI default setting that prevents pulling Styrene from the public repo
+git config --global --unset url."ssh://git@github.com".insteadOf "https://github.com"
```
venimus: CircleCI sets `git config --global url.ssh://git@github.com.insteadOf https://github.com` on the 'checkout' step and forgets to remove it. However it prevents to download from public repos without a key. So we need to revert this setting to be able to checkout Styrene (which has no release binary)

.circleci/windows/package.cfg
venimus: Version is no longer hard-coded and it will be resolved on build time

.circleci/windows/test.bat
```diff
@@ -0,0 +1,70 @@
...
+@for /f "tokens=1* delims=, " %%i in ("%TEST_STEPS%") do @(
+    @call:log Run test %%i
+	call :TEST_%%i
+)
+
+@call:log Finished test phase
+exit /b %ERRORLEVEL%
+
+:: Subroutines
+:TEST_ct
```
venimus: These steps are not used by CI but I kept them for BC

.circleci/windows/test.bat
```diff
@@ -0,0 +1,70 @@
...
+@rem    PROJECT_ROOT
+
+SETLOCAL ENABLEEXTENSIONS
+@call:log Set the paths appropriately
+
+:: Appropriately set paths
+SET "PATH=%WIN_MSYS2_ROOT%\mingw64\bin;%WIN_MSYS2_ROOT%\usr\bin;%WIN_MSYS2_ROOT%;%PATH%"
+SET "PATH=%WIN_OTP_PATH%\bin;%WIN_OTP_PATH%\erts-%ERTS_VERSION%\bin;%WIN_OTP_PATH%\erts-%ERTS_VERSION%;%PATH%"
+
+:: Construct unix paths
```
venimus: Next set of commands will try to resolve where the artifacts are, so this script could be used locally. In addition because msys is a posix environment we need to prepare both versions of windows and unix forms of the paths

scripts/windows/install_erlang.bat
```diff
@@ -0,0 +1,73 @@
...
+IF NOT "%OTP_VERSION%"=="%~1" GOTO:INSTALL
+IF EXIST "%WIN_OTP_PATH%\bin\" @call:log Erlang/OTP %OTP_VERSION% && GOTO:DONE
+:INSTALL
+@call:log Install Erlang/OTP %~1 into %WIN_OTP_PATH%
+IF "%OTP_URL%"=="" SET "OTP_URL=https://packages.erlang-solutions.com/erlang/erlang/esl-erlang/FLAVOUR_1_general/esl-erlang_%~1~windows_amd64.exe"
+@call:log Download from %OTP_URL%
+PowerShell (New-Object System.Net.WebClient).DownloadFile('%OTP_URL%', '%TMP%\erl.exe')
+@IF NOT "%ERRORLEVEL%"=="0" (call:log Download failed && exit /b %ERRORLEVEL%)
+
+:: Uninstall the old version
```
venimus: Just in case we have old version of Erlang in the same directory we have to uninstall it first. If we install in a separate folder <install_path> we can have multiple versions simultaneously. Then switching the active can be done via ENV

scripts/windows/install_msys2.bat
```diff
@@ -0,0 +1,66 @@
...
+:START
+IF NOT "%~1"=="" SET "WIN_MSYS2_ROOT=%~1"
+IF "%WIN_MSYS2_ROOT%"=="" SET "WIN_MSYS2_ROOT=C:\tools\msys64"
+IF "%MSYS2_URL%"=="" SET "MSYS2_URL=https://netcologne.dl.sourceforge.net/project/msys2/Base/x86_64/msys2-base-x86_64-20190524.tar.xz"
+
+IF EXIST "%WIN_MSYS2_ROOT%\mingw64" call:log Ok && GOTO :MSYS2_INSTALLED
+@call:log Target directory %WIN_MSYS2_ROOT%
+
+IF EXIST "%WIN_MSYS2_ROOT%" rd /s/q "%WIN_MSYS2_ROOT%"
+SET "PSAdmin=powershell Start-Process -wait -WindowStyle hidden 'powershell.exe'"
```
venimus: This 'Inception' command is to be able to prepare a self-elevating shell for executing stuff requiring admin rights

scripts/windows/install_msys2.bat
```diff
@@ -0,0 +1,66 @@
...
+
+@call:log Download from %MSYS2_URL%
+powershell (New-Object System.Net.WebClient).DownloadFile('%MSYS2_URL%', '%TMP%\msys2.tar.xz')
+
+@call:log Download PowerShell 7z helper
+%PSAdmin% 'Install-Package -Scope CurrentUser -Force 7Zip4PowerShell'
```
venimus: Unfortunately msys2 is distributed in this inadequate .xz compressed format and we need some way to extract it. Seems the easiest way is to use this PS module that adds 7z to powershell

scripts/windows/install_msys2.bat
```diff
@@ -0,0 +1,66 @@
+@echo off
+:: Pass -v as first argument to enable debug echo
+:: Script to download and install Msys2 from %MSYS2_URL%
```
venimus: Installing MSYS2 is quite complex and does not have an installer, so we need our own script to do it. It could be very helpful for users too

scripts/windows/msys2_prepare.bat
```diff
@@ -11,26 +13,101 @@
...
+:: Dump detected version
+@echo SET MSVC_VERSION=%MSVC_VERSION%
+
+@call:log Find and execute the VS env preparation script for %PLATFORM%
+call:VCVARSALL
+IF NOT "%ERRORLEVEL%"=="0" exit /b %ERRORLEVEL%
+
+:: Set the paths appropriately avoiding duplicates
+SET "_PATHS=%WIN_MSYS2_ROOT%\mingw64\bin;%WIN_MSYS2_ROOT%\usr\bin;%WIN_MSYS2_ROOT%"
+ECHO."%PATH%" | FINDSTR /V /C:"%_PATHS%">nul && SET "PATH=%_PATHS%;%PATH%"
```
venimus: This "ugly" mess is to check if we already have changed the PATH in a previous run. Users may run this script multiple times and we don't want to fill the PATH with duplicates. Shockingly it has a limit of 1024 chars. Visual studio preparation script that is executed earlier shamelessly adds dozens of entries already

scripts/windows/msys2_prepare.bat
```diff
@@ -1,4 +1,6 @@
 @echo off
```
venimus: Changes in this script are basically to add more checks for env and refactor it to subroutines for easier maintenance and clarity. It does not do anything fundamentally different from the old one

scripts/windows/msys2_prepare.bat
```diff
@@ -66,74 +143,41 @@ mingw-w64-x86_64-python3-pip ^
...
-
-rem Set the paths appropriately
-SET "PATH=%WIN_MSYS2_ROOT%\mingw64\bin;%WIN_MSYS2_ROOT%\usr\bin;%WIN_OTP_PATH%\bin;%WIN_OTP_PATH%\erts-%OTP_VERSION%;%PATH%"
+@call:log Set up msys2 env variables
+:: 1>nul to hide the Copied 1 file message
+COPY %~dp0msys2_env_build.sh %WIN_MSYS2_ROOT%\etc\profile.d\env_build.sh 1>nul
 
-echo Current time: %time%
-rem Set up msys2 env variables
+SET BASH="%WIN_MSYS2_ROOT%\usr\bin\bash.exe"
```
venimus: Unfortunately setting Path is not always enough to guarantee that we execute the proper bash (i.e. we can have WSL, or git bash already installed). So we better implicitly refer to the executable.

venimus:
I'm not sure if it is not better to build with otp 21.0 instead of 20.1

venimus:
Latest commits:
* Rebased and added OTP21.3 build (also removed redundant ci params)
* In addition added some UX improvements to install scripts
* Updated docs

.circleci/windows/package.cfg
dincho: I don't really like the `VERSION_PLACEHOLDER`, perhaps I'd expect `master` or something "real" - but OK

venimus: it does not really matter what it will be anyway. It does not stop building the package. Also replace will look for `version =`. While missing this step and using it directly it will make this more obvious for debugging

scripts/windows/install_erlang.bat
```diff
@@ -0,0 +1,82 @@
+@echo off
```
venimus: This script is to streamline installation of erlang supporting multiple simultaneous versions and uses erlang-solutions as mirror (by default) as it is way too faster to download. CI download and install will require a script anyway. Chocolatey script does not support installing multiple versions properly and also leaves a running process which bugs the job execution (waits forever for it to finish)

scripts/windows/msys2_prepare.bat
```diff
@@ -11,26 +12,119 @@
...
-IF "%WIN_JDK_BASEPATH%"=="" SET "WIN_JDK_BASEPATH=C:\Program Files\Java"
-IF "%JAVA_VERSION%"=="" SET "JAVA_VERSION=11.0.2"
-IF "%WIN_JDK_PATH%"=="" SET "WIN_JDK_PATH=C:\Program Files\Java\jdk-%JAVA_VERSION%"
 IF "%WIN_MSYS2_ROOT%"=="" FOR /F %%F IN ('where msys2') DO SET "WIN_MSYS2_ROOT=%%~dpF"
-IF "%WIN_OTP_PATH%"=="" SET "WIN_OTP_PATH=C:\Program Files\erl%ERTS_VERSION%"
+:: default if autodetect failed
+IF "%WIN_MSYS2_ROOT%"=="" SET "WIN_MSYS2_ROOT=C:\tools\msys64"
+
+IF "%FORCE_STYRENE_REINSTALL%"=="" SET "FORCE_STYRENE_REINSTALL=false"
+IF "%JDK_URL%"=="" SET "JDK_URL=https://download.java.net/java/GA/jdk11/9/GPL/openjdk-11.0.2_windows-x64_bin.zip"
```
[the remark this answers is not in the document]
venimus: It is probably not used in CI, but this script is also used by 'users/devs' which may do other tasks that require it -> build UML diagrams, so we need to get rid of it there first. Or just let them install it manually as with Visual Studio

.circleci/windows/package.cfg
```diff
@@ -0,0 +1,79 @@
...
+name = peer_key
+comment = peer_key utility for Aeternity
+exec = ./usr/lib/aeternity/bin/aeternity.cmd peer_key
+icon = input-tablet
+terminal = false
+
+[rename_db.desktop]
+name = rename_db
+comment = rename_db utility for Aeternity
+exec = ./usr/lib/aeternity/bin/aeternity.cmd rename_db
```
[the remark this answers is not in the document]
venimus: I'm not sure which of these are still valid (and what they should do). May be someone more qualified should check it. @tolbrino ?

venimus:
> I see 3 windows build jobs using 3 different OTP versions - it does not make sense to me:

Yep, It is the main concern to me too. I actually left this just to decide which one should stay, as only one of the resulting packages will be distributed anyway.
I did split the test to make sure we use clean environment. But if the CPU-time is an issue there is no problem to merge it.
There is also a possibility to have 1 job that do 2-3 builds at the same time

venimus:
> Blocking mainly because of the old `rename_db` command and multiple build jobs.

I merged both steps and only left the 20.3 build as it seems most stablehttps://github.com/aeternity/aeternity/pull/2692/commits/3076b9298c25bf42581c2eb0f8a29bbe859449fa.

And removed the cmd https://github.com/aeternity/aeternity/pull/2692/commits/aac271eed91b01d4b20cbff9a538e950b4b66e0d

.circleci/config.yml
```diff
@@ -952,13 +1090,41 @@ jobs:
...
+
 # CircleCI skips a job for a tag by default.
 # A job must have a filters tags section to run as a part of a tag push
 # and all its transitively dependent jobs must also have a filters tags section.
 workflows:
   version: 2
   build_test_deploy:
     jobs:
+      - windows_package:
+          otp_version: "20.3"
```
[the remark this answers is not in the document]
venimus: I would too, but was not sure if it is officially supported

.circleci/config.yml
```diff
@@ -399,6 +403,40 @@ references:
...
+  # Windows
+  windows_cache_version: &windows_cache_version "v1"
+  windows_artifacts: &windows_artifacts C:\Users\circleci\project\packages
+  windows_working_directory: &windows_working_directory C:\Users\circleci\project
+
+  windows_env: &windows_env
+    OTP_VERSION: << parameters.otp_version >>
+    WIN_OTP_PATH: << parameters.otp_install_path >>
+    WIN_MSYS2_ROOT: << parameters.msys_install_path >>
+    JAVA_VERSION: 12.0.2
```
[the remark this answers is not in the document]
venimus: I left it so preparation script will skip installing it. It is already preinstalled in the orb.
May be I will remove it from there and note to devs that they should install it manually if needed

venimus:
last commit:
* Removed Java requirements (and references) and updated doc
* Renamed most .bat files to .cmd (as they have cmd extensions by default)
I didn't rename `msys2_prepare.bat` and `msys2_shell.bat` for BC
* Switched build to OTP 21.3

PR #2693, opened 2019-08-22
Wrap network and timing sensitive CI operations in a retry loop

tolbrino:
Often CI jobs fail due to small random network errors or host CPU
stalls. This change introduces a simple retry loop around operations
which are prone to such issues. If a job still fails after a retry the
author of a change can at least be confident that the issue wasn't a
one-off problem.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,6 +73,8 @@ references:
   install_os_deps: &install_os_deps
     run:
       name: Install OS dependancies
+      environment:
+        MAX_RETRIES: 5
       command: |
         ./ci/try sudo apt-get -qq -y install git curl \
           autoconf build-essential ncurses-dev libssl-dev
@@ -83,6 +85,7 @@ references:
       environment:
         # See minimum_otp_vsn in rebar.config
         OTP_VERSION: "20.1"
+        MAX_RETRIES: 5
       command: |
         # Install OTP package deps
         ./ci/try sudo apt-get -qq -y update
@@ -100,6 +103,7 @@ references:
       name: Install libsodium
       environment:
         LIBSODIUM_VERSION: "1.0.16"
+        MAX_RETRIES: 5
       # source: https://github.com/aeternity/docker-builder/blob/master/Dockerfile#L23
       command: |
         LIBSODIUM_DOWNLOAD_URL="https://github.com/jedisct1/libsodium/releases/download/${LIBSODIUM_VERSION}/libsodium-${LIBSODIUM_VERSION}.tar.gz"
@@ -216,9 +220,9 @@ references:
         PACKAGE_TESTS_DIR: *package_tests_workspace
       command: |
         epmd -daemon
-        ./ci/try make python-env
+        MAX_RETRIES=5 ./ci/try make python-env
         mkdir ${PACKAGE_TESTS_DIR:?}
-        MAX_RETRIES=2 ./ci/try make python-package-win32-test python-release-test WORKDIR=${PACKAGE_TESTS_DIR:?} PACKAGE=${PACKAGE_TARBALL:?}
+        ./ci/try make python-package-win32-test python-release-test WORKDIR=${PACKAGE_TESTS_DIR:?} PACKAGE=${PACKAGE_TARBALL:?}
 
   test_arch_os_dependencies: &test_arch_os_dependencies
     run:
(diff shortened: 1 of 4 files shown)


ci/try
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env sh
+#
+# This script executes a given COMMAND until it succeeds or MAX_RETRIES have
+# been reached. In between retries the script waits for
+# (INITIAL_TIMEOUT)^RETRIES , such that the time between retries increases.
+
+MAX_RETRIES=${MAX_RETRIES:-5}
```
venimus: I would suggest this to be 2 by default, IMHO 5 is too much retries and just set the var for jobs that tend to recover with more.

.circleci/config.yml
```diff
@@ -101,12 +102,15 @@ references:
...
       command: |
-        LIBSODIUM_DOWNLOAD_URL="https://github.com/jedisct1/libsodium/releases/download/${LIBSODIUM_VERSION}/libsodium-${LIBSODIUM_VERSION}.tar.gz" \
-          && curl -fsSL -o libsodium-src.tar.gz "$LIBSODIUM_DOWNLOAD_URL" \
-          && mkdir libsodium-src \
-          && tar -zxf libsodium-src.tar.gz -C libsodium-src --strip-components=1 \
-          && cd libsodium-src \
-          && ./configure && make -j$(nproc) && sudo make install && sudo ldconfig
+        LIBSODIUM_DOWNLOAD_URL="https://github.com/jedisct1/libsodium/releases/download/${LIBSODIUM_VERSION}/libsodium-${LIBSODIUM_VERSION}.tar.gz"
+          ./ci/try curl -fsSL -o libsodium-src.tar.gz "$LIBSODIUM_DOWNLOAD_URL"
+          mkdir libsodium-src
```
venimus: I see you removed `&&` which means that the command will not check if the previous one failed and will execute it. Is this on purpose?

tolbrino: These commands are execute with `/bin/bash -eo pipefail` so any command which exits with >0 should also make the job fail.

venimus: That makes sense. Does the try script return an exit code when failed?

dincho:
I'm not really after this, can you post a list what intermittent CI failures this is trying to solve?

In my opinion this would only extend the failure agony, being a fan of "fail fast" principle, I don't think this will help 99.99% of the cases.

Network issues and CPU stalls are platform issues (in this case CircleCI) we shouldn't try to workaround but report to the provider. Otherwise they become lazy and don't fix it forever, take for example TravisCI - you can't really run anything there without retries.

In regards to test retries, maybe that makes a bit of more sense, given our tests are very fragile and time sensitive, however I still don't think it's the CI job to do that but the test-running so that it can restart a single test or suite, not the all the tests.

So, my conclusion would be to not merge this as is, however I'll not block it as someone might prove me wrong.

tolbrino:
I would not trust any CI to be bullet proof in terms of network issues. I support the "fail fast" goal, but for jobs which fail due to such issues after 20 minutes "fast" is relative. In that case I'd rather have it retry automatically since my attention is already somewhere else.

However, this PR is merely a proposal to improve the current situation where there are plenty of cases where failed jobs are simply re-run and pass.

.circleci/config.yml
```diff
@@ -355,7 +363,7 @@ references:
         name: Test
         command: |
           epmd -daemon
-          MAX_RETRIES=2 ./ci/try make ${MAKE_TARGET:?} CT_TEST_FLAGS="--suite=$(.circleci/split_suites.sh)"
+          ./ci/try make ${MAKE_TARGET:?} CT_TEST_FLAGS="--suite=$(.circleci/split_suites.sh)"
```
dincho: Note that this may run different set of sets if it's run multiple times, I'm not aware how the CircleCI internal mechanics works and if they always return the same results for a commit hash or not.

That script runs:
```
circleci tests split --split-by=timings --timings-type=classname
```

The split is statistical based on latest run times.

.circleci/config.yml
tolbrino: Now it does.

ci/try
tolbrino: Changed.

dincho:
> However, this PR is merely a proposal to improve the current situation where there are plenty of cases where failed jobs are simply re-run and pass.

I do agree with the idea, and indeed this is the fast workaround, but the better workaround as I mentioned is retry in the test-runners

dincho:
> I do agree with the idea, and indeed this is the fast workaround, but the better workaround as I mentioned is retry in the test-runners

Maybe something like this is even better ? https://github.com/ferd/cth_retry ?

tolbrino:
I've changed the PR such that it should use the retry functionality from `common_test` where possible.

tolbrino:
Closing this since CircleCI seems to have stabilized.

PR #2698, opened 2019-08-23
also pass version opts at reestablish

uwiger:
See [PT #168074662](https://www.pivotaltracker.com/story/show/168074662)

System tests fail since proper encoding versions are not passed along in `reestablish`, causing the newer node to use the wrong encoding version for `offchain_update`.


uwiger:
pushed to the `system-tests` branch ...

uwiger:
The `aest_channels` test passed in the `system-tests` job

PR #2704, opened 2019-08-26
Pt 167889960 disallow claiming test registrar

ThomasArts:
Improved version of PR #2667 where systems tests and others are depending on height to not break when we switch to Lima


apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -3544,7 +3544,7 @@ sophia_signatures_aens(Cfg) ->
                   VMVersion when ?IS_FATE_SOPHIA(VMVersion) -> Name1
               end,
     NameAccSig      = sign(<<NameAcc/binary, Ct/binary>>, NameAcc),
-    {ok, NameHash}  = aens:get_name_hash(aens_test_utils:fullname(<<"bla">>)),
+    {ok, NameHash}  = aens:get_name_hash(Name1),
     NameSig         = sign(<<NameAcc/binary, NameHash/binary, Ct/binary>>, NameAcc),
```
ThomasArts: This is weird, I would expect that it depends on the protocol whether to use this name hash or the actual name

apps/aens/src/aens_utils.erl
```diff
@@ -59,12 +60,15 @@ check_claimed_status(Name) ->
         revoked -> {error, name_revoked}
     end.
 
+name_parts(Name) ->
+    binary:split(Name, ?LABEL_SEPARATOR, [global, trim]).
+
 validate_name(Name) ->
-    case binary:split(Name, ?LABEL_SEPARATOR, [global, trim]) of
+    case name_parts(Name) of
```
hanssv: This isn't optimal - better to know what height you are at and reject `.test`/`.aet` accordingly...

apps/aens/test/aens_test_utils.erl
```diff
@@ -98,6 +102,24 @@ set_account(Account, State) ->
     AccTree = aec_accounts_trees:enter(Account, aec_trees:accounts(Trees)),
     set_trees(aec_trees:set_accounts(Trees, AccTree), State).
 
+%% protocol dependent
+
+latest_protocol_version() ->
```
hanssv: This is `aect_test_utils:latest_protocol_version()`?

apps/aens/src/aens_utils.erl
ThomasArts: ticket created for after Lima release: https://www.pivotaltracker.com/story/show/168246084

PR #2630, opened 2019-08-08
[PT-164136571] State channel cache encryption

grzegorz225:
Corresponding protocol PR: https://github.com/aeternity/protocol/pull/396

PR #2709, opened 2019-08-26
WIP let recconnect try reestablish (or kill idle client)

uwiger:
See [PT #167752251](https://www.pivotaltracker.com/story/show/167752251) and [PT #168073522](https://www.pivotaltracker.com/story/show/168073522)

Currently, no extra tests have been added, but old tests seem to pass.

This PR introduces a few new things:
* If, at reconnect, the FSM still thinks the client is alive, it returns an error with the client pid
* The reconnecting client then tries to detect if the old client is passive (currently, no activity in the last 5 seconds). If so, it kills it, sleeps a bit and tries once more.
* If there is no FSM, the reconnecting client tries to fetch cached state, and if it finds it, does a reestablish. Before this, the `reconnect_tx` has been validated.
* In order to support reestablish from a reconnect request, the FSM opts are also cached together with the state (but only added when the cache instance is initiated, not updated each time.)

The goal is to be able to use `reconnect` for all sorts of reconnect/reestablish.

Conflicts with #2630 are expected.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -2659,8 +2659,9 @@ on_chain_id(D, SignedTx) ->
     {PubKey, D#data{on_chain_id = PubKey}}.
 
 initialize_cache(#data{ on_chain_id = ChId
-                      , state       = State } = D) ->
-    aesc_state_cache:new(ChId, my_account(D), State).
+                      , state       = State
+                      , opts        = Opts } = D) ->
+    aesc_state_cache:new(ChId, my_account(D), State, Opts).
 
 cache_state(#data{ on_chain_id = ChId
                  , state       = State } = D) ->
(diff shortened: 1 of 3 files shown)


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -3489,27 +3490,31 @@ close_(Reason, D) ->
...
-            keep_state(D2, [{reply, From, ok}]);
-        {error, _} = Err ->
-            lager:debug("Request failed: ~p", [Err]),
-            keep_state(D, [{reply, From, Err}])
-    catch
-        error:E ->
-            lager:debug("CAUGHT ~p / ~p", [E, erlang:get_stacktrace()]),
-            keep_state(D, [{reply, From, E}])
+handle_call(_, {?RECONNECT_CLIENT, Pid, Tx} = Msg, From, #data{} = D0) ->
+    case D0 of
```
tolbrino: I'd rather have these 2 case clauses as function clauses.

apps/aechannel/src/aesc_state_cache.erl
```diff
@@ -36,15 +36,15 @@
              mons_ref  = ets:new(mons_ref, [set]),
              watchers  = ets:new(watchers, [set]),
              min_depth = ?MIN_DEPTH}).
--record(ch, {id, state}).
--record(pch, {id, pubkeys = [], state}).
+-record(ch, {id, state, opts}).
```
tolbrino: Can you document the semantics and use of `opts` throughout the cache?

apps/aechannel/src/aesc_state_cache.erl
```diff
@@ -194,11 +195,15 @@ move_state_to_persistent(ChId) ->
               ?TAB, [{ #ch{id = key(ChId, '_'), _ = '_'}, [], ['$_'] }]),
     case Found of
         [] -> ok;
-        [#ch{id = {_,PKa}, state = A}, #ch{id = {_,PKb}, state = A}] ->
-            Pch = #pch{id = ChId, pubkeys = [PKa, PKb], state = A},
+        [#ch{id = {_,PKa}, state = A, opts = Oa},
```
tolbrino: Can you add some high-level comments to these pattern matches? The detail gets easily lost when scanning the code.

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -23,6 +25,38 @@
 -define(ERROR_TO_CLIENT(Err), {?MODULE, send_to_client, {error, Err}}).
 
 
+%% ===========================================================================
+%% API to check if session is potentially hanging, waiting for socket close
```
tolbrino: 👍 Nice docs, just forgot to add `@doc`.

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -62,35 +96,85 @@ websocket_init(Params) ->
...
+        error ->
+            OnError(not_found)
+    end.
+
+check_existing_client(Client, Opts, Handler) ->
+    OnError = fun(E) ->
+                      handler_init_error(E, Handler)
+              end,
+    case time_since_last_dispatch(Client) of
+        T when is_integer(T), T < 5000 ->
```
tolbrino: The threshold should be defined somewhere at the top and definitely be documented.

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -62,35 +96,85 @@ websocket_init(Params) ->
...
+            handler_init_error(client_still_active, Handler);
+        T when is_integer(T) ->
+            %% It is actually a WS client
+            MRef = erlang:monitor(process, Client),
+            exit(Client, kill),
+            receive {'DOWN', MRef, _, _, _} ->
+                    timer:sleep(100),
```
tolbrino: Same as above. Otherwise the value seems somewhat arbitrary.

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -62,35 +96,85 @@ websocket_init(Params) ->
...
+                    reconnect_to_fsm_(Opts, Handler, OnError)
+            end;
+        _ ->
+            timer:sleep(100),
```
tolbrino: Same as above.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -3489,27 +3490,31 @@ close_(Reason, D) ->
...
+                    MRef = erlang:monitor(process, Pid),
+                    D2 = D1#data{ client           = Pid
+                                , client_mref      = MRef
+                                , client_connected = true },
+                    keep_state(D2, [{reply, From, ok}]);
+                {error, _} = Err ->
+                    lager:debug("Request failed: ~p", [Err]),
+                    keep_state(D, [{reply, From, Err}])
+            catch
+                error:E ->
```
grzegorz225:
```suggestion
                error:E:Stacktrace ->
```

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -3489,27 +3490,31 @@ close_(Reason, D) ->
...
+                    lager:debug("CAUGHT ~p / ~p", [E, erlang:get_stacktrace()]),
```
grzegorz225:
```suggestion
                    lager:debug("CAUGHT ~p / ~p", [E, Stacktrace]),
```

apps/aechannel/src/aesc_state_cache.erl
```diff
@@ -36,15 +36,15 @@
              mons_ref  = ets:new(mons_ref, [set]),
              watchers  = ets:new(watchers, [set]),
              min_depth = ?MIN_DEPTH}).
--record(ch, {id, state}).
--record(pch, {id, pubkeys = [], state}).
+-record(ch, {id, state, opts}).
+-record(pch, {id, pubkeys = [], state, opts = #{}}).
```
grzegorz225: This is backwards incompatible, a db migration is needed.

tolbrino:
@uwiger Can we close this?

PR #2713, opened 2019-08-27
Pt 168132312 new name hash

ThomasArts:
From that we accept the .aet domain, we use a different commit hash and name hash.

This is based upon PR 2704...

### DIFF
--- a/apps/aens/src/aens_hash.erl
+++ b/apps/aens/src/aens_hash.erl
@@ -60,7 +60,7 @@ name_hash(NameAscii) ->
 
 -spec pre_lima_commitment_hash(binary(), integer()) -> commitment_hash().
 pre_lima_commitment_hash(NameAscii, Salt) ->
-    NameHash = name_hash(NameAscii),
+    NameHash = pre_lima_name_hash(NameAscii),
     SaltBin = int_to_bin(Salt),
     hash(<<NameHash/binary, SaltBin/binary>>).
 
--- a/apps/aens/src/aens_hash.erl
+++ b/apps/aens/src/aens_hash.erl
@@ -2,6 +2,12 @@
 %%% @copyright (C) 2018, Aeternity Anstalt
 %%% @doc
 %%%    Naming System hashing functions
+%%%
+%%%    We hash names in .test domain in a legacy format, whereas names
+%%%    in the .aet domain are hashed differently.
+%%%    If additional domains are added, we need to make their hash function
+%%%    explicit in this code.
+%%%
 %%% @end
 %%%=============================================================================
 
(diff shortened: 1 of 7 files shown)


apps/aens/src/aens_hash.erl
```diff
@@ -28,14 +34,37 @@
 %%% API
 %%%===================================================================
 
+
 -spec commitment_hash(binary(), integer()) -> commitment_hash().
 commitment_hash(NameAscii, Salt) ->
+    case aens_utils:name_domain(NameAscii) of
+        {ok, Domain} when Domain =:= <<"aet">> ->
```
skkw: This pulling of domain (split binary, reverse list, check first elem) seems rather inefficient, but it's probably not an issue in reality.

Could be made faster and cleaner though:

```
is_registrar(<<Name/binary>>, <<Registrar/binary>>) when size(Name) > size(Registrar) ->
    Reg = <<".", Registrar/binary>>,
    RegSize = size(Reg),
    binary:part(Name, size(Name) - RegSize, RegSize) == Reg;
is_registrar(_, _) ->
    false.
```

apps/aens/src/aens_hash.erl
```diff
@@ -28,14 +34,37 @@
...
+            hash(<<NameAscii/binary, SaltBin/binary>>);
+        _ ->
+            %% This could be .test or any other wrong name backward compatible
+            pre_lima_commitment_hash(NameAscii, Salt)
+    end.
+
+-spec name_hash(binary()) -> name_hash().
+name_hash(NameAscii) ->
+    case aens_utils:name_domain(NameAscii) of
+        {ok, Domain} when Domain =:= <<"aet">> ->
```
skkw: same here

apps/aens/src/aens_hash.erl
```diff
@@ -28,14 +34,37 @@
...
+            hash(NameAscii);
+        _ ->
+            %% This could be .test or any other wrong name backward compatible
+            pre_lima_name_hash(NameAscii)
+    end.
+
+
+-spec pre_lima_commitment_hash(binary(), integer()) -> commitment_hash().
+pre_lima_commitment_hash(NameAscii, Salt) ->
     NameHash = name_hash(NameAscii),
```
skkw: Shouldn't this call `pre_lima_name_hash` instead?

apps/aens/src/aens_hash.erl
ThomasArts: No need for efficieny here... clear code prefered

apps/aens/src/aens_hash.erl
ThomasArts: it will automatically take that one, but sure, that could save a function call. Well spotted, shall fix it for next rebase

apps/aens/src/aens_hash.erl
```diff
@@ -28,14 +34,38 @@
...
 %%%===================================================================
 
+
 -spec commitment_hash(binary(), integer()) -> commitment_hash().
 commitment_hash(NameAscii, Salt) ->
-    NameHash = name_hash(NameAscii),
-    SaltBin = int_to_bin(Salt),
-    hash(<<NameHash/binary, SaltBin/binary>>).
+    case aens_utils:name_domain(NameAscii) of
+        {ok, Domain} when Domain =:= <<"aet">> ->
```
hanssv: Shouldn't it be related to `aec_governance` somehow?

docs/release-notes/RELEASE-NOTES-5.0.0-rc.1.md
```diff
@@ -33,6 +33,8 @@ Changelog:
   affect the offchain state trees. This is an incompatible change in the serialization of offchain
   updates, but old offchain updates can still be deserialized. When creating a channel with a party
   running an older node version, add `version_offchain_update=1` to the channel params.
+  * Name preclaims are no longer allowed to use 0 as Salt value
```
hanssv: You can't backport the release notes...

apps/aens/src/aens_hash.erl
ThomasArts: Yes, indeed.

PR #2716, opened 2019-08-27
PT-168026587 GA in dry run

hanssv:
Improve dry-run - handle GA transactions, etc

- Add GAAttachTx
- Allow normal (unsigned) transactions from generalized accounts
- Add DryRunCallRequest - where you can pass a subset of contract call parameters and set the context (i.e. Auth context).


py/hack_dry_run_input.py
```diff
@@ -0,0 +1,176 @@
+# coding: utf-8
+
```
dincho: 😢

apps/aecore/src/aec_dry_run.erl
```diff
@@ -64,8 +78,65 @@ add_account(#{pub_key := PK, amount := A}, AccountsTree) ->
...
+        Context    = TxHashCtxt ++ StateCtxt,
+        {tx, Context, dummy_sign(CallTx)}
+    catch _:_R ->
+        error({bad_dry_run_call_request, ReqMap})
+    end.
+
+prepare_env(Env, Opts) ->
+    case proplists:get_value(auth_tx, Opts, undefined) of
+        undefined -> Env;
+        TxHash    -> aetx_env:set_ga_tx_hash(Env, TxHash)
```
ThomasArts: Would it add value to check that TxHash is a binary hash?

apps/aeprimop/src/aeprimop.erl
```diff
@@ -507,6 +509,12 @@ inc_account_nonce({Pubkey, Nonce, Force}, #state{} = S) ->
                 _ ->
                     Account1 = aec_accounts:set_nonce(Account, Nonce),
                     put_account(Account1, S1)
+            end;
+        false when DryRun andalso AccType == generalized ->
```
ThomasArts: What if we are in Dryrun and AccType == basic?

hanssv: Then we do the same checks as if we are not dry-running.

apps/aecore/src/aec_dry_run.erl
hanssv: It comes from `aeser_api_encoder:safe_decode/2` so seems like overkill

PR #2717, opened 2019-08-28
Generic GAMetaTx error message

hanssv:


hanssv:
Already taken care of elsewhere.

PR #2721, opened 2019-08-28
Generic errors in meta TX

hanssv:


apps/aega/src/aega_meta_tx.erl
```diff
@@ -204,11 +204,11 @@ process(#ga_meta_tx{} = Tx, Trees, Env0) ->
...
 set_meta_result(ok, _Tx, Trees, _Env) ->
     Trees;
-set_meta_result(Err = {error, _}, Tx, Trees, Env) ->
-    %% ct:pal("Setting error: ~p\n", [Err]),
-    SetInstructions =
-        aeprimop:ga_set_meta_tx_res_instructions(
-            ga_pubkey(Tx), auth_data(Tx), Err),
+set_meta_result({error, Reason0}, Tx, Trees, Env) ->
+    Reason = inner_transaction_failed,
+    lager:info("Inner transaction failed with reason: ~p", [Reason0]),
```
tolbrino: Shouldn't this be a debug log? A node operator should normally not be interested in the error.

hanssv: Yes...

PR #2722, opened 2019-08-29
Fix Call.caller and Call.origin in the context of Forced progress

velzevur:
[Forced Progress context of `Caller`primops](https://www.pivotaltracker.com/story/show/166477826)

Off-chain transactions calls and Force progress transactions have a bug: the result of a `Call.origin` or `Call.caller` calls is not the caller of the contract but rather the owner of the contract. This PR fixes it in a backwards compatible manner: the fix kicks in from Lima hard fork on.

TODOs:
- [ ]  Add off-chain contract call tests
- [ ]  Add a release notes line

### DIFF
--- a/apps/aechannel/src/aesc_offchain_update.erl
+++ b/apps/aechannel/src/aesc_offchain_update.erl
@@ -85,6 +85,8 @@
 -export([type2swagger_name/1]).
 -endif.
 
+-include_lib("aecontract/include/hard_forks.hrl").
+
 -spec from_db_format(update() | tuple()) -> update().
 from_db_format(#transfer{} = U) ->
     U;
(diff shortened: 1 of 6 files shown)


apps/aechannel/src/aesc_offchain_update.erl
```diff
@@ -208,15 +210,28 @@ apply_on_trees(Update, Trees0, OnChainTrees, OnChainEnv, Round, Reserve) ->
...
                        call_data = CallData, call_stack = CallStack,
                        gas_price = GasPrice, gas = Gas} ->
-            Caller = account_pubkey(CallerId),
+            Height = aetx_env:height(OnChainEnv),
             ContractPubKey = contract_pubkey(ContractId),
-            Trees1 = remove_tokens(Caller, Amount, Trees0, Reserve),
+            Trees1 = remove_tokens(account_pubkey(CallerId), Amount, Trees0, Reserve),
             Trees2 = add_tokens(ContractPubKey, Amount, Trees1),
             Call = aect_call:new(CallerId, Round, ContractId, Round, GasPrice),
+            PostLima =
```
grzegorz225: 1. What should happen if a pre-lima deployed contract is called after lima.
2. What should happen if a post lima deployed contract calls an pre lima deployed contract.
3. What should happen if a pre lima deployed contract calls an post lima deployed contract.
4. How many existing channels https://roma-net.mdw.aepps.com/channels are currently using smart contracts ;) ?

tolbrino:
Fixed and merged as part of #2726

PR #2726, opened 2019-08-29
SC Call.origin and Call.caller cleanup and tests

velzevur:
[PT-166477826](https://www.pivotaltracker.com/story/show/166477826)

Off-chain transactions calls and Force progress transactions have a bug: the result of a `Call.origin` or `Call.caller` calls is not the caller of the contract but rather the owner of the contract.

### DIFF
--- a/apps/aecontract/src/aect_channel_contract.erl
+++ b/apps/aecontract/src/aect_channel_contract.erl
@@ -4,7 +4,7 @@
 
 -export([new/6,
          run_new/6,
-         run/11,
+         run/12,
          get_call/4,
          insert_failed_call/6
         ]).
@@ -38,7 +38,7 @@ run_new(ContractPubKey, Call, CallData0, Trees0, OnChainTrees,
     CallData = assert_init_function(CallData0, VmVersion, Code),
     {Contract, Trees1} = prepare_init_call(VmVersion, Contract0, Trees0),
     Store = aect_contracts:state(Contract),
-    CallDef = make_call_def(OwnerPubKey, ContractPubKey,
+    CallDef = make_call_def(OwnerPubKey, OwnerPubKey, ContractPubKey,
                             _Gas = 1000000, _GasPrice = 1,
                             _Amount = 0, %TODO: make this configurable
                             CallData, CallStack, Code, Store, Call, OnChainTrees, OnChainEnv, Trees1),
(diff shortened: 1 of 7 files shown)


apps/aecontract/src/aect_channel_contract.erl
```diff
@@ -120,7 +119,7 @@ run(ContractPubKey, ABIVersion, Call, CallData, CallStack, Trees0,
         true                                 -> erlang:error(wrong_vm_version);
         false                                -> erlang:error(wrong_abi_version)
     end,
-    CallDef = make_call_def(OwnerPubKey, ContractPubKey, Gas, GasPrice, Amount,
+    CallDef = make_call_def(CallerPubkey, ContractPubKey, Gas, GasPrice, Amount,
```
hanssv: I think `make_call_def` is using `OwnerPubKey`?

apps/aecontract/src/aect_channel_contract.erl
```diff
@@ -145,7 +145,7 @@ make_call_def(OwnerPubKey, ContractPubKey, GasLimit, GasPrice, Amount,
     , tx_env          => OnChainEnv
     , off_chain       => true
     , on_chain_trees  => OnChainTrees
-    , origin          => OwnerPubKey
+    , origin          => CallerPubkey
```
hanssv: Is this always true? Looks strange...

grzegorz225: I'm unsure how calls from inside an contract are handled.

docs/release-notes/RELEASE-NOTES-4.2.1.md
```diff
@@ -4,6 +4,7 @@
 It includes the following changes:
 * Change to a generic error message for the inner transaction of a GAMetaTX.
 * Long overdue bump of info-field in key block header from pre-fortuna value.
+* Improves State Channels tests to cover `Call.origin` and `Call.caller`
```
grzegorz225: Shouldn't this line summarize the changes instead of referencing tests?

apps/aecontract/src/aect_channel_contract.erl
hanssv: I think it is only the top-level call that enters `aect_channel_contract:run` (there is only one place in the code where this function is called) so this is correct. But I will double check.

docs/release-notes/RELEASE-NOTES-4.2.1.md
hanssv: I agree

tolbrino:
```suggestion
* Fix incorrect setting of `Call.origin` and `Call.caller` in contract calls from state channels
```

hanssv: off-chain contract calls?

PR #2734, opened 2019-09-02
Fix SC system tests

velzevur:
I remember fixing this but I could not find it anywhere

### DIFF
--- a/apps/aechannel/src/aesc_session_noise.erl
+++ b/apps/aechannel/src/aesc_session_noise.erl
@@ -107,9 +107,8 @@ close(Session) ->
             unlink(Session),
             ok;
         exit:R ->
-            lager:error("CAUGHT exit:~p, ~p", [R, erlang:get_stacktrace()]),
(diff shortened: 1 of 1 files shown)


apps/aechannel/src/aesc_session_noise.erl
```diff
@@ -107,9 +107,8 @@ close(Session) ->
             unlink(Session),
             ok;
         exit:R ->
-            lager:error("CAUGHT exit:~p, ~p", [R, erlang:get_stacktrace()]),
+            lager:debug("CAUGHT exit:~p, ~p", [R, erlang:get_stacktrace()]),
```
tolbrino:
```suggestion
            lager:error("Unhandled exit error during session closing: ~p, ~p", [R, erlang:get_stacktrace()]),
```

tolbrino:
> Can't remember the details, but maybe `R` is significant. If it isn't `normal` or `shutdown` it might actually be an error?

I agree it should be logged at least.

PR #2736, opened 2019-09-03
PT-168256282 Disallow calls to init

hanssv:
https://github.com/aeternity/aesophia/pull/139
https://github.com/aeternity/aebytecode/pull/75

### DIFF
--- a/apps/aecontract/test/aecontract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_SUITE.erl
@@ -141,6 +141,7 @@
         , sophia_bits/1
         , sophia_aevm_bad_code/1
         , sophia_aevm_bad_init/1
+        , sophia_no_calls_to_init/1
         , sophia_int_to_str/1
         , sophia_events/1
         , sophia_crypto/1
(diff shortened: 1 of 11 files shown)


apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -806,7 +808,6 @@ create_contract_(ContractCreateTxGasPrice) ->
     %% Check that the created contract has the correct details from the contract create tx
     ?assertEqual(PubKey, aect_contracts:owner_pubkey(Contract)),
     ?assertEqual(aect_create_tx:vm_version(aetx:tx(Tx)), aect_contracts:vm_version(Contract)),
-    ?assertEqual(aect_create_tx:code(aetx:tx(Tx)), aect_contracts:code(Contract)),
```
UlfNorell: The code on chain and the code in the contract is no longer the same. I see.

UlfNorell:
We could (should) provide a function in the compiler that checks if a given on-chain binary was produced by compiling some given source code. This could easily take care of the `init` stripping.

What I like about removing the init function completely is that we're not relying on the compiler to forbid local calls to `init` (which it does, but there could be bugs). Since FATE `init` now really does write the state it would be catastrophic if there were any way to have it run after contract creation.

PR #2725, opened 2019-08-29
SC user abort & recoverable errors

uwiger:
See [#168160333](https://www.pivotaltracker.com/story/show/168160333)

First commit (WIP). Old test suites seem to pass, but no new tests added yet.

Approach:
* Allow client to answer a signing request with an `{error: ErrorCode}` parameter
* Timeouts during ongoing updates/deposits/... lead to an appropriate error message (`error_code: ?ERR_TIMEOUT`), reverting to stable offchain state and moving to the `open` FSM state
* Reserve error codes >= 128 to users. The `error_code` field is two bytes.

PR #2737, opened 2019-09-03
Improve client protocol error handling in SC FSMs

uwiger:
New PR since there are quite a few changes to the old one (#2725).

Fixes tickets:

[PT-168160333](https://www.pivotaltracker.com/story/show/168160333)
[PT-159425103](https://www.pivotaltracker.com/story/show/159425103)
[PT-167532107](https://www.pivotaltracker.com/story/show/167532107)
[PT-164144646](https://www.pivotaltracker.com/story/show/164144646)
[PT-167944617](https://www.pivotaltracker.com/story/show/167944617)

This PR introduces the ability to reject a signing request with an `{error: Code}` object, where `Code` is an integer. Some error codes have been added, and the range `128...65535` have been reserved for user-defined (application-level) errors. Also, the flow for close-mutual (shutdown) has been extended so that the FSMs wait for on-chain confirmation before terminating. This also creates room for aborting/rejecting a shutdown request. For this purpose, a `shutdown_error` protocol message has been added.

### DIFF
--- a/apps/aechannel/src/aesc_chain_watcher.erl
+++ b/apps/aechannel/src/aesc_chain_watcher.erl
@@ -1,9 +1,9 @@
 -module(aesc_chain_watcher).
 -behaviour(gen_server).
 
--export([start_link/5,       %% (TxHash, ChanId, MinimumDepth, Mod) -> {ok, Pid}
+-export([start_link/2,       %% (ChanId, Mod) -> {ok, Pid}
+         start_link/5,       %% (TxHash, ChanId, MinimumDepth, Mod) -> {ok, Pid}
          watch/5,            %% (WatcherPid, Type, TxHash, MinimumDepth, Mod) -> ok
-         restart_watcher/2,  %% (ChanId, Mod) -> {ok, Pid}
          watch_for_channel_close/3,
          watch_for_unlock/2,
          watch_for_min_depth/5]).
@@ -190,6 +190,16 @@ get_txs_since_(StopCond, ChId) ->
     get_txs_since(StopCond, aec_chain:top_block_hash(), ChId, #{}).
 
 
+start_link(ChanId, Mod) ->
+    I = #{ callback_mod => Mod
+         , parent       => self() },
+    Reqs = [#{ mode => watch
+             , info => I#{ type => watch } }],
+    gen_server:start_link(?MODULE, #{ parent  => self()
+                                    , chan_id => ChanId
+                                    , requests => Reqs },
+                          ?GEN_SERVER_OPTS).
+
 start_link(Type, TxHash, ChanId, MinDepth, Mod) ->
     I = #{ parent       => self()
          , type         => Type
@@ -203,16 +213,6 @@ start_link(Type, TxHash, ChanId, MinDepth, Mod) ->
     gen_server:start_link(?MODULE, #{parent   => self(),
                                      chan_id  => ChanId,
                                      requests => Reqs},
-                          ?GEN_SERVER_OPTS).
-
-restart_watcher(ChanId, Mod) ->
-    I = #{ callback_mod => Mod
-         , parent       => self() },
-    Reqs = [#{ mode => watch
-             , info => I#{ type => watch } }],
-    gen_server:start_link(?MODULE, #{ parent  => self()
-                                    , chan_id => ChanId
-                                    , requests => Reqs },
                           ?GEN_SERVER_OPTS).
 
 watch(Watcher, Type, TxHash, MinDepth, Mod) ->
--- a/apps/aechannel/src/aesc_chain_watcher.erl
+++ b/apps/aechannel/src/aesc_chain_watcher.erl
@@ -3,6 +3,7 @@
 
 -export([start_link/5,       %% (TxHash, ChanId, MinimumDepth, Mod) -> {ok, Pid}
          watch/5,            %% (WatcherPid, Type, TxHash, MinimumDepth, Mod) -> ok
+         restart_watcher/2,  %% (ChanId, Mod) -> {ok, Pid}
          watch_for_channel_close/3,
          watch_for_unlock/2,
          watch_for_min_depth/5]).
@@ -202,6 +203,16 @@ start_link(Type, TxHash, ChanId, MinDepth, Mod) ->
     gen_server:start_link(?MODULE, #{parent   => self(),
                                      chan_id  => ChanId,
                                      requests => Reqs},
+                          ?GEN_SERVER_OPTS).
+
+restart_watcher(ChanId, Mod) ->
+    I = #{ callback_mod => Mod
+         , parent       => self() },
+    Reqs = [#{ mode => watch
+             , info => I#{ type => watch } }],
+    gen_server:start_link(?MODULE, #{ parent  => self()
+                                    , chan_id => ChanId
+                                    , requests => Reqs },
                           ?GEN_SERVER_OPTS).
 
 watch(Watcher, Type, TxHash, MinDepth, Mod) ->
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -72,7 +72,8 @@
 -export([message/2]).
 
 %% Used by client
--export([signing_response/3]).          %% (Fsm, Tag, Obj)
+-export([ signing_response/3          %% (Fsm, Tag, Obj)
+        , error_code_to_msg/1]).      %% (Code)
 
 %% Used by min-depth watcher
 -export([ minimum_depth_achieved/4      %% (Fsm, OnChainId, Type, TxHash)
@@ -367,6 +368,15 @@ message(Fsm, {T, _} = Msg) when ?KNOWN_MSG_TYPE(T) ->
 signing_response(Fsm, Tag, Obj) ->
     gen_statem:cast(Fsm, {?SIGNED, Tag, Obj}).
 
+error_code_to_msg(?ERR_VALIDATION) -> <<"validation error">>;
+error_code_to_msg(?ERR_CONFLICT  ) -> <<"conflict">>;
+error_code_to_msg(?ERR_TIMEOUT   ) -> <<"timeout">>;
+error_code_to_msg(?ERR_ABORT     ) -> <<"abort">>;
+error_code_to_msg(Code) when Code >= ?ERR_USER ->
+    <<"user-defined">>;
+error_code_to_msg(_) ->
+    <<"unknown">>.
+
 %% ==================================================================
 %% Used by min-depth watcher
 
@@ -573,8 +583,9 @@ awaiting_reestablish(cast, {?CH_REESTABL, Msg}, #data{role = responder} = D) ->
     case check_reestablish_msg(Msg, D) of
         {ok, D1} ->
             report(info, channel_reestablished, D1),
-            gproc_register(D1),
-            next_state(open, send_reestablish_ack_msg(D1));
+            D2 = maybe_restart_watcher(D1),
+            gproc_register(D2),
+            next_state(open, send_reestablish_ack_msg(D2));
         {error, _} = Error ->
             close(Error, D)
     end;
@@ -603,8 +614,9 @@ awaiting_reestablish(Type, Msg, D) ->
 awaiting_signature(enter, _OldSt, _D) -> keep_state_and_data;
 awaiting_signature(cast, {?SIGNED, Tag, {error, Code}} = Msg,
                    #data{op = #op_sign{ tag = Tag }} = D) ->
-    D1 =log(rcv, msg_type(Msg), Msg, D),
-    handle_recoverable_error(#{code => Code}, D1);
+    lager:debug("Error code as signing reply (Tag = ~p): ~p", [Tag, Code]),
+    D1 = log(rcv, msg_type(Msg), Msg, D),
+    handle_recoverable_error(#{code => Code, msg_type => Tag}, D1);
 awaiting_signature(cast, {Req, _} = Msg, #data{ongoing_update = true} = D)
   when ?UPDATE_CAST(Req) ->
     %% Race detection!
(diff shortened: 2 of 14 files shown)


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2595,10 +2716,25 @@ has_my_signature(Me, SignedTx) ->
...
+            T = erlang:get_stacktrace(),
+            lager:debug("CAUGHT E=~p / T = ~p", [E, pr_stacktrace(T)]),
+            error({caught,E})
+    end.
+start_min_depth_watcher_(Type, SignedTx, Updates,
                         #data{ watcher = Watcher0
                              , op = Op
-                             , opts = #{minimum_depth := MinDepth}} = D) ->
+                             , opts = Opts } = D) ->
+    MinDepth = maps:get(minimum_depth, Opts, ?MINIMUM_DEPTH),
```
uwiger: This may seem odd... The purpose for this was to avoid a `function_clause` error in one of the `aest_db_SUITE` system test runs.

uwiger:
Perhaps also [#164144646](https://www.pivotaltracker.com/story/show/164144646) can be said to be addressed in this PR.

apps/aechannel/src/aesc_chain_watcher.erl
```diff
@@ -204,6 +205,16 @@ start_link(Type, TxHash, ChanId, MinDepth, Mod) ->
                                      requests => Reqs},
                           ?GEN_SERVER_OPTS).
 
+restart_watcher(ChanId, Mod) ->
```
velzevur: So this is exactly like `start_link/4` without the initial transaction to be tracked?

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2595,10 +2716,25 @@ has_my_signature(Me, SignedTx) ->
             ok =:= aetx_sign:verify_one_pubkey(Me, SignedTx, CurrHeight)
     end.
 
-start_min_depth_watcher(Type, SignedTx, Updates,
+
+maybe_restart_watcher(#data{ watcher = undefined
```
velzevur: I like the assertive approach but possibly drop the `maybe_` prefix as it is a bit misleading: we either start a new watcher or we crash.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -759,8 +777,8 @@ awaiting_signature(cast, {?SIGNED, ?SHUTDOWN_ACK, SignedTx} = Msg,
             D1 = send_shutdown_ack_msg(SignedTx, D),
             D2 = D1#data{ op = ?NO_OP
                         , log = log_msg(rcv, ?SIGNED, Msg, D1#data.log)},
-            report_on_chain_tx(close_mutual, SignedTx, D1),
-            close(close_mutual, D2)
+            report_on_chain_tx(close_mutual, SignedTx, D2),
```
velzevur: Nice catch

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -866,14 +884,30 @@ channel_closed(cast, {?CHANNEL_CHANGED, _Info} = Msg, D) ->
...
 
+mutual_closed(enter, _, D) ->
+    keep_state(enter_closing(D));
+mutual_closed(cast, {?MIN_DEPTH_ACHIEVED, ChainId, ?WATCH_CLOSED, TxHash},
+              #data{ on_chain_id = ChainId
+                   , op = #op_min_depth{ tag = ?WATCH_CLOSED
+                                       , tx_hash = TxHash  % same tx hash
+                                         }} = D) ->
+    close(close_confirmed, D);
+mutual_closed(cast, {?SHUTDOWN_ERR, Msg}, D) ->
```
velzevur: I think this opens an attack vector (Bob is malicious):
1. Alice and Bob co-sign a mutual close, post it on-chain and enter `mutual_closed` state
2. Bob being malicious sends a `?SHUTDOWN_ERR` message to Alice and she rolls back to `open` state
3. Alice and Bob continue making off-chain updates. Bob spends some tokens and now Bob has less tokens than he did in step 1
4. Since the on-chain transaction had been posted on-chain - it gets included with the balances posted in step 1

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -990,15 +1034,31 @@ half_signed(Type, Msg, D) ->
     handle_common_event(Type, Msg, error_all, D).
 
 mutual_closing(enter, _OldSt, _D) -> keep_state_and_data;
+mutual_closing(cast, {?SHUTDOWN_ERR, Msg}, D) ->
```
velzevur: Possibly related to [previous comment](https://github.com/aeternity/aeternity/pull/2737/files#r320393959) for an attack vector. In this case malicious Bob keeps the close mutual Alice solo-authenticated and Bob can co-authenticate it and post it at any point now. Only protection Alice has is posting a new transaction (of any type) on-chain thus invalidating the `nonce` in the `channel_close_mutual` transaction Bob holds.

velzevur: Although this will not work for the specific case of `generalized` accounts that don't have a meaningful replay-attack protection, I don't think this is a realistic issue

apps/aechannel/src/aesc_chain_watcher.erl
tolbrino: How about making this `start_link/2`?

apps/aehttp/src/sc_ws_api_jsonrpc.erl
```diff
@@ -629,3 +646,14 @@ check_params(<<"channels.update.new">>, Params) when is_map(Params) ->
                                , mandatory => false})]);
 check_params(Method, Params) ->
     {error, {unknown, Method, Params}}.
+
+valid_error_code(ErrorCode) when is_integer(ErrorCode) ->
+    ErrorCode >= 1 andalso ErrorCode =< 16#ffFF;
```
tolbrino: Why are you mixing upper and lower case?

velzevur:
I pushed thee branch to `system-tests` in order to run `docker_system_tests` job

apps/aehttp/src/sc_ws_api_jsonrpc.erl
uwiger: > Why are you mixing upper and lower case?

I find that it's easier to count the number of letters then. Perhaps overkill when it's only four of them. :)

uwiger:
This PR should also address [#167944617](https://www.pivotaltracker.com/story/show/167944617)

apps/aechannel/src/aesc_fsm.erl
tolbrino: I will change that later too. So not a problem.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -4006,8 +4168,30 @@ block_hash_from_op(#op_min_depth{data = #op_data{block_hash = BlockHash}}) ->
...
 block_hash_from_op(?NO_OP) -> %% in case of unexpected close
     ?NOT_SET_BLOCK_HASH.
 
 tx_env_and_trees_from_top(Type) ->
     aesc_tx_env_cache:tx_env_and_trees_from_top(Type).
+
+%% This function is meant to prune and prettify stack traces
+%% containing #data{} records, which often are so large that printouts become
+%% unreadable.
+pr_stacktrace(T) ->
```
tolbrino: Why not use `lager:pr_stacktrace/2` instead?

uwiger: It doesn't do quite the same thing. For one thing, it inserts line breaks, which messes with the Datadog side, and the main feature of my `pr_stacktrace/1` is that it removes some bulky record elements, which lager obviously can't do on its own.

apps/aechannel/src/aesc_fsm.erl
uwiger: > Possibly related to [previous comment](https://github.com/aeternity/aeternity/pull/2737/files#r320393959) for an attack vector. In this case malicious Bob keeps the close mutual Alice solo-authenticated and Bob can co-authenticate it and post it at any point now. Only protection Alice has is posting a new transaction (of any type) on-chain thus invalidating the `nonce` in the `channel_close_mutual` transaction Bob holds.

I'm not sure how to automatically protect against that, but Alice _will_ get notified about the situation and could possibly glean that Bob now may have a co-signed shutdown tx armed and ready. If it appears on-chain later, the watcher will at least detect it.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2015,6 +2058,28 @@ check_withdraw_error_msg(Msg, D) ->
...
+        #op_close{data = #op_data{signed_tx = SignedTx}} ->
+            TxHash = aetx_sign:hash(SignedTx),
+            case aec_chain:find_tx_location(TxHash) of
+                Block when is_binary(Block) ->
+                    %% Tx is already on chain - ignore error msg
+                    {error, tx_already_on_chain};
+                mempool ->
+                    %% For now, treat it as if it's on chain
+                    {error, tx_already_on_chain};
+                _ ->
```
velzevur:
```suggestion
                not_found->
                    {error, tx_not_found};
                none ->
```

I think it is worth differentiating between the heads: `aec_chain:find_tx_location/1` can return also `not_found|none` with different semantics:
* `not_found` means that the node has not yet seen this transaction. In this case either our node had not yet synced the posted transaction or it wasn't posted on-chain in the first place
* `none` means it was rejected as an invalid one (for ex. `nonce` didn't match). I think in this particular case it is OK to rollback to `open` state

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2024,7 +2089,7 @@ check_op_error_msg(Op, #{ channel_id := ChanId
     case aesc_offchain_state:get_fallback_state(State) of
         {LastRound, State1} -> %% same round
             Log1 = log_msg(rcv, Op, Msg, Log),
-            {ok, ErrCode, D#data{state = State1, log = Log1}};
+            {ok, ErrCode, D#data{state = State1, log = Log1, op = ?NO_OP}};
```
velzevur: I see this changes the semantics of `check_op_error_msg/3` a bit but this actually fixes a bug :)

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2043,7 +2108,8 @@ send_withdraw_created_msg(SignedTx, Updates,
            , data       => #{tx      => TxBin,
                              updates => UBins}},
     aesc_session_noise:wdraw_created(Sn, Msg),
-    log(snd, ?WDRAW_CREATED, Msg, Data).
+    log(snd, ?WDRAW_CREATED, Msg,
+        set_ongoing(?WDRAW_CREATED, Data)).
```
velzevur: All of those are nice catches

apps/aechannel/src/aesc_fsm.erl
velzevur: This is fixed now

apps/aechannel/src/aesc_fsm.erl
uwiger: So if it wasn't posted on on-chain in the first place, why would we reject the error message?

apps/aechannel/src/aesc_fsm.erl
velzevur: Since Alice was the author of the `channel_close_mutual` transaction, it is Alice who is the `origin` of the transaction. If Alice posts a new transaction on-chain (spend transaction to herself) with the same `nonce` and this gets included, the transaction that Bob now holds will be invalid and he will not be able to post it ever again. This will work with `basic` and `generalized` accounts alike if a replay-attack is present in the GA contract.

Not sure if this is part of the FSM in the first place.

docs/release-notes/next-lima/PT-168160333-improved-error-handling.md
```diff
@@ -0,0 +1,4 @@
+* State Channels: It is now possible to abort a signing request by replying with an error code.
```
tolbrino: Link to protocol documentation missing.

docs/release-notes/next-lima/PT-168160333-improved-error-handling.md
```diff
@@ -0,0 +1,4 @@
...
+* State Channels: The FSMs now stay alive until a Close Mutual (shutdown) has been confirmed on-chain
+  It is also possible to abort/reject a Close Mutual by rejecting the signing request.
+* State Channels: The chain watcher was not restarted after leave/reestablish. This has been fixed.
```
tolbrino: I'd rather describe which bug is fixed here. The chain watcher is an implementation detail.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2015,23 +2059,52 @@ check_withdraw_error_msg(Msg, D) ->
...
+check_shutdown_err_msg(#{ channel_id := ChanId } = Msg,
+                       #data{ on_chain_id = ChanId
+                            , op = Op } = D) ->
+    case Op of
+        #op_close{data = #op_data{signed_tx = SignedTx}} ->
+            TxHash = aetx_sign:hash(SignedTx),
+            case aec_chain:find_tx_location(TxHash) of
+                Block when is_binary(Block) ->
+                    %% Tx is already on chain - ignore error msg
+                    {error, tx_already_on_chain};
```
tolbrino: Debug logs for these 2 cases would be nice as this could help troubleshoot issue and/or usage patterns later.

apps/aechannel/src/aesc_fsm.erl
tolbrino: 👍

velzevur:
Looks good. I pushed to `system-tests` just to make sure we are safe

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2015,23 +2059,54 @@ check_withdraw_error_msg(Msg, D) ->
...
+            TxHash = aetx_sign:hash(SignedTx),
+            case aec_chain:find_tx_location(TxHash) of
+                Block when is_binary(Block) ->
+                    lager:debug("Tx already on-chain. Ignore error msg", []),
+                    {error, tx_already_on_chain};
+                mempool ->
+                    lager:debug("Tx in mempool. Treat as if already on-chain. Ignore error", []),
+                    {error, tx_already_on_chain};
+                not_found->
+                    lager:debug("Tx not found. Assume it may later appear on-chain. Ignore error", [])
```
tolbrino:
```suggestion
                    lager:debug("Tx not found. Assume it may later appear on-chain. Ignore error", []),
```

PR #2748, opened 2019-09-05
Swagger cleanup

dincho:
- Stop deploying swagger.json to API docs repo
- Use swagger.yaml instead of swagger.json to server /api endpoint
- Cleanup swagger spec (wrong required fields)
- Use Bravado dynamic swagger client instead of generated client


dincho:
> A nice side-effect is that you can also remove Java from the Windows CI setup.

I was hoping to remove it at all, but now we have that nice PlantUML diagrams 😞

PR #2749, opened 2019-09-05
Bump erlang patch version

dincho:
Follow-up of https://github.com/aeternity/docker-builder/pull/22


dincho:
Not waiting the windows builds as there is no related changes to it.

PR #2752, opened 2019-09-05
Pt 168245174 memory gas

happi:
Spend gas for memory usage (stack, tuples, strings and large integers).
Each used underlying machine word (64-bits), called a cell, cost 1 unit.
For each 1024 used cells (8 KiB) the price per unit increases by 1.
When the stack shrinks the number of used cells decreases, this does not give you back gas spent for using the stack cell but the price of new cells might go down.
Not all integer operations uses new cells and only operations that can use many cells counts towards the gas cost.

TODO: describe this model in the protocol documentation.

### DIFF
--- a/apps/aefate/src/aefa_engine_state.erl
+++ b/apps/aefate/src/aefa_engine_state.erl
@@ -365,20 +365,18 @@ spend_gas(X, #es{gas = Gas} = ES) ->
         false -> ES#es{gas = NewGas}
     end.
 
--define(CHEAP_CELLS, 10000). %% TODO: move to governance.
--define(CELL_COST_GROWTH, 0.6).
-
+%% The gas price per cell increases by 1 for each kibiword (each 1024 64-bit word) used.
 -spec spend_gas_for_new_cells(non_neg_integer(), state()) -> state().
-spend_gas_for_new_cells(NewCells, #es{ created_cells = Cells } = ES) when NewCells + Cells =< ?CHEAP_CELLS ->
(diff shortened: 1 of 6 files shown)


apps/aefate/src/aefa_engine_state.erl
```diff
@@ -344,6 +365,20 @@ spend_gas(X, #es{gas = Gas} = ES) ->
...
+    TotalCells = Cells + NewCells,
+    spend_gas(NewCells, ES#es{ created_cells = TotalCells });
+spend_gas_for_new_cells(1, #es{ created_cells = Cells } = ES) ->
+    TotalCells = Cells + 1,
+    CellCost = 1 + (TotalCells bsr 10),
+    spend_gas(CellCost, ES#es{ created_cells = TotalCells });
+spend_gas_for_new_cells(NewCells, #es{ created_cells = Cells } = ES) ->
+    TotalCells = Cells + NewCells,
+    CellCost = 1 + (TotalCells bsr 10),
+    spend_gas(NewCells * CellCost, ES#es{ created_cells = TotalCells }).
```
UlfNorell: Why special case for `NewCells = 1` and `NewCells + Cells =< 1024`? As far as I can see the general case does the right thing in both of these cases.

happi: just wanted to avoid a multiplication in the most used case.

happi: Premature optimizations to the rescue.

happi: The computation was much more convoluted in previous version, I can tidy it up in a later PR.

PR #2753, opened 2019-09-06
Use latest aesophia and aebytecode with new op numbers and gas.

happi:
Use new number of byte codes and new base gas for fate instructions.

### DIFF
--- a/apps/aehttp/test/aehttp_contracts_SUITE.erl
+++ b/apps/aehttp/test/aehttp_contracts_SUITE.erl
@@ -1155,7 +1155,7 @@ remote_gas_test_contract(Config) ->
     call_get(APub, APriv, EncC2Pub, Contract, 100),
     force_fun_calls(Node),
     Balance1 = get_balance(APub),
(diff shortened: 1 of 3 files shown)


apps/aehttp/test/aehttp_contracts_SUITE.erl
```diff
@@ -1181,7 +1181,7 @@ remote_gas_test_contract(Config) ->
     [] = call_func(APub, APriv, EncC1Pub, Contract, "call", [ZeroContract, "2", "1"], error),
     force_fun_calls(Node),
     Balance5 = get_balance(APub),
-    ?assertMatchVM(900000, 800017, (Balance4 - Balance5) div ?DEFAULT_GAS_PRICE),
+    ?assertMatchVM(900000, 800141, (Balance4 - Balance5) div ?DEFAULT_GAS_PRICE),
```
hanssv: So I see the FATE execution became 8 times as expensive...

happi: Yes. Base cost is now 10/instruction instead of 3. But we will reduce the base tx.fee. And it will still be much cheaper than aevm.

PR #2754, opened 2019-09-06
Make hash function on bytes hash the raw bytes

UlfNorell:
For AEVM we do this in the compiler (we don't have the size of the
byte array at runtime). PR: https://github.com/aeternity/aesophia/pull/146

### DIFF
--- a/apps/aecontract/test/aecontract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_SUITE.erl
@@ -157,6 +157,7 @@
         , sophia_address_checks/1
         , sophia_remote_gas/1
         , sophia_higher_order_state/1
+        , sophia_bignum/1
         , create_store/1
         , read_store/1
         , store_zero_value/1
@@ -284,7 +285,8 @@ all() ->
                        , sophia_polymorphic_entrypoint
                        ]).
 
--define(FATE_TODO, [sophia_state_gas_arguments
+-define(FATE_TODO, [sophia_state_gas_arguments,
+                    sophia_bignum       %% Dynamic gas for arithmetic ops!
                    ]).
 
 groups() ->
@@ -387,6 +389,7 @@ groups() ->
                                  sophia_bytes_to_x,
                                  sophia_address_checks,
                                  sophia_too_little_gas_for_mem,
+                                 sophia_bignum,
                                  sophia_higher_order_state
                                ]}
     , {sophia_oracles_ttl, [],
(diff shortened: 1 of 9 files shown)


apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -4579,6 +4582,52 @@ sophia_crypto(_Cfg) ->
...
+
+            ResSha3_B1    = ?call(call_contract, Acc, IdC, sha3_b17,    word, {Bytes1}),
+            ResSha256_B1  = ?call(call_contract, Acc, IdC, sha256_b17,  word, {Bytes1}),
+            ResBlake2b_B1 = ?call(call_contract, Acc, IdC, blake2b_b17, word, {Bytes1}),
+
+            ResSha3_B2    = ?call(call_contract, Acc, IdC, sha3_b52,    word, {Bytes2}),
+            ResSha256_B2  = ?call(call_contract, Acc, IdC, sha256_b52,  word, {Bytes2}),
+            ResBlake2b_B2 = ?call(call_contract, Acc, IdC, blake2b_b52, word, {Bytes2}),
+
+            ?assertMatchAEVM1OOG(Sha3_B1_N   , ResSha3_B1),
```
hanssv: could be just one sequence of `?assertMatchVM`?

test/contracts/sophia_2/crypto.aes
```diff
@@ -15,6 +15,14 @@ contract CryptoTest =
   function sha256  (x : complex) : hash = Crypto.sha256(x)
   function blake2b (x : complex) : hash = Crypto.blake2b(x)
 
+  function sha3_b17    (x : bytes(17)) : hash = Crypto.sha3(x)
```
hanssv: Is this one necessary? Didn't the tests exclude minerva?

UlfNorell: Huh, how does it compile on minerva in that case..? 🤔

apps/aecontract/test/aecontract_SUITE.erl
UlfNorell: Yeah I guess. I copied the existing pattern.

test/contracts/sophia_2/crypto.aes
hanssv: No, I just meant the bit with the newly added functions?

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -4633,6 +4655,52 @@ sophia_crypto(_Cfg) ->
     ?assertMatchFATE({bytes, Sha256},    ResSha256),
     ?assertMatchFATE({bytes, Blake2b},   ResBlake2b),
 
+    %% Tests for bytes (which should hash as raw strings from lima)
+
+    case vm_version() >= ?VM_AEVM_SOPHIA_3 of   %% No bytes in minerva
```
tolbrino: This looks like it could make use of `?skipRest` to reduce the nesting.

UlfNorell: There are tests below that shouldn't be skipped...

PR #2755, opened 2019-09-06
Charge gas for size of log entries in FATE

gorillainduction:
https://www.pivotaltracker.com/story/show/168340855

### DIFF
--- a/apps/aefate/src/aefa_fate_op.erl
+++ b/apps/aefate/src/aefa_fate_op.erl
@@ -771,6 +771,8 @@ log(Arg0, Arg1, Arg2, Arg3, Arg4, EngineState) ->
 log(Arg0, Arg1, Arg2, Arg3, Arg4, Arg5, EngineState) ->
     log_([Arg0, Arg1, Arg2, Arg3, Arg4, Arg5], EngineState).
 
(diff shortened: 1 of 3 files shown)


apps/aefate/src/aefa_fate_op.erl
```diff
@@ -771,14 +771,18 @@ log(Arg0, Arg1, Arg2, Arg3, Arg4, EngineState) ->
...
                (?FATE_BYTES(Bin))  -> Bin;
                (Other)             -> aefa_fate:abort({value_does_not_match_type, Other, string})
             end,
-    ToWord = fun(N) when is_integer(N) -> <<N:256>>;
+    ToWord = fun(N) when is_integer(N),
+                         N >= 0,
+                         N =< ?MAX_WORD_INT -> <<N:256>>;
                 (?FATE_FALSE)          -> <<0:256>>;
                 (?FATE_TRUE)           -> <<1:256>>;
                 (?FATE_BITS(N))        -> <<N:256>>;
```
UlfNorell: There can be more than 256 bits here.

gorillainduction: Good catch!

PR #2758, opened 2019-09-09
Let reconnect try reestablish (or kill idle client)

uwiger:
This PR is an evolution of #2709. Not all comments on that PR have been applied yet.
Remaining TODO:

- [ ] Fix failing `leave_reestablish_responder_stays()` test case (strange amounts inconsistency in the following `deposit_()`
- [ ] Write test cases for the new `reconnect` functionality

See 
* [#167752251](https://www.pivotaltracker.com/story/show/167752251) (SC: Harmonize client reconnect and reestablish)
* [#168073522](https://www.pivotaltracker.com/story/show/168073522) (re-connect should allow re-connect to sockets which appear to be alive.)
* [#167956796](https://www.pivotaltracker.com/story/show/167956796) (SC: leaking FSM and noise session)

The above issues seem interrelated, and this PR will attempt to address them all.
Main themes:
* The `reconnect` API method should detect if the fsm is not running, and then instead try to fetch the cached state and do a reestablish. For this to work, the connect options need to be cached as well, and the responder needs to actually be up and running, even after a `leave`.
* The responder staying after `leave` (if the `keep_running` option is set) is something that ought to be desired in the server scenario. The responder must then start a new acceptor if the initiator disconnects (either suddenly or via `leave`), and then be prepared for an `attach_responder` request followed by a `reestablish` while in the `open` state.

WIP let responder optionally keep running

WIP leave_reestablish_responder_stays() test


tolbrino:
This PR can be closed.

uwiger:
Yes, closing.

PR #2759, opened 2019-09-09
Refactor aesc_offchain_update versioning

velzevur:
PT [Improve off-chain update versioning](https://www.pivotaltracker.com/story/show/168370492)

This PR keeps the canonical serialization of Force Progress transactions:
```
(aeternity@localhost)3> aeser_api_encoder:encode(tx_hash, aetx_sign:hash(FP)).
<<"th_tDn4WaKAqweS2m7WbG3X2hBWfe4U1CmuFUVtk19uvdHJYAsNv">>
```

It takes heavy advantange of `aesc_offchain_update:update()` not being part of [what is accepted Lima hard fork on](https://github.com/aeternity/aeternity/blob/master/apps/aechannel/src/aesc_offchain_tx.erl#L242-L249) so if participants send each other `#meta{}` off-chain updates, those will not impact the off-chain transaction being produced and thus - pre-Lima meta updates will not impact the ability of users to use co-authenticated transactions in a Force Progress.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -65,7 +65,6 @@
         , record_fields/1
         , report_tags/0
         , timeouts/0
-        , version_tags/0
         , bh_deltas/0
         ]).
 
@@ -305,8 +304,6 @@ record_fields(Other) -> aesc_offchain_state:record_fields(Other).
 report_tags() ->
     maps:keys(?DEFAULT_REPORT_FLAGS).
 
-version_tags() ->
-    [offchain_update].
 
 bh_deltas() ->
     record_info(fields, bh_delta).
@@ -2771,12 +2768,12 @@ send_recoverable_error_msg(?WDRAW_ERR, Sn, Msg) ->
     lager:debug("send withdraw_error: ~p", [Msg]),
     aesc_session_noise:wdraw_error(Sn,Msg);
 send_recoverable_error_msg(?SHUTDOWN_ERR, Sn, Msg) ->
-    case aesc_offchain_update:get_vsn() of
-        V when V > 1 ->
+    case was_fork_activated(?LIMA_PROTOCOL_VSN) of
+        true ->
             lager:debug("send shutdown_error: ~p", [Msg]),
             aesc_session_noise:shutdown_error(Sn,Msg);
         _ ->
-            %% This message won't be defined in earlier FSMs
+            %% This message won't be defined pre Lima
             ok
     end.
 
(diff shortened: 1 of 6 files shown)


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2793,12 +2790,12 @@ send_recoverable_error_msg(?WDRAW_ERR, Sn, Msg) ->
     lager:debug("send withdraw_error: ~p", [Msg]),
     aesc_session_noise:wdraw_error(Sn,Msg);
 send_recoverable_error_msg(?SHUTDOWN_ERR, Sn, Msg) ->
-    case aesc_offchain_update:get_vsn() of
-        V when V > 1 ->
+    case was_fork_activated(?LIMA_PROTOCOL_VSN) of
```
velzevur: Taking advantage of this being Lima already. We could also remove the other head as well

apps/aechannel/src/aesc_offchain_update.erl
```diff
@@ -113,7 +110,6 @@ from_db_format(Tuple) ->
         #deposit{} -> Updated;
         #create_contract{} -> Updated;
         #call_contract{} -> Updated;
-        #meta{} -> Updated;
```
velzevur: Not possible to get here

velzevur:
I pushed to `system-tests` branch

velzevur:
Adding a release notes line is debatable

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -2791,7 +2791,7 @@ send_recoverable_error_msg(?WDRAW_ERR, Sn, Msg) ->
     aesc_session_noise:wdraw_error(Sn,Msg);
 send_recoverable_error_msg(?SHUTDOWN_ERR, Sn, Msg) ->
     case was_fork_activated(?LIMA_PROTOCOL_VSN) of
-        true ->
+        false ->
```
hanssv: This makes no sense (and the tests seem to agree?!) The comment in the next clause indicates that this is backwards, or just wrong...

apps/aechannel/src/aesc_fsm.erl
velzevur: Actually since we dropped the support for older FSMs and don't expect syncing with them, this whole case is not necessary

velzevur:
```

===> Failed to fetch and copy dep: {git,
                                  "https://github.com/aeternity/erl-base58.git",
                                  {ref,
                                      "60a335668a60328a29f9731b67c4a0e9e3d50ab6"}}

velzevur:
Failing system tests seem to be unrelated to changes at hand. Same failure happens in `master` branch as well

PR #2762, opened 2019-09-09
Bytes.concat and Bytes.split

UlfNorell:
- Pivotal: [PT-168370742](https://www.pivotaltracker.com/story/show/168370742)
- aeternity/aebytecode#80
- aeternity/aesophia#147


happi:
Perhaps create a PT for adding gas for bytes concat when gas for memory ops is merged.

PR #2764, opened 2019-09-10
Fix osx package deps

dincho:


dincho:
Unrelated `test_fortuna` failure. Merging.

PR #2767, opened 2019-09-10
Dont benchmark in CI + less logging in aesc_fsm_SUITE

hanssv:


apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -350,8 +348,8 @@ init_per_group_(Config) ->
         true -> Config;
         false ->
             aecore_suite_utils:start_node(dev1, Config),
-            aecore_suite_utils:connect(aecore_suite_utils:node_name(dev1), [block_pow, sc_cache_kdf]),
-            ct:log("dev1 connected", []),
+            aecore_suite_utils:connect(aecore_suite_utils:node_name(dev1), [block_pow]),
```
grzegorz225: Why did you disable mocking of the key derivation function inside aesc_state_cache? Without the mock each channel opening requires evaluating Argon2 which is expensive.

tolbrino: I suspect this is a rebase issue where @hanssv accidentally removed it.

hanssv: Ask @tolbrino 🤷‍♂

hanssv: Fixed.

PR #2769, opened 2019-09-11
Simplify getting protocol version in primops

jur0:
Instead of extensive use of `aec_hard_forks:protocol_effective_at_height`, it's simpler just to get the protocol version from state.


jur0:
I'm trying to get the harmless refactoring PRs in now. The ones that break API/change functionality will go after Lima.

PR #2777, opened 2019-09-12
Pt 168370198 deploy contract in hard fork

gorillainduction:
https://www.pivotaltracker.com/story/show/168370198


apps/aecore/src/aec_chain_state.erl
```diff
@@ -277,6 +277,8 @@ is_key_block(N) -> node_type(N) =:= key.
 
 is_micro_block(N) -> node_type(N) =:= micro.
 
+node_header(#node{header = H}) -> H.
+
```
sennui: just fyi, there was `export_header` (?!)

gorillainduction: Right. I was looking for it, but couldn't find it. I might fix this together with other review comments.

sennui: `export_` name is kind of out of convention

gorillainduction: It used to indicate more than just getting a field, but the name just stuck on. If I change this, I'll probably go for `node_header/1` all around instead.

gorillainduction:
Rebase for CI fix. No other change.

apps/aecore/src/aec_block_fork.erl
```diff
@@ -36,3 +37,152 @@ apply_accounts_file(Trees, Accounts) ->
...
+    Trees1        = fork_contracts_prepare_owner(Txs, Specs, Static, Trees),
+    ApplyTxs      = fun(Tx, T) -> apply_hard_fork_contract_tx(Tx, T, TxEnv) end,
+    {Contracts, Trees2} = lists:mapfoldl(ApplyTxs, Trees1, Txs),
+    Nonce         = get_nonce(OwnerPubkey, Trees2),
+    hard_fork_contracts_post_processing(Contracts, OwnerPubkey, Nonce, Trees).
+
+fork_contracts_static_specs(TxEnv) ->
+    OwnerPubkey   = aec_governance:locked_coins_holder_account(),
+    GasLimit      = aec_governance:block_gas_limit(),
+    GasPrice      = aec_governance:minimum_gas_price(aetx_env:height(TxEnv)),
```
hanssv: OK, since we don't go through the mempool. (I was worried about the miner minimum gas price for a short while.)

PR #2778, opened 2019-09-12
Pass protocol version instead of height to aetx_sign

jur0:
[PT-168428278](https://www.pivotaltracker.com/n/projects/2124891/stories/168428278)

The protocol version is known in all places where the functions from `aetx_sign` are called. `aec_hard_forks:protocol_effective_at_height` can be therefore removed from `aetx_sign` module, and the protocol version can be passed directly instead of block height.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -1851,6 +1851,10 @@ curr_height() ->
     {_, Height} = curr_hash_and_height(),
     Height.
 
+curr_protocol() ->
+    TopHeader = aec_chain:top_header(),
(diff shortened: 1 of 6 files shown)


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1851,6 +1851,10 @@ curr_height() ->
     {_, Height} = curr_hash_and_height(),
     Height.
 
+curr_protocol() ->
```
sennui: you saved 3 letters ;)

jur0: there is also `curr_height` so didn't want to break function naming consensus :)

PR #2780, opened 2019-09-13
Fix install_os_deps CI job

dincho:
Add missing apt-get update step


.circleci/config.yml
```diff
@@ -78,6 +78,7 @@ references:
     run:
       name: Install OS dependancies
```
ThomasArts: Next time we may rename it to dependencies

dincho:
I'll not wait `appveyor` as it's unrelated.

PR #2781, opened 2019-09-13
implement aens_update support for sophia

skkw:


UlfNorell:
Are there corresponding PRs for aesophia and aebytecode?

skkw:
> Are there corresponding PRs for aesophia and aebytecode?

Yes, I can see you've found them:

https://github.com/aeternity/aebytecode/pull/81
https://github.com/aeternity/aesophia/pull/155

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -556,6 +558,20 @@ init_per_testcase(TC, Config) when TC == sophia_aens_resolve;
...
     meck:expect(aec_governance, name_claim_bid_timeout, fun(_, _) -> 0 end),
     init_per_testcase_common(TC, Config);
+
+init_per_testcase(sophia_aens_update_transaction, Config) ->
+    ProtocolVsn = aec_hard_forks:protocol_vsn(?config(protocol, Config)),
+    case ProtocolVsn >= ?IRIS_PROTOCOL_VSN of
+        true ->
+            meck:expect(aec_governance, name_claim_bid_timeout, fun(_, _) -> 0 end),
+            init_per_testcase_common(sophia_aens_update_transaction,
+                                     [{sophia_version, ?SOPHIA_LIMA_FATE},
```
hanssv: I really hope you mean SOPHIA_IRIS_FATE?

hanssv: But also, these  (sophia/vm/abi_version) should already be set by `init_per_group` should not be necessary to set here.

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -556,6 +558,20 @@ init_per_testcase(TC, Config) when TC == sophia_aens_resolve;
...
+                                      {vm_version, ?VM_FATE_SOPHIA_1},
```
hanssv: Isn't this testing the new feature? `VM_FATE_SOPHIA_1` is already released and can't have new features?

apps/aecontract/test/aecontract_SUITE.erl
skkw: it's fixed now

apps/aecontract/test/aecontract_SUITE.erl
skkw: Fixed now, we have ?VM_FATE_SOPHIA_2.

apps/aecore/src/aec_hard_forks.erl
```diff
@@ -70,6 +72,21 @@ ensure_env() ->
 protocol_effective_at_height(Height) ->
     protocol_effective_at_height(Height, protocols()).
 
+-spec protocol_vsn_name(protocol_vsn()) -> atom().
```
velzevur: I see this is not being used but it is rather a complementary function to protocol_vsn/1

apps/aecore/src/aec_hard_forks.erl
```diff
@@ -70,6 +72,21 @@ ensure_env() ->
...
+protocol_vsn_name(?ROMA_PROTOCOL_VSN)    -> roma;
+protocol_vsn_name(?MINERVA_PROTOCOL_VSN) -> minerva;
+protocol_vsn_name(?FORTUNA_PROTOCOL_VSN) -> fortuna;
+protocol_vsn_name(?LIMA_PROTOCOL_VSN)    -> lima;
+protocol_vsn_name(?IRIS_PROTOCOL_VSN)    -> iris.
+
+-spec protocol_vsn(atom()) -> protocol_vsn().
```
velzevur: I see this is being used only in `aecontract_SUITE`, so it could be contained there? My concern is those two functions being used in code instead of tests and we have those names all over the place again. @jur0 worked hard to refactor those :D An alternative approach would be `-ifdef(TEST).`-s if we want to keep those here.

apps/aecore/src/aec_hard_forks.erl
skkw: This was used in (now removed) generic "assert_min_protocol" function usable in aeprimop, which was able to turn protocol version number to nicer `{required, iris, some_feature_name}` failure. 
But since the check was moved to fate level, it's not used right now... We should refactor the min protocol requirements in aeprimop (and tests) to use that to get nicer errors...

apps/aecore/src/aec_hard_forks.erl
jur0: If you don't use `protocol_effective_at_height`, I'm ok with that. I'd maybe put those names to the same hrl file as `*_PROTOCOL_VSN` defines.

apps/aeprimop/src/aeprimop.erl
```diff
@@ -856,23 +862,52 @@ name_transfer({OwnerPubkey, RecipientType, RecipientHash, NameHash}, S) ->
...
+name_update_op(OwnerPubkey, NameHash, TTL, ClientTTL, Pointers
+              ) when ?IS_HASH(OwnerPubkey),
+                     ?IS_HASH(NameHash),
+                     ?IS_CHAIN_TTL_OPTION(TTL),
+                     ?IS_CHAIN_TTL_OPTION(ClientTTL),
+                     (Pointers == undefined orelse is_list(Pointers)) ->
+    {name_update, {OwnerPubkey, NameHash, TTL, ClientTTL, Pointers}}.
+
+
+name_update({OwnerPubkey, NameHash, TTL, ClientTTL, Pointers}, S) ->
```
hanssv: This will work, but I wonder if it would not be nicer to split `aens_names:update` into `aens_names:update_ttl`, `aens_names:update_pointers` and `aens_names:update_client_ttl`... And only update those values that are **not** `undefined` and thus not a have to dig out the old values...

apps/aeprimop/src/aeprimop.erl
```diff
@@ -856,23 +862,52 @@ name_transfer({OwnerPubkey, RecipientType, RecipientHash, NameHash}, S) ->
...
-name_update({OwnerPubkey, NameHash, DeltaTTL, MaxTTL, ClientTTL, Pointers}, S) ->
-    [runtime_error(ttl_too_high) || DeltaTTL > MaxTTL],
-    {Name, S1} = get_name(NameHash, S),
-    assert_name_owner(Name, OwnerPubkey),
-    assert_name_claimed(Name),
-    AbsoluteTTL = S#state.height + DeltaTTL,
-    Name1 = aens_names:update(Name, AbsoluteTTL, ClientTTL, Pointers),
-    put_name(Name1, S1).
+    {name_update,
+     {OwnerPubkey, NameHash, {relative_ttl, DeltaTTL}, {relative_ttl, ClientTTL}, Pointers}};
```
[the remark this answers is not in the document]
hanssv: Sorry, changed my mind, it is just wrong :-) It should definitely not be `relative_ttl` since it has nothing to do with the chain height!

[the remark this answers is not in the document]
skkw: Regarding the TTL, I'm reusing Chain.TTL type from Sophia. From the manual:

Chain.ttl - Time-to-live (fixed height or relative to current block).

Example: FixedTTL(1050) or RelativeTTL(50)

I think relative ttl is useful, as you don't need to get the current height first to add your delta to it, right?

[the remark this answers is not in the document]
hanssv: I repeat then... ClientTTL has nothing to do with chain height, so why would you want to add the current height to it?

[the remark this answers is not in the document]
hanssv: And yes, this means that the `aesophia` commit probably had the wrong type in it...

apps/aeprimop/src/aeprimop.erl
skkw: The usage from Sophia looks like this:

```
  stateful entrypoint update(owner      : address,
                             name       : string,
                             ttl        : option(Chain.ttl),
                             client_ttl : option(Chain.ttl),
                             pointers   : option(map(string, AENS.pointee))) : unit =
    AENS.update(owner, name, ttl, client_ttl, pointers)
```
So it's possible to update only what's needed without digging values just to be passed around. 

Agree on the ergonomics though, that ideally we would have all variants. (maybe at 5.3.0? it would all call the same thing at FATE level)

apps/aecore/src/aec_hard_forks.erl
skkw: I admit that it should be probably better to have this in separate PR... The idea was just to have a way to translate protocol version to name and back - for better error reporting (which wasn't realized at the end here as the check went out from aeprimop)... I checked with Juraj, it doesn't seem to interfere with his work.

apps/aega/src/aega_attach_tx.erl
```diff
@@ -136,8 +136,10 @@ is_legal_version_at_protocol(#{vm := VMVersion, abi := ABIVersion}, Protocol) ->
             VMVersion == ?VM_AEVM_SOPHIA_3;
         P when P >= ?LIMA_PROTOCOL_VSN, ABIVersion == ?ABI_AEVM_SOPHIA_1 ->
             VMVersion == ?VM_AEVM_SOPHIA_4;
-        P when P >= ?LIMA_PROTOCOL_VSN, ABIVersion == ?ABI_FATE_SOPHIA_1 ->
+        P when P == ?LIMA_PROTOCOL_VSN, ABIVersion == ?ABI_FATE_SOPHIA_1 ->
             VMVersion == ?VM_FATE_SOPHIA_1;
+        P when P >= ?IRIS_PROTOCOL_VSN, ABIVersion == ?ABI_FATE_SOPHIA_1 ->
+            VMVersion == ?VM_FATE_SOPHIA_2;
```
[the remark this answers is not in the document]
skkw: Hmm, I'm confused a bit - for IRIS we check that we VM version is ?VM_FATE_SOPHIA_2, sounds right?

apps/aeprimop/src/aeprimop.erl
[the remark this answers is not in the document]
skkw: Ah I see, ClientTTL is just int, not Chain.ttl. Will fix that.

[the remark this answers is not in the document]
hanssv: Yeah, exactly, my bad for not catching it already in `aesophia`...

apps/aega/src/aega_attach_tx.erl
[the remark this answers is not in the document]
hanssv: Yes once my remark over there is fixed they will align, just making sure we double check!

apps/aecontract/src/aect_contracts.erl
```diff
@@ -170,6 +170,15 @@ is_legal_version_at_protocol_(create, #{vm := ?VM_FATE_SOPHIA_1, abi := ?ABI_FAT
         ?LIMA_PROTOCOL_VSN            -> true;
         P when P > ?LIMA_PROTOCOL_VSN -> true
```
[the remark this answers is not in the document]
skkw: I changed this, but then all hell break loose ale everything stopped working :)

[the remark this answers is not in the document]
hanssv: That kind of indicates that this PR is not complete 🤔

apps/aega/src/aega_attach_tx.erl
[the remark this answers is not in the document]
hanssv: This is still inconsistent... Here you only allow `VM_FATE_SOPHIA_1` in Lima, in `aect_contracts` it is also allowed in Iris...

apps/aecontract/src/aect_contracts.erl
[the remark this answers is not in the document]
hanssv: It should be possible...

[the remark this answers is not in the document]
skkw: Should be fixed now

apps/aecontract/src/aect_contracts.erl
```diff
@@ -196,6 +205,13 @@ is_legal_version_at_protocol_(call, #{vm := VMVersion}, Protocol) ->
                                    (VMVersion =:= ?VM_FATE_SOPHIA_1)
                                    ) ->
             true;
+        P                     when (P >= ?IRIS_PROTOCOL_VSN)
+                                   and
+                                   (
+                                   (VMVersion =:= ?VM_FATE_SOPHIA_1) or
```
hanssv: This is completely redundant. It is already covered by the case above...

apps/aega/src/aega_attach_tx.erl
```diff
@@ -136,8 +136,10 @@ is_legal_version_at_protocol(#{vm := VMVersion, abi := ABIVersion}, Protocol) ->
             VMVersion == ?VM_AEVM_SOPHIA_3;
         P when P >= ?LIMA_PROTOCOL_VSN, ABIVersion == ?ABI_AEVM_SOPHIA_1 ->
             VMVersion == ?VM_AEVM_SOPHIA_4;
-        P when P >= ?LIMA_PROTOCOL_VSN, ABIVersion == ?ABI_FATE_SOPHIA_1 ->
+        P when P == ?LIMA_PROTOCOL_VSN, ABIVersion == ?ABI_FATE_SOPHIA_1 ->
```
hanssv: This is getting repetitive... This is _still_ not consistent with L165-172 in `aect_contracts`

apps/aecontract/src/aect_contracts.erl
skkw: fixed

apps/aega/src/aega_attach_tx.erl
skkw: fixed

PR #2782, opened 2019-09-13
Improve swagger

velzevur:
This fixes the swagger.json being produced to include a `parameters` key even when it is empty.

This also fixes [JS SDK issue 657](https://github.com/aeternity/aepp-sdk-js/issues/657)

### DIFF
--- a/apps/aehttp/priv/swagger.yaml
+++ b/apps/aehttp/priv/swagger.yaml
@@ -49,6 +49,7 @@ paths:
       description: 'Get the top block (either key or micro block)'
       produces:
         - application/json
+      parameters: []
       responses:
         '200':
           description: 'Successful operation'
@@ -67,6 +68,7 @@ paths:
       description: 'Get the current key block'
       produces:
         - application/json
+      parameters: []
       responses:
         '200':
           description: 'Successful operation'
@@ -85,6 +87,7 @@ paths:
       description: 'Get the hash of the current key block'
       produces:
         - application/json
+      parameters: []
       responses:
         '200':
           description: 'Successful operation'
@@ -106,6 +109,7 @@ paths:
       description: 'Get the height of the current key block'
       produces:
         - application/json
+      parameters: []
       responses:
         '200':
           description: 'Successful operation'
@@ -127,6 +131,7 @@ paths:
       description: 'Get the pending key block'
       produces:
         - application/json
+      parameters: []
       responses:
         '200':
           description: 'Successful operation'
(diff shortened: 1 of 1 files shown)


apps/aehttp/priv/swagger.yaml
```diff
@@ -1665,15 +1705,6 @@ paths:
 ##
 
 parameters:
-  HeightIn:
```
ThomasArts: Why is it an improvement to inline this? Just wondering

hanssv:
The title is wrong it should be "Improve" I think...

apps/aehttp/priv/swagger.yaml
velzevur: Old behaviour was that this was inlined for us by our tools that produced [swagger.json](https://github.com/aeternity/aeternity/blob/v4.2.0/apps/aehttp/priv/swagger.json#L224-L232). Since we depricated those tools for good, now suddenly we don't inline this and it causes issues client-side.

Not really improvement but more of returning to initial behaviour from client's perspective

dincho:
Indeed I would not say that this is an improvement but backward compatibility kind of PR. While the empty parameters is clearly because of broken parsers downstream (the spec is clear that it's optional), the path parameters reference is still not clear if it's OK in the spec, even the online [validator complains about it (master)](https://online.swagger.io/validator/debug?url=https%3A%2F%2Fraw.githubusercontent.com%2Faeternity%2Faeternity%2Fmaster%2Fapps%2Faehttp%2Fpriv%2Fswagger.yaml).

The [swagger v2 specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject) is not explicit about in path parameters, but it's more general:
> A list of parameters that are applicable for all the operations described under this path. These parameters can be overridden at the operation level, but cannot be removed there. The list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of a name and location. The list can use the Reference Object to link to parameters that are defined at the Swagger Object's parameters. There can be one "body" parameter at most.

Note: **The list can use the Reference Object to link to parameters that are defined at the Swagger Object's parameters.**

This PR online swagger validator [does not throw errors](https://online.swagger.io/validator/debug?url=https://raw.githubusercontent.com/aeternity/aeternity/6afa712cb03bf358efafbb029a9f60e88a1a16e2/apps/aehttp/priv/swagger.yaml)

@velzevur I'd like to see more details commit message (during merge) that explains both changes

tolbrino:
> If we merge #2784 this PR should be closed and the changes from lima synced to master
> 
> Blocking until decision

So the approach would be to forward-port any changes from `lima` to `master` and never backport changes from `master` to `lima` ? I'm asking because its usually the other way around.

PR #2784, opened 2019-09-13
Improve swagger

velzevur:
Same as #2782, opened against `releases/lima-rc` so SDK works there as well


hanssv:
Seems to break the SC FSM somehow, otherwise it looks good.

PR #2786, opened 2019-09-17
Upload windows artifacts

venimus:
Upload windows CircleCI artifacts
and small consistency changes
PKG_SUFFIX now includes the extension as windows builds `.zip`


.circleci/config.yml
```diff
@@ -42,7 +42,6 @@ references:
   container_otp21_plt: &container_otp21_plt _build/default/rebar3_21.3.8.6_plt
 
   infrastructure_config_environment: &infrastructure_config_environment
-    resource_class: large
```
venimus: seems a copy/paste leftover

.circleci/config.yml
```diff
@@ -53,6 +52,11 @@ references:
     environment:
       <<: *infrastructure_config_environment
 
+  upload_container_config: &upload_container_config
+    executor: infrastructure_container_stable
+    resource_class: large
```
venimus: i'm not sure we need `large`

.circleci/config.yml
```diff
@@ -53,6 +52,11 @@ references:
     environment:
       <<: *infrastructure_config_environment
 
+  upload_container_config: &upload_container_config
```
dincho: What's the point of this additional config ?

.circleci/config.yml
```diff
@@ -241,6 +245,12 @@ references:
       command: |
         envdir /secrets aws s3 cp --acl public-read ${PACKAGE_TARBALL:?} s3://${S3_RELEASES_BUCKET:?}/
 
+  upload_latest_artifacts: &upload_latest_artifacts
```
dincho: Please create a proper CircleCI command, there is no full config migration scheduled, so we have to do it gradually when adding/changing it

.circleci/config.yml
```diff
@@ -392,17 +402,15 @@ references:
...
-    - *set_package_path
-    - setup_secrets
-    - attach_workspace:
-        at: *packages_workspace
-    - run:
-        name: Upload build artifacts to central repository (AWS)
-        command: |
-          envdir /secrets aws s3 cp --acl public-read ${PACKAGE_TARBALL:?} s3://${S3_RELEASES_BUCKET:?}/aeternity-latest-${PKG_SUFFIX:?}.tar.gz
-    - *fail_notification
+  upload_latest_package: &upload_latest_package
```
dincho: I don't like the naming, not each package is artifact

.circleci/config.yml
```diff
@@ -1035,41 +1043,53 @@ jobs:
       - DEPLOY_DOWNTIME: 1800 #30M
     steps: *deploy_steps
 
+  # Upload artifacts
   upload_packages_osx:
-    executor: infrastructure_container_stable
```
dincho: I prefer to keep the explicit executors for each job, with CircleCI 2.1 it's much readable than current mix of partials

.circleci/config.yml
```diff
@@ -1471,7 +1496,7 @@ workflows:
           type: approval
           filters:
             branches:
-              ignore: /.*/
+              ignore: /v.*/
```
dincho: Is this for testing purposes? Otherwise it does not make sense

venimus: yes will be reverted after simulations

.circleci/config.yml
venimus: it is because the Upload instances could use different (smaller) resource than build

.circleci/config.yml
venimus: Then it might be better to refactor the other command too, as we could use the track of gradual changes

.circleci/config.yml
venimus: it was for consistency with `upload_package`

venimus: also there is already `&upload_latest_artifacts`

venimus:
Some review notes:
ci build that simulates release is https://circleci.com/workflow-run/7d55b944-9153-4190-a230-d1c421ffe254
However i skipped the upload of 'latest' versions to AWS as it will override the release artifacts with junk, so this step is not there, however there are 3 'versioned' copies successfully uploaded https://s3.console.aws.amazon.com/s3/buckets/aeternity-node-releases/?region=eu-north-1&tab=overview
Github release artifacts can be found here in this Draft release https://github.com/aeternity/aeternity/releases/tag/untagged-bbc315360d6d9e96b471

Those leftovers will be cleaned up after merge

PR #2797, opened 2019-09-25
add querying of oracle via name hash

skkw:


apps/aeoracle/src/aeo_query_tx.erl
```diff
@@ -244,9 +242,14 @@ serialization_template(?ORACLE_QUERY_TX_VSN) ->
 version(_) ->
     ?ORACLE_QUERY_TX_VSN.
 
+
 -spec valid_at_protocol(aec_hard_forks:protocol_vsn(), tx()) -> boolean().
-valid_at_protocol(_, _) ->
-    true.
+valid_at_protocol(Proto, Tx) ->
+    case {Proto >= ?LIMA_PROTOCOL_VSN, aeser_id:specialize_type(oracle_id(Tx))} of
```
hanssv: 😱 This should be valid only Post Lima!

skkw: This will have to be changed in tests too... It's there for development at the moment.

apps/aeprimop/src/aeprimop.erl
```diff
@@ -643,28 +643,59 @@ oracle_query_op(OraclePubkey, SenderPubkey, SenderNonce, Query, QueryFee,
...
+oracle_query({OracleId, SenderPubkey, SenderNonce,
+              Query, QueryFee, QTTL, RTTL, Return}, S) ->
+    case aeser_id:specialize(OracleId) of
+        {oracle, OraclePubkey} ->
+            oracle_query({OraclePubkey, OracleId, SenderPubkey, SenderNonce,
+                          Query, QueryFee, QTTL, RTTL, Return}, S);
+        {name, NameHash} ->
+            S#state.protocol >= ?IRIS_PROTOCOL_VSN orelse
+                runtime_error(oracle_query_by_name_hash_not_available_at_protocol),
+            {OraclePubkey, S1} = int_resolve_name(NameHash, <<"oracle_pubkey">>, S),
```
sennui: I think we should extract well know keys somewhere more visible

PR #2798, opened 2019-09-25
Integration test for Lima token migration contract

hanssv:


test/contracts/token_migration.aes
```diff
@@ -0,0 +1,80 @@
+payable contract TokenMigration =
```
dincho: Maybe a reference to the original source/commit would be helpful in future. I noticed somewhere it is tracked in some repo

hanssv: I'm waiting for the contract to be merged in that repo, but yes, good point.

apps/aecontract/src/aect_state_tree.erl
```diff
@@ -156,19 +156,31 @@ enter_store_nodes(Prefix, Writes, CtTree) ->
...
+                      aect_contracts:contract().
 get_contract(Pubkey, #contract_tree{ contracts = CtTree }, Options) ->
     Contract = aect_contracts:deserialize(Pubkey, aeu_mtrees:get(Pubkey, CtTree)),
     case proplists:get_value(no_store, Options, false) of
-        false -> add_store(Contract, CtTree);
+        false -> add_store(Contract, CtTree, Options);
         true  -> Contract
     end.
 
 add_store(Contract, CtTree) ->
```
ThomasArts: This function is used in lookup. But the weird thing is that the no_store option there is removed and [] is created as option list. Not sure this is intentional or a slip.

https://github.com/aeternity/aeternity/blob/57d82c7ee59ccf49082b68fff11c94a9cb917b4f/apps/aecontract/src/aect_state_tree.erl#L204

gorillainduction: Agreed. It makes sense to pass on the options there as well, even if it is not currently used.
I'll add it.

PR #2800, opened 2019-09-25
Remove confusing comments

jur0:
~No PT, based on [this comment](https://github.com/aeternity/aeternity/pull/2697#discussion_r327553125). It removes dead code.~

It actually removes just confusing comments as it can happen that a malicious node can send a block not connected to a key block.


apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1295,21 +1295,17 @@ get_onchain_env(MicroHeader) ->
                 key ->
                     {PrevHeader, PrevHeader};
                 micro ->
-                    case aec_chain:get_header(PrevKeyHash) of
-                        {ok, PrevKeyHeader} -> {PrevHeader, PrevKeyHeader};
-                        error               -> {PrevHeader, not_found}
-                    end
+                    {ok, PrevKeyHeader} = aec_chain:get_header(PrevKeyHash),
```
hanssv: Does this mean that you can crash the remote peer-process by sending an unconnected micro-block?

hanssv: Wasn't this exactly what I asked about in another PR? 😕

PR #2805, opened 2019-09-26
SC: WS Handler hogs the jobs pool

### LINKED ISSUE #2823: SC: WS Handler hogs the jobs pool
The sc_ws_handler instance created when starting a channel via the WebSocket API doesn't acknowledge completion to the load regulator until it reaches terminate(). This means that no more concurrent channels can be allowed than the size of the worker pool. This is not how it's supposed to work.

One could debate whether it should call jobs:done(JobId). The most reasonable place should be after it has tried to spawn the fsm.

### LINKED ISSUE #2876: Expose maximum FSM limit to config
Currently there is no way to reduce the maximum amount of simultaneously running FSMs

uwiger:
See [PT #168782034](https://www.pivotaltracker.com/story/show/168782034)

Fixes #2823 
Fixes #2876 

This change makes the SC WS handler release the jobs worker after spawning (or reconnecting to) the fsm. Before this change, the handler would hold on to the worker until `terminate()` which meant that the size of the jobs `counter` pool effectively limited the number of currently active state channels.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -3499,7 +3499,23 @@ callback_mode() ->
 -spec init(map()) -> {ok, InitialState, data(), [{timeout, Time::pos_integer(),
                                                    InitialState}, ...]}
                           when InitialState :: state_name().
-init(#{opts := Opts0} = Arg) ->
+init(#{opts := Opts} = Arg) ->
+    case check_limits(Opts) of
+        ok ->
+            init_(Arg);
+        {error, Reason} ->
+            {stop, Reason}
+    end.
+
+check_limits(Opts) ->
+    case maps:is_key(existing_channel_id, Opts) of
+        true ->
+            aesc_limits:register_returning();
+        false ->
+            aesc_limits:allow_new()
+    end.
+
+init_(#{opts := Opts0} = Arg) ->
     %% Protect the password from leakage
     StatePasswordWrapper = aesc_state_password_wrapper:init(maps:find(state_password, Opts0)),
     Opts1 = maps:remove(state_password, Opts0),
--- a/apps/aechannel/src/aesc_limits.erl
+++ b/apps/aechannel/src/aesc_limits.erl
@@ -1,0 +1,122 @@
+%%% @doc This module implements a simple registry to keep track of the number of FSMs.
+%%%
+%%% The `aesc_limits' server monitors each FSM, and inserts an entry into an ETS table.
+%%% The size of the ETS table reflects the number of active FSMs in the node.
+%%% We define two entry points in order to distinguish between reestablish attempts
+%%% (`register_returning()'), for channels which we have already agreed to serve, and
+%%% attempts to create new channels (`allow_new()'), where a limit check is actually
+%%% imposed. This means that in certain corner cases, if lots of channels happen to
+%%% leave and later reestablish roughly simultaneously, we might overshoot and end
+%%% up with more channels than the configured limit (theoretically, many times as
+%%% many). To avoid this, a count of channels which are currently 'on leave' would
+%%% need to be kept. This could perhaps be kept current by the state cache.
+%%%
+-module(aesc_limits).
+
+-behavior(gen_server).
+
+-export([ allow_new/0
+        , register_returning/0 ]).
+
+-export([ start_link/0
+        , init/1
+        , handle_call/3
+        , handle_cast/2
+        , handle_info/2
+        , terminate/2
+        , code_change/3
+        ]).
+
+-record(st, { }).
+
+-define(PIDS, aesc_limits_pids).
+
+allow_new() ->
+    gen_server:call(?MODULE, allow_new).
+
+register_returning() ->
+    gen_server:call(?MODULE, register_returning).
+
+start_link() ->
+    %% We make the ets table public, since it's created by the supervisor.
+    %% An alternative would be to set an `heir' option and use `ets:give_away/3`
+    %% to give the newly (re-)started server control; this way, the ets table
+    %% could be private. In either case, if the server restarts, it will have to
+    %% either establish new monitors for all pids in the surviving table, or
+    %% find all running FSMs, e.g. via gproc, to set up new monitors.
+    ets:new(?PIDS, [set, public, named_table]),
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+
+init([]) ->
+    %% If this is a restart, ensure that all existing pids have working mrefs
+    ets:foldl(fun refresh_monitor/2, ok, ?PIDS),
+    {ok, #st{}}.
+
+handle_call(register_returning, {Pid, _}, S) ->
+    MRef = monitor(process, Pid),
+    case ets:insert_new(?PIDS, {Pid, MRef}) of
+        true ->
+            lager:debug("Returning session (~p) allowed", [Pid]),
+            {reply, ok, S};
+        false ->
+            lager:debug("Returning session (~p) denied: already exists", [Pid]),
+            demonitor(MRef),
+            {reply, {error, exists}, S}
+    end;
+handle_call(allow_new, {Pid,_}, S) ->
+    Limit = get_limit(),
+    case ets:info(?PIDS, size) of
+        Sz when Sz >= Limit ->
+            lager:debug("New session (~p) denied; Sz = ~p, Limit = ~p",
+                        [Pid, Sz, Limit]),
+            {reply, {error, channel_count_limit_exceeded}, S};
+        Sz ->
+            MRef = monitor(process, Pid),
+            case ets:insert_new(?PIDS, {Pid, MRef}) of
+                true ->
+                    lager:debug("New session (~p) allowed; Size = ~p, Limit = ~p",
+                                [Pid, Sz, Limit]),
+                    {reply, ok, S};
+                false ->
+                    lager:debug("New session (~p) denied: already exists", [Pid]),
+                    demonitor(MRef),
+                    {reply, {error, exists}, S}
+            end
+    end;
+handle_call(_, _, S) ->
+    {reply, {error, unknown_call}, S}.
+
+handle_cast(_Msg, S) ->
+    {noreply, S}.
+
+handle_info({'DOWN', MRef, process, Pid, _Reason}, S) ->
+    NDeleted = ets:select_delete(?PIDS, [{ {Pid, MRef}, [], [true] }]),
+    lager:debug("'DOWN' received; ~p entries deleted for ~p", [NDeleted, Pid]),
+    {noreply, S};
+handle_info(_, S) ->
+    {noreply, S}.
+
+terminate(_, _) ->
+    ok.
+
+code_change(_FromVsn, S, _Extra) ->
+    {ok, S}.
+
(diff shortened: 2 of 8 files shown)


AleksandarFilipov:
I don't understand why this fix is needed. All the old connections will be terminated once they are replaced with a new one. Are calls to terminate missing, and we have lingering handlers? Please help me understand.

uwiger:
> I don't understand why this fix is needed. All the old connections will be terminated once they are replaced with a new one. Are calls to terminate missing, and we have lingering handlers? Please help me understand.

The jobs regulation is only meant to be applied to the start of the job, i.e. the creation of the channel. If the job is short-lived, it's ok to simply ask jobs for permission, then run to completion, but since a state channel can essentially live forever, the job may run on indefinitely. This is not what the `counter` regulation in jobs was meant for.

If we really want to limit the maximum number of WS connections, `cowboy` has a `max_connections` option that we can use. The `counter` setting is only meant to limit how many concurrently ongoing channel setup requests we have, so the `sc_ws_handler` should reply to jobs as soon as it has reached a point where it's 'done' with the initial work - roughly when it's time to wait for peer or user input, for example.

velzevur:
I see this maximum cap of open WebSockets was broken and possibly was a number much lower than the specified one. This PR fixes the issue as limiting the jobs to the WebSockets that are currently being upgraded. This is actually limited by `cowboy`'s default value for [Ranch `max_connections`](https://ninenines.eu/docs/en/ranch/2.0/guide/listeners/), so having the `jobs` do the same thing could be misleading.

Also - removing the check for opened WebSocket connections. we should probably limit the total amount of running FSMs on a node. Tracked as #2876

uwiger:
> Also - removing the check for opened WebSocket connections. we should probably limit the total amount of running FSMs on a node. Tracked as #2876

This PR actually does that, via the `channels.max_count` config variable (default: `1000`)

velzevur:
> This PR actually does that, via the channels.max_count config variable (default: 1000)
Yes, I had the PR opened in a tab. Sorry for the noise

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -461,6 +484,30 @@ strict_checks(Fsm, Strict) when is_boolean(Strict) ->
...
+number_of_fsms() ->
+    gproc:lookup_value({a,l,?MODULE}).
+
+within_fsm_number_limit() ->
+    {ok, Max} = aeu_env:find_config([ <<"channels">>, <<"max_count">> ] , [ user_config
+                                                                          , schema_default
+                                                                          , {value, 1000} ]),
+    N = number_of_fsms(),
+    lager:debug("Number of fsms: ~p; Max = ~p", [N, Max]),
+    N < Max.
```
velzevur: Here `N` does not include the include the FSM being started at the moment so the maximum amount of FSMs is indeed according to [description](https://github.com/aeternity/aeternity/pull/2805/files#diff-fc6905deb828c517288fba567781e76fR323). Ok

velzevur:
I would expect a release notes line as this is client visible. Probably a simple test would be nice as well.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -206,16 +211,19 @@ inband_msg(Fsm, To, Msg) ->
...
+                              aesc_fsm_sup:start_child([#{ host => Host
+                                                         , port => Port
+                                                         , opts => Opts }]);
+                          {error, _Reason} = Err ->
+                              Err
+                      ?CATCH_LOG(_E)
+                          {error, _E}
+                      end
+              end,
+    check_limits(Proceed, Opts).
```
tolbrino: Why is the `check_limits` check not added to `init_checks` instead?

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -225,15 +233,30 @@ respond(Port, #{} = Opts0) ->
...
+                          ok ->
+                              aesc_fsm_sup:start_child([#{ port => Port
+                                                         , opts => Opts }]);
+                          {error, _Reason} = Err ->
+                              Err
+                      ?CATCH_LOG(_E)
+                          {error, _E}
+                      end
+              end,
+    check_limits(Proceed, Opts).
```
tolbrino: Same as above.

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -461,6 +484,30 @@ strict_checks(Fsm, Strict) when is_boolean(Strict) ->
     gen_statem:call(Fsm, {strict_checks, Strict}).
 -endif.
 
+%% ======================================================================
```
tolbrino: These functions should be moved to another module to not extend the complexity of `aesc_fsm`. I'd suggest `aesc_fsm_stats`.

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -246,6 +246,7 @@ handler_init_error(Err, Handler) ->
                     {channel_reserve_too_low      , value_too_low},
                     {push_amount_too_low          , value_too_low},
                     {lock_period_too_low          , value_too_low},
+                    {limit_exceeded               , limit_exceeded},
```
tolbrino: Which limit? The error message should be more precise.

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -315,6 +315,17 @@
                 }
             }
         },
+        "channels" : {
+            "type" : "object",
+            "additionalProperties" : false,
+            "properties" : {
+                "max_count" : {
+                    "description" : "Max number of fsms allowed on node",
```
tolbrino: We should not leak implementation-level details into the documentation. FSM is such a detail. Instead I'd call it `active state channels`.

tolbrino:
> I would expect a release notes line as this is client visible. Probably a simple test would be nice as well.

I agree, a test is missing.

apps/aehttp/src/sc_ws_handler.erl
uwiger: That's partly a matter of strategy: do we have many detailed error types, or more general, perhaps complemented with more detailed info?

apps/aeutils/priv/aeternity_config_schema.json
uwiger: Yes, except there's a difference between serving only one-half of a state channel and having both halves running on the same node.

apps/aechannel/src/aesc_fsm.erl
uwiger: Partly because `init_checks()` is checking the static properties of the provided arguments, while the `check_limits()` function checks the dynamic headroom of the node, and partly because the dynamic check should be performed as late as possible - or as early as possible. There are valid arguments for both, where an argument for early checking would be that the node should minimize the amount of work spent on rejected requests, and an argument for late checking might be that we want to treat new channels and 'returning' channels differently. Either way, the limit check is different from the parameter checking.

apps/aeutils/priv/aeternity_config_schema.json
tolbrino: That is correct, my suggestion was incomplete. Still I'd prefer to have a more high level approach.

apps/aehttp/src/sc_ws_handler.erl
tolbrino: So far most error messages are only used in a single case, so I expected the first option. But I don't have a preference.

apps/aechannel/src/aesc_fsm.erl
tolbrino: I agree that the limit check should be performed early. However, the same result could be achieved by simply appending it to the list of init checks, since `aeu_validation:run/1` will stop as soon as a check fails.

apps/aeutils/priv/aeternity_config_schema.json
uwiger: Suggestions are welcome, but we actually use "FSM" already in the published documentation, e.g.
[here](https://github.com/aeternity/protocol/blob/master/channels/OFF-CHAIN.md#channel_reestablish) and [here](https://github.com/aeternity/protocol/blob/master/node/api/channels_api_usage.md#authenticated-channel_create_tx). I agree that it's not a particularly user-friendly term.

tolbrino: `active state channel clients` ?

tolbrino:
@uwiger This PR needs to be rebased on master.

velzevur:
@uwiger please consider some tests improvements in #2906

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -233,9 +233,18 @@ respond(Port, #{} = Opts0) ->
...
-        {error, _Reason} = Err -> Err
+        {error, _Reason} = Err ->
+            Err
     ?CATCH_LOG(_E)
-        {error, _E}
+            {error, _E}
+    end.
+
+check_limits(Opts) ->
+    case maps:is_key(existing_channel_id, Opts) of
```
tolbrino: What if this value is not at least a valid channel id? E.g. `undefined`?

apps/aechannel/src/aesc_limits.erl
```diff
@@ -0,0 +1,90 @@
+-module(aesc_limits).
```
tolbrino: This deserves documentation.

apps/aechannel/src/aesc_limits.erl
```diff
@@ -0,0 +1,90 @@
...
+-define(PIDS, aesc_limits_pids).
+
+allow_new() ->
+    gen_server:call(?MODULE, allow_new).
+
+register_returning() ->
+    gen_server:call(?MODULE, register_returning).
+
+start_link() ->
+    ets:new(?PIDS, [set, public, named_table]),
```
tolbrino: Why public?

uwiger: Easier to debug in some corner cases, but they certainly don't have to be public.

apps/aechannel/src/aesc_fsm.erl
uwiger: Then the reestablish will fail, except a bit later. It might be noted that `aesc_fsm:init_/1` also uses `maps:is_key(existing_channel_id, Opts1)` to determine whether a reestablish should be attempted.

apps/aechannel/src/aesc_limits.erl
uwiger: I actually had documentation in the first version, that was unfortunately lost as I extracted myself from the mess caused by forgetting that I was in the middle of a rebase. :)

I will add some documentation again.

apps/aechannel/src/aesc_limits.erl
uwiger: Actually, also because the ets table is created and owned by the supervisor.
In order to make it private, the table will need to be inherited by the `aesc_limits` server, which is of course doable.

apps/aechannel/src/aesc_limits.erl
```diff
@@ -0,0 +1,122 @@
+%%% @doc This module implements a simple registry to keep track of the number of FSMs
```
tolbrino:
```suggestion
%%% @doc This module implements a simple registry to keep track of the number of FSMs.
```

PR #2807, opened 2019-09-27
Lima sync

dincho:
merge without squash


velzevur:
Intermittent failures kicked in

PR #2808, opened 2019-09-27
Document database backend configuration notes

venimus:
in scope of https://www.pivotaltracker.com/story/show/166283878
I added some documentation that seemed important for users.
It's open for suggestions

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -279,13 +279,13 @@ If the file is valid YAML but does not contain a valid configuration, it prints 
 
 Aeternity nodes support several types of database persistence backends:
 
- - RocksDB (default for Unix, available for Unix)
- - Mnesia (default for Win32, available for all)
- - Leveled (experimental, available for all)
+ - RocksDB (default for Unix, supported by Unix)
+ - Mnesia (default for Win32, supported by all OS'es)
+ - Leveled (experimental, supported by all OS'es)
 
 You may choose the database backend by setting `chain.db_backend` to the corresponding value `rocksdb`, `mnesia`, `leveled`
 
-Example:
+Example (switch to Leveled):
 
 ```yaml
 chain:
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -225,6 +225,7 @@ Place the config file in one of the locations specified in the [File name and lo
 Make sure you amend the `sync` > `port` parameter with your actual value.
 
 ```yaml
+
```
venimus: some MD pareses consider `---` to be a 'horizontal rule' if not separated by NL

docs/configuration.md
```diff
@@ -272,3 +273,41 @@ You shall read output like the following:
 OK
 ```
 If the file is valid YAML but does not contain a valid configuration, it prints a helpful output.
+
+
+## Database backend
+
+Aeternity nodes support several types of database persistence backends:
+ - RocksDB (default for Unix, available for Unix)
```
dincho: Could you please add newline here as well, it's required in the standard markdown. Github preview unfortunately support only their own frankenstein.

However, docs.aeternity.com uses the standard markdown parser.

docs/configuration.md
```diff
@@ -272,3 +273,41 @@ You shall read output like the following:
...
+RocksDB does not work with NTFS volumes.
+
+**Leveled** is designed to be a better alternative to RocksDB and is available for all OS'es.
+However currently it has an experimental support.
+
+**Mnesia** is the DB backend that is distributed with Erlang/OTP but is considered less reliable than the other two.
+It is currently the default database when RocksDB is not available (i.e. Win32)
+
+Notes:
+ - If using RocksDB, `db_path` should not point to an NTFS volume (like a mapped windows drive in WSL or volume mounts
```
dincho: Newline is required here as well.

docs/configuration.md
```diff
@@ -272,3 +273,43 @@ You shall read output like the following:
...
+    db_backend: leveled
+```
+
+**RocksDB** is only available under Unix compatible systems (including OS X and WSL) and is used there by default.
+RocksDB does not work with NTFS volumes.
+
+**Leveled** is designed to be a better alternative to RocksDB and is available for all OS'es.
+However currently it has an experimental support.
+
+**Mnesia** is the DB backend that is distributed with Erlang/OTP but is considered less reliable than the other two.
```
tolbrino: I wouldn't say so. Technically mnesia is also used when choosing the other 2 db backends. Just that the mnesia disk storage is not used. It has some drawbacks compared to the other options, but I don't think reliability is it.

venimus: I probably didn't use the proper term, may be less performant?

PR #2809, opened 2019-09-30
Remove protocol_affective_at_height from key candidate create

jur0:
[PT-168320388](https://www.pivotaltracker.com/n/projects/2124891/stories/168320388)

It removes `aec_hard_forks:protocol_effective_at_height` from `aec_chain_state:fake_key_node` and protocol is passed to the function instead.

It also saves one db read in `aec_block_key_candidate:create`.

### DIFF
--- a/apps/aecore/src/aec_block_key_candidate.erl
+++ b/apps/aecore/src/aec_block_key_candidate.erl
@@ -6,8 +6,7 @@
 %%%=============================================================================
 -module(aec_block_key_candidate).
 
--export([ create/2
-        ]).
+-export([create/2]).
 
 -ifdef(TEST).
 -export([adjust_target/2]).
@@ -17,19 +16,13 @@
 -include("blocks.hrl").
 
 %% -- API functions ----------------------------------------------------------
--spec create(aec_blocks:block() | aec_blocks:block_header_hash(),
-             aec_keys:pubkey()) ->
+-spec create(aec_blocks:block(), aec_keys:pubkey()) ->
                     {ok, aec_blocks:block()} | {error, term()}.
-create(BlockHash, Beneficiary) when is_binary(BlockHash) ->
-    case aec_chain:get_block(BlockHash) of
-        {ok, Block} ->
-            int_create(BlockHash, Block, Beneficiary);
-        error ->
-            {error, block_not_found}
-    end;
-create(Block, Beneficiary) ->
-    {ok, BlockHash} = aec_blocks:hash_internal_representation(Block),
-    int_create(BlockHash, Block, Beneficiary).
+create(PrevBlock, Beneficiary) ->
+    Height = aec_blocks:height(PrevBlock) + 1,
+    {ok, PrevBlockHash} = aec_blocks:hash_internal_representation(PrevBlock),
+    Protocol = aec_hard_forks:protocol_effective_at_height(Height),
+    int_create(PrevBlockHash, PrevBlock, Beneficiary, Protocol).
 
 -spec adjust_target(aec_blocks:block(), list(aec_headers:header())) ->
                            {ok, aec_blocks:block()} | {error, term()}.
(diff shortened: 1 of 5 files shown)


apps/aecore/src/aec_conductor.erl
```diff
@@ -287,9 +290,11 @@ handle_call(is_leader, _From, State = #state{ consensus = Cons }) ->
 handle_call(reinit_chain, _From, State1 = #state{ consensus = Cons }) ->
     %% NOTE: ONLY FOR TEST
     ok = reinit_chain_state(),
-    TopBlockHash = aec_chain:top_block_hash(),
+    TopBlock = aec_chain:top_block(),
+    {ok, TopBlockHash} = aec_blocks:hash_internal_representation(TopBlock),
```
sennui: Is there no risk here? Now we re-compute the headers, before we read it from the database. 
Why the change? (just curious.)

apps/aecore/src/aec_conductor.erl
```diff
@@ -240,17 +242,18 @@ handle_call({add_synced_block, Block},_From, State) ->
...
 handle_call(get_key_block_candidate,_From, #state{beneficiary = undefined} = State) ->
     {reply, {error, beneficiary_not_configured}, State};
-handle_call(get_key_block_candidate,_From, State) ->
+handle_call(get_key_block_candidate,_From,
+            #state{top_block = TopBlock, top_block_hash = TopBlockHash} = State) ->
     {Res, State1} =
         case State#state.pending_key_block of
             undefined ->
-                get_pending_key_block(aec_chain:top_block_hash(), State);
+                get_pending_key_block(TopBlock, State);
```
hanssv: This looks like a change in behavior? Using the local state rather than the current chain state?

apps/aecore/src/aec_block_key_candidate.erl
```diff
@@ -17,19 +16,13 @@
 -include("blocks.hrl").
 
 %% -- API functions ----------------------------------------------------------
--spec create(aec_blocks:block() | aec_blocks:block_header_hash(),
-             aec_keys:pubkey()) ->
+-spec create(aec_blocks:block(), aec_keys:pubkey()) ->
```
sennui: OK, and you can do it, because you added top block to conductor's state

apps/aecore/src/aec_block_key_candidate.erl
```diff
@@ -51,54 +44,44 @@ adjust_target(Block, AdjHeaders) ->
...
-    Block = int_create_block(PrevBlockHash, PrevBlock, Miner, Beneficiary, Trees),
+int_create(PrevBlockHash, PrevBlock, Miner, Beneficiary, AdjChain, Protocol) ->
+    {ok, Trees} =
+        aec_chain_state:calculate_state_for_new_keyblock(PrevBlockHash, Miner, Beneficiary, Protocol),
+    Block = int_create_block(PrevBlockHash, PrevBlock, Miner, Beneficiary, Trees, Protocol),
     case adjust_target(Block, AdjChain) of
-        {ok, AdjBlock} -> {ok, AdjBlock};
+        {ok, AdjBlock}  -> {ok, AdjBlock};
         {error, Reason} -> {error, {failed_to_adjust_target, Reason}}
     end.
```
sennui: all whitespace changes?

sennui:
Agree with Hans, could have more comment.

apps/aecore/src/aec_conductor.erl
jur0: When a new block is successfully added, then the `TopBlock` is changed in `preempt_if_new_top`. I assume the `TopBlock` from the state and `aec_chain:top_block` and `top_block_hash` always refer to the `TopBlock`.

apps/aecore/src/aec_block_key_candidate.erl
jur0: This saves 1 db read - instead of passing the hash and reading the block from the hash, the whole block is passed.

apps/aecore/src/aec_conductor.erl
ThomasArts: The conductor is a program running in parallel with other stuff. I can imagine a race in which the top block has changed between the state update and the call to get the key block. Don't recall whether there is a reason exactly here, but in general we have had issues with concurrent updates in the past.

apps/aecore/src/aec_conductor.erl
jur0: The top block is read from the db using the top block hash, so I suppose there is no risk here.
```
-spec top_block() -> 'undefined' | aec_blocks:block().
top_block() ->
    case top_block_hash() of
        undefined -> undefined;
        Hash -> aec_db:get_block(Hash)
    end.

-spec top_block_hash() -> 'undefined' | binary().
top_block_hash() ->
    aec_db:get_top_block_hash().
```

apps/aecore/src/aec_conductor.erl
jur0: My understanding is that even if there are several workers that are able to add new blocks (sync, gossip, HTTP API,...), all of these components do it using the conductor process, they just use `post_block` or `add_synced_block` functions from the conductor - these are calls. The conductor just needs to make sure each new added block is valid and connected to the genesis. Based on this, I thought it'd safe to just use the `TopBlock` from the state, not the one read from the db, as they are always the same. If there is something else that can update the db, I can just read the `TopBlock` from the db.

sennui:
@jur0 I advise to split it in two PRs. Get the protocol injected to candidate and later think about optimization

jur0:
There is #2847 - it doesn't touch the conductor...

PR #2825, opened 2019-10-02
SC: Update documentation for dynamic min_depth calculation

### LINKED ISSUE #2821: SC: Fix documentation of dynamic min depth
The documentation and changelog entry are out of sync with the implementation.

tolbrino:
Fixes #2821

### DIFF
--- a/apps/aechannel/src/aesc_fsm.hrl
+++ b/apps/aechannel/src/aesc_fsm.hrl
@@ -227,8 +227,11 @@
                         | ?DEP_ERR
                         | ?WDRAW_ERR.
 
--opaque opts() :: #{ minimum_depth          => non_neg_integer() %% Defaulted for responder, not for initiator.
-                   , minimum_depth_strategy => txfee
+-type minimum_depth_factor()   :: non_neg_integer().
+-type minimum_depth_strategy() :: txfee.
+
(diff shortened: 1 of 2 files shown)


apps/aechannel/src/aesc_fsm.hrl
```diff
@@ -227,8 +227,11 @@
                         | ?DEP_ERR
                         | ?WDRAW_ERR.
 
--opaque opts() :: #{ minimum_depth          => non_neg_integer() %% Defaulted for responder, not for initiator.
-                   , minimum_depth_strategy => txfee
+-type minimum_depth_factor()   :: non_neg_integer().
+-type minimum_depth_strategy() :: txfee.
```
grzegorz225: Should we add an instant strategy in which we only require only a micro block confirmation?

tolbrino:
> Perhaps it would be good with a table showing example values?

I will put a table with numbers into the documentation in the protocol repo.

apps/aechannel/src/aesc_fsm.hrl
tolbrino: Yes, we should add a `static` strategy where the factor equals the min depth and `0` means only a micro block confirmation is needed.

PR #2837, opened 2019-10-03
Remove protocol_affective_at_height from micro candidate create

jur0:
[PT-168825403](https://www.pivotaltracker.com/n/projects/2124891/stories/168825403)

### DIFF
--- a/apps/aecore/src/aec_block_micro_candidate.erl
+++ b/apps/aecore/src/aec_block_micro_candidate.erl
@@ -101,22 +101,15 @@ int_create(BlockHash, Block, KeyBlock, Trees) ->
     int_create_block(BlockHash, Block, KeyBlock, Trees, Txs).
 
 int_create_block(PrevBlockHash, PrevBlock, KeyBlock, Trees, Txs) ->
+    %% Micro block always has the same protocol version as the last key block.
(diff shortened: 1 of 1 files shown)


apps/aecore/src/aec_block_micro_candidate.erl
```diff
@@ -101,22 +101,15 @@ int_create(BlockHash, Block, KeyBlock, Trees) ->
...
     PrevBlockHeight = aec_blocks:height(PrevBlock),
 
-    %% Assert correctness of last block protocol version, as minimum
-    %% sanity check on previous block and state (mainly for potential
-    %% stale state persisted in DB and for development testing).
-    ExpectedPrevBlockVersion =
-        aec_hard_forks:protocol_effective_at_height(PrevBlockHeight),
-    {ExpectedPrevBlockVersion, _} = {aec_blocks:version(PrevBlock),
-                                     {expected, ExpectedPrevBlockVersion}},
-
```
skkw: Shouldn't this sanity check stay there in some new form?

jur0: All the blocks are added via the conductor process which makes sure each added block is valid and connected to the genesis. So we can rely on the previous key block to be valid and so having the right protocol version.

PR #2848, opened 2019-10-04
Add build switch to skip compilation of rocksdb

### LINKED ISSUE #2843: Rocksdb gets build even when disabled
There is a switch for Windows which should disable building rocksdb, which doesn't work.

tolbrino:
By setting the environment variable `AE_DISABLE_ROCKSDB` to any value one can skip
building rocksdb entirely. This also requires the user to use another
database backend, as trying to use rocksdb would result in a runtime
error.

Fixes #2843

### DIFF
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -2,21 +2,29 @@
 
 IsWindows = case os:type() of {win32, _} -> true; {_, _} -> false end,
 
+FilterRocksDbCheck = (IsWindows orelse os:getenv("DISABLE_ROCKSDB") == "1"),
+
 FilterRelxRocksDb = fun(Apps) ->
-                        case IsWindows of
+                        case FilterRocksDbCheck of
                             false ->
                                 Apps;
                             true ->
-                                Apps -- [rocksdb, mnesia_rocksdb]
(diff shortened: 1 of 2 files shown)


rebar.config.script
```diff
@@ -2,21 +2,29 @@
 
 IsWindows = case os:type() of {win32, _} -> true; {_, _} -> false end,
 
+FilterRocksDbCheck = (IsWindows orelse os:getenv("DISABLE_ROCKSDB") == "1"),
```
dincho: I've 2 propositions:
Adding REBAR prefix seems more appropriate as now it's too generic/gloabal. e.g. `REBAR_DISABLE_ROCKSDB`

If we can ditch the == "1" check, so that it's exists or not would also be nice I guess :)

dincho:
> Help... rebar.lock.script
> ... what are we doing now. No, please don't make live harder than it already is with rebar3

What's the alternatives ? We need to be able to disable rocksdb for ARM builds.

rebar.config.script
tolbrino: Changed.

PR #2859, opened 2019-10-07
Fix fate VM bug

UlfNorell:
Stupid Erlang match semantics...


grzegorz225:
Do we need a release note for this :upside_down_face:

UlfNorell:
I'd say it's enough to post a reply to the forum (https://forum.aeternity.com/t/error-when-performing-dryrun-lima-v5-0-0-rc-2/4840/15) when the fix is released.

PR #2865, opened 2019-10-07
Fix system tests

velzevur:
Here we go again. Fixing 2 more issues

### DIFF
--- a/system_test/common/aest_channels_SUITE.erl
+++ b/system_test/common/aest_channels_SUITE.erl
@@ -95,7 +95,8 @@
     name    => node2,
     peers   => [node1],
     backend => aest_docker,
-    source  => {pull, "aeternity/aeternity:local"}
+    source  => {pull, "aeternity/aeternity:local"},
+    mining => #{autostart => false}
 }).
 
 %=== COMMON TEST FUNCTIONS =====================================================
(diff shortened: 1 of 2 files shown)


system_test/common/aest_channels_SUITE.erl
```diff
@@ -95,7 +95,8 @@
     name    => node2,
     peers   => [node1],
     backend => aest_docker,
-    source  => {pull, "aeternity/aeternity:local"}
+    source  => {pull, "aeternity/aeternity:local"},
+    mining => #{autostart => false}
```
velzevur: If both are mining simultaniously, using the fast mining setup, there is a constant struggle between the two nodes that produces a ton of race conditions.

system_test/common/aest_mempool_SUITE.erl
```diff
@@ -137,7 +137,7 @@ test_mempool_ttl_cleanup(Cfg) ->
 
     %% Check the first one got on the chain
     aest_nodes:wait_for_value({txs_on_chain, [DepositHash1]},
-                              [node1, node2], {blocks_delta, 2}, []),
+                              [node1, node2], {blocks_delta, 5}, []),
```
velzevur: Loosen up the expectation for the deposit to make it in the next 2 generations (with fast mining this can be a fragile assumption)

PR #2874, opened 2019-10-10
Add Iris consensus protocol

### LINKED ISSUE #2873: Add Iris protocol skeleton

dincho:
closes #2873


apps/aecore/src/aec_trees.erl
```diff
@@ -261,6 +261,10 @@ perform_pre_transformations(Trees, _TxEnv, _Protocol, undefined) ->
 perform_pre_transformations(Trees, _TxEnv, Protocol, Protocol) ->
     %% No version change in block.
     Trees;
+perform_pre_transformations(Trees, _TxEnv, Protocol, _PrevProtocol)
+  when Protocol > ?LIMA_PROTOCOL_VSN ->
```
ThomasArts: Here one should just say `Protocol > PrevProtocol` or if one wants to be very strict `Procotol == PrevProtocol + 1` 

But a few lines down, one needs to add something in the style of:
```
apply_pre_transformations(?IRIS_PROTOCOL_VSN, Trees, TxEnv) ->
    %% To be updated for Iris
    Trees.
```

You add `aec_block_fork:apply_iris` and call it.

ThomasArts: Forget that comment, I read the code wrongly. It was correct.

jur0: @ThomasArts I think your comment was right. It'd be nicer to have `apply_pre_transformations(?IRIS_PROTOCOL_VSN, ...)` and just return `Trees` - it'd eliminate a case where we accidentally put a protocol version higher than Iris, which would crash.

dincho: > where we accidentally put a protocol version higher than Iris, which would crash.

Don't we want that (crash) ?

jur0: yes, we want to crash if there is a higher protocol version than Iris. So I'd use `apply_pre_transformations(?IRIS_PROTOCOL_VSN, ...)` - if you put something like Iris + 1 it'd crash.

hanssv: I'm not sure that `perform_pre_transformations` is the place where we have to defend against that type of errors? But why not...

dincho: I'm rethinking this and it was a bit painful to add a new protocol, therefore I think we should keep it as is, so that future me won't be frustrated to go over 22 files to have a new protocol, but 21.

PR #2879, opened 2019-10-11
Refactor fsm message log and reporting

tolbrino:
This fixes the order of log messages and ensures that the chain watcher is
started before performing a dependent action.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -702,157 +702,146 @@ awaiting_signature(cast, {?SIGNED, withdraw_tx, SignedTx} = Msg,
                               D#data{op = #op_ack{ tag = withdraw_tx
                                                  , data = OpData}})))
         end, D);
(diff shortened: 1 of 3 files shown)


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -3202,7 +3185,7 @@ log(Op, Type, M, #data{log = Log0} = D) ->
     D#data{log = log_msg(Op, Type, M, Log0)}.
 
 log_msg(Op, Type, M, Log) ->
-    aesc_window:add({Op, Type, os:timestamp(), M}, Log).
+    aesc_window:add_new({Op, Type, os:timestamp(), M}, Log).
```
uwiger: Given that the added item contains an `os:timestamp()` entry (microsecond timestamp), what are the odds that duplicates will be detected and handled?

tolbrino: Odds are low in this case, but I figured its the right thing to do semantically.

PR #2885, opened 2019-10-12
SC FSM: add optional fees to on-chain transactions

velzevur:
Protocol PR: https://github.com/aeternity/protocol/pull/423

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -1828,7 +1828,7 @@ wdraw_tx_for_signing(Opts, D) ->
 
 close_mutual_tx_for_signing(Opts, D) ->
     Account = my_account(D),
-    new_close_mutual_tx(Opts#{ acct => Account }, D).
+    new_close_mutual_tx(Opts#{acct => Account}, D).
 
 %% @doc The responding side creates a 'throwaway' close_mutual_tx, in order to
 %% validate the tx received from the initiating side. The critical parts to
@@ -1847,7 +1847,7 @@ fake_close_mutual_tx(Mod, Tx, D) ->
 new_close_mutual_tx(Opts, D) ->
     new_onchain_tx_for_signing(channel_close_mutual_tx, Opts, D).
 
-slash_tx_for_signing(Opts, #data{ state = St } = D) ->
+slash_tx_for_signing(Opts, #data{state = St} = D) ->
     {Round, SignedTx} = aesc_offchain_state:get_latest_signed_tx(St),
     slash_tx_for_signing(Round, SignedTx, Opts, D).
 
@@ -1855,7 +1855,7 @@ slash_tx_for_signing(_Round, SignedTx, Opts, D) ->
     Account = my_account(D),
     new_onchain_tx_for_signing(channel_slash_tx,
                                Opts#{ acct => Account
-                                    , payload    => aetx_sign:serialize_to_binary(SignedTx)
+                                    , payload => aetx_sign:serialize_to_binary(SignedTx)
                                     }, D).
 
 settle_tx_for_signing(Opts, D) ->
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -1811,7 +1811,7 @@ create_tx_for_signing(#data{opts = #{ initiator        := Initiator
          , lock_period      => LockPeriod },
     %% nonce is not exposed to the client via WebSocket but is used in tests
     %% shall we expose it to the client as well?
-    Optional = maps:with([nonce], Opts),
+    Optional = maps:with([nonce, fee], Opts),
     new_onchain_tx_for_signing(channel_create_tx,
                                maps:merge(Obligatory, Optional), D).
 
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -30,8 +30,8 @@
 %% API
 -export([ start_link/1
         , attach_responder/2      %% (fsm(), map())
-        , close_solo/1            %% (fsm())
-        , snapshot_solo/1         %% (fsm())
+        , close_solo/2            %% (fsm(), map())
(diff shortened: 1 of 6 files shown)


velzevur:
Still have to add tests for the close solo - slash -settle and a protocol description

tolbrino:
@velzevur Release note is missing.

apps/aehttp/test/aehttp_sc_SUITE.erl
```diff
@@ -110,6 +118,8 @@
 -define(CHECK_INFO(Timeout), check_info(?LINE, Timeout)).
 -define(PEEK_MSGQ, peek_msgq(?LINE)).
 
+-define(RANDOM_BIG_FEE, 123456789876543).
```
uwiger: Isn't this rather 'arbitrary' rather than 'random'? ;-)

apps/aehttp/test/aehttp_sc_SUITE.erl
```diff
@@ -3316,19 +3355,21 @@ sc_ws_withdraw(Config) ->
...
-                      lock_period => 10,
-                      push_amount => 1,
-                      initiator_amount => IAmt,
-                      responder_amount => RAmt,
-                      channel_reserve => 2,
-                      keep_running => false,
-                      protocol => sc_ws_protocol(Config)
-                    }, Other), Config), Config).
+  %% TODO FIX THOSE MAYBES
+    maybe_fee(
```
uwiger: I guess these could be generalized to something like `maybe_set([slogan, fee], maps:merge(...))`, but the benefit would be marginal.

tolbrino: @velzevur You could use a similar approach as I've added here: https://github.com/aeternity/aeternity/blob/master/apps/aechannel/test/aesc_fsm_SUITE.erl#L3821

apps/aehttp/src/sc_ws_api_jsonrpc.erl
```diff
@@ -529,15 +535,21 @@ process_request(#{<<"method">> := <<"channels.message">>,
...
     case aesc_fsm:upd_deposit(FsmPid, XOpts) of
         ok -> no_reply;
         {error, _Reason} = Err -> Err
     end;
 process_request(#{<<"method">> := <<"channels.withdraw">>,
                     <<"params">> := #{<<"amount">>  := Amount} = Props}, FsmPid) ->
     assert_integer(Amount),
-    XOpts = maps:merge(maybe_read_bh(Props), #{amount => Amount}),
+    XOpts = maps:merge(optional_params([ bh_params()
+                                        , fee_params()],
```
tolbrino: Formatting off.

apps/aehttp/src/sc_ws_api_jsonrpc.erl
```diff
@@ -529,15 +535,21 @@ process_request(#{<<"method">> := <<"channels.message">>,
...
+    XOpts = maps:merge(optional_params([ bh_params()
+                                       , fee_params()],
+                                       Props),
+                        #{amount => Amount}),
     case aesc_fsm:upd_deposit(FsmPid, XOpts) of
         ok -> no_reply;
         {error, _Reason} = Err -> Err
     end;
 process_request(#{<<"method">> := <<"channels.withdraw">>,
                     <<"params">> := #{<<"amount">>  := Amount} = Props}, FsmPid) ->
```
tolbrino: In the other functions this value is called `Params`.

apps/aehttp/src/sc_ws_api_jsonrpc.erl
```diff
@@ -529,15 +535,21 @@ process_request(#{<<"method">> := <<"channels.message">>,
 process_request(#{<<"method">> := <<"channels.deposit">>,
                     <<"params">> := #{<<"amount">>  := Amount} = Props}, FsmPid) ->
     assert_integer(Amount),
-    XOpts = maps:merge(maybe_read_bh(Props), #{amount => Amount}),
+    XOpts = maps:merge(optional_params([ bh_params()
```
tolbrino: Same as below.

apps/aehttp/test/aehttp_sc_SUITE.erl
```diff
@@ -3350,6 +3391,16 @@ maybe_slogan(Map, Config) ->
             Map#{slogan => S}
     end.
 
+maybe_fee(#{fee := _} = Map, _) ->
```
tolbrino: If this function stays, it should be renamed since its not a getter, but a setter.

apps/aehttp/test/aehttp_sc_SUITE.erl
velzevur: I honestly forgot about this TODO :)

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1811,7 +1816,7 @@ create_tx_for_signing(#data{opts = #{ initiator        := Initiator
          , lock_period      => LockPeriod },
     %% nonce is not exposed to the client via WebSocket but is used in tests
     %% shall we expose it to the client as well?
-    Optional = maps:with([nonce], Opts),
+    Optional = maps:with([nonce, fee], Opts),
```
uwiger: For consistency, perhaps this could have been changed to `XOpts` as well? :)

tolbrino: Generally this should be cleaned up across the SC code base. This originated from the occasion where there were option, and some extra options I believe. The naming is confusing at least.

apps/aehttp/test/aehttp_sc_SUITE.erl
tolbrino: I see that the other setters are named in a similar fashion, so this should be addresses separately for all of these functions.

apps/aechannel/src/aesc_fsm.erl
velzevur: @uwiger `XOpts` are the accumulation of optional and mandatorary maps of properties. This is only the optional one. I do agree with @tolbrino that this is a bit inconsistent :D

PR #2897, opened 2019-10-14
Added garbage collector for account state nodes

### LINKED ISSUE #2845: Core: introduce GC for state
## Expected Behavior

Database of states doesn't explode

## Actual Behavior

Database of states grows really fast

## Design

To be published. 

It focuses around re-writing trees at given height to new table. The old table holding past states can be dropped.

### LINKED ISSUE #2989: Implement garbage collector of account states
[PR 2897](https://github.com/aeternity/aeternity/pull/2897) implements the GC

skkw:
Closes #2845 
Closes #2989 

Rationality and approach described in this [report](https://github.com/aeternity/papers/blob/master/Garbage%20collector%20for%20account%20state%20data.pdf).

### DIFF
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -1122,7 +1122,7 @@ handle_successfully_added_block(Block, Hash, Events, State, Origin) ->
                     ok; %% Don't spend time when we are the leader.
                 false ->
                     aec_tx_pool:garbage_collect(),
-                    [ aec_db_gc:maybe_garbage_collect() || BlockType == key ]
+                    [ maybe_garbage_collect_accounts() || BlockType == key ]
             end,
             {ok, setup_loop(State2, true, IsLeader, Origin)}
     end.
@@ -1187,3 +1187,18 @@ get_pending_key_block(TopHash, #state{beneficiary = Beneficiary} = State) ->
         {ok, Block} -> {{ok, Block}, State#state{ pending_key_block = Block }};
         {error, _}  -> {{error, not_found}, State}
     end.
+
+
+%% Called only from the conductor.
+%% Putting it here avoids us to clutter aec_db_gc's public API
+%%
+%% To avoid starting of the GC process just for EUNIT
+-ifdef(EUNIT).
+maybe_garbage_collect_accounts() -> nop.
+-else.
+
+%% This should be called when there are no processes modifying the block state
+%% (e.g. aec_conductor on specific places)
+maybe_garbage_collect_accounts() ->
+    gen_statem:call(aec_db_gc, maybe_garbage_collect).
+-endif.
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -726,7 +726,7 @@ preempt_if_new_top(#state{ top_block_hash = OldHash,
 
                     [ aec_keys:promote_candidate(aec_blocks:miner(NewBlock)) || KeyOrNewForkMicro == key ],
 
-                    {changed, NewBlock, create_key_block_candidate(State5)}
+                    {changed, KeyOrNewForkMicro, NewBlock, create_key_block_candidate(State5)}
             end
     end.
 
@@ -1115,10 +1115,15 @@ handle_successfully_added_block(Block, Hash, Events, State, Origin) ->
             {ok, State};
         {micro_changed, State2 = #state{ consensus = Cons }} ->
             {ok, setup_loop(State2, false, Cons#consensus.leader, Origin)};
-        {changed, NewTopBlock, State2} ->
+        {changed, BlockType, NewTopBlock, State2} ->
             IsLeader = is_leader(NewTopBlock),
-            %% Don't spend time when we are the leader.
-            [ aec_tx_pool:garbage_collect() || not IsLeader ],
+            case IsLeader of
+                true ->
+                    ok; %% Don't spend time when we are the leader.
+                false ->
+                    aec_tx_pool:garbage_collect(),
+                    [ aec_db_gc:maybe_garbage_collect() || BlockType == key ]
+            end,
             {ok, setup_loop(State2, true, IsLeader, Origin)}
     end.
 
--- a/apps/aecore/src/aec_db_gc.erl
+++ b/apps/aecore/src/aec_db_gc.erl
@@ -355,9 +355,13 @@ interval(ConfigInterval) -> ConfigInterval.
 
 -ifdef(INTERVAL_VARIANCE).
 interval(ConfigInterval) ->
-    Variance = trunc(ConfigInterval * ?INTERVAL_VARIANCE_PERCENT / 100.0),
-    Delta    = rand:uniform(Variance) - (Variance div 2),
-    max(3, ConfigInterval + Delta).
+    case trunc(ConfigInterval * ?INTERVAL_VARIANCE_PERCENT / 100.0) of
+        0 ->
+            ConfigInterval;
+        Variance when Variance > 0 ->
+            Delta = rand:uniform(Variance) - (Variance div 2),
+            max(3, ConfigInterval + Delta)
+    end.
 -else.
 interval(ConfigInterval) -> ConfigInterval.
 -endif.
--- a/apps/aecore/src/aec_db_gc.erl
+++ b/apps/aecore/src/aec_db_gc.erl
@@ -318,7 +318,8 @@ get_mpt(Height) ->
     try get_mpt_from_hash(Hash) of
         MPT -> MPT
     catch
-        error:{hash_not_present_in_db, MissingHash}:Stacktrace ->
+        error:{hash_not_present_in_db, MissingHash} ->
+            Stacktrace = erlang:get_stacktrace(),
             error({hash_not_present_in_db_at_height, Height, MissingHash, Stacktrace})
     end.
 
--- a/apps/aecore/src/aec_db_gc.erl
+++ b/apps/aecore/src/aec_db_gc.erl
@@ -1,5 +1,20 @@
 -module(aec_db_gc).
 
+%% Implementation of Garbage Collection of Account state.
+%% To run, the GC needs to be `enabled` (config param), and node synced.
+%% If key block `interval` (config parameter) is hit, GC starts collecting
+%% of the reachable nodes from most recent generations (`history` config parameter).
+%%
+%% If #data.enabled and #data.synced are true, #data.hashes indicates the GC status:
+%% - undefined    - not the right time to run GC
+%% - is pid       - we are scanning reachable nodes
+%% - is reference - waiting for signal to execute the `swap_nodes` phase
+%%
+%% GC scan is performed on the background. The only synchronous part of GC is `swap_nodes` phase.
+%% (initiated by aec_conductor on key block boundary)
+%%
+%% More details can be found here: https://github.com/aeternity/papers/blob/master/Garbage%20collector%20for%20account%20state%20data.pdf
+
 -behaviour(gen_statem).
 
 %% API
@@ -29,9 +44,9 @@
 -define(LOG(Fmt), lager:info(Fmt, [])).         % io:format(Fmt"~n")
 -define(LOG(Fmt, Args), lager:info(Fmt, Args)). % io:format(Fmt"~n", Args)
 
-%% %%%===================================================================
-%% %%% API
-%% %%%===================================================================
+%%%===================================================================
+%%% API
+%%%===================================================================
 
 start_link() ->
     #{enabled := Enabled, interval := Interval, history := History} = config(),
@@ -41,10 +56,12 @@ start_link(Enabled, Interval, History) ->
     gen_statem:start_link({local, ?MODULE}, ?MODULE, [Enabled, Interval, History], []).
 
 
+%% To avoid starting of the GC process just for EUNIT
 -ifdef(EUNIT).
 maybe_garbage_collect() -> nop.
 -else.
-%% this should be called when there are no processes modifying the block state
+
+%% This should be called when there are no processes modifying the block state
 %% (e.g. aec_conductor on specific places)
 maybe_garbage_collect() ->
     gen_statem:call(?MODULE, maybe_garbage_collect).
@@ -57,10 +74,9 @@ stop() ->
 %%% gen_statem callbacks
 %%%===================================================================
 
-init([true, Interval, History]) when Interval =< History ->
-    lager:error("GC interval ~p must be greater than history ~p", [Interval, History]),
-    {stop, interval_too_short};
-init([Enabled, Interval, History]) ->
+%% Change of configuration parameters requires restart of the node.
+init([Enabled, Interval, History])
+  when is_integer(Interval), Interval > 0, is_integer(History), History > 0 ->
     if Enabled ->
             aec_events:subscribe(top_changed),
             aec_events:subscribe(chain_sync);
--- a/apps/aecore/src/aec_db_gc.erl
+++ b/apps/aecore/src/aec_db_gc.erl
@@ -1,0 +1,195 @@
+-module(aec_db_gc).
+
+-behaviour(gen_statem).
+
+%% API
+-export([start_link/0,
+         start_link/3,
+         maybe_garbage_collect/0,
+         stop/0]).
+
+%% gen_statem callbacks
+-export([callback_mode/0, init/1, terminate/3, code_change/4]).
+-export([handle_event/4]).
+
+-record(data,
+        {enabled  :: boolean(),             % do we garbage collect?
+         interval :: non_neg_integer(),     % how often (every `interval` blocks) GC runs
+         history  :: non_neg_integer(),     % how many block state back from top to keep
+         synced   :: boolean(),             % we only run GC if chain is synced
+         height   :: non_neg_integer(),     % latest height of MPT hashes stored in tab
+         hashes   :: pid() | ets:table()}). % hashes tab (or process filling the tab 1st time)
+
+-include_lib("aecore/include/aec_db.hrl").
+
+-define(DEFAULT_INTERVAL, 10000).
+-define(DEFAULT_HISTORY, 500).
+
+-define(TIMED(Expr), (timer:tc(fun () -> Expr end))).
+-define(LOG(Fmt), lager:info(Fmt, [])).         % io:format(Fmt"~n")
+-define(LOG(Fmt, Args), lager:info(Fmt, Args)). % io:format(Fmt"~n", Args)
+
+%% %%%===================================================================
+%% %%% API
+%% %%%===================================================================
+
+start_link() ->
+    #{enabled := Enabled, interval := Interval, history := History} = config(),
+    start_link(Enabled, Interval, History).
+
+start_link(Enabled, Interval, History) ->
+    gen_statem:start_link({local, ?MODULE}, ?MODULE, [Enabled, Interval, History], []).
+
+%% this should be called when there are no processes modifying the block state
+%% (e.g. aec_conductor on specific places)
+maybe_garbage_collect() ->
+    gen_statem:call(?MODULE, maybe_garbage_collect).
+
+stop() ->
+    gen_statem:stop(?MODULE).
+
+%%%===================================================================
+%%% gen_statem callbacks
+%%%===================================================================
+
+init([true, Interval, History]) when Interval =< History ->
+    lager:error("GC interval ~p must be greater than history ~p", [Interval, History]),
+    {stop, interval_too_short};
+init([Enabled, Interval, History]) ->
+    if Enabled ->
+            aec_events:subscribe(top_changed),
+            aec_events:subscribe(chain_sync);
+       true ->
+            ok
+    end,
+    Data = #data{enabled  = Enabled,
+                 interval = Interval,
+                 history  = History,
+                 synced   = false,
+                 height   = undefined,
+                 hashes   = undefined},
+    {ok, idle, Data}.
+
+
+%% once the chain is synced, there's no way to "unsync"
+handle_event(info, {_, chain_sync, #{info := {chain_sync_done, _}}}, idle,
+             #data{enabled = true} = Data) ->
+    aec_events:unsubscribe(chain_sync),
+    {keep_state, Data#data{synced = true}};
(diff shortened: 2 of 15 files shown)


apps/aecore/src/aec_db_gc.erl
```diff
@@ -0,0 +1,206 @@
...
+    Header = aec_chain:top_header(),
+    case aec_headers:type(Header) of
+        key ->
+            Height  = aec_headers:height(Header),
+            {ok, _} = range_collect_reachable_hashes(Height, Data),
+            {ok, N} = swap_nodes(Hashes),
+            ets:delete(Hashes),
+            {next_state, idle, Data#data{height = undefined, hashes = undefined},
+             {reply, From, {ok, N}}};
+        micro ->
```
sennui: this should not happen given you added check in conductor?

apps/aecore/src/aec_db_gc.erl
```diff
@@ -0,0 +1,206 @@
...
+    {ok, Hashes} = collect_reachable_hashes_fullscan(FromHeight),
+    {ok, Hashes} = range_collect_reachable_hashes(FromHeight, ToHeight, Hashes),
+    {ok, Hashes}.
+
+collect_reachable_hashes_fullscan(Height) ->
+    Tab = ets:new(gc_reachable_hashes, [public]), %% TODO: remove public
+    MPT = get_mpt(Height),
+    {ok, aeu_mp_trees:visit_reachable_hashes(MPT, Tab, fun store_hash/3)}.
+
+%% assumes Height - 1, Height - 2, ... down to Height - History
```
sennui: If the chain reorganized between the `Height` and `Height - History`, it emits top_change as well for middle blocks? (if not we can miss some blocks in `hashes`)

apps/aecore/src/aec_conductor.erl
```diff
@@ -1114,10 +1114,15 @@ handle_successfully_added_block(Block, Hash, Events, State, Origin) ->
             {ok, State};
         {micro_changed, State2 = #state{ consensus = Cons }} ->
             {ok, setup_loop(State2, false, Cons#consensus.leader, Origin)};
-        {changed, NewTopBlock, State2} ->
+        {changed, BlockType, NewTopBlock, State2} ->
```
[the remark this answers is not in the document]
skkw: I was thinking it's better to have the block type explicit there...

apps/aecore/src/aec_db_gc.erl
```diff
@@ -0,0 +1,206 @@
...
+    ets:foldl(fun ({Hash, Node}, ok) ->
+                      Rec = #aec_account_state{key = Hash, value = Node},
+                      mnesia:write(Rec)
+              end, ok, Hashes).
+
+-spec get_mpt(non_neg_integer()) -> aeu_mp_trees:tree().
+get_mpt(Height) ->
+    {ok, Hash0} = aec_chain_state:get_key_block_hash_at_height(Height),
+    {ok, Trees} = aec_chain:get_block_state(Hash0),
+    AccountTree = aec_trees:accounts(Trees),
```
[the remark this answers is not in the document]
skkw: Yes. It seemed that history of balance changes are the most useless to keep around.

apps/aecore/src/aec_db_gc.erl
```diff
@@ -0,0 +1,206 @@
...
+
+%% assumes Height - 1, Height - 2, ... down to Height - History
+%% are in Hashes from previous runs
+collect_reachable_hashes_delta(Height, Hashes) ->
+    MPT = get_mpt(Height),
+    {ok, aeu_mp_trees:visit_reachable_hashes(MPT, Hashes, fun store_unseen_hash/3)}.
+
+range_collect_reachable_hashes(ToHeight, #data{height = LastHeight, hashes = Hashes}) ->
+    range_collect_reachable_hashes(LastHeight, ToHeight, Hashes).
+range_collect_reachable_hashes(LastHeight, ToHeight, Hashes) ->
```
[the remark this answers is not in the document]
skkw: Yeah, it seems a little inconsistent wrt the function above, but that function above has different arity. 
The `range_collect_reachable_hashes` with 3 params tried to keep the height parameters ordered in from..to fashion.

apps/aecore/src/aec_db_gc.erl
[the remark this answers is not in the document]
hanssv: AENS should be a good candidate for GC pre-claims (and the new auctions) aren't very useful once they have gone stale 🤔

skkw:
Closes #2989

skkw:
closes #2845

apps/aecore/src/aec_db_gc.erl
```diff
@@ -0,0 +1,206 @@
+-module(aec_db_gc).
```
tolbrino: Please add portions of the report as documentation here, since this is also where it will be read.

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -419,6 +419,29 @@
...
+                    "properties" : {
+                        "enabled": {
+                            "description": "If true, node will perform garbage collection of account state, removing unrechable merkle patricia nodes and free some disk space",
+                            "type": "boolean",
+                            "default": true
+                        },
+                        "interval" : {
+                            "description": "How often (every `interval` block) should garbage collection run",
+                            "type": "integer",
+                            "minimum": 50,
```
tolbrino: I'd expect this to be allowed as low as one wants? I tested with a value of 2 which worked fine.

apps/aecore/src/aec_db_gc.erl
```diff
@@ -0,0 +1,206 @@
...
+-endif.
+
+stop() ->
+    gen_statem:stop(?MODULE).
+
+%%%===================================================================
+%%% gen_statem callbacks
+%%%===================================================================
+
+init([true, Interval, History]) when Interval =< History ->
```
tolbrino: Why is this not allowed? If I'd like to check every 10 blocks, but with a history of 500 blocks, that sounds like a valid scenario...

apps/aecore/src/aec_db_gc.erl
```diff
@@ -0,0 +1,206 @@
...
+start_link() ->
+    #{enabled := Enabled, interval := Interval, history := History} = config(),
+    start_link(Enabled, Interval, History).
+
+start_link(Enabled, Interval, History) ->
+    gen_statem:start_link({local, ?MODULE}, ?MODULE, [Enabled, Interval, History], []).
+
+
+-ifdef(EUNIT).
+maybe_garbage_collect() -> nop.
```
tolbrino: It should at least be documented why the call is disabled during testing. Also it would be safer to move this check into the call handling and not the API.

apps/aecore/src/aec_db_gc.erl
```diff
@@ -0,0 +1,206 @@
...
+-export([callback_mode/0, init/1, terminate/3, code_change/4]).
+-export([handle_event/4]).
+
+-record(data,
+        {enabled  :: boolean(),                         % do we garbage collect?
+         interval :: non_neg_integer(),                 % how often (every `interval` blocks) GC runs
+         history  :: non_neg_integer(),                 % how many block state back from top to keep
+         synced   :: boolean(),                         % we only run GC if chain is synced
+         height   :: undefined | non_neg_integer(),     % latest height of MPT hashes stored in tab
+         hashes   :: undefined | pid() | reference()}). % hashes tab (or process filling the tab 1st time)
```
tolbrino: The name `hashes` is misleading. I'd expect a list here, but its actually a reference (to a process or table).

hanssv:
What will happen if I try to dry-run from a state that has been GC:ed? Is this tested? Is it handled gracefully?

tolbrino:
@skkw Please also add a note to the release notes and configuration entry that the GC can leave the node in a inconsistent state which would require a complete re-sync.

apps/aecore/src/aec_db_gc.erl
```diff
@@ -0,0 +1,211 @@
...
+-define(TIMED(Expr), timer:tc(fun () -> Expr end)).
+-define(LOG(Fmt), lager:info(Fmt, [])).         % io:format(Fmt"~n")
+-define(LOG(Fmt, Args), lager:info(Fmt, Args)). % io:format(Fmt"~n", Args)
+
+%% %%%===================================================================
+%% %%% API
+%% %%%===================================================================
+
+start_link() ->
+    #{enabled := Enabled, interval := Interval, history := History} = config(),
```
uwiger: Now, there isn't currently any API for changing config params during run-time (except during testing), but perhaps there should still be a one-line comment noting that the implementation expects the node to be restarted in order to change the GC parameters? Specifically, the fsm never gets out of `synced = false` unless a `chain_sync_done` event is received, which only happens if subscription is enabled, which only happens in `init/1` if `enabled = true` from the start.

apps/aecore/src/aec_db_gc.erl
```diff
@@ -0,0 +1,211 @@
...
+callback_mode() -> handle_event_function.
+
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+
+%% From - To (inclusive)
+collect_reachable_hashes(FromHeight, ToHeight) when FromHeight < ToHeight ->
+    {ok, Hashes} = collect_reachable_hashes_fullscan(FromHeight),
+    {ok, Hashes} = range_collect_reachable_hashes(FromHeight, ToHeight, Hashes),
```
uwiger: Why is `Hashes` expected to be unchanged?

uwiger: > Why is `Hashes` expected to be unchanged?

Presumably because it's actually the reference to the ets table. This could be explicitly stated.

apps/aecore/src/aec_db_gc.erl
```diff
@@ -0,0 +1,211 @@
...
+    {ok, Hashes}.
+
+collect_reachable_hashes_fullscan(Height) ->
+    Tab = ets:new(gc_reachable_hashes, [public]), %% TODO: remove public
```
uwiger: What if there is already an existing ets table?

apps/aecore/src/aec_db_gc.erl
```diff
@@ -0,0 +1,211 @@
...
+             #data{interval = Interval, history = History,
+                   enabled = true, synced = true,
+                   height = undefined, hashes = undefined} = Data)
+  when Height rem Interval == 0 ->
+    Top = self(),
+    Pid = spawn_link(
+            fun () ->
+                    FromHeight = max(Height - History, 0),
+                    {Time, {ok, Hashes}} = ?TIMED(collect_reachable_hashes(FromHeight, Height)),
+                    ets:give_away(Hashes, Top, {{FromHeight, Height}, Time})
```
uwiger: Suggest `Top` -> `Parent`

apps/aecore/src/aec_db_gc.erl
```diff
@@ -0,0 +1,211 @@
...
+            end),
+    {keep_state, Data#data{height = Height, hashes = Pid}};
+handle_event(info, {'ETS-TRANSFER', Hashes, _, {{FromHeight, ToHeight}, Time}}, idle,
+             #data{enabled = true, hashes = Pid} = Data)
+  when is_pid(Pid) ->
+    ?LOG("scanning of ~p reachable hashes in range <~p, ~p> took ~p seconds",
+         [ets:info(Hashes, size), FromHeight, ToHeight, Time / 1000000]),
+    {next_state, ready, Data#data{hashes = Hashes}};
+
+handle_event(info, {_, top_changed, #{info := #{block_type := key, height := Height}}}, ready,
```
uwiger: I suggest that the `top_changed` events are either handled subsequently (not separated by another event), or handled in one callback and distinquished with a `case`.

apps/aecore/src/aec_db_gc.erl
```diff
@@ -0,0 +1,227 @@
...
+range_collect_reachable_hashes(ToHeight, #data{height = LastHeight, hashes = Hashes}) ->
+    range_collect_reachable_hashes(LastHeight, ToHeight, Hashes).
+range_collect_reachable_hashes(LastHeight, ToHeight, Hashes) ->
+    [collect_reachable_hashes_delta(H, Hashes) || H <- lists:seq(LastHeight + 1, ToHeight)],
+    {ok, Hashes}.
+
+%% the actual GC: clear whole table + insert reachable nodes
+swap_nodes(Hashes) ->
+    NodesCount = ets:info(Hashes, size),
+    %% clearing tab can't run in transaction
```
uwiger: It can, if you do something like this:
```
mnesia:transaction(fun() ->
                        {_Mod,Tid,Ts} = get(mnesia_activity_state),
                        mnesia:clear_table(Tid,Ts,Tab,'_')
                   end).
```

skkw: That looks very helpful, going to try that!

tolbrino: Those lovely Mnesia internal APIs :-)

apps/aecore/src/aec_db_gc.erl
skkw: Fixed

apps/aecore/src/aec_db_gc.erl
skkw: Normally I would do it that way, but here the flow is more natural - after top_changed (under specific circumstances), the ets transfer happens, and after that, we process top_changed differently. So it's nice top to bottom reading...

apps/aecore/src/aec_db_gc.erl
skkw: Hmm, since the tab isn't named, it would continue to work but leak the ETS. I'll add name to this ETS, so then it crashes in that case, so GC process restarts. Shouldn't happen as we only do the fullscan when there's no table, but may be nicer for reading.

apps/aecore/src/aec_db_gc.erl
skkw: I've added comment there.

apps/aecore/src/aec_db_gc.erl
skkw: Yes, that's correct. There doesn't seem to be practical utility to support runtime reconfiguration, adding comment there.

skkw:
> What will happen if I try to dry-run from a state that has been GC:ed? Is this tested? Is it handled gracefully?

> @skkw Please also add a note to the release notes and configuration entry that the GC can leave the node in a inconsistent state which would require a complete re-sync.

Adding the comment.

apps/aecore/src/aec_db_gc.erl
skkw: It's allowed in new version.

apps/aecore/src/aec_db_gc.erl
skkw: Matter of preference really :).. I understand the plural indicates any collection (not just list) OR means to create such collection.

apps/aecore/src/aec_db_gc.erl
skkw: Added a comment.

apps/aeutils/priv/aeternity_config_schema.json
skkw: Lowered to 5, for the adventurous :)

apps/aecore/src/aec_db_gc.erl
skkw: Added small section linking to the paper if anybody wants to read it.

apps/aecore/src/aec_db_gc.erl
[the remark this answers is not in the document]
skkw: There may be a time to come back to make it work for AENS, but at the moment AENS DB isn't issue. (or is it?)

apps/aecore/src/aec_db_gc.erl
tolbrino: But in the possible cases the API to interact with the value is different, one is an ETS table, the other one a process. I'm not going to block because of this, but this is clearly an overloaded semantic.

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -428,6 +428,29 @@
...
                     "default" : ["ak_2KAcA2Pp1nrR8Wkt3FtCkReGzAi8vJ9Snxa4PcmrthVx8AhPe8:109"]
+                },
+                "garbage_collection" : {
+                    "type" : "object",
+                    "additionalProperties" : false,
+                    "properties" : {
+                        "enabled": {
+                            "description": "If true, node will perform garbage collection of account state, removing unrechable merkle patricia nodes and free some disk space",
+                            "type": "boolean",
+                            "default": true
```
tolbrino:
```suggestion
                            "default": false```

File: docs/release-notes/next/GH-2845-accounts-state-gc.md [hunk 18]
```diff
@@ -0,0 +1,6 @@
+* Adds garbage collector for removing old account states to free up space on disk.
+  By default the GC is disabled, and does the cleanup every 50000 blocks.
+
+  The cleanup and swap of nodes can not be done transactionally.
```
tolbrino:
```suggestion
  The cleanup and swap of nodes can not be done atomically.
```

apps/aecore/src/aec_db_gc.erl
```diff
@@ -0,0 +1,249 @@
...
+    {WriteTime, ok} =
+        ?TIMED(aec_db:ensure_transaction(
+                 fun () ->
+                         ets:foldl(
+                           fun ({Hash, Node}, ok) ->
+                                   aec_db:write_accounts_node(Hash, Node)
+                           end, ok, Hashes)
+                 end, [], sync_transaction)),
+    ?LOG("writing reachable account state nodes took ~p seconds", [WriteTime / 1000000]),
+    {atomic, DBCount} = mnesia:transaction(fun () -> length(mnesia:all_keys(aec_account_state)) end),
```
uwiger: Ouch! Does this have to be done inside a transaction?
`length(mnesia:dirty_select(aec_account_state, [{'_',[],[1]}]))` ought to be much cheaper.

skkw:
> Did clearing the table inside the transaction context not work out?

No, for some reason only a fraction (1/3, sometime 1/2) of keys to write were written after clear_table.

But it shouldn't be an issue, as we are blocking aec_conductor which drives the writes to the DB.

apps/aecore/src/aec_db_gc.erl
skkw: Shouldn't be needed in transaction, but serves more as a sanity check at least in this phase. I expect that the whole counting will be removed later.

apps/aecore/src/aec_dry_run.erl
```diff
@@ -15,8 +15,11 @@
 dry_run(TopHash, Accounts, Txs) ->
     try setup_dry_run(TopHash, Accounts) of
         {Env, Trees} -> dry_run_(Txs, Trees, Env)
-    catch _E:_R ->
-        {error, <<"Failed to get block state and environment">>}
+    catch
```
ThomasArts: This looks weird... why would the refactoring change the behaviour so much?

apps/aetx/src/aetx_env.erl
```diff
@@ -119,21 +119,30 @@ tx_env_and_trees_from_top(Type) when Type == aetx_contract;
...
+                                {KH, KHash, aec_headers:time_in_msecs(Header)};
+                            key ->
+                                {Header, Hash,
+                                 aec_headers:time_in_msecs(Header) +
+                                     aec_block_micro_candidate:min_t_after_keyblock()}
+                        end,
+                    Env = tx_env_from_key_header(KeyHeader, KeyHash, Time, Hash),
+                    {set_context(Env, Type), Trees}
+            catch
+                error:{hash_not_present_in_db, _} ->
```
ThomasArts: We really, really should never get here, should we?

apps/aecore/src/aec_dry_run.erl
hanssv: The PR introduces new behavior (before we had all old states...) so it is nice if that is signalled to the user when he tries to use a state that is garbage collected!

docs/release-notes/next/GH-2845-accounts-state-gc.md
```diff
@@ -0,0 +1,6 @@
+* Adds garbage collector for removing old account states to free up space on disk.
+  By default the GC is disabled, and does the cleanup every 50000 blocks.
```
ThomasArts: When it is disabled, it should not do the garbage collection at all, not even after 50k blocks

docs/release-notes/next/GH-2845-accounts-state-gc.md
```diff
@@ -0,0 +1,6 @@
...
+
+  The cleanup and swap of nodes can not be done atomically.
```
ThomasArts: What does that mean? Cannot be done atomically?

apps/aetx/src/aetx_env.erl
skkw: It is now possible if the root of MP trie was garbage collected. 
If you want to get a "too old" state, it will contain such root hash for account MP-trie.

apps/aecore/src/aec_dry_run.erl
skkw: It's improvement in a way that we can differentiate between block hash not in DB, and situation where block hash is in DB, but state for that block points to garbage collected root of MP-trie.

docs/release-notes/next/GH-2845-accounts-state-gc.md
skkw: Bad wording, will fix this.

docs/release-notes/next/GH-2845-accounts-state-gc.md
skkw: The non atomicity here means that the modification of the aec_account_state table is done in 2 steps:
1.) clearing of the table
2.) writing of the reachable nodes

If, between 1 and 2 the node exits, the node needs resync. (But considering the GC should run once in 3 months, and the critical section takes 2-3 seconds, the risk is low.)

uwiger: How does one detect that the node needs resync? Is this done automatically?

uwiger: Also, is a sync actually required? Wouldn't a local audit suffice? All the blocks are still there ...

PR #2900, opened 2019-10-15
Rename AENS suffix from aet to crypto

sennui:
Still running tests in parallel.


apps/aens/test/aens_utils_test.erl
```diff
@@ -13,15 +13,15 @@ to_ascii_test() ->
     ?assertEqual({error, no_registrar},
                  aens_utils:to_ascii(<<"abcdefg">>)),
     ?assertEqual({error, multiple_namespaces},
-                 aens_utils:to_ascii(<<"abcd.efgh.aet">>)),
+                 aens_utils:to_ascii(<<"abcd.efgh.crypto">>)),
     ?assertEqual({error, registrar_unknown},
-                 aens_utils:to_ascii(<<"abcd.aettt">>)),
+                 aens_utils:to_ascii(<<"abcd.cryptott">>)),
```
ThomasArts: would actually be cool test to make this "abcd.aet"
```suggestion
                 aens_utils:to_ascii(<<"abcd.aet">>)),
```

sennui: @ThomasArts we could benefit from both (so we exclude some prefix matching error)

PR #2902, opened 2019-10-15
SC: Add missing channel_id in channel_reestablished report

### LINKED ISSUE #2854: SC: channel_id not available in reestablish info message
I would like to get the channel back on reestablish. Remainder: channel_id is required for performing reestablish. 

https://github.com/aeternity/protocol/blob/master/node/api/examples/channels/json-rpc/sc_ws_leave_reestablish.md#responder-----node-15

```
{
  "jsonrpc": "2.0",
  "method": "channels.info",
  "params": {
    "channel_id": null,
    "data": {
      "event": "channel_reestablished"
    }
  },
  "version": 1
}
```

Channel id is available on the next message which contains `"event" : open` but it's still confusing

tolbrino:
Fixes #2854

### DIFF
--- a/apps/aehttp/src/sc_ws_handler.erl
+++ b/apps/aehttp/src/sc_ws_handler.erl
@@ -294,10 +294,9 @@ websocket_info(?ERROR_TO_CLIENT(Err), #handler{protocol = Protocol} = H) ->
     {reply, {text, jsx:encode(Resp)}, H};
 websocket_info(stop, #handler{} = H) ->
     {stop, H};
-websocket_info({aesc_fsm, FsmPid, Msg}, #handler{fsm_pid = FsmPid,
-                                                 enc_channel_id = ChannelId,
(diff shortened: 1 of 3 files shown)


apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -294,10 +294,9 @@ websocket_info(?ERROR_TO_CLIENT(Err), #handler{protocol = Protocol} = H) ->
     {reply, {text, jsx:encode(Resp)}, H};
 websocket_info(stop, #handler{} = H) ->
     {stop, H};
-websocket_info({aesc_fsm, FsmPid, Msg}, #handler{fsm_pid = FsmPid,
-                                                 enc_channel_id = ChannelId,
```
tolbrino: This undefined value was previously used irrespective of the result of the `set_channel_id/2` call.

Issue #2904, opened 2019-10-15
aest_channels_SUITE ==> {{badmatch,{ok,#{<<"info">> => <<"channel_changed">>,


velzevur:
## Expected Behavior
Tests to pass

## Actual Behavior
```

%%% aest_channels_SUITE ==> {{badmatch,{ok,#{<<"info">> => <<"channel_changed">>,
                 <<"tx">> =>
                     <<"tx_+QEDCwH4hLhAYBj2C1P6pUsA6F9JsXf4AvseeWtc5o6ehUM/RRJ2plrG/ZKN+QC3lPK6OPZchu2mvk3GgDOqDM9Fro9qrizrDrhAdCvx9a+1zCOxPJ4ocuLjOv/q4bLBBZDUZPDw9UyelCOE6jTlhwCugo1f5e4zAUKzEiFGy6ipjJEPsqqbeMoSArh5+Hc0AaEGxTb3qHbfgee9CAvFHHTI43pMGcXnCFymREnezjU8U8uhAbG1d7zTJ8s55V5sAmvWp0obNd5sBlDErlHvq3WeQVtmhQukO3QAAIYP1+jpeACg+qkbkYbUeot8Zo8Suvb99JpHnLuu37++QJ5oUGv+f50DAQjjCrs=">>,
                 <<"type">> => <<"channel_withdraw_tx">>}}},
 [{aest_api,sc_close_mutual,2,
            [{file,"/home/circleci/aeternity/_build/system_test+test/extras/system_test/common/helpers/aest_api.erl"},
             {line,166}]},
  {aest_channels_SUITE,simple_channel_test,4,
                       [{file,"/home/circleci/aeternity/_build/system_test+test/extras/system_test/common/aest_channels_SUITE.erl"},
                        {line,214}]},
  {test_server,ts_tc,3,[{file,"test_server.erl"},{line,1539}]},
  {test_server,run_test_case_eval1,6,[{file,"test_server.erl"},{line,1055}]},
  {test_server,run_test_case_eval,9,[{file,"test_server.erl"},{line,987}]}]}
```
[CI](https://circleci.com/gh/aeternity/aeternity/89921#artifacts/containers/0)

[...]

PR #2905, opened 2019-10-15
ST channels test: fix race

velzevur:
Fixes: #2904

### DIFF
--- a/system_test/common/aest_channels_SUITE_data/aeternity.yaml.mustache
+++ b/system_test/common/aest_channels_SUITE_data/aeternity.yaml.mustache
@@ -32,7 +32,7 @@ chain:
     persist: true
 
 mining:
-    autostart: true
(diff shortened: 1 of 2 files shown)


system_test/common/aest_channels_SUITE_data/aeternity.yaml.mustache
```diff
@@ -32,7 +32,7 @@ chain:
     persist: true
 
 mining:
-    autostart: true
+    autostart: {{mining.autostart}}
```
velzevur: Actual fix. Without this, the settings for only `node1` to be mining do not kick in, hence - both `node1` and `node2` are mining simultaneously and are competing for leadership. This plays really bad with fast mining setup

tolbrino: 👍

PR #2907, opened 2019-10-16
SC: Refactor outer fsm API functions

### LINKED ISSUE #2887: SC: Split initiate/respond API functions into new module
These functions represent an API which is used to start the FSM. It is conceptually a layer above the FSM and should be kept in a separate module.

E.g. `aesc_client`.

### LINKED ISSUE #2888: SC: Refactor check accounts logic
Refactor function `check_accounts/3` such that only a single account is checked at a time, to reduce work after a mismatch has been found.

Also error returns from checks for `check_account` are not very clear, need to be somewhat distinct from each other and documented properly.

### LINKED ISSUE #2889: SC: Split account_type function into new module
Extract function `account_type/1` into separate module to make the fsm agnostic to account data structure. This allows the function to be mocked more easily since we don't care about the entire account data structure.

tolbrino:
- Make account checks sequential
- Add new aechannel_eqc module to rebar configuration

Fixes #2887 
Fixes #2888 
Fixes #2889

### DIFF
--- a/apps/aechannel/src/aesc_checks.erl
+++ b/apps/aechannel/src/aesc_checks.erl
@@ -1,0 +1,101 @@
+%% @doc This modules provides stateless checks for various properties throughout
+%% the channels application.
+-module(aesc_checks).
+
+-export([ state_password/1
+        , state_password_in_opts/1
+        , known_role/1
+        , amounts/1
+        , account/1
+        , accounts/3
+        , lock_period/1
+        ]).
+
+-include("aesc_codec.hrl").
+-include("aechannel.hrl").
+
+%% TODO: Make this configurable
+%% No need for a stronger password policy
+%% This check is only here to ensure that someone doesn't enter a 1-2 character password
+-define(STATE_PASSWORD_MINIMUM_LENGTH, 6).
+
+%% ==================================================================
+%% Checks
+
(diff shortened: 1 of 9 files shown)


rebar.config
```diff
@@ -234,6 +234,7 @@
                                     , "eqc/aega_eqc"
                                     , "eqc/aeminer_eqc"
                                     , "eqc/aesophia_eqc"
+                                    , "eqc/aechannel_eqc"
```
tolbrino: @ThomasArts Is this ok?

tolbrino:
@uwiger @velzevur I rebased on `master`.

apps/aechannel/src/aesc_checks.erl
```diff
@@ -0,0 +1,100 @@
...
+
+%% TODO: Make this configurable
+%% No need for a stronger password policy
+%% This check is only here to ensure that someone doesn't enter a 1-2 character password
+-define(STATE_PASSWORD_MINIMUM_LENGTH, 6).
+
+%% ==================================================================
+%% Checks
+
+-spec state_password_in_opts(map()) -> ok | {error, password_required_since_lima | invalid_password}.
```
AleksandarFilipov: I see that code is just moved around, this got my attention, will never return 'password_required_since_lima`

PR #2908, opened 2019-10-17
Upgrade rebar3 to 3.12.0

### LINKED ISSUE #2871: Update rebar3 to v3.12.0
That version includes relx v3.33.0 which includes Windows path fix https://github.com/erlware/relx/pull/730

tolbrino:
Since we don't have any custom changes at the moment,
the binary file has been taken directly from
https://github.com/erlang/rebar3/releases/tag/3.12.0

Fixes #2871


venimus:
Since this includes 'Program Files' bugfix. We may consider distribution of `.exe` too, although users should be advised to put `aeternity.yml` in some alternate place, and upgrading may not be very reliable as it does not remove automatically the old files.

PR #2914, opened 2019-10-18
Fix fate call caller

hanssv:
Correct handling of Call.caller in FATE

When returning from a remote contract call we have
to reset the Caller from the call_stack.


tolbrino:
@hanssv release note missing?

tolbrino:
> Since the State Channels FSM - tests failed _again_ in Minerva I took the opportunity to add a Release note.

👍

PR #2919, opened 2019-10-18
Remove lager_transform when running eqc

tolbrino:
This allows the mocking of calls to `lager`.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -20,13 +20,12 @@
 
 {project_app_dirs, ["apps/*"]}.
 
-{overrides, [{override, [ {erl_opts, [debug_info, {parse_transform, lager_transform},
-                                      {lager_extra_sinks, [ epoch_mining
-                                                          , epoch_metrics
-                                                          , epoch_sync
-                                                          , aestratum
-                                                          ]}]}
-                        ]}]}.
+{erl_opts, [debug_info, {parse_transform, lager_transform},
+            {lager_extra_sinks, [ epoch_mining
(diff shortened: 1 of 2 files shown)


hanssv:
Fancy things with rebar3... On a Friday... 🤕

tolbrino:
> Fancy things with rebar3... On a Friday... 🤕

Its all for the greater good of using eqc :-)

rebar.config
```diff
@@ -229,13 +230,16 @@
...
-                                    ]}]}
+            {eqc, [ {project_app_dirs, ["eqc_test"]}
+                  , {extra_src_dirs, [ "eqc/aecore_eqc"
+                                     , "eqc/aega_eqc"
+                                     , "eqc/aeminer_eqc"
+                                     , "eqc/aechannel_eqc"
+                                     , "eqc/aesophia_eqc"
+                                     , "eqc/aeutils_eqc"
+                                     ]}
+                  , {overrides, [{override, [{erl_opts, [debug_info]}]}]}
```
ThomasArts: But this means that all other QuickCheck tests that do not use lager mocking are failing... that will give me a lot of emails from CI running the QuickCheck task every few hours...

tolbrino: That is correct. I didn't notice that. So we have to configure 2 profiles. (1) `eqc` which works as before; (2) `eqc_lager_mock` which mocks lager.

tolbrino:
Tagging this as wontfix, since it seems obsolete by now.

PR #2924, opened 2019-10-21
Enable HTTP cache for testnet API gateway

IvayloPapazov:
rel: https://www.pivotaltracker.com/story/show/166982041


sennui:
This is intended merge to master - why don't you override the config during deployment? Just asking - maybe we want to have client cache by default on master?

venimus:
I think the issue is confusing, this is not what was intended there. It had to be configuration for the deploy environment, not the default config

PR #2925, opened 2019-10-21
Remove type check for CONS

UlfNorell:
- It always failed for maps
- It only checked tuple sizes and variant arities


apps/aefate/src/aefa_fate_op.erl
```diff
@@ -1697,16 +1697,7 @@ op(map_delete, Map, Key) when ?IS_FATE_MAP(Map),
 op(map_delete, ?FATE_STORE_MAP(Cache, Id), Key) ->
     ?FATE_STORE_MAP(Cache#{ Key => ?FATE_MAP_TOMBSTONE }, Id);
 op(cons, Hd, Tail) when ?IS_FATE_LIST(Tail) ->
-    case ?FATE_LIST_VALUE(Tail) of
```
grzegorz225: This allows to deploy a contract which could create "polymorphic" lists - lists with nonuniform types for instance: [int, list, map]. Is there any place in which an attacker could attack another contract by doing so?

PR #2930, opened 2019-10-22
Catch inner aetx:process crash in aega_meta_tx

hanssv:
QuickCheck finally convinced me that this is a potential issue - with this approach we can detect crashes and convert them into proper errors without changing consensus.


sennui:
Do we need corresponding change in some protocol doc?

sennui:
Don't see how to cancel my review...

I think I approved to fast - is this message broadcasted?

To have consensus all nodes would have to have this message and behave in identical way.

sennui:
Approving - tx lands on chain, so it's safe

hanssv:
I don't think the protocol description is detailed enough to catch this level of change. The Tx was intended to go on chain from the beginning.

PR #2933, opened 2019-10-23
GH 2932 Fix test issues

### LINKED ISSUE #2932: %%% aesc_fsm_SUITE ==> {{badmatch,[{key_block,{key_header
## Expected Behavior
mine blocks functionality

## Actual Behavior
tests fail

## Logs, error output, etc.
https://circleci.com/gh/aeternity/aeternity/93542

https://circleci.com/gh/aeternity/aeternity/93532

velzevur:
Hopefully fixes #2932

### DIFF
--- a/apps/aecore/test/aecore_suite_utils.erl
+++ b/apps/aecore/test/aecore_suite_utils.erl
@@ -287,7 +287,9 @@ mine_blocks(Node, NumBlocksToMine, MiningRate, Opts) ->
 mine_blocks(Node, NumBlocksToMine, MiningRate, Type, Opts) ->
     ok = rpc:call(Node, application, set_env,
                   [aecore, expected_mine_rate, MiningRate], 5000),
-    [] = flush_new_blocks(),
+    %% ensure not mining
+    stopped = rpc:call(Node, aec_conductor, get_mining_state, [], 5000),
+    _ = flush_new_blocks(), %% flush potential hanging message queue messages
     subscribe(Node, block_created),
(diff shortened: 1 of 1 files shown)


tolbrino:
Wouldn't this just crash now instead of proceeding?

velzevur:
@tolbrino this would crash only if we are mining in a parallel, which would be an erronous test anyway

tolbrino:
Then how does that improve anything?

velzevur:
The process that is responsible for mining is the `aec_conductor`. Its mining state can be `running` or `stopped`. We start the mining with `aec_conductor:start_mining/0` and stop it with `aec_conductor:stop_mining/0`. When a new block is mined, the subscribed processes receive a `{gproc_ps_event, block_created | micro_block_created, Info}` that informs them a new block had been mined.

The test setup is counting the received `{gproc_ps_event, block_created | micro_block_created, Info}` messages and then calls the `aec_conductor:stop_mining/0`. There is a small chance that, because of fast mining setup, a block is mined before the stop message is actually processed by the `aec_conductor` process. In this case the mining is indeed stopped but the test process receives an unexpected `{gproc_ps_event, block_created | micro_block_created, Info}` message in its message queue. Later on we try flushing and assertively check that no blocks were mined while flushing. This assertion fails if we have a hanging message in the queue.

The changes in this PR lower the assertion that there was no message in the queue. They also make sure that there are no parallel mining attempts, which is intended to keep the same level of safety, so no lowering

PR #2934, opened 2019-10-23
Align OTP versions with docker image configuration

tolbrino:
This should unblock the test in CI


venimus:
This PR has it too https://github.com/aeternity/aeternity/pull/2935 (lima back-merge)

PR #2941, opened 2019-10-24
Update erlang-lz4 to latest upstream

tolbrino:
The upstream patch has been merged: https://github.com/szktty/erlang-lz4/pull/12


tolbrino:
> Cannot read file '/usr/lib/erlang/lib/parsetools-2.1.5/include/leexinc.hrl': no such file or directory

Will check that out.

PR #2953, opened 2019-10-28
Call db_find_node inside mnesia transaction

### LINKED ISSUE #2951: fix call to mnesia outside the transaction context in chain state
## Expected Behavior

Update chain and state in transactional way

## Actual Behavior

https://github.com/aeternity/aeternity/blob/master/apps/aecore/src/aec_chain_state.erl#L423

The check if we have a node is done outside the transaction context.
We have a case when two nodes slipped into chain state and the second one is failing internal check with found_already_calculated_state.

Result: can't sync any further

## Steps to reproduce

...on the way

jur0:
Closes #2951

### DIFF
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -420,22 +420,21 @@ hash_is_in_main_chain(Hash, TopHash) ->
 %%%-------------------------------------------------------------------
 
 internal_insert(Node, Block, Origin) ->
-    case db_find_node(hash(Node)) of
-        error ->
-            %% To preserve the invariants of the chain,
-            %% Only add the block if we can do the whole
-            %% transitive operation (i.e., calculate all the state
-            %% trees, and update the pointers)
-            Fun = fun() -> internal_insert_transaction(Node, Block, Origin)
-                  end,
-            try aec_db:ensure_transaction(Fun)
-            catch exit:{aborted, {throw, ?internal_error(What)}} -> internal_error(What)
-            end;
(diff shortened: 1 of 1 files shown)


apps/aecore/src/aec_chain_state.erl
```diff
@@ -420,22 +420,21 @@ hash_is_in_main_chain(Hash, TopHash) ->
...
-            catch exit:{aborted, {throw, ?internal_error(What)}} -> internal_error(What)
-            end;
+        error      -> internal_insert_transaction1(Node, Block, Origin);
         {ok, Node} -> ok;
-        {ok, Old} -> internal_error({same_key_different_content, Node, Old})
+        {ok, Old}  -> internal_error({same_key_different_content, Node, Old})
     end.
 
-internal_insert_transaction(Node, Block, Origin) ->
+internal_insert_transaction1(Node, Block, Origin) ->
```
tolbrino: Either this or the calling function should be renamed to capture the true semantics. Just adding a number suffix is not a good idea to get rid of nesting.

jur0:
This can be closed as #2961 will fix the issue.

tolbrino:
> This can be closed as #2961 will fix the issue.

Wouldn't it still be sensible to move the find_node call into the transaction context?

jur0:
`aec_chain_state:insert_block` is only used in `aec_conductor` process when handling a `gen_server` call, so inside this call the db is supposed to be consistent state. I think `db_find_node` doesn't have to be part of the transaction. Maybe @sennui can comment on this?

PR #2954, opened 2019-10-28
Propagate noise session error to the client

### LINKED ISSUE #2892: SC: Handle result of `start_session/2` gracefully
Currently a `badmatch` error will be generated, which is not useful for the client trying to establish the client. Thus, handle the error gracefully and provide a meaningful error message to the caller.

tolbrino:
Previously any error would result in a `badmatch` error.

Moreover, the name for the noise session supervisor module has been
clarified and the init routine of the FSM itself was refactored.

Closes #2892

### DIFF
--- a/apps/aechannel/src/aesc_session_noise.erl
+++ b/apps/aechannel/src/aesc_session_noise.erl
@@ -30,7 +30,7 @@
         , wdraw_signed/2
         , wdraw_locked/2
         , wdraw_error/2
-        , error/2
+        , generic_error/2
         , inband_msg/2
         , leave/2
         , leave_ack/2
@@ -95,7 +95,7 @@ wdraw_created      (Session, Msg) -> cast(Session, {msg, ?WDRAW_CREATED, Msg}).
 wdraw_signed       (Session, Msg) -> cast(Session, {msg, ?WDRAW_SIGNED , Msg}).
 wdraw_locked       (Session, Msg) -> cast(Session, {msg, ?WDRAW_LOCKED , Msg}).
 wdraw_error        (Session, Msg) -> cast(Session, {msg, ?WDRAW_ERR    , Msg}).
-error              (Session, Msg) -> cast(Session, {msg, ?ERROR        , Msg}).
+generic_error      (Session, Msg) -> cast(Session, {msg, ?ERROR        , Msg}).
 inband_msg         (Session, Msg) -> cast(Session, {msg, ?INBAND_MSG   , Msg}).
 leave              (Session, Msg) -> cast(Session, {msg, ?LEAVE        , Msg}).
 leave_ack          (Session, Msg) -> cast(Session, {msg, ?LEAVE_ACK    , Msg}).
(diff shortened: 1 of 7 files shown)


apps/aechannel/src/aesc_session_noise.erl
```diff
@@ -135,152 +167,24 @@ patterns() ->
 
 record_fields(st) -> record_info(fields, st);
 record_fields(_ ) -> no.
-%% ==================================================================
-
-%% Connection establishment
-
-connect(Host, Port, Opts) ->
-    start_link(#{fsm => self(), op => {connect, Host, Port, Opts}}).
```
tolbrino: I changed the connect call to start the process through the supervisor. @uwiger Why wasn't that the case before?

apps/aechannel/src/aesc_session_noise.erl
```diff
@@ -135,152 +167,24 @@ patterns() ->
...
+%% ==================================================================
+%% gen_server API
 
-start_link(Arg) when is_map(Arg) ->
-    %% Res = gen_server:start_link(?MODULE, Arg#{parent => self()}, ?GEN_SERVER_OPTS),
-    %% We don't use gen_server:start_link/3, since we want to use gen_server:enter_loop/3
-    %% in the init/1 function, and we want to avoid redundant `init_ack()` messages
-    %% sent by the `gen_server:init_it/6 function (a different kind of support has been
-    %% introduced in later versions of OTP.)
-    Res = proc_lib:start_link(?MODULE, init, [Arg#{parent => self()}]),
```
tolbrino: I moved from this setup to using straight up `gen_server` with the connection establishing logic being called after the process has finished its `init` routine.

apps/aechannel/src/aesc_session_noise.erl
```diff
@@ -94,36 +146,16 @@ wdraw_created      (Session, Msg) -> cast(Session, {msg, ?WDRAW_CREATED, Msg}).
 wdraw_signed       (Session, Msg) -> cast(Session, {msg, ?WDRAW_SIGNED , Msg}).
 wdraw_locked       (Session, Msg) -> cast(Session, {msg, ?WDRAW_LOCKED , Msg}).
 wdraw_error        (Session, Msg) -> cast(Session, {msg, ?WDRAW_ERR    , Msg}).
-error              (Session, Msg) -> cast(Session, {msg, ?ERROR        , Msg}).
```
tolbrino: This was changed to get rid of the overloading of the `erlang:error` function.

apps/aechannel/src/aesc_session_noise.erl
```diff
@@ -59,10 +71,12 @@
 -type responder_accept_opts() :: #{ reestablish   => false
                                   , initiator := aec_keys:pubkey() | any
                                   , responder := aec_keys:pubkey()
+                                  , lsock       := term()
```
tolbrino: Was required to please dialyzer.

velzevur:
```suggestion
                                  , lsock     := term()
```

PR #2961, opened 2019-10-30
Fixes Make duplicated block inserts silent and not crashing due to incompat…

### LINKED ISSUE #2951: fix call to mnesia outside the transaction context in chain state

sennui:
Fixes #2951

### DIFF
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -431,7 +431,7 @@ internal_insert(Node, Block, Origin) ->
             try aec_db:ensure_transaction(Fun)
             catch exit:{aborted, {throw, ?internal_error(What)}} -> internal_error(What)
             end;
-        {ok, Node} -> ok;
+        {ok, Node} -> {error, already_in_db};
         {ok, Old} -> internal_error({same_key_different_content, Node, Old})
     end.
 
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -1080,18 +1080,18 @@ handle_add_block(Header, Block, State, Origin) ->
             handle_add_block(Block, Hash, Prev, State, Origin)
     end.
 
-handle_add_block(Block, Hash, Prev, State, Origin) ->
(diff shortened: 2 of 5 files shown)


apps/aecore/src/aec_chain_state.erl
```diff
@@ -431,7 +431,7 @@ internal_insert(Node, Block, Origin) ->
             try aec_db:ensure_transaction(Fun)
             catch exit:{aborted, {throw, ?internal_error(What)}} -> internal_error(What)
             end;
-        {ok, Node} -> ok;
+        {ok, Node} -> {ok, #{}};
```
hanssv: On line 444 there is also an incorrect return of `ok`

sennui: yes, but it's not relevant to return of the function

tolbrino: I find the semantics weird. Why not simply return an error like `already_present` or so?

tolbrino:
@sennui Entry in release note is missing.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -431,7 +431,7 @@ internal_insert(Node, Block, Origin) ->
             try aec_db:ensure_transaction(Fun)
             catch exit:{aborted, {throw, ?internal_error(What)}} -> internal_error(What)
             end;
-        {ok, Node} -> ok;
+        {ok, Node} -> {error, already_in_db};
```
hanssv: `internal_insert_transaction` below returns `ok` in one branch - should also be `{error, already_in_db}`?

hanssv: Nevermind, I can't read.

PR #2967, opened 2019-11-01
FATE: Speedup by caching Contract.creator

hanssv:
+ disable unused operation.


apps/aefate/src/aefa_fate_op.erl
```diff
@@ -765,11 +765,22 @@ address(Arg0, EngineState) ->
...
 
 contract_creator(Arg0, EngineState) ->
-    Pubkey  = aefa_engine_state:current_contract(EngineState),
-    API     = aefa_engine_state:chain_api(EngineState),
-    Creator = aefa_chain_api:creator(Pubkey, API),
-    Address = aeb_fate_data:make_address(Creator),
-    write(Arg0, Address, EngineState).
+    Pubkey       = aefa_engine_state:current_contract(EngineState),
+    CreatorCache = aefa_engine_state:creator_cache(EngineState),
+    {CreatorKey, ES1} =
```
tolbrino: Somewhat confusing to move from `EngineState` to `ES1` but ok.

hanssv: Well, when changing an existing module you should follow the style of the rest of the code 🤷‍♂

tolbrino: I saw that and I agree. Still weird.

PR #2966, opened 2019-10-31
GH 2964 SC: tighten up checks

velzevur:
Fixes #2964

PR #2968, opened 2019-11-01
Improve close solo and force progress checks and tests

velzevur:
Replaces #2966, opened agains `releases/lima`


hanssv:
You can just change the base of a PR?! But still looks good.

PR #2979, opened 2019-11-05
Avoid processing (and logging) gossiped blocks while syncing

### LINKED ISSUE #2978: Avoid processing (and logging) uninteresting gossiped information
Currently while syncing the console is filled with `epoch_sync:info`-messages like:
```
16:09:32.958 [info] Couldn't insert received block ({illegal_orphan,<<253,40,11,203,120,254,4,98,220,60,63,90,226,27,103,59,163,138,254,67,19,78,187,14,163,184,7,132,75,212,83,71>>})
...
16:21:16.927 [info] Got invalid light micro_block (<<38,194,93,89,...,11,196,154,37>>): {error,prev_key_block_not_found}
```

This is because of gossiped key-blocks and micro-blocks respectively. Near the end of sync it might be beneficial to start looking at gossiped information, but when the received blocks are thousands of generations into the future they can be ignored.

hanssv:
Closes #2978 

It should even reduce the load of the syncing node, since we skip some processing of the gossiped blocks.

### DIFF
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -1100,6 +1100,9 @@ handle_add_block(Block, Hash, Prev, State, Origin) ->
                 {error, Reason} when Origin == block_created; Origin == micro_block_created ->
                     lager:error("Couldn't insert created block (~p)", [Reason]),
                     {{error, Reason}, State};
+                {error, Reason = {illegal_orphan, H}} ->
+                    lager:info("Couldn't insert received block ({not_attached_to_chain, ~s})", [aeu_debug:pp(H)]),
+                    {{error, Reason}, State};
                 {error, Reason} ->
                     lager:info("Couldn't insert received block (~p)", [Reason]),
                     {{error, Reason}, State}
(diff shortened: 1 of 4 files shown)


apps/aecore/src/aec_conductor.erl
```diff
@@ -1100,6 +1100,9 @@ handle_add_block(Block, Hash, Prev, State, Origin) ->
                 {error, Reason} when Origin == block_created; Origin == micro_block_created ->
                     lager:error("Couldn't insert created block (~p)", [Reason]),
                     {{error, Reason}, State};
+                {error, Reason = {illegal_orphan, H}} ->
+                    lager:info("Couldn't insert received block ({not_attached_to_chain, ~s})", [aeu_debug:pp(H)]),
```
uwiger: Why the double brackets? Wouldn't e.g. `(not_attached_to_chain: ~s)` suffice?

hanssv: Wanted it to looks similar to the one below... Just wanted to get rid of that `illegal_orphan` since it sounded more severe than it is.

PR #2980, opened 2019-11-05
Improve test framework for FATE

UlfNorell:
Mainly handling store in `aefa_sophia_test`

### DIFF
--- a/apps/aefate/test/aefa_sophia_test.erl
+++ b/apps/aefate/test/aefa_sophia_test.erl
@@ -147,7 +147,7 @@ print_run_stats(Time, ES) ->
     GasUsed    = ?CALL_GAS - aefa_engine_state:gas(ES),
     Trace      = aefa_engine_state:trace(ES),
     Red        = fun({_, {reductions, R}}) -> R end,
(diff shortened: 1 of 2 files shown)


apps/aefate/test/aefa_sophia_test.erl
```diff
@@ -109,15 +147,82 @@ print_run_stats(Time, ES) ->
     GasUsed    = ?CALL_GAS - aefa_engine_state:gas(ES),
     Trace      = aefa_engine_state:trace(ES),
     Red        = fun({_, {reductions, R}}) -> R end,
-    Reductions = Red(hd(Trace)) - Red(lists:last(Trace)),
+    Reductions = Red(hd(Trace ++ [{bla, {reductions, 0}}])) - Red(lists:last([{bla, {reductions, 0}} | Trace])),
```
tolbrino: For what is `bla` being used?

UlfNorell: I don't want to crash if `Trace` is empty. As you can see in the line above, the first component of the pair is ignored, `bla` is just a dummy value.

tolbrino: Ah I see, thanks.

PR #2982, opened 2019-11-05
Add protocol activated by miner singnalling to HTTP /status

### LINKED ISSUE #2957: Add protocol to HTTP /status endpoint after it's activated by miner signalling
The `/status` endpoint only returns the scheduled hard forks. After a miner signalled consensus upgrade is successful, it should also return info about this hard fork.

jur0:
Closes #2957.

### DIFF
--- a/apps/aehttp/src/aehttp_dispatch_ext.erl
+++ b/apps/aehttp/src/aehttp_dispatch_ext.erl
@@ -599,9 +599,18 @@ handle_request_('GetStatus', _Params, _Context) ->
     Syncing = aec_sync:is_syncing(),
     SyncProgress = aec_sync:sync_progress(),
     Listening = true, %% TODO
-    Protocols = maps:fold(fun(Vsn, Height, Acc) ->
+    Protocols =
+        maps:fold(fun(Vsn, Height, Acc) ->
                           [#{<<"version">> => Vsn, <<"effective_at_height">> => Height} | Acc]
-                 end, [], aec_hard_forks:protocols()),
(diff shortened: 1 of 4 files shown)


apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -599,9 +599,18 @@ handle_request_('GetStatus', _Params, _Context) ->
...
-    Protocols = maps:fold(fun(Vsn, Height, Acc) ->
+    Protocols =
+        maps:fold(fun(Vsn, Height, Acc) ->
                           [#{<<"version">> => Vsn, <<"effective_at_height">> => Height} | Acc]
-                 end, [], aec_hard_forks:protocols()),
+                  end, [], aec_hard_forks:protocols()),
+    Version = aec_blocks:version(TopKeyBlock),
+    Protocols2 =
+        case aeu_env:get_env(aecore, fork, undefined) of
+            #{version := Version, signalling_end_height := SigEndHeight} ->
```
jur0: If the `Version` is the same as top block's `Version`, it means the new protocol was activated.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -599,9 +599,18 @@ handle_request_('GetStatus', _Params, _Context) ->
...
+                [#{<<"version">> => Version, <<"effective_at_height">> => SigEndHeight} | Protocols];
+            _Other ->
```
tolbrino: Since you already define a default, you should match it here.

PR #2984, opened 2019-11-05
Add database write retry

### LINKED ISSUE #2974: Database write error is not propagated to caller
## Expected Behavior

The database layer should inform the caller that a write error has occurred.

## Actual Behavior

The database reports `ok`, even though the underlying database engine couldn't write to disk. This leads to caller processes to crash because of wrong expectations about the data in the database.

Not reproducible at the moment. This issue has been observed on servers on AWS as well as laptops.

## Logs, error output, etc.

```
2019-11-04 11:55:19.233 [error] <0.27245.4> gen_server aec_conductor terminated with reason: {aborted,{{hash_not_present_in_db,<<143,229,136,89,66,31,166,237,72,113,137,173,48,64,224,160,47,180,182,35,205,242,147,142,152,55,215,248,255,145,185,112>>},[{aeu_mp_trees,db_get,2,[{file,"/home/builder/aeternity/apps/aeutils/src/aeu_mp_trees.erl"},{line,1108}]},{aeu_mp_trees,new,2,[{file,"/home/builder/aeternity/apps/aeutils/src/aeu_mp_trees.erl"},{line,126}]},{aec_trees,deserialize_from_db,1,[{file,"/home/builder/aeternity/apps/aecore/src/aec_trees.erl"},{line,416}]},{aec_db,find_block_state_and_data,...},...]}} in mnesia:wrap_trans/6 line 493
2019-11-04 11:55:19.233 [error] <0.27245.4> CRASH REPORT Process aec_conductor with 0 neighbours exited with reason: {aborted,{{hash_not_present_in_db,<<143,229,136,89,66,31,166,237,72,113,137,173,48,64,224,160,47,180,182,35,205,242,147,142,152,55,215,248,255,145,185,112>>},[{aeu_mp_trees,db_get,2,[{file,"/home/builder/aeternity/apps/aeutils/src/aeu_mp_trees.erl"},{line,1108}]},{aeu_mp_trees,new,2,[{file,"/home/builder/aeternity/apps/aeutils/src/aeu_mp_trees.erl"},{line,126}]},{aec_trees,deserialize_from_db,1,[{file,"/home/builder/aeternity/apps/aecore/src/aec_trees.erl"},{line,416}]},{aec_db,find_block_state_and_data,...},...]}} in mnesia:wrap_trans/6 line 493
```

tolbrino:
If the database encounters a write problem during execution of a
transaction the operation will exit and subsequently take down the node.
This change introduces the option to retry transactions to handle short
temporary disk write issues. At worst the previous behaviour will occur
after all retries have failed, at best the problem doesn't persist and
the write is successful after a retry.

The number of retries can be configured in the user configuration.

Closes #2974 

Depends on https://github.com/aeternity/mnesia_rocksdb/pull/15

### Microbenchmark Results

These are the raw numbers, I didn't bother to eliminate some high outliers (especially possible with on disc tests).

#### Before PR

```
Throughput testing results (in microseconds) in ram

# of blocks inserted	= 100000
Total runtime		= 3319927

Min	= 25
Max	= 10329
Mean	= 33
Mode	= 124
Range	= 10304
Median	= 32

Throughput testing results (in microseconds) on disc

# of blocks inserted	= 100000
Total runtime		= 9356080

Min	= 64
Max	= 4883
Mean	= 93
Mode	= 4883
Range	= 4819
Median	= 85
```

#### With PR

```
Throughput testing results (in microseconds) in ram

# of blocks inserted	= 100000
Total runtime		= 3449935

Min	= 30
Max	= 8999
Mean	= 34
Mode	= 124
Range	= 8969
Median	= 33

Throughput testing results (in microseconds) on disc

# of blocks inserted	= 100000
Total runtime		= 12208141

Min	= 75
Max	= 47134
Mean	= 122
Mode	= 4843
Range	= 47059
Median	= 104
```


tolbrino:
@uwiger @hanssv Opinions?

apps/aecore/src/aec_db.erl
```diff
@@ -860,3 +854,24 @@ default_dir() ->
...
+%% operation will not be retried upon failure. If retries is greater than 0, the
+%% operation will be retried upon failure and the retry counter will be
+%% decremented.
+try_activity(Type, Fun, Retries) when Retries =< 0 ->
+    mnesia:activity(Type, Fun);
+try_activity(Type, Fun, Retries) ->
+    try
+        mnesia:activity(Type, Fun)
+    catch
+        exit:Reason ->
```
gorillainduction: This is very dangerous. If you catch unqualified stuff like this inside of a mnesia transaction, you disrupt mnesia's internal retry mechanism.

Probably this is all right, but if this happens to be called inside of a (wrapped) transaction, this will get the wrong retry semantics and might corrupt data.

tolbrino: Good point. Since this function is called out of `ensure_transaction` which does check that it is not executed inside another transaction, I figured it should be ok. Or should I check here as well to prevent future misuse?

gorillainduction: You are correct. This is currently safe, at least in the transaction case. I would add a comment about this rather than a check, and skip the `try_activity/2` to make it more obvious that this should not be used from any call point.

gorillainduction:
I did not check now, but if I recall correctly, the user config thingie might end up reading from disk. Did you check the overhead of adding such a read on every transaction?

tolbrino:
> I did not check now, but if I recall correctly, the user config thingie might end up reading from disk. Did you check the overhead of adding such a read on every transaction?

No, I assumed there is some caching going on. But now that you mention it that assumption seems too easy. I will test it.

gorillainduction:
Btw, @tolbrino, these are just drive-by-comments on something that caught my eye.

You will have to look elsewhere for approval of this PR ;-)

tolbrino:
> Btw, @tolbrino, these are just drive-by-comments on something that caught my eye.
> 
> You will have to look elsewhere for approval of this PR ;-)

Haha, I get it. Even then feedback is appreciated.

tolbrino:
I have verified that reading the configuration variable doesn't add considerable overhead. The configuration file is not re-read, only in-memory reads of the configuration values are performed.

uwiger:
So in the updated `mnesia_rocksdb`, the default behavior when encountering a low-level write error is to log it using `mnesia_lib:[dbg_out|verbose|warning]/2`. This will not cause a transaction abort.

Is the assumption that the transaction will fail when the object is read back?

tolbrino:
@hanssv @uwiger I adapted the PR according to the feedback. Reviews needed.

tolbrino:
@hanssv @skkw I have to adapt this PR yet again a bit. @uwiger highlighted some shortcomings which require tighter assertions for the writes, which I shall add. I will ping you once this is ready for reviews again.

tolbrino:
@uwiger @hanssv @gorillainduction I improved the PR as requested and added a GC for the error store.

apps/aecore/src/aec_db_error_store.erl
```diff
@@ -0,0 +1,109 @@
...
+
+%% API
+-export([ start_link/0
+	, stop/0
+        , check/1
+	]).
+
+%% gen_server callbacks
+-export([ init/1
+        , handle_call/3
```
ThomasArts: are there tabs. inyour layout here?

tolbrino:
> Icannot say I fully oversee the consequences of these changes. It looks solid, but it is deep down in the core.
> 
> The tests give some guarantee that things work better. Did you do a sync against mainnet to see if anything unexpected pops-up?

I only did so in testnet. Will repeat that for mainnet.

tolbrino:
@ThomasArts mainnet syncs completely.

apps/aecore/src/aec_db_error_store.erl
tolbrino: That might be, will fix that.

Issue #2962, opened 2019-10-31
aehttp_sc_SUITE ==> plain.with_open_channel.sc_ws_update_abort: FAILED timeout


tolbrino:
## Expected Behavior

Tests pass.

## Actual Behavior

```
%%% aehttp_sc_SUITE ==> plain.with_open_channel.sc_ws_update_abort: FAILED
%%% aehttp_sc_SUITE ==> {{timeout,{messages,[{<0.14644.0>,websocket_event,channel,conflict,
                      #{<<"jsonrpc">> => <<"2.0">>,
                        <<"method">> => <<"channels.conflict">>,
                        <<"params">> =>
                            #{<<"channel_id">> =>
                                  <<"ch_2mYFVhAbGMgwQPyPuHS9prC7WmKLMFyy89cqqtRR6CZ9kjcYDA">>,
                              <<"data">> =>
                                  #{<<"channel_id">> =>
                                        <<"ch_2mYFVhAbGMgwQPyPuHS9prC7WmKLMFyy89cqqtRR6CZ9kjcYDA">>,
                                    <<"error_code">> => 2,
                                    <<"error_msg">> => <<"conflict">>,
                                    <<"round">> => 5}},
                        <<"version">> => 1}}]}},
 [{aehttp_ws_test_utils,wait_for_msg,5,
                        [{file,"/home/builder/aeternity/apps/aehttp/test/aehttp_ws_test_utils.erl"},
                         {line,316}]},
  {aehttp_sc_SUITE,wait_for_channel_event_,3,
                   [{file,"/home/builder/aeternity/apps/aehttp/test/aehttp_sc_SUITE.erl"},
                    {line,3676}]},
  {aehttp_sc_SUITE,wait_for_channel_event_match,4,
                   [{file,"/home/builder/aeternity/apps/aehttp/test/aehttp_sc_SUITE.erl"},
                    {line,3641}]},
  {aehttp_sc_SUITE,channel_abort_sign_tx,4,
                   [{file,"/home/builder/aeternity/apps/aehttp/test/aehttp_sc_SUITE.erl"},
                    {line,509}]},
  {aehttp_sc_SUITE,sc_ws_update_abort,1,
                   [{file,"/home/builder/aeternity/apps/aehttp/test/aehttp_sc_SUITE.erl"},
                    {line,3096}]},
  {test_server,ts_tc,3,[{file,"test_server.erl"},{line,1755}]},
[...]

PR #2993, opened 2019-11-06
Improve slash tests

velzevur:


velzevur:
Test roma failed with an issue already tracked in #2962, restarting the job

PR #2996, opened 2019-11-07
SC: improve force progress stability

velzevur:


apps/aechannel/src/aesc_channels.erl
```diff
@@ -20,7 +20,8 @@
          serialize_for_client/1,
          close_solo_last_onchain/3,
          close_solo_with_payload/4,
-         force_progress/6,
+         force_progress_latest_onchain/6,
```
hanssv: `last_onchain` vs. `latest_onchain`

velzevur: Addressed

PR #2999, opened 2019-11-08
New transaction - PayingForTx

### LINKED ISSUE #2851: Add `PayingForTX` transaction
The idea surfaced in the forum [Allow transaction fee being payed by another account](https://forum.aeternity.com/t/allow-transaction-fee-being-payed-by-another-account/3887)

In short, we should add a transaction (more or less just a wrapping transaction) that allows an account to sign its intention to pay for another transaction. 

```
Tx   = SpendTX(from = A, to = B, amount = 100, ...)
STx  = Sign(PrivKeyA, Tx)
Tx2  = PayingForTX(paying = C, tx = STx)
STx2 = Sign(PrivKeyC, Tx2)
```

This should be possible to combine with #2849 to allow for intricate paying-for schemes

hanssv:
Allows an account to pay for a transaction on behalf of someone else.
Paying for fees and gas cost, will not cover the amount spent by the
transaction, just "the cost of the transaction" (and the extra size
added by wrapping the original transaction).

Note: this changes how "gas" for transaction size is calculated for GAs
from IRIS onwards. Instead of the outer/wrapping TX paying the size-gas
the whole TX it just has to cover its own part of the TX. This should be
less confusing, and easier to reason about. 

Closes #2851

### DIFF
--- a/apps/aechannel/src/aesc_utils.erl
+++ b/apps/aechannel/src/aesc_utils.erl
@@ -212,7 +212,7 @@ check_solo_close_payload(ChannelPubKey, FromPubKey, Nonce, Fee, Payload,
         {error, _} = E -> E;
         {ok, [Channel, last_onchain]} ->
             Checks =
-                [fun() -> aetx_utils:check_account(FromPubKey, Trees, Nonce, Fee, Env) end,
+                [fun() -> check_account(FromPubKey, Trees, Nonce, Fee, Env) end,
                  fun() -> check_is_peer(FromPubKey, aesc_channels:peers(Channel)) end,
                  fun() -> check_is_active(Channel) end,
                  fun() -> check_root_hash_in_channel(Channel, PoI) end,
@@ -221,7 +221,7 @@ check_solo_close_payload(ChannelPubKey, FromPubKey, Nonce, Fee, Payload,
             aeu_validation:run(Checks);
         {ok, [Channel, {SignedState, PayloadTx}]} ->
             Checks =
-                [ fun() -> aetx_utils:check_account(FromPubKey, Trees, Nonce, Fee, Env) end,
+                [ fun() -> check_account(FromPubKey, Trees, Nonce, Fee, Env) end,
                   fun() -> check_is_active(Channel) end,
                   fun() -> check_payload(Channel, PayloadTx, FromPubKey, SignedState,
                                           Trees, Env, solo_close) end,
@@ -239,7 +239,7 @@ check_slash_payload(ChannelPubKey, FromPubKey, Nonce, Fee, Payload,
             {error, slash_must_have_payload};
         {ok, [Channel, {SignedState, PayloadTx}]} ->
             Checks =
-                [ fun() -> aetx_utils:check_account(FromPubKey, Trees, Nonce, Fee, Env) end,
+                [ fun() -> check_account(FromPubKey, Trees, Nonce, Fee, Env) end,
                   fun() -> check_is_closing(Channel) end,
                   fun() -> check_payload(Channel, PayloadTx, FromPubKey, SignedState,
                                          Trees, Env, slash) end,
@@ -250,9 +250,8 @@ check_slash_payload(ChannelPubKey, FromPubKey, Nonce, Fee, Payload,
 
 check_force_progress(Tx, Payload, OffChainTrees, Trees, Env) ->
     Protocol = aetx_env:consensus_version(Env),
-    _Height = aetx_env:height(Env),
     ?TEST_LOG("Checking force progress:\nTx: ~p,\nPayload: ~p,\nOffChainTrees: ~p,\nHeight: ~p",
-              [Tx, Payload, OffChainTrees, _Height]),
+              [Tx, Payload, OffChainTrees, aetx_env:height(Env)]),
     ChannelPubKey = aesc_force_progress_tx:channel_pubkey(Tx),
     FromPubKey = aesc_force_progress_tx:origin(Tx),
     Nonce = aesc_force_progress_tx:nonce(Tx),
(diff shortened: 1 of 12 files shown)


hanssv:
Associated documentation https://github.com/aeternity/protocol/pull/443

apps/aechannel/src/aesc_utils.erl
```diff
@@ -734,30 +752,30 @@ process_solo_close(ChannelPubKey, FromPubKey, Nonce, Fee,
                    Payload, PoI, Height, Trees, Env) ->
     add_event(
       process_solo_close_slash(ChannelPubKey, FromPubKey, Nonce, Fee,
-                               Payload, PoI, Height, Trees),
+                               Payload, PoI, Height, Trees, Env),
```
ThomasArts: Height is already part of Env, but that optimization is not part of this PR.

apps/aecore/src/aec_governance.erl
```diff
@@ -158,7 +158,8 @@ tx_base_gas(name_update_tx, _)            -> ?TX_BASE_GAS;
 tx_base_gas(oracle_extend_tx, _)          -> ?TX_BASE_GAS;
 tx_base_gas(oracle_query_tx, _)           -> ?TX_BASE_GAS;
 tx_base_gas(oracle_register_tx, _)        -> ?TX_BASE_GAS;
-tx_base_gas(oracle_response_tx, _)        -> ?TX_BASE_GAS.
+tx_base_gas(oracle_response_tx, _)        -> ?TX_BASE_GAS;
+tx_base_gas(paying_for_tx, _)             -> ?TX_BASE_GAS div 5.
```
ThomasArts: Somewhere we should record the ratio behind this 1/5th.

apps/aecore/src/aec_paying_for_tx.erl
```diff
@@ -0,0 +1,213 @@
...
+
+%%%===================================================================
+%%% Internal functions
+
+check_tx(STx) ->
+    Tx = aetx_sign:tx(STx),
+    {Type, InnerTx} = aetx:specialize_type(Tx),
+    case Type of
+        paying_for_tx                   -> {error, not_allowed_to_pay_for_recursively};
+        ga_meta_tx                      -> check_tx(aega_meta_tx:tx(InnerTx));
```
ThomasArts: May the ga_meta_tx in its turn contain a paying_for_tx ? Then there is some indirect recursion.

apps/aega/test/aega_SUITE.erl
```diff
@@ -418,17 +434,23 @@ basic_minimum_fee(_Cfg) ->
...
-    %% inner transaction is paid for by the meta transaction.
+    %% Up until IRIS:
+    %%   When the transaction is wrapped in a meta transaction, the size of
+    %%   the inner transaction is paid for by the meta transaction.
+    %% From IRIS:
+    %%   The inner transaction should cover its own "size-cost"
     SizeFee = aetx:size(DummyTx) * aec_governance:byte_gas() * MinGP,
-    MinimumInnerFee = MinimumSpendFee - SizeFee,
+    MinimumInnerFee = case Protocol =< ?LIMA_PROTOCOL_VSN of
+                        true  -> MinimumSpendFee - SizeFee;
```
ThomasArts: This is not new code, but just wonder whether this could ever become a negative MinimumInnerFee?

apps/aeprimop/src/aeprimop.erl
```diff
@@ -215,7 +216,7 @@ oracle_extend_tx_instructions(Pubkey, DeltaTTL, Fee, Nonce) ->
 oracle_query_tx_instructions(OraclePubkey, SenderPubkey, Query,
                              QueryFee, QTTL, RTTL, TxFee, Nonce) ->
     [ force_inc_account_nonce_op(SenderPubkey, Nonce)
-    , spend_fee_op(SenderPubkey, TxFee + QueryFee)
+    , spend_fee_op(SenderPubkey, TxFee, QueryFee)
```
ThomasArts: Ah, Query fee still has to be paid by inner Tx

apps/aeprimop/src/aeprimop.erl
```diff
@@ -566,13 +578,31 @@ lock_namefee(Kind, From, Amount, #state{protocol = Protocol} = S) ->
...
-
-spend_fee({From, Amount}, #state{} = S) when is_integer(Amount), Amount >= 0 ->
-    {Sender1, S1}   = get_account(From, S),
-    assert_account_balance(Sender1, Amount),
-    {ok, Sender2}   = aec_accounts:spend_without_nonce_bump(Sender1, Amount),
-    put_account(Sender2, S1).
+    {spend_fee, {From, Amount, 0}}.
+
+spend_fee_op(From, DelegatedAmount, Amount)
+        when ?IS_HASH(From), ?IS_NON_NEG_INTEGER(DelegatedAmount + Amount) ->
```
ThomasArts: Actually, one would like to enforce the stronger ?IS_NON_NEG_INTEGER(DelegateAmount), ?IS_NON_NEG_INTEGER(Amount).

apps/aeprimop/src/aeprimop.erl
```diff
@@ -1122,17 +1164,20 @@ ga_attach({OwnerPubkey, GasLimit, GasPrice, ABIVersion,
...
-    assert_account_balance(Account, TotalAmount),
+    assert_account_balance(Account, OwnerAmount),
     assert_ga_create_version(ABIVersion, VMVersion, S),
     Code           = assert_ga_attach_byte_code(ABIVersion, SerializedCode,
                                                 CallData, AuthFun, S),
     %% Charge the fee, the gas (the unused portion will be refunded)
     %% and the deposit (stored in the contract) to the contract owner (caller),
     CTVersion      = #{vm => VMVersion, abi => ABIVersion},
-    S2             = account_spend(Account, TotalAmount, S1),
+    S2             = if OwnerAmount > 0 -> account_spend(Account, OwnerAmount, S1);
```
ThomasArts: I have a hard time to see when the OwnerAmount == 0 kicks in? Let alone <0.

ThomasArts:
A future feature request is to disable to possibilty to take the inner transaction and post it. As it is now, someone can promisse to pay for a transaction, but if the signer of that transaction actually has an account with tokens, the inner transaction can just be posted and inner transaction signer is charged for fee and gas.

apps/aecore/src/aec_governance.erl
hanssv: How about here: https://github.com/aeternity/protocol/pull/443 (you have already approved it!)

apps/aecore/src/aec_paying_for_tx.erl
hanssv: No, that is exactly what we are checking here...

apps/aega/test/aega_SUITE.erl
hanssv: No it can't. Since `SizeFee` is part of `MinimumSpendFee`.

apps/aeprimop/src/aeprimop.erl
hanssv: Yes, the `QueryFee` does not go to the miner so it should be payed by the TX-actor.

apps/aeprimop/src/aeprimop.erl
hanssv: OwnerAmount can't be `< 0` and if it is `0` the spending `0` is a no-op so we skip that.

apps/aeprimop/src/aeprimop.erl
hanssv: Yes 👍

hanssv:
Fixed Thomas' comment on asserts in `aeprimop` and rebased.

Did nothing about the potential problem that the inner TX is a valid transaction, and confirmed that `aetx_sign:serialize_for_client_inner` is correct for `aec_paying_for_tx`.

apps/aecore/src/aec_governance.erl
tolbrino: There should be an inline comment as well.

hanssv:
We have a plan for making the inner transaction invalid (in the post to mempool sense) by signing something slightly different (probably dropping the network id or similar). But let's postpone this to a separate PR/Issue.

PR #3008, opened 2019-11-13
add vm_version to fate's engine state record

### LINKED ISSUE #3007: Add vm_version to Fate's engine state
It's useful to know what version of the VM is running the contract code at the aefa_fate_op level.

skkw:
closes #3007

### DIFF
--- a/apps/aefate/src/aefa_engine_state.erl
+++ b/apps/aefate/src/aefa_engine_state.erl
@@ -30,6 +30,7 @@
         , memory/1
         , stores/1
         , trace/1
+        , vm_version/1
         ]).
 
 %% Setters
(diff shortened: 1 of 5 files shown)


apps/aefate/src/aefa_engine_state.erl
```diff
@@ -6,7 +6,7 @@
 %%%-------------------------------------------------------------------
 -module(aefa_engine_state).
 
--export([ new/6
+-export([ new/7
```
hanssv: I think we need an access function as well ;-)

skkw: right!

apps/aefate/src/aefa_engine_state.erl
```diff
@@ -110,14 +111,15 @@
                                    %% Call stack of contracts (including tail calls)
             , stores            :: aefa_stores:store()
             , trace             :: list()
+            , vm_version        :: non_neg_integer()
```
tolbrino: The type seems reasonable but is not enforced anywhere. Is that the intention?

PR #3012, opened 2019-11-14
Add aec_db write throughput test

tolbrino:
Ref #2974 

I'd like to merge this test before functional changes regarding the linked issue in order to see performance overhead of any changes.

### DIFF
--- a/apps/aecore/test/aec_db_tests.erl
+++ b/apps/aecore/test/aec_db_tests.erl
@@ -108,9 +108,9 @@ write_chain_test_() ->
 
                ok
        end},
-      {"Throughput test building chain with 100 blocks",
+      {"Throughput test building chain with 1000 blocks",
        fun() ->
-               TotalBlockCount = 100,
+               TotalBlockCount = 1000,
 
                %% Setup
                [GB | Blocks] = aec_test_utils:gen_blocks_only_chain(TotalBlockCount + 1),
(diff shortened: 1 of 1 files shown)


tolbrino:
For reference these numbers are the result at https://circleci.com/gh/aeternity/aeternity/99362 :

```
Throughput testing results (in microseconds):

# of block inserted	= 100
Total runtime		= 5435

Min	= 34
Max	= 83
Mean	= 54
Mode	= 73
Range	= 49
Median	= 53
```

dincho:
What's the benefit having this run for each build ? 🤔

tolbrino:
> What's the benefit having this run for each build ? 🤔

I'd like to get at least a rough indication whether any change adds 100% overhead to database operations. Once can certainly do more extensive testing manually which is more geared towards a specific change, but this minor test is a good rough guide. We've done something similar in the state channels CT suite.

dincho:
> I'd like to get at least a rough indication whether any change adds 100% overhead to database operations. Once can certainly do more extensive testing manually which is more geared towards a specific change, but this minor test is a good rough guide. We've done something similar in the state channels CT suite.

That's OK, but then let's add 100% time increase assert then, so that one can notice it at all :)

tolbrino:
> 
> That's OK, but then let's add 100% time increase assert then, so that one can notice it at all :)

For that to work you'd need some form of stateful statistics store which you can query against to compare the current result. Would be nice, but out of scope.

sennui:
I really like these kind of checks. Given that eunit_lima is not significantly longer I am all for it

hanssv:
👍

tolbrino:
@dincho I'll keep your feedback in mind as a feature request.

PR #3014, opened 2019-11-15
Back-merge releases/lima

venimus:
Sync master with 5.1.0


dincho:
This is not the way to sync, because as you see the release CI workflow is applied to this PR.

You must branch from master then merge releases/lima into that new branch and open PR agains master with the new branch.

dincho:
Moreover, this process is kind of wrong at all. You must actually merge master into releases/lima and then release from releases/lima

venimus:
> Moreover, this process is kind of wrong at all. You must actually merge master into releases/lima and then release from releases/lima

This is done already
Ok will make separate PR

PR #3025, opened 2019-11-21
aeo_oracles:serialize_for_client should encode FATE_ABI data

### LINKED ISSUE #3021: Oracle FATE queries need encoding like AEVM queries
Oracle queries and responses are send as binaries encoded in a certain way. At the moment FATE queries are just arbitrary binaries, whereas AEVM binaries are encoded as contract bytecode.

```
  EncFormat = fun(F) -> if ABIVersion == ?ABI_AEVM_SOPHIA_1 ->
                                 aeser_api_encoder:encode(contract_bytearray, F);
                             true ->
                                 F
                          end end,
```
https://github.com/aeternity/aeternity/blob/082a7d47a9f71e90cfaca9ee934cc9cca73b2762/apps/aeoracle/src/aeo_oracles.erl#L176

The FATE ABI version is different and an encoding is needed.

hanssv:
Fixes #3021

### DIFF
--- a/apps/aeoracle/src/aeo_oracles.erl
+++ b/apps/aeoracle/src/aeo_oracles.erl
@@ -173,7 +173,8 @@ serialize_for_client(#oracle{id              = Id,
                              ttl             = TTL,
                              abi_version     = ABIVersion
                             }) ->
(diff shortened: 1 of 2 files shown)


hanssv:
> Release note entry is missing, since its a user-visible change.

There are no such oracles, so no one will notice. But I've added a release note anyway.

PR #3029, opened 2019-11-26
Log peer connection count periodically

### LINKED ISSUE #2168: Number of peers needed in logs
Would be useful to see additional info like number of peers, chain height, blocks to go in info log entry `Synced blocks 3611 - 3684`.

jur0:
Closes #2168

### DIFF
--- a/apps/aecore/src/aec_peers.erl
+++ b/apps/aecore/src/aec_peers.erl
@@ -80,7 +80,7 @@
 -define(DEFAULT_RESOLVE_MAX_RETRY,             7).
 -define(DEFAULT_RESOLVE_BACKOFF_TIMES,
         [5000, 15000, 30000, 60000, 120000, 300000, 600000]).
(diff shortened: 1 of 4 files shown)


apps/aecore/src/aec_peers.erl
```diff
@@ -80,6 +80,7 @@
 -define(DEFAULT_RESOLVE_MAX_RETRY,             7).
 -define(DEFAULT_RESOLVE_BACKOFF_TIMES,
         [5000, 15000, 30000, 60000, 120000, 300000, 600000]).
+-define(LOG_PEER_COUNT_TIMEOUT,             5000).
```
hanssv: I think this should be configurable?! 🤔

jur0: Everything in sync is configurable, so I made this interval configurable too.

PR #3034, opened 2019-11-26
Add number of peer connections to HTTP /status

### LINKED ISSUE #3033: Add number of peer connections to HTTP /status
The number of returned peers is not the number of currently connected peers, but the number of all known peers there have ever been present in the network.

jur0:
Now it returns the sum of inbound and outbound connections.

Closes #3033


hanssv:
While I agree that this is vastly more useful than the old number, I can give a heads up warning that there will be a bit of confusion since this number is used to report (an incorrect value for) "the size of the network"...

dincho:
It's debatable if this is a fix, maybe adding an additional field would be better approach to show the connected peers together with the current number (of ever seen peers in the network)

PR #3037, opened 2019-11-26
Alternative accounts GC approach using partial cleaning

tolbrino:
The goal of this PR is to only perform database operations which are guaranteed to be performed atomically. Therefore, GC operations are only deleting a subset of entries. The GC on the initial dataset will take longer, but be safer.


tolbrino:
Not needed anymore.

Issue #2850, opened 2019-10-04
Add "fancy" crypto-primitives to FATE/Sophia


hanssv:
It isn't feasible to implement advanced crypto-primitives directly in Sophia. Instead we should provide users with a good set of primitives that can act as a build-blocks in smart contracts (for example to be used in GA authentication functions).

This includes but isn't limited to Zero-knowledge proofs, multisignature schemes, etc.

@knarz to collect a wish-list, and suggested libraries - then we need to wrap in Erlang and add to FATE/Sophia.

hanssv:
#3043 is the first bunch of these - documentation is missing...

And possibly more flavors of crypto.

PR #3043, opened 2019-11-27
GH-2850 Add FATE support for BLS12-381 crypto operations

hanssv:
Depends on https://github.com/aeternity/aebytecode/pull/88
and https://github.com/aeternity/aesophia/pull/187

Refs #2850


hanssv:
The `emcl` dependency doesn't yet build on Windows, so WIP. cc @tolbrino

.circleci/config.yml
```diff
@@ -863,6 +863,8 @@ jobs:
             brew install libsodium erlang@20
             brew link libsodium
             brew link --force erlang@20
+            brew install gmp erlang@20
```
dincho: why not adding it to the install row above ? this looks like copy-paste nonsense to me 🤔

.circleci/config.yml
```diff
@@ -863,6 +863,8 @@ jobs:
...
+            brew link --force gmp erlang@20
```
dincho: erlang is already linked above, also do you really need force?

Dockerfile
```diff
@@ -14,7 +14,7 @@ FROM ubuntu:18.04
 COPY --from=builder /app/_build/prod/rel/aeternity /home/aeternity/node
 
 # OpenSSL is shared lib dependency
-RUN apt-get -qq update && apt-get -qq -y install libssl1.0.0 curl libsodium23 \
+RUN apt-get -qq update && apt-get -qq -y install libssl1.0.0 curl libsodium23 libgmp-dev \
```
dincho: could please try a runtime package only without -dev?

hanssv: @dincho I blame @tolbrino for everything, just doing what he told me :-)

hanssv: I guess it should be `libgmp10` then? 🤔

hanssv: Fixed

PR #3044, opened 2019-11-27
Gh3009 swagger validation

ThomasArts:
Using a webservice to validate our swagger specification.

### DIFF
--- a/apps/aehttp/test/aehttp_spec_SUITE.erl
+++ b/apps/aehttp/test/aehttp_spec_SUITE.erl
@@ -80,19 +80,34 @@ validate_api(Config) ->
     Spec = json_from_yaml(Config),
 
     Url = "http://validator.swagger.io/validator/",
-    {ok, {_, _Headers, Body}} = httpc:request(post, {Url, [],  "application/json", Spec}, [], []),
-    %% For manual verification visit https://github.com/swagger-api/validator-badge
-    %% A picture is returned in the body
-    case aec_hash:blake2b_256_hash(list_to_binary(Body)) of
-        <<85,250,111,235,97,122,213,232,25,130,236,210,99,16,245,146,67,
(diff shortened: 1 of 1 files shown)


apps/aehttp/test/aehttp_spec_SUITE.erl
```diff
@@ -73,3 +75,31 @@ get_api(Config) ->
...
     ok.
+
+validate_api(Config) ->
+    Spec = json_from_yaml(Config),
+
+    Url = "http://validator.swagger.io/validator/",
+    {ok, {_, _Headers, Body}} = httpc:request(post, {Url, [],  "application/json", Spec}, [], []),
+    %% For manual verification visit https://github.com/swagger-api/validator-badge
+    %% A picture is returned in the body
+    case aec_hash:blake2b_256_hash(list_to_binary(Body)) of
```
hanssv: Compairing against a picture? Would it be an alternative to use the `debug/` endpoint instead and not have to rely on that badge picture?

apps/aehttp/test/aehttp_spec_SUITE.erl
```diff
@@ -73,3 +75,31 @@ get_api(Config) ->
     {ok,{{"HTTP/1.1",405,"Method Not Allowed"},_,_}} =
          httpc:request(post, {URL, [], Type, Body}, [], []),
     ok.
+
+validate_api(Config) ->
+    Spec = json_from_yaml(Config),
+
+    Url = "http://validator.swagger.io/validator/",
+    {ok, {_, _Headers, Body}} = httpc:request(post, {Url, [],  "application/json", Spec}, [], []),
```
hanssv: Maybe handling some more responses (and indicate that there is a connection problem/remote problem)?

PR #3045, opened 2019-11-28
Remove python tests

### LINKED ISSUE #3009: Retire Python UAT suite
~~Currently python2.7 is used, which will be EOL in early 2020. Thus, we should update to the latest LTS version of python.~~

~~This requires changes in the CI configuration as well and must be tested on all platforms.~~

The Python UAT suite shall be retired. The tests need to be inspected for cases which are not handled by the systems tests yet. Such cases shall be ported to the system tests.

Some tests covered by the Python UAT shall be handled outside of the new UAT, see #3065 and https://github.com/aeternity/aeminer/issues/17

ThomasArts:
Closes #3009

Blocked by #3062


tolbrino:
I've rebased this PR on latest master and removed some more artifacts.

PR #3060, opened 2019-12-11
Remove duplicate check for block presence in DB

jur0:
The check is [here](https://github.com/aeternity/aeternity/blob/master/apps/aecore/src/aec_chain_state.erl#L423), so no need to have it twice.

### DIFF
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -1082,31 +1082,29 @@ handle_add_block(Header, Block, State, Origin) ->
 
 handle_add_block(Block, Hash, Prev, #state{top_block_hash = TopBlockHash} = State, Origin) ->
     epoch_mining:debug("trying to add block (hash=~w, prev=~w)", [Hash, Prev]),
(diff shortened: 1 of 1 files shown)


apps/aecore/src/aec_conductor.erl
```diff
@@ -1082,31 +1082,29 @@ handle_add_block(Header, Block, State, Origin) ->
...
-    case aec_chain:has_block(Hash) of
-        true ->
-            epoch_mining:debug("Block (~p) already in chain when top is (~p) [conductor]", [Hash, TopBlockHash]),
+    %% Block validation is performed in the caller's context for
+    %% external (gossip/sync) blocks and we trust the ones we
+    %% produce ourselves.
+    case aec_chain_state:insert_block(Block, Origin) of
+        {ok, Events}  ->
+            handle_successfully_added_block(Block, Hash, Events, State, Origin);
+        {pof,_PoF,Events} ->
```
hanssv:
```suggestion
        {pof, _PoF, Events} ->
```

PR #3062, opened 2019-12-11
Prepare test support on win32 for multi-node CT tests

tolbrino:
This PR fixes issues with the setup of CT suites. The goal is to be able to run CT suites on win32 as needed. It is not complete, but improvements will be handled in another PR.

This PR also include general fixes for the win32 scripts.

Refs #3009


.circleci/windows/build.cmd
```diff
@@ -22,10 +22,12 @@ IF NOT "%ARTIFACTS_PATH%"=="" FOR /f %%i IN ('cygpath -a %ARTIFACTS_PATH%') DO S
 
 IF "%PACKAGES_PATH%"=="" SET "PACKAGES_PATH=%PROJECT_ROOT%/packages"
 
-call "%~dp0vcvarsall" || exit /b %ERRORLEVEL%
+call "%~dp0vcvarsall"
+IF ERRORLEVEL 1 exit /b %ERRORLEVEL%
```
venimus:
```suggestion
IF %ERRORLEVEL% NEQ 0 exit /b %ERRORLEVEL%
```

.circleci/windows/build.cmd
```diff
@@ -22,10 +22,12 @@ IF NOT "%ARTIFACTS_PATH%"=="" FOR /f %%i IN ('cygpath -a %ARTIFACTS_PATH%') DO S
...
 
 @call:log Build production release
-"%WIN_MSYS2_ROOT%\usr\bin\bash.exe" -lc "${PROJECT_ROOT}/.circleci/windows/build.sh" || exit /b %ERRORLEVEL%
+"%WIN_MSYS2_ROOT%\usr\bin\bash.exe" -lc "${PROJECT_ROOT}/.circleci/windows/build.sh"
+IF ERRORLEVEL 1 exit /b %ERRORLEVEL%
```
venimus:
```suggestion
IF %ERRORLEVEL% NEQ 0 exit /b %ERRORLEVEL%
```

.circleci/windows/test.cmd
```diff
@@ -38,7 +38,8 @@ IF "%TEST_STEPS%"=="" SET "TEST_STEPS=release"
 
 @for /f "tokens=1* delims=, " %%i in ("%TEST_STEPS%") do @(
     @call:log Run test %%i
-	call :TEST_%%i || exit /b %ERRORLEVEL%
+        call :TEST_%%i
+        IF ERRORLEVEL 1 exit /b %ERRORLEVEL%
```
venimus:
```suggestion
        IF %ERRORLEVEL% NEQ 0 exit /b %ERRORLEVEL%
```

.circleci/windows/vcvarsall.cmd
```diff
@@ -21,7 +21,8 @@
 @FOR /F "tokens=* USEBACKQ delims=" %%F IN (`where /r "%ProgramFiles(x86)%\Microsoft Visual Studio" vcvarsall`) DO SET "VCVARSALL=%%F"
 :VCVARSALLFOUND
 
-call "%VCVARSALL%" %PLATFORM% || exit /b %ERRORLEVEL%
+call "%VCVARSALL%" %PLATFORM%
+IF ERRORLEVEL 1 exit /b %ERRORLEVEL%
```
venimus:
```suggestion
IF %ERRORLEVEL% NEQ 0 exit /b %ERRORLEVEL%
```

.circleci/windows/build.cmd
venimus: I think this is more appropriate

venimus: Seems I missed the fact that ERRORLEVEL is resolved before the actual execution
so it had to be
```
SETLOCAL EnableDelayedExpansion

command-1 || exit /b !ERRORLEVEL!
```
However it requires to have setlocal which is not put everywhere on purpose (because it creates a new SET scope and thus can not work with the global scope, which is required on few places)

tolbrino: Correct. But I also like the check on a separate line as it is more explicit. I am fine with using your suggestions.

tolbrino:
> Not really into windows scripts, rebar lock check refactoring looks OK to me

I'm not into it either.

PR #3066, opened 2019-12-12
Gh 2988 sc fix nonces

### LINKED ISSUE #2988: SC: node enters non recoverable mode
rerunning scenario [close on chain](https://github.com/aeternity/ae-channel-service/blob/reproduce_%232988/apps/ae_socket_connector/test/ae_socket_connector_test.exs#L200)

fires the attached node crash

Subsequent connects are possible but expected socket messages are missing.

***Short description of scenario***
1. connect
2. transfer from initiator to responder
3. gets poi
4. initiator post a `close_solo_tx`
5. initiator/responder expects which **doesn't** arrive :on_chain, 0, :transient, "solo_closing" 
6. reconnect is possible, however messages are not delivered as expected.

Intermittent failing test case (isolated) can be found here
https://circleci.com/gh/aeternity/ae-channel-service/88

channel service reference:
https://github.com/aeternity/ae-channel-service/tree/reproduce_%232988

testcase
https://github.com/aeternity/ae-channel-service/blob/reproduce_%232988/apps/ae_socket_connector/test/ae_socket_connector_test.exs#L200-L260

***crash***

[crash_on_chain_ops.log](https://github.com/aeternity/aeternity/files/3814615/crash_on_chain_ops.log)

**Below is same or similar issue/ kept as reference**

clients are able to reconnect but shortly after initiator gets `funding_signed` the connection gets `channels info, died` 

[aeternity_4.log](https://github.com/aeternity/aeternity/files/3809827/aeternity_4.log)
[crash_4.log](https://github.com/aeternity/aeternity/files/3809828/crash_4.log)

*error:{badmatch,{error,nonce_too_high}*

velzevur:
Fixes #2988 

Issue was that the FSM was not prepared to handle a transaction failing to be posted in the pool. What exposed the test was the Elixir code that was producing competing transactions sharing the same nonce.

The same ussue put a bit more abstract would be:
1. Alice initiates an on-chain channel transacion, deposit for example
1. Alice's FSM prepares the transaction for her and puts what it consideres to be the next correct nonce for Alice
1. Alice authenticates it, gives it back to her FSM, which sends it to Bob's FSM which prompts Bob to co-authenticate, which Bob happily does and the now co-authenticated deposit goes back to Alice's FSM
1. Alice produces a transaction with the same nonce and posts in a pool and it is included in a block
1. Alice's FSM tries posting the transaction but fails because it is now invalid. What happened is that the FSM crashed in flames, which arguably is not the expected UX. 

With this fix, Alice will get notified, so will Bob. Bob's FSM, protecting his interest, will also try posting the transaction on-chain. If it fails as well - we are safe, indeed it is invalid.

----
This will be further improved with allowing Alice to optionally specify her desired nonce. This is the next thing on my plate.

### DIFF
--- a/apps/aechannel/src/aesc_codec.hrl
+++ b/apps/aechannel/src/aesc_codec.hrl
@@ -64,8 +64,9 @@
 -define(CHANNEL_CLOSED, channel_closed).
 
 %% Error codes
--define(ERR_VALIDATION, 1).
--define(ERR_CONFLICT  , 2).
--define(ERR_TIMEOUT   , 3).
--define(ERR_ABORT     , 4).
+-define(ERR_VALIDATION        , 1).
+-define(ERR_CONFLICT          , 2).
+-define(ERR_TIMEOUT           , 3).
+-define(ERR_ABORT             , 4).
+-define(ERR_ONCHAIN_REJECTED  , 5).
 -define(ERR_USER      , 128).  % anything >= 128 is a user error
--- a/apps/aecore/test/aecore_suite_utils.erl
+++ b/apps/aecore/test/aecore_suite_utils.erl
@@ -290,7 +290,7 @@ mine_all_txs(Node, MaxBlocks) ->
     case rpc:call(Node, aec_tx_pool, peek, [infinity]) of
         {ok, []} -> {ok, []};
         {ok, Txs} ->
-            TxsHs = [ aeser_api_encoder:encode(tx_hash, aetx_sign:hash(T)) || T <- Txs],
+            TxsHs = [ aetx_sign:hash(T) || T <- Txs],
             mine_blocks_until_txs_on_chain(Node, TxsHs, MaxBlocks)
     end.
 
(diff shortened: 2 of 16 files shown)


apps/aechannel/src/aesc_codec.hrl
```diff
@@ -64,8 +64,9 @@
...
 %% Error codes
--define(ERR_VALIDATION, 1).
--define(ERR_CONFLICT  , 2).
--define(ERR_TIMEOUT   , 3).
--define(ERR_ABORT     , 4).
+-define(ERR_VALIDATION        , 1).
+-define(ERR_CONFLICT          , 2).
+-define(ERR_TIMEOUT           , 3).
+-define(ERR_ABORT             , 4).
+-define(ERR_ONCHAIN_REJECTED  , 5).
```
uwiger: This should have a corresponding clause in `aesc_fsm:error_code_to_msg/1` (line 363)

apps/aecore/test/aecore_suite_utils.erl
```diff
@@ -309,29 +309,30 @@ mine_blocks_until_txs_on_chain(Node, TxHashes, MiningRate, Max, Opts) ->
...
+            {ok, Blocks} = mine_safe_setup(Node, expected_mine_rate(), #{}, Fun),
+            [key, micro] = [aec_blocks:type(B) || B <- Blocks],
+
+            case rpc:call(Node, aec_tx_pool, peek, [infinity]) of
+                {ok, []} ->
+                    {ok, Blocks};
+                {ok, [_|_]} ->
+                    %% So, Tx(s) is/are back in the mempool, this means (unless some Txs arrived
+                    %% from thin air) that we had a micro-fork. Let's check what state we stopped in
+                    {ok, NewBlocks} = mine_safe_setup(Node, expected_mine_rate(), #{}, Fun),
```
hanssv: So we make it less likely that we run into this problem, but we don't fix it?!

apps/aecore/test/aecore_suite_utils.erl
```diff
@@ -343,20 +344,22 @@ mine_blocks_until_txs_on_chain_loop(_Node, _TxHashes, 0, _Acc) ->
 mine_blocks_until_txs_on_chain_loop(Node, TxHashes, Max, Acc) ->
     case mine_blocks_loop(1, key) of
         {ok, [Block]} -> %% We are only observing key blocks
+            ct:log("mined a key block", []),
```
hanssv: Is this really helpful? It will be many, many log entries...

apps/aecore/test/aecore_suite_utils.erl
```diff
@@ -343,20 +344,22 @@ mine_blocks_until_txs_on_chain_loop(_Node, _TxHashes, 0, _Acc) ->
...
-            case txs_not_on_chain(Node, Block, TxHashes) of
+            TxsNotOnChain = TxHashes -- txs_on_chain(Node, TxHashes),
+            case TxsNotOnChain of
                 []        -> {ok, NewAcc};
                 TxHashes1 -> mine_blocks_until_txs_on_chain_loop(Node, TxHashes1, Max - 1, NewAcc)
             end;
         {error, _} = Error -> Error
     end.
 
-txs_not_on_chain(Node, Block, TxHashes) ->
```
hanssv: This is quite a drastic change... I don't have the energy to figure out what is the problem with the existing approach? 

It will mean that we have to change the input to `mine_txs` everywhere?

Issue #3070, opened 2019-12-16
Add metrics to improve visibility of network state


tolbrino:
- Current difficulty
- Block propagation time
- Inter-block time
- fix the non-working micro-forks metric

Follow-up #3079

ThomasArts:
Current difficulty is, as far as I can see, directly available from endpoint v2/status.

Tx round-trip time has been discussed in chat and can mean 2 things:
1) Time from posting a transaction to receiving it via gossip
2) Confirmation time: time from posting to being included in a micro-block.

Number 1) is tricky. If you post a transaction and get it returned, you could get it returned from your neighbouring node. There is no guarantee that more than you and your neighbour actually have received the Tx... We have no hop-count in gossip messages.

Number 2 is almost in place. We post monitoring transactions that contain a timestamp in the payload. I believe that for example account: ak_zvU8YQLagjcfng7Tg8yCdiZ1rpiWNp1PBn3vtUs44utSvbJVR is ours and has posted a payload looking like:
```183982:kh_JwhH8QCTtJyoLZ4Zo6DtkVfXUPQSJ7bqBzbcAPtnM7zazmxiS:mh_2AohoUhCChdtoNSZPYsjHimAmY2VPH5aU8N9VJWi9gDHf9hrPC:15765745514Ú_d```
The first number is the key-block it has observed, the last number 15765745514 is posix timestamp.  This timestamp can be compared with the micro-block time stamp we find it in.

ThomasArts:
Block propagation time can be a metric locally. A node can compare the incoming key-block (or micro-block) and compare the timestamp in that block with local time. This is not guaranteed to be the real propagation time, it could even be negative if clocks drift or miners are not completely honest, but it's an indication and relatively easy to add.

dincho:
> Number 1) is tricky. If you post a transaction and get it returned, you could get it returned from your neighbouring node. There is no guarantee that more than you and your neighbour actually have received the Tx... We have no hop-count in gossip messages.

Maybe we should report the RTT of the transaction for each peer that it's received from, this way we can have some more details and more meaningful numbers ? 🤔

tolbrino:
Closing this as more metrics are handled in a different task.

PR #3071, opened 2019-12-17
Adds new metrics to ae.epoch.aemon namespace

tolbrino:
- `block.propagation_time.key`   : Time until key-blocks reached this node in milliseconds
- `block.propagation_time.micro` : Time until micro-blocks reached this node in milliseconds
- `block.time_since_prev.key`    : Time between key-blocks in milliseconds
- `block.time_since_prev.micro`  : Time between micro-blocks in milliseconds
- `chain.top.difficulty`         : Difficulty of the top block
- `forks.micro.count`             : Count of observed micro-forks
- `forks.micro.height`             : Height difference of observed micro-forks

Refs #3070

### DIFF
--- a/apps/aemon/src/aemon_metrics.erl
+++ b/apps/aemon/src/aemon_metrics.erl
@@ -1,64 +1,37 @@
+%% @doc This module provides helper functions which mask exometer and
+%% aec_metric calls for the user.
 -module(aemon_metrics).
 
+%% generic API
 -export([create/1]).
 
+%% specific metrics
 -export([ fork_micro/0
         , confirmation_delay/1
-        ]).
--export([ gen_stats_tx/1
+        , gen_stats_tx/1
         , gen_stats_tx_monitoring/1
         , gen_stats_microblocks/1
-        ]).
--export([ publisher_balance/1
+        , publisher_balance/1
         , publisher_post_tx/1
-        ]).
--export([ queue_size/1
+        , queue_size/1
         , ttl_expired/1
+        , block_propagation_time/2
+        , block_time_since_prev/2
+        , chain_top_difficulty/1
         ]).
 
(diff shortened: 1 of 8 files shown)


apps/aemon/src/aemon_mon_on_chain.erl
```diff
@@ -33,70 +52,84 @@ code_change(_FromVsn, St, _Extra) ->
...
+    case [B || B <- Blocks, not lists:member(B, BlocksOnChain1)] of
+        [] ->
+            %% All blocks are in the generation.
+            ok;
+        BlocksNotOnChain ->
+            lager:debug("Monitoring observed micro-fork, missing blocks in generation ~p = ~p",
+                        [Height, BlocksNotOnChain]),
+            ok = aemon_metrics:fork_micro(length(BlocksNotOnChain))
+    end,
+    St#st{micro_blocks = []}.
```
sennui: So, we collect microblocks and on a key block we zero out them. Also we cross check the generation with our accumulator to see how long is microfork. Looks OK.

apps/aemon/src/aemon_metrics.erl
```diff
@@ -46,42 +80,35 @@ publisher_balance(Balance) ->
...
+%% ==================================================================
+%% internal functions
 
 create(Name, Type) ->
-    exometer:ensure(metric(Name), Type, []),
-    ok.
+    create(Name, Type, []).
+
+create(Name, Type, Opts) ->
+    ok = exometer:ensure(metric(Name), Type, Opts).
```
ThomasArts: Are you sure you ant to crash here in case we cannot create the metric?

tolbrino: I agree we could just log a failure. I just found the previous approach of ignoring it entirely out of place.

PR #3090, opened 2019-12-19
Check max number of gossiped peers

### LINKED ISSUE #3086: Enforce max number of peers in ping message
Max number of peers should be 32 and the node should check there are no more than that.

jur0:
Closes #3086 

There is no `aec_peer_connection_tests` so no test for this change...

### DIFF
--- a/apps/aecore/src/aec_peer_connection.erl
+++ b/apps/aecore/src/aec_peer_connection.erl
@@ -56,6 +56,7 @@
 -define(DEFAULT_NOISE_HS_TIMEOUT, 5000).
 -define(DEFAULT_CLOSE_TIMEOUT, 3000).
 %% The number of peers sent in ping message.
+-define(MAX_GOSSIPED_PEERS_COUNT, 32).
 -define(DEFAULT_GOSSIPED_PEERS_COUNT, 32).
 
 %% gen_server callbacks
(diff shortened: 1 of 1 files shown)


jur0:
Yes, it's configurable and the maximum is 32: https://github.com/aeternity/aeternity/blob/master/apps/aeutils/priv/aeternity_config_schema.json#L137

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1225,6 +1236,11 @@ gossiped_peers_count() ->
                                aecore, gossiped_peers_count,
                                ?DEFAULT_GOSSIPED_PEERS_COUNT).
 
+max_gossipep_peers_count() ->
```
uwiger:
```suggestion
max_gossiped_peers_count() ->
```

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -712,10 +712,21 @@ handle_ping_msg(S, RemotePingObj) ->
...
 decode_remote_ping(#{ genesis_hash := GHash,
                       best_hash    := THash,
+                      share        := Share,
                       peers        := Peers,
                       difficulty   := Difficulty,
                       sync_allowed := SyncAllowed}) ->
-    {ok, SyncAllowed, GHash, THash, Difficulty, Peers};
+    case length(Peers) =< Share of
+        true ->
+            case Share =< max_gossipep_peers_count() of
```
uwiger:
```suggestion
            case Share =< max_gossiped_peers_count() of
```

Issue #3079, opened 2019-12-18
Add more metrics to improve visibility of network state


tolbrino:
- Gas price per Tx
- Contract state size, size of data, gas spent
- Tx round-trip time
- Gas per microblock
- Txs per microblock
- Tx size
- Block execution time
- Tx round-trip time

PR #3094, opened 2019-12-19
Added new metrics to cover more block and transaction information

tolbrino:
- `ae.epoch.aemon.block.tx.total.micro`             : Number of transactions in a microblock
- `ae.epoch.aemon.block.gas.total.micro`            : Gas used per microblock
- `ae.epoch.aemon.block.gas.per_tx.micro`           : Gas used per transaction in a microblock
- `ae.epoch.aecore.blocks.micro.txs_execution_time.success` : Execution time of all transaction in a microblock in microseconds
- `ae.epoch.aecore.blocks.micro.txs_execution_time.error` : Execution time of all transaction in a microblock in microseconds, when an error was encountered
- `ae.epoch.aecore.blocks.micro.insert_execution_time.success` : Execution time of insertion of a microblock in microseconds
- `ae.epoch.aecore.blocks.micro.insert_execution_time.error` : Execution time of failed insertion of a microblock in microseconds
- `ae.epoch.aecore.blocks.key.insert_execution_time.success` : Execution time of insertion of a keyblock in microseconds
- `ae.epoch.aecore.blocks.key.insert_execution_time.error` : Execution time of failed insertion of a keyblock in microseconds

Refs #3079

### DIFF
--- a/apps/aecore/priv/exometer_predefined.script
+++ b/apps/aecore/priv/exometer_predefined.script
@@ -1,29 +1,40 @@
 %% -*- erlang-indent-level: 4; indent-tabs-mode: nil -*-
 %%
-%% Predefined metrics for the aecore app
+%% Predefined metrics for the aecore app.
 %% Note that exometer entries can be added dynamically as well.
 %% See the docs at https://github.com/Feuerlabs/exometer_core
-%% for information about how to create, update and read metrics
-[
- {[ae,epoch,aecore,mining,blocks_mined]      , counter     , []},
- {[ae,epoch,aecore,mining,retries]           , counter     , []},
- {[ae,epoch,aecore,mining,interval]          , histogram   , []},
- {[ae,epoch,aecore,chain,height]             , gauge       , []},
- {[ae,epoch,aecore,chain,top_change,interval], histogram   , []},
- {[ae,epoch,aecore,peers,count]              , gauge       , []},
- {[ae,epoch,aecore,peers,inbound]            , gauge       , []},
- {[ae,epoch,aecore,peers,outbound]           , gauge       , []},
- {[ae,epoch,aecore,peers,verified]           , gauge       , []},
- {[ae,epoch,aecore,peers,unverified]         , gauge       , []},
- {[ae,epoch,aecore,peers,blocked]            , gauge       , []},
- {[ae,epoch,aecore,peers,errored]            , gauge       , []},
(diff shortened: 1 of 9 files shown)


apps/aecore/src/aec_chain_state.erl
```diff
@@ -426,18 +432,19 @@ internal_insert(Node, Block, Origin) ->
...
-            Fun = fun() -> internal_insert_transaction(Node, Block, Origin)
+            Fun = fun() ->
+                          internal_insert_transaction(Node, Block, Origin)
                   end,
             try
                 aec_db:ensure_transaction(Fun)
             catch
                 exit:{aborted, {throw, ?internal_error(What)}} ->
-                    internal_error(What)
+                    {error, What}
```
sennui: that's a big change. internal_insert is not exported - OK.

tolbrino: Yes. The previous setup would double throw the error. Since only this tagged error was caught anyway, we can get rid of the second throw entirely.

apps/aecore/priv/exometer_predefined.script
```diff
@@ -1,29 +1,40 @@
...
+
+%% The execution time metrics are captured in micro-seconds.
+
+%% Keep 10 minutes of data, ~= 200 values, with a new micro-block every 3 seconds.
+, {[ae,epoch,aecore,blocks,micro,txs_execution_time,success]    , histogram , [{time_span, 10 * 60 * 1000}]}
+, {[ae,epoch,aecore,blocks,micro,txs_execution_time,error]      , histogram , [{time_span, 10 * 60 * 1000}]}
+, {[ae,epoch,aecore,blocks,micro,insert_execution_time,success] , histogram , [{time_span, 10 * 60 * 1000}]}
+, {[ae,epoch,aecore,blocks,micro,insert_execution_time,error]   , histogram , [{time_span, 10 * 60 * 1000}]}
+
+%% Keep 4 hours of data, ~= 120 values, with a new key-block every 2 minutes.
```
dincho: It could be metrics related, but we expect key-blocks every 3 minutes

tolbrino: Right, I tried to be conservative here. This way we don't get too many values into the histogram to make computation slow. If its every 3 minutes with these settings there will be 80 entries which is close enough to 100 to be useful.

PR #3102, opened 2020-01-08
Improve agree_on_height in sync

hanssv:
The previous implementation is sub-optimal and will become worse as the chain grows...

### DIFF
--- a/apps/aecore/src/aec_sync.erl
+++ b/apps/aecore/src/aec_sync.erl
@@ -724,14 +724,7 @@ do_work_on_sync_task(PeerId, Task, LastResult) ->
             delayed_run_job(PeerId, Task, sync_tasks,
                             fun() -> do_work_on_sync_task(PeerId, Task) end, 5000);
         {agree_on_height, Chain} ->
-            #chain{ blocks = [#chain_block{ hash = TopHash, height = TopHeight } | _] } = Chain,
-            LocalHeader = aec_chain:top_header(),
-            LocalHeight = aec_headers:height(LocalHeader),
-            MinAgreedHash = aec_chain:genesis_hash(),
-            MaxAgree = min(LocalHeight, TopHeight),
-            GenesisHeight = aec_block_genesis:height(),
-            case  agree_on_height(PeerId, TopHash, TopHeight,
-                                  MaxAgree, MaxAgree, GenesisHeight, MinAgreedHash) of
+            case agree_on_height(PeerId, Chain) of
                 {ok, AHeight, AHash} ->
                     epoch_sync:debug("Agreed upon height (~p): ~p", [ppp(PeerId), AHeight]),
                     Agreement = {agreed_height, #chain_block{ height = AHeight, hash = AHash }},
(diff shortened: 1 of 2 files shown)


apps/aecore/src/aec_sync.erl
```diff
@@ -756,6 +749,65 @@ do_work_on_sync_task(PeerId, Task, LastResult) ->
...
+        RemoteHash =
+            case TopHeight == MinHeight of
+                true  -> TopHash;
+                false -> get_header_by_height(PeerId, LocalHeight, TopHash)
+            end,
+        case aec_chain:hash_is_connected_to_genesis(RemoteHash) of
+            true ->
+                {ok, MinHeight, RemoteHash};
+            false ->
+                agree_on_height(PeerId, TopHash, MinHeight, 1)
```
ThomasArts: we keep the top hash of the remote as reference

apps/aecore/src/aec_sync.erl
```diff
@@ -756,6 +749,65 @@ do_work_on_sync_task(PeerId, Task, LastResult) ->
...
+        {error, Reason}
+    end.
+
+%% Height is a height where we disagree, so ask for Height - Step
+agree_on_height(PeerId, RemoteTop, Height, Step) ->
+    NewHeight = max(aec_block_genesis:height(), Height - Step),
+    RHash = get_header_by_height(PeerId, NewHeight, RemoteTop),
+    case aec_chain:hash_is_connected_to_genesis(RHash) of
+        true ->
+            agree_on_height(PeerId, RemoteTop, NewHeight, Height, RHash);
```
ThomasArts: If Height - Step < 0 (genesis height) then NewHeight == 0. We cannot talk to each other if we do not agree ongenesis height. If, however, other party sends us (malliciously) a different hash RHash then genesis Hash, then we loop for ever.

This is a unlikely attack vector, but one could polute the code a bit by `when Height > 0` as guard. Just give up if the we disagree on height 0.

apps/aecore/src/aec_sync.erl
```diff
@@ -756,6 +749,65 @@ do_work_on_sync_task(PeerId, Task, LastResult) ->
...
+                {ok, MinHeight, RemoteHash};
+            false ->
+                agree_on_height(PeerId, TopHash, MinHeight, 1)
+        end
+    catch throw:{get_header_by_height, {error, Reason}} ->
+        {error, Reason}
+    end.
+
+%% Height is a height where we disagree, so ask for Height - Step
+agree_on_height(PeerId, RemoteTop, Height, Step) ->
```
ThomasArts:
```suggestion
agree_on_height(PeerId, RemoteTop, Height, Step) when Height > 0 ->
```

ThomasArts: we cannot diasagree onHeight == 0

apps/aecore/src/aec_sync.erl
hanssv: @ThomasArts This is handled in the implementation of `get_header_by_height`? If we are indeed at `aec_chain:genesis_height()` (not necessarily 0) we actually don't ask the other node... We just take the genesis_hash from locally since we know they agree... There is even a comment indicating that!

apps/aecore/src/aec_sync.erl
hanssv: That change makes no sense...

PR #3117, opened 2020-01-13
Filter peers in bucket with each peer addition

### LINKED ISSUE #3114: Check max update lapse with each peer addition
Both verified and unverified pool are made of buckets, each bucket can hold up to some number of peers. For example, the unverified pool has 1024 buckets, each can have 64 peers.

When a new peer is added to a pool, it's assigned to a bucket (by hashing peer's IP address). If this bucket is full, with each peer addition a filtering function is called. It tries to find peers for removal to make space for the new peer. A peer that hasn't been updated for more than `max_update_lapse` is removed. If no such peer exists in the bucket, some of the peers is evicted instead (based on some algorithm). This works well when the number of peers in the network is much greater than we currently have. The buckets are not getting full, so the filtering function is not called and peers that are not updated anymore stay in the buckets.

Instead of waiting for a bucket to become full, the filtering function can be called with each peer addition to a bucket (even if the bucket is not full).

jur0:
Closes #3114

### DIFF
--- a/apps/aecore/src/aec_peers_pool.erl
+++ b/apps/aecore/src/aec_peers_pool.erl
@@ -1621,23 +1621,24 @@ pool_make_space(Pool, RSt, BucketIdx, FilterFun, SortKeyFun) ->
     #pool{buckets = Buckets, bucket_size = MaxBucketSize, skew = Skew} = Pool,
     Bucket = array:get(BucketIdx, Buckets),
     BucketSize = length(Bucket),
-    case BucketSize < MaxBucketSize of
-        true ->
-            {free_space, [], undefined, RSt, Pool};
-        false ->
-            {Bucket2, Removed, KeyedEvictable, EvictableSize}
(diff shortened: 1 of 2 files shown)


jur0:
> Isn't this DoS vector ?

Can you explain why you think so?

hanssv:
I'm not sure I understand how this will solve the issue with dead peers. Each node will now filter possibly dead peers more often, but unless they all do so at the same the dead peers will be redistributed? Or do we keep track of filtered peers somehow?

jur0:
This will call the filtering function removing the old peers more often. A peer is old when it hasn't been updated for longer than `max_update_lapse` (there hasn't been a ping which included this peer). Since this old peer is removed it won't be propagated to the other nodes (and updated there) so there is more likelihood it will get removed from them, too. This will work when we decrease `max_update_lapse` to something more reasonable (it's set to 30 days now).

We don't keep track of the filtered peers (yet).

PR #3123, opened 2020-01-17
Select random peers from verified pool instead of all peers

### LINKED ISSUE #3122: Propagate only peers from verified pool
Dead peers won't be sent in ping messages anymore.

jur0:
Closes #3122 

This changes behaviour of `aec_peers_pool:random_subset` which returns random peers from verified pool instead of random peers from both verified and unverified pools. These random peers are used when creating a ping message.


apps/aecore/src/aec_peers_pool.erl
```diff
@@ -1141,11 +1147,12 @@ verified_add(St, Now, Peer) ->
             Peer2 = get_peer(St3, PeerId),
             Peer3 = Peer2#peer{vidx = BucketIdx},
             St4 = set_peer(St3, PeerId, Peer3),
-            case is_available(St4, PeerId) of
+            St5 = add_lookup_verif_all(St4, PeerId),
+            case is_available(St5, PeerId) of
```
skkw: `is_available` can also return `undefined` - the callers `(un)verified_add` should handle that result value?

jur0: yes, it's handled in `verified_maybe_add`.

PR #3126, opened 2020-01-22
SC FSM: allow optional nonces for on-chain transactions

velzevur:
Whoever starts an on-chain action for the State Channel, pays the fee. It is also their `nonce` that is part of the non-replay attack protection of the transaction. This PR introduces the functionality of optionally providing the `nonce` to be used. If none is specified, the default behaviour of FSM picking what it considers to be the correct one is used instead.

This PR also fixes a bug in the FSM flow: once the chain watcher detects a malicious `channel_close_solo_tx` on-chain, the FSM enters a `channel_closing` state and participants are allowed to `slash`. Then any of them could trigger a `slash`, optionally providing `gas_price`, `fee` or `nonce` and the FSM produces the transaction for them. If for some reason the client decides not to authenticate this transaction but to abort it instead, the FSM rollsback to `open` state. Once back in `open` state - the client can not trigger a `slash` or `settle` actions anymore.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -4175,21 +4175,30 @@ handle_call(_St, _Req, From, D) ->
     keep_state(D, [{reply, From, {error, unknown_request}}]).
 
 handle_call_(awaiting_signature, {abort_update, Code, Tag}, From,
-             #data{op = #op_sign{tag = Tag}} = D) ->
+             #data{ op = #op_sign{tag = Tag}
+                  , on_chain_id = ChannelId} = D) ->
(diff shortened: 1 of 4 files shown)


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -4175,21 +4175,30 @@ handle_call(_St, _Req, From, D) ->
     keep_state(D, [{reply, From, {error, unknown_request}}]).
 
 handle_call_(awaiting_signature, {abort_update, Code, Tag}, From,
```
velzevur: Bugfix starts here and ends in line 4201

PR #3135, opened 2020-01-30
Add TCP connection check of unverified peers

### LINKED ISSUE #3134: Add TCP connection check of unverified peers
The node has issue finding peers that are alive in unverified pool and connect to them since there can be too many dead peers. Periodic TCP checks of random unverified peers could be helpful since the peers that can be connected to are moved to the verified pool. When a connection is going to be established, these available verified peers are selected first (not some random and most likely dead peers).

jur0:
Closes #3134

### DIFF
--- a/apps/aecore/src/aec_peer_connection.erl
+++ b/apps/aecore/src/aec_peer_connection.erl
@@ -70,8 +70,11 @@ start_link(Ref, Socket, Transport, Opts) ->
 
 %% -- API --------------------------------------------------------------------
 
-connect(#{} = Options) ->
+connect(#{conn_type := noise} = Options) ->
     epoch_sync:debug("New peer connection to ~p", [Options]),
+    aec_peer_connection_sup:start_peer_connection(Options);
+connect(#{conn_type := tcp} = Options) ->
+    epoch_sync:debug("New TCP probe to ~p", [Options]),
     aec_peer_connection_sup:start_peer_connection(Options).
 
 connect_start_link(Port, Opts) ->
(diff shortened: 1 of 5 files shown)


dincho:
I wonder why TCP check and not full Noise check?

I see that tcp checked peers are added to the verified pool, so now it mixes peers that are know to accept Noise, and also peers that accept TCP but may not accept Noise or are part of other network ... this sounds very weird to me. It invalidate the metrics and we don't know anymore how many verified peers that accept Noise we have in the pool. If it's decided to keep this implementation at all, please also fix the metrics.

jur0:
Full Noise connection is more expensive than just TCP. This is meant just to quickly check if a peer is alive.

dincho:
> Full Noise connection is more expensive than just TCP. This is meant just to quickly check if a peer is alive.

Do you have numbers, how much more expensive ?

jur0:
> > Full Noise connection is more expensive than just TCP. This is meant just to quickly check if a peer is alive.
> 
> Do you have numbers, how much more expensive ?

I don't, but if I was an attacker and wanted to fake live nodes, it's expensive to spin these nodes and it's pretty cheap to make the nodes accept Noise on the top of TCP. The difference between TCP and Noise check is negligible I'd say. The expensive part is to have an IP which is alive (in terms of TCP).

hanssv:
The scenario where it would probably make a difference is when a node from the wrong network accidentally gets into a pool... But otherwise I agree with Juraj's assessment.

PR #3145, opened 2020-02-07
Fix GC bug in Fate store maps

UlfNorell:
Fixes aeternity/aesophia#204.

This bug cause nested maps to be more expensive than they should be. Protected by Iris hard-fork.


tolbrino:
@UlfNorell This can be merged?

UlfNorell:
Yes indeed.

PR #3148, opened 2020-02-13
GH-3006 Fix compiler version "unknown"

### LINKED ISSUE #3006: Compiler version is always 'unknown' in Lima
There is a bug in `aect_sophia:serialize_` where we look up `version` rather than `compiler_version` in the result map from the compiler. This in combination with all code being re-fitted in `aeprimop:prepare_init_call` makes the compiler version being `<<"unknown">>` for all contracts in LIMA. 

This is under consensus (as it is part of the bytecode in the contract state tree), so we can't fix it until IRIS. Also, the compiler is using `aeserialization` for doing this serialization, it is probably a good idea to refactor `aect_sophia` (remove it?!) at the same time. Sadly we will have to be backwards bug-compatible here so the best approach is probably to bump the version of the serialization.

hanssv:
Fixes #3006

Replaces the buggy serialization in `aect_sophia` with the one from `aeserialization`. And keeps the buggy behavior for LIMA by explicitly setting the compiler version to "unknown"...

### DIFF
--- a/apps/aeprimop/src/aeprimop.erl
+++ b/apps/aeprimop/src/aeprimop.erl
@@ -2157,12 +2157,12 @@ timed_contract_call(Type, Fun, CallData, CTVersion) ->
 
     {Call, State}.
 
-process_timed_contract_call(Time, Type, Call, CallData, State,
+process_timed_contract_call(Time, Type, Call, CallData, _State,
                             #{vm := CtVMVsn, abi := CtABIVsn}) ->
     ReturnType = aect_call:return_type(Call),
     GasUsed = aect_call:gas_used(Call),
     CallDataSize = byte_size(CallData),
-    CtPK = aect_call:contract_pubkey(Call),
+    % CtPK = aect_call:contract_pubkey(Call),
     Metrics = [ {gas_used, GasUsed}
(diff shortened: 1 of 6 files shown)


apps/aeprimop/src/aeprimop.erl
```diff
@@ -2157,12 +2169,12 @@ timed_contract_call(Type, Fun, CallData, CTVersion) ->
...
     {Call, State}.
 
-process_timed_contract_call(Time, Type, Call, CallData, State,
+process_timed_contract_call(Time, Type, Call, CallData, _State,
                             #{vm := CtVMVsn, abi := CtABIVsn}) ->
     ReturnType = aect_call:return_type(Call),
     GasUsed = aect_call:gas_used(Call),
     CallDataSize = byte_size(CallData),
-    CtPK = aect_call:contract_pubkey(Call),
+    % CtPK = aect_call:contract_pubkey(Call),
```
tolbrino: Why do you keep that in here?

hanssv: I just fixed the compiler warnings that someone (you?) left behind... No great thought went into that process.

PR #3150, opened 2020-02-18
allow arbitrary callback function to handle non-core specific table s…

skkw:
…etup


skkw:
> Tests?

Since the only addition is `apply(M,F,Args)` I thought it's not needed.

PR #3152, opened 2020-02-19
Cleanup and polish FSM from unused terms

velzevur:


apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -279,9 +279,6 @@ groups() ->
     ].
 
 ga_sequence() ->
-    [ {group, signatures} ].
```
velzevur: Actual bug

apps/aechannel/src/aesc_fsm.erl
```diff
@@ -1896,8 +1894,7 @@ create_with_minimum_fee(Mod, Opts, CurrHeight, D, Attempts) ->
                                      %% the update as well
                         Opts1#{update =>
                                aesc_offchain_update:set_gas_price(GasPrice,
-                                                                  Update)},
-                        Opts1
```
velzevur: Actual bug

PR #3164, opened 2020-03-05
Batch channel setup for scalability tests

uwiger:
Some proposed changes to simplify scalability tests for State Channels.

The `multiple_channels` test still seems to fail with timeouts (and the target node remaining) when trying with 500 channels or more.


uwiger:
Solved in a different way now

PR #3165, opened 2020-03-05
Gh3154 fetch log via sc ws

### LINKED ISSUE #3154: SC: Allow client to fetch n last events using the websocket api
During channel creation the initiator can go of line. Once the initiator reconnects it will get an fsm_up event. At this point the initiator will need to know what state the FSM is in. Particularly the **channels_info open** event is interesting, where the client can tell if the channel is present on chain.

Suggested solution, make the event log available through the websocket api.

uwiger:
Closes #3154 

Some Dialyzer warnings, rebase needed, and documentation needs to be written.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -59,6 +59,7 @@
         , dry_run_contract/2   %% (fsm(), map()) -> {ok, call()} | {error, _}
         , get_balances/2       %% (fsm(), {ok, [key()]) -> [{key(), amount()}]} | {error, _}
         , get_history/1        %% (fsm()) -> [Event]
+        , get_history/2        %% (fsm(), Opts) -> [Event]
         , get_round/1          %% (fsm()) -> {ok, round()} | {error, _}
         , patterns/0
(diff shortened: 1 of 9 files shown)


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -3279,19 +3284,25 @@ report(Tag, Info, D) ->
...
             lager:debug("~p report_info(true, Client = ~p, Msg = ~p)",
                         [Role, Client, Msg]),
-            Client ! {?MODULE, self(), Msg};
+            maybe_send_rpt(Connected, Client, Msg),
+            log(rpt, maps:get(tag, Msg), Msg, D);
        true  ->
             lager:debug("~p report_info(~p, ~p)", [Role, DoRpt, Msg0]),
-            ok
-    end,
+            D
```
velzevur: This is to cause some funny bugs in my PR. Note to self and future reviewers 😄

uwiger: Yes, careful rebasing will surely be required

uwiger:
You mean more applications of `check_fsm_log()`?

Issue #3169, opened 2020-03-13
SC WS API doesn't allow reconnect until the channel object is on-chain


uwiger:
## Expected Behavior

The SC FSM allows the client to disconnect once the `create_tx` signing reply has arrived. It should also then be possible to reconnect immediately.

## Actual Behavior

But the reconnect logic in the WS API aborts if it can't find the `existing_channel_id` on-chain, which could be significantly later.

## Steps to Reproduce the Problem

  1. Begin setting up a channel
  1. Sign the `create_tx`, but don't produce any microblocks to pick it up
  1. Disconnect one of the clients. The FSM should remain
  1. Try reconnecting. This should fail.

uwiger:
Note that this is partly related to issue #3027

uwiger:
Addressed in #3224.

PR #3170, opened 2020-03-16
Allow SC reconnect after signing create_tx

uwiger:
See issue #3169 

Some refactoring of the sc suite, to allow for partial setup of a channel, pausing before the create_tx has been mined onto the chain.

In local test runs, `aehttp_sc_SUITE` fails with errors seemingly related to contract cache issues. This seems separate from the present fix.

### DIFF
--- a/apps/aechannel/src/aesc_fsm.erl
+++ b/apps/aechannel/src/aesc_fsm.erl
@@ -1365,7 +1365,6 @@ send_open_msg(#data{ opts    = Opts
            },
     aesc_session_noise:channel_open(Sn, Msg),
     Data#data{ channel_id   = ChannelPubKey
-             , temp_chan_id = ChannelPubKey
              , log = log_msg(snd, ?CH_OPEN, Msg, Log) }.
 
 check_open_msg(#{ chain_hash           := ChainHash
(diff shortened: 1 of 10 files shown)


apps/aechannel/src/aesc_fsm.erl
```diff
@@ -3184,8 +3215,19 @@ watch_req() ->
 on_chain_id(#data{on_chain_id = ID} = D, _) when ID =/= undefined ->
     {ID, D};
 on_chain_id(D, SignedTx) ->
-    {ok, PubKey} = aesc_utils:channel_pubkey(SignedTx),
-    {PubKey, D#data{on_chain_id = PubKey}}.
+    try aesc_utils:channel_pubkey(SignedTx) of
+        {ok, PubKey} ->
+            {PubKey, D#data{on_chain_id = PubKey}}
+    ?CATCH_LOG(_E)
+        if D#data.strict_checks == false ->
```
velzevur: `strict_checks` is for AliceFSM, in the case Alice tries cheating Bob. That indeed would be a test scenario. I am wondering if we get here at all.

velzevur: Yep, the `check_incorrect_create` goes right there, this is indeed needed

velzevur:
Since a new state had been added, please update the UML

uwiger:
Superseded by #3224

PR #3178, opened 2020-03-23
Add db persistance to all docker examples

dincho:


AleksandarFilipov:
I guess you could write a line or two in the topic. Also that it's optional

PR #3187, opened 2020-03-25
Whitepaper contract

velzevur:
This is the contract that is intended to go in the whitepaper as a showcase for using a State Channel's off-chain Smart Contract using on-chain data. The dataitself is represented by an on-chain weather Oracle that publishes on-chain if there had been an hailstorm at a certain block height. This is later on used in the off-chain contract to trigger an insurance coverage event.

The Oracle's data publishing lifecycle requires a `query` to precede the `response`. This is either to be done in two steps as distinct transactions or as a single contract call as with the second smart contract provided. Note that there is no tests coverage for it.

### DIFF
--- a/test/contracts/channel_whitepaper_example.aes
+++ b/test/contracts/channel_whitepaper_example.aes
@@ -24,7 +24,7 @@ contract ChannelWhitepaperExample =
       active_to       = 0
       }
 
-  stateful payable entrypoint insure(generations: int) =
+  stateful payable entrypoint insure() =
     if (Call.caller == Contract.creator)
       abort("service_provider")
     if (Contract.balance + Call.value < state.compensation)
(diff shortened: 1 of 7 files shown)


test/contracts/channel_whitepaper_example.aes
```diff
@@ -0,0 +1,76 @@
...
+      put(state{ active_to = state.active_to + period })
+
+  stateful entrypoint withdraw(amount: int) =
+    if (Call.caller != Contract.creator)
+      abort("not_service_provider")
+    if (Chain.block_height =< state.active_to) // insured
+      if (Contract.balance - amount < state.compensation)
+        abort("not_enough_compensation")
+    Chain.spend(Contract.creator, amount)
+    ()
```
thepiwo: no need to return empty tuples

velzevur: As Hans noted, `unit`s save gas, so this is an optimization

hanssv: Not here, `Chain.spend` already returns unit :-)

test/contracts/channel_whitepaper_example.aes
```diff
@@ -0,0 +1,75 @@
...
+    { oracle          = oracle,
+      city            = city,
+      price_per_gen   = price_per_gen,
+      compensation    = compensation,
+      active_from     = 0,
+      active_to       = 0
+      }
+
+  stateful payable entrypoint insure() =
+    if (Call.caller == Contract.creator)
```
uwiger:
```suggestion
    require (Call.caller != Contract.creator, "service_provider")
```

uwiger: (would be applied generally to all such tests)

Issue #3200, opened 2020-04-01
Occasional timeouts in fsm_SUITE throughput tests


uwiger:
## Expected Behavior

The throughput tests create several concurrent channels, and should complete in reasonable time.

## Actual Behavior

Sometimes, there are timeouts, where some of the channel proxies don't get their expected response from the `miner_loop` process used to coordinate block mining requests.

## Steps to Reproduce the Problem

Happens mostly in CI, occasionally.

A first step should be to refactor the test code so that more informative error reports are produced when it happens.

## Logs, error output, etc.

```
*** System report during aesc_fsm_SUITE:many_chs_msg_loop/1 in throughput 2020-04-01 12:47:13.617 ***

=ERROR REPORT==== 1-Apr-2020::12:47:13 ===
Error in process <0.13699.0> on node aeternity_ct@localhost with exit value:
{timeout,[{aesc_fsm_SUITE,wait_for_signed_transaction_in_block,3,
                          [{file,"/home/builder/aeternity/apps/aechannel/test/aesc_fsm_SUITE.erl"},
                           {line,3033}]},
          {aesc_fsm_SUITE,create_channel_from_spec,7,
                          [{file,"/home/builder/aeternity/apps/aechannel/test/aesc_fsm_SUITE.erl"},
                           {line,2351}]},
          {aesc_fsm_SUITE,create_multi_channel_,3,
                          [{file,"/home/builder/aeternity/apps/aechannel/test/aesc_fsm_SUITE.erl"},
                           {line,2286}]}]}
```

uwiger:
The throughput tests were moved in PR #3277 to a separate benchmark suite, which is not run automatically. Suggest that we close this, as occasional timeouts are much less of a problem when tests are run on-demand.

PR #3201, opened 2020-04-01
Refactor miner_loop interaction for better debugging

uwiger:
See issue #3200 

Initial push is to improve the chance of gleaning something useful from failures in CI, since it's hard to reproduce this locally.

### DIFF
--- a/apps/aechannel/test/aesc_fsm_SUITE.erl
+++ b/apps/aechannel/test/aesc_fsm_SUITE.erl
@@ -301,11 +301,11 @@ update_sequence() ->
     [ check_incorrect_update
     ].
 
-update_sequence_() ->
-    [ check_incorrect_deposit
-    , check_incorrect_withdrawal
-    , check_incorrect_update
-    ].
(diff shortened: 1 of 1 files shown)


apps/aechannel/test/aesc_fsm_SUITE.erl
```diff
@@ -301,11 +301,11 @@ update_sequence() ->
     [ check_incorrect_update
     ].
 
-update_sequence_() ->
-    [ check_incorrect_deposit
-    , check_incorrect_withdrawal
-    , check_incorrect_update
-    ].
+%% update_sequence_() ->
```
velzevur, line 304: Tracked as #3203

uwiger:
The test failures are odd. They seem related to the newly merged FP PR (#3197)

velzevur:
> The test failures are odd. They seem related to the newly merged FP PR (#3197)

Tracked as #3204. It is not yet approved, cc @tolbrino

PR #3208, opened 2020-04-06
Gh 3204 fix test

velzevur:
Fixes #3204

PR #3211, opened 2020-04-07
Fix (with height protection) current contract issue

hanssv:
It needs a release note, etc... But this is the technical part.


velzevur:
Tests failed with a known issue that is addressed in #3208

PR #3222, opened 2020-04-15
future proof rebar.lock.script

tsloughter:
The upcoming rebar3 release expands on the hashes kept in `rebar.lock` to be safer. This patch ensures that that change and any future addition to the lock file won't crash the script.


tolbrino:
The failing test case is unrelated and fixed in `master`. So merging since the change looks good.

PR #3224, opened 2020-04-17
Gh3169 sc reestablish early 2

uwiger:
Same as #3170 , but rebased against current master.


AleksandarFilipov:
Temporary revived this branch to be able to refrence it.

PR #3240, opened 2020-04-24
Re-ordered variables

### LINKED ISSUE #3171: The Gas parameter is in the second position, the first is Db.
https://github.com/aeternity/aeternity/blob/39b5d3e2e942167f6ad54b842caeeb8670a2b0ae/apps/aecore/src/aec_tx_pool.erl#L434

So the int_get_candidate should be following: 
int_get_candidate(_,Gas, _, _, _, _, _, Acc) ->

DanielaIvanova:
Fixes #3171


dincho:
Also not sure why my review is requested here

radrow:
Is it possible to add a test case to cover this?

PR #3241, opened 2020-04-24
sending generation info to datadog

### LINKED ISSUE #3215: Send generation's info to datadog
Each generation's keyblock has an `info` that represents the protocol version of the miner. Expose this to datadog

skkw:
Fixes #3215

### DIFF
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -1124,7 +1124,7 @@ handle_successfully_added_block(Block, Hash, Events, State, Origin) ->
         {changed, BlockType, NewTopBlock, State2} ->
             BlockType == key andalso
                 aec_metrics:try_update(
-                  [ae,epoch,aecore,mining,info],
+                  [ae,epoch,aecore,blocks,key,info],
                   aec_headers:info(aec_blocks:to_key_header(NewTopBlock))),
             IsLeader = is_leader(NewTopBlock),
             case IsLeader of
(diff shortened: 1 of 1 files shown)


dincho:
What is this about? Link to issue/description ?

apps/aecore/src/aec_conductor.erl
```diff
@@ -1122,6 +1122,10 @@ handle_successfully_added_block(Block, Hash, Events, State, Origin) ->
         {micro_changed, State2 = #state{ consensus = Cons }} ->
             {ok, setup_loop(State2, false, Cons#consensus.leader, Origin)};
         {changed, BlockType, NewTopBlock, State2} ->
+            Origin == block_created andalso
```
velzevur, line 1125: I see this changes the report to be done only by the miner that produced the block. Why? Do we intend to run this on our non-mining seed nodes?

tolbrino:
@dincho @uwiger @skkw So this is done? At least release notes are missing.

apps/aecore/src/aec_conductor.erl
dincho: Hmm, this shouldn't be the case. It should post info for each block that's "appended" to the node database. Including received from sync/p2p gossip.

dincho:
> @dincho @uwiger @skkw So this is done? At least release notes are missing.

No we've review comments above.

dincho:
closing in favour of https://github.com/aeternity/aeternity/pull/3265

PR #3255, opened 2020-05-04
WIP on-chain checking bef reestablish, preparation

uwiger:
First commit.
* Export watcher support for getting all channel on-chain txs since a given hash
* fsm caches the hash of the last seen tx
* TODO: process the txs since last time and update state


uwiger:
Solved by #3274 - closing

PR #3262, opened 2020-05-08
Remove logs from attached console

tolbrino:


tolbrino:
Lager log levels can be changed at runtime, so we could certainly add OS env vars which are read by setup at system start and used to adapt the console log levels. In that case you'd want 2 different vars to selectively enable the logging, correct?

/cc @uwiger

dincho:
> Lager log levels can be changed at runtime, so we could certainly add OS env vars which are read by setup at system start and used to adapt the console log levels. In that case you'd want 2 different vars to selectively enable the logging, correct?
> 
> /cc @uwiger

That would be the best, first one to enable console logging and the second to change the level.

dincho:
> > Lager log levels can be changed at runtime, so we could certainly add OS env vars which are read by setup at system start and used to adapt the console log levels. In that case you'd want 2 different vars to selectively enable the logging, correct?
> > /cc @uwiger
> 
> That would be the best, first one to enable console logging and the second to change the level.



> Lager log levels can be changed at runtime, so we could certainly add OS env vars which are read by setup at system start and used to adapt the console log levels. In that case you'd want 2 different vars to selectively enable the logging, correct?
> 
> /cc @uwiger

It's debatable if we should use ENV vars tho, as the node takes mostly command arguments right now, we should probably keep using that or unify it somehow to accept env vars as well.

(we accept only config path as env var AFAIK right now)

hanssv:
Adding a couple of simple functions that can be called in the shell when you'd like to use it?
I've got the following two lines in my NOTES file:
```
lager:set_loglevel(lager_console_backend, warning).
lager:set_loglevel(epoch_sync_lager_event,lager_console_backend,undefined,warning).
```
It silences the logger more than enough to make it nice and quiet.

tolbrino:
> Adding a couple of simple functions that can be called in the shell when you'd like to use it?
> I've got the following two lines in my NOTES file:
> 
> ```
> lager:set_loglevel(lager_console_backend, warning).
> lager:set_loglevel(epoch_sync_lager_event,lager_console_backend,undefined,warning).
> ```
> 
> It silences the logger more than enough to make it nice and quiet.

You are correct, that would be the simplest solution. @dincho I am leaning towards just adding helper functions for that instead of extra env vars. Ok for you?

tolbrino:
I have followed @hanssv suggestion.

tolbrino:
> Returning the previous log level makes perfect sense to me, but might be confusing for less experienced programmers!?

You might be right. Let's see whether anybody complains.

tolbrino:
@dincho ping

PR #3269, opened 2020-05-11
(macos) Add docker memory note

nikita-fuchs:
add note for macos users to eventually increase their memory allocation


tolbrino:
> The amended document has nothing to do with with building docker images, but only running. The node does not need that significant amount of memory to run (only to build).
> 
> However, this document can be enriched with build section and the sentence proposed added there.

I agree. The file should have a separate section on using Docker for building.

velzevur:
As discussed, a this shall be part of a broader document that describes the build process via Docker.

PR #3270, opened 2020-05-12
Wrong ulimit suggestion

### LINKED ISSUE #3188: Wrong ulimit suggestion
## Expected Behavior

Suggestion works as is.

## Actual Behavior

```
-bash: ulimit: open files: cannot modify limit: Invalid argument
```

  1. Start a node
  1. Get a message `WARNING: ulimit -n is $OPEN_FILES_LIMIT; $OPEN_FILES_RECOMMENDED is the recommended minimum.`
  1. Try to run the suggestion on latest MacOS

## Specifications

  - OS: macOS Catalina
  - Node Version: 5.5.0

DanielaIvanova:
resolves #3188


DanielaIvanova:
@dincho as I understood, we might run into issues if we don't set `OPEN_FILES_RECOMMENDED=24576`, as the node keeps running because it has open network connections and file descriptors, so we have to keep inform the users that they need to set `OPEN_FILES_RECOMMENDED`.

tolbrino:
The problem is that macOS requires a different step before running `ulimit` to raise the limit that high. I don't think we can explain all of that in the shell. This problem is the same on some Linux platforms, but luckily there the limits are still high enough, but that might change.

PR #3271, opened 2020-05-12
Added CORS headers for error responses

### LINKED ISSUE #3192: Missing CORS headers for error responses
## Expected Behavior

CORS headers should be send (when enabled) with error responses as well.

It does not behave correct both for 4xx and 5xx responses.

## Actual Behavior

Not send

  ```curl localhost:3013/v2/not_found -v -H "Origin: https://test.com"```

## Specifications

  - OS: any
  - Node Version: master

DanielaIvanova:
resolves #3192


apps/aehttp/src/aehttp_app.erl
```diff
@@ -99,8 +99,8 @@ start_http_api(Target, LogicHandler) ->
             {ip, ListenAddress},
             {num_acceptors, PoolSize}],
     Env = #{env => #{dispatch => Dispatch},
-            middlewares => [cowboy_router,
-                            aehttp_cors_middleware,
+            middlewares => [aehttp_cors_middleware,
```
tolbrino, line 102: Why was the order a problem?

DanielaIvanova: Because by fixing the ordering we are fixing the problems with headers, our `cors` headers will **always** be added, instead of default headers of the cowboy router.

docs/release-notes/next/GH-3192-cors-headers.md
```diff
@@ -0,0 +1 @@
+* Added CORS headers for error responses
```
tolbrino, line 1: I could be nitpicking here. Let's merge this. But in the future when writing these, consider the reader has no context. Doesn't know the change. So I would wonder, responses from what API.

PR #3273, opened 2020-05-15
Add contract call timing test

uwiger:
Reusing the scaffolding for the simple benchmark tests in the FSM suite, add a test that calls the simple counter contract already present.

Rough numbers indicate that simple contract calls are some 40% slower than token transfers, which seems like a pretty good result.


uwiger:
I also reduced the `N` for the update_transfer bench test. This shaved off ca 5 seconds on the test execution time, and it doesn't seem to reduce the usefulness of the information much.

uwiger:
> I don't see that this tests anything that wasn't tested before? Continuous integration isn't where you do benchmarking.
> 
> +1 on reducing the other benchmarking, that should be elsewhere entirely.

Addressed by merging #3277 into this PR. New reviews needed

PR #3275, opened 2020-05-26
Gh 3203 fix tests

### LINKED ISSUE #3203: Fix aesc_fsm_SUITE:update_sequence/0
## Expected Behavior

only `check_incorrect_update` is being tested. Expectaion is `check_incorrect_deposit` and `check_incorrect_withdrawal` are tested as well

velzevur:
Fixes #3203


velzevur:
What happened is that the FSM that had the raised flag `strict_checks` to false would skip checks of incoming transactions. This in the case of withdraw and deposit made it accept an erroneously signed transaction and entering in a `awaiting_locked` state waiting for the withdrawal or deposit respectively. The flag is tests only. Also worth noting is that this happens for the cheating party: the malicious party gets in the wrong FSM state.

One approach would be teaching the FSM to better handle the player cheating while flag is raised. I think this makes the wrong assumption that somehow this is OK and I rather took the other path: accomodate the test to leave the FSM in whatever state it is.

PR #3292, opened 2020-06-10
WIP Initial approach to pluggable core functionality

uwiger:
A first sketch of how to handle pluggable functionality. Some thoughts:

* Allowing free redefinition of records etc would mess up the type checking pretty badly, so I added an `extra` attribute (type: `map()`) to records that may need extending. The overriding module controls the content and interpretation of that. By default, it's the empty map.
* A new module `aec_plugin.erl` supports a "plugin map", where known semantic tags (currently, only `[aec_headers]`) can be mapped to new modules. The new modules must be loadable for the mapping to succeed. In OTP 21 and later, the plugin map is handled via `persistent_term`; in earlier OTP versions, as an `aecore` application env variable (perhaps it should instead be hard-coded as disabled?)
* An include file `aec_plugin.hrl` provides a macro `?PLUGGABLE(Args, DefaultExpr)` which checks if the current module is redefined, and if so, if the current function/arity is exported in the redefining module. If so, the redefining function is called with `Args` (expression including the parentheses surrounding the arguments); otherwise, `DefaultExpr` is evaluated. Example:

```erlang
deserialize_from_binary(Bin) ->
    ?PLUGGABLE((Bin), deserialize_from_binary_(Bin)).

deserialize_from_binary_(Bin) ->
    case deserialize_from_binary_partial(Bin) of
        {key, Header} -> Header;
        {micro, Header, <<>>} -> Header;
        {error, What} -> error(What)
    end.
```

### DIFF
--- a/apps/aecore/src/aec_headers.erl
+++ b/apps/aecore/src/aec_headers.erl
@@ -60,6 +60,7 @@
 -include_lib("aeminer/include/aeminer.hrl").
 -include_lib("aecontract/include/hard_forks.hrl").
 -include("blocks.hrl").
+-include("aec_plugin.hrl").
 
 %%%===================================================================
 %%% Records and types
@@ -80,7 +81,8 @@
           signature    = <<0:?BLOCK_SIGNATURE_BYTES/unit:8>>   :: block_signature(),
           txs_hash     = <<0:?TXS_HASH_BYTES/unit:8>>          :: txs_hash(),
           time         = 0                                     :: non_neg_integer(),
-          version                                              :: non_neg_integer()
+          version                                              :: non_neg_integer(),
+          extra        = #{}                                   :: map()
          }).
 
 -record(key_header, {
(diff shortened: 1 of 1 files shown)


apps/aecore/src/aec_plugin.erl
```diff
@@ -0,0 +1,80 @@
...
+            lager:error("Module ~p not loaded: ~p", [M, E]),
+            false
+    end.
+
+%% We want to use persistent_term for storing the registry, since
+%% fetching it will have the same overhead as a remote function call.
+%% The functionality was introduced in OTP 21.2. Also in OTP 21, the
+%% predefined macro ?OTP_RELEASE was introduced. Let's assume that anyone
+%% using OTP 21 is using at least 21.2.
+-ifdef(OTP_RELEASE).
```
tolbrino, line 57: Keep in mind to do https://github.com/aeternity/aeternity/issues/3104 first. Should be fairly fast.

gorbak25:
Merged the parse transform PR.
Before merging this branch to master:
1) Sync mainet node from scratch
2) Load existing DB
3) CI pipeline green

gorbak25:
Rebased on master

uwiger (commented):
LGTM (since I'm the originator, I can't approve it)

gorbak25:
Ok, synced with mainnet in 41.14227%(top: kh_2rTkGyTpMZkzMEEZ8LkSKpkeajLqbMNHxsXz91W3afK6ScmFH4) and no hiccups so far. I'm aborting the full sync as I need my nvme for more important stuff :P Will try to start from an existing db.

gorbak25 (approved):

radrow (approved):

PR #3294, opened 2020-06-10
Use parse_transform w -pluggable() attrs

uwiger:
A PR against #3292 

As requested, a parse_transform acting on `-pluggable([f/1, g/2])` attributes instead. The core code then can remain untouched.

### DIFF
--- a/apps/aecore/include/aec_plugin.hrl
+++ b/apps/aecore/include/aec_plugin.hrl
@@ -1,14 +1,3 @@
 
--define(PLUGGABLE(Args, Default),
-        case aec_plugin:get_module(?MODULE) of
-            undefined ->
-                Default;
-            __M ->
-                case erlang:function_exported(
-                       __M, ?FUNCTION_NAME, ?FUNCTION_ARITY) of
-                    true ->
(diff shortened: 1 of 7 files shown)


uwiger:
Fixing test failures and trying to add some enhanced support, to make for more maintainable code.

uwiger:
@hanssv, @velzevur Please give new reviews, since quite a bit has been added since the approvals were given.

velzevur:
Since CircleCI had been fixed, I restarted the upload artifacts job

erlmachinedev:
I consider this approach is elegant and flexible but I want to touch some minuses (I also respect so much @uwiger effort applied here)

1) Every time I faced parse transform I got the same debug problem - correlations between source code and produced error log. So from that time, we need to take into account this ambiguity which covers the main functionality and potentially will be appeared often as we deal with structure serialization/deserialization and other IO processing stuff (especially isn't clear for the new developers);

2) I guess from #3292 the main purpose of parse transform is to hide the pluggable approach and complexity as such. But I consider the plugin choice has to be expressed as explicit and transparent as possible (maybe it can change the general architecture) and to be emphasized into appropriate modules for clearness.

PR #3316, opened 2020-06-23
Remove integration sync triggers from CI config

IvayloPapazov:
Because the context does not work with a schedule as we expected.


dincho:
Unrelated test failure, merging.

Issue #3296, opened 2020-06-15
Annotated full reference config file


dincho:
Create a full reference configuration file annotated with extensive documentation.

Should be tested agains the scheme and if it matches its defaults and if the reference configuration is missing configuration options defined in the schema.

Should be shipped with the package, TBD in what dir.

PR #3323, opened 2020-08-03
Annotated config file for #3296

akovari:
Annotated config file for #3296. I also added some default values to the schema based on defaults I was able to find in the code. Please let me know if this is not ok.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -444,6 +444,9 @@ test-arch-os-dependencies: KIND=test
 test-arch-os-dependencies:
 	make ct-latest SUITE=apps/aecontract/test/aecontract GROUP=sophia TEST=sophia_crypto
 
+validate-reference-config:
+	./_build/prod/bin/check_config docs/reference-config.yaml
+
 .PHONY: \
 	all console \
 	local-build local-start local-stop local-attach \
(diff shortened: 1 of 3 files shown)


dincho:
This does not fully covers the idea of #3296 and especially:

> Should be tested agains the scheme and if it matches its defaults and if the reference configuration is missing configuration options defined in the schema.

That is, there should be a tool that checks if the reference configuration is missing some option and if it's default diverts from what's in the schema.

A second better approach would be to have a tool that generates the reference configuration based on the schema.

Having 2 files with the same information that has to be updated manually proved to fail in the past, because developers often forgot to update them both.

On the formatting I'd recommend to have the comment above each option and use 2 spaces indentation, also additional section empty lines as separators would improve the readability. Some nice examples:
- https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-reference-yml.html
- https://www.appveyor.com/docs/appveyor-yml/
- https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#workflow-file-example

I'd also recommend having the configuration file with it's default name `aeternity.yaml` so that operator can just simply do `cp docs/config/aeternity.yaml .` in the app working dir without writing the destination file name.

Issue #3324, opened 2020-08-04
 aehttp_spec_SUITE:validate_api: FAILED


velzevur:
Our approach for swagger.yaml validation seems to be no longer up to date with their API.
The test fails with:

```
2020-08-03 20:04:44.719
Connection problem: {ok,{{"HTTP/1.1",400,"Bad Request"},
                         [{"connection","keep-alive"},
                          {"date","Mon, 03 Aug 2020 20:04:44 GMT"},
                          {"server","Jetty(9.4.18.v20190429)"},
                          {"content-length","76"},
                          {"content-type","application/json"},
                          {"access-control-allow-origin","*"},
                          {"access-control-allow-methods",
                           "GET, POST, DELETE, PUT"},
                          {"access-control-allow-headers",
                           "Content-Type, api_key, Authorization"},
                          {"access-control-expose-headers",
                           "Content-Disposition"}],
                         "{\"code\":400,\"message\":\"Input error: missing required query parameter `url`\"}"}}


%%% aehttp_spec_SUITE ==> validate_api: FAILED
```

PR #3325, opened 2020-08-04
Swagger validator URL has changed #3324

erlmachinedev:
The location of validator needs to be passed as https://validator.swagger.io/validator/ (instead of http). So that is the largest PR I ever made :).
P.S. I suggest to relocate this public endpoint to the appropriate config (it's should be visible to understand our behaviour from the external world perspective).


hanssv:
The title isn't very descriptive ;-)

erlmachinedev:
> The title isn't very descriptive ;-)

Got you (I will take into account)

gorbak25 (approved):

PR #3330, opened 2020-08-21
Config structure for Hyperchains; Module names;

erlmachinedev:
@gorbak25 , @radrow Take a look, hyperchain section and jobs configuration (potentially will be used to control capacity, seems like the general approach here).

### DIFF
--- a/apps/aeutils/priv/aeternity_config_schema.json
+++ b/apps/aeutils/priv/aeternity_config_schema.json
@@ -494,6 +494,45 @@
                             "default": 500
                         }
                     }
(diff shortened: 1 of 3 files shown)


apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -494,6 +494,45 @@
                             "default": 500
                         }
                     }
+                },
+                "hyperchain": {
```
gorbak25, line 498: s/hyperchain/hyperchains

PR #3340, opened 2020-08-29
Improve tests to expose server mode of SC FSM

### LINKED ISSUE #3069: SC: Feature, addition allow merchant to subscribe to incoming client requests
Feature request.

In order to make state channels more approachable we should allow the merchant subscribe to incoming client requests.
A coffee shop should be able to provide a service where the client start their app which should establish/re-establish a connection with the merchant where the merchant should enable automatically start the corresponding peer (channel client).

This would allow the following suggested end to end user flow.
1 Customer (bob) downloads app.
2 bob starts app, app targets predefined node where his fsm is hosted. Once app is started a channel open/reestablish is triggered
3 bobs action triggers an event at merchant (alice). Upon this action merchant opens the corresponding fsm.
4 bob and alice now have an connection.

In the initial embodiment bob could provide contracts (if needed) and the alice app (web app) could be fairly simple.

Prerequisites for this to work is that there is a hosting node which can be reached by both parties.  ref: @dincho (to be added)

velzevur:
Closes #3069 

Ulf's comment was right and this indeed is already present. I added some tests to validate it.

This PR had been produced with the support of the Aeternity Foundation. You can read about this in [the forum](https://forum.aeternity.com/t/active-aeternity-node-maintenance-iris-hard-fork-release-candidate/7365/).

### DIFF
--- a/apps/aehttp/test/aehttp_sc_SUITE.erl
+++ b/apps/aehttp/test/aehttp_sc_SUITE.erl
@@ -91,7 +91,9 @@
     sc_ws_abort_shutdown/1,
     sc_ws_abort_slash/1,
     sc_ws_abort_settle/1,
-    sc_ws_can_not_abort_while_open/1
+    sc_ws_can_not_abort_while_open/1,
+    sc_ws_responder_can_use_any_initiator/1,
+    sc_ws_responder_can_use_any_initiator_and_initiator_can_reattach/1
    ]).
 -export([sc_ws_open_/4,
          sc_ws_close_/1,
@@ -315,6 +317,12 @@ groups() ->
       , sc_ws_leave_reconnect
       , sc_ws_reconnect_early
       , sc_ws_leave_responder_does_not_timeout
+      ]},
+
+      %% possible to leave and reestablish channel
+     {any_initiator, [sequence],
+      [ sc_ws_responder_can_use_any_initiator
+      , sc_ws_responder_can_use_any_initiator_and_initiator_can_reattach
       ]}
     ].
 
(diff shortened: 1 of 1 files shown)


radrow (approved):

apps/aehttp/test/aehttp_sc_SUITE.erl
```diff
@@ -5741,3 +5756,47 @@ sc_ws_leave_responder_does_not_timeout(Config0) ->
...
+                                                ])),
+    ct:log("channel opened", []),
+    ct:log("Config = ~p", [Config]),
+    ct:log("*** Leaving channel ***", []),
+
+    Config1 = [{responder_leaves, false}|Config],
+    ct:log("Config1 = ~p", [Config1]),
+    ReestablishOptions = sc_ws_leave_(Config1),
+    Config2 = reconnect_client_(ReestablishOptions, initiator,
+                                [{reconnect_scenario, reestablish} | Config1]),
```
gorbak25, line 5783: Odd formatting compared to other places in the codebase.
```suggestion
                                [{reconnect_scenario, reestablish}|Config1]),
```

apps/aehttp/test/aehttp_sc_SUITE.erl
```diff
@@ -5741,3 +5756,47 @@ sc_ws_leave_responder_does_not_timeout(Config0) ->
...
+    ok = sc_ws_update_(Config2),
+    ct:log("*** Closing ... ***", []),
+    ok = sc_ws_close_(Config2).
+
+
+custom_responder_opts(L) ->
+    Opts =
+        lists:foldl(
+            fun({keep_running, Bool}, Accum) when is_boolean(Bool) ->
+                Accum#{keep_running => true};
```
gorbak25, line 5794: Might be confusing for future users who might call custom_responder_opts([{keep_running, false}])

PR #3349, opened 2020-09-04
Basic HC abstract data structures, Persistent parent DB

gorbak25:
1) Basic HC abstract data structures
2) Persistent parent DB
3) Primitive for retrieving election candidates
4) Tests

### DIFF
--- a/apps/aecore/src/aec_block_micro_candidate.erl
+++ b/apps/aecore/src/aec_block_micro_candidate.erl
@@ -15,6 +15,12 @@
 
 -export([ min_t_after_keyblock/0]).
 
+-ifdef(TEST).
+-export([ create_with_txs/2
+        , create_with_txs_and_state/3
+        ]).
+-endif.
+
 -export_type([block_info/0]).
 
 -include("blocks.hrl").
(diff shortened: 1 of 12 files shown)


erlmachinedev:
This is an important piece of our project and I suggest to merge that. For conversation log purposes I have to mention some points which would act as an optimization advises:

1) If we can improve **gen_server** as the main provider of this API we can avoid transaction overhead in each call (gen_server guarantees sequential access to the related tables). This approach (gen_server as API provider) would allow using different access layers, for example, https://github.com/rabbitmq/gen-batch-server when the fetched data volume is large (for example the current height of Aeternity is 311027,  100 commitments per block can lead to 311027 * 100 transaction calls and for now this code can lead to freeze of Db for sufficient amount of time);

2) For me as API consumer could be useful to manage all protocol via aehc_commitment.erl module (To be able to call aehc_commitment:new/2 I have to prepare Header and PoGF and to construct objects which are internal part of the commitment itself);

3) We use some duplicated stuff from aecore part, for example: table_specs/1,  define(TAB(Record)), -define(t(Expr), aec_db:ensure_transaction(fun() -> Expr end)), I guess in the future time it should be relocated to appropriate header file and to be shared between our app and core part too;

I see the well-documented module and suggest to merge (we can face with additional API improvements and Db access optimizations but I guess not now)

erlmachinedev (approved):

gorbak25:
> This is an important piece of our project and I suggest to merge that. For conversation log purposes I have to mention some points which would act as an optimization advises:
> 
>     1. If we can improve **gen_server** as the main provider of this API we can avoid transaction overhead in each call (gen_server guarantees sequential access to the related tables). This approach (gen_server as API provider) would allow using different access layers, for example, https://github.com/rabbitmq/gen-batch-server when the fetched data volume is large (for example the current height of Aeternity is 311027,  100 commitments per block can lead to 311027 * 100 transaction calls and for now this code can lead to freeze of Db for sufficient amount of time);
> 
Yup, this could be also done for the core db part.

>     2. For me as API consumer could be useful to manage all protocol via aehc_commitment.erl module (To be able to call aehc_commitment:new/2 I have to prepare Header and PoGF and to construct objects which are internal part of the commitment itself);
> 
Ok, just add any helpers you might need when implementing other functionality.

>     3. We use some duplicated stuff from aecore part, for example: table_specs/1,  define(TAB(Record)), -define(t(Expr), aec_db:ensure_transaction(fun() -> Expr end)), I guess in the future time it should be relocated to appropriate header file and to be shared between our app and core part too;
> 
Yes, this refactoring needs to be done - I will do it in a separate PR.
> 
> I see the well-documented modules and suggest to merge (we can face with additional API improvements and Db access optimizations but I guess not now)

radrow (approved):

Issue #3301, opened 2020-06-15
HTTP cache tests


dincho:
HTTP cache options has very thin layer of tests, so a bugs are found while we tried to deploy this to production.

Never had time to reproduce it locally and extensively document it, but IIRC it was in the mickroblocks API

dincho:
details in the PR, but hard to rebase to master

PR #3352, opened 2020-09-11
Add more HTTP cache integration tests

dincho:
rel #3301, however it's not completed


dincho:
very hard to align to master

PR #3361, opened 2020-09-24
JSON ACI encoder

gorbak25:
This pull request improves Sophia smart contract handling in the node.
While compiling a contract we generate in parallel an JSON ACI and store it in an ETS table.
All calls to the compiler for encoding call data were replaced by https://github.com/aeternity/aesophia_aci_encoder
This library will be used in Hyperchains in order to efficiently interact with the staking contracts and possibly other contracts(generalized accounts defined by parent chain connectors). Decoding call data will be added soon to the ACI library.


rebar.lock
```diff
@@ -20,6 +20,18 @@
   {git,"https://github.com/aeternity/aeserialization.git",
       {ref,"80ff8a49f4765db390d6194beae4662e6872ad85"}},
   0},
+ {<<"aesophia">>,
```
hanssv, line 23: Let's not put a random commit from the `master/iris` branch of `aesophia` in the aeternity node...

hanssv (changes requested):

apps/aechannel/test/aesc_txs_SUITE.erl
```diff
@@ -5769,7 +5769,8 @@ encode_call_data(ContractName, Function, Arguments) ->
 
 address_encode(Type, Binary) ->
     case protocol_version() of
-        Vsn when Vsn < ?FORTUNA_PROTOCOL_VSN; Type == hash ->
+        %%Vsn when Vsn < ?FORTUNA_PROTOCOL_VSN; Type == hash ->
```
radrow, line 5772: Do we need this commented line?

apps/aecontract/test/aect_test_utils.erl
```diff
@@ -400,6 +434,20 @@ encode_call_data(Code, Fun, Args) ->
...
+        true ->
+            Aci = generate_json_aci(Vsn, Code),
+            Args1 = string:join([ to_str(Arg) || Arg <- Args ], ", "),
+            Args2 = to_str(Fun) ++ "(" ++ Args1 ++ ")",
+            aeaci_aci:encode_call_data(Aci, Args2);
+        false ->
+            slow_encode_call_data(Vsn, Code, Fun, Args)
+    end.
+
+slow_encode_call_data(Vsn, Code, Fun, Args) ->
```
radrow, line 450: I love your explicitness

radrow (approved):

apps/aechannel/test/aesc_txs_SUITE.erl
gorbak25: I left it here for clarity

hanssv (commented):
I'm missing the option to turn this off... ~If~When you want to develop the compiler/FATE further it would be useful to be able to run the tests without having to rely on a third component (that might not yet be updated). Would that make sense?

Also out of curiosity, what is the speedup compared to the (cached) call data creation? It should be really significant compared to calling the CLI compiler :-)

gorbak25:
> I'm missing the option to turn this off... 
Will add it in a moment.

~If~When you want to develop the compiler/FATE further it would be useful to be able to run the tests without having to rely on a third component (that might not yet be updated). Would that make sense?
Good point, especially the ACI library misses some features right now(Bits type, etc..).
 
> Also out of curiosity, what is the speedup compared to the (cached) call data creation? It should be really significant compared to calling the CLI compiler :-)
Well, the main speedup comes from the ets cache, looking at the CI runs of ct-lima the aci compiler only improves the situation marginally by at most 1-2 minutes which was expected. 
The main reason for writing this library is not to speed up tests slightly but to have a flexible method for calling contracts in production, without this component any Hyperchain related contract would require a hardcoded encoder/decoder which would need to be kept in sync with the contract source code. The HC staking contract will be modified pretty often and also we have plans for parent chain connector defined smart contracts which also will be changed a lot in development - I know that we could just hardcode all encodings/decodings but this would lack the flexibility we need.

hanssv:
By (only) using this library we do loose a bit of testing of the compiler though... But calldata is fairly stable so no big issue I think.

apps/aecontract/test/aect_test_utils.erl
```diff
@@ -529,8 +535,10 @@ generate_json_aci(Vsn, Code) ->
     AciId = #aci_cache_id{code_hash = crypto:hash(md5, Code), backend = Backend},
     case ets:lookup(?ACI_TAB, AciId) of
         [#aci_cache_entry{result = Result}] ->
+            ct:log("ACI cache HIT :)"),
             Result;
         [] ->
+            ct:log("ACI cache MISS :)"),
```
radrow, line 544: There should be a sad face imo :(

hanssv (approved):
LGTM

radrow (approved):

PR #3364, opened 2020-10-01
Aesophia rebar plugin

gorbak25:
This PR adds a rebar plugin for interacting with the sophia compiler when building the node.
This will be used in the build system of hyperchains.

The plugin introduces 2 neat operations:
- Compile and generate the ACI:
```
./rebar3 aesophia -c contract.aes -o result.json
```
- Verify the contract bytecode:
```
./rebar3 aesophia -c contract.aes -o result.json -v
```

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -108,7 +108,10 @@
 
        ]}.
 
-{plugins, [{swagger_endpoints, {git, "https://github.com/aeternity/swagger_endpoints", {ref, "d5a9f07"}}}]}.
+{plugins, [
+        {swagger_endpoints, {git, "https://github.com/aeternity/swagger_endpoints", {ref, "d5a9f07"}}},
+        {aesophia_rebar_plugin, {git, "https://github.com/aeternity/aesophia_rebar_plugin", {ref, "df113cc"}}}
+    ]}.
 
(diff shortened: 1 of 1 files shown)


hanssv (commented):
Does the plugin have a dependency on a compiler? If so, this should be a test dependency (is there such a thing for a plugin?)?

gorbak25:
@hanssv 
The plugin depends on aesophia_cli. 
I've checked that in the final release plugins are stripped away - the plugin is meant only to be used in the build process. 
Plugin dependencies are kept separate from the normal dependencies.

hanssv:
The build process of the normal node should _not_ depend on the compiler...

radrow (approved):

erlmachinedev (approved):

hanssv (commented):
This is the Aeternity node? Surely the HyperChains node will be a separate thing?

hanssv:
For the Hyperchain node it sounds reasonable, but I can't see why it wouldn't be sufficient to just have that fixed version of `aesophia_cli` as a dependency and just do `aesophia_cli ...` instead of going through `rebar3` - though I am perhaps overly negative about `rebar3`, it isn't on my Xmas card list 😄

PR #3370, opened 2020-10-12
Improve block insertion time

gorbak25:
Preliminary refactoring of block validation to prepare for a customizable consensus engine.
1. Reduce the consumed IO resources by caching blocks required for target calculation in ram
2. Fail when the block is clearly invalid before starting the DB transaction 
3. Don't write the genesis hash to the db on every single block
4. Remove some unnecessary reads for the previous block/keyblock
5. Better benchmarks

My next PR will move the rest of the header validation logic outside the DB transaction - the validation of keyblocks will be made in the customizable consensus engine.

### DIFF
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -1169,10 +1169,6 @@ db_get_header(Hash) when is_binary(Hash) ->
     {value, Header} =  aec_db:find_header(Hash),
     Header.
 
-dirty_db_get_header(Hash) when is_binary(Hash) ->
-    {value, Header} =  aec_db:dirty_find_header(Hash),
-    Header.
-
 db_find_key_nodes_at_height(Height) when is_integer(Height) ->
     case aec_db:find_headers_and_hash_at_height(Height) of
         [_|_] = Headers ->
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -514,60 +514,109 @@ internal_insert(Node, Block, Origin) ->
     end.
 
 %% Builds the insertion context from cached data and the node to insert
-%% Please note that this is called in dirty context
+%% Please note that this is called in dirty context without a try
+%% clause - don't hard crash here.
+%% Performs basic checks to ensure the chain consistency
 build_insertion_ctx(Node, micro, undefined) ->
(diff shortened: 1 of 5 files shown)


apps/aecore/src/aec_chain_state.erl
```diff
@@ -1310,3 +1412,27 @@ fork_result(Count, #{signalling_block_count := SigCount}) when Count >= SigCount
...
     false.
+
+setup_etc_cache() ->
+    %% Lager might not be up so don't bother with logging
+    Self = self(),
+    Tab = ?RECENT_CACHE,
+    Keypos = #recent_blocks.key,
+    case ets:info(Tab, name) of
+        undefined ->
+            spawn(fun() ->
```
radrow, line 1423: Why timeout?

radrow (approved):

erlmachinedev:
Oh, here is a great amount of significant work (I even afraid to put any criticism here) You are 🤖. @gorbak25 if you don't mind I share some thoughts about this great сontribution :

1) It could be nice to separate this functionality into the appropriate component like  "**aec_chain_recent_state**";
2) To describe this component and to reflect it's purpose via **exported API**;   
3) I know that Erlang processes don't break without reason (but ?RECENT_CACHE table which is spawned and stubbed by infinite sleep looks scary, taking into account the consequences of denying). Did you consider to put it into the **supervision tree**?

So my opinion is that complexity is decreased and we have to think about future devs too. I consider that functionality is important and which is deserved to be decoupled from app design perspective.

erlmachinedev (approved):

gorbak25:
More cleanup will be done with the consensus engine PR - the insertion context/cache will become part of the consensus module. The sleeping process will be moved under the supervision tree when I will introduce periodic GC of orphaned caches.

radrow (approved):

gorbak25:
Before merging I will run a sync test from genesis to ensure that switching to dirty reads didn't break the node.

gorbak25:
So far no hiccups
image

apps/aecore/src/aec_chain_state.erl
uwiger: I think this is good practice, even though it is very unlikely to trigger (but not impossible).

uwiger (approved):

PR #3372, opened 2020-10-15
Optional dirty tree backend

gorbak25:
Using dirty reads for our merkle trees is fine, as the in ram cache we got in place is enough to ensure that we don't mutate outdated/unavailable data(which might be returned by a dirty read). By enabling dirty reads, it turned out that one test case in CI randomly failed - aehttp_sc_SUITE ==> plain.sc_ws_min_depth_is_modifiable
Some components in our node, like the state channel chain watcher might try to read the channel state instantly after the new block got inserted to the DB - as the tree cache is not globally shared between processes we cannot rely on dirty reads in this case.

This PR changes the DB to return a "safe" tree db backend unless the caller explicitly asks for a dirty version.

### DIFF
--- a/apps/aechannel/src/aesc_state_tree.erl
+++ b/apps/aechannel/src/aesc_state_tree.erl
@@ -16,6 +16,7 @@
          lookup/2,
          mtree_iterator/1,
          new_with_backend/1,
+         new_with_dirty_backend/1,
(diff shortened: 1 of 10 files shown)


uwiger (approved):
LGTM

radrow (approved):

PR #3373, opened 2020-10-16
Improve peer pools handling

velzevur:
This PR improves handling of peers:
* Significantly speeds up the TCP probes. Those used to be done roughly once
  every 2 minutes while now they are executed every second. Note that this is
  done only if there are peers to be checked: if a peer is missing, there is a
  grace period in which we mark it as 'suspended', waiting for its return. If
  all peers are checked, we wait longer - until either a new peer is added or
  there is a not-suspended peer. Basically the probes come on demand. Even
  despite the probes are fast, there is a maximum size of currently ongoing
  probes.
* Fixes the process of probing `verified` peers. Until now we were probing
  only `unverified` peers. If a peer responds to the probe - we consider it
  `verified`. We downgrade a `verified` peer only after we fail connecting to
  it. This resulted in amounting peers that used to be `verified` at one point
  of time. This is fixed now as we probe `verified` peers as well. What is
  more is, that this PR plays really well with PR #3365.
  This means that if a peer goes down, we will find it significantly sooner
  and will downgrade it as an `unverified` one. This is important as a nodes
  shares to other nodes only peers one considers to be `verified`.

### DIFF
--- a/apps/aecore/src/aec_peers.erl
+++ b/apps/aecore/src/aec_peers.erl
@@ -452,7 +452,7 @@ handle_call({connection_closed, PeerId, PeerCon}, _From, State0) ->
     {reply, ok, update_peer_metrics(schedule_connect(State))};
 handle_call({peer_dead, PeerId, PeerCon}, _From, State0) ->
     State = on_peer_dead(PeerId, PeerCon, State0),
-    {reply, ok, update_peer_metrics(State)};
+    {reply, ok, update_peer_metrics(maybe_start_tcp_probe_timers(State))};
 handle_call({peer_connected, PeerId, PeerCon}, _From, State0) ->
     {Result, State} = on_peer_connected(PeerId, PeerCon, State0),
     {reply, Result, update_peer_metrics(schedule_connect(State))};
@@ -461,7 +461,7 @@ handle_call({peer_accepted, PeerAddr, PeerInfo, PeerCon}, _From, State0) ->
     {reply, Result, update_peer_metrics(State)};
 handle_call({peer_alive, PeerId, PeerCon}, _From, State0) ->
     {Result, State} = on_peer_alive(PeerId, PeerCon, State0),
-    {reply, Result, update_peer_metrics(State)}.
+    {reply, Result, update_peer_metrics(maybe_start_tcp_probe_timers(State))}.
 
 handle_cast({log_ping, Outcome, PeerId, _Time}, State) ->
     update_ping_metrics(Outcome),
@@ -864,12 +864,12 @@ maybe_start_tcp_probe_timers(State) ->
     _State = maybe_start_tcp_probe_timer(unverified, State1).
 
 -spec maybe_start_tcp_probe_timer(peer_pool_name(), state()) -> state().
-maybe_start_tcp_probe_timer(verified, #state{ last_tcp_verified_probe_time = undefined } = State) ->
+maybe_start_tcp_probe_timer(verified, #state{ tcp_verified_probe_ref = undefined } = State) ->
     Delay = initial_tcp_probe_delay(),
     RefVerified = start_timer({tcp_probe, verified}, Delay),
     State#state{ last_tcp_verified_probe_time = timestamp(),
                  tcp_verified_probe_ref = RefVerified };
-maybe_start_tcp_probe_timer(unverified, #state{ last_tcp_unverified_probe_time = undefined } = State) ->
+maybe_start_tcp_probe_timer(unverified, #state{ tcp_unverified_probe_ref = undefined } = State) ->
     Delay = initial_tcp_probe_delay(),
     RefUnverified = start_timer({tcp_probe, unverified}, Delay),
     State#state{ last_tcp_unverified_probe_time = timestamp(),
--- a/apps/aecore/src/aec_peers.erl
+++ b/apps/aecore/src/aec_peers.erl
@@ -74,12 +74,13 @@
 -define(MAX_CONNECTION_INTERVAL,       30 * 1000). % 30 seconds
 
 -define(MIN_TCP_PROBE_INTERVAL,              100).
--define(MAX_TCP_PROBE_INTERVAL,       120 * 1000). % 2 minutes
+-define(MAX_TCP_PROBE_INTERVAL,       1 * 1000). % 1 second 
 
 -define(DEFAULT_PING_INTERVAL,        120 * 1000). %  2 minutes
 -define(DEFAULT_UNBLOCK_INTERVAL, 15 * 60 * 1000). % 15 minutes
 -define(DEFAULT_MAX_INBOUND,                 100).
 -define(DEFAULT_MAX_OUTBOUND,                 10).
+-define(DEFAULT_MAX_TCP_PROBES,              100).
 -define(DEFAULT_SINGLE_OUTBOUND_PER_GROUP,  true).
 -define(DEFAULT_RESOLVE_MAX_SIZE,            100).
 -define(DEFAULT_RESOLVE_MAX_RETRY,             7).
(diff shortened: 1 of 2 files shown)


velzevur:
In order to verify and measure the improvements, I will kindly ask @dincho to spawn 2 nodes:
ControlGroupNode - a node running the latest release version
TestNode - a node running suggested changes on top of master (so  #3365 is included as well)

Those will run in a parallel for a few days and we would check how their pools are going. The main focus goes on the `verified` pool.

gorbak25 (dismissed):

apps/aecore/src/aec_peers.erl
```diff
@@ -734,21 +746,27 @@ connect(State0) ->
...
-                             "waiting ~b ms", [Delay]),
-            schedule_tcp_probe(Delay, State2);
-        {selected, Peer, State2} ->
-            schedule_tcp_probe(undefined, start_connection(tcp, Peer, State2))
+%% Connects to a random peer from verified or the unverified pool to check if
+%% the selected peer is alive; schedule the next TCP probe timeout
+-spec tcp_probe(peer_pool_name(), state()) -> state().
+tcp_probe(PeerPoolName, State0) ->
+    State = cancel_tcp_probe_timer(PeerPoolName, State0),
+    case is_tcp_probe_allowed(State0) of
```
uwiger, line 754: Is the reference to `State0` instead of `State` deliberate?

uwiger: If so, perhaps this could be made clear with e.g.
```erlang
ProbeAllowed = is_tcp_probe_allowed(State0),
State = cancel_tcp_probe_timer(PeerPoolName, State0),
case ProbeAllowed of
   ...
end
```

apps/aecore/src/aec_peers.erl
```diff
@@ -734,21 +746,27 @@ connect(State0) ->
...
+        false ->
+            schedule_tcp_probe(?MAX_TCP_PROBE_INTERVAL, PeerPoolName, State0);
+        true ->
+            case pool_random_select(PeerPoolName, State) of
+                {unavailable, State2} ->
+                    epoch_sync:debug("No peers available for TCP probe", []),
```
uwiger, line 760: Should the `PeerPoolName` be included in the debug msg?

apps/aecore/src/aec_peers.erl
```diff
@@ -734,21 +746,27 @@ connect(State0) ->
...
+                    schedule_tcp_probe(?MAX_TCP_PROBE_INTERVAL, PeerPoolName, State2);
+                {wait, Delay, State2} ->
+                    epoch_sync:debug("No peers available for TCP probe, "
```
uwiger, line 763: Include `PeerPoolName`?

apps/aecore/src/aec_peers.erl
```diff
@@ -787,18 +805,17 @@ schedule_connect(MinDelay, State0) ->
...
--spec schedule_tcp_probe(undefined | non_neg_integer(), state()) -> state().
-schedule_tcp_probe(MinDelay, State0) ->
-    State = cancel_tcp_probe_timer(State0),
-    NextDelay = next_tcp_probe_delay(State),
+-spec schedule_tcp_probe(undefined | non_neg_integer(), peer_pool_name(), state()) -> state().
+schedule_tcp_probe(MinDelay, PeerPoolName, State0) ->
+    State = cancel_tcp_probe_timer(PeerPoolName, State0),
+    NextDelay = next_tcp_probe_delay(PeerPoolName, State),
     Delay = safe_max(MinDelay, NextDelay),
     epoch_sync:debug("Waiting ~b ms for the next TCP probe", [Delay]),
```
uwiger, line 813: Include `PeerPoolName`?

apps/aecore/src/aec_peers.erl
```diff
@@ -814,19 +831,19 @@ next_connect_delay(State) ->
...
-next_tcp_probe_delay(State) ->
-    %% TODO: this probably needs some tweaks, it should be also dependent on the
-    %% number of available unverified peers, number of verified peers vs
-    %% max_outbound, and maybe it could start to be more frequent when there are
-    %% more timeouts while connecting and the number of outbound connections is
-    %% low, etc.
-    #state{ last_tcp_probe_time = LastTime, outbound = Outbound } = State,
+-spec next_tcp_probe_delay(peer_pool_name(), state()) -> non_neg_integer().
+next_tcp_probe_delay(PeerPoolName, State) ->
+    LastTime =
```
uwiger, line 841: Could be its own function.

apps/aecore/src/aec_peers.erl
```diff
@@ -1272,10 +1315,12 @@ on_connect_timeout(_Ref, State) ->
...
-on_tcp_probe_timeout(Ref, #state{ tcp_probe_ref = Ref } = State) ->
-    tcp_probe(State#state{ tcp_probe_ref = undefined });
-on_tcp_probe_timeout(_Ref, State) ->
+-spec on_tcp_probe_timeout(reference(), peer_pool_name(), state()) -> state().
+on_tcp_probe_timeout(Ref, verified, #state{ tcp_verified_probe_ref = Ref } = State) ->
+    tcp_probe(verified, State#state{ tcp_verified_probe_ref = undefined });
+on_tcp_probe_timeout(Ref, unverified, #state{ tcp_unverified_probe_ref = Ref } = State) ->
+    tcp_probe(unverified, State#state{ tcp_unverified_probe_ref = undefined });
+on_tcp_probe_timeout(_Ref, _, State) ->
     epoch_sync:info("TCP probe timeout with invalid reference", []),
```
uwiger, line 1324: Perhaps also include the `PeerPoolName` here as well, since now the `Ref` could be valid, but the pool name wrong...

apps/aecore/src/aec_peers.erl
uwiger: But the test doesn't seem to need to reference `State0` ...

apps/aecore/src/aec_peers.erl
velzevur: Since it is only used in this small function, I'd rather keep it self contained in it.

uwiger (dismissed):

gorbak25:
@velzevur @uwiger I've set up the experiment on my k8s cluster :)
Control node: https://control-ae-sync-test.gorbak25.eu/v2/status
PR node: https://pr-ae-sync-test.gorbak25.eu/v2/status
Public dashboard: https://ae-sync-test.gorbak25.eu/d/RBn2zktGk/ae-sync-test?orgId=1&refresh=10s&from=now-6h&to=now
So far it looks that indeed this PR improves the situation:
image

I've used the following sync config for both the control and test node:
```
aecore:
    - sync_max_outbound: 100
    - sync_max_inbound: 1000
    - ping_interval: 200 
```

velzevur:
Great work @gorbak25 !
It looks like a few hours ago the `unverified` peers started crawling up:
Screenshot 2020-10-19 at 10 45 00
It could be a symptom for a bug in the PR above or maybe the node is receiving suspiciously a lot of new peers...

gorbak25:
@velzevur the logs are flooded with the following:
image

uwiger (approved):

velzevur:
@gorbak25 those error seem unrelated to the (dead) peers issue. The messages above are a symptom for a race in the sync itself

gorbak25:
@uwiger @velzevur 
New version deployed on: https://pr-v2-ae-sync-test.gorbak25.eu/v2/status
Stats should be available on the public dashboard
image

gorbak25:
For the record I'm leaving the latest state of the dashboard here in case I need to reclaim some resources on my server:
image

gorbak25 (approved):

PR #3385, opened 2020-10-30
Conductor benchmarks

gorbak25:
#3377 got really big and I'm not even close finishing this :(
I've started to separate some parts into more manageable PR's which can be merged independently of #3377.

This PR introduces benchmarks for aec_conductor:add_synced_block - it turned out that the overhead of postprocesing blocks was absurdly large(#3377 has the fix for it).

PR #3377, opened 2020-10-19
Customizable consensus engine

gorbak25:
This feature is in very early stages and development is rather slow, I'm only opening a PR so It's clear for my teammates on what I'm working on right now.

PR #3386, opened 2020-10-30
Refactor MPT to use behaviours for the db backend

gorbak25:

This PR refactors the DB backend of MPT trees to use behaviors instead of tuple funs - this surprisingly increased the sync performance dramatically as calls to validate_exported were performed unbelievably often and were expensive.
Before(This includes DB and conductor optimisations not present in this PR):
image
After(the previous optimisations + this PR):
image

Merge after #3387 #3385 #3388


velzevur:
Great work!

velzevur (approved):

radrow (approved):

PR #3387, opened 2020-10-30
E2E sync benchmarks

gorbak25:

This PR introduces end to end synchronization benchmarks between two nodes - the measurements are more representative compared to other benchmarks as they consider things like network throughput, aec_sync, aec_conductor, etc...
Those tests are pretty heavy and are really useful while tweaking the DB - Those benchmarks are meant to be run manually in the console by a dev - they are disabled by default and they log timings to the console - to activate those tests one passes in the environment SYNC_BENCHMARK=1

### DIFF
--- a/apps/aecore/test/aecore_sync_SUITE.erl
+++ b/apps/aecore/test/aecore_sync_SUITE.erl
@@ -35,7 +35,9 @@
     report_metrics/1,
     check_metrics_logged/1,
     crash_syncing_worker/1,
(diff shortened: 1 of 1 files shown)


radrow (approved):

erlmachinedev (approved):
It could be nice to make some refactoring later (via common_test convenient configuration). As a developer, I may want to see some documented info about execution scenarios (we can forget SYNC_BENCHMARK param).

PR #3388, opened 2020-10-30
Fix timestamp check in contract tests

gorbak25:

#3377 and #3386 improve performance so much that this check fails as most of the time Time1 == Timestamp :)


radrow (approved):

erlmachinedev (approved):
Eventually, os:timestamp() is global blocking call, if you are fast enough you can match sure :)

PR #3391, opened 2020-10-30
Why?

gorbak25:

Fixes execution time of some SC tests.
@uwiger is there a good reason for this sleep? Should we leave it in production and just remove it in tests?

### DIFF
--- a/apps/aehttp/src/sc_ws_handler.erl
+++ b/apps/aehttp/src/sc_ws_handler.erl
@@ -266,7 +266,7 @@ websocket_handle_(_Data, H) ->
 
 websocket_info(?ERROR_TO_CLIENT(Err), #handler{protocol = Protocol} = H) ->
     {reply, Resp} = sc_ws_api:error_response(Protocol, Err),
-    timer:sleep(1000),
+    %%timer:sleep(1000),
     lager:info("Handler critical error: ~p", [Err]),
     {reply, {text, jsx:encode(Resp)}, H};
 websocket_info(stop, #handler{} = H) ->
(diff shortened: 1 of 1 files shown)


uwiger:
> @uwiger is there a good reason for this sleep? Should we leave it in production and just remove it in tests?

I imagine it's left behind from some debugging spree.

uwiger (approved):

radrow:
grafik

apps/aehttp/src/sc_ws_handler.erl
```diff
@@ -266,7 +266,7 @@ websocket_handle_(_Data, H) ->
 
 websocket_info(?ERROR_TO_CLIENT(Err), #handler{protocol = Protocol} = H) ->
     {reply, Resp} = sc_ws_api:error_response(Protocol, Err),
-    timer:sleep(1000),
+    %%timer:sleep(1000),
```
radrow, line 269:
```suggestion
```
We got the allusion from that comment

radrow (changes requested):

radrow (approved):

erlmachinedev:
The timer:sleep(1000) isn't a solution and hides the real async issue. The best way is to remove sleep and to provide a fix of the bug that listed on the screenshot. 

I propose to open an appropriate issue to avoid loss of that discussion.

gorbak25:
@erlmachinedev I think that random failures of plain.sc_ws_min_depth_is_modifiable might be fixed by #3389

erlmachinedev (approved):

PR #3394, opened 2020-11-03
Fix CI builds

gorbak25:


.circleci/config.yml
```diff
@@ -928,6 +932,8 @@ jobs:
   osx_package:
     macos:
       xcode: "10.1.0" # 10.1.0 is macOS 10.13.6
+    environment:
+      ERLANG_ROCKSDB_BUILDOPTS: "-j2"
```
dincho, line 936: this runs in VM and I don't think you need that, that would slow the builds probably

erlmachinedev (dismissed):

dincho (approved):
better drop the -j2 from OSX

velzevur (approved):

uwiger (approved):

erlmachinedev (approved):

PR #3403, opened 2020-11-06
Update mkdocs.yml

nikita-fuchs:
in order to improve the user experience, I use these two plugins to add usefull information and annotations in the other docs, please merge so we have the available here, too.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -8,6 +8,11 @@ theme:
         - yaml
         - json
         - bash
+        
+markdown_extensions:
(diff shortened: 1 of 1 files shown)


omar-saadoun:
What else is needed here to get this merged?

velzevur:
Not needed, anyone can add those plugins locally if they want to.

dincho:
The mentioned plugins are not used in this repo. Please install them upstream or provide a detailed explanation why we should consider this.

PR #3414, opened 2020-11-25
Fix issue in aec_chain_state

### LINKED ISSUE #3311: When syncing from backup, accept previous states in DB if they don't differ
After unzipping DB from https://downloads.aeternity.io

and starting node, there are errors during sync: `found_already_calculated_state`

Coming from: aec_chain_state:update_state_tree/4

If the states are identical, we should ignore this error.

velzevur:
The service download.aeternity.io is fed some database snapshots. Those are
produced by simply grabbing mnesia files and publishing them. If at the moment
of moving the data a block was in the process of being consumed by the node
itself, there is a risk of a race condition: the block (wrapped as a MPT node)
is being included in the DB while it is not yet set to be the top block. This
PR identifies if a node's hash is already known and if so - checks if it
produces the same expected result. If so - it is accepted. In the case of a DB
node being found with the same hash but different state trees - the node is
still crashing, no change there.

While working on the PR, 2 small bugs were identified and fixed, please
consult the comments below.

Fixes #3311

### DIFF
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -987,7 +987,7 @@ update_fraud_info(ForkInfoIn, Node, State) ->
 
 handle_top_block_change(OldTopHash, NewTopDifficulty, Node, Events, State) ->
     case get_top_block_hash(State) of
-        OldTopHash -> {State, []};
+        OldTopHash -> {State, no_events()};
         NewTopHash when OldTopHash =:= undefined ->
             NewTopHash = get_genesis_hash(State), %% Internal inconsistency check
             State1 = update_main_chain(NewTopHash, NewTopHash, NewTopHash, State),
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -103,7 +103,11 @@
         ]).
 
 -ifdef(TEST).
--export([calc_rewards/6]).
+-export([calc_rewards/6,
+         wrap_block/1,
+         internal_insert_transaction/4,
+         build_insertion_ctx/3
+        ]).
 -endif.
 
 -include_lib("aeminer/include/aeminer.hrl").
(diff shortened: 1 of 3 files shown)


apps/aecore/src/aec_chain_state.erl
```diff
@@ -972,7 +987,7 @@ update_fraud_info(ForkInfoIn, Node, State) ->
 
 handle_top_block_change(OldTopHash, NewTopDifficulty, Node, Events, State) ->
     case get_top_block_hash(State) of
-        OldTopHash -> State;
+        OldTopHash -> {State, []};
```
velzevur, line 990: Small bug: the function returns a tuple of a new state and events. If the change had not changed, no new events.

This module would surely benefit from some `-spec`-s

apps/aecore/src/aec_db.erl
```diff
@@ -734,7 +734,7 @@ find_signal_count(Hash) ->
 add_tx_location(STxHash, BlockHash) when is_binary(STxHash),
                                          is_binary(BlockHash) ->
     ?t(mnesia:write(#aec_tx_location{key = STxHash, value = BlockHash}),
-       [aec_tx_location, STxHash]).
+       [{aec_tx_location, STxHash}]).
```
velzevur, line 737: Possibly a typo.

apps/aecore/test/aec_chain_state_tests.erl
```diff
@@ -487,3 +487,45 @@ make_spend_tx(Sender, SenderNonce, Recipient, Fee, Amount) ->
...
+     ]}.
+
+accept_existing_db_node() ->
+    [B1, B2, B3] = prep_micro_blocks(3),
+    {ok, _} = ?TEST_MODULE:insert_block(B1),
+    {ok, _} = ?TEST_MODULE:insert_block(B2),
+    Node = ?TEST_MODULE:wrap_block(B3),
+    Ctx = ?TEST_MODULE:build_insertion_ctx(Node, micro, undefined),
+    {ok, _} = ?TEST_MODULE:insert_block(B3),
+    ?TEST_MODULE:internal_insert_transaction(Node, B3, undefined, Ctx),
```
velzevur, line 529: This goes around the dirty check if the node is present in the DB. In normal cases those requests are serialised by the `aec_conductor` process so the dirty read shall be sufficient.

apps/aecore/src/aec_chain_state.erl
gorbak25, line 990:
```suggestion
        OldTopHash -> {State, no_events()};
```

gorbak25 (dismissed):

uwiger (approved):

gorbak25 (approved):

Issue #3355, opened 2020-09-14
Fast synchronization


velzevur:
Right now all nodes operate as “archive” nodes, most people want to quickly get the latest state so optional fast synchronization algorithms need to be implemented and provided as an option for users/node operators - after fast sync is done it would be nice to optionally sync older states(configurable policies as in geth)

PR #3415, opened 2020-11-26
Gb 3355 parallel tx eval 2

uwiger:
See issue #3355. This PR attempts to (eventually) speed up block evaluation by processing heavy transactions in parallel.

Currently, the implementation has been tested with spend txs, either accessing shared resources (same account) or being independent. The `aehttp_contract_SUITE` also double-checks with parallel eval (via dry-run over HTTP), but using only singleton lists of contract call transactions.

A preliminary performance test indicates that for now, the overhead of parallel eval is roughly 10x that of sequential eval. This should be possible to reduce, e.g. by removing the `lager:debug()` calls, but it is clear that e.g. parallel eval of `spend` txs will never pay off, since these basically only update the state trees. The expectation is that especially complex contract evaluations will benefit from parallel evaluation.

The tracing improvements rely on the uwiger/trace_runner#3 PR. The commit hash will change once that PR is merged.


apps/aecore/src/aec_trees.erl
```diff
@@ -105,13 +134,58 @@
...
                                      , aens_state_tree
-                                     , aeo_state_tree ]}.
+                                     , aeo_state_tree | aetx:tx_callbacks() ]}.
+
+-dialyzer([{nowarn_function, pp_term/1}, no_opaque]).
+pp_term(#trees{ accounts  = A
+              , calls     = C
+              , channels  = Ch
+              , contracts = Co
+              , ns      = Ns
```
gorbak25, line 151:
```suggestion
              , ns        = Ns
```

gorbak25 (changes requested):
1. Doesn't fully work - #3424 shows an failing test case
2. When restarting a worker due to dependencies we redo the (probably) already done signature check - By running the tests with {dont_verify_signature, true} the overhead between parallel and sequential TX processing becomes more reasonable
3. I don't see the option of passing the Eval CTX from the outside - If parallel tx eval got enabled by default then we will build the deps before entering aec_conductor.
4. I think we might enable this by default in case we implement #3419 - the preprocesing will be done while other blocks are being inserted - minimizing the hot path :)

apps/aecore/test/aec_trees_tests.erl
```diff
@@ -738,8 +835,160 @@ ct_create_tx(Sender, VmVersion, ABIVersion) ->
     {ok, Tx} = aect_create_tx:new(Spec),
     Tx.
 
+apply_txs_on_state_trees(SignedTxs, TreesIn, Env, Opts) ->
+    T0 = erlang:system_time(microsecond),
+    SeqRes = ?TEST_MODULE:apply_txs_on_state_trees(SignedTxs, TreesIn, Env, Opts),
```
gorbak25, line 840: I would rather use timer:tc :)

apps/aecore/src/aec_trees_proxy.erl
```diff
@@ -0,0 +1,592 @@
...
+
+tx_down(_Ix, Pid, {'$caught', Type, What}, #st{strict = true} = St) ->
+    throw_error(Type, What, Pid, St);
+tx_down(Ix, _Pid, Reason, #st{ status = Status
+                             , events = Events0 } = St) ->
+    %% All valid txs must have at least some claims.
+    {IsValid, Events} =
+        case Reason of
+            {ok, NewEvents} ->
+                {true, Events0 ++ NewEvents};
```
gorbak25, line 174: This is linear in the first list - this might become an issue when many events were emitted

apps/aecore/src/aec_trees.erl
```diff
@@ -583,59 +780,112 @@ internal_commit_to_db(Trees) ->
...
+-spec apply_txs_on_state_trees(signed_txs(), trees(), env(), opts()) -> apply_result().
+apply_txs_on_state_trees(SignedTxs, Trees, Env, Opts) ->
+    apply_txs_on_state_trees(SignedTxs, [], [], Trees, Env, Opts).
+
+-spec par_apply_txs_on_state_trees(signed_txs(), valid_txs(), invalid_txs(),
+                                   trees(), env(), opts()) -> apply_result().
+par_apply_txs_on_state_trees(SignedTxs, Valid, Invalid, Trees, Env, Opts) ->
+    ParMaxDeps = proplists:get_value(par_max_deps_ratio, Opts, undefined),
+    case aec_trees_proxy:prepare_for_par_eval(SignedTxs) of
+        #{tx_count := TxC, max_dep_count := DepC} when (DepC / TxC) > ParMaxDeps ->
```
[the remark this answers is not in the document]
uwiger: Yes, this was a a deliberate trick to short-circuit that check. Should have added a comment about it, of course. :)

velzevur:
Left open so we can extract the block witness in case we consider it useful.

PR #3416, opened 2020-11-26
Remove TTL from aega_meta_tx

### LINKED ISSUE #3056: meta_tx's TTL
`aetx:ttl/1` specializes the inner tx and calls its callback's `ttl/1`. In the case of a channel co-authenticated transaction when both participants are GAs, that would result in two embedded meta transactions. Calling `aetx:ttl/1` that would result in the outermost meta_tx's ttl. 

Instead `aetx:ttl/1` should return the innermost transaction's `ttl/1`.

velzevur:
Closes #3056

### DIFF
--- a/apps/aega/src/aega_meta_tx.erl
+++ b/apps/aega/src/aega_meta_tx.erl
@@ -46,7 +46,8 @@
 -export([set_tx/2
         ]).
 
--define(GA_META_TX_VSN, 1).
+-define(NO_TTL_VSN, 2).
+-define(PRE_IRIS_VSN, 1).
 -define(GA_META_TX_TYPE, ga_meta_tx).
 
 %% Should this be in a header file somewhere?
@@ -63,7 +64,7 @@
           gas         :: amount(),
           gas_price   :: amount(),
           fee         :: amount(),
-          ttl         :: aetx:tx_ttl(),
+          ttl         :: aetx:tx_ttl() | no_ttl,
           tx          :: aetx_sign:signed_tx()
         }).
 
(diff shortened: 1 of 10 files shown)


apps/aega/src/aega_meta_tx.erl
```diff
@@ -144,7 +148,7 @@ new(#{ga_id       := GAId,
                      gas         = Gas,
                      gas_price   = GasPrice,
                      fee         = Fee,
-                     ttl         = maps:get(ttl, Args, 0),
```
hanssv, line 147: This will not work? We are not always running Iris :-)

hanssv (changes requested):

velzevur:
Tests failed with 
```

%%% aecore_sync_SUITE ==> all_nodes.performance.persistence.trusted_peer_is_untrusted_after_a_restart: FAILED
%%% aecore_sync_SUITE ==> {{badmatch,0},
 [{aecore_sync_SUITE,trusted_peer_is_untrusted_after_a_restart,1,
                     [{file,"/home/builder/aeternity/apps/aecore/test/aecore_sync_SUITE.erl"},
                      {line,970}]},
  {test_server,ts_tc,3,[{file,"test_server.erl"},{line,1754}]},
  {test_server,run_test_case_eval1,6,[{file,"test_server.erl"},{line,1263}]},
  {test_server,run_test_case_eval,9,[{file,"test_server.erl"},{line,1195}]}]}
```
This is orthogonal to this PR but fixed by #3418

apps/aega/src/aega_meta_tx.erl
hanssv: Only an issue in tests...

hanssv (approved):

uwiger (approved):

PR #3429, opened 2020-12-14
Mnesia bypass

gorbak25:
This bypasses the mnesia transaction manager and batches writes when using rocksdb as the DB backend - the E2E sync benchmarks noted a 2x speedup in sync speed. I've managed to sync the first 10k mainnet generations very quickly(2-3 minutes).
Most importantly this PR ensures that the DB is always consistent. This is a temporary fix before migrating our entire node to a proper DB(It's still possible to get corruption but at least we will NEVER write the MPT tree nodes partially).


apps/aecore/src/aec_db.erl
```diff
@@ -1052,6 +1096,85 @@ handle_activity_result(Res, ErrorKeys) ->
             exit(Err)
     end.
 
+do_activity(transaction, Fun) ->
+    try
+    mnesia:activity(transaction, fun() ->
+        R = Fun(),
+        case persistent_term:get(?BYPASS, no_bypass) of
+            {rocksdb, Handles} ->
+                {mnesia, _, {tidstore, TStore, _, _}} = get(mnesia_activity_state),
```
uwiger, line 1105: The TidStore table is a bag, and can contain sequences of writes/deletes to the same key. Thus, order is critical. At least at first glance, the walk seems to reverse the order of updates.

uwiger: Sorry, mnesia already resolves this before commit.

apps/aecore/src/aec_db.erl
```diff
@@ -837,6 +856,24 @@ wait_for_tables(Tabs, Sofar, _, _) ->
...
 
+%% Prepare bypass
+prepare_mnesia_bypass() ->
+    P = mnesia:table_info(aec_account_state, all),
+    TNames = [N || {N,_} <- tables()],
+    case proplists:get_value(rocksdb_copies, P, []) of
+        [] -> ok;
+        [_] ->
+            HIndex = 'mnesia_ext_proc_aec_headers-4-_ix',
+            {HIndexRef, set} = gen_server:call(HIndex, get_ref),
```
uwiger, line 867: Perhaps it would be cleaner to export `mnesia_rocksdb:get_ref(rocksdb_copies,Tab)`, which also by design resolves `mnesia_rocksdb:get_ref(rocksdb_copies, {Tab, index, IxPos})`.

BTW, the index positions (and type of index - with rocksdb, always ordered) can be fetched using `mnesia_lib:val(Tab, index_info)`

radrow (approved):

uwiger (dismissed):

velzevur (dismissed):

velzevur (approved):

PR #3437, opened 2021-01-01
Make inner tx of PayingForTx non-valid

### LINKED ISSUE #3054: Make inner transaction of PayingForTx non-valid
The initial implementation has a fully fledged normally signed inner transaction. This means that it is possible to unwrap the PayingForTx and post the inner transaction on-chain. Since the inner transaction isn't intended to be used like this it would be nice to disallow it.

The idea is to change what is signed for the inner transaction - the obvious idea is to drop the network-id from the signing schema but perhaps there are other ways as well.

velzevur:
Fixes #3054

There is a suggested approach in the issue: use empty `network_id` for signing the inner transaction. This has one big flaw: the inner transaction is repayable on a different network with a different `network_id`. Instead I took a slightly different approach: the inner transaction uses `<netwok_id>-<encoded payer_id>`, so for example it would be used `ae_mainnet-ak_123...`. Note that the payer account's ID is Base58c encoded.

### DIFF
--- a/apps/aecore/test/aec_test_utils.erl
+++ b/apps/aecore/test/aec_test_utils.erl
@@ -66,6 +66,7 @@
         , substitute_innermost_tx/2
         , sign_tx/3
         , sign_tx_hash/2
+        , sign_tx_with_specific_network_id_prefix/3
         , signed_spend_tx/1
         , wait_for_pubkey/0
         , min_gas_price/0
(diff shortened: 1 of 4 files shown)


apps/aecore/test/aec_test_utils.erl
```diff
@@ -631,17 +632,30 @@ sign_tx(Tx, PrivKey) ->
 sign_tx_hash(Tx, PrivKey) ->
     sign_tx(Tx, PrivKey, true).
 
+sign_tx_with_specific_network_id_prefix(Tx, PrivKey, PubKeyPrefix) ->
```
hanssv, line 635: Function is named "...network.." and variable "pubkey"... Not obvious.

apps/aega/test/aega_SUITE.erl
```diff
@@ -1033,7 +1033,9 @@ paying_for_outer(_Cfg) ->
                   nonce        => aect_test_utils:next_nonce(Acc2, State)}),
 
     PrivKey  = aect_test_utils:priv_key(Acc2, State),
-    SignedTx = aec_test_utils:sign_tx(SpendTx, PrivKey),
+    SignedTx = aec_test_utils:sign_tx_with_specific_network_id_prefix(SpendTx,
```
hanssv, line 1036: The long line is easier to read, but if you are on a 640x480 monitor, sure 😛

hanssv (commented):
I think we should have something static instead of the payer pubkey, it should be possible for "anyone" to pick up the inner tx.

uwiger (approved):

gorbak25 (changes requested):
Please forbid using "-inner_tx" in the network id.
If at some point someone launches "ae_mainnet-inner_tx" then inner transactions from ae_mainnet could be published on "ae_mainnet-inner_tx". When starting/booting the node please hard crash when the network id ends with "-inner_tx" so someone in the future doesn't accidentally shot himself in the foot.

hanssv:
I wonder if it isn't better to make a quiz with random detailed questions about AE that you have to answer correctly in order to be allowed to start the node? I don't think we will be able to stop truly people from shooting themselves in the foot otherwise?

hanssv (approved):

gorbak25 (approved):

PR #3441, opened 2021-01-05
Improve aestratum integration test stability

gorbak25:
I never reproduced the CI failure locally but I think this might be the fix for it.


velzevur (approved):

uwiger (approved):

jdgcs:
Is it possible for users to use aestratum as the mining pool which can communicate with mining client(e.g Gminer) in the next version?

uwiger:
> Is it possible for users to use aestratum as the mining pool which can communicate with mining client(e.g Gminer) in the next version?

It should be possible yes - as possible as it has been in previous versions.

The documentation does state that the application is in 'alpha' state, which is not least because we know of no users (even experimental), and have received no user feedback. However, there are test suites that are integrated into our CI. The last substantive change to the aestratum code was made in Oct 2019.

Please do try it out and let us know if you encounter any problems.

Issue #3282, opened 2020-06-09
AENS: Review and simplify pointers


hanssv:
- [ ] Check what is really enforced right now
- [ ] Adapt to something more strict (if necessary), restrict the number of pointers, the order(?), etc.
- [ ] Update tests to cover this

hanssv:
https://forum.aeternity.com/t/pointer-key-with-multiple-values/6024
https://github.com/aeternity/aeternity/pull/2684

Some references that can be useful - also see #core-ns in Rocket.chat

uwiger:
Done. Closing

PR #3447, opened 2021-01-07
GH-3282 AENS pointers cleanup

hanssv:
Limit AENS pointers from IRIS onward:
- no duplicate keys on chain
- not more than 32 pointers per name
- pointer key shouldn't be more than 256 bytes.

This is enforced in AENSUpdateTx. The main complication is when handling legacy (bad) pointers in Sophia, we take the approach of simply discarding bad pointers if/when encountered in Sophia.

Protocol: https://github.com/aeternity/protocol/pull/477
Sophia: https://github.com/aeternity/aesophia/pull/291

### DIFF
--- a/apps/aecontract/test/aecontract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_SUITE.erl
@@ -663,7 +663,6 @@ init_per_testcase(TC, Config) when TC == sophia_auth_tx;
 init_per_testcase(TC = bad_aens_pointer_handling_lima_to_iris, Config) ->
     case aect_test_utils:latest_protocol_version() of
         ?IRIS_PROTOCOL_VSN ->
-            LHeight = 20,
             IHeight = 25,
             Fun = fun(H) when H <  IHeight -> ?LIMA_PROTOCOL_VSN;
                      (H) when H >= IHeight -> ?IRIS_PROTOCOL_VSN
@@ -672,9 +671,7 @@ init_per_testcase(TC = bad_aens_pointer_handling_lima_to_iris, Config) ->
             meck:expect(aec_governance, name_claim_bid_timeout, fun(_, _) -> 0 end),
             Config1 =
                 [{sophia_version, ?SOPHIA_IRIS_FATE}, {vm_version, ?VM_FATE_SOPHIA_2},
-                 {fork_heights, #{ lima    => LHeight,
-                                   iris    => IHeight
-                                 }},
+                 {fork_heights, #{ iris    => IHeight }},
                  {protocol, iris} | Config],
             init_per_testcase_common(TC, Config1);
         _ ->
@@ -6293,7 +6290,6 @@ sophia_aens_update_transaction(_Cfg) ->
     ok.
 
 bad_aens_pointer_handling_lima_to_iris(Cfg) ->
-    _LimaHeight = maps:get(lima, ?config(fork_heights, Cfg)),
     IrisHeight = maps:get(iris, ?config(fork_heights, Cfg)),
 
     state(aect_test_utils:new_state()),
--- a/apps/aecontract/test/aecontract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_SUITE.erl
@@ -187,6 +187,7 @@
         , store_onetype_random_ops/1
         , fate_vm_cross_protocol_store_big/1
         , fate_vm_cross_protocol_store_multi_small/1
+        , bad_aens_pointer_handling_lima_to_iris/1
         ]).
 
 -include_lib("common_test/include/ct.hrl").
@@ -330,6 +331,7 @@ all() ->
                        , sophia_auth_tx
                        , sophia_strings
                        , {group, store_map_ops}
+                       , bad_aens_pointer_handling_lima_to_iris
                        ]).
 
 -define(FATE_TODO, [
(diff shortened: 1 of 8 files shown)


apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -614,6 +617,26 @@ init_per_testcase(TC, Config) when TC == sophia_auth_tx;
         true  -> init_per_testcase_common(TC, Config);
         false -> {skip, {requires_protocol, iris, TC}}
     end;
+init_per_testcase(TC = bad_aens_pointer_handling_lima_to_iris, Config) ->
+    case aect_test_utils:latest_protocol_version() of
+        ?IRIS_PROTOCOL_VSN ->
+            LHeight = 20,
+            IHeight = 25,
+            Fun = fun(H) when H <  IHeight -> ?LIMA_PROTOCOL_VSN;
```
uwiger, line 625: Perhaps it doesn't matter, but shouldn't it be `H >= LHeight, H < IHeight`?

apps/aeprimop/src/aeprimop.erl
```diff
@@ -953,12 +953,17 @@ name_update({OwnerPubkey, NameHash, TTL, ClientTTL, Pointers}, S) ->
...
                     true -> ClientTTL
                  end,
-    Pointers1  = if Pointers == undefined -> aens_names:pointers(Rec);
-                    is_list(Pointers) -> Pointers
-                 end,
     assert_ttl(TTL1, S1),
     assert_name_owner(Rec, OwnerPubkey),
     assert_name_claimed(Rec),
+    %% 'undefined' means we got here from FATE - thus we don't expect to
+    %% fail on legcy invalid pointers - so explicitly sanitize them.
```
uwiger, line 960:
```suggestion
    %% fail on legacy invalid pointers - so explicitly sanitize them.
```

uwiger (approved):

apps/aecontract/test/aecontract_SUITE.erl
hanssv: Yeah, the particular Lima height is not important.

skkw:
Would it be possible to enforce check on pointer keys (possibly values as well) so that they are printable?

Otherwise we would have to ensure that the results are encoded to handle the binary variants...

On testnet - we have this name_update_tx (note the binary pointer key):

```
{<<42, 175, 171, 92, 66, 180, 129, 211, 36, 154, 132, 2, 11, 119, 57, 236, 242,
   240, 164, 154, 146, 112, 119, 161, 89, 43, 88, 123, 204, 133, 156, 54>>,
 {:signed_tx,
  {:aetx, :name_update_tx, :aens_update_tx, 160,
   {:ns_update_tx,
    {:id, :account,
     <<96, 70, 239, 88, 30, 239, 116, 157, 73, 35, 96, 177, 84, 44, 123, 233,                                                                                                                     
       151, 181, 221, 202, 13, 46, 81, 10, 67, 18, 178, 23, 153, 139, 252,                                                                                                                        
       116>>}, 449,
    {:id, :name,
     <<158, 252, 133, 15, 146, 139, 214, 236, 224, 44, 148, 32, 117, 109, 239,                                                                                                                    
       240, 230, 47, 104, 248, 153, 101, 252, 222, 94, 208, 136, 247, 20, 222,                                                                                                                    
       174, 11>>}, 49999,
    [
      {:pointer,
       <<11, 180, 237, 121, 39, 249, 123, 81, 225, 188, 181, 225, 52, 13, 18,                                                                                                                     
         51, 91, 42, 43, 18, 200, 188, 82, 33, 214, 60, 75, 203, 57, 212, 30,                                                                                                                     
         97>>,
       {:id, :account,
        <<11, 180, 237, 121, 39, 249, 123, 81, 225, 188, 181, 225, 52, 13, 18,                                                                                                                    
          51, 91, 42, 43, 18, 200, 188, 82, 33, 214, 60, 75, 203, ...>>}}
    ], 50000, 18200000000000, 0}},
  [
    <<72, 146, 109, 105, 217, 248, 85, 240, 60, 19, 110, 153, 222, 189, 121, 88,                                                                                                                  
      119, 37, 5, 17, 61, 207, 65, 87, 53, 16, 6, 224, 153, 11, 119, 6, 184,                                                                                                                      
      132, 180, 245, 217, 207, 236, 6, 161, 234, 179, 168, ...>>
  ]}}
```

hanssv:
The values are either accounts, contracts, oracles or channels(!?) so they are inherently binary - but for them presenting the base58 encoded version is the obvious solution.

The keys are a different story... I kind of like the idea of having them printable - but is there a good (stable!) definition of printable?

skkw:
> The keys are a different story... I kind of like the idea of having them printable - but is there a good (stable!) definition of printable?

I guess this could do:

```
iex(aeternity@localhost)95> :io_lib.printable_unicode_list('Ńaïve')
true
iex(aeternity@localhost)96> :io_lib.printable_unicode_list(<<1,2,3,4>>)
false
```

hanssv:
Yeah, and hope that the `io_lib` (or the backing code) never change...

hanssv:
But I'll have a look and see if there is a sensible and stable implementation/definition. After all we do some String validation in (the upcoming version of) FATE.

hanssv:
Still it won't fix any historical pointer keys, so any code handling them should care for the non-printable case - maybe present them base64 encoded in that case?

velzevur (dismissed):

uwiger (approved):

cytadela8 (dismissed):

uwiger (approved):

gorbak25 (approved):

cytadela8 (approved):

velzevur (approved):

PR #3461, opened 2021-01-12
First stab at sync_allowed_height_from_top

uwiger:
WIP prototype of fork resistance.

Modeled after the check of `gossip_allowed_height_from_top`, which rejects block insertions that are at least a given distance from the top, the introduction of `sync:sync_allowed_height_from_top` (default: `0` for now, which means disabled), makes it possible to resist a node connecting with a longer fork.


uwiger:
See issue #3463

uwiger:
Functionality merged in PR #3478. Closing

PR #3462, opened 2021-01-12
Monitoring mode

gorbak25:
This PR is part of a joint effort between the core team and the SH team to give the community more transparency about the network.
Adds a special "monitoring" mode in which:

1. Peer connection limits are disregarded - the node tries to connect to as many peers as possible and as agressively as possible
2. The remote top hashes and difficulty are kept locally - might be useful for determining the source node of a new competing fork
3. We ask the peers once an hour about their version and vendor string
4. Keeps the data even when the peer disconnects - a "last pinged" timer is maintained - it's up to the frontend to filter out active nodes(last seen < 3-5 minutes)

No tests so far besides testing that it works in an erlang console:
image
image

### DIFF
--- a/apps/aecore/src/aec_peer_analytics.erl
+++ b/apps/aecore/src/aec_peer_analytics.erl
@@ -16,12 +16,15 @@
         , handle_call/3
         , handle_cast/2
         , handle_info/2
+        , terminate/2
+        , code_change/3
         ]).
 
 %% Peer stats
 -export([ log_peer_status/4
         , log_temporary_peer_status/4
         , get_stats/0
+        , get_stats_for_client/0
         , enabled/0
         ]).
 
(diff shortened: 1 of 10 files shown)


gorbak25:
image

gorbak25:
@mradkov @thepiwo 
Use this aeternity.yaml:
```
sync:
  peer_analytics: true
```

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -629,7 +630,19 @@ handle_request_('GetStatus', _Params, _Context) ->
        <<"peer_pubkey">>                => aeser_api_encoder:encode(peer_pubkey, PeerPubkey),
        <<"top_key_block_hash">>         => aeser_api_encoder:encode(key_block_hash, TopBlockHash),
        <<"top_block_height">>           => TopBlockHeight}};
-
+handle_request_('GetNetworkStatus', _Params, _Context) ->
+    case aec_peer_analytics:enabled() of
+        false ->
+            {400, [], #{reason => <<"Network analytics disabled in node config">>}};
```
dincho, line 636: This is wrong HTTP code:
```
400 Bad Request
The server cannot or will not process the request due to an apparent client error (e.g., malformed request syntax, size too large, invalid request message framing, or deceptive request routing).
```

This should return standard 404 IMO

dincho (commented):
I heavily disagree this endpoint being part of the external interface, it should be internal regardless that it's configuration enabled.

mradkov:
> I heavily disagree this endpoint being part of the external interface, it should be internal regardless that it's configuration enabled.

How else would you go about this?

gorbak25:
> > I heavily disagree this endpoint being part of the external interface, it should be internal regardless that it's configuration enabled.
> 
> How else would you go about this?

There are 2 kinds of endpoint - internal and external. Right now I've put it on the external endpoint - After I move it to the internal endpoint you can still access it.

dincho:
Please also update "epoch_full.yaml"

thepiwo:
should only be in the debug endpoints, right?

apps/aecore/src/aec_peer_analytics.erl
```diff
@@ -0,0 +1,141 @@
...
+             , last_seen => unix_time()
+             , genesis_hash => GHash
+             , top_hash => THash
+             , difficulty => Diff
+             },
+    St#{stats => Stats#{ Pub => PS }}.
+
+unix_time() ->
+    {MegaSecs, Secs, _MicroSecs} = now(),
+    MegaSecs * 1000000 + Secs.
```
uwiger, line 114: `erlang:system_time(second)` will give the same result, only more straightforwardly.
(Except `now()` is defined to always increase monotonically. If this is desired,
`erlang:monotonic_time(second) + erlang:time_offset(second)` should do essentially the same.)

apps/aecore/src/aec_peers.erl
```diff
@@ -787,10 +787,15 @@ next_connect_delay(State) ->
     case is_outbound_allowed(State) of
         false  -> infinity;
         true ->
-            #state{ last_connect_time = LastTime, outbound = Outbound } = State,
-            ExpDelay = floor(math:pow(2, Outbound - 1)) * 1000,
-            BoundDelay = min(ExpDelay, ?MAX_CONNECTION_INTERVAL),
-            max(?MIN_CONNECTION_INTERVAL, BoundDelay - (timestamp() - LastTime))
+            case aeu_env:user_config([<<"sync">>, <<"peer_analytics">>], false) of
```
uwiger, line 790: Whenever there is a schema default (and there should be), I think it should be honored.
```
{ok, Bool} = aeu_env:find_config([<<"sync">>, <<"peer_analytics">>],
                                 [user_config, schema_default, {value,false}]),
case Bool of
...
```

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -629,7 +630,19 @@ handle_request_('GetStatus', _Params, _Context) ->
...
+        true ->
+            Stats0 = maps:to_list(aec_peer_analytics:get_stats()),
+            Stats1 = lists:map(fun({K, V}) ->
+                { aeser_api_encoder:encode(peer_pubkey, K)
+                , encode_peer_details(V)} end,
+                Stats0),
```
uwiger, line 642: You could use a maps:fold here:
```
Stats1 = maps:fold(fun(K, V, Acc) ->
                        Acc#{ aeser_api_encoder:encode(peer_pubkey, K) => encode_peer_details(V) }
                   end, #{}, aec_peer_analytics:get_stats())
```

gorbak25:
Available on the internal endpoint at:
http://localhost:3113/v2/status/network

thepiwo:
@gorbak25 all other internal endpoints are under v2/debug, so the proxy configuration we are having won't work for an internal non-debug route

uwiger (approved):

apps/aehttp/priv/swagger.yaml
```diff
@@ -947,7 +947,25 @@ paths:
           description: 'Successful operation'
           schema:
             $ref: '#/definitions/Status'
-
+  /debug/network:
+    get:
+      tags:
+        - internal
+        - node_info
```
dincho, line 954: missing debug tag ?

dincho (approved):

radrow (approved):

dincho (approved):

Issue #3463, opened 2021-01-13
Configurable resistance to forks injected via sync


uwiger:
The discussion below outlines how to address resistance to long forks being injected via the sync protocol (note, long forks are already rejected when injected via gossip).

The suggested approach is to let the node have a "confirmation time", much in the same way as when users confirm their own transactions: at a configurable depth, the chain becomes explicitly unwilling to fork. The nature of POW mining requires the ability to handle short forks resulting from natural race conditions. However, these should not go deeper than 1 key block.

## Introduction
In a proof-of-work (POW) blockchain system, it is possible to attack a network with brute force ("51% attack"): if you control enough mining resources to out-mine the rest of the network, you can essentially go back and rewrite history.

This can be realised in the following type of attack (by attacker 'A'):
1. `A` buys value on the chain. This will contribute to driving up the value of the chain token, at least on a fairly small chain.
2. `A` prepares a mining rig, syncs up with the chain, perhaps mines a few blocks on the main chain as a test run.
3. `A` then starts e.g. buying some other coin that is hard to trace. The exchanges may require a verification period of 500 blocks (or even more), but will eventually approve the transactions.
4. Meanwhile, `A` disconnects the mining rig from the main network, and secretly builds a block history where the transactions in (3) are not included.
5. Once the alternative history has a significantly greater height than the main chain, `A` reconnects the mining rig, and the sync protocol will instruct the main chain nodes to accept the longer fork.
6. The transactions in (3) are now no longer on the chain, and `A` was able to purchase external coin _and_ get all the money back.

At one level, this is perfectly in line with the protocol, and expectations of POW systems. Still, we would like to have some automatic protections, at least making it less profitable to do this.

## One-time fix: Whitelist
In the last attack, we were able to spot the buildup (steps 2 and 3 above), causing us to suspect that (4) was underway. We made a secret emergency patch where our block insertion logic checked a 'whitelist', which authoritatively specified which block hashes were expected at given heights. This seems to have worked since (a) the miners and exchanges accepted our solution, and authoritative view of the world, and (b), we were able to start fielding this (minutes) before the attack (step 5) was initiated.

An obvious problem with this is that it's a one-off, and the manually injected whitelist can only protect against attacks that have already gone through steps 1 - 3 above; it will be ineffective against future attacks.

### Automatically reject long forks
[...]

PR #3478, opened 2021-01-20
Uw fork resilience 2

uwiger:
Updated version of PR #3461, See issue #3463
This PR currently fails the forking_SUITE (thus, separate PR). Debugging ongoing

### DIFF
--- a/apps/aecore/src/aec_resilience.erl
+++ b/apps/aecore/src/aec_resilience.erl
@@ -30,12 +30,13 @@ fork_resistance_active() ->
     Res.
 
 start_link() ->
+    ensure_tab(),   % ets table will survive process restart
     gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
 
 init([]) ->
-    ensure_tab(),   % ets table will survive process restart
     aec_events:subscribe(chain_sync),
     aec_events:subscribe(update_config),
+    maybe_enable_fork_resistance_at_startup(),
     {ok, #st{}}.
 
 handle_call(_Req, _From, St) ->
@@ -71,6 +72,20 @@ terminate(_Reason, _St) ->
 
 code_change(_FromVsn, St, _Extra) ->
     {ok, St}.
+
+maybe_enable_fork_resistance_at_startup() ->
+    case lookup(fork_resistance, undefined) of
+        undefined ->
+            case aeu_env:find_config([<<"sync">>, <<"resist_forks_from_start">>],
+                                     [ user_config, schema_default ]) of
+                {ok, true} ->
+                    maybe_enable_fork_resistance();
+                _ ->
+                    ok
+            end;
+        _ ->
+            ok
+    end.
 
 maybe_enable_fork_resistance() ->
     case get_fork_resistance() of
--- a/apps/aecore/src/aec_resilience.erl
+++ b/apps/aecore/src/aec_resilience.erl
@@ -1,0 +1,121 @@
+-module(aec_resilience).
+-behaviour(gen_server).
+
+%%% This module implements
+
+-export([ start_link/0 ]).
+
+-export([ fork_resistance_active/0
+        ]).
+
+-export([ init/1
+        , handle_call/3
+        , handle_cast/2
+        , handle_info/2
+        , terminate/2
+        , code_change/3
+        ]).
+
+-record(st, {}).
+
+-define(TAB, ?MODULE).
+
+-type height() :: non_neg_integer().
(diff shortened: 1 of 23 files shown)


apps/aecore/src/aec_resilience.erl
```diff
@@ -0,0 +1,121 @@
...
+        [{_, Value}] ->
+            Value;
+        [] ->
+            Default
+    end.
+
+ensure_tab() ->
+    case ets:info(?TAB, name) of
+        undefined ->
+            ets:new(?TAB, [set, public, named_table]);
```
gorbak25, line 118: Isn't the table cleaned up when aec_resilience crashes?

apps/aecore/src/aec_resilience.erl
```diff
@@ -0,0 +1,121 @@
...
+    {ok, #st{}}.
+
+handle_call(_Req, _From, St) ->
+    {reply, {error, unknown_call}, St}.
+
+handle_cast(_Msg, St) ->
+    {noreply, St}.
+
+handle_info({gproc_ps_event, chain_sync, #{info := {chain_sync_done, _}}}, St) ->
+    %% For now, at least, we try to enable fork resistance as soon as we are in sync
```
gorbak25, line 48: Well if the node restarts then it can sync with a malicious chain

apps/aecore/test/aecore_forking_SUITE.erl
```diff
@@ -131,7 +143,89 @@ sync_fork_in_wrong_order(Config) ->
...
+    T1 = os:timestamp(),
+    aecore_suite_utils:start_node(dev3, Config),
+    aecore_suite_utils:connect(N3),
+    aecore_suite_utils:await_sync_complete(T1, [N3]),
+    Top = rpc:call(N3, aec_chain, top_block, [], 5000),
+    %% Nodes are all in sync
+    ok = stop_and_check([dev1, dev2, dev3], Config),
+    ok.
+
+dev3_failed_attack(Config) ->
```
gorbak25, line 168: Looks OK if the fork resistance is in manual mode - in automatic mode it will also work unless a node restarts:
1) dev1, dev2 mine 120 blocks
2) dev3 mines silently 140 more difficult blocks
3) dev2 stops
4) dev3 comes online
5) dev2 starts
Should we do anything in this case?

gorbak25 (changes requested):

apps/aecore/src/aec_resilience.erl
uwiger: Hmpf! The call to `ensure_tab()` was supposed to be made from `start_link()`, not `init/1`.

apps/aecore/src/aec_resilience.erl
uwiger: Yes, but this will need to be prevented using whitelisting

apps/aecore/test/aecore_forking_SUITE.erl
uwiger: One could imagine adding some config setting that enables fork resistance from the start, _or_ we rely on whitelisting in that case.

apps/aecore/src/aec_resilience.erl
gorbak25: Not always - by persisting the time when we were synced it's possible to automatically enforce fork resistance:
If the fork resistance value is 100 and we were offline 30 minutes after restarting(and being synced before) we can still safely set the fork resistance value to 90.

gorbak25: Of course when being offline for a long period we need to disable fork resistance. But in most cases crashes/downtimes are small.

gorbak25: @uwiger Another possibility:
Instead of keeping the date when we were last "fully synced" we can keep a finalized hash which once enabled by fork resistance must be respected. IMO the whitelist is a countermeasure for when the node is fresh and needs to sync from genesis. After we synced at least once we might be sure that the block at TopHeight-100 is finalized - this should be respected after restarts.

apps/aecore/test/aecore_forking_SUITE.erl
gorbak25: IMO the whitelist is useful when syncing from genesis - by knowing that we were synced at least once at some point in time we have a gurantee that Top-100 is finalized so we should reject forks starting before Top-100 and reenable resistance once we synced fully.

gorbak25 (approved):

velzevur (approved):

PR #3480, opened 2021-01-21
Minimal change to provoke 'wrong_target' error in forking suite

uwiger:
This PR is for debugging only, revealing an error in target calculation in the forking suite.
The error manifests when the height exceeds the `BLOCKS_TO_CHECK_DIFFICULTY_COUNT` (default: 17) enough to cause divergence.


gorbak25:
:(

gorbak25 (approved):

velzevur (approved):

PR #3491, opened 2021-02-01
Fix swagger versioning

velzevur:
It looks like the `sed` command was broken, resulting in `VERSION` file being `v5.7.0` instead of expected `5.7.0` which broke the semver.

The work on this PR is supported by Aeternity Crypto Foundation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -244,7 +244,7 @@ REVISION:
 	@git rev-parse HEAD > $@ || echo "unknown" > $@
 
 VERSION:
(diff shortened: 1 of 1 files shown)


hanssv (approved):

gorbak25 (dismissed):
Shall I make and deploy 5.7.1 ?

uwiger (approved):

gorbak25 (approved):

uwiger (approved):

Issue #3496, opened 2021-02-02
Abnormal  env consumption during database access 


erlmachinedev:
The point is the application:get_env/2 call executes through the shared internal gen_server and maybe has to be used with care.

The function `aec_db:try_activity/3 ` accesses `aeu_env:user_config_or_env([<<"chain">>, <<"db_write_max_retries">>],
                                            aecore, db_write_max_retries, 1)` abnormally often, which leads to a lack of productivity.

It queries it twice, as usually hits the second branch of `user_config_or_env\4`.

The first idea was to implement access via persistent term, but looks like the recent fix of @gorbak25 allows us to remove this at all.

The branch which solves the issue (supplied with measurements) will be attached here to be reviewed by the core team.

uwiger:
Not that I disagree that there is no need to call a particular `get_env` more than once, but as a general point, the calls do _not_ go through any central process.

First, the execution path of `application:get_env/2` is:
```erlang
get_env(Application, Key) -> 
    application_controller:get_env(Application, Key).
```
which leads to the following code in `application_controller.erl`:
```erlang
get_env(AppName, Key) ->
    case ets:lookup(ac_tab, {env, AppName, Key}) of
        [{_, Val}] -> {ok, Val};
        _ -> undefined
    end.
```

(`ac_tab` is also created with `{read_concurrency, true}` to be particularly fast to read by concurrent processes.)

Now, the `aeu_env` access functions are mainly designed for flexibility, and not for speed, so they go through `setup:get_env/3`, which will lookup expansion metadata for the application (through more `get_env` lookups), then do a deep traversal of the found term to see if there are expansion variables that need to be substituted.

Of course, this is all a feature, not a bug. ;-)
But it was designed to be accessed once, typically in `setup` start phase `110`, and if needed, optimized for performance.

uwiger:
> Now, the `aeu_env` access functions are mainly designed for flexibility, and not for speed, so they go through `setup:get_env/3`, which will lookup expansion metadata for the application (through more `get_env` lookups), then do a deep traversal of the found term to see if there are expansion variables that need to be substituted.
> 
> Of course, this is all a feature, not a bug. ;-)
> But it was designed to be accessed once, typically in `setup` start phase `110`, and if needed, optimized for performance.

[...]

PR #3504, opened 2021-02-09
Issue #3496

erlmachinedev:
This PR removes configuration access call due to the DB transaction processing. 
he discussion and attached measurements are placed on [3496](https://github.com/aeternity/aeternity/issues/3496). 

The configuration schema is marked as "deprecated" to be removed at the next release.


radrow:
Why are all of these tests failing?

radrow:
`{case_clause,ok}` – sounds like something undesirable :p

PR #3506, opened 2021-02-09
Hyperchains consensus

gorbak25:
- [x] Integrate the staking contract with the ACI library and the node build system
- [x] Keep track of where the staking contract is deployed
- [x] If Hyperchains were requested at genesis then deploy the staking contract in the HC genesis block
- [x] When switching the consensus to HC and no predeployed staking contract is present then deploy the staking contract upon changing the consensus
- [x] When using a predeployed staking contract then validate the bytecode of the onchain staking contract upon consensus change
- [x] When starting the node validate our local bytecode against the persisted staking contract
- [ ] Decrease the finalized height delta to 4-6 when HC got activated
- [x] Turn on Hyperchain consensus only when the staking contract accumulates enough economic security, fallback to PoW otherwise
- [x] Embed "Delegate" and "Parent hash" in the PoW field of the header
- [ ] Integrate with Parent chain manager - block in "generate" key block seal in order to wait for a new parent block to get mined
- [x] Static staking contract calls - introspection etc...
- [x] Restricted protocol level staking contract calls
- [ ] HC protocol CT suite - 2 nodes one of which runs a chain simulator and the second one is a hyperchain


zjn6868:
@ #

- [ 

- ]

erlmachinedev:
If you don’t mind I suggest to merge this branch as soon. Even if some tests haven’t passed yet (the reason maybe outside of the scope of this functionality).

erlmachinedev (approved):

cytadela8:
@erlmachinedev In terms of tests, everything,but some non-trivial dialyzer issues is resolved.

cytadela8 (approved):

PR #3512, opened 2021-02-16
Introduce the infrastructure for OAS3

velzevur:
This PR provides the means to have both old swagger v2 with the new OAS v3
specifications. The new specifications go to `oas3.yaml` while we keep the old
`swagger.yaml` untouched. Those two can share `OperationId`-s but those must
have the same expectations for params and response schemas. The idea is to
keep the old specifications as long as we need them while preparing the new
API. This would provide the time needed for SDKs and client apps to catch up.

### DIFF
--- a/apps/aecore/test/aecore_suite_utils.erl
+++ b/apps/aecore/test/aecore_suite_utils.erl
@@ -1267,23 +1267,21 @@ use_swagger(SpecVsn) ->
         end,
     put(api_prefix, Prefix).
 
+get(Key, Default) ->
+    case get(Key) of
+        undefined -> Default;
+        V -> V
+    end.
+
 http_request(Host, get, Path, Params) ->
-    Prefix =
-        case get(api_prefix) of
-            undefined -> "/v2/";
-            V -> V
-        end,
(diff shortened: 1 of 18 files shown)


apps/aecore/test/aecore_suite_utils.erl
```diff
@@ -1258,14 +1259,32 @@ await_new_jobs_pid_recurse(N, OldP, TRef) ->
...
     ct:log("GET ~p", [URL]),
     R = httpc_request(get, {URL, []}, [], []),
     process_http_return(R);
 http_request(Host, post, Path, Params) ->
-    URL = binary_to_list(iolist_to_binary([Host, "/v2/", Path])),
+    Prefix =
+        case get(api_prefix) of
+            undefined -> "/v2/";
+            V -> V
+        end,
```
uwiger, line 1286: An opportunity for a helper function, perhaps? (c.f. lines 1271-1275)

uwiger (approved):

uwiger (approved):

gorbak25 (approved):

Issue #3501, opened 2021-02-05
Random CI failure in aecore_txs_SUITE:missing_tx_gossip


gorbak25:
```
%%% aecore_txs_SUITE ==> missing_tx_gossip: FAILED
%%% aecore_txs_SUITE ==> {max_reached,
    [{aecore_suite_utils,mine_safe_setup,4,
         [{file,
              "/home/builder/aeternity/apps/aecore/test/aecore_suite_utils.erl"},
          {line,1391}]},
     {aecore_txs_SUITE,missing_tx_gossip,1,
         [{file,
              "/home/builder/aeternity/apps/aecore/test/aecore_txs_SUITE.erl"},
          {line,233}]},
     {test_server,ts_tc,3,[{file,"test_server.erl"},{line,1754}]},
     {test_server,run_test_case_eval1,6,
         [{file,"test_server.erl"},{line,1263}]},
     {test_server,run_test_case_eval,9,
         [{file,"test_server.erl"},{line,1195}]}]}
```

PR #3514, opened 2021-02-17
GH 3501:  aecore txs failures

velzevur:
Improves #3501

### DIFF
--- a/apps/aecore/test/aecore_suite_utils.erl
+++ b/apps/aecore/test/aecore_suite_utils.erl
@@ -414,9 +414,9 @@ assert_not_already_on_chain(Node, TxHashes) ->
             error({already_on_chain, AlreadyOnChain})
     end.
 
-tx_included_or_can_be_included(Node, TxHash) ->
+is_tx_onchain(Node, TxHash) ->
     case rpc:call(Node, aec_chain, find_tx_location, [TxHash]) of
-        mempool -> true;
+        mempool -> false;
         X when is_binary(X) ->
             ct:log("Transaction already included ~p",
                   [TxHash]),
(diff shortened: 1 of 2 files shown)


gorbak25 (approved):

velzevur:
Tests passed, restarting them to make sure the issue is gone

gorbak25:
@velzevur Could you restart them with instant mining? We could shave off a few seconds here ;)

gorbak25 (approved):

gorbak25:
```
%%% aecore_txs_SUITE: Makefile:228: recipe for target 'ct-minerva' failed
make: *** [ct-minerva] Terminated
Makefile:431: recipe for target 'internal-ct' failed
make[1]: *** [internal-ct] Hangup

Too long with no output (exceeded 10m0s): context deadline exceeded
``` 
:thinking:

radrow (approved):

velzevur:
It feels like I broke it even more now...

gorbak25:
@velzevur try enabling deterministic mining and then find the hidden assumption in the test.
When I was fighting txs_gc the assumptions were:
1. You receive at least one block with a beneficiary different than us(otherwise GC won't kick in)
2. One specific generation needs at least 2 microblocks -.-

velzevur:
So tests fail with
```
*** User 2021-02-17 15:37:37.844 ***
Transaction not received yet <<"th_2snpKf7ic5XGYygogQV6CXjmMscRawcEQgpCquNBbrM9Ukyd7o">>
```
where this is the hash of the last transaction. It was posted to `dev2` but it is unknown to it later on. _Investigating further more_

velzevur:
I declare a partial success: it seems like the intensity of failing this test had dramatically decreased. I might have fixed _a few_ bugs there. Most prominent would be that the tx pool accepts txs which nonces are `Nonce =< AccountNonce + Offset`. This is not the case with new accounts, where the tx pool rejects all txs with nonce greater than `nonce_baseline` setting (1 by default). Hence providing a tx with a `nonce=5` on `dev2` can only succeed if it has seen a tx with a `nonce=1`. This is now fixed with bumping the setting `nonce_baseline`.

The tests still fail, but less often. I will investigate further more.

uwiger (approved):

uwiger:
I'll take a partial improvement for now. :)

gorbak25 (approved):

PR #3542, opened 2021-03-09
Add generated files to gitignore

cytadela8:
I wonder why nobody before me was annoyed enough to make this change.


cytadela8:
Seems I had some legacy stuff. Sorry :/

PR #3543, opened 2021-03-10
AESC: new tx - set delegates

### LINKED ISSUE #3288: AESC: new transaction type - set delegates
Currently delegates can only be set at channel creation time which makes no sense for long running channels.

velzevur:
Fixes #3288 

FSM and WebSocket support are not part of the on-chain protocol and not included here.

### DIFF
--- a/apps/aechannel/src/aesc_set_delegates_tx.erl
+++ b/apps/aechannel/src/aesc_set_delegates_tx.erl
@@ -50,6 +50,8 @@
           responder_delegate_ids :: [aeser_id:id()],
           from_id                :: aeser_id:id(),
           payload                :: binary(),
+          state_hash             :: binary(),
+          round                  :: non_neg_integer(),
           ttl                    :: aetx:tx_ttl(),
           fee                    :: non_neg_integer(),
           nonce                  :: non_neg_integer()
@@ -69,6 +71,8 @@ new(#{channel_id             := ChannelId,
       initiator_delegate_ids := IDelegates,
       responder_delegate_ids := RDelegates,
       payload                := Payload,
+      state_hash             := StateHash,
+      round                  := Round,
       fee                    := Fee,
       nonce                  := Nonce} = Args) ->
     channel = aeser_id:specialize_type(ChannelId),
@@ -81,6 +85,8 @@ new(#{channel_id             := ChannelId,
             initiator_delegate_ids = IDelegates,
             responder_delegate_ids = RDelegates,
             payload                = Payload,
+            state_hash             = StateHash,
+            round                  = Round,
             ttl                    = maps:get(ttl, Args, 0),
             fee                    = Fee,
             nonce                  = Nonce},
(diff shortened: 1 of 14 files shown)


velzevur:
As a comment why do we have both `payload` and `state_hash` + `round`: this is to fix the following situation:
1. Alice and Bob create a channel with no delegates
1. Alice and Bob co-sign on-chain tx, let's say a deposit, post it on-chain
2. Alice and Bob co-sign a `channel_set_delegates_tx` authorising Dave to provide txs for Alice, post it on-chain
3. Alice decides she doesn't want Dave to be able to represent her, they sign a new `channel_set_delegates_tx` where Dave is not present
4. Bob posts a non-closing on-chain tx (deposit, snapshot and etc) where he provides the latest on-chain state
5. Now depending on Bob's authentication method, he could replay the tx in (2) and thus return Dave in the list of delegates

This works in a similar way replaying all txs and thus adding/removing delegates. It heavily depends on the ability of a tx authentication to be replayed....

apps/aechannel/src/aesc_set_delegates_tx.erl
```diff
@@ -0,0 +1,276 @@
...
+    , {responder_delegate_ids   , [id]}
+    , {payload                  , binary}
+    , {state_hash               , binary}
+    , {round                    , int}
+    , {ttl                      , int}
+    , {fee                      , int}
+    , {nonce                    , int}
+    ].
+
+-spec version(tx()) -> non_neg_integer().
```
uwiger, line 268: If this function were moved to the top of the "behaviour API" section, it would be closer to the macro definition.

apps/aehttp/priv/oas3.yaml
```diff
@@ -1544,6 +1544,35 @@ paths:
...
               schema:
                 $ref: "#/components/schemas/Error"
+  /debug/channels/set-delegates:
+    post:
+      tags:
+        - internal
+        - channel
+        - debug
+      operationId: PostChannelSetDelegates
+      description: 'Get a channel_set_delegates_tx transaction object'
```
uwiger, line 1554: "Get"?

uwiger (approved):

apps/aechannel/src/aesc_set_delegates_tx.erl
```diff
@@ -0,0 +1,276 @@
...
+         state_hash/1,
+         round/1]).
+
+-include_lib("aecontract/include/hard_forks.hrl").
+%%%===================================================================
+%%% Types
+%%%===================================================================
+
+-define(INITIAL_VSN, 1).
+-define(TX_TYPE, channel_set_delegates_tx).
```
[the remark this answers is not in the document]
velzevur: This is mostly of compile-time safety so we don't make a typo with the atom

apps/aehttp/priv/oas3.yaml
velzevur: This is a big topic but: since the goal of this API is to fetch a channel representation - _get_ seems appropriate. Because HTTP GET having a request body is considered a bad practice we use POST instead.

gorbak25:
> As a comment why do we have both `payload` and `state_hash` + `round`: this is to fix the following situation:
> 
>     1. Alice and Bob create a channel with no delegates
> 
>     2. Alice and Bob co-sign on-chain tx, let's say a deposit, post it on-chain
> 
>     3. Alice and Bob co-sign a `channel_set_delegates_tx` authorising Dave to provide txs for Alice, post it on-chain
> 
>     4. Alice decides she doesn't want Dave to be able to represent her, they sign a new `channel_set_delegates_tx` where Dave is not present
> 
>     5. Bob posts a non-closing on-chain tx (deposit, snapshot and etc) where he provides the latest on-chain state
> 
>     6. Now depending on Bob's authentication method, he could replay the tx in (2) and thus return Dave in the list of delegates
> 
> 
> This works in a similar way replaying all txs and thus adding/removing delegates. It heavily depends on the ability of a tx authentication to be replayed....

If you can replay a transaction I think you have bigger problems...

gorbak25 (approved):

PR #3544, opened 2021-03-10
CREATE, CLONE and BYTECODE_HASH

radrow:
Related: https://github.com/aeternity/aebytecode/pull/94

This PR adds support for `CREATE`, `CLONE` and `BYTECODE_HASH` instructions. Unformal semantics:

- `CREATE` deploys a contract of a given **serialized code**. It executes its `init` function with arguments from the stack. Returns `contract`
- `CLONE` works just like `CREATE` but takes an existing contract instead. Can be protected.
- `CLONE_G` works just like `CLONE` but takes an additional parameter for gas limitation.
- `BYTECODE_HASH` calculates a hash of the **bytecode** of a given contract. Wraps with `Some` or returns `None` on fail.

Scope:

- [x] `CREATE` handling
- [x] `CREATE` test
- [x] `BYTECODE_HASH` handling
- [x] `BYTECODE_HASH` test
- [x] `CLONE` handling
- [x] `CLONE` test
- [x] `CLONE_G` handling
- [x] `CLONE_G` test
- [x] <s>Ban calldata with `fate_code`</s> – discussed not to be implemented


apps/aefate/src/aefa_fate_op.erl
```diff
@@ -1239,6 +1242,37 @@ spend(Arg0, Arg1, ES0) ->
...
+                    CodeCache1 = maps:put(Pubkey, ContractCode, CodeCache),
+                    ES2 = aefa_engine_state:set_code_cache(CodeCache1, ES1),
+                    ES3 = aefa_engine_state:set_chain_api(APIState1, ES2),
+                    Hashed = aeb_fate_data:make_hash(aec_hash:hash(fate_bytecode, ContractCode)),
+                    write(Arg0, Hashed)
+                error ->
+                    abort({trying_to_get_contract_hash, Pubkey}, ES1)
+            end;
+        ContractCode ->
+            Hashed = aeb_fate_data:make_hash(aec_hash:hash(fate_bytecode, ContractCode)),
```
gorbak25, line 1272: It this a hash over the code or is this the hash over bytecode? Also is this the hash over the serialized fcode or is it over the deserialized version of the code?

radrow: Bytecode, I believe serialized. Must check tho

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -1239,6 +1242,37 @@ spend(Arg0, Arg1, ES0) ->
...
+                {ok, ContractCode, APIState1} ->
+                    CodeCache1 = maps:put(Pubkey, ContractCode, CodeCache),
+                    ES2 = aefa_engine_state:set_code_cache(CodeCache1, ES1),
+                    ES3 = aefa_engine_state:set_chain_api(APIState1, ES2),
+                    Hashed = aeb_fate_data:make_hash(aec_hash:hash(fate_bytecode, ContractCode)),
+                    write(Arg0, Hashed)
+                error ->
+                    abort({trying_to_get_contract_hash, Pubkey}, ES1)
+            end;
+        ContractCode ->
```
gorbak25, line 1271: If https://github.com/aeternity/aeternity/pull/3546/files gets merged would that be the code with or without the INIT function?

radrow: As long as init isn't stripped from the bytecode (in the `fcode` field) it shall be included

cytadela8: I believe I decided to strip `init` before putting a contract to cache so this need revisiting

apps/aefate/src/aefa_fate.erl
```diff
@@ -249,6 +249,7 @@ abort(remote_type_mismatch, ES) ->
 abort({function_arity_mismatch, Got, Expected}, ES) ->
     ?t("Expected ~p arguments, got ~p", [Expected, Got], ES);
 abort({value_does_not_match_type, Val, Type}, ES) ->
+    error({jebudu, Val, Type}),
```
cytadela8: Is this supposed to be here? :P

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -1239,6 +1248,89 @@ spend(Arg0, Arg1, ES0) ->
...
+
+create_contract(Code, Amount, ES0) ->
+    Current = aefa_engine_state:current_contract(ES0),
+
+    AS0              = aefa_engine_state:chain_api(ES0),
+    {ok, Nonce, AS1} = aefa_chain_api:next_nonce(Current, AS0),
+
+    Contract0  = aect_contracts:new(
+                     Current, Nonce,
+                     #{vm => aefa_engine_state:vm_version(ES0)
```
cytadela8, line 1289: Make sure to allow only FATE2

apps/aefate/src/aefa_fate.erl
radrow: uhm, no

radrow:
**TODO** BAN REMOTE CALLS WITH CODE

apps/aefate/src/aefa_fate_op.erl
radrow: This is a helper function, the CREATE handler makes this check

apps/aefate/src/aefa_chain_api.erl
```diff
@@ -242,6 +253,38 @@ account_balance(Pubkey, #state{primop_state = PState,
...
+-spec next_nonce(pubkey(), state()) -> 'error' |
+                                       {'ok', integer(), state()}.
+next_nonce(Pubkey, #state{primop_state = PState0,
+                          onchain_primop_state = Onchain} = S) ->
+    case aeprimop_state:find_account(Pubkey, PState0) of
+        {Account, PState1} ->
+            Nonce = aec_accounts:nonce(Account) + 1,
+            case eval_primops(
+                   [aeprimop:force_inc_account_nonce_op(Pubkey, Nonce)],
+                   S#state{onchain_primop_state = {onchain, PState1}}) of
```
cytadela8, line 265: You are putting `primop_state` to `onchcain_primeop_state` is this correct?

rebar.lock
```diff
@@ -1,190 +0,0 @@
-{"1.2.0",
```
cytadela8, line 1: Fix (readd) `rebar.lock`

cytadela8 (commented):
`do-not-strip-init` needs review to make sure compatibility with your changes.

apps/aefate/src/aefa_fate_op.erl
cytadela8: OK!

apps/aefate/src/aefa_chain_api.erl
radrow: uh, good catch

rebar.lock
radrow: Done

cytadela8 (commented):
Good!

apps/aefate/src/aefa_fate_op.erl
radrow: @cytadela8 could you confirm?

apps/aefate/src/aefa_chain_api.erl
```diff
@@ -208,19 +221,41 @@ contract_fate_code(Pubkey, #state{primop_state = PState} = S) ->
...
+                    {ok, Pubkey};
+                {ref, Ref} ->
+                    RefContractPK = aeser_id:specialize(Ref, contract),
+                    contract_find_final_ref(RefContractPK, S0)
+            end
+    end.
+
+-spec remove_contract(pubkey(), state()) -> state().
+remove_contract(Pubkey, #state{primop_state = PState0} = S0) ->
+    PState1 = aeprimop_state:delete_contract(Pubkey, PState0),
```
cytadela8, line 256: What about the account?

apps/aefate/src/aefa_chain_api.erl
```diff
@@ -250,6 +285,38 @@ account_balance(Pubkey, #state{primop_state = PState,
...
+            case eval_primops(
+                   [aeprimop:force_inc_account_nonce_op(Pubkey, Nonce)],
+                   S#state{primop_state = PState1}) of
+                {ok, S1} -> {ok, Nonce, S1};
+                Err -> Err
+            end;
+        none when Onchain =:= none ->
+            error;
+        none ->
+            %% Try to find this onchain as well.
```
cytadela8, line 304: I have to say I'm quite unsure if we should look onchain. I though that state channels contracts can't interact in statefull way with onchain

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -201,7 +205,7 @@ return(EngineState) ->
     aefa_fate:pop_call_stack(EngineState).
 
 returnr(Arg0, EngineState) ->
-    ES1 = un_op(get, {{stack, 0}, Arg0}, EngineState),
+    ES1 = push(Arg0, EngineState),
```
cytadela8, line 208: What this change does?

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -1247,6 +1261,134 @@ spend(Arg0, Arg1, ES0) ->
...
+    ES1 = aefa_engine_state:set_chain_api(AS3, ES0),
+    {ContractPK, ES1}.
+
+bytecode_hash(Arg0, Arg1, ES0) ->
+    ?AVAILABLE_FROM(?VM_FATE_SOPHIA_2, ES0),
+    {?FATE_CONTRACT(Pubkey), ES1} = get_op_arg(Arg1, ES0),
+    case aefa_engine_state:contract_fate_bytecode(Pubkey, ES1) of
+        {ok, ByteCode, ?VM_FATE_SOPHIA_2, ES2} ->
+            SerByteCode = aeb_fate_code:serialize(ByteCode),
+            Hashed  = aeb_fate_data:make_hash(aec_hash:hash(fate_code, SerByteCode)),
```
cytadela8, line 1386: Minor concern. Should the (gas) price not depend somehow on the hashed contract bytecode size? What if somebody makes a big contract and keeps asking for its bytecode hash...

cytadela8: Keeping a local cache at the node would also mitigate this

rebar.config
```diff
@@ -67,7 +67,7 @@
         {aeminer, {git, "https://github.com/aeternity/aeminer.git",
                    {ref, "0a82f0f"}}},
 
-        {aebytecode, {git, "https://github.com/aeternity/aebytecode.git", {ref, "135a27c"}}},
+        {aebytecode, {git, "https://github.com/aeternity/aebytecode.git", {branch, "factories"}}},
```
cytadela8, line 70: but proper hash when available or if you really want use 'master'

radrow: I will fix it when aebytecode gets merged

apps/aefate/src/aefa_fate_op.erl
radrow: Nothing, just more readable syntax

apps/aefate/src/aefa_chain_api.erl
radrow: But this is not a stateful interaction from the perspective of chain, is it?

apps/aefate/src/aefa_chain_api.erl
radrow: Good catch! Fixed

apps/aefate/src/aefa_fate_op.erl
radrow: Caching it would solve the problem I guess.

apps/aefate/src/aefa_chain_api.erl
```diff
@@ -194,25 +198,65 @@ traverse_to_key_hash(H, KeyHash) ->
         _Height -> traverse_to_key_hash(H, aec_headers:prev_key_hash(Header))
     end.
 
--spec contract_fate_code(pubkey(), state()) -> 'error' |
-                                               {'ok', {term(), aect_contracts:vm_version()}, state()}.
-contract_fate_code(Pubkey, #state{primop_state = PState} = S) ->
+-spec put_contract(aect_contracts:contract(), state()) -> state().
+put_contract(Contract, #state{primop_state = PS0} = S) ->
```
gorbak25, line 202: You know it's possible to predict a contract address without deploying it and send money to that account?
1. Foo pre-calculates the address of Bar - X 
2. Foo sends funds to address X
3. Foo deploys Bar at address X(it's enough to predict the nonces)

Your code would nuke the account balance.
BTW. @hanssv @UlfNorell How does the node handle this case generally in create_tx? Should the tokens be lost or still present after deploy?

gorbak25 (changes requested):
Don't merge until:
1. #3551 gets merged to master
2. #3554 gets rebased on master
3. We discuss what to do when funds are already present at the deployer account

apps/aefate/test/aefate_engine_test.erl
```diff
@@ -273,20 +293,60 @@ make_call(Contract, Function0, Arguments) ->
...
+      fun({C, Functions}, {Ts0, Nonce}) ->
+              FateCode = setup_contract(Functions),
+              ByteCode = aeb_fate_code:serialize(FateCode),
+              Code     = aeser_contract_code:serialize(
+                           #{byte_code => ByteCode,
+                             type_info => [],
+                             contract_source =>
+                                 %% Ode to the idea of making this field mandatory
+                                 "main :: IO ()\n" ++
+                                 "main = putStrLn \"what are you staring at?\"\n"
```
gorbak25, line 323: Add more monads and apply a ton of functors ;)

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -1239,6 +1261,134 @@ spend(Arg0, Arg1, ES0) ->
...
```
gorbak25, line 1386: How easy is it to calculate this hash on the client side?
Could you give me a quick snippet how to get an bytecode_hash from contract_create_tx ?

gorbak25 (commented):
Does the bytecode hash include the init function?

radrow:
TODO: remove contract on `CREATE` fail!!!

apps/aefate/src/aefa_fate_op.erl
cytadela8: Now init is never stripped we just check we check if the function called is not init when calling. (unless AllowInit=true)

apps/aefate/src/aefa_fate_op.erl
radrow: No, you need to run erlang console.

radrow:
FIXME: non_payable account for non payable contracts

radrow:
TODO: handle existing account

radrow:
TODO: emit events

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -1247,6 +1261,134 @@ spend(Arg0, Arg1, ES0) ->
...
+    ReplaceInitResult
+        = fun(ES0_) ->
+                  %% init returns unit, so we get rid of it
+                  %% if the call was protected, rewrap the result
+                  case Protected of
+                      unprotected ->
+                          {{tuple, {}}, ES1_} = get_op_arg({stack, 0}, ES0_),
+                          write({stack, 0}, ?FATE_CONTRACT(ContractPK), ES1_);
+                      protected ->
+                          case get_op_arg({stack, 0}, ES0_) of
```
hanssv, line 1323: I don't remember exactly how `protected` works, but the compiler will generate code that wraps the result with a `Some`? If so, then we should never get the second (`None`) case or I am missing something?

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -1247,6 +1261,134 @@ spend(Arg0, Arg1, ES0) ->
...
+                              {{variant,[0,1],1,{{tuple,{}}}}, ES1_} ->
```
hanssv, line 1324:
```suggestion
                              {{variant,[0, 1], 1, {{tuple, {}}}}, ES1_} ->
```

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -1247,6 +1261,134 @@ spend(Arg0, Arg1, ES0) ->
...
+                                  write( {stack, 0}, make_some(?FATE_CONTRACT(ContractPK))
+                                       , ES1_
+                                       );
+                              {{variant, [0,1],0,{}}, ES1_} ->
```
hanssv, line 1328:
```suggestion
                              {{variant, [0, 1], 0, {}}, ES1_} ->
```

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -1247,6 +1261,134 @@ spend(Arg0, Arg1, ES0) ->
...
+    Contract0  =
+        case CodeOrPK of
+            {new, Code} -> aect_contracts:new(
+                              Current, Nonce,
+                              #{vm => aefa_engine_state:vm_version(ES0)
+                              , abi => ?ABI_FATE_SOPHIA_1
+                              }, Code, 0);
+            {clone, PK} -> aect_contracts:new_clone(
+                             Current, Nonce,
+                             #{vm => aefa_engine_state:vm_version(ES0)
```
hanssv, line 1366: Doesn't this potentially change the VM version of a contract? For now we only allow to clone FATEv2 contracts, and any contract doing clone will also be on FATEv2. But in the future if there is a FATEv3, etc?

apps/aefate/src/aefa_fate_op.erl
hanssv: I think it should have a dynamic cost for the size, it is done for other functions/operations!

hanssv (commented):
Initial impression is good! 

Not sure about the `delete_contract` and `delete_account` functions are we planning on deleting accounts?

Let's look at payable/existing accounts carefully at some point.

apps/aefate/src/aefa_utils.erl
```diff
@@ -133,4 +135,27 @@ decode_variant([], [], _N, _Values, '$undefined$') ->
 decode_variant([], [], _N, _Values, Acc) ->
     Acc.
 
-
```
hanssv, line 136: This is an internal test format, it should not be moved from `aefate/test` to `aefate/src`

hanssv (changes requested):

apps/aefate/src/aefa_chain_api.erl
hanssv: The code does not consider `payable`/`non_payable`

But I wonder if this code should not use `aeprimop` accessing `aeprimop_state` directly is not right?

apps/aefate/src/aefa_fate_op.erl
radrow: In the future I would assert that the version is the same as caller's

apps/aefate/src/aefa_fate_op.erl
hanssv: Ulf explained...

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -1247,6 +1261,134 @@ spend(Arg0, Arg1, ES0) ->
...
+        get_op_args([Arg0, Arg1, Arg2, Arg3], ES0),
+    clone_common(CloneePK, InitArgsTypes, Value, no_gas_cap, Prot, ES1).
+
+clone_g(Arg0, Arg1, Arg2, Arg3, Arg4, ES0) ->
+    ?AVAILABLE_FROM(?VM_FATE_SOPHIA_2, ES0),
+    {[?FATE_CONTRACT(CloneePK), InitArgsTypes, Value, GasCap, Prot], ES1} =
+        get_op_args([Arg0, Arg1, Arg2, Arg3, Arg4], ES0),
+    clone_common(CloneePK, InitArgsTypes, Value, {gas_cap, GasCap}, Prot, ES1).
+
+clone_common(CloneePK, InitArgsTypes, Value, GasCap, Prot, ES0) ->
```
hanssv, line 1302: `clone_common` and `create_common` are sufficiently similar to merit a refactoring it could take `{new, Code}` or `{clone, PK}` as argument and set `Protected` to `unprotected` in the create case.

apps/aefate/src/aefa_fate_op.erl
hanssv: I would assume the VM-version should be the same as the cloned contract originally have?!

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -1247,6 +1261,134 @@ spend(Arg0, Arg1, ES0) ->
...
+                      unprotected ->
+                          {{tuple, {}}, ES1_} = get_op_arg({stack, 0}, ES0_),
+                          write({stack, 0}, ?FATE_CONTRACT(ContractPK), ES1_);
+                      protected ->
+                          case get_op_arg({stack, 0}, ES0_) of
+                              {{variant,[0, 1], 1, {{tuple, {}}}}, ES1_} ->
+                                  write( {stack, 0}, make_some(?FATE_CONTRACT(ContractPK))
+                                       , ES1_
+                                       );
+                              {{variant, [0, 1], 0, {}}, ES1_} ->
```
hanssv, line 1328: `aefa_engine_state:remove_contract` should be called here rather than below?! Or else the contract will remain if the init-function fails...

UlfNorell:
Looked through this with @hanssv, and one comment is that there's quite a lot of machinery for handling protected cloning (remove contraction functions and a new kind of continuation frame on the call stack). Is there a strong business case for this to warrant the added complexity? If not, I would be tempted to drop it.

radrow:
@UlfNorell Regarding continuation – I couldn't invent any other way to overwrite the results of the `init` call – as it returns `{}` at this point. On the other hand we could just make it always return the PK of the newly created contract, but I am afraid that it could break some other stuff. But if not – I am absolutely for it.

apps/aefate/src/aefa_fate_op.erl
radrow: I don't get the issue. It works in tests, the compiler should be adjustable to this behavior?

apps/aefate/src/aefa_fate_op.erl
radrow: done

apps/aefate/src/aefa_utils.erl
radrow: done

apps/aecontract/include/address.hrl
```diff
@@ -0,0 +1,9 @@
+%% Arguments like <<_:256>> are encoded as addresses.
```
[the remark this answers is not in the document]
radrow: done

UlfNorell:
I understand why the continuation is there. It's not a bad solution to the problem. My question was whether having protected cloning is worth the extra complexity.

apps/aefate/src/aefa_fate_op.erl
radrow: Why? `{failed_protected_call, ES4} ->` case should trigger right on the `init` fail, won't it?

radrow:
Continuations are there not only because of protected clones, but also for normal clones and create. The only added complexity related uniquely to the protected clone is contract removal which

- isn't very much complex in the code
- may be useful later in some other cases anyway

The only problematic thing here is to agree on some edge case behaviors (like creating a payable account by positive valued spend, cloning a non-payable contract on it, failing in init and wondering what should happen with the account and the funds).

apps/aefate/src/aefa_chain_api.erl
radrow: Should work now. I fix the accounts' payability on collision and I leave the account on init fail – just in case it was there before.

apps/aefate/src/aefa_fate_op.erl
radrow: done

apps/aefate/src/aefa_fate_op.erl
hanssv: I think the `{failed_protected_call, ES4}` will only hit if the checks for the remote call fail? This is why we have the continuation 🤔

radrow: wait, it won't be called if the remote call aborts?!

hanssv:
The roll-back case when init fails is tricky (even without the special case where the account is pre-existing) we would like to simply rollback to an earlier engine state, except for gas spent etc.

apps/aefate/src/aefa_fate_op.erl
radrow: The protection seems to handle `abort` as well as all other errors or do I misunderstand something?
```
protect(unprotected, Action, _) -> Action();
protect(protected, Action, Recover) ->
    try Action() catch _:_ -> Recover() end.
```

apps/aecore/src/aec_accounts.erl
```diff
@@ -129,6 +132,13 @@ type(#account{ ga_contract = C })         -> contract = aeser_id:specialize_type
 is_payable(#account{ flags = 0 }) -> true;
 is_payable(#account{ flags = N }) -> not get_flag(non_payable, N).
 
+-spec set_payable(account(), boolean()) -> account().
+set_payable(Acc = #account{ flags = N }, false) ->
+    Acc#account{ flags = (N bsr 1) bsl 1 };
+set_payable(Acc, true) ->
+    #account{ flags = N } = set_payable(Acc, false),
+    Acc#account{ flags = N + ?FLAG_NON_PAYABLE_VALUE }.
```
hanssv, line 140: The first clause should use `?FLAG_NON_PAYABLE_VALUE` (or rather `?FLAG_NON_PAYABLE` which isn't there).

We need to fix flag handling once and for all, and we also need to handle `non_payable` when creating a contract where the account pre-exists. Let's do that in another issue (#3579)

apps/aefate/src/aefa_fate_op.erl
hanssv: You also have the `check_return_type_protected` involved

apps/aecore/src/aec_accounts.erl
radrow: Now should be nicer

apps/aefate/src/aefa_fate_op.erl
radrow: Now it is linear to the size of the hashed contract with coef=1. May be a bit too low, doesn't it?

apps/aefate/src/aefa_fate_op.erl
UlfNorell: `remote_call_common` returns before jumping to the actual code (this is why you need the continuation!). The `{jump, 0, ES4}` in the `ok` case is what starts the execution of the `init` function. So clearly if the init call fails there's no way we could end up in `failed_protected_call` branch.

radrow:
# GAS OF `BYTECODE_HASH`
I have measured the gas and execution time on my PC. It *is* inaccurate, but the rank of the numbers should be reasonable enough to be considered;

So, the simple function
``` 
function f(n) = if(n == 0) 0 else f(n-1)
```
Called as `f(10000000)` takes about `130 000 000` gas and evaluates for about 13 seconds which gives 1 second of execution worth 10kk gas.

Hashing `6 000 000 000` bytes takes 10 seconds on my PC.  Therefore hashing 1 byte should take 1/60 of gas. 

## Conclusion

1 gas per byte should be fair in this case #checkmymath

apps/aefate/src/aefa_fate_op.erl
radrow: done

apps/aecontract/include/fate_type_macros.hrl
```diff
@@ -0,0 +1,9 @@
+%% Arguments like <<_:256>> are encoded as addresses.
+%% For FATE we need to distinguish between the type of object.
+%% For test cases that is run both for FATE and AEVM, wrap args in
```
cytadela8, line 3: Maybe move this to aecontract/test/include?

hanssv (approved):
👍

apps/aecore/src/aec_accounts.erl
```diff
@@ -129,6 +132,12 @@ type(#account{ ga_contract = C })         -> contract = aeser_id:specialize_type
 is_payable(#account{ flags = 0 }) -> true;
 is_payable(#account{ flags = N }) -> not get_flag(non_payable, N).
 
+-spec set_payable(account(), boolean()) -> account().
+set_payable(Acc = #account{ flags = N }, false) ->
+    Acc#account{ flags = N band (2#11111111 - ?FLAG_NON_PAYABLE_VALUE) };
```
cytadela8, line 137:
```suggestion
    Acc#account{ flags = N band (bnot ?FLAG_NON_PAYABLE_VALUE) };
```

Works as expected:
obraz

cytadela8 (dismissed):

hanssv (approved):
👍

apps/aefate/src/aefa_chain_api.erl
```diff
@@ -194,9 +201,41 @@ traverse_to_key_hash(H, KeyHash) ->
...
+-spec put_contract(aect_contracts:contract(), state()) -> state().
+put_contract(Contract, #state{primop_state = PS} = S) ->
+    PK = aect_contracts:pubkey(Contract),
+    {Payable, PS0} = case aect_contracts:code(Contract) of
+                  {code, SerCode} ->
+                      #{payable := P} = aeser_contract_code:deserialize(SerCode),
+                      {P, PS};
+                  {ref, Clonee} ->
+                      case aeprimop_state:find_account(aeser_id:specialize(Clonee, contract), PS) of
+                          none -> error;
```
gorbak25, line 213: Is it OK that you hard crash with a match error in that case?

apps/aefate/src/aefa_chain_api.erl
```diff
@@ -208,19 +247,43 @@ contract_fate_code(Pubkey, #state{primop_state = PState} = S) ->
...
+contract_find_final_ref(Pubkey, #state{primop_state = PState} = S0) ->
+    case aeprimop_state:find_contract_without_store(Pubkey, PState) of
+        none -> error;
+        {value, Contract} ->
+            case aect_contracts:code(Contract) of
+                {code, _} ->
+                    {ok, Pubkey, aect_contracts:vm_version(Contract)};
+                {ref, Ref} ->
+                    RefContractPK = aeser_id:specialize(Ref, contract),
+                    contract_find_final_ref(RefContractPK, S0)
```
gorbak25, line 276: Infinite loop in case of cyclic references - Good thing we enforce that we can't create cyclic references :)

apps/aefate/src/aefa_chain_api.erl
```diff
@@ -208,19 +247,43 @@ contract_fate_code(Pubkey, #state{primop_state = PState} = S) ->
...
+            end
+    end.
+
+-spec remove_contract(pubkey(), state()) -> state().
+remove_contract(Pubkey, #state{primop_state = PState0} = S0) ->
+    PState1 = aeprimop_state:delete_contract(Pubkey, PState0),
+    %% We don't delete the account as it could have existed before
```
gorbak25, line 283: But you still leave this account when it was created ;)

apps/aefate/src/aefa_chain_api.erl
radrow: This code actually prevents cyclic references from happening – unless someone puts in their DB such a one manually (which would be invalid anyways).

apps/aefate/src/aefa_chain_api.erl
radrow: Yes, we have agreed on that behavior as it would be ambiguous in case the account has existed before.

apps/aefate/src/aefa_chain_api.erl
radrow: This is an internal error which happens only in case someone refers to a contract without an account which should be invalid in the first place. The match crash shall point to a code that clearly shows what was the issue

gorbak25 (approved):

PR #3546, opened 2021-03-12
Include full FATE 2 code with init in trees

cytadela8:
Init removal is done in aefate unless `allow_init` is passed.

aeternity/protocol#482

### DIFF
--- a/apps/aecontract/src/aect_channel_contract.erl
+++ b/apps/aecontract/src/aect_channel_contract.erl
@@ -75,14 +75,15 @@ run_new(ContractPubKey, Call, CallData, Trees0, OnChainTrees,
 
 prepare_init_call(VmVersion, Protocol, Contract, Trees0) when ?IS_FATE_SOPHIA(VmVersion) ->
     Code = #{ byte_code := ByteCode } = aeser_contract_code:deserialize(aect_contracts:code(Contract)),
-    FateCode  = aeb_fate_code:deserialize(ByteCode),
-    FateCode1 =
+    ByteCode1 =
         if
-        % Before FATE 2 we do not include init function in code put in trees
-            VmVersion == ?VM_FATE_SOPHIA_1 -> aeb_fate_code:strip_init_function(FateCode);
-            true -> FateCode
+            % Before FATE 2 we do not include init function in code put in trees
+            VmVersion == ?VM_FATE_SOPHIA_1 ->
+                FateCode  = aeb_fate_code:deserialize(ByteCode),
+                FateCode1 = aeb_fate_code:strip_init_function(FateCode),
+                aeb_fate_code:serialize(FateCode1);
+            true -> ByteCode
         end,
-    ByteCode1 = aeb_fate_code:serialize(FateCode1),
     Code1     = Code#{ byte_code := ByteCode1 },
     %% The serialization was broken in the Lima release - setting the
     %% compiler version to "unknown" regardless of the actual value.
--- a/apps/aecontract/src/aect_channel_contract.erl
+++ b/apps/aecontract/src/aect_channel_contract.erl
@@ -158,7 +158,7 @@ run(ContractPubKey, ABIVersion, Call, CallData, CallStack, Trees0,
         false                                -> erlang:error(wrong_abi_version)
     end,
     CallDef = make_call_def(CallerPubkey, OwnerPubkey, ContractPubKey, Gas, GasPrice, Amount,
-              CallData, false, CallStack, Code, Store, Call, OnChainTrees, OnChainEnv, Trees0),
+              CallData, _AllowInit = false, CallStack, Code, Store, Call, OnChainTrees, OnChainEnv, Trees0),
     {CallRes, Trees, _} = aect_dispatch:run(#{vm => VmVersion, abi => ABIVersion}, CallDef),
     UpdatedTrees = aect_utils:insert_call_in_trees(CallRes, Trees),
     {aec_trees:gc_cache(UpdatedTrees, [accounts, contracts]), CallRes}.
--- a/apps/aecontract/src/aect_channel_contract.erl
+++ b/apps/aecontract/src/aect_channel_contract.erl
@@ -43,7 +43,7 @@ run_new(ContractPubKey, Call, CallData, Trees0, OnChainTrees,
     CallDef = make_call_def(OwnerPubKey, OwnerPubKey, ContractPubKey,
                             _Gas = 1000000, _GasPrice = 1,
                             _Amount = 0, %TODO: make this configurable
-                            CallData, CallStack, Code, Store, Call, OnChainTrees, OnChainEnv, Trees1),
+                            CallData, _AllowInit = true, CallStack, Code, Store, Call, OnChainTrees, OnChainEnv, Trees1),
     CTVersion = aect_contracts:ct_version(Contract),
     {CallRes, Trees, _} = aect_dispatch:run(CTVersion, CallDef),
     case aect_call:return_type(CallRes) of
@@ -76,7 +76,12 @@ run_new(ContractPubKey, Call, CallData, Trees0, OnChainTrees,
 prepare_init_call(VmVersion, Protocol, Contract, Trees0) when ?IS_FATE_SOPHIA(VmVersion) ->
     Code = #{ byte_code := ByteCode } = aeser_contract_code:deserialize(aect_contracts:code(Contract)),
     FateCode  = aeb_fate_code:deserialize(ByteCode),
-    FateCode1 = aeb_fate_code:strip_init_function(FateCode),
+    FateCode1 =
+        if
+        % Before FATE 2 we do not include init function in code put in trees
+            VmVersion == ?VM_FATE_SOPHIA_1 -> aeb_fate_code:strip_init_function(FateCode);
+            true -> FateCode
+        end,
     ByteCode1 = aeb_fate_code:serialize(FateCode1),
     Code1     = Code#{ byte_code := ByteCode1 },
     %% The serialization was broken in the Lima release - setting the
@@ -153,7 +158,7 @@ run(ContractPubKey, ABIVersion, Call, CallData, CallStack, Trees0,
         false                                -> erlang:error(wrong_abi_version)
     end,
     CallDef = make_call_def(CallerPubkey, OwnerPubkey, ContractPubKey, Gas, GasPrice, Amount,
-              CallData, CallStack, Code, Store, Call, OnChainTrees, OnChainEnv, Trees0),
+              CallData, false, CallStack, Code, Store, Call, OnChainTrees, OnChainEnv, Trees0),
     {CallRes, Trees, _} = aect_dispatch:run(#{vm => VmVersion, abi => ABIVersion}, CallDef),
     UpdatedTrees = aect_utils:insert_call_in_trees(CallRes, Trees),
     {aec_trees:gc_cache(UpdatedTrees, [accounts, contracts]), CallRes}.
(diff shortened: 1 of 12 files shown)


apps/aecontract/src/aect_channel_contract.erl
```diff
@@ -76,7 +76,12 @@ run_new(ContractPubKey, Call, CallData, Trees0, OnChainTrees,
 prepare_init_call(VmVersion, Protocol, Contract, Trees0) when ?IS_FATE_SOPHIA(VmVersion) ->
     Code = #{ byte_code := ByteCode } = aeser_contract_code:deserialize(aect_contracts:code(Contract)),
     FateCode  = aeb_fate_code:deserialize(ByteCode),
-    FateCode1 = aeb_fate_code:strip_init_function(FateCode),
+    FateCode1 =
+        if
+        % Before FATE 2 we do not include init function in code put in trees
+            VmVersion == ?VM_FATE_SOPHIA_1 -> aeb_fate_code:strip_init_function(FateCode);
```
gorbak25, line 82: BTW. It's a good question whether we shall enable the init function for FATE2 or after the IRIS hard fork.

gorbak25: If we disable creation of FATE1 contracts after IRIS this will be enough.

cytadela8:
```
is_legal_version_at_protocol_(create, #{vm := ?VM_FATE_SOPHIA_1, abi := ?ABI_FATE_SOPHIA_1}, Protocol) ->
    case Protocol of
        ?ROMA_PROTOCOL_VSN            -> false;
        ?MINERVA_PROTOCOL_VSN         -> false;
        ?FORTUNA_PROTOCOL_VSN         -> false;
        ?LIMA_PROTOCOL_VSN            -> true;
        P when P > ?LIMA_PROTOCOL_VSN -> false
    end;
    ```

Comment (cytadela8) [reply apps/aecontract/src/aect_channel_contract.erl:82]: We currently dissalow creation of FATE1 contracts on IRIS. I think there is no reason for anybody to deploy FATE1 after IRIS. The behaviour changes _should not_ affect anybody

**Comment by cytadela8** [comment]
Ready unless somebody we want tests :sweat_smile:

**Review by gorbak25 — CHANGES_REQUESTED** [review]
Add at least one test that the init function is present.

**Comment by cytadela8** [comment]
We have a test for chcking `sophia_no_calls_to_init`. I will add a test to check that the init is present in FATE2 and not present in FATE1

**Comment by gorbak25** [comment]
Document the change in https://github.com/aeternity/protocol/

**Comment by cytadela8** [comment]
I needed fate-interop for tests, also there was a confilic with that change actually

**Comment by radrow** [comment]
Is this one a subset of #3551 ?

**Review by gorbak25 — DISMISSED** [review]

**Review by radrow — DISMISSED** [review]

**Comment by cytadela8** [comment]
DO NOTE MERGE BEFORE AESERALIZATION(update in rebar.config needed)

**Review by radrow — DISMISSED** [review]

File: apps/aecontract/src/aect_channel_contract.erl (hunk 1 above)
Comment (gorbak25) [inline apps/aecontract/src/aect_channel_contract.erl:82]: Move the check one line higher so for FATE2 we avoid unnecessary de-serialization and serialization of the code.

File: apps/aecontract/src/aect_channel_contract.erl [hunk 2]
```diff
@@ -153,15 +158,15 @@ run(ContractPubKey, ABIVersion, Call, CallData, CallStack, Trees0,
         false                                -> erlang:error(wrong_abi_version)
     end,
     CallDef = make_call_def(CallerPubkey, OwnerPubkey, ContractPubKey, Gas, GasPrice, Amount,
-              CallData, CallStack, Code, Store, Call, OnChainTrees, OnChainEnv, Trees0),
+              CallData, false, CallStack, Code, Store, Call, OnChainTrees, OnChainEnv, Trees0),
```
gorbak25, line 161:
```suggestion
              CallData, _AllowInit = false, CallStack, Code, Store, Call, OnChainTrees, OnChainEnv, Trees0),
```

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -6754,6 +6756,44 @@ fate_vm_version_switching(Cfg) ->
     ok.
 
 
+contract_init_on_chain_fate(Cfg) ->
+    state(aect_test_utils:new_state()),
+    ForkHeights   = ?config(fork_heights, Cfg),
```
gorbak25, line 6761: The amount of spaces after the variable jumps from one line to the other. Please fix the formatting

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -6754,6 +6756,44 @@ fate_vm_version_switching(Cfg) ->
...
+    LimaHeight = maps:get(lima, ForkHeights),
+    IrisHeight = maps:get(iris, ForkHeights),
+    Protocol = aec_hard_forks:protocol_effective_at_height(LimaHeight),
+    GasPrice = aec_governance:minimum_gas_price(Protocol),
+    MinerMinGasPrice= aec_tx_pool:minimum_miner_gas_price(),
```
gorbak25, line 6766: Like here

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -6754,6 +6756,44 @@ fate_vm_version_switching(Cfg) ->
...
+    MinGasPrice   = max(GasPrice, MinerMinGasPrice),
+    Acc           = ?call(new_account, 10000000000 * MinGasPrice),
+    LimaSpec      = #{height => LimaHeight, vm_version => ?VM_FATE_SOPHIA_1,
+                      amount => 100,
+                      gas_price => MinGasPrice,
+                      fee => 1000000 * MinGasPrice},
+    IrisSpec      = #{height => IrisHeight, vm_version => ?VM_FATE_SOPHIA_2,
```
gorbak25, line 6773: Move the version to a new line

apps/aefate/src/aefa_fate.erl
```diff
@@ -113,7 +113,7 @@ final_trees(EngineState) ->
     aefa_chain_api:final_trees(aefa_engine_state:chain_api(EngineState)).
 
 verify_init_calldata(CallData) ->
-    Init = aeb_fate_code:symbol_identifier(<<"init">>),
+    Init = ?FATE_INIT_ID,
     case decode_calldata(CallData) of
         {Init, _Args} -> ok;
```
gorbak25, line 118: Perhaps inline?

apps/aefate/src/aefa_fate.erl
```diff
@@ -341,28 +342,35 @@ set_remote_function(Caller, ?FATE_CONTRACT(Pubkey), Function, CheckPayable, Chec
...
             case CheckPrivate andalso is_function_private(Function, ES) of
-                false -> set_local_function(Function, ES);
+                false -> set_local_function(Function, AllowInit, ES);
                 true  -> abort({function_is_private, Function}, ES)
             end;
         false -> abort({function_is_not_payable, Function}, ES)
     end.
 
-set_local_function(Function, ES) ->
+set_local_function(Function, true, ES) ->
```
gorbak25, line 366:
```suggestion
set_local_function(Function, _AllowInit = true, ES) ->
```
Let's clarify what that parameter is supposed to do.

apps/aefate/src/aefa_fate.erl
```diff
@@ -341,28 +342,35 @@ set_remote_function(Caller, ?FATE_CONTRACT(Pubkey), Function, CheckPayable, Chec
...
+    set_local_function_(Function, ES);
+set_local_function(?FATE_INIT_ID, false, ES) ->
+    abort({trying_to_call_function, ?FATE_INIT_ID}, ES);
```
gorbak25, line 369: Is this error message under consensus or just a local artifact?

gorbak25: In either case I would suggest running a "sync from genesis test" before we release IRIS

apps/aefate/src/aefa_fate.erl
```diff
@@ -614,15 +622,17 @@ pop_call_stack(ES) ->
             ES2 = check_return_type_protected(Protected, RetType, TVars, Stores, API, ES1),
             pop_call_stack(ES2);
         {local, Function, TVars, BB, ES1} ->
-            ES2 = set_local_function(Function, ES1),
+            % Allow returning to init
+            ES2 = set_local_function(Function, true, ES1),
```
gorbak25, line 626: Is it possible to manipulate the callstack to change the return function?
Foo -> Bar -> Hack -> NotBar
I don't think it's the case but I want to be fully sure.

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -210,14 +210,22 @@ call(Arg0, EngineState) ->
     {Args, ES2} = aefa_fate:pop_args_from_signature(Signature, ES1),
     ES3         = aefa_fate:push_return_address(ES2),
     ES4         = aefa_fate:bind_args(Args, ES3),
-    {jump, 0, aefa_fate:set_local_function(Fun, ES4)}.
+    AllowInit = case aefa_engine_state:vm_version(ES3) of
```
gorbak25, line 213: Good catch :)

apps/aeprimop/src/aeprimop.erl
```diff
@@ -1473,7 +1473,12 @@ prepare_init_call(Code, Contract, S = #state{protocol = Protocol}) ->
         #{vm := V} when ?IS_FATE_SOPHIA(V) ->
             #{ byte_code := ByteCode } = Code,
             FateCode  = aeb_fate_code:deserialize(ByteCode),
-            FateCode1 = aeb_fate_code:strip_init_function(FateCode),
+            FateCode1 =
```
gorbak25, line 1476: Move the check one line higher to avoid unnecesary deserialization/serialization

docs/release-notes/next-iris/GH-3546-fate-include-init.md
```diff
@@ -0,0 +1 @@
+* Include full FATE 2 code with init in trees this improvement includes state channels
```
gorbak25, line 1:
```suggestion
* Include full FATE 2 code with init in state trees. This also applies to offchain contracts in state channels
```

gorbak25 (changes requested):

apps/aefate/src/aefa_fate.erl
cytadela8: No idea if it is under consensus. Would need to look into how the errors are handled in call tree or some tx tree. I kept the error as before to avoid headache.

apps/aefate/src/aefa_fate.erl
cytadela8: So if this would be the case it would ofcourse affect not only init. Anyway.

As I understand our callstack assures strict separation of functions even inside one contract. `call_stack` is not accessible in any direct way. It's only changed by the VM when a call is made or we are returning. We have also `accumulator` and `accumulator_stack` that are saved on each call to make sure callee can't affect calculations of the caller.

apps/aefate/src/aefa_fate.erl
cytadela8: After some investigation, I don't believe it is under consensus.

apps/aecontract/test/aecontract_SUITE.erl
```diff
@@ -272,6 +273,10 @@
 
 -define(assertMatchFATE(__ExpVm1, __ExpVm2, __Res),
     case vm_version() of
+        ?VM_AEVM_SOPHIA_1 -> ok;
```
gorbak25, line 276: ???

gorbak25 (approved):

radrow (approved):

PR #3548, opened 2021-03-15
Primeop contract clone

cytadela8:
On top of #3546


cytadela8:
We are not going to do this through primeop.

Issue #3549, opened 2021-03-16
Mempool: allow including already GCed transactions


velzevur:
Garbage collecting transactions is good but the current implementation forbids an already GCed transaction to be added to the mempool. This was initially a spam prevention mechanism which is trivial to work around if a malicious user wants to. Transactions that are already GCed shall be allowed to be included in the mempool as long as they are valid.

PR #3550, opened 2021-03-16
Fix forking orphan txs

velzevur:
Exposes and fixs an issue with oprhaned transactions: while they rightfully
are returned to the tx pool, this only helps if the node actually becomes a
generation leader. Those were not broadcasted.

This fixes the following scenario:

We have 3 miners: `Alice`, `Bob` and `Carol`. The current leader is `Alice`.
`Alice` receives a transaction `Tx1` and imidiately includes it in a
microblock, as she should. Since she is including it in a block, she
broadcasts it to `Bob` but she also sends him the microblock. Meanwhile
`Carol` produces a new keyblock but since `Carol` had not seen yet this
microblock, the keyblock is based on a previous block instead. Once `Carol`'s
keyblock propagates the network, it orphans Alice's microblock and rolls back
`Tx1`. This transaction now is in the mempool of both `Alice` and `Bob` but
they don't broadcast it.  `Alice`'s microblock could reach `Carol` via gossip
but since there is a heavier fork with more mining being done (`Carol`'s fork)
- it is rejected and `Carol` ignores those transactions that come in the
microblock. Because of races in the network, `Carol` might also not receive
the microblock at all. This opens the door for a race condition where the
transaction is initially included in a microblock but then rolled back and not
re-included as long as `Carol` is the generation leader. It would be included
only after either `Alice` or `Bob` are elected for a generation leader.

What this PR does is once a transaction is orphaned and returned to the
mempool, an event is created so that the transaction is also picked by `sync`
and being gossiped as it would be if received for the first time. Instead of
broadcasting it to only already fully synced nodes, it is being sent to all
connected peers (even not synced ones yet), which aims at further improving
the network propagation time.

It is worth mentioning that this does not impact consensus and if a
participant had already seen this transaction, they would simply ignore the
gossiped message. If they have not yet received it, though, it would end up in
their mempool, which is what we aim for.

This could be further improved by #3549.

### DIFF
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -758,7 +758,9 @@ remove_locations(Hash, Hash) ->
 remove_locations(StopHash, CurrentHash) ->
     lists:foreach(fun(TxHash) ->
                           aec_db:remove_tx_location(TxHash),
-                          aec_db:add_tx_hash_to_mempool(TxHash)
+                          aec_db:add_tx_hash_to_mempool(TxHash),
+                          Tx = aec_db:get_signed_tx(TxHash),
+                          aec_events:publish(tx_received, Tx)
                   end, db_safe_get_tx_hashes(CurrentHash)),
     remove_locations(StopHash, db_get_prev_hash(CurrentHash)).
 
--- a/apps/aecore/test/aecore_forking_SUITE.erl
+++ b/apps/aecore/test/aecore_forking_SUITE.erl
@@ -14,7 +14,6 @@
 %% test case exports
 -export(
    [
-    sync_fork_in_wrong_order/1,
     add_dev3_node/1,
     dev3_failed_attack/1,
     dev3_syncs_to_community/1,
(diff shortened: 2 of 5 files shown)


apps/aecore/test/aecore_forking_SUITE.erl
```diff
@@ -288,3 +321,104 @@ patterns() ->
...
+    ct:log("top of chain ~p: ~p", [ CorrectForkNode, CFTop ]),
+    stop_and_check([CorrectForkNode], Config), 
+
+    start_node(OtherNode, Config), 
+    ForkNName = aecore_suite_utils:node_name(OtherNode),
+    ForkTop = rpc:call(ForkNName, aec_chain, top_block, [], 5000),
+    ct:log("top of chain ~p: ~p", [ OtherNode, ForkTop ]),
+
+    false = (ForkTop == CFTop),
+    timer:sleep(100),
```
gorbak25, line 396: aec_test_utils:wait_for_it ?

apps/aecore/src/aec_chain_state.erl
```diff
@@ -758,7 +758,9 @@ remove_locations(Hash, Hash) ->
 remove_locations(StopHash, CurrentHash) ->
     lists:foreach(fun(TxHash) ->
                           aec_db:remove_tx_location(TxHash),
-                          aec_db:add_tx_hash_to_mempool(TxHash)
+                          aec_db:add_tx_hash_to_mempool(TxHash),
+                          Tx = aec_db:get_signed_tx(TxHash),
```
gorbak25, line 762: The good thing is that this won't regress on sync performance - this only executes on microforks.

gorbak25 (approved):

uwiger (approved):

PR #3551, opened 2021-03-16
Store contract code by reference (for contract clone)

cytadela8:
In order for contract cloning to be efficient we do not want to store another copy of contract code after cloning. This PR introduces a way to store reference to contract containing actual code in contract tree structure.

### DIFF
--- a/apps/aecontract/src/aect_contracts.erl
+++ b/apps/aecontract/src/aect_contracts.erl
@@ -68,8 +68,7 @@
         id           :: aeser_id:id(),
         owner_id     :: aeser_id:id(),
         ct_version   :: version(),
-        code         :: binary(),     %% The byte code when code_ref=no_ref
-        code_ref     :: no_ref | {ref, aeser_id:id()},     %% no_ref or DIRECT reference to contract with bytecode
+        code         :: {code, binary()} | {ref, aeser_id:id()}, %% the bytecode or DIRECT reference to contract with bytecode
         store        :: store(),      %% The current state/store (stored in a subtree in mpt)
         log          :: binary(),     %% The current event log
         active       :: boolean(),    %% false when disabled, but don't remove unless referrer_ids == []
@@ -247,21 +246,20 @@ new(RTx) ->
 -spec new(aec_keys:pubkey(), ct_nonce(), version(), binary(), amount()) -> contract().
 %% NOTE: Should only be used for contract execution without transaction
 new(Owner, Nonce, CTVersion, Code, Deposit) ->
-    new_internal(Owner, Nonce, CTVersion, Code, no_ref, Deposit).
+    new_internal(Owner, Nonce, CTVersion, {code, Code}, Deposit).
 
 %% NOTE: When cloning passed CodeRefPK has to have code_ref=no_ref (NO INDIRECT CODE REFERENCE)
 -spec new_clone(aec_keys:pubkey(), ct_nonce(), version(), aec_keys:pubkey(), amount()) -> contract().
 new_clone(Owner, Nonce, CTVersion, CodeRefPK, Deposit) ->
     CodeRef = aeser_id:create(contract, CodeRefPK),
-    new_internal(Owner, Nonce, CTVersion, <<>>, {ref, CodeRef}, Deposit).
+    new_internal(Owner, Nonce, CTVersion, {ref, CodeRef}, Deposit).
 
-new_internal(Owner, Nonce, CTVersion, Code, CodeRef, Deposit) ->
+new_internal(Owner, Nonce, CTVersion, Code, Deposit) ->
     Pubkey = compute_contract_pubkey(Owner, Nonce),
     C = #contract{ id           = aeser_id:create(contract, Pubkey),
                    owner_id     = aeser_id:create(account, Owner),
                    ct_version   = CTVersion,
                    code         = Code,
-                   code_ref     = CodeRef,
                    store        = aect_contracts_store:new(),
                    log          = <<>>,
                    active       = true,
@@ -275,7 +273,7 @@ new_internal(Owner, Nonce, CTVersion, Code, CodeRef, Deposit) ->
 -spec serialize(contract()) -> serialized().
 serialize(#contract{owner_id     = OwnerId,
                     ct_version   = CTVersion,
-                    code         = Code,
+                    code         = {code, Code},
                     log          = Log,
                     active       = Active,
                     referrer_ids = ReferrerIds,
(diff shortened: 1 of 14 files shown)


apps/aecontract/src/aect_contracts.erl
```diff
@@ -245,11 +251,13 @@ new(Owner, Nonce, CTVersion, Code, Deposit) ->
...
                    ct_version   = CTVersion,
                    code         = Code,
+                   code_ref     = no_ref,
                    store        = aect_contracts_store:new(),
                    log          = <<>>,
                    active       = true,
                    referrer_ids = [],
-                   deposit      = Deposit
+                   deposit      = Deposit,
+                   version      = serialization_version_for_vm(CTVersion)
```
gorbak25, line 260: You need to transform the state trees when enabling IRIS - btw. take a look at how we manage multiple versions of account nodes in account state trees.

gorbak25 (changes requested):

apps/aecontract/src/aect_contracts.erl
cytadela8: Why do I need to do this? This way I managed to avoid doing that.

gorbak25: You have still old DB records present in the DB. The only reason why the CI suite didn't fail is that our DB smoke tests only test ~1k blocks - essentially no contracts were present back then.
How to crash your code:
1. Take a snapshot of mainnet(https://downloads.aeternity.io/)
2. Start the node
3. Observe as block insertion crashes as contract_call tx crash your code ;P

apps/aecontract/src/aect_contracts.erl
```diff
@@ -294,42 +330,100 @@ serialize_for_client(#contract{id           = Id,
 
 -spec deserialize(pubkey(), serialized()) -> contract().
 deserialize(Pubkey, Bin) ->
+    {?CONTRACT_TYPE, Vsn, RawFields} = aeser_chain_objects:deserialize_type_and_vsn(Bin),
+    deserialize(Pubkey, Vsn, RawFields).
+
+deserialize(Pubkey, ?PRE_IRIS_VSN, RawFields) ->
```
gorbak25, line 336: I'm not fully sure that we store the contracts in serialized form in the DB

apps/aecontract/src/aect_contracts.erl
cytadela8: It will not cause I made sure old serialization and deserialization is preserved. Actually your solution will not work as we can't change old serialization, becouse it would change hashes. Please look how serialization and deserailization is handled.

apps/aecontract/src/aect_contracts.erl
cytadela8: We store in serialized AFAIK. How else would we store it? See `aect_state_tree:insert_contract`

apps/aecontract/src/aect_contracts.erl
```diff
@@ -67,8 +68,8 @@
         id           :: aeser_id:id(),
         owner_id     :: aeser_id:id(),
         ct_version   :: version(),
-        code         :: binary(),     %% The byte code
-        code_ref     :: no_ref | {ref, aeser_id:id()},
+        code         :: binary(),     %% The byte code when code_ref=no_ref
+        code_ref     :: no_ref | {ref, aeser_id:id()},     %% no_ref or DIRECT reference to contract with bytecode
```
radrow, line 72: Why not 
```suggestion
        code     :: {code, binary()} | {ref, aeser_id:id()},
```

cytadela8: To be honest... no reason. I will change that. This is kind of internal working of aect_contracts. And it will be better if I combine the fields. First I did it like this to reflect serialized and deserialized form where we don't have access to complex types.

cytadela8: Done

radrow (approved):
Looks good!

radrow (dismissed):

gorbak25 (changes requested):
Wait for merging #3546

apps/aecontract/src/aect_contracts.erl
```diff
@@ -60,18 +61,20 @@
...
         ct_version   :: version(),
-        code         :: binary(),     %% The byte code
+        code         :: {code, binary()} | {ref, aeser_id:id()}, %% the bytecode or DIRECT reference to contract with bytecode
         store        :: store(),      %% The current state/store (stored in a subtree in mpt)
         log          :: binary(),     %% The current event log
         active       :: boolean(),    %% false when disabled, but don't remove unless referrer_ids == []
         referrer_ids :: list(aeser_id:id()), %% List of contracts depending on this contract
-        deposit      :: amount()
+        deposit      :: amount(),
+        version      :: serialization_version()
```
gorbak25, line 77: Why is version last not first @cytadela8?

apps/aecontract/src/aect_contracts.erl
```diff
@@ -240,6 +246,15 @@ new(RTx) ->
 -spec new(aec_keys:pubkey(), ct_nonce(), version(), binary(), amount()) -> contract().
 %% NOTE: Should only be used for contract execution without transaction
 new(Owner, Nonce, CTVersion, Code, Deposit) ->
+    new_internal(Owner, Nonce, CTVersion, {code, Code}, Deposit).
+
+%% NOTE: When cloning passed CodeRefPK has to have code_ref=no_ref (NO INDIRECT CODE REFERENCE)
+-spec new_clone(aec_keys:pubkey(), ct_nonce(), version(), aec_keys:pubkey(), amount()) -> contract().
+new_clone(Owner, Nonce, CTVersion, CodeRefPK, Deposit) ->
+    CodeRef = aeser_id:create(contract, CodeRefPK),
```
gorbak25, line 254: Why do we use a full id here? Why not just the bytes?

apps/aecontract/src/aect_contracts.erl
```diff
@@ -249,31 +264,59 @@ new(Owner, Nonce, CTVersion, Code, Deposit) ->
...
+                    active       = Active,
+                    referrer_ids = ReferrerIds,
+                    deposit      = Deposit,
+                    version      = ?CODE_REF_VSN}) ->
+    {CodeMaybe, RefMaybe} =
+        case Code of
+            {code, C} -> {C, []};
+            {ref, Ref} -> {<<>>, [Ref]}
+        end,
+    aeser_chain_objects:serialize(
```
gorbak25, line 307: <3

gorbak25 (approved):

radrow (approved):

apps/aecontract/src/aect_contracts.erl
cytadela8: No reason. I appended the new field.

apps/aecontract/src/aect_contracts.erl
cytadela8: Cause we have a serialization type for IDs and not for fixed size byte fields. Also we use IDs for other simillar fields like the `id` field.

PR #3552, opened 2021-03-16
Improve the release process

velzevur:
This automates the block info field in the keyblock. It also provides caching
capabilities for the revision and version. Mining tests nonces do not have to
be updated on every new info as well.

A tribute to @dincho for pushing for this :)

### DIFF
--- a/apps/aecore/test/aec_eunit_SUITE.erl
+++ b/apps/aecore/test/aec_eunit_SUITE.erl
@@ -130,10 +130,13 @@ application_test(Config) ->
     meck:expect(aec_fork_block_settings, lima_extra_accounts, 0, []),
     meck:expect(aec_fork_block_settings, lima_contracts, 0, []),
     meck:expect(aec_fork_block_settings, block_whitelist, 0, #{}),
+    meck:new(aeu_info, []),
+    meck:expect(aeu_info, block_info, 0, 591),
     {ok,_} = application:ensure_all_started(aecore),
     timer:sleep(100),
     ok = application:stop(aecore),
     meck:unload(aec_fork_block_settings),
+    meck:unload(aeu_info),
 
     app_stop(StartedApps -- ?TO_BE_STOPPED_APPS_BLACKLIST, TempDir).
 
--- a/apps/aecore/test/aec_mining_tests.erl
+++ b/apps/aecore/test/aec_mining_tests.erl
@@ -46,7 +46,8 @@ mine_block_test_() ->
                              ?IRIS_PROTOCOL_VSN    -> 9446698485151902999 
                          end,
                  {BlockCandidate,_} = aec_test_utils:create_keyblock_with_state(
-                                        [{TopBlock, aec_trees:new()}], ?TEST_PUB),
+                                        [{TopBlock, aec_trees:new()}],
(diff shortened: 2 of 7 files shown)


apps/aeutils/src/aeu_info.erl
```diff
@@ -37,3 +48,24 @@ read_trimmed_file(Filename) ->
...
+cached_file(Filename, ReadFun) ->
+    EtsKey = {file, Filename},
+    try
+        [{EtsKey, Data}] = ets:lookup(?ETS_TABLE, EtsKey),
+        Data
+    catch
+        _:_ ->
+            case ets:whereis(?ETS_TABLE) of
+                undefined ->
+                    ets:new(?ETS_TABLE, [named_table, {read_concurrency, true}, public]);
```
uwiger, line 64: Which process becomes the owner of the table?

uwiger (approved):

apps/aecore/test/aec_eunit_SUITE.erl
```diff
@@ -130,10 +130,13 @@ application_test(Config) ->
     meck:expect(aec_fork_block_settings, lima_extra_accounts, 0, []),
     meck:expect(aec_fork_block_settings, lima_contracts, 0, []),
     meck:expect(aec_fork_block_settings, block_whitelist, 0, #{}),
+    meck:new(aeu_info, []),
+    meck:expect(aeu_info, block_info, 0, 591),
```
gorbak25, line 134: Will this need to be updated every release?

apps/aecore/test/aec_mining_tests.erl
```diff
@@ -45,8 +45,14 @@ mine_block_test_() ->
                              ?LIMA_PROTOCOL_VSN    -> 2331533446344578375;
                              ?IRIS_PROTOCOL_VSN    -> 9446698485151902999 
                          end,
+                 Info =
+                    case aec_hard_forks:protocol_effective_at_height(Height + 1) of
+                        ?ROMA_PROTOCOL_VSN -> default;
+                        _ -> 591
```
gorbak25, line 51: Same here?

apps/aeutils/src/aeu_info.erl
```diff
@@ -37,3 +48,24 @@ read_trimmed_file(Filename) ->
 
 trim_ending_whitespace(Binary) ->
     re:replace(Binary, "\\s+$", "", [{return, binary}, global]).
+
+cached_file(Filename) ->
+    cached_file(Filename, fun read_trimmed_file/1).
+
+cached_file(Filename, ReadFun) ->
```
gorbak25, line 55: Why not persistent_term?

apps/aecore/test/aec_eunit_SUITE.erl
velzevur: no :)

apps/aeutils/src/aeu_info.erl
uwiger: I think persistent_term should be used very sparingly, considering the update costs.
> Storing or updating a term (using put/2) is proportional to the number of already created persistent terms because the hash table holding the keys will be copied. In addition, the term itself will be copied.

http://erlang.org/doc/man/persistent_term.html

gorbak25 (approved):

PR #3554, opened 2021-03-17
Fix OAS3 issues

### LINKED ISSUE #3539: "consumes" is not defined in names endpoints in Swagger 2
- /debug/names/claim
- /debug/names/preclaim
- /debug/names/transfer
- /debug/names/revoke
- /debug/names/update 

I assume it should be
```json
        "consumes": [
          "application/json"
        ],
```
as in other endpoints. The lack of this hint breaks the default behavior of `swagger-client`.

velzevur:
This fixes #3539 

While the initial bug report was regarding missing `consumes` section and in `OAS3` [specification](https://swagger.io/specification/) there is no such keyword (there is in `swagger 2`, the old version) - there was an unrelated bug. Addressed in this PR and now JS client can easily be built by:

```
$ curl -X POST \
  https://generator3.swagger.io/api/generate \
  -H 'content-type: application/json' \
  -d '{
  "specURL" : "https://raw.githubusercontent.com/aeternity/aeternity/09e3aa7348f17b66e92719d781345b16199ebd8a/apps/aehttp/priv/oas3.yaml",
  "lang" : "javascript",
  "type" : "CLIENT",
  "codegenVersion" : "V3"
}' -o js.zip
```

@davidyuk please confirm this resolves the issue on your end.

### DIFF
--- a/apps/aehttp/src/aehttp_api_validate.erl
+++ b/apps/aehttp/src/aehttp_api_validate.erl
@@ -127,12 +127,23 @@ populate_param(Param, Req, Validator, EndpointsMod) ->
     Name = proplists:get_value("name", Param),
     case get_param_value(In, Name, Req) of
         {ok, Value, Req1} ->
-            case prepare_param(Param, Value, Name, Validator, EndpointsMod) of
+            case prepare_param(params_priority(Param), Value, Name, Validator, EndpointsMod) of
                 {ok, NewName, NewValue} -> {ok, NewName, NewValue, Req1};
                 {error, Reason} -> {error, Reason, Req1}
             end;
         Error -> Error
     end.
+
+params_priority(Rules) ->
+    Priority =
+        fun("default") -> 1;
+           (_) -> 100
+        end,
+    lists:sort(
+        fun({K1, _}, {K2, _}) ->
+            Priority(K1) =< Priority(K2)
+        end,
+        Rules).
 
 prepare_param([], Value, Name, _, _) -> {ok, Name, Value};
 prepare_param([ Rule | Rules ], Value, Name, Validator, EndpointsMod) ->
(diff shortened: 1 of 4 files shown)


apps/aehttp/src/aehttp_api_validate.erl
```diff
@@ -192,6 +201,7 @@ prepare_param_({"type", "float"}, Value, Name, _, _) ->
 prepare_param_({"type", "integer"}, Value, Name, _, _) ->
     try {ok, to_int(Value)}
     catch
+        error:function_clause  -> 1 = {domat, Name, Value};
```
gorbak25, line 204: wut? This will never match

velzevur: oops, one of my debug messages leaked out :D

gorbak25 (dismissed):

apps/aehttp/src/aehttp_api_validate.erl
```diff
@@ -127,13 +127,24 @@ populate_param(Param, Req, Validator, EndpointsMod) ->
     Name = proplists:get_value("name", Param),
     case get_param_value(In, Name, Req) of
         {ok, Value, Req1} ->
-            case prepare_param(Param, Value, Name, Validator, EndpointsMod) of
+            case prepare_param(params_priority(Param), Value, Name, Validator, EndpointsMod) of
```
velzevur, line 130: It is an implicit requirement that `default` goes first as otherwise it gets ignored :)

davidyuk:
I have built this PR locally (`node_revision: "09e3aa7348f17b66e92719d781345b16199ebd8a" `) and as I see, this does not resolve the issue on my end:  `consumes` still is not present in the definitions of names endpoints in Swagger 2 definition (`/v2/api`).

velzevur:
@davidyuk why do you use `/v2/`? This is the old `swagger2` API, you should build the SDK against the new `/v3` one instead.

davidyuk:
Well, I'm doing my work step by step and noticed the inconsistency in the old swagger definitions, my issue was about that. It is not a blocker for me, I have a workaround. But fixing this may improve compatibility with other tools. In the end, I will use the latest definitions (v3), but I can't confirm that this PR fixes the issue I reported (#3539).

uwiger (approved):

velzevur:
@davidyuk Is this working for you?

davidyuk:
> @davidyuk Is this working for you?

I have checked, it works. Thank you!

gorbak25 (approved):

PR #3571, opened 2021-03-31
Check the signature of inner tx in PayingForTx

### LINKED ISSUE #3570: Tx pool: paying for inner transaction with invalid signatures
How `PayingForTx` works:
 Let's have `Alice` and `Bob`. `Alice` produces a transaction, signs it using a special network id - `<network_id>-inner_tx` and gives this transaction to Bob. The transaction is using a special network id and can not be included in the blockchain alone as the authentication will not match. It must be wrapped by a `PayingFor` transaction and here comes `Bob` - he wraps it and authenticates it. Bob's authentication is based on `Alice`'s transaction (*the signed one*).

There is an inconsistency: if Bob's authentication does not match - the transaction does not enter the transaction pool; if `Alice`'s transaction is erroneously authenticated - we allow it in the transaction pool. The latter one seems like a bug.

hanssv:
Fixes #3570 

### DIFF
--- a/apps/aecore/src/aec_paying_for_tx.erl
+++ b/apps/aecore/src/aec_paying_for_tx.erl
@@ -122,9 +122,17 @@ check(#paying_for_tx{}, Trees, _Env) ->
     %% Checks in process/3
     {ok, Trees}.
 
--spec signers(tx(), aec_trees:trees()) -> {ok, [aec_keys:pubkey()]}.
-signers(#paying_for_tx{} = Tx, _) ->
-    {ok, [payer_pubkey(Tx)]}.
+-spec signers(tx(), aec_trees:trees()) ->
+    {ok, [aec_keys:pubkey()]} | {error, inner_tx_incorrectly_signed}.
+signers(#paying_for_tx{tx = STx} = Tx, Trees) ->
+    %% PayingForTx was introduced in IRIS, so pass that as protocol version -
+    %% since signers don't get protocol version argument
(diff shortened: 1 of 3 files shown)


apps/aecore/src/aec_paying_for_tx.erl
```diff
@@ -122,9 +122,15 @@ check(#paying_for_tx{}, Trees, _Env) ->
     %% Checks in process/3
     {ok, Trees}.
 
--spec signers(tx(), aec_trees:trees()) -> {ok, [aec_keys:pubkey()]}.
-signers(#paying_for_tx{} = Tx, _) ->
-    {ok, [payer_pubkey(Tx)]}.
+-spec signers(tx(), aec_trees:trees()) ->
+    {ok, [aec_keys:pubkey()]} | {error, inner_tx_incorrectly_signed}.
+signers(#paying_for_tx{tx = STx} = Tx, Trees) ->
+    case aetx_sign:verify(STx, Trees, ?IRIS_PROTOCOL_VSN, <<"-inner_tx">>) of
```
velzevur, line 128: I am a bit concerned with the hardcoded protocol version here.

hanssv: I added a soothing comment :-)

velzevur (approved):

uwiger (approved):

velzevur (approved):

PR #3581, opened 2021-04-07
Proper ordering in FATE opcodes

cytadela8:
- where maps:to_list is used, make sure behaviour stays consistent by using `lists:keysort(1, maps:to_list(M))` instead
- introduce new sorting for MAPS_TO_LIST opcode from IRIS
- test MAPS_TO_LIST works as expected
- change ordering of comparison opcodes to `aeb_fate_data:lt` ordering


radrow (approved):
Looks good...

apps/aecontract/src/aect_contracts_store.erl
```diff
@@ -100,6 +100,7 @@ find_keys({Key, Val, Iter}, Map) ->
 serialize_for_client(Store) ->
     maps:from_list(
         [ {aeser_api_encoder:encode(contract_store_key, Key),
-           aeser_api_encoder:encode(contract_store_value, Val)} || {Key, Val} <- maps:to_list(contents(Store)) ]
+           aeser_api_encoder:encode(contract_store_value, Val)}
+          || {Key, Val} <- lists:keysort(1, maps:to_list(contents(Store))) ]
```
gorbak25, line 104: +1

apps/aefate/src/aefa_stores.erl
```diff
@@ -190,7 +190,7 @@ store_map_to_list(Pubkey, MapId, #store{cache = Cache} = S) ->
     ?METADATA(RawId, _, _) = get_map_meta(MapId, Meta),
     Subtree = aect_contracts_store:subtree(map_data_key(RawId), Store),
     {[ {aeb_fate_encoding:deserialize(K), aeb_fate_encoding:deserialize(V)}
-      || {K, V} <- maps:to_list(Subtree) ], S1}.
+      || {K, V} <- lists:keysort(1,maps:to_list(Subtree)) ], S1}.
```
uwiger, line 193: Why not instead use a single `map_to_sorted_list/1` helper function?

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -600,18 +602,45 @@ map_to_list(Arg0, Arg1, EngineState) ->
...
+        true ->
+            [aeb_fate_data:make_tuple(KV)
+             || KV <- lists:sort(fun ({K1,_}, {K2,_}) -> aeb_fate_data:lt(K1, K2) end, List)]
+    end,
+    aeb_fate_data:make_list(Tuples).
+
+pre_iris_map_ordering() ->
+    case persistent_term:get(?PRE_IRIS_MAP_ORDERING, notloaded) of
+        notloaded ->
+            MapOrdering = aec_fork_block_settings:pre_iris_map_ordering(),
```
uwiger, line 638: I think it would be better to load the ordering at the same time as e.g. the whitelist is loaded.

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -600,18 +602,45 @@ map_to_list(Arg0, Arg1, EngineState) ->
...
             aefa_engine_state:spend_gas_for_new_cells(Size * 2, ES3)
     end.
 
+map_to_sorted_list(Map, ES) ->
+    List = maps:to_list(?FATE_MAP_VALUE(Map)),
+    ConsensusVersion = aefa_engine_state:consensus_version(ES),
+    Tuples = if
+        ConsensusVersion < ?IRIS_PROTOCOL_VSN ->
+            MapOrdering = pre_iris_map_ordering(),
+            case maps:get(?FATE_MAP_VALUE(Map), MapOrdering, default) of
```
gorbak25, line 623: +1

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -600,18 +602,45 @@ map_to_list(Arg0, Arg1, EngineState) ->
...
+                     || KV <- lists:keysort(1, List)];
+                Ordering -> Ordering
+            end;
+        true ->
+            [aeb_fate_data:make_tuple(KV)
+             || KV <- lists:sort(fun ({K1,_}, {K2,_}) -> aeb_fate_data:lt(K1, K2) end, List)]
+    end,
+    aeb_fate_data:make_list(Tuples).
+
+pre_iris_map_ordering() ->
```
gorbak25, line 635: Yes, load this when the node boots

data/aecore/.pre_iris_map_ordering.json
```diff
@@ -0,0 +1 @@
...QXFtcdb4iVTE4GkY1PwCufAKBwlqApp3MHrxuIYF1dzA+eVA9cXsIkl8H2Ea/lNxPQrm+IC8L8lb/2b8ArnwCg200wfdMxaDIHr6zGzksxpPsc4z9+ttsNkswY0gKSejhviB4X2oKWCZ/AK58AoF3y/lu2m5taXUzE1BkOnNub2cLMkS9GmQphdrByW6atb4kHmQPV4xXMv8ArnwCg+wS+G5e2Pza5YlhDClmsZPW3rJilGKej5BZO2WXonh1viI0NyCYZEf/AK58AoDcqX4r/baqNOVPH9LwZVEg8RMI8LWWvkGcSkRLC44lsb4kaumJViIGrj8ArnwCga70kniODSWgEdku6G+zSai0QUX8dr/Xg5QJhIZPoMAVviCAONg+2YxfAK58AoDmIQi/GfK0GdIUq2QyH7YWRYzsLnSbMXhaTTwhpf6egb4kCfI91khDXH8ArnwCgY2DHTLwYbB3hF15JoK/tSnoXLKAAf9G9uMISndEZjV9viGlASOAjp/fAK58AoHfvzlUnwiDZEE/XAOFhQGVdD+EuLr49nqiNCEWzEHkZb4gJ34iLwu1vwCufAKCOBfyLVzw0CU6SfHH2Q0Y/MilnGTQDxI/a6/4qrAMFY2+IDm1qsZju38ArnwCgukfcIwgrKFfRryJN90CoMaRoRU06BezbGRzjtONmtmpvhynGkPj8H8ArnwCgLktmTaMGLCOVJf4bDcMWoU/3VU/9Y72hHpy9dw3C6qNviQEa22kC9AL/wCufAKDVsYnGnzUieZjy278QrlJWqghFT9B3nsiAYZlxX1HCA2+HAnpfX7sXwCufAKCRD9DatoKgUsAAaIxE1LNq/uDE+a45lZEg6+BDVV/ZzG+IZck13v1nl8ArnwCgJyGEGMrHpYRToZZ3ezoxrB7A+06uI0zIkK0sdoPBxchviAlnBGTRdpfAK58AoIc6RD34okq7QSUJ3fkAD+I5tjWN4VNLIFV0Zdk0CKGGb4hUlRN9ZN7XwCufAKBGIHzPXgoDL9sj1vIIj8s7VF47uw1ljg5x0v3y2eiNFm+JJG5y4A76oYfAK58AoAxVSDWXiL1W2FsopwMN5CV/f4Yefw2ZwuZTQUhqAy/tb4kTEE4baxUHJ8ArnwCgAEsIMLLx7tNrm/sBZ/EgQvQ17ZNuhUm7wZoAUOiITqtviAWEwgZuzX/AJL04wg=="}
```
gorbak25, line 1: not_a_backdoor.exe
Can you remove the newlines?

gorbak25 (changes requested):
I will deploy it and see whether it syncs with mainnet

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -600,18 +602,45 @@ map_to_list(Arg0, Arg1, EngineState) ->
...
+            MapOrdering = pre_iris_map_ordering(),
+            case maps:get(?FATE_MAP_VALUE(Map), MapOrdering, default) of
+                default ->
+                    [aeb_fate_data:make_tuple(KV)
+                     || KV <- lists:keysort(1, List)];
+                Ordering -> Ordering
+            end;
+        true ->
+            [aeb_fate_data:make_tuple(KV)
+             || KV <- lists:sort(fun ({K1,_}, {K2,_}) -> aeb_fate_data:lt(K1, K2) end, List)]
```
hanssv, line 631: We should update the comparison operation to use `aeb_fate_data:lt/2` also in IRIS.

apps/aefate/src/aefa_stores.erl
hanssv: The result of this function is immediately turned back into a map with `maps:from_list` so we could refactor and let it return a map.

apps/aevm/src/aevm_ae_primops.erl
```diff
@@ -831,7 +831,7 @@ map_call_tolist(_Gas, Data, State) ->
     [MapId] = get_args([word], Data),
     {KeyType, ValType} = aevm_eeevm_maps:map_type(MapId, State),
     {ok, Map} = aevm_eeevm_maps:get_flat_map(MapId, State),
-    List = maps:to_list(Map),
+    List = lists:keysort(1, maps:to_list(Map)),
```
hanssv, line 834: Theoretically AEVM is still active, you can post new contracts until the IRIS protocol upgrade, so maybe we should have the same/a similar mechanism with a file containing mappings here?

hanssv (commented):
It is not height-limited in any way, so this require a synchronized update (or hoping that we don't get more conflicts during the update period) of nodes.

Looks good (once the helper function is in place!).

apps/aefate/src/aefa_stores.erl
cytadela8: Yeah, this could use more refactoring....

cytadela8:
Ok, so there are 2 open questions:
- consider returning to undefined behaviour for before IRIS (so we avoid a fork if not everybody switches to this fix and a big map_to_list call is performed)
- look into aevm

apps/aefate/src/aefa_stores.erl
cytadela8: Also, here we operate on map bytes => bytes. So the FATE ordering doesn't apply. This ordering doesn't rely also on any internal representation as out serialization is well defined.

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -614,18 +617,40 @@ map_to_list(Arg0, Arg1, EngineState) ->
...
 
+map_to_sorted_list(Map, ES) ->
+    List = maps:to_list(?FATE_MAP_VALUE(Map)),
+    ConsensusVersion = aefa_engine_state:consensus_version(ES),
+    Tuples = if
+        ConsensusVersion < ?IRIS_PROTOCOL_VSN ->
+            MapOrdering = persistent_term:get(?PRE_IRIS_MAP_ORDERING),
+            case maps:get(?FATE_MAP_VALUE(Map), MapOrdering, default) of
+                default ->
+                    [aeb_fate_data:make_tuple(KV)
```
gorbak25, line 640: revisit after IRIS

gorbak25 (approved):

hanssv (dismissed):

cytadela8:
I need to investigate lima_tests there seems to be a minor bug somewhere.

hanssv (dismissed):

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -2618,6 +2658,16 @@ comp(egt, A, B) -> A >= B;
 comp( eq, A, B) -> A =:= B;
 comp(neq, A, B) -> A =/= B.
 
+comp_iris( lt, A, B) -> aeb_fate_data:lt(A, B);
+comp_iris( gt, A, B) -> aeb_fate_data:lt(B, A);
+% equal or less than <==> not greater than
+comp_iris(elt, A, B) -> not aeb_fate_data:lt(B, A);
```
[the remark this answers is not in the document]
cytadela8: Preferably all of those should be implemented in aeb_fate_data. Equality could be also optimized...

`aeb_fate_data:elt/2` makes an assumption that two equal fate terms have matching internal representation.

cytadela8:
Updated with aebytecode master (incl. revert of changes to string & bytes ordering)

cytadela8:
lima_tests were missing loading of the ordering data (201f732)

hanssv (approved):

gorbak25 (approved):

PR #3592, opened 2021-04-26
Gas model adjustment for FATEv2/Iris

### LINKED ISSUE #3579: non_payable for contracts when the account pre-exists
When a contract is created, it is created at a predictable address. This is by design, and there are use-cases where this is useful. However, there is a bug in the handling of `ContractCreateTx`, namely  that if the account pre-exists and the contract is `non_payable` we fail to set this flag for the account. This should be fixed for `Iris`.

Also, while fixing that, the flag handling in `aec_accounts` is pretty bad - it started as a single flag and then became a half-assed general thing for a single flag - do this properly!

hanssv:
Improve gas model in FATEv2/Iris, lots of minor fixes and improvements. Mainly we try harder to charge up front for gas, in order to make it possible to stop early when we are about to run out of gas.

The PR also adds more strict type checks for some FATE operations, again to fail earlier and in a more controlled way. 

The PR also fixes #3579 


cytadela8 (dismissed):

radrow (approved):
Looks decent

gorbak25 (changes requested):
...

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -1095,7 +1095,7 @@ make_fate_tx(Aetx, GAMetas, PayFor, ES) ->
              make_some(Res)}
     end.
 
-make_fate_base_tx(BaseTxType, BaseTx) ->
+make_fate_base_tx(BaseTxType, BaseTx, ES) ->
```
radrow, line 1098: For what?

cytadela8: Yes, seems to not be needed. (Note make_fate_tx is available only from FATE_2)

cytadela8 (approved):

cytadela8 (dismissed):

radrow (dismissed):

apps/aefate/src/aefa_fate_op.erl
hanssv: It is used now, missed an argument to abort in the first version.

radrow (approved):

cytadela8 (approved):

cytadela8:
Please wait for @gorbak25 review

gorbak25 (approved):

PR #3597, opened 2021-04-28
Improve developer workflow

lrascao:
Just some niceties i found missing banging my head around while standing up a node


Makefile
```diff
@@ -421,6 +436,9 @@ internal-build: VERSION REVISION internal-compile-deps endpoints
 internal-start:
 	@./_build/$(KIND)/$(CORE) start
 
+internal-console:
+	@./_build/$(KIND)/$(CORE) console
```
lrascao, line 440: maybe more people prefer running interactive console better

config/dev1/sys.config
```diff
@@ -36,8 +36,8 @@
   {aecore, [
       {sync_port, 3015},
 
-      {peers, [<<"aenode://pp_23YdvfRPQ1b1AMWmkKZUGk2cQLqygQp55FzDWZSEUicPjhxtp5@localhost:3025">>,
```
lrascao, line 39: these hardcoded addresses did not match the generated-on-startup peer keys, so nodes weren't able to connect

rebar.config
```diff
@@ -200,7 +200,46 @@
...
+                            {vm_args, "./config/dev2/vm.args"},
+                            {overlay, [{mkdir, "{{output_dir}}/data"},
+                                       {copy,
+                                        "_build/dev2/bin/check_config",
+                                        "{{output_dir}}/bin/check_config"},
+                                       {copy,
+                                       "_build/dev2/lib/aeutils/priv/aeternity_config_schema.json",
+                                       "{{output_dir}}/data/aeternity_config_schema.json"},
+                                       {copy,
+                                       "config/dev2/keys/{peer_key,peer_key.pub}",
```
lrascao, line 223: relx supports this copy format `{X,Y,Z}`, might be worth using in other copy sections above and reduce the verbosity

config/dev1/sys.config
lrascao: i've noticed now that these are set at `aecore_suite_utils` so they shouldn't be changed, it's probably the `peer_key` and `peer_key.pub` that need to contain these keypairs

lrascao:
dev{1-3} keypairs extracted from `aecore_suite_utils` and encrypted to disk with `peer_password`

cytadela8 (approved):
Works and I like it :smiley:

PR #3623, opened 2021-05-14
Fix transaction API etag response header

### LINKED ISSUE #3622: cache invalidation for transaction endpoint doesn't seem to work as expected
sending the same request over and over doesn't invalidate the cache as would be expected, see attached video

https://www.youtube.com/watch?v=K94G3AcxCNc

dincho:
fixes #3622

### DIFF
--- a/apps/aehttp/src/aehttp_cache_etag.erl
+++ b/apps/aehttp/src/aehttp_cache_etag.erl
@@ -59,12 +59,10 @@ generate('GetGenerationByHeight', Req, State) ->
     handle_generation_height(Req, State);
 
 generate('GetTransactionByHash', Req, State) ->
-    Hash = cowboy_req:binding(hash, Req),
-    {{strong, Hash}, Req, State};
+    handle_transaction_hash(Req, State);
 
 generate('GetTransactionInfoByHash', Req, State) ->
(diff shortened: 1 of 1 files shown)


apps/aehttp/src/aehttp_cache_etag.erl
```diff
@@ -157,6 +170,19 @@ etag_generation_hash(Hash) ->
             undefined
     end.
 
+-spec etag_transaction_hash(binary()) -> {ok, binary()} | undefined.
+etag_transaction_hash(Hash) ->
+    case aec_chain:find_tx_with_location(Hash) of
+        {mempool, _Tx} ->
+            {ok, aeu_hex:bin_to_hex(Hash)};
+        {BlockHash, _Tx} when is_binary(BlockHash) ->
```
hanssv, line 178: What about micro-forks (or forks...)

hanssv (dismissed):

apps/aehttp/src/aehttp_cache_etag.erl
hanssv: Nah, that is fine

velzevur (dismissed):

hanssv (approved):

velzevur (approved):

PR #3628, opened 2021-05-17
New HTTP endpoint: dry-run

### LINKED ISSUE #3624: Make DryRunTxs a part of external API
In comparison with other debugging-related internal endpoints, the dry-run is heavily used in production primarily for static calls of smart contracts. From the side of JS SDK, this seems to be the one internal endpoint that is used in production, adding it to external API would allow improving user experience:
- by omitting internal URL in SDK initialization (and related questions about the reasoning of internal URL)
- simplifying running a fully-functional node because won't be needed to setup a reverse proxy in some cases (see #3616)
- simplify custom node setup for Superhero wallet and Base aepp (currently apps accepts only one URL that is used for both internal and external apis, so reverse proxy is required)

I understand the concerns about possible abusing of this endpoint, maybe its availability in external endpoints may be configurable in settings? (with a reasonable default value)

cc: @marc0olo @nikita-fuchs

velzevur:
Fixes #3624

Introduce a new endpoint in the `external` interface. It is to slowly replace
the `internal` debug one. I've also made `aehttp_dryrun_SUITE` aware of the
two specs (oas3 and swagger2).

### DIFF
--- a/apps/aeutils/priv/aeternity_config_schema.json
+++ b/apps/aeutils/priv/aeternity_config_schema.json
@@ -306,6 +306,11 @@
                             "description" : "Number of acceptors in external pool",
                             "type" : "integer",
                             "default" : 10
+                        },
+                        "gas_limit" : {
+                            "description" : "Maximum gas allowed to be available to a single dry-run call.",
+                            "type" : "integer",
+                            "default" : 1000000
                         }
                     }
                 },
@@ -366,6 +371,10 @@
                     },
                     "obsolete" : {
                         "description" : "Old endpoints that will be removed",
+                        "type" : "boolean"
+                    },
+                    "dry-run" : {
+                        "description" : "External protected dry run endpoint",
                         "type" : "boolean"
                     }
                 },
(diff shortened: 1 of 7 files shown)


apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -367,6 +372,10 @@
                     "obsolete" : {
                         "description" : "Old endpoints that will be removed",
                         "type" : "boolean"
+                    },
+                    "dry-run" : {
```
dincho, line 376: so:
```yaml
http:
  endpoints:
    dry-run: true
```
 should enable the new endpoint without disabling the other endpoints ?
I'm asking because I never used this configuration path and the schema does not say what's it's type? Is it object or list ? What's the default.

dincho: Shouldn't we also set the defaults here ? I see there is no defaults of the old endpoints, but it's never late to fix it :)

velzevur: yes, this is how you enable it.

uwiger (approved):

dincho (approved):

mspaansen:
Ok

PR #3630, opened 2021-05-18
rebar.config tidying up

rustaem:
Self explanatory

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -135,7 +135,7 @@
 			aeminer, aecore, aehttp, enacl, enoise,
 			aebytecode, aeserialization, aevm, aechannel,
 			aefate, aemon, aestratum, ecrecover,
-			aesophia_aci_encoder, aehyperchains
+			aesophia_aci_encoder
 		]
 	},
 	{sys_config, "./config/sys.config"},
--- a/rebar.config
+++ b/rebar.config
@@ -20,234 +20,276 @@
 
 {project_app_dirs, ["apps/*"]}.
 
-{erl_opts, [debug_info, {parse_transform, lager_transform},
-            {lager_extra_sinks, [epoch_mining,
-                                 epoch_metrics,
-                                 epoch_sync,
-                                 aestratum]}]}.
+{erl_opts, [
+	debug_info, {parse_transform, lager_transform},
+	{lager_extra_sinks, [
+		epoch_mining,
+		epoch_metrics,
+		epoch_sync,
+		aestratum
+	]}
+]}.
 
 %% NOTE: When possible deps are referenced by Git ref to ensure consistency between builds.
 {deps, [
-        %% off-the-shelf depdendencies
+	%% off-the-shelf depdendencies
+	{enoise, {git, "https://github.com/aeternity/enoise.git", {ref, "1e6ee67"}}},
+	{gproc, "0.6.1"},
+	{jobs, "0.9.0"},
+	{exometer_core, "1.5.7"},
+	{yamerl, "0.7.0"},
+	{eper, "0.99.1"},
+	{ephemeral, "2.0.4"},
+	{sext, "1.8.0"},
 
-        {enoise, {git, "https://github.com/aeternity/enoise.git",
-                 {ref, "1e6ee67"}}},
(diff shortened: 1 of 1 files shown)


cytadela8:
@rustaem `===> Application needed for release not found: aehyperchains`

rebar.config
```diff
@@ -20,234 +20,276 @@
...
+		{ref, "66e5a51"}}},
+
+	{mnesia_leveled, {git, "https://github.com/aeternity/mnesia_leveled.git",
+		{ref, "11c480b"}}},
+
+	{aeminer, {git, "https://github.com/aeternity/aeminer.git",
+		{ref, "0a82f0f"}}},
+
+	{aebytecode, {git, "https://github.com/aeternity/aebytecode.git",
+		{ref, "951db9f"}}},
```
cytadela8, line 72: You should not change any refs!

rebar.config
```diff
@@ -20,234 +20,276 @@
...
+	{trace_runner, {git, "git://github.com/uwiger/trace_runner.git",
+		{ref, "69dc388"}}},
+
+	% for hyperchains and the testing suite
+	{aesophia_aci_encoder, {git, "https://github.com/aeternity/aesophia_aci_encoder",
+		{ref, "41d0427"}}},
+
+	% For automatic building of the StakingContract.json
+	% Do NOT use 'tag' or 'ref' clause here as the source code of the contract is subject to change.
+	{hyperchains_contract, {git, "git://github.com/aeternity/hyperchains-contract.git",
```
cytadela8, line 112: This shoudn't be on master

cytadela8 (changes requested):

rebar.config
cytadela8: (accidently :P)

cytadela8 (commented):
It looks good now. I would like to leave it to others if we want to introduce those style changes.

cytadela8 (approved):

hanssv (commented):
Unless rebar3 starts to require TABs in the config I don't think we should have them...

rebar.config
```diff
@@ -20,234 +20,271 @@
...
-
-        % upstream is not maintained anymore
-        {sha3, {git, "https://github.com/aeternity/erlang-sha3",
-               {ref, "b5f27a2"}}},
-
-        % for lightweight trace-based debugging
-        {trace_runner, {git, "git://github.com/uwiger/trace_runner.git",
-                        {ref, "69dc388"}}},
+	%% off-the-shelf depdendencies
+	{enoise, {git, "https://github.com/aeternity/enoise.git", {ref, "1e6ee67"}}},
```
hanssv, line 36: If tidying up, this one isn't off-the shelf 😅

hanssv (changes requested):
Looks tidy, but no tabs please.

rustaem:
> Unless rebar3 starts to require TABs in the config I don't think we should have them...

You know it's one stroke in any IDE, right?.. :) Changed anyway.

rebar.config
rustaem: I wouldn't want to cut any pre-existing comments -- there is a reason for their existence probably. Are you certain it must go?

hanssv: No, I mean the dependency should be moved further down, to a group where it belongs.

hanssv:
> > Unless rebar3 starts to require TABs in the config I don't think we should have them...
> 
> You know it's one stroke in any IDE, right?.. :) Changed anyway.

Yes, and any sane editor will replace that key-stroke with the right amount of white space.

rustaem:
> > > Unless rebar3 starts to require TABs in the config I don't think we should have them...
> > 
> > 
> > You know it's one stroke in any IDE, right?.. :) Changed anyway.
> 
> Yes, and any sane editor will replace that key-stroke with the right amount of white space.

No, I mean it's a single key-stroke to change tabs to spaces or vv :)
And the dep is moved.

hanssv (approved):

cytadela8 (approved):

cytadela8 (approved):

radrow (approved):

PR #3631, opened 2021-05-19
Hc 3851

rustaem:
Refactoring & dialyzer coverage for aehc_consensus_hyperchains (mostly)

PR #3637, opened 2021-05-25
#3856 hc and core refactoring

rustaem:
Addressing #3681.

HC and supporting parts of core refactoring

Cross-apps records properly encapsulated
Coverage
Other refactoring and fixes


apps/aecore/src/aec_chain_state.erl
```diff
@@ -1149,7 +1170,7 @@ maybe_pogf(Node, [Sibling|T]) ->
 % don't award her with mining reward but we lock the excess of coins on the
 % next granting of fees. This way we can compute properly the locked amount.
 calc_rewards(FraudStatus1, FraudStatus2, GenerationFees,
-             K2MineReward, K1FraudReward, IsKey1Genesis) ->
+    K2MineReward, K1FraudReward, IsKey1Genesis) ->
```
cytadela8, line 1173: Wasn't the indentation nicer here before? Overall the indentation is better, but some reduction in indentation may be confusing.

apps/aecore/src/aec_block_insertion.erl
```diff
@@ -228,25 +275,35 @@ build_insertion_ctx_prev(Node, PrevKeyNode) ->
...
+-spec build_insertion_ctx_check_prev_height(
+    chain_node(), chain_node(), chain_node()) ->
+    {ok, chain_node(), chain_node()} | {error, atom()}.
+build_insertion_ctx_check_prev_height(Node, PrevNode, PrevKeyNode) ->
+    case aec_chain_node:type(Node) of
+        key ->
+            build_insertion_ctx_check_prev_height_key(Node, PrevNode, PrevKeyNode);
+        micro ->
+            build_insertion_ctx_check_prev_height_micro(Node, PrevNode, PrevKeyNode)
+    end.
```
cna0, line 298: How about moving this `case` to the `build_insertion_ctx_check_prev_height` clause pattern-matching?
```erlang
build_insertion_ctx_check_prev_height(key, ...) -> ...
build_insertion_ctx_check_prev_height(micro, ...) -> ...
```

apps/aecore/src/aec_block_insertion.erl
```diff
@@ -228,25 +275,35 @@ build_insertion_ctx_prev(Node, PrevKeyNode) ->
...
                     end
             end
    end.
 
-build_insertion_ctx_check_prev_height(#node{type = key} = Node, PrevNode, PrevKeyNode) ->
+-spec build_insertion_ctx_check_prev_height(
+    chain_node(), chain_node(), chain_node()) ->
+    {ok, chain_node(), chain_node()} | {error, atom()}.
+build_insertion_ctx_check_prev_height(Node, PrevNode, PrevKeyNode) ->
+    case aec_chain_node:type(Node) of
```
cna0, line 293: Is it not simpler to work directly with the `#node{}` record? I think now we have to use `aec_chain_node` module everywhere introducing some more `case ... of`

rustaem: > Is it not simpler to work directly with the `#node{}` record? I think now we have to use `aec_chain_node` module everywhere introducing some more `case ... of`

Sure it's easier. Like Dark Side is easier than Light Side, and eating sugar for energy is easier than exercising and sleeping well :)

apps/aehyperchains/src/aehc_app.erl
```diff
@@ -33,9 +38,9 @@ stop(_State) ->
 check_env() ->
     case aehc_utils:hc_enabled() of
         true ->
-            lager:info("Hyperchains are enabled");
+            linfo("Hyperchains are enabled");
         false ->
-            lager:info("Hyperchains are disabled"),
+            linfo("Hyperchains are disabled"),
```
cna0, line 43: I'm very fond of consistency in the code. Now I'm a bit confused – what should we use for logging from now on, `lager` or `aehc_log`?

cytadela8:
I'm fine with `aec_chain_node:type` instead of accessing the record directly, but I also don't like adding those small helpers like "linfo" or `-type chain_node() :: aec_chain_node:chain_node().` I think inlining such things makes the code more readable the same applies to #3631

apps/aecore/src/aec_block_insertion.erl
cna0: Haha. I see your point, but having a top-level module does not always reduce complexity but adds more. This could introduce a tiny bit of complexity for working with the Node record. Just an opinion; your call.

apps/aehyperchains/src/aehc_app.erl
cytadela8: We should use lager

apps/aecore/src/aec_block_insertion.erl
cytadela8: I think we should use `aec_chain_node:type` this is consistant with how things are done in other parts.

apps/aehyperchains/src/aehc_app.erl
rustaem: > I'm very fond of consistency in the code. Now I'm a bit confused – what should we use for logging from now on, `lager` or `aehc_log`?

My opinion is we should abstract such things. Lager? logger? io:format? whatever?.. Those changes shouldn't be a concern for a higher level code. And abstracting is the only way to move them out of concern. Again, just my opinion here.

cna0:
Generally, we are all fond of code readability and consistency, our opinions may only differ, Here is an opinion: I see the reason behind adding helpers for third-party libraries, but I think in the long run that might force us into having helpers for all sorts of libraries.

What did you use for code formatting? We could use it as part of the development process if the guys agree.

Overall, this pull-request looks good.

apps/aecore/src/aec_block_insertion.erl
rustaem: > Haha. I see your point, but having a top-level module does not always reduce complexity but adds more. This could introduce a tiny bit of complexity for working with the Node record. Just an opinion; your call.

Actually, it reduces the overall complexity, and sometimes greatly so. Imagine someone from the core team add a field in `chain_node` or changes the order of fields. Even better, imagine something like that in a hot-swap scenario. It's scary, been there.

So basically we don't put records in .hrl and we definitely don't do cross-app matching. It's ok for a quick drafting sometimes, but never in a prod.

rustaem:
> I think in the long run that might force us into having helpers for all sorts of libraries.

Which is good. Ideally, we shouldn't have more than, say, 5 lines per function, and one can only achieve that by factoring everything into most simple pieces.

> What did you use for code formatting? We could use it as part of the development process if the guys agree.

IntelliJ IDEA mostly, I really don't like rebar3_format. Not saying it's bad, it's just IDEA is so much easier for those things.

apps/aecore/src/aec_block_insertion.erl
cna0: @rustaem good points!

apps/aehyperchains/src/aehc_log.erl
```diff
@@ -0,0 +1,35 @@
+%%% -*- erlang-indent-level: 4 -*-
```
cytadela8, line 1: This module should not be created. Those helpers only introduce confusion.

cytadela8 (commented):
Please remove some helper as they make the code unnecesserly complex. aehc_log should go and inline trivial types (like `-type chain_node() :: aec_chain_node:chain_node().`)

apps/aehyperchains/src/aehc_consensus_hyperchains.erl
```diff
@@ -497,16 +486,19 @@ state_pre_transform_micro_node(_Node, _PrevNode, _PrevKeyNode, Trees) -> Trees.
...
 %% -------------------------------------------------------------------
 %% PoGF
-pogf_detected(_H1, _H2) -> ok. %% TODO: we can't punish for forks due to forking on the parent chain - inspect the key headers at the given height
+pogf_detected(_H1, _H2) ->
+    ok. %% TODO: we can't punish for forks due to forking on the parent chain - inspect the key headers at the given height
 
 %% -------------------------------------------------------------------
 %% Genesis block
+-spec genesis_transform_trees(trees(), map()) -> trees() | no_return().
+-dialyzer({nowarn_function, genesis_transform_trees/2}).
```
cytadela8, line 501: Hey, this is cheating :P Could you maybe add a comment why this dialyzer error can't be resolved?

apps/aehyperchains/src/aehc_consensus_hyperchains.erl
```diff
@@ -595,133 +596,136 @@ new_unmined_key_node(PrevNode, PrevKeyNode, Height, Miner, Beneficiary, Protocol
             end
     end.
 
+%%TODO Finish the BL here.
+-dialyzer({nowarn_function, new_pos_key_node/8}).
```
cytadela8, line 600: Here also please comment.

apps/aecore/src/aec_block_insertion.erl
erlmachinedev: I also agree to encapsulate protocol related stuff. Function call can be useful in mocking etc..

erlmachinedev (approved):

erlmachinedev:
There is a huge amount of great work! Besides that it's worth to mention few design issues which would be nice to solve in the future:

1. **aec_block_insertion.erl** was introduced as the "optimization" layer  which keeps a context of blockchain interaction. This makes some complexity (even take into account that fact that optimization impact is tremendous).  It could be nice to cover it by appropriate **behavioral API** and decouple like switchable part
2. **Mnesia bypass** functionality should be triggered to remove in the future
3. **aec_consensus behaviour** is redundant (don’t tell Gregozh about :)) and could be reduced into the more compact form
4. To organize and structure test bunches into something more manageable (at some point we discussed simulation test engine which can cover the correct resources allocation and perform garbage collection of blockchain related artifacts)

apps/aecore/src/aec_chain_state.erl
```diff
@@ -341,17 +346,16 @@ persist_state(OldState, NewState) ->
             true
     end.
 
-db_write_top_block_node(#node{ header = Header
-                             , hash   = Hash }) ->
+-spec db_write_top_block_node(chain_node()) -> ok | no_return().
+db_write_top_block_node(N) ->
```
cytadela8, line 350: Consider moving this function to aec_chain_node module

apps/aehyperchains/include/aehc_types.hrl
```diff
@@ -0,0 +1,37 @@
...
+
+-define(COMMITER_PUB_BYTES, 32).
+-define(POGF_HASH_BYTES, 32).
+-define(COMMITMENT_HASH_BYTES, 32).
+
+-type commiter_pubkey() :: <<_:(?COMMITER_PUB_BYTES*8)>>.
+-type pogf_hash() :: <<_:(?POGF_HASH_BYTES*8)>>.
+-type commitment_hash() :: <<_:(?COMMITMENT_HASH_BYTES*8)>>.
+
+-type pubkey() :: aec_keys:pubkey().
```
cytadela8, line 25: I don't really like putting things here that have nothing todo with hc. Especially non trivial definition like signature() or query()

apps/aehyperchains/src/aehc_commitment.erl
```diff
@@ -40,16 +39,14 @@ has_pogf(#hc_commitment{pogf = no_pogf}) -> false;
 has_pogf(#hc_commitment{pogf = _}) -> true.
 
 -spec pogf(commitment()) -> aehc_pogf:pogf().
```
cytadela8, line 41: returns `aehc_pogf:maybe_pogf()`

apps/aehyperchains/src/aehc_consensus_hyperchains.erl
```diff
@@ -1039,9 +1062,77 @@ protocol_staking_contract_call(Trees0, TxEnv, Query) ->
...
+
+-spec encode_contract_pubkey(pubkey()) -> binary().
+encode_contract_pubkey(Pubkey) -> aeser_api_encoder:encode(contract_pubkey, Pubkey).
+
+-spec encode_account_pubkey(pubkey()) -> binary().
+encode_account_pubkey(Pubkey) -> aeser_api_encoder:encode(account_pubkey, Pubkey).
+
+%% Typing here is not perfect, we need to have known_type() and payload()
+%% exported from aeser_api_encoder
+-spec decode(binary()) -> {atom(), binary()}.
```
cytadela8, line 1125: This is same case as `ldebug` helper. Remove this helper as it decreases readability.

apps/aehyperchains/src/aehc_consensus_hyperchains.erl
```diff
@@ -1039,9 +1062,77 @@ protocol_staking_contract_call(Trees0, TxEnv, Query) ->
...
+apply_fallback(Function, Args) -> apply(fallback_consensus(), Function, Args).
+
+-spec jsx_decode(jsx:json_text()) -> map() | no_return().
+jsx_decode(Data) ->
+    case jsx:decode(Data, [{return_maps, true}, {labels, binary}]) of
+        Map when is_map(Map) -> Map;
+        [Map] when is_map(Map) -> Map
+    end.
+
+-spec encode_contract_pubkey(pubkey()) -> binary().
```
cytadela8, line 1117: Remove this helper as below.

apps/aehyperchains/src/connectors/aehc_aeternity_connector.erl
```diff
@@ -78,6 +78,11 @@ handle_info(_Info, State) ->
 terminate(_Reason, _State) ->
     ok.
 
+
+%% Implementation
+
+%%TODO Fix the situation with this stub.
```
cytadela8, line 84: What's the situation, here?

rebar.config
```diff
@@ -304,7 +304,7 @@
 {dialyzer, [
             {warnings, [unknown]},
             {plt_apps, all_deps},
-            {base_plt_apps, [erts, kernel, stdlib, crypto, mnesia, trace_runner, rocksdb, mnesia_rocksdb]}
+            {base_plt_apps, [erts, kernel, stdlib, crypto, mnesia, trace_runner, rocksdb, mnesia_rocksdb, aesophia_aci_encoder]}
```
cytadela8, line 307: OK, we can do this for now, but let's actually fix the typing in aesophia_aci_encoder later. Ok?

cytadela8 (commented):
Still not a fan of removing vertical alignment, but we can live without it.

cna0 (approved):

apps/aecore/src/aec_chain_state.erl
rustaem: I think I've addressed most of the issues except this one. Please elaborate, why would we want to move any DB functionality into a data type module? I may be wrong, but is there any perf improvements to be achieved here, and if any, would it be worth loosing the abstraction purity?

rebar.config
rustaem: We'll fix it definitely, but just out of this scope.

cytadela8 (commented):
Overall good, but tests are failing.

cna0 (approved):

seanhinde:
Superseded by new HC design

PR #3639, opened 2021-05-26
iris hardfork height attempt 1

velzevur:
This PR sets the `iris` hardfork to happen at block ~441450~ 441444.

Keyblock 434272 was mined on 2021-05-26 10:35:49 UTC

15 days * 24 hours * 20 keyblocks = 7200 keyblocks.

This means we expect roughly `iris` to happen on 10th June, ~8:30~ 9:11am UTC.

**Reviewers, please double check this.**

Please note that this is to produce a RC4. If any blocker are found, the value
can be still changed before producing the actual `v6.0.0` release.


velzevur:
My calculations were a bit off so I readjusted the height as well.

uwiger (approved):

radrow (approved):

PR #3647, opened 2021-06-01
Upgrade Hyperchains to Iris

cna0:
This PR addresses the issue #3676, switching hyperchains from Lima to Iris.

### DIFF
--- a/apps/aehyperchains/src/aehc_consensus_hyperchains.erl
+++ b/apps/aehyperchains/src/aehc_consensus_hyperchains.erl
@@ -23,9 +23,9 @@
 
 -define(STAKING_CONTRACT, {?MODULE, staking_contract}).
 -define(STAKING_CONTRACT_ADDR, {?MODULE, staking_contract_addr}).
-%% Lima or Iris as we need the FATE VM at genesis
+%% Iris as we need the FATE VM at genesis
 %% In case that's unwanted then start up another consensus before hyperchains
--define(HC_GENESIS_VERSION, ?LIMA_PROTOCOL_VSN).
+-define(HC_GENESIS_VERSION, aec_hard_forks:protocol_vsn(iris)).
 
 %% Magic nonces
 -define(NONCE_HC_ENABLED, 16#ffffffffffffffff - 1).
@@ -112,8 +112,9 @@
         , deserialize_pos_pow_field/1
         ]).
 
+-include_lib("aecontract/include/aecontract.hrl").
 -include_lib("aecontract/include/hard_forks.hrl").
--include("../../aecore/include/blocks.hrl").
+-include_lib("aecore/include/blocks.hrl").
 -include_lib("aeminer/include/aeminer.hrl").
 
 -record(activation_criteria, {
@@ -507,7 +508,7 @@ pogf_detected(_H1, _H2) -> ok. %% TODO: we can't punish for forks due to forking
 
 %% -------------------------------------------------------------------
 %% Genesis block
-genesis_transform_trees(Trees0, #{}) ->
+genesis_transform_trees(Trees, #{}) ->
     %% At genesis no ordinary user could possibly deploy the contract
     case get_predeploy_address() of
         {ok, Address} ->
@@ -520,12 +521,7 @@ genesis_transform_trees(Trees0, #{}) ->
             TxEnv = genesis_tx_env(),
             %% We don't need to check the protocol version against the block
             %% The insertion of the genesis block bypasses the version check
-            Trees1 = case hc_genesis_version() of  %% Call a function here to make dialyzer happy
-                ?LIMA_PROTOCOL_VSN -> aec_block_fork:apply_lima(Trees0, TxEnv);
-                ?IRIS_PROTOCOL_VSN -> Trees0; %% No special changes
-                _ -> aec_consensus:config_assertion_failed("Hyperchains from genesis require at least LIMA at genesis", "", [])
-            end,
-            deploy_staking_contract_by_system(Trees1, TxEnv)
+            deploy_staking_contract_by_system(Trees, TxEnv)
     end.
 genesis_raw_header() ->
     aec_headers:new_key_header(
@@ -810,8 +806,8 @@ get_hc_activation_criteria() ->
 
 %% Deploys the staking contract using a free system account :)
 %% TODO: For now just hardcode the settings
--define(VM_VERSION, 5).
--define(ABI_VERSION, 3).
+-define(VM_VERSION, ?VM_FATE_SOPHIA_2).
+-define(ABI_VERSION, ?ABI_FATE_SOPHIA_1).
 -define(DEPOSIT_DELAY, 5).
 -define(STAKE_RETRACTION_DELAY, 5).
 -define(WITHDRAW_DELAY, 10).
@@ -904,15 +900,7 @@ verify_existing_staking_contract(Address, Trees, TxEnv) ->
                         {value, Contract, OnchainCode} ->
                             OnchainAbi = aect_contracts:abi_version(Contract),
                             OnchainVm = aect_contracts:vm_version(Contract),
-
-                            %% TODO: Remove the stripping for IRIS...
-                            Code0 = aeser_contract_code:deserialize(get_staking_contract_bytecode()),
-                            FateCode0  = aeb_fate_code:deserialize(maps:get(byte_code, Code0)),
-                            FateCode1 = aeb_fate_code:strip_init_function(FateCode0),
-                            Bytecode = aeb_fate_code:serialize(FateCode1),
-                            LocalCode = aeser_contract_code:serialize(Code0#{ byte_code => Bytecode
-                                                                            , compiler_version => <<"unknown">>
-                                                                            }),
+                            LocalCode = get_staking_contract_bytecode(),
 
                             if  OnchainAbi /= ?ABI_VERSION -> ErrF("Wrong ABI version");
                                 OnchainVm /= ?VM_VERSION -> ErrF("Wrong VM version");
@@ -1043,5 +1031,3 @@ protocol_staking_contract_call(Trees0, TxEnv, Query) ->
 %% TODO: customize
 fallback_consensus() ->
     aec_consensus_bitcoin_ng.
-
-hc_genesis_version() -> ?HC_GENESIS_VERSION.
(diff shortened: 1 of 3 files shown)


apps/aehyperchains/src/aehc_consensus_hyperchains.erl
```diff
@@ -905,14 +905,12 @@ verify_existing_staking_contract(Address, Trees, TxEnv) ->
...
-                            Code0 = aeser_contract_code:deserialize(get_staking_contract_bytecode()),
-                            FateCode0  = aeb_fate_code:deserialize(maps:get(byte_code, Code0)),
-                            FateCode1 = aeb_fate_code:strip_init_function(FateCode0),
-                            Bytecode = aeb_fate_code:serialize(FateCode1),
-                            LocalCode = aeser_contract_code:serialize(Code0#{ byte_code => Bytecode
-                                                                            , compiler_version => <<"unknown">>
-                                                                            }),
+                            Code = aeser_contract_code:deserialize(get_staking_contract_bytecode()),
+                            FateCode  = aeb_fate_code:deserialize(maps:get(byte_code, Code)),
+                            Bytecode = aeb_fate_code:serialize(FateCode),
```
cytadela8, line 910: This is actually redundant. If you look closely you should see that:
```
maps:get(byte_code, Code) = Bytecode
```
Only think that may still be needed is setting the compiler version to "unknown"

cytadela8 (changes requested):
You have also tests failing. They are passing on hyperchains-consensus so either:
1. You didn't build on the current version of hyperchains-consensus
2. You forgot to modify something in the tests that is relevant (probably change `?VM_FATE_SOPHIA_1 -> ?VM_FATE_SOPHIA_2`)

apps/aehyperchains/src/aehc_consensus_hyperchains.erl
cytadela8: (deserialized just to serialize back is a no-op (or to be more precise an irrelevant assert on format))

cna0: Thanks for pointing out.

cytadela8 (approved):

cytadela8 (approved):

apps/aehyperchains/src/aehc_consensus_hyperchains.erl
```diff
@@ -112,8 +112,9 @@
         , deserialize_pos_pow_field/1
         ]).
 
+-include_lib("aecontract/include/aecontract.hrl").
 -include_lib("aecontract/include/hard_forks.hrl").
--include("../../aecore/include/blocks.hrl").
+-include_lib("aecore/include/blocks.hrl").
```
erlmachinedev, line 117: Hyperchains is a dependency of aecore in mono-repo app structure:

- aecore 
  - aehyperchains
  - ...

When we link aecore as "include_lib" we make circular dependency inside (take a look into another included apps, they don't use it).

erlmachinedev (commented):
I think **aec_hard_forks** module should contain the new configuration related to the Iris protocol switch:

Take a look here and be sure that ?IRIS_PROTOCOL_VSN is listed here:

``` protocols_from_network_id ```

File: apps/aehyperchains/src/aehc_consensus_hyperchains.erl [hunk 3]
```diff
@@ -520,12 +521,7 @@ genesis_transform_trees(Trees0, #{}) ->
             TxEnv = genesis_tx_env(),
             %% We don't need to check the protocol version against the block
             %% The insertion of the genesis block bypasses the version check
-            Trees1 = case hc_genesis_version() of  %% Call a function here to make dialyzer happy
-                ?LIMA_PROTOCOL_VSN -> aec_block_fork:apply_lima(Trees0, TxEnv);
-                ?IRIS_PROTOCOL_VSN -> Trees0; %% No special changes
-                _ -> aec_consensus:config_assertion_failed("Hyperchains from genesis require at least LIMA at genesis", "", [])
-            end,
-            deploy_staking_contract_by_system(Trees1, TxEnv)
+            deploy_staking_contract_by_system(Trees, TxEnv)
```
rustaem, line 524: So we can remove one of those dia cheats finally.

apps/aehyperchains/src/aehc_consensus_hyperchains.erl
cytadela8: @cna0 Please respond, I don't have enough understanding of erlang dependencies to judge here.
@erlmachinedev What exactly would break if such a circular dependency is introduced?

cytadela8 (commented):
> I think **aec_hard_forks** module should contain the new configuration related to the Iris protocol switch:
> 
> Take a look here and be sure that ?IRIS_PROTOCOL_VSN is listed here for ae_hc_devnet:
> 
> `aec_hard_forks:protocols_from_network_id/1`

Yes! line 110 in aec_hard_forks. That line defines protocol version at block 0 of hyperchains devnet.

erlmachinedev:
> > I think **aec_hard_forks** module should contain the new configuration related to the Iris protocol switch:
> > Take a look here and be sure that ?IRIS_PROTOCOL_VSN is listed here for ae_hc_devnet:
> > `aec_hard_forks:protocols_from_network_id/1`
> 
> Yes! line 110 in aec_hard_forks. That line defines protocol version at block 0 of hyperchains devnet.

@cytadela8 Hi, sure. Circular dependency means that software which manages dependency tree may get into recursion on attempt to solve relationship between modules.

I think @cna0 may just conform the current  statement which already takes place in other deps like aetx, aeprimop, etc.

```erlang
include("../../aecore/include/aec_hash.hrl").
```

cytadela8:
> > > I think **aec_hard_forks** module should contain the new configuration related to the Iris protocol switch:
> > > Take a look here and be sure that ?IRIS_PROTOCOL_VSN is listed here for ae_hc_devnet:
> > > `aec_hard_forks:protocols_from_network_id/1`
> > 
> > 
> > Yes! line 110 in aec_hard_forks. That line defines protocol version at block 0 of hyperchains devnet.
> 
> @cytadela8 Hi, sure. Circular dependency means that software which manages dependency tree may get into recursion on attempt to solve relationship between modules.
> 
> I think @cna0 may just conform the current statement which already takes place in other deps like aetx, aeprimop, etc.
> 
> ```erlang
> include("../../aecore/include/aec_hash.hrl").
> ```

Ok, I thought rebar can handle circular dependencies correctly. I guess it causes problems sometimes...

apps/aehyperchains/src/aehc_consensus_hyperchains.erl
cna0: @erlmachinedev @cytadela8 as fas I as I know, erlang takes care of circular deps; I don't think `include_lib` or `include` would make a difference regarding that circular deps. Or am I missing something?

cna0:
> > I think @cna0 may just conform the current statement which already takes place in other deps like aetx, aeprimop, etc.
> > ```erlang
> > include("../../aecore/include/aec_hash.hrl").
> > ```
> 
> Ok, I thought rebar can handle circular dependencies correctly. I guess it causes problems sometimes...

Does it not have to do with Erlang's preprocessor? `include` and `include_lib` only include a file before compilation.

cna0:
@cytadela8 @erlmachinedev Shouldn't the tests fail if there was a problem with use of `include_lib`? I think it shouldn't even compile if there was a circular deps. I've never seen circular deps in Erlang, because there's no importing. Or again, I might be missing something.

cytadela8:
> @cytadela8 @erlmachinedev Shouldn't the tests fail if there was a problem with use of `include_lib`? I think it shouldn't even compile if there was a circular deps. I've never seen circular deps in Erlang, because there's no importing. Or again, I might be missing something.

I agree, I will ask somebody from core maintnance.

erlmachinedev:
> obraz
> Let's use include_lib as it seems to work :)

We may may mention Robert Virding at the beginning of this statements [commits](https://github.com/aeternity/aeternity/commits?author=rvirding), and a lot of other names but it's not the case.  

@cytadela8 If this form is correct please **reassign** the PR to make the changes into the other dependency list (we should fix around 20+ deps which should be adjusted accordingly to the Ulf recommendation) 

P.S Somebody (very similar, don't point him) suggested to make map ordering at the beginning of VM creation and what we had in result?


Screen Shot 2021-06-08 at 2 53 46 PM

cna0:
> @cna0 please adjust `aec_hard_forks:protocols_from_network_id/1` as mentioned above and we are going to merge this after refactoring

@cytadela8 I don't understand it. Could you elaborate, please?

cna0:
@cytadela8 I changed the code to use `aec_hard_forks:protocol_vsn(iris)`, if that's what we needed. `protocols_from_network_id` is an internal function which I don't understand how you meant to use it.

cytadela8:
> @cytadela8 I changed the code to use `aec_hard_forks:protocol_vsn(iris)`, if that's what we needed. `protocols_from_network_id` is an internal function which I don't understand how you meant to use it.

Using the define ?IRIS_PROTOCOL_VSN directly was actually fine.

`protocols_from_network_id` is used by the core code to tell what protocol version/hard fork we are on at different block height. The `HC_GENESIS_VERSION` define is only used by some code inside our hyperchain consensus. 

The different network ids signify just that -> different networks. Protocol version changes depending on block height on mainnet and testnet differently. Also we are defining many "local testnets" there and those typically use only two protocol versions and the signifficant one is starting on block 1. For hyperchain testnet (playground) we are using the network id "ae_hc_devnet" please change `protocols_from_network_id` so that this network uses IRIS protocol version.

cytadela8 (approved):

erlmachinedev (approved):
Seems ok

apps/aehyperchains/test/aehc_consensus_tests.erl
```diff
@@ -626,8 +627,8 @@ make_spend_tx(Sender, SenderNonce, Recipient, Amount) ->
                                        payload => <<>>}),
     SpendTx.
 
-staking_contract_vm() -> 5.
-staking_contract_abi() -> 3.
+staking_contract_vm() -> ?VM_FATE_SOPHIA_2.
+staking_contract_abi() -> ?ABI_FATE_SOPHIA_1.
```
rustaem, line 631: This is exactly what I'm working on right now. We should coordinate efforts more.

rustaem (approved):
Yes, include_lib is for in-tree including; include is for grabbing anything from anywhere. We should use include_lib anywhere possible.

Looks good and solves some of my own troubles :)

Issue #3656, opened 2021-06-05
update lager


uwiger:
As part of preparing the source for OTP 24 (see issue #3655), we should get closer to the latest version of lager.
We have at least [one unmerged feature](https://github.com/erlang-lager/lager/pull/516): silencing certain debug traces (exometer_core).

PR #3657, opened 2021-06-08
Gh3656 update lager

uwiger:
See issue #3656 

As we have [merged some needed changes](https://github.com/erlang-lager/lager/pull/556) into upstream `lager`, and as we're preparing to run on top of OTP 24, this PR switches to the latest upstream `lager` commit (no published version yet has the necessary functionality).

Some changes to the config were needed, in order to ensure that the log files still end up in the right place. While touching up some of the code, a few lingering references to "epoch" were also removed.


velzevur (approved):

seanhinde (approved):
LGTM

PR #3663, opened 2021-06-16
Don't read the whole subtree unnecessarily

hanssv:


uwiger (approved):
I think that some database reads could be elliminated by making an mp_tree iteration which is aware of the keys requested, but this should be a drastic improvement already.

velzevur (approved):

PR #3673, opened 2021-07-05
Update aeternity/swagger_endpoints dep

velzevur:
Update the dep so other methods than `GET` and `POST` are allowed. This is
required for the endpoint to delete a specific transaction from the mempool
(where we need a `DELETE` method) but probably for other REST APIs as well.

### DIFF
--- a/apps/aehttp/src/aehttp_dispatch_int.erl
+++ b/apps/aehttp/src/aehttp_dispatch_int.erl
@@ -47,7 +47,7 @@ patterns() ->
 -spec forbidden( Mod :: module(), OperationID :: atom() ) -> boolean().
 forbidden(Mod, OpId) ->
     OpSpec = Mod:operation(OpId),
-    [ #{ tags := Tags } | _ ] = maps:values(OpSpec),
+    Tags = maps:get(tags, OpSpec),
     case lists:member(<<"debug">>, Tags) of
         true -> not aehttp_app:enable_internal_debug_endpoints();
         false -> false
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,13 +149,13 @@ references:
     restore_cache:
       key: *build_nix_cache_key
 
-  machine_build_cache_key: &machine_build_cache_key machine-build-cache-v14-{{ .Branch }}-{{ .Revision }}
+  machine_build_cache_key: &machine_build_cache_key machine-build-cache-v15-{{ .Branch }}-{{ .Revision }}
   restore_machine_build_cache: &restore_machine_build_cache
     restore_cache:
       keys:
         - *machine_build_cache_key
-        - machine-build-cache-v14-{{ .Branch }}-
-        - machine-build-cache-v14-
+        - machine-build-cache-v15-{{ .Branch }}-
+        - machine-build-cache-v15-
 
   save_machine_build_cache: &save_machine_build_cache
     save_cache:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,13 +149,13 @@ references:
     restore_cache:
       key: *build_nix_cache_key
 
-  machine_build_cache_key: &machine_build_cache_key machine-build-cache-v15-{{ .Branch }}-{{ .Revision }}
+  machine_build_cache_key: &machine_build_cache_key machine-build-cache-v16-{{ .Branch }}-{{ .Revision }}
   restore_machine_build_cache: &restore_machine_build_cache
     restore_cache:
       keys:
         - *machine_build_cache_key
-        - machine-build-cache-v15-{{ .Branch }}-
-        - machine-build-cache-v15-
+        - machine-build-cache-v16-{{ .Branch }}-
+        - machine-build-cache-v16-
 
   save_machine_build_cache: &save_machine_build_cache
     save_cache:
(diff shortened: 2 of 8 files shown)


velzevur:
Complementary to aeternity/swagger_endpoints#15

uwiger (dismissed):

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -47,7 +47,7 @@ patterns() ->
 -spec forbidden( Mod :: module(), OperationID :: atom() ) -> boolean().
 forbidden(Mod, OpId) ->
     OpSpec = Mod:operation(OpId),
-    [ #{ tags := Tags } | _ ] = maps:values(OpSpec),
```
earlyriser99, line 50: Looks like you can't set debug on individual methods. Unless each OpId will only contain a single method - so you would need a standalone DeleteTx OpId with only the DELETE method. In which case this is guaranteed to be a single list entry so could safely match on a single entry.

velzevur: Basically the combination of URI + method defines a single operation that comes with a unique OperationID. The old structure used to be
```
#{ OpID1 => #{method1 => OpSpec},
   OpID2 => #{method2 => OpSpec2},
   ....
   }```
Since there can be only one method per OperationID, the key-value relation is a bit unnecessary and could be misleading (implying that there could be another method there) so I simplified it, moving the method to the OpSpec instead.

Comment (earlyriser99) [reply apps/aehttp/src/aehttp_dispatch_int.erl:50]: So this could safely be `[ #{ tags := Tags } ] = maps:values(OpSpec)`

To save future readers hunting around for the case there might be multiple methods

Comment (velzevur) [reply apps/aehttp/src/aehttp_dispatch_int.erl:50]: Well not really: the old code was a bit misleading.
The structure used to be:
```
operations() ->
    #{'GetChannelByPubkey' =>
      #{get =>
            #{parameters =>
                  [[{"in","query"}, 
                    {"name","int-as-string"},
                    {"description",
                     "If this flag is set to true, the response will have all integers set as strings"},
                    {"required",false},
                    {"schema",[{"type","boolean"},{"default",false}]}],
                   [{"in","path"},
                    {"name","pubkey"},
                    {"description","The pubkey of the channel"},
                    {"required",true},
                    {"type","string"},
                    {"example",
                     "ch_2tceSwiqxgBcPirX3VYgW3sXgQdJeHjrNWHhLWyfZL7pT4gZF4"}]],
              path => <<"/v3/channels/{pubkey}">>,
              responses =>   
                  #{200 => #{<<"$ref">> => <<"/components/schemas/Channel">>},
                    400 => #{<<"$ref">> => <<"/components/schemas/Error">>},
                    404 => #{<<"$ref">> => <<"/components/schemas/Error">>}},
              tags => [<<"external">>,<<"channel">>]}},

....
```
where `method1` and `method2` were placeholders for `get` or `post`. Calling ` Mod:operation/1` with a Operation ID (for example with `GetChannelByPubkey`) would return the whole nested map with one key (being `get` from the example above). This was named as `OpSpec` above and thus the call to `maps:values/1` to retrieve the actual OpSpec :) Now the same structure would look like this:

```
```
operations() ->
    #{'GetChannelByPubkey' =>
            #{parameters =>
                  [[{"in","query"}, 
                    {"name","int-as-string"},
                    {"description",
                     "If this flag is set to true, the response will have all integers set as strings"},
                    {"required",false},
                    {"schema",[{"type","boolean"},{"default",false}]}],
                   [{"in","path"},
                    {"name","pubkey"},
                    {"description","The pubkey of the channel"},
                    {"required",true},
                    {"type","string"},
                    {"example",
                     "ch_2tceSwiqxgBcPirX3VYgW3sXgQdJeHjrNWHhLWyfZL7pT4gZF4"}]],
              path => <<"/v3/channels/{pubkey}">>,
              responses =>   
                  #{200 => #{<<"$ref">> => <<"/components/schemas/Channel">>},
                    400 => #{<<"$ref">> => <<"/components/schemas/Error">>},
                    404 => #{<<"$ref">> => <<"/components/schemas/Error">>}},
              tags => [<<"external">>,<<"channel">>],
              method => "get"},

....
```
...and the ` Mod:operation/1` would return the actual OpSpec :) Since the method is only one, it is moved to the OpSpec itself.

Calling `maps:values/1` on the OpSpec would make little sense...

earlyriser99: My bad - looking at it backwards

earlyriser99 (dismissed):
Looks good now I understand it

earlyriser99 (dismissed):
LGTM

zxq9:
What exactly is messed up with the CircleCI checks?
Lima fails, but only because of a missing path `_build/tests/reports`.
The smoke check seems more significant. Are the tests just a bit broken? Do we usually fix that in a second PR?
```Test test_peer_discovery in aest_peers_SUITE failed!
{{badmatch,#{get =>```

uwiger (approved):

zxq9 (approved):

velzevur:
@zxq9 there is a CI cache that holds _usually_ immutable stuff, like deps...

earlyriser99 (dismissed):
LGTM

dincho:
Not sure if it is related, but any GET endpoint should also support HEAD

uwiger (approved):

earlyriser99 (approved):

PR #3667, opened 2021-06-19
Fix Oas3 typos in ProtectedDryRunTxs and GetPendingAccountTransactionsByPubkey

davidyuk:
fixes partially #3658

PR #3694, opened 2021-07-26
fix oas3 spec and add new definitions

marc0olo:
This PR contains changes included in #3667.

When working on https://github.com/kryptokrauts/aepp-sdk-java/issues/174 I noticed that there are certain attributes in the response definition that don't rely on a specific schema definition. This results in the generation of randomly named classes in the Java codegen. For the array of `SignedTx` this is quite annoying because it needs to be used in multiple places of our code.

I hope we can merge this. I am open for changing the name of the schema definitions if necessary.


velzevur (approved):

uwiger (approved):

marc0olo:
@velzevur can we merge this? do I need to rebase?

PR #3700, opened 2021-08-16
Fix system tests

velzevur:


velzevur:
The new structure of channel create (expecting actual delegates) made a few tests fail while using the old structure.
Moving from `aevm` to `fate` is mandatory as well.

I pushed to branch `system-tests` that runs the STs to check if there is something else missing.

uwiger (approved):

cna0 (approved):

PR #3702, opened 2021-08-17
Gh3545 admin mode app ctrl

uwiger:
This PR enables switching modes - specifically between `normal` and `maintenance` mode.
It introduces [`app_ctrl`](https://github.com/aeternity/app_ctrl) as a new dependency, and a mode configuration in the `aecore.app` file
(also, an API for modifying mode and role configurations at runtime).

Currently, runtime mode switching is done using `app_ctrl:set_mode(normal | maintenance)` or `app_ctrl:set_and_await_mode(Mode, Timeout)`. See e.g. `aecore_oam_SUITE`.

Starting a node in maintenance mode can be accomplished by setting the user config `system:maintenance_mode` to true (either in the user config file, or by setting the OS env variable `AE__SYSTEM__MAINTENANCE_MODE`.

Note that the PR rearranges some processes and app dependencies, primarily splitting `aecore` and moving the sync functionality into a separate app: `aesync`. Also, rather than having `aecore` depend on other ae apps, like `aechannel`, these apps now depend on `aecore`, which keeps running during maintenance mode.

The `app_ctrl` application starts `gproc` and `setup` very early, during 'protected_mode', before enabling the node's default mode. Among other things, this means that `setup` start phase hooks will run to completion before any of the user apps not specifically included in protected_mode will be allowed to run.


earlyriser99 (approved):
Looks really clean.

PR #3704, opened 2021-08-24
Gh3468 db rollback script

uwiger:
Rebased on PR #3702 

See the `whitelist_and_rollback` test case in `aecore_forking_SUITE.erl`

This PR introduces `argparse` for more structured extension script handling. It also adds a whitelisting script and a db rollback script. The rollback script can check the whitelist (if given the `-w` option), and will then roll back to either the greatest height where the chain and whitelist match, or the height below where they first diverge.


uwiger:
The diff base is PR #3702 for easier review. Once that PR is merged, change this PR to be against `master`.

velzevur (approved):

earlyriser99 (approved):
Nice!

Issue #3303, opened 2020-06-15
Dev mode


dincho:
While this has been discussed multiple times I feel it was lost somewhere in the forums.

Develop mode should enable application developers to run local environment/network with a "fast" non resource intensive mining. Include transactions once they appear. No cuckoo/simulation.

Currently we burn developer laptops for no reason.

uwiger:
> The mining mocks are already present in the codebase - exposing them in the releases won't involve much work.

It might be better not to use `meck` for this, though.

marc0olo:
just saw this due to the PR. I guess I created a duplicate with https://github.com/aeternity/aeternity/issues/3692, right?

@sborrazas do you also want to share your thoughts about that? I am referring to https://github.com/aeternity/ae_mdw/issues/220 and possible also https://github.com/aeternity/aeternity/issues/3693

cc @thepiwo

uwiger:
@marc0olo So far, the intent of the PR (#3708) is to lay a foundation so that specific dev mode features might evolve on its own timeline, more in sync with the dev tools. This is why I decided to favor the plugin route.

sborrazas:
Ideally it would just be a configuration in the node which would allow instant transactions, pretty much what ganache does. It would be useful if there was a docker image that starts this sandboxed/dev mode locally without much configuring.

Further down the line it would also be useful to have a clear public Erlang API on the node (similar to the HTTP API) that would allow using this devmode node from the Erlang running node directly, to avoid having to go through HTTP. This Erlang API would also make it clearer for plugins to know what's internal and what's not.

Related to #3693, it can also be useful (but not as useful as having devmode), to have some sort of DSL that would allow to specify a blockchain state in a dev-friendly way, so as to be able to avoid having to create the transactions for every test ran.

uwiger:
> Ideally it would just be a configuration in the node which would allow instant transactions, pretty much what ganache does. It would be useful if there was a docker image that starts this sandboxed/dev mode locally without much configuring.

Yes, this is pretty much the current idea.

> Further down the line it would also be useful to have a clear public Erlang API on the node (similar to the HTTP API) that would allow using this devmode node from the Erlang running node directly, to avoid having to go through HTTP. This Erlang API would also make it clearer for plugins to know what's internal and what's not.

[...]

PR #3708, opened 2021-08-31
Load plugins, on-demand consensus mode

uwiger:
See #3303 
This initial push adds support for plugins, and introduces an `on_demand` consensus mode.

The `on_demand` consensus mode is basically a copy of the `ct_tests` consensus mode, used for immediate block production in test suites.

A `dev` role for `app_ctrl` has been added. Plugins designed for dev mode can add themselves to this role, by adding this to their `.app` file's environment:

```erlang
    {'$app_ctrl', [
        {modify, [
           {roles, [{dev, [join]}]}
         ]}
```

To add plugins, specify a lib root for plugin applications, using the user config option `system:plugin_path`.
Under this root, erlang applications are expected to reside, as under `ERLANG_LIBS` lib roots.
The config option `system:plugins` (array of strings) lists the names of the plugin applications that are to be loaded.
The detection and loading of plugin apps happens in start phase `102` (`aeu_plugins:load_plugins()`).


uwiger:
Example of starting a node in "dev mode" with a plugin called `aep_debug` (which only exists locally on my machine):

```
AE__FORK_MANAGEMENT__NETWORK_ID="ae_dev" \
  AE__CHAIN__CONSENSUS='{"0":{"name": "on_demand"}}' \
  AE__SYSTEM__PLUGIN_PATH=/home/uwiger/ae/plugins/lib AE__SYSTEM__PLUGINS='["aep_debug"]' \
  _build/prod/rel/aeternity/bin/aeternity console
```

apps/aecore/src/aec_hard_forks.erl
```diff
@@ -141,6 +141,8 @@ protocols_from_network_id(<<"local_iris_testnet">>) ->
      %%, ?LIMA_PROTOCOL_VSN     => Excluded for testing new protocol
      , ?IRIS_PROTOCOL_VSN     => 1
      };
+protocols_from_network_id(<<"ae_dev">>) ->
```
dincho, line 144: What's the purpose of this hardcoded network id ?

apps/aecore/src/aecore_app.erl
```diff
@@ -49,10 +36,13 @@ stop(_State) ->
...
-    Mode = case ok(aeu_env:find_config([<<"system">>, <<"maintenance_mode">>],
-                                       [user_config, schema_default])) of
-               true  -> maintenance;
-               false -> normal
+    MMode = ok(aeu_env:find_config([<<"system">>, <<"maintenance_mode">>],
+                                   [user_config, schema_default])),
+    DevMode = is_dev_mode_consensus(),
+    Mode = case {MMode, DevMode} of
+               {true , _}     -> maintenance;
+               {false, true}  -> dev_mode;
```
dincho, line 44: I'm a bit confused to see this feature as special "mode"

uwiger: The 'on_demand' consensus protocol is special: once activated, it cannot be deactivated. Also, e.g. sync needs to be disabled, since a bunch of integrity checks are disabled. It seems reasonable that the node is aware of this.

apps/aecore/src/aec_hard_forks.erl
nikita-fuchs: it would be nice to be able to provide a custom one at some point.

uwiger: I believe you already can. But you have to specify the consensus properties in detail then. Also, the on-demand consensus mode is really not meant for networking.

earlyriser99 (approved):
LGTM

velzevur (approved):

PR #3714, opened 2021-09-09
Fix failing test in aec_block_micro_candidate_tests

hanssv:


apps/aefate/test/aefa_sophia_test.erl
```diff
@@ -232,6 +232,10 @@ run_call(Code, Fun, Args, Options) ->
             io:format("Store:\n  ~p\n", [Store1]),
             print_logs(EventMap, Logs),
             aefa_engine_state:accumulator(ES);
+        {Time, {revert, Reason, ES}} ->
+            print_run_stats(Time, ES),
+            io:format("Revert: ~ts\n", [Reason]),
```
uwiger, line 237: Is this supposed to remain here?

uwiger (dismissed):

apps/aefate/test/aefa_sophia_test.erl
hanssv: I was doing something else, and this is unrelated but it is a good thing to have.

velzevur (approved):

velzevur (approved):

uwiger (approved):

PR #3722, opened 2021-09-20
Simplified plugin config check + apply_os_env

uwiger:
See issue #3303 
This PR simplifies the config checking for the plugin, and allows the plugin to define an OS env prefix, similar to the `AE` prefix used by the main node. This will allow for parameterization using OS env variables like `DEV__KEYBLOCK_INTERVAL=10`.

### DIFF
--- a/apps/aeutils/src/aeu_env.erl
+++ b/apps/aeutils/src/aeu_env.erl
@@ -585,7 +585,7 @@ lst(E) -> [E].
 update_config(Map) when is_map(Map) ->
     Schema = application:get_env(aeutils, '$schema', #{}),
     ConfigMap = application:get_env(aeutils, '$user_map', #{}),
-    ConfigMap1 = update_config(Map, Schema, ConfigMap, _Notify = true),
+    ConfigMap1 = update_config(Map, ConfigMap, Schema, _Notify = true),
     cache_config(ConfigMap1),
     ok.
 
--- a/apps/aeutils/src/aeu_env.erl
+++ b/apps/aeutils/src/aeu_env.erl
@@ -22,7 +22,8 @@
 -export([find_config/2]).
 -export([nested_map_get/2]).
 -export([read_config/0]).
--export([apply_os_env/0]).
+-export([apply_os_env/0,
+         apply_os_env/3]).
 -export([parse_key_value_string/1]).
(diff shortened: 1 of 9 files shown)


velzevur (dismissed):

marc0olo:
I don't feel confident in reviewing Erlang code. will leave that up to others :)

apps/aeutils/src/aeu_env.erl
```diff
@@ -305,23 +310,34 @@ read_config(Mode) when Mode =:= silent; Mode =:= report ->
 
 apply_os_env() ->
     ok = application:ensure_started(gproc),
-    try
-    %% Make sure gproc is started, since we use it in update_config/1
-    application:ensure_started(gproc),
     Pfx = "AE",  %% TODO: make configurable
```
nikita-fuchs, line 313: shouldn't this be `DEV` or so then ?

apps/aeutils/src/aeu_env.erl
```diff
@@ -559,17 +582,30 @@ lst(L) when is_list(L) -> L;
 lst(E) -> [E].
 
 update_config(Map) when is_map(Map) ->
+    update_config(Map, _Notify = true).
```
nikita-fuchs, line 585: does this generally mean we can update some or all configs also in runtime ?

uwiger: Generally, it sorta does, but also generally, not all code in the system may be prepared for runtime changes of the user config, so your mileage may vary at this time.

apps/aeutils/src/aeu_env.erl
uwiger: This function is called during the normal boot. The support function, called from `aeu_plugins` is flexible in regard to prefix and schema. The thing that is not yet configurable is the os env prefix for overriding the parameters in the main schema.

velzevur (approved):

velzevur (approved):

marc0olo:
any chance to get other reviews here? again, I am probably not the right guy to review Erlang code :D

hanssv:
Ulf promises to fix everything later, so I'm good!

hanssv (approved):

Issue #3749, opened 2021-10-15
Rollback confuses tx_pool, leaves orphaned transactions


uwiger:
When performing a rollback while in dev_mode (with the plugin active), a crash happened in `aec_tx_pool`.
This serves as a reminder that the initial rollback function didn't do anything to the transactions in the blocks being deleted.
The tx_pool crash comes from trying to find the common ancestor where the OldHash has already been deleted.

An idea is that rollback could reuse the logic for fork switches, with only minor adjustments.

```
14:46:33.166 [info] New key block - height: 101
14:47:46.433 [info] Application aeplugin_dev_mode exited with reason: stopped
14:47:46.434 [info] Application aehttp exited with reason: stopped
14:47:46.472 [warning] Could not fetch plugin config object (<<"aeplugin_dev_mode">>)

14:47:54.149 [error] gen_server aec_tx_pool terminated with reason: no match of right hand value {error,unknown_hash} in aec_tx_pool:do_top_change/4 line 759
14:47:54.149 [error] CRASH REPORT Process aec_tx_pool with 0 neighbours crashed with reason: no match of right hand value {error,unknown_hash} in aec_tx_pool:do_top_change/4 line 759
14:47:54.149 [error] Supervisor aecore_sup had child aec_tx_pool started with aec_tx_pool:start_link() at <0.1750.0> exit with reason no match of right hand value {error,unknown_hash} in aec_tx_pool:do_top_change/4 line 759 in context child_terminated
14:47:54.150 [error] gen_server aec_conductor terminated with reason: {{{badmatch,{error,unknown_hash}},[{aec_tx_pool,do_top_change,4,[{file,"/app/apps/aecore/src/aec_tx_pool.erl"},{line,759}]},{aec_tx_pool,handle_call_,3,[{file,"/app/apps/aecore/src/aec_tx_pool.erl"},{line,368}]},{timer,tc,1,[{file,"timer.erl"},{line,166}]},{aec_tx_pool,handle_call,3,[{file,"/app/apps/aecore/src/aec_tx_pool.erl"},{line,357}]},{gen_server,try_handle_call,4,[{file,"gen_server.erl"},{line,661}]},{gen_server,handle_msg,6,[{file,"gen_server.erl"},{line,690}]},{proc_lib,init_p_do_apply,...}]},...} in gen_server:call/3 line 223
[...]

PR #3750, opened 2021-10-18
tx_pool understands rollback; stricter rollback check in forking_SUITE

uwiger:
See issue #3749 
This PR makes the tx_pool check if the chain has retreated, and (1) doesn't crash, (2) treats it the same way as when the top jumps backwards due to sync (i.e. adjusting the tx TTLs for GC).

It also ensures that the location objects of txs in deleted blocks are deleted.

Lastly, the `aecore_forking_SUITE` checks more thoroughly that the rollback performed actually succeeds and results in the expected new top height.

### DIFF
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -367,7 +367,7 @@ new_state_from_persistence() ->
 persist_state(OldState, NewState) ->
     case {get_top_block_hash(OldState), get_top_block_hash(NewState)} of
         {TH, TH} -> false;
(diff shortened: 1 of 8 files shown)


bogdan-manole:
While i'm using rolling back for different tests, at some point I get http 400 error code (bad request). could that be caused by the same issue? At minute 3:23 you will see the error in this screen cast [https://asciinema.org/a/3UEH81AzPpT8sMzGZrZX6RYix]( https://asciinema.org/a/3UEH81AzPpT8sMzGZrZX6RYix)

PR #3763, opened 2021-10-29
Fully remove txs during rollback

uwiger:
See issue #3749 - this is another attempt at solving the issue.

The transactions need to be fully removed. Otherwise, the tx_pool will detect that the tx was present before, and may deny re-entry (configurable).

Also note that if an action should be completely reverted by rollback (e.g. in dev mode), it's not enough to save the current height, perform the action and then roll back to the saved height. If the generation at the saved height is still open, the action may well remain on-chain even after the rollback. Currently, the rollback takes only whole generations.

So, save the current height, then make sure to post a new keyblock to ensure that the reversible actions are posted at `height + 1`.


thepiwo (approved):
haven't checked the code, but this resolves my earlier issues when using the devmode

velzevur (approved):

seanhinde (approved):
Looks pretty comprehensive to me

Issue #3781, opened 2021-11-25
Migrate from lager to logger


uwiger:
## Is your feature request related to a problem? Please describe.

As of OTP 21, `logger` should be the preferred logging tool, and `lager` is no longer as actively maintained.

PR #3782, opened 2021-11-25
Redirect logger logging to lager

uwiger:
See issue #3781
This commit moves log level adjustment to an earlier start phase,
moves the adjustment code to aeutils, and also installs a logger
handler that redirects logger log output to lager, with the same
metadata as that captured by the lager_transform.


seanhinde (approved):
Happy to see rebar.config get tidier :)

hanssv (approved):

PR #3784, opened 2021-11-26
Fix two HTTP API bugs: missing CORS and crash in names

### LINKED ISSUE #3732: Missing CORS headers for errors on /v3/names endpoint
A followup to #3192 . There are still endpoints that return no CORS headers on errors. The example I stumbled upon is this: 

https://mainnet.aeternity.io/v3/names/%21.chain

I already checked with @dincho and it seems to be an issue with the node itself, not the infrastructure setup.

velzevur:
Fixes #3732.

There are two different bugs:
1. CORS not set when a request crashes
2. There is a crashing endpoint :)

### DIFF
--- a/apps/aehttp/priv/oas3.yaml
+++ b/apps/aehttp/priv/oas3.yaml
@@ -1986,6 +1986,28 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  "/debug/crash":
+    get:
+      tags:
+        - internal
+        - debug
+      operationId: GetCrashRequest
+      description: This is a sample URI to simulate a request crash. Shall be used only for test purposes only
+      parameters:
+        - $ref: '#/components/parameters/intAsString'
+      responses:
+        "200":
+          description: This will never happen
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/TokenSupply"
+        "500":
+          description: We always end up here
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
   "/node/operator/mempool/hash/{hash}":
     delete:
       tags:
--- a/apps/aehttp/priv/oas3.yaml
+++ b/apps/aehttp/priv/oas3.yaml
@@ -1992,7 +1992,7 @@ paths:
         - internal
         - debug
       operationId: GetCrashRequest
-      description: This is a sample URI to simulate a request crash. Shall be used only for test purposes only
+      description: This is a sample URI to simulate a request crash. Shall be used for test purposes only
       parameters:
         - $ref: '#/components/parameters/intAsString'
       responses:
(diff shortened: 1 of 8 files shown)


apps/aens/src/aens.erl
```diff
@@ -85,11 +85,14 @@ get_name_hash(Name) when is_binary(Name) ->
 is_name(Bin) ->
     length(aens_utils:name_parts(Bin)) > 1.
 
+-spec name_to_name_hash(binary()) -> {ok, binary()} | {error, atom()}.
```
velzevur, line 88: Later on there is the expectation that this function returns either an ok tuple or an `{error, atom()}` one. In rare cases (names invalid according to IDNA) the function was returning a nested tuple instead. This is point 2 - the crash.

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -68,7 +68,10 @@ run(Queue, F) ->
         error:{rejected, _} ->
             {503, [], #{reason => <<"Temporary overload">>}};
         error:timeout ->
-            {503, [], #{reason => <<"Not yet started">>}}
+            {503, [], #{reason => <<"Not yet started">>}};
+        Class:Reason:Stacktrace ->
```
velzevur, line 72: If we don't catch the crash - it propagates all the way to `cowboy_rest` where it is caught. In this case a standard code 500 response is being returned, sadly it does not have CORS headers. So we handle the error and the response follows the expected path (and has CORS headers added to it).

apps/aehttp/src/aehttp_dispatch_int.erl
```diff
@@ -74,7 +74,10 @@ handle_request(OperationID, Req, Context) ->
         error:{rejected, _} ->
             {503, [], #{reason => <<"Temporary overload">>}};
         error:timeout ->
-            {503, [], #{reason => <<"Not yet started">>}}
+            {503, [], #{reason => <<"Not yet started">>}};
+        Class:Reason:Stacktrace ->
```
velzevur, line 78: Same as in `aehttp_dispatch_ext`

apps/aehttp/priv/oas3.yaml
```diff
@@ -1986,6 +1986,28 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  "/debug/crash":
+    get:
+      tags:
+        - internal
+        - debug
+      operationId: GetCrashRequest
+      description: This is a sample URI to simulate a request crash. Shall be used only for test purposes only
```
hanssv, line 1995: Repeated words

apps/aehttp/priv/swagger.yaml
```diff
@@ -1895,6 +1895,28 @@ paths:
           description: Height not available
           schema:
             $ref: '#/definitions/Error'
+  /debug/crash:
+    get:
+      tags:
+        - internal
+        - debug
+      operationId: GetCrashRequest
+      description: This is a sample URI to simulate a request crash. Shall be used only for test purposes only
```
hanssv, line 1904: Same here

hanssv (approved):

hanssv (approved):

dincho (approved):

PR #3798, opened 2021-12-10
GH-3706: Only count used gas towards micro block limit

### LINKED ISSUE #3706: Tx gas used being used for validations
The transaction gas is a measure of how much computations are required to apply a transaction. A transaction comes with a gas limit. This is the upper boundary of how much gas it can use. The actual gas used is known only after the transaction had been applied.

Currently when validating incoming blocks the gas limit is being used. This means that if a transaction has a big limit but it actually consumes significantly less gas - then the actual computations in the block are less and there is "space" for more transactions. In order this to happen when validating an incoming transaction the actual gas used shall be used instead. This might be a separate mapping in the transaction's list between tx hash and actual gas used.

hanssv:
Fixes #3706 

Protocol update: aeternity/protocol#495

### DIFF
--- a/apps/aecore/src/aec_block_micro_candidate.erl
+++ b/apps/aecore/src/aec_block_micro_candidate.erl
@@ -18,6 +18,7 @@
 -export_type([block_info/0]).
 
 -include("blocks.hrl").
+-include_lib("aecontract/include/hard_forks.hrl").
 
 -opaque block_info() :: #{trees := aec_trees:trees(),
                           txs_tree := aec_txs_trees:txs_tree(),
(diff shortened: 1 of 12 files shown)


apps/aecore/src/aec_block_micro_candidate.erl
```diff
@@ -247,10 +256,15 @@ add_txs_to_trees(MaxGas, Trees, [Tx | Txs], Acc, Env) ->
 report_failed_txs([])  -> ok;
 report_failed_txs(Txs) -> aec_tx_pool:failed_txs(Txs).
 
-used_gas(Block, Txs) ->
+used_gas(Block, Txs, Trees) ->
     Version = aec_blocks:version(Block),
     Height = aec_blocks:height(Block),
-    lists:foldl(fun(Tx, Acc) -> aetx:gas_limit(aetx_sign:tx(Tx), Height, Version) + Acc end, 0, Txs).
+    GasFun =
```
uwiger, line 262: Seems like this pattern appears in three places. Candidate for a separate function, perhaps?

uwiger (approved):

apps/aecore/src/aec_block_micro_candidate.erl
hanssv: I think there are subtle differences though? But maybe that was in an earlier stage.

uwiger: Yes, perhaps

velzevur (approved):

PR #3808, opened 2022-01-04
fix: bump xcode image as used one is deprecated

michalbagrowski:


michalbagrowski:
closing because build do not support `/` in branch name

PR #3827, opened 2022-01-20
Update installation.md, add docker commands for windows

nikita-fuchs:


dincho (changes requested):
that should go in the docker docs, this doc is about “native” packages

dincho:
stale

PR #3828, opened 2022-01-20
Add read cache for contract store

hanssv:
Reading from the Merkle-Patricia Tree is rather expensive, reading a full subtree is even more expensive - by adding a read cache the iterators used to fetch the subtree will become (much) more efficient. 

### DIFF
--- a/apps/aecontract/src/aect_contracts_store.erl
+++ b/apps/aecontract/src/aect_contracts_store.erl
@@ -8,6 +8,7 @@
 -module(aect_contracts_store).
 
 -record(store, { cache :: #{key() => val()},
+                 read_cache :: #{key() => val()},
                  mtree :: aeu_mtrees:mtree() }).
 
 -opaque store() :: #store{}.
@@ -17,6 +18,7 @@
 -export([ contents/1,
           size/1,
           get/2,
+          get_w_cache/2,
           mtree/1,
           new/0,
           new/1,
@@ -24,6 +26,7 @@
           put_map/2,
           remove/2,
           subtree/2,
+          subtree_w_cache/2,
           write_cache/1,
           serialize_for_client/1 ]).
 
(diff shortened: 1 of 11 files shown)


hanssv:
Some numbers from a live node to go with the PR (^^) - I’ve run some benchmarks on my node, it iterates through the contract store, and reads the first (lexicographically by address) N contracts. Then it proceeds to read all store maps from those contracts while logging the time. Lots of use of iterators and subtrees:
Before fixes (N = 200):
```
Cache: 0 / 0
200 contracts found in 108100.02 ms
Cache: 0 / 0
Mean Avg time per subtree node: 1126.96 us
Avg time per subtree node: 1043.22 us
Total unfolding time: 61792.5 ms
200 contracts contained 189 store maps with 58998 keys
```
And after fixes:
```
Cache: 1364488 / 87075
200 contracts found in 24177.89 ms
Cache: 752306 / 85363
Mean avg time per subtree node: 650.02 us
Avg time per subtree node: 222.12 us
Total unfolding time: 13317.3 ms
200 contracts contained 189 store maps with 58998 keys
```

apps/aecontract/src/aect_contracts_store.erl
```diff
@@ -41,14 +44,26 @@ new() ->
...
 
 %% Returns empty binary if key is not in the store.
 -spec get(key(), store()) -> val().
-get(Key, #store{ cache = Cache, mtree = Tree }) ->
+get(Key, Store) ->
+    {Val, _NewStore} = get_w_cache(Key, Store),
+    Val.
+
+-spec get_w_cache(key(), store()) -> {val(), store()}.
+get_w_cache(Key, #store{ cache = Cache, read_cache = RCache, mtree = Tree } = S) ->
```
uwiger, line 56: If `get_w_cache/2` were moved to `aeu_mp_trees`, it could help us move towards producing block witnesses (where we should cache every read, and the read cache + write cache would make up the witness).

uwiger (approved):

velzevur (approved):

apps/aecontract/src/aect_contracts_store.erl
hanssv: That is an entirely different beast though :-)

PR #3830, opened 2022-01-20
Permanent storage of whether GC was ever enabled in database

seanhinde:
Should we refuse to start node if someone attempts to start a previously GC's node with GC disabled in the config?

- Allowing the node to start and ignoring the config is kind of confusing, but friendlier
- Refusing to start feels more proper, but less friendly

DECISION - allowing node to start anyway violates principle of least surprise, so refuse to start.

Still needs some test suitery -- ADDED

### DIFF
--- a/apps/aecore/src/aec_chain_state.erl
+++ b/apps/aecore/src/aec_chain_state.erl
@@ -92,6 +92,10 @@
         , ensure_chain_ends/0
         , ensure_key_headers_height_store/0
         , grant_fees/5
+        , is_gc_configured/0
+        , is_gc_enabled/0
+        , was_gc_enabled/0
+        , persist_gc_enabled/0
         ]).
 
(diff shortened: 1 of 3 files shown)


zxq9 (approved):
The only question I have is whether GC will eventually merge with fast sync/sparse node ideas we've been discussing lately. Not having a way to go from GC to non-GC feels a little odd, but definitely is correct in the current situation -- and having a specific error for it is much nicer than having things mysteriously break.

cmullaparthi:
Should we document what the user has to do if they encounter this situation? I'm guessing they have to either turn GC back on or rebuild the node from scratch? The former is easy enough, but the steps required for a rebuild are probably worth spelling out somewhere in the docs?

uwiger:
Closing, as the current GC implementation (most recent merged PR: #4147) persistently notes the last GC scan.

PR #3833, opened 2022-01-21
fix release bundle tag

michalbagrowski:
resolve https://github.com/aeternity/aeternity/issues/3844

### DIFF
--- a/.circleci/config/jobs/@docker.yml
+++ b/.circleci/config/jobs/@docker.yml
@@ -15,7 +15,7 @@ docker-image:
     - setup_docker
     - when:
         condition:
-          matches: { pattern: "^latest-bundle.*$", value: << parameters.tag >> }
+          matches: { pattern: "^latest-bundle$", value: << parameters.tag >> }
         steps:
         - docker_image:
             tag: << parameters.tag >>
--- a/.circleci/config/jobs/@docker.yml
+++ b/.circleci/config/jobs/@docker.yml
@@ -15,6 +15,14 @@ docker-image:
     - setup_docker
     - when:
         condition:
+          matches: { pattern: "^latest-bundle.*$", value: << parameters.tag >> }
+        steps:
+        - docker_image:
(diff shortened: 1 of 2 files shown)


venimus (dismissed):

.circleci/config/jobs/@docker.yml
```diff
@@ -6,6 +6,9 @@ docker-image:
       default: false
     tag:
       type: string
+    src:
```
dincho, line 9: Not sure if you tested this and it works by indirect passing the parameter down to the all job steps, but could you please pass it explicitly to the `docker_image` and then to do `docker_tag` command

venimus (approved):

.circleci/config/jobs/@docker.yml
```diff
@@ -13,6 +13,14 @@ docker-image:
         steps:
           - aeplugin_install_devmode
     - setup_docker
+    - when:
```
dincho, line 16: I'm not sure this is best approach as the following condition will also match

michalbagrowski: so what is your proposition, as my this pipelien understanding it's best place...

michalbagrowski: ok now it should  match when required

venimus (approved):

.circleci/config/jobs/@docker.yml
dincho: What I mean is that when this job is invoked from:
- docker-image-latest (tag: latest) AND
- docker-image-latest-bundle (tag: latest-bundle)

The following step that matches for tag: `^latest.*$` will be run twice, although it would do nothing wrong it's definitely wrong and unwanted behaviour.

michalbagrowski:
closing as @dincho will take over this task.

dincho:
closing in favour of #3856 3856

PR #3843, opened 2022-01-27
Use -1 when name_ttl is not specified on name_update events

sborrazas:
Right now, there's no way to distinguish on the env event if the AENS
update was called using `Some(0)` (update the name to expire
immediately) or `None` (don't update the name TTL).

This implementation now returns name_ttl = 0 when it was supposed to
expire immediately and `name_ttl = -1` when the name_ttl is supposed
to remain untouched.


hanssv:
Isn't there exactly the same issue with pointers (and client_ttl) - where we also sometimes pass `None` and keep the previous value?

sborrazas:
@hanssv what default values do you suggest using for these two cases?

hanssv:
For`client_ttl` I guess `-1` would work. But for pointers I don't know, there are no type correct values that are easily distinguishable 🤔

PR #3851, opened 2022-02-03
Added missing exports for MAC (m1?)

omar-saadoun:
Added missing exports for MAC


ThomasArts:
Good catch, but this is part of building, not installing.
See https://github.com/aeternity/aeternity/blob/master/docs/build.md#macos

Issue #3553, opened 2021-03-16
Refactor rocksdb database plugin for better stability/performance


uwiger:
Related to e.g. #3407, #2974, #3466 and #3088, we have identified the need to refactor the rocksdb database plugin to reduce IO saturation, improve failure recovery and speed.

There are a couple of core issues:
* The current backend plugin creates a separate rocksdb instance for each table (and index, checkpoint, etc.) Each instance will keep its own Write-Ahead Log (WAL). Mnesia also keeps a WAL - one central transaction log. This causes multiplication of write-ahead logging for no benefit, increasing the write pressure, and also driving up the number of open file descriptors.
* Rocksdb can reject writes if the IO system is saturated. Since writes take place beyond Mnesia's point of no return, Mnesia has no ability to handle this, and the write is essentially lost. We have previously tried to catch this at the transaction level and with special error handling in the plugin, but that is a band-aid.

While it's doubtful that we can remove the gap completely, it should be much better to have one central rocksdb instance for all tables, utilizing rocksdb column families instead. This should also allow use of rocksdb transactions for efficient atomic updates, perhaps entirely bypassing mnesia, at least in some cases.

Some support for transitioning from the old db structure will be needed, such that the plugin can detect and use separate instances for some, or all, tables.

uwiger:
Addressed in #3948

PR #3869, opened 2022-03-17
Gh3553 integrate new rocksdb plugin 220317

uwiger:
See issue #3553 

This PR uses the most recent `mnesia_rocksdb` plugin, utilizing `rocksdb` column families for ae tables.
State trees use `{raw, raw}` encoding, since they are sets, and both key and value holds binary data.

A `db_migrate` extension script is added to migrate a legacy database to column familes. For a full db, this should take 1.5 - 2 hrs.


uwiger:
Functionality merged in other PRs. Closing.

Issue #3825, opened 2022-01-18
Include some connected peers as well as verified peers in ping messages


seanhinde:

seanhinde:
Further discussion suggested an enhancement to this PR to allow up to 10% rather than always 10%. PR: https://github.com/aeternity/aeternity/pull/3878

PR #3878, opened 2022-04-08
Allow up to 10% not always 10% connected peers

seanhinde:
Another go at including connected peers in ping messages.

Make it a bit more unpredictable by having a random small subset of connected peers rather than always a small subset

Addresses gh-3825


velzevur (approved):
IMO 10% is a bit small. The connected peers are the ones we really want to propagate. The verified are the ones we are _thinking_ are online but might have saturated connection pools. Or it could be our outgoing pool is saturated and that's why we are not connecting to them.
Having said this - I can only speculate if 10% is small or not. My 2 cents.

velzevur:
Actually it would be so much better making it configurable from the config :)

velzevur (dismissed):

hanssv (approved):
Looks good to me! Should it perhaps be documented somewhere what is a good value? I.e. 5-20% is good, but 80-100% might not be great 🤔

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -1099,6 +1099,13 @@
                     "maximum" : 32,
                     "default" : 32
                 },
+                "ping_connected_peer_percentage" : {
```
uwiger, line 1102: I'm wondering if there should be a sub-group for ping messages? Wouldn't it, for example, also be reasonable to be able to set an upper limit on the actual number of peers included?

dincho, line 1102: I'd go with just `ping_peers_percentage` name :)

mitchelli: I'm adding a subgroup for pings, so I'll change the parameter to peer_percentage.

uwiger (approved):

hanssv (approved):

PR #3890, opened 2022-05-03
Fix `int-as-string` in dry-run

davidyuk:
While working on https://github.com/aeternity/aepp-sdk-js/pull/1233 I found that dry-run endpoint is not fully compatible with `int-to-string` parameter:
- it doesn't accept account amount as string
- it returns an array of results that doesn't have nested fields converted to strings

### DIFF
--- a/apps/aehttp/src/aehttp_api_handler.erl
+++ b/apps/aehttp/src/aehttp_api_handler.erl
@@ -125,10 +125,10 @@ cache_enabled() ->
     aeu_env:user_config_or_env([<<"http">>, <<"cache">>, <<"enabled">>],
                                aehttp, [cache, enabled], ?DEFAULT_HTTP_CACHE_ENABLED).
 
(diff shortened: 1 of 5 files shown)


thepiwo:
it's a breaking change though :/

davidyuk:
But it breaks only in the scope of `int-as-string`. I would consider this more as a bug fix, because `int-as-string` wasn't working properly in those cases.

hanssv:
Release note missing, otherwise I have no issues with it. But I'm not the SemVer fanatic around here 😅

velzevur:
A test case exposing the bug and (and confirming it is fixed) would be really nice to have.

davidyuk:
I've added tests and the release note 🚀

velzevur (approved):
Great work, thanks!

hanssv (approved):

PR #3897, opened 2022-05-30
WIP towards simulated HC parent chains

seanhinde:
Not yet finished, but should pass CI

### DIFF
--- a/apps/aecore/src/aec_parent_connector.erl
+++ b/apps/aecore/src/aec_parent_connector.erl
@@ -8,8 +8,6 @@
 
 -module(aec_parent_connector).
 
-%% FSM states: Not clear what's needed
-%% 
 %% Functionality:
 %% - at intervals check parent chain to understand when new top block is
 %%   available.
@@ -24,7 +22,10 @@
 %%%=============================================================================
 
 %% External API
--export([start_link/0]).
+-export([start_link/0, start_link/3, stop/1]).
+
+%% Use in test only
+-export([trigger_fetch/0]).
 
 %% Callbacks
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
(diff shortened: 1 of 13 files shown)


apps/aecore/src/aec_parent_connector.erl
```diff
@@ -100,7 +123,10 @@ handle_info(check_parent, #state{parent_nodes = ParentNodes,
             %% - Call the smart contract to elect new leader.
             %% - Notify conductor of new status
     end,
-    erlang:send_after(State#state.fetch_interval, self(), check_parent),
+    if is_integer(FetchInterval) ->
```
velzevur, line 126: Is it possible that `FetchInverval` becomes non-integer? If so - does it make sense silently not scanning the parent chain?

apps/aehttp/src/aehttpc_aeternity.erl
```diff
@@ -37,59 +38,75 @@ get_generation(Host, Port, Hash) ->
...
-    {ok, Commitments}.
+            end, [], Txs).
 
-post_commitment_tx(Host, Port, AccountId, Signature, HCAccountId, CurrentTop) ->
-    %% 1. get the next nonce for our account
-    NoncePath = <<"/v2/accounts/", AccountId/binary, "next-nonce">>,
+post_commitment_tx(Host, Port, AccountId, StakerPrivKey, HCAccountId, CurrentTopHash) ->
+    %% 1. get the next nonce for our account over at the parent chain
+    SenderEnc = aeser_api_encoder:encode(account_pubkey, AccountId),
+    NoncePath = <<"/v2/accounts/", SenderEnc/binary, "/next-nonce">>,
```
velzevur, line 73: please use `/v3/` prefix or make the prefix configurable. Our plan is ditching `/v2/`

velzevur (commented):
Looks great

apps/aecore/src/aec_parent_connector.erl
seanhinde: I could probably ifdef(TEST). out the place that could happen. For test purposes it's possible to start the process without timed checks. Then test code can trigger parent chain checks at will, and we can have deterministic testing

velzevur (approved):

Issue #3558, opened 2021-03-20
 all_nodes.orphaned_txs_get_included.on_key_block.spend_on_dev1


velzevur:
```
%%% aecore_forking_SUITE ==> all_nodes.orphaned_txs_get_included.on_key_block.spend_on_dev1: FAILED
%%% aecore_forking_SUITE ==> {{badmatch,
     {ok,[{signed_tx,
              {aetx,spend_tx,aec_spend_tx,86,
                  {spend_tx,
                      {id,account,
                          <<217,202,108,173,192,99,13,10,129,124,71,86,232,
                            121,148,177,243,254,160,88,174,204,22,114,15,42,
                            51,71,75,19,135,16>>},
                      {id,account,
                          <<217,202,108,173,192,99,13,10,129,124,71,86,232,
                            121,148,177,243,254,160,88,174,204,22,114,15,42,
                            51,71,75,19,135,16>>},
                      1,20000000000000,0,1,<<"foo">>}},
              [<<81,167,139,30,9,255,177,19,189,128,101,29,34,247,152,191,
                 171,80,86,215,40,40,16,99,254,2,201,236,42,48,184,90,212,
                 245,75,205,221,98,21,140,43,136,69,74,209,99,124,205,181,99,
                 106,129,157,227,162,96,137,8,5,67,142,119,199,1>>]},
          {signed_tx,
              {aetx,spend_tx,aec_spend_tx,86,
                  {spend_tx,
                      {id,account,
                          <<217,202,108,173,192,99,13,10,129,124,71,86,232,
                            121,148,177,243,254,160,88,174,204,22,114,15,42,
                            51,71,75,19,135,16>>},
                      {id,account,
                          <<217,202,108,173,192,99,13,10,129,124,71,86,232,
                            121,148,177,243,254,160,88,174,204,22,114,15,42,
                            51,71,75,19,135,16>>},
                      1,20000000000000,0,2,<<"foo">>}},
              [<<228,224,36,33,88,98,95,166,236,44,84,246,36,49,54,203,
                 185,204,45,138,51,192,240,4,48,121,215,195,181,38,170,
                 240,97,252,72,16,145,201,64,201,8,246,11,103,18,40,248,
                 205,204,123,0,155,174,209,70,3,0,29,35,47,198,104,127,14>>]}]}},
 [{aecore_forking_SUITE,spend,1,
      [{file,
           "/home/builder/aeternity/apps/aecore/test/aecore_forking_SUITE.erl"},
[...]

PR #3913, opened 2022-06-16
Fix ct forking SUITE

seanhinde:
Fix timing issue with the forking suite

Issue was that occasionally the original SpendTx was still left in the mempool on dev1 after being mined on dev2. Fix is to wait until it is really gone from dev1 before carrying on with test.

This PR is sponsored by the ACF


uwiger (approved):

velzevur (approved):

seanhinde:
Fixes #3558

PR #3935, opened 2022-07-01
Long name fix

zxq9:
Accept more error types than atom() as the second argument for aec_tx_pool_failures:limit/2.
This patch was initiated to fix the case where too-long AENS names are entered into the mempool, but is also more generally useful than the previous strictly atom-only error case.

### DIFF
--- a/apps/aecore/test/aecore_mempool_SUITE.erl
+++ b/apps/aecore/test/aecore_mempool_SUITE.erl
@@ -24,6 +24,7 @@
     skipped_nonce_specific_cleanup/1,
     insufficient_funds_specific_cleanup/1,
     name_claim_to_unknown_commitment_cleanup/1,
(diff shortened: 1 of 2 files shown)


apps/aecore/test/aecore_mempool_SUITE.erl
```diff
@@ -495,11 +498,39 @@ name_claim_to_unknown_commitment_cleanup(Config) ->
             _ -> name_not_preclaimed
         end,
     {ok, 1} = rpc:call(NodeName, aec_tx_pool_failures, limit, [SignedTx, Error]),
-    make_microblock_attempts(1, Config),
```
hanssv, line 498: This change can be undone it was just while debugging, sorry.

uwiger (approved):
LGTM

velzevur (approved):

fabiankrol (approved):

PR #3877, opened 2022-04-08
Gh3553 integrate new rocksdb plugin 220408

uwiger:
See issue #3553 

This PR uses the most recent `mnesia_rocksdb` plugin, utilizing `rocksdb` column families for ae tables.
State trees use `{raw, raw}` encoding, since they are sets, and both key and value holds binary data.

A `db_migrate` extension script is added to migrate a legacy database to column familes. For a full db, this should take 1.5 - 2 hrs.

Update of PR #3869 as part of ongoing debugging

PR #3777, opened 2021-11-17
Gh3553 integrate new rocksdb plugin

uwiger:
Refactor database access to take advantage of the new direct access API in the `mnesia_rocksdb` plugin.

Incorporates the majority of changes proposed in #3610

PR #3948, opened 2022-07-08
Gh3553 integrate new rocksdb plugin + direct api

uwiger:
This PR should replace #3777, #3869, and #3877.

See issue #3553

In this PR, the default mode of the node is to:
* Create the db using rocksdb column families to represent mnesia tables
* Still using the mnesia access API
* Peeling the rocksdb data out of the transaction store and updating as batch (faster, likely safer)
* No longer going via the table-owning gen_server for each access.

It's possible to turn off the optimization at mnesia commit, using the `chain:db_commit_bypass` option (boolean, default: `true`).
It's also possible to enable use of the `mrdb` direct access API, i.e. rocksdb transactions rather than mnesia ones.
This is done with the option `chain:db_direct_access` (boolean, default: `false`). This mode has been tested in a full mainnet sync, and seems to be somewhat faster. There is also a new CT target, `make ct-mnesia-mrdb`, for it. This is automatically run by the `commit` workflow.


hooks/pre_start.sh
```diff
@@ -1,10 +1,16 @@
...
-if [ "$OPEN_FILES_RECOMMENDED" -gt "$OPEN_FILES_LIMIT" ]; then
-    echo "WARNING: ulimit -n is $OPEN_FILES_LIMIT; $OPEN_FILES_RECOMMENDED is the recommended minimum."
+if [ "$OPEN_FILES_RECOMMENDED" -gt "$AE_OPEN_FILES_LIMIT" ]; then
+    echo "WARNING: ulimit -n is $ULIMIT; $OPEN_FILES_RECOMMENDED is the recommended minimum."
     echo "You are recommended to ensure the node is stopped and raise the maximum number of open files before starting the node."
     echo "Visit https://github.com/aeternity/aeternity/wiki/Setting-Number-of-Max-Open-Files for additional pointers."
 fi
+if [ "$ULIMIT" -lt "$AE_OPEN_FILES_LIMIT" ]; then
+    echo "Trying to raise fd limit to $AE_OPEN_FILES_LIMIT"
+    prlimit --pid=$PPID --nofile=$AE_OPEN_FILES_LIMIT
```
dincho, line 15:
```
dincho at Dinchos-Mac-mini in ~
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.15.7
BuildVersion:	19H1824

dincho at Dinchos-Mac-mini in ~
$ prlimit
-bash: prlimit: command not found
```

uwiger: Arguably, we don't even need the check anymore, since I've been syncing against mainnet with `NOFILE=1024`

uwiger: I removed the highlighted part above.

zxq9 (approved):
Yay!

dincho:
Is it backward/forward compatible changing both new options ?

hooks/pre_start.sh
```diff
@@ -1,10 +1,18 @@
...
+export AE_OPEN_FILES_LIMIT
 
-if [ "$OPEN_FILES_RECOMMENDED" -gt "$OPEN_FILES_LIMIT" ]; then
-    echo "WARNING: ulimit -n is $OPEN_FILES_LIMIT; $OPEN_FILES_RECOMMENDED is the recommended minimum."
+if [ "$OPEN_FILES_RECOMMENDED" -gt "$AE_OPEN_FILES_LIMIT" ]; then
+    echo "WARNING: ulimit -n is $ULIMIT; $OPEN_FILES_RECOMMENDED is the recommended minimum."
     echo "You are recommended to ensure the node is stopped and raise the maximum number of open files before starting the node."
     echo "Visit https://github.com/aeternity/aeternity/wiki/Setting-Number-of-Max-Open-Files for additional pointers."
 fi
+# prlimit doesn't automatically work on MaxOSx
```
dincho, line 13: Could you please remove this non portable commented script

uwiger: Done

config/sys.config
```diff
@@ -9,6 +9,12 @@
     ]
   },
 
+  { hut,
+    [
+     {level, debug}
```
dincho, line 14: defaults to debug ? this goes into the release if I'm not mistaken ?

dincho:
I wonder what the migrate script does in details, also is it idempotent ?

This PR/change deserve a nice release notes :)

uwiger:
> Is it backward/forward compatible changing both new options ?

Yes. If `db_direct_access` is enabled, `db_commit_bypass` will be ignored.

uwiger:
> I wonder what the migrate script does in details, also is it idempotent ?

For each 'standalone' table, it creates a corresponding column family, then traverses the former, writing a batch at a time to the CF, then deleting the batch from the standalone. It does this one table at a time.

If run again, it will find no standalone tables (hopefully).

> This PR/change deserve a nice release notes :)

I added release notes, at least. Whether they're nice enough is up to you to decide. ;-)

uwiger:
I've found an issue with the migration code. Working on a fix.

zxq9 (approved):
LGTM

zxq9 (approved):
One of these days my review will actually take. :-)

velzevur (approved):

Issue #3961, opened 2022-07-23
HC: return the validator address in the MainStaiking get_state() endpoint


velzevur:
Currently the MainStaking looks like this:
```
entrypoint get_state() : get_state_response =
  let validators = map_keys(state.online_validators) ++
           map_keys(state.offline_validators)
  { staking_validator_ct = state.staking_validator_ct,
   validators      = [get_validator_state(v) | v <- validators],
   total_stake      = state.total_stake,
   validator_min_stake  = state.validator_min_stake,
   stake_minimum     = state.stake_minimum
   }


entrypoint get_validator_state(v : address) : get_validator_response =
  let validator = get_validator(v)
  let online =
    switch(validator_bucket(v))
      ONLINE => true
      OFFLINE => false
  {ct = validator.ct,
   stake = validator.stake,
   is_online = online,
   state = validator.ct.get_state()}
  ```
  
  There are 2 closely related improvements:
  1. `get_validator_state` endpoint must return the validator address in the structure. Otherwise we don't know who that is :)
  2. `get_state` has all the contract addresses needed by `get_validator_state` so no need for maps lookup: instead of iterating just the map keys, we should rather iterate on map tuples (key, value) and call the `get_validator_state` with those instead

PR #3964, opened 2022-07-25
Add validator address to get_validator_state response

fabiankrol:
ref: `HC: return the validator address in the MainStaiking get_state() endpoint` #3961


velzevur (approved):
Looks great

PR #3976, opened 2022-08-15
add servers to oas3 spec

marc0olo:
contact email and terms of service url needs clarification


marc0olo:
I decided to remove the info. also https://github.com/aeternity/api-docs requires on that servers info right now. I didn't find a better solution and I don't wanna waste more time on this.

marc0olo:
closing as we go a different route now, see: https://github.com/aeternity/api-docs/pull/13

anyway, I think we should remove terms of service and the email for the api team in the spec 🤷‍♂️

PR #3978, opened 2022-08-18
Read custom accounts file when in devmode

nikita-fuchs:
This introduces the behaviour to try reading a custom file with to-be prefilled accounts. It kicks in whenever devmode is enabled and a value for the config `AE__CHAIN__DB_PATH` is passed. (The latter shall become the default behaviour when devmode is started, as the db path shall serve as "work spaces" in devmode. )

To see how it works together with devmode, test this with my latest devmode PR: https://github.com/aeternity/aeplugin_dev_mode/pull/17 It is not strictly necessary though, as these changes affect the node's behaviour only when devmode is active.

Regarding the additional deps: In order to generate the accounts file early enough, I had to utilise [ this ](https://github.com/nikita-fuchs/aeplugin_dev_mode/blob/e9d921b9b51a5b895486b5744fde9bb9de98ba9d/src/aeplugin_dev_mode.app.src#L13) lifecycle hook function, which is ran prior to any deps from `devmode` being loaded. If I missed anything about this, I'm open for suggestions of course.


apps/aecore/src/aec_fork_block_settings.erl
```diff
@@ -244,7 +244,20 @@ read_preset_contracts(?LIMA_PROTOCOL_VSN = Release) ->
...
+        DevmodeAccsFilePath -> DevmodeAccsFilePath
+    end.
+    
+is_devmode_accounts_present() ->
+    Workspace = lists:last(filename:split(os:getenv("AE__CHAIN__DB_PATH"))),
+    FilePath = filename:join(os:getenv("AE__CHAIN__DB_PATH"), "devmode_prefunded_accs_" ++ Workspace ++ ".json"),
+    case filelib:is_regular(FilePath) of
+        true ->  FilePath;
+        false -> false        
+    end.
```
hanssv, line 260: I still think this is backwards, the node should not, I repeat not, have to know about the plugins, etc. What would be an ok solution here is to add a flag `YES_PLEASE_DO_GENERATE_ACCOUNTS` or `YES_PLEASE_USE_THESE_ACCOUNTS`. Where the first one is a boolean and the second one a path... 

Would that make it a lot more complicated?

rebar.config
```diff
@@ -107,7 +107,17 @@
...
+
+        % all four for devmode:
+        {ebip39,           
+                {git, "https://github.com/aeternity/ebip39.git", {ref, "7520340"}}},
+        {eaex10,
+                {git, "https://github.com/aeternity/eaex10.git", {ref, "958046a"}}},
+        {epbkdf2, 
+                {git, "https://github.com/aeternity/epbkdf2", {tag, "v1.0.0"}}},
+        {ec_utils,
+                {git, "https://github.com/hanssv/ec_utils.git", {ref, "e557b1c"}}}
```
hanssv, line 120: Nope, the node should not depend on these 🙀

apps/aecore/src/aec_fork_block_settings.erl
nikita-fuchs: I can make devmode set an `AE_USE_CUSTOM_ACCOUNTS` env var and check for that 🤷‍♂️ ?

rebar.config
nikita-fuchs: I'm not opinionated about where to put them, maybe @uwiger can hint me how to make this available in his `setup` [hook function](https://github.com/nikita-fuchs/aeplugin_dev_mode/blob/e9d921b9b51a5b895486b5744fde9bb9de98ba9d/src/aeplugin_dev_mode.app.src#L13)  ? I tried grasping the workings of `setup` et al., but to my best understanding only the _node's_ dependencies are available at this point. Open to find out what I missed and learn a thing or two.

For curiosity, what's so bad about having those in the node deps after all @hanssv ?

Issue #3871, opened 2022-03-28
node crashing while running javscript client tests


subhod-i:
The client calls either `emit_kb` or `height()` API, The node crashes while running the tests in [aepp-sophia-examples](https://github.com/aeternity/aepp-sophia-examples.git). It's hard to identify for which of the above two APIs the node is crashing.

Client code where the debugger is stopping.
https://github.com/aeternity/aeproject/blob/59bbdedac9e66f6ccb8402f2536349f4ef200b6d/src/lib/utils.js#L98

hack: I was able to fix this by introducing sleep() between the `emit_kb` and `awaitHeight` method.

## Expected Behavior
node able to send an error response if there is an error or otherwise execute as expected

## Actual Behavior
node crashes without sending any response to the client.

## Steps to Reproduce the Problem

  1.  git clone https://github.com/aeternity/aepp-sophia-examples.git
  2. cd aepp-sophia-examples
  3. git checkout fix/tests-and-ci
  4. npm install
  5. npm install -g @aeternity/aeproject
  6. aeproject env
  7. npm run test:exchangeOracles 

## Configuration, logs, error output, etc.

- `aeternity.yaml` configuration file: https://github.com/aeternity/aepp-sophia-examples/blob/fix/tests-and-ci/docker/aeternity.yaml
I have copied the complete log from the beginning. the error logs specifically at the tail.

- `log/aeternity.log` log file -  https://gist.github.com/subhod-i/fae6cfa2a0a1928e4a252d4d1445bb64

## Specifications

  - Virtualization: container

PR #3979, opened 2022-08-18
Gh3871 crash during rollback

uwiger:
See issue #3871 (and aeternity/aeplugin_dev_mode#15)

After reproducing the problem in the dev_mode_SUITE, the suggested fix is to notify the conductor (and tx_pool) after completed rollback to give them a chance to update their state. The observed crash appears to be due to the conductor notifying aec_tx_pool of top change with outdated parameters.


velzevur (approved):

seanhinde (approved):
LGTM

Issue #3967, opened 2022-08-02
Remove array module use in aec_peer*


zxq9:
The aec_peer* modules keep track of network peers. As with all peer networks, the vast majority of addresses in the network are ephemeral in nature. Unfortunately the expanding and contracting nature of the list of known network peers tends to conflicts with the way the peers modules have been implemented using Erlang "array" structures (here: https://github.com/aeternity/aeternity/blob/master/apps/aecore/src/aec_peers_pool.erl#L211 ), which do not have a smooth way of shrinking in size (only coming ever more nested as time goes on). This causes a miniscule memory leak which, over a period of months, can grow the size of he aec_peers process to a few dozen megabytes. This is, of course, not much memory leakage over time, but it is enough to trigger memory alarms and really should be cleaned up regardless.

The goal of this ticket is to refactor the data structures that aec_peers uses to track node information to not use the array module.

zxq9:
Fixed!

PR #3988, opened 2022-08-28
3967 pool leak

zxq9:
Fix memory leak noted in #3967 by converting from arrays to lists.


apps/aecore/src/aec_peers_pool.erl
```diff
@@ -773,8 +782,8 @@ reference_count(St, PeerId) ->
 find_peer(St, PeerId) ->
     #?ST{peers = Peers} = St,
     case maps:find(PeerId, Peers) of
```
velzevur, line 784: Maybe `maps:get/3` would be a better fit here. Just a note, not a blocker

apps/aecore/test/aec_peers_pool_tests.erl
```diff
@@ -1514,6 +1520,7 @@ validate_counters() ->
 
     % Delete all the peers.
     Pool4 = lists:foldl(fun({I, _}, P) ->
+        erlang:display(P, I),
```
velzevur, line 1523: TIL: `Erlang:display/1` - thanks!

velzevur (approved):
Great work!

apps/aecore/src/aec_peers_pool.erl
zxq9: That's true all over the module. The structure and choice of calls is pretty bizarre. If you check where `find_peer/2` is called and the way it is called, it is a wrapper around `maps:find/2` that reduces it to the *less* succinct `lists:keyfind/3`. I want to kill all of this weirdness in a cleanup commit later, but there are a *lot* of stupid things like this and I don't have time right now to care about it. In three weeks I will come back and chop all of this nonsense up.

apps/aecore/test/aec_peers_pool_tests.erl
zxq9: Dhoh! I should remove that. It adds nothing to the pass/fail nature of the tests themselves.

velzevur (approved):

uwiger (approved):

PR #3992, opened 2022-08-29
In HC, genesis block has an empty info

velzevur:
Ever since Minerva protocol upgrade, keyblocks come with an info field. It
represents the node's version. This field is not validated in any way and is
intended to be used for signaling soft upgrades. All networks so far had
started from a Roma genesis block (empty info field).

Genesis block is being computed according to the latest protocol. Since HCs
start from CERES protocol, the genesis block includes the node's version at
that height. Although this is not validated, it impacts heavily the syncing
capabilities of nodes - it limits HC nodes to the same fork as the one the
genesis block was produced on.

This PR makes the genesis to have an empty field for HC nodes.

### DIFF
--- a/apps/aecore/src/aec_headers.erl
+++ b/apps/aecore/src/aec_headers.erl
@@ -296,16 +296,18 @@ new_key_header(Height, PrevHash, PrevKeyHash, RootHash, Miner, Beneficiary,
                     key_seal     = KeySeal,
                     nonce        = Nonce,
                     time         = Time,
-                    info         = make_info(Version, Info),
+                    info         = make_info(Height, Version, Info),
                     version      = Version
                }).
 
-make_info(Version, Info) when (Version >= ?MINERVA_PROTOCOL_VSN), ?IS_INT_INFO(Info) ->
+make_info(0, _Version, default) ->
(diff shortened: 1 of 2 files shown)


velzevur:
This PR only covers HC consensus, I could make it for BitcoinNG and PoS consensus as well. Comments are welcome

uwiger (approved):

dincho:
Can I kindly ask to fix this in general not only for HC consensus please?

apps/aecore/src/aec_headers.erl
```diff
@@ -358,7 +360,7 @@ info(#key_header{info = <<I:?OPTIONAL_INFO_BYTES/unit:8>>}) ->
 
 -spec set_info(key_header(), info()) -> key_header().
 set_info(#key_header{version = Vsn} = H, I) ->
```
hanssv, line 362: matching on height? But the compiler might be good enough here...

hanssv (approved):

uwiger (approved):

hanssv (approved):

uwiger (approved):

PR #4015, opened 2022-10-04
More aggressive connector cleanup, more general connect retries

uwiger:
This PR seeks to address two problems observed during State Channel demo development:

* Initiator connectors sometimes don't get stopped when the FSM terminates due to the WebSocket client disconnecting. The connectors seem blocked in some function, and don't see the polite `close` request until (sometimes very much) later. As a remedy, the FSM will do `exit(Connector, kill)` if the polite request fails.
* Connectors already had a timeout-retry loop to handle the case when the initiator starts before the responder, but once connected at the TCP level, the initiator would fail immediately if it couldn't pair with the right responder. Now, the retry loop covers both the TCP connect and the initiator-responder pairing.


velzevur (dismissed):

uwiger:
Superseded by #4051 (merged). Closing.

Issue #4017, opened 2022-10-10
Warnings about deprecated use of `ranch` options


uwiger:
## Expected Behavior

A minimum of unnecessary warnings should be issued by the node

## Actual Behavior

When starting the HTTP server and WebSocket listener, deprecation warnings are issued:

```
2022-10-10 15:25:25.086 [warning] <0.1787.0> Setting Ranch options together with socket options is deprecated. Please use the new map syntax that allows specifying socket options separately from other options.
```

## Steps to Reproduce the Problem

  1. Start a node
  1. Look at the log output

uwiger:
Fixed in #4018

PR #4018, opened 2022-10-10
Reformat Ranch options to avoid deprecation warnings

uwiger:
See Issue #4017 

The change complies with the suggestion of the deprecation warnings issued at startup.


seanhinde (approved):
Lovely :)

Issue #3883, opened 2022-04-14
Improve and extend GC


uwiger:
With the refactored `mnesia_rocksdb` plugin, the problem of further increasing the demand for open file descriptors by adding tables is essentially eliminated (at least as far as rocksdb and mnesia are concerned).

The GC could be generalized and made more robust by keeping A-B tables for each state tree.

uwiger:
Fixed in #4106

PR #4021, opened 2022-10-13
Use A-B table pair for account states (GC) - no init:restart()

uwiger:
This is derived from the larger work under issue #3883 - a quick fix in order to avoid blocking the conductor for an extended period of time and to avoid doing `init:restart()` in order to perform the GC swap.

Strategy:
* Use an A-B pair for the `accounts` state tree, and an `aec_chain_state` value pointing to the current primary
* The above pointer is also cached as a persistent_term, since it's checked at each state tree access.
* The fullsweep process spawned by `aec_gb_gc` now also dumps the live state tree to the secondary table in the db
* The 'diff' sweeps for subsequent blocks update both the ets cache and the secondary table
* The call from the conductor now only modifies the pointer to the new primary (after verifying that the cache is up-to-date)


velzevur (dismissed):

velzevur (approved):

uwiger:
Closing. Merged as #4034

PR #4025, opened 2022-10-23
Allow provisioning of custom prefunded accounts file

nikita-fuchs:
This is intended to be replacement for https://github.com/aeternity/aeternity/pull/3978. it's

1. rebased on the latest changes on that module in `master`
2. accounting for @hanssv requests to be a
 2.1 generic, non-devmode focused solution and 
 2.2 not relying on any new dependencies to the node, which thanks to https://github.com/aeternity/aeternity/pull/4020 is not necessary anymore.

PR #4039, opened 2022-11-03
Uw devmode account creation

uwiger:
Same as #4025, but from a trusted account in order to be able to run the full CI.


uwiger:
Merged in other PR

PR #4040, opened 2022-11-03
Fix nonce returned on Oracle.query event

uwiger:
(Same as #4027, but pushed from a trusted account in order to run full CI)
Nonce was being hard-coded to 0 when triggering Oracle.query events and thus generating the wrong oracle IDs.


uwiger:
Merged in other PR. Closing.

Issue #4043, opened 2022-11-07
Random failure in aest_sync_SUITE


uwiger:
## Expected Behavior

Passing system test, even in the presence of trivial errors.

## Actual Behavior

Any entry in the crash.log will cause the test to fail, for example:
```
*** User 2022-11-07 11:04:54.485 ***
Node node1's crash logs is not empty:
2022-11-07 11:04:46 =SUPERVISOR REPORT====
     Supervisor: {local,aec_connection_sup}
     Context:    shutdown_error
     Reason:     {shutdown,{gen_server,call,[aec_peers,{get_connection,<<110,24,196,223,45,184,46,170,246,151,230,202,58,123,68,155,139,69,7,137,185,188,110,90,7,215,166,71,48,55,219,109>>}]}}
     Offender:   [{pid,<0.2004.0>},{id,aec_tx_pool_sync},{mfargs,{aec_tx_pool_sync,start_link,[]}},{restart_type,permanent},{shutdown,5000},{child_type,worker}]
```

## Steps to Reproduce the Problem

  1. Happens pretty regularly during CI

## Suggestion
In this particular case, it might be warranted to catch that particular error and return an `{error, not_running}` instead, given that the API already supports sync being stopped, e.g. during maintenance or offline mode.

uwiger:
#3475

uwiger:
Fixed in #4044

PR #4049, opened 2022-11-16
shutdown exit is wrapped by gen_server

uwiger:
See issue #4043 
Previous fix didn't solve the problem


velzevur (approved):

uwiger:
To be clear, that code represented the original:
```erlang
2> sd:start().
{ok,<0.87.0>}
3> try sd:test() catch C:E -> {caught, C, E} end.
...
{caught,exit,{foo,{gen_server,call,[sd,crash]}}}
```

fabiankrol (approved):

Issue #4011, opened 2022-09-22
State Channel initiator will not sleep and retry if the TCP connect accidentally succeeds


uwiger:
## Expected Behavior

If a State Channel initiator tries and fails to connect to a responder, it should sleep and try again a few times.

## Actual Behavior

This happens if there is no TCP acceptor available at the endpoint, but if another responder is waiting on the same port (a supported pattern), the TCP connect will succeed, and the initiator may then fail to pair up with any of the waiting responders.

## Steps to Reproduce the Problem

  1. Have multiple initiator/responder pairs connecting on the same port number, I1/R1 and I2/R2
  1. Open an endpoint for R1
  1. Open an endpoint for I2

(This _should_ recreate the problem - I haven't actually tried it)

PR #4051, opened 2022-11-23
Refactoring of State Channels responder connection handling

uwiger:
See issue #4011 

We introduce an acceptor dispatcher (`aesc_acceptors`), which ensures that a new acceptor is started as soon as an acceptor connects at the TCP level. The newly connected acceptor needs to negotiate a noise session, then find a waiting responder and match to it.

Some preparation has been made to support pre-configured listen ports, together with the size of the acceptor pool. The `aesc_listeners` server no longer worries about responder pubkeys. Since we need to support dynamically requested listen ports, the server differentiates between `preconfigured` and `dynamic` listeners: the latter are automatically removed when there are no processes alive that expressed interest via `aesc_listeners:ensure_listener/2`.

Default acceptor pool size is `1`. Given that it takes less than a millisecond from a connect to the creation of a new acceptor, it's unlikely that this will be a bottlenect, except possibly for heavily loaded SC server applications (in which case, the chain is much more likely to be the bottleneck).

Also, some reduction of debug log noise by compacting hashes and keys (keeping the first and last 5 bytes, and sticking 4 zeroes in the middle).

### DIFF
--- a/apps/aechannel/src/aesc_acceptors.erl
+++ b/apps/aechannel/src/aesc_acceptors.erl
@@ -1,0 +1,179 @@
+-module(aesc_acceptors).
+-behavior(gen_server).
+
+-export([start_link/0]).
+
+-export([ start_acceptors/3
+        , start_acceptors/5
+        , stop_acceptors/1
+        , worker_done/1
+        ]).
+
+-export([ init/1
+        , handle_call/3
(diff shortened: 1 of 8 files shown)


seanhinde (approved):
LGTM

PR #4053, opened 2022-11-28
Add plugin config management support

uwiger:
A plugin-specific version of `aeu_env:find_config/2` is added:

`aeu_plugins:find_config(PluginName, Key, Path) -> {ok, any()} | undefined`

The plugin schema is also cached, once the schema and the corresponding user config are validated.

### DIFF
--- a/apps/aeutils/src/aeu_plugins.erl
+++ b/apps/aeutils/src/aeu_plugins.erl
@@ -3,6 +3,7 @@
 -export([load_plugins/0]).
 
 -export([ check_config/3
+        , find_config/3
         , validate_config/2
         , is_dev_mode/0
         , suggest_config/2 ]).
@@ -17,14 +18,16 @@ check_config(PluginName, SchemaFilename, OsEnvPrefix) ->
                 [Config] ->
                     Schema = load_schema(SchemaFilename),
                     {ok, Config1} = validate(Config, Schema),
-                    case aeu_env:apply_os_env(OsEnvPrefix, Schema, Config1) of
-                        no_change ->
-                            Config1;
-                        {error, E} ->
-                            error(E);
-                        Config2 when is_map(Config2) ->
-                            Config2
-                    end;
+                    Result = case aeu_env:apply_os_env(OsEnvPrefix, Schema, Config1) of
+                                 no_change ->
+                                     Config1;
+                                 {error, E} ->
+                                     error(E);
+                                 Config2 when is_map(Config2) ->
+                                     Config2
+                             end,
+                    cache_plugin_schema(NameBin, Schema),
+                    Result;
                 [] ->
                     lager:warning("Could not fetch plugin config object (~p)",
                                   [PluginName]),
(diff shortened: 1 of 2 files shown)


velzevur (approved):

apps/aeutils/src/aeu_plugins.erl
```diff
@@ -36,6 +39,39 @@ check_config(PluginName, SchemaFilename, OsEnvPrefix) ->
...
+    case find_config_step(PluginName, Key, H) of
+        undefined -> find_config_(PluginName, Key, T);
+        {ok, _} = Ok -> Ok
+    end;
+find_config_(_, _, []) ->
+    undefined.
+
+find_config_step(PluginName, Key, user_config) ->
+    aeu_env:find_config([<<"system">>, <<"plugins">>,
+                         {<<"name">>, PluginName, <<"config">>} | Key], [user_config]);
```
seanhinde, line 70: In the path `[<<"system">>, <<"plugins">>,{<<"name">>, PluginName, <<"config">>} | Key] `am I thinking correctly that `{<<"name">>, PluginName, <<"config">>}` is essentially a compound key in a list element (plugins)? If so it seems a bit unnecessary to have the `<<"name">>` and `<<"config">>` things as part of the list key - could be clearer to have them outside, or just drop them, depending on whether there are any other element types under plugins. So something like:` [<<"system">>, <<"plugins">>,{PluginName}, <<"config">> | Key]`

(I'm thinking of this in terms of future work to regularise the schema, which is currently yang and otherwise compatible)

uwiger: The meaning of the pattern is: for `{Key1, Val1, Key2}`, find the first object where `Key1 : Val1`, then return the value corresponding to `Key2`. That is, in the following config, we want to match on `name` and traverse into `config`:

```yaml
system:
    dev_mode: true
    plugins:
        -
          name: aeplugin_dev_mode
          config:
              keyblock_interval: 0
              microblock_interval: 0
        -
```

uwiger: For further clarification, this is the internal representation of the above config (or similar):
```erlang
{<<"system">> =>
     #{<<"dev_mode">> => true,
       <<"plugins">> =>
           [#{<<"config">> =>
                  #{<<"keyblock_interval">> => 0,
                    <<"microblock_interval">> => 0},
              <<"name">> => <<"aeplugin_dev_mode">>},
            #{<<"name">> => <<"aeplugin_scm">>}]}}
```

And the normal operation of `find_config()` is that `[K1, K2 | ...]` corresponds to map keys, where the corresponding value is a new map - the subtree to traverse into.

uwiger: Your example would assume that the object is always keyed on name?
We have one other keyed object array in the schema (`mining:cuckoo:miners`), and it is keyed on "executable". Another object array (`metrics:rules`) does have a "name" attribute, but it's not a key, and not required (actually, there is no suitable key attribute, so one would presumably fetch the whole array instead).

seanhinde: All good then. Understood. That works. Non keyed items could just have a implicit index in some future config setup.

seanhinde (approved):

PR #4055, opened 2022-11-30
Move dev mode emitter into core node as part of new app

uwiger:
As the basic dev mode functionality turns out to be useful for many different scenarios, it was suggested that it move into the core node. Here is a first implementation.

This would require some adaptations in the aeplugin_dev_mode application.

### DIFF
--- a/apps/aedevmode/src/aedevmode_app.erl
+++ b/apps/aedevmode/src/aedevmode_app.erl
@@ -1,0 +1,28 @@
+-module(aedevmode_app).
+-behavior(application).
+
+-export([ start/2
+        , stop/1 ]).
+
+-export([post_check/0]).
+
+start(_StartType, _StartArgs) ->
+    aedevmode_sup:start_link().
+
+stop(_State) ->
+    lager:info("Stopping aedevmode_app", []),
+    ok.
+
+%% Run at 300, i.e. after normal env checks AND the db start
+post_check() ->
+    case aecore_env:is_dev_mode() of
+        true ->
+            Pub = patron_pubkey_for_testing(),
+            aeu_env:suggest_config([<<"mining">>, <<"beneficiary">>], Pub);
+        false ->
+            ok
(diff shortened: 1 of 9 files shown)


apps/aedevmode/src/aedevmode_app.erl
```diff
@@ -0,0 +1,28 @@
...
+-export([post_check/0]).
+
+start(_StartType, _StartArgs) ->
+    aedevmode_sup:start_link().
+
+stop(_State) ->
+    lager:info("Stopping aedevmode_app", []),
+    ok.
+
+%% Run at 300, i.e. after normal env checks AND the db start
```
uwiger, line 16: I realize that it's not clear why I put this check so late in the flow.
Initially, I had it test for (dev mode AND `top_height() == 0`). Then I removed the top_height check ...

apps/aedevmode/src/aedevmode_app.erl
```diff
@@ -0,0 +1,28 @@
...
+post_check() ->
+    case aecore_env:is_dev_mode() of
+        true ->
+            Pub = patron_pubkey_for_testing(),
+            aeu_env:suggest_config([<<"mining">>, <<"beneficiary">>], Pub);
```
nikita-fuchs, line 21: can't see from here where this is ran, but it would be nice if it wouldn't override the mining beneficiary I set from the pool of generated (and used) accounts.

uwiger: The `post_check()` is run in start phase `300`, that is, after all the 'official' start phase hooks.
Also, the `suggest_config/2` function only sets the value _if_ it was previously undefined. So it explicitly won't override anything set previously.

uwiger:
Rebased to current master (no issues)

seanhinde (approved):
Great stuff

velzevur (approved):

PR #4057, opened 2022-12-02
Protect Fee and GasPrice in GaMetaTx

### LINKED ISSUE #3417: Meta TX fee and gas_price
If the account is a `generalized_account`, the authentication method is not the `basic` signatures produced by a private key but a contract call wrapped in a `ga_meta_tx`. The issue is that this wrapper transaction is not authenticated, thus could be thinkered by the miner before inclusion. As [per documentation](https://github.com/aeternity/protocol/blob/master/serializations.md#generalized-accounts-meta-transaction), a meta transaction is

```
[ <ga_id>       :: id()
, <auth_data>   :: binary()
, <abi_version> :: int()
, <fee>         :: int()
, <gas>         :: int()
, <gas_price>   :: int()
, <ttl>         :: int()
, <tx>          :: binary()
]
```
Changing any of `ga_id`, `auth_data`, `abi_version` or `tx` would instantly render it invalid. The `gas` has a protocol defined maximum, so it can not be bumped much and even so - this does not change the actual consumed gas. The `ttl` could be modified and this could make the tx sort of invalid or maybe not, this is addressed in #3416. What is left are `fee` and `gas_price` which a malicious miner could change right before including a meta transaction. Those (and maybe the rest as well?) should be subject to the `auth_data`

ThomasArts:
fixes #3417

By making fee and gas_price part of the authentication data, it is no longer possible for a miner to increase the fee or gas_price without invalidating the transaction.

### DIFF
--- a/apps/aechannel/src/aesc_utils.erl
+++ b/apps/aechannel/src/aesc_utils.erl
@@ -702,7 +702,7 @@ verify_meta_tx(SignerId, StoreKey, AuthContractId, MetaTx, Trees, Env, TxType)
                        , store       => Store
                        , call        => Call
                        , trees       => Trees
-                       , tx_env      => set_auth_tx_hash(aega_meta_tx:tx(MetaTx), Env, TxType)
+                       , tx_env      => set_auth_tx_hash(MetaTx, Env, TxType)
                        , off_chain   => false
                        , origin      => SignerPK
                        , creator     => aect_contracts:owner_pubkey(Contract)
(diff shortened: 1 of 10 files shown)


ThomasArts:
TO DO: 

- [ ] better/real serialization
- [ ] consensus breaking note
- [ ] release note
- [ ] rebase

hanssv (commented):
Looks good, and the checklist seems to accurately describe what is still missing 😅

velzevur (approved):

PR #4067, opened 2022-12-28
Support loading plugins from archive files

uwiger:
The `aeu_plugins` module is extended with some useful support functions for plugins, and employs a different loading strategy:

* Plugins can now be put under `plugins/` as `.ez` zip archives (e.g. created using the `aeplugin_rebar3` rebar plugin)
* The node creates a subdir, `plugins/lib/`, where it unpacks needed applications, extracting the version and adding to the application directory name, if necessary
* Dependencies are detected, and only those that are not already in the system are also unpacked to `plugins/lib/` and loaded/started

The [aeplugin_rebar3](https://github.com/aeternity/aeplugin_rebar3) build plugin allows for creating an archive file with the needed dependencies, using the command `rebar3 ae_plugin`.

New API functions:
* `aeu_plugins:data_dir(PluginName)` - returns a plugin-specific data area under the node `data/` directory
* `aeu_plugins:log_dir(PluginName)` - returns a plugin-specific log area under the node `log/` directory
*

### DIFF
--- a/apps/aeutils/priv/aeternity_config_schema.json
+++ b/apps/aeutils/priv/aeternity_config_schema.json
@@ -47,6 +47,10 @@
                         }
                     },
                     "default": []
+                },
+                "custom_prefunded_accs_file" : {
+                    "description" : "Overrides the automatically derived file name for prefunded accounts. No default value",
+                    "type" : "string"
                 }
             }
         },
--- a/.circleci/config/commands/@install.yml
+++ b/.circleci/config/commands/@install.yml
@@ -4,8 +4,8 @@ aeplugin_install_devmode:
         name: Download devmode plugin
         command: |
           export PLUGIN_VERSION=<< pipeline.parameters.aeplugin_devmode_version >>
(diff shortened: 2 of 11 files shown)


apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -47,6 +47,10 @@
                         }
                     },
                     "default": []
+                },
+                "custom_prefunded_accs_file" : {
```
dincho, line 51: why plugins suddenly started polluting the node configuration ?

uwiger: > why plugins suddenly started polluting the node configuration ?

The way it normally works is that the prefunded accounts are created through hardcoded files.
This way, it's possible to create a custom chain using unmodified source without inheriting a truckload of predefined accounts specific to mainnet.

dincho: Nevermind, that seems just ENV->config variable change, unrelated to the plugin update

velzevur (dismissed):

velzevur (approved):

seanhinde (approved):

PR #4069, opened 2023-01-06
Improvements to Rosetta uncovered during use

seanhinde:
Improvements to Rosetta uncovered during use

* Improve error handling


velzevur (approved):

apps/aehttp/src/aehttp_dispatch_rosetta.erl
```diff
@@ -177,6 +186,8 @@ handle_request_(networkStatus, _, _Context) ->
               <<"peers">> => PeersFormatted},
         {200, [], Resp}
     catch
+        invalid_network ->
```
uwiger, line 189: So, omitting the class in the catch pattern, it defaults to `throw`. Sneaky.

uwiger (approved):

PR #4070, opened 2023-01-13
Fix cors config parsing

velzevur:
CORS configs can be setup in the node config. The old code made the wrong
assumption that they were delivered as a proplist. This PR fixes it as well as
it provides exhaustive test suite for CORS so we don't suffer a regression.


dincho (approved):
seems to work for my usecase

dincho (approved):

uwiger (approved):

PR #4075, opened 2023-01-18
Extend get_generation timeout to handle larger generations

seanhinde:
Thinking that increasing this single timeout ought to allow clean sync against the largest generations in UAT. Let's see.


velzevur (approved):

seanhinde:
Also tested full sync against UAT network, which has caused previous issues. Sync completed.

hanssv (approved):

uwiger (approved):

PR #4079, opened 2023-01-30
FATE Debugger

ghallak:


apps/aefate/src/aefa_engine_state.erl
```diff
@@ -100,9 +122,22 @@
...
 -type void_or_fate() :: ?FATE_VOID | aeb_fate_data:fate_type().
 -type pubkey() :: <<_:256>>.
 
+-record(debug_info, { status = continue     :: debugger_status()
+                    , location = none       :: debugger_location()
+                    , breakpoints = []      :: list()
+                    , skip_instructions = 0 :: integer()
+                    , vars_registers = #{}  :: map()
+                    , call_stack = []       :: [{string(), pos_integer()}]
+                    }).
```
radrow, line 137: Move to `aefa_dbg`?

apps/aefate/src/aefa_engine_state.erl
```diff
@@ -126,6 +161,7 @@
             , stores            :: aefa_stores:store()
             , trace             :: list()
             , vm_version        :: non_neg_integer()
+            , debug_info        :: disabled | debug_info()
```
radrow, line 164: Try to make existence of this field conditional by a preprocessor directive `debug_enabled`

apps/aefate/src/aefa_fate.erl
```diff
@@ -260,7 +260,16 @@ abort(E) -> throw({add_engine_state, E}).
 
 execute(EngineState) ->
     Instructions = aefa_engine_state:current_bb_instructions(EngineState),
-    loop(Instructions, EngineState).
```
radrow, line 263: If debugging is enabled via macro, engine state can just assume the debugger is in

apps/aefate/src/aefa_fate.erl
```diff
@@ -295,6 +306,16 @@ step([I|Is], EngineState0) ->
             catch_protected(Err, EngineState0)
     end.
 
+break_or_step(Is, ES) ->
+    case aefa_engine_state:debugger_status(ES) of
+        break ->
+            InstructionsCount = length(aefa_engine_state:current_bb_instructions(ES)),
+            InstructionsSkip = InstructionsCount - length(Is),
+            {break, aefa_engine_state:set_skip_instructions(InstructionsSkip, ES)};
```
radrow, line 314: Consider adding instruction counter in dbg state

apps/aefate/src/aefa_engine_state.erl
```diff
@@ -863,3 +923,88 @@ vm_version(#es{vm_version = X}) ->
...
+
+%%%------------------
+
+debugger_resume(ES = #es{debug_info = Info = #debug_info{status = step}}) ->
+    ES#es{debug_info = Info#debug_info{status = break}};
+debugger_resume(ES = #es{debug_info = Info = #debug_info{status = {next, From}}}) ->
+    case ES#es.current_function == From of
+        true  -> ES#es{debug_info = Info#debug_info{status = break}};
+        false -> ES
+    end;
```
radrow, line 1003: This should be resolved using callstack --- otherwise you will always enter recursive calls (indirect ones too!)

apps/aefate/src/aefa_engine_state.erl
```diff
@@ -100,9 +122,22 @@
...
+
 -type void_or_fate() :: ?FATE_VOID | aeb_fate_data:fate_type().
 -type pubkey() :: <<_:256>>.
 
+-record(debug_info, { status = continue     :: debugger_status()
+                    , location = none       :: debugger_location()
+                    , breakpoints = []      :: list()
+                    , skip_instructions = 0 :: integer()
+                    , vars_registers = #{}  :: map()
+                    , call_stack = []       :: [{string(), pos_integer()}]
```
radrow, line 136: Maybe also keep contract call stack that would remember contract names and addresses?

apps/aefate/src/aefa_engine_state.erl
ghallak: I'd keep this as it as since there is no east way of having a conditional field and I don't want to duplicate the `es` record.

apps/aefate/src/aefa_fate.erl
ghallak: This will do 79686760aa7bbb12c21ef7d4026f4eab0743ec0f ?

apps/aefate/src/aefa_fate.erl
ghallak: Something like this 4e0f209ed631428d104bdafee71d65cf42aa62bf ?

apps/aefate/src/aefa_engine_state.erl
ghallak: This should be resolved after this 74fd3e8dc98909c3590de895ec5153c35a03dcae

apps/aefate/src/aefa_engine_state.erl
```diff
@@ -863,3 +924,93 @@ vm_version(#es{vm_version = X}) ->
...
+        []        -> undefined;
+        [Reg | _] -> Reg
+    end.
+
+%%%------------------
+
+debugger_resume(ES = #es{debug_info = Info = #debug_info{status = step}}) ->
+    ES#es{debug_info = Info#debug_info{status = break}};
+debugger_resume(ES = #es{debug_info = Info = #debug_info{status = {next, Stack}}}) ->
+    case length(ES#es.call_stack) == length(Stack) of
```
ghallak, line 1005: I'm passing the entire call stack along with `{next, Stack}` but I'm not sure that this is needed. Possibly only the size of the call stack or the head of the call stack (if it's unique) should work.

apps/aefate/src/aefa_fate.erl
radrow: Yes

apps/aefate/src/aefa_fate.erl
radrow: yes

apps/aefate/src/aefa_engine_state.erl
radrow: I got lost, why does it depend on the stack? Isn't it supposed to just execute until next `DBG_LOC`?

apps/aefate/src/aefa_engine_state.erl
```diff
@@ -100,9 +123,22 @@
...
+-type debugger_location() :: none | {string(), integer()}.
+
 -type void_or_fate() :: ?FATE_VOID | aeb_fate_data:fate_type().
 -type pubkey() :: <<_:256>>.
 
+-record(debug_info, { status = continue       :: debugger_status()
+                    , location = none         :: debugger_location()
+                    , breakpoints = []        :: list()
+                    , current_instruction = 0 :: integer()
+                    , vars_registers = #{}    :: map()
```
radrow, line 136: You can specify types of keys and values of a map.

ghallak: Done here 1205b532673694f7d341bd59acae9e47a4e809f5

apps/aefate/src/aefa_engine_state.erl
ghallak: Solved here by adding `DBG_CONTRACT` https://github.com/aeternity/aeternity/pull/4079/commits/b8213daa5f6d63d0cc312e8a5da11b88f2b07b12.

apps/aefate/src/aefa_fate_op.erl
```diff
@@ -2976,3 +2984,34 @@ gt_to_emcl(X1, X2, X3, X4, X5, X6, X7, X8, X9, X10, X11, X12) ->
...
+
+-ifdef(DEBUG_INFO).
+dbg_loc({immediate, File}, {immediate, Line}, EngineState) ->
+    Info  = aefa_debug:set_debugger_location({File, Line}, aefa_engine_state:debug_info(EngineState)),
+    Stack = aefa_engine_state:call_stack(EngineState),
+    UpdatedStatus = 
+    case lists:member({File, Line}, aefa_debug:breakpoints(Info)) of
+        true  ->
+            case aefa_debug:debugger_status(Info) of
+                continue -> aefa_debug:set_debugger_status(break, Info);
```
ghallak, line 2996: I have checked the Erlang debugger, and it does not break neither on stepover (called next) nor on stepout (called finish). That's why the breakpoints will be hit only with continue, or stepin (which executes every line anyway, so there's no need to mention it here).

rebar.config
```diff
@@ -205,18 +205,19 @@
...
                             {include_erts, false},
                             {include_src, true},
                             {sys_config, "./config/dev1/sys.config"},
                             {vm_args, "./config/dev1/vm.args"}]},
+                    {erl_opts, [{d, 'DEBUG_INFO'}]},
                     {dist_node, [{setcookie, 'aeternity_cookie'},
                                  {sname, 'aeternity_ct@localhost'}]},
                     {deps, [{meck, "0.8.12"},
-                            {aesophia, {git, "https://github.com/aeternity/aesophia.git", {ref,"4ae2472"}}},
+                            {aesophia, {git, "https://github.com/aeternity/aesophia.git", {ref,"122a6e0"}}},
```
hanssv, line 220: You need to have a Ceres compiler...

ghallak: The ceres branch should be rebased on top of master after merging the debugger PR in the aesophia repository, only after that we can have a ceres compiler here.

Alternatively, the ceres branch in aesophia could be merged to master, and then the debugger branch could be rebased on top of the new master.

I think the first option is more likely to be the right one. Could you suggest another solution for this?

hanssv: The ceres branch of aesophia should be kept separate until we get nearer to the hard fork, so first option is the right one. In the meantime you could rebase the ceres branch (new_ceres?)  on top of the debugger PR in aesophia to make sure it works with this PR 😅

radrow (approved):
uh finally

radrow (approved):

radrow:
party!

PR #4095, opened 2023-02-16
force hc contracts on iris

jyeshe:


jyeshe:
No need to include IRIS in the condition as `aect_state_tree:get_contract/1` returns the contract (on top hash)

PR #4097, opened 2023-02-21
fix stratum test

velzevur:
Stratum config includes keys to sign rewards distribution. They are configured
on the disc. They live a path that is set in the config. This path can be an
absolute or a relative one. In the case of a relative path, its root was
`apps/aestratum/priv`. Although not impossible, this made it inconvinient to
configure the keys, especially in the context of a release.

~This PR changes the config so the root dir for relative paths is the root dir
of the project. This allows the user to configure their keys in `data` or in
the project root if they really want to.~

~This change is breaking backwards compatibility with regards of the root of a
relative path for the keys in the config.~

This PR changes the config of the test so that it uses absolute paths but we keep backwards compatibility (even if Stratum is not working).

### DIFF
--- a/apps/aestratum/src/aestratum_config.erl
+++ b/apps/aestratum/src/aestratum_config.erl
@@ -153,7 +153,7 @@ check_keypair_roundtrips({PK, SK}) ->
 
 read_keys(Dir) ->
     AbsDir = case filename:pathtype(Dir) of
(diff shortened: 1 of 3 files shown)


uwiger (dismissed):
A more strictly correct approach would be to use `setup:data_dir()`, but the practical result would be the same in our case.

dincho (approved):

PR #4104, opened 2023-03-02
Generalized GC

uwiger:
See issue #3883 

This PR introduces a rewritten GC, using A/B table pairs for each state tree.
State tree object lookups on GC:d trees will try the A table first, then the B table. If found in B, the object is automatically promoted into A. Writes always go to A.

A GC switch (triggered by `aec_conductor` when all conditions are met), involves clearing B, then making it the new A, then starting a scanner process on the given height, which (slowly, in the background) touches each reachable node, ensuring that all reachable state is promoted into the new A. Meanwhile, all new blocks added to the chain will automatically write to the new A. When all such passes are complete for the given height, _and_ the requested history depth has been reached, a new switch can be performed.

The GC can run from the beginning, and adds ca 30% to the chain sync time with `history: 10000`.

The gc scans are regulated by `jobs`. If they run too slowly, more state than the requested history depth may be kept.


zxq9 (approved):

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -1588,22 +1588,50 @@
                     "type" : "object",
                     "additionalProperties" : false,
                     "properties" : {
-                        "enabled": {
+                        "enabled" : {
                             "description": "If true, node will perform garbage collection of account state, removing unreachable merkle patricia nodes and free some disk space",
                             "type": "boolean",
                             "default": false
                         },
+                        "from_start" : {
```
dincho, line 1596: I don't think this property name is technically correct or I'm expecting something a bit different from this functionality. Should be something more like "during_sync", "desync" or something like that?

Also, the description should mention how "in sync" is defined for this feature to work, percentage or somehow else ?

docs/garbage_collection.md
```diff
@@ -0,0 +1,116 @@
+# Introduction
```
dincho, line 1: perhaps add this to the docs.aeternity.io by referring this file in mkdocs.yml

docs/release-notes/next/GH-3883-generalized-gc.md
```diff
@@ -0,0 +1,3 @@
+* Garbage collection has been rewritten, and now operates on all state trees. It is also
```
dincho, line 1: Is this backward compatible with a database created by:
- previous version with GC enabled
- previous version without GC enabled

uwiger: Hopefully. Not tested yet :)

uwiger:
Due to rebasing issues, this PR seems to revert some previous commits. Will create a new PR.

PR #4106, opened 2023-03-02
Generalized GC (corrected base)

uwiger:
(replaces #4104)

See issue #3883 

This PR introduces a rewritten GC, using A/B table pairs for each state tree.
State tree object lookups on GC:d trees will try the A table first, then the B table. If found in B, the object is automatically promoted into A. Writes always go to A.

A GC switch (triggered by `aec_conductor` when all conditions are met), involves clearing B, then making it the new A, then starting a scanner process on the given height, which (slowly, in the background) touches each reachable node, ensuring that all reachable state is promoted into the new A. Meanwhile, all new blocks added to the chain will automatically write to the new A. When all such passes are complete for the given height, _and_ the requested history depth has been reached, a new switch can be performed.

The GC can run from the beginning, and adds ca 30% to the chain sync time with `history: 10000`.

The gc scans are regulated by `jobs`. If they run too slowly, more state than the requested history depth may be kept.

### DIFF
--- a/apps/aechannel/src/aesc_state_tree.erl
+++ b/apps/aechannel/src/aesc_state_tree.erl
@@ -17,7 +17,8 @@
          mtree_iterator/1,
          new_with_backend/1,
          new_with_dirty_backend/1,
-         root_hash/1]).
+         root_hash/1,
+         db/1]).
 
 -export([ from_binary_without_backend/1
         , to_binary_without_backend/1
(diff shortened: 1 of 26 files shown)


uwiger:
Re. previous comments:

https://github.com/aeternity/aeternity/pull/4104#pullrequestreview-1321677315

> Is this backward compatible with a database created by:
> 
> previous version with GC enabled
> previous version without GC enabled

Yes. I've tested with a database created using the `aec_db_gc_SUITE` on `master`, and also verified through code review.

uwiger:
Previous comment:

https://github.com/aeternity/aeternity/pull/4104#pullrequestreview-1321666035

> I don't think this property name is technically correct or I'm expecting something a bit different from this functionality. Should be something more like "during_sync", "desync" or something like that?

Changed to `during_sync`.

zxq9 (approved):

PR #4107, opened 2023-03-02
Limit body request length

jyeshe:
- Allows setting request body limit by the Mdw (default is 1MB)
- It would be a safer call to set a lower limit to the Node as well but as requested this depends on monitoring how the `/dry-run` have been used so the config value is kept to the one default to cowboy.

### DIFF
--- a/apps/aehttp/src/aehttp_app.erl
+++ b/apps/aehttp/src/aehttp_app.erl
@@ -23,7 +23,6 @@
 -define(DEFAULT_CHANNEL_WEBSOCKET_LISTEN_ADDRESS, <<"127.0.0.1">>).
 -define(DEFAULT_CHANNEL_ACCEPTORS, 10).
 
--define(MAX_REQUEST_LINE_LENGTH, 1024).
 % 6M gas per microblock/tx / 20 gas per byte = 300kB
 % 400k after encoding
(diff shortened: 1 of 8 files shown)


sborrazas (approved):

dincho:
What's the actual max body limit ?

jyeshe:
> What's the actual max body limit ? Perhaps I don't fully understand this option, can you elaborate a bit more than the couple sentences in the docs ?

It stops reading from the socket and closes it when the body is bigger than the config value. As discussed in the meeting, the cowboy default value will be kept since the dry run accepts multiple txs until it's better known how the dry run is actually been used.

dincho (changes requested):
config schema and epoch_full.yml also needs the corresponding updates

dincho (dismissed):

velzevur:
For every user visible change we add a markdown file in `docs/release-notes/next` (ex `docs/release-notes/next/limit_body_req_lenght.md`) that describes the change. When we do a release, we gather (and delete) all those files and put all collected change notes in the release notes. Please check #4093 for more details.

Please add a release notes line :)

velzevur (approved):

dincho (approved):

PR #4110, opened 2023-03-10
Set explicit user id for docker user

jyeshe:
Sets the current user id explicitly to make sure it's the same as the one set on the Mdw:
https://github.com/aeternity/ae_mdw/blob/a94b1d2acf0df30a16572d301aaf68eaf4d2a5cc/Dockerfile#L130

This allows setting the permission of existing `mdw` and `mnesia` database under Mdw `data/` without inspecting to which id the `aeternity` user is mapped to.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ RUN ln -fs librocksdb.so.6.13.3 /usr/local/lib/librocksdb.so.6.13 \
 COPY --from=builder /app/_build/prod/rel/aeternity /home/aeternity/node
 
 # Aeternity app won't run as root for security reasons
-RUN useradd --shell /bin/bash aeternity \
+RUN useradd --uid 1000 --shell /bin/bash aeternity \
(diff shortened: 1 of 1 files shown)


dincho:
It's already 1000 implicitly? What this explicitness fix/add ?

jyeshe:
> It's already 1000 implicitly? What this explicitness fix/add ?
> 
> ```
> aeternity@2f3da897796c:~/node$ id
> uid=1000(aeternity) gid=1000(aeternity) groups=1000(aeternity)
> ```

It's not a fix. it's setting the very same/current one that I could find in `aeternity/aeternity` explicitly to make sure that both users (from node and mdw docker images) are the same for the purpose of giving permission to the database. With the `--uid 1000` it makes it safer on the deployment env to use `chwon -R 1000 data` for the mdw database.

sborrazas (approved):

Issue #4109, opened 2023-03-10
Buggy event subscription code in test suite libs


uwiger:
## Expected Behavior

Test cases sometimes need to subscribe to node events and act accordingly

## Actual Behavior

The subscription code in aecore_suite_utils is broken in several places.

Example:
https://github.com/aeternity/aeternity/blob/master/apps/aecore/test/aecore_suite_utils.erl#L705-L725

```erlang
wait_for_tx_in_pool(Node, SignedTx, Timeout) ->
    true = rpc:call(Node, aec_events, subscribe , [tx_created]),
    true = rpc:call(Node, aec_events, subscribe , [tx_received]),
    Hash = aetx_sign:hash(SignedTx),
    case rpc:call(Node, aec_chain, find_tx_location, [Hash]) of
```

Since aec_events uses gproc registrations, the process that actually registers for evens is the temporary process spawned by the rpc server. This registration will be cleaned up once the process completes, and in any case, events will _not_ be published to the caller.

There is an API function in the same module, `aecore_suite_utils:subscribe(Node, Event)`, but this is buggy too.
https://github.com/aeternity/aeternity/blob/master/apps/aecore/test/aecore_suite_utils.erl#L1645-L1654

```erlang
                            case catch aec_events:subscribe(Event) of
                                ok ->
                                    From ! {Ref, ok},
                                    proxy_loop(
                                      [{From, Event}|
```
The problem with this code is that `aec_events:subscribe(Event) -> true`, which will be interpreted as an erroneous result, and the subscription is ignored by the proxy process.

uwiger:
Fixed in #4111

PR #4111, opened 2023-03-14
Fix event subscription in test suite lib

uwiger:
See issue #4109 

Test suites pass locally with the corrected event subscription logic.


uwiger:
To clarify, the bug analysis is actually in the issue description. The bug was found through code review.

velzevur (approved):

zxq9 (approved):
LGTM

PR #4115, opened 2023-03-17
Add bigint format type hint

loxs:
Here is a suggestion to add type hints, which can be used by the REST client generators, in order to encourage usage of the native JS type `bigint` (and similar types in other languages).

The format keyword is probably the right way to do it, according to the OAS spec. Here are some docs: https://swagger.io/docs/specification/data-models/data-types/
Search for "format modifier" in the page.

I am still not sure about the implications of this, let's discuss it first.

### DIFF
--- a/apps/aehttp/priv/oas3.yaml
+++ b/apps/aehttp/priv/oas3.yaml
@@ -286,6 +286,7 @@ paths:
           required: true
           schema:
             type: integer
+            format: "bigint"
             minimum: 1
             example: 1
       responses:
@@ -819,6 +820,7 @@ paths:
           required: false
           schema:
             type: integer
+            format: "bigint"
             default: 20
             minimum: 1
             maximum: 1000
@@ -1801,6 +1803,7 @@ paths:
           required: true
           schema:
             type: integer
+            format: "bigint"
             minimum: 0
       responses:
         "200":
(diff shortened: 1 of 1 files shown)


hanssv:
Do you really need bigint for values that are defined with a minimum of 1 and a maximum of 1000?? Are JS integers 8bit?

loxs:
JS does not have a proper integer primitive, instead it has `number` which can be either an int or a float. Even when using TypeScript, one can't be sure that a number is an int at compile time.

Bigint on the other hand is guaranteed to be an `int`.  It has the cost of not supporting older JS environments of course. I often rely on bigints even just for correctness.

Regarding the overhead, it's very minimal and with a large degree of certainty, the smallest performance concern in any reasonably big JS project. If you are interested, read some React horror story about performance :) 

So we don't strictly need bigints for many of the numbers, we can safely use regular numbers.

On the other hand we provide the `string-to-int` query params which serializes every int, even the small ones to a string. This flag is always set to `true` in the SDK. I'd speculate that the overhead of sending strings over the wire is much greater than any potential problems of using bigints on the other side. The ultimate solution to this would be to not allow `string-to-int` flag (or not honor it) for schemas that have max value that fits in the JS number range. And the client can decide for themselves how to use these numbers.

Also this PR was created when I was toying with REST (client) code generators - an idea which I afterwards abandoned in favor of writing all the handlers (to all endpoints) by hand. So in reality for this new project I don't even need it any more. I guess I'm just closing the PR...

dincho:
> I'd speculate that the overhead of sending strings over the wire is much greater than any potential problems of using bigints on the other side.

There shouldn't be any "wire" overhead as both endup in a big string called JSON :) Well, except for the extra " denoting a string in that format :)

davidyuk:
> format: "bigint"

I would guess is it a big number by `maximum` value.

> writing all the handlers (to all endpoints) by hand

Seems to be a cumbersome approach.

I don't think we need to enforce all numbers to be bigints, isn't it the same as replacing all Objects with Maps? In js sdk I'm trying to keep Numbers for block height, gas, nonce. And bigint for ae amounts, fees, gasPrice. So the user won't be able to multiply fee by block height without explicit conversion to the same type.

PR #4116, opened 2023-03-17
Revisit GA contract ga_main_w_temporary.aes

velzevur:


velzevur:
We will discuss adding another layer of super accounts that can change the master.

loxs:
Are we sure that wallets can sign with this `ASDF` prefix? I thought it needs to be exactly the same as the "message" prefix, which seems to be `"aeternity Signed Message:\n"` and otherwise, the wallet will refuse to sign it?

velzevur:
The best case would be dry-running the `to_sign` entrypoint to generate the hash to be signed for you so you don't generate the hash yourself. The `prefix` is fed to the contract in the `init` so it could be any message.

hanssv (approved):

hanssv:
What we would really like is an arbitrary sized binary/bytes type in Sophia - let's add that to the wish-list

uwiger (approved):

rumyanar (approved):
Tested and it works ok with the new Master changes.

PR #4118, opened 2023-03-21
CI maintenance

dincho:


.circleci/config/jobs/@packages.yml
```diff
@@ -116,6 +116,7 @@ macos-standalone:
       type: string
   macos:
     xcode: "13.2.1" # macOS 11.6.2 (Big Sur)
```
hanssv, line 118: Is the comment still relevant?

hanssv (approved):

.circleci/config/jobs/@packages.yml
dincho: should be, I haven't actually updated xcode/macos version

Issue #4096, opened 2023-02-20
Queries beyond GC point fail with "Internal server error" rather than a notice about GC


zxq9:
## Expected Behavior
When a query is sent to a GC'd node over the JSON API endpoint (for example, "/v2/accounts/[AccID]/height/[BlockHeight]") if the object does appear on chain but the requested block height is unavailable due to GC the response should be something along the lines of "GC: Out of range" to inform the user that the data exists, but only a full node can service the query.

## Actual Behavior
When a query is sent to a GC'd node over the JSON API endpoint if the object does appear on chain but the requested block height is unavailable due to GC the response that is returned is "Internal server error". This is wrong, of course -- the server is experiencing no error, it is functioning correctly as a GC'd node.

An example of covering for the current behavior is the current way the Vanillae Erlang lib handles this case:
https://github.com/aeternity/Vanillae/blob/36ddcfc97f4d884106dc644e36f0977c4e064c0a/bindings/erlang/src/vanillae.erl#L491-L497

uwiger:
Fixed in #4119. Closing

PR #4119, opened 2023-03-22
Return 410 code for queries beyond GC point

uwiger:
See issue #4096 

This implementation assumes that if GC is enabled and the requested state is at a height lower than the height of the last GC switch, then the state may be either fully or partially deleted, and we don't even try to retrieve it. The return code `410` means `gone`, i.e. the data has been deleted.

The PR is based on PR #4106 (not yet merged)


hanssv (approved):
Release note?

velzevur (approved):

hanssv (approved):

PR #4124, opened 2023-04-10
Bitcoin parent chain hyperchain connector

seanhinde:
Basic functionality for bitcoin parent chain connector

TODO:
- [x] Configuration params
- [x] Put child chain network id in commitment (only 15 bytes remain in BTC OP_RETURN)?
- [x] Testing in a system with a full bitcoin node

Pushed to future PR
- [ ] Filtering to specific account UTXOs
- [ ] Consider having a pool of client accounts to avoid re-using UTXOs with low confirmation counts (definitely not 0)
- [ ] Allow different host/port for offline bitcoind based wallet node


hanssv (approved):
Can't see that this would affect the normal node operations, only glanced through the actual bitcoin/dogecoin stuff 👍

Issue #3088, opened 2019-12-19
Unexpected database corruption


dincho:
## Expected Behavior

No database corruption

## Actual Behavior

In case of data storage issues, e.g. very hight latency, reaching open file limits etc. the database is always corrupted.

## Steps to Reproduce the Problem

Set low open files limit

## Logs, error output, etc.

```
2019-12-19 10:25:43.096 [error] <0.14878.13> gen_server aec_conductor terminated with reason: {aborted,{{found_already_calculated_state,<<120,88,27,160,154,99,215,157,247,7,45,255,216,108,25,252,171,178,98,115,98,64,239,48,111,226,97,65,44,130,20,221>>},[{aec_chain_state,update_state_tree,4,[{file,"/home/builder/aeternity/apps/aecore/src/aec_chain_state.erl"},{line,707}]},{aec_chain_state,update_state_tree,2,[{file,"/home/builder/aeternity/apps/aecore/src/aec_chain_state.erl"},{line,698}]},{aec_chain_state,internal_insert_transaction,3,[{file,"/home/builder/aeternity/apps/aecore/src..."},...]},...]}} in mnesia:wrap_trans/6 line 493
```

This is not caused by open file limits but still crashes the nodes in the same sense.

## Specifications

  - Virtualization: AWS
  - Hardware specs:
    * RAM: enough 
    * CPU: slow
  - OS: Ubuntu 14.04
  - Node Version: 5.2.0

dincho:
Additional example:
```
2019-12-19 15:32:31.031 [warning] <0.1507.0>@aec_db:try_activity:925 Mnesia activity Type=transaction exit with Reason={aborted,{badmatch,[]}}, retrying
2019-12-19 15:32:31.064 [error] <0.1507.0> CRASH REPORT Process <0.1507.0> with 0 neighbours exited with reason: {aborted,{badmatch,[]}} in mnesia:wrap_trans/6 line 493
2019-12-19 15:32:31.065 [error] <0.1501.0> Supervisor aecore_sup had child aec_tx_pool started with aec_tx_pool:start_link() at undefined exit with reason {aborted,{badmatch,[]}} in context start_error
[...]

PR #4129, opened 2023-04-27
Make db_direct_access the default

uwiger:
See issue #3088 

The most (_only_?) robust way to avoid database corruption during IO saturation is to run with the `db_direct_access` setting. This makes the rocksdb plugin use rocksdb transactions rather than mnesia transactions. This ensures atomicity in the case when writes are rejected due to congestion.

This is an initial push, partly to ensure that we don't forget to make this the detault.


apps/aecore/src/aec_db.erl
```diff
@@ -1498,6 +1495,32 @@ prepare_mnesia_bypass() ->
...
 
+can_enable_direct_access() ->
+    AllStandalone = all_standalone_tables(),
+    case aeu_env:find_config([<<"chain">>, <<"db_direct_access">>], [ user_config
+                                                                    , schema_default
+                                                                    , {value, false} ]) of
+        {ok, true} ->
+            case AllStandalone of
+                [] ->
+                    lager:debug("Enabling direct access for rocksdb", []),
```
dincho, line 1506: INFO would be more appropriate IMO

dincho:
A release note is somehow important for this change IMO, although it is BC

seanhinde (dismissed):

dincho (dismissed):

PR #4135, opened 2023-05-17
Fixes for db_direct_access + optimizations

uwiger:
Based on debugging of mainnet sync with aggressive GC during sync, a bug was found in `aec_db:ensure_activity/2` for the `db_direct_access` case. The bug caused many operations that should have been dirty to run as transactions instead. While this probably didn't cause any crashes, it would have caused overhead, and complicated debugging.

Transaction retries in case of resource conflict (using Rocksdb's optimistic transaction support) were using the default setting of only one retry. This was changed to 50, and based on debugging output, some transactions were seen to retry up to 43 times.

The primary source of resource contention is still being investigated, but so far, the main suspect is the tx pool gc, interfering with block insertion.


uwiger:
CI failures probably due to the modified tx pool gc behavior (run synchronously from the conductor rather than triggered asynchronously from the aec_tx_pool)

velzevur (approved):

zxq9 (approved):

PR #4146, opened 2023-06-03
Pin down current hyperchain config in schema

seanhinde:
Partly for documentation reasons, but also it feel like this could be generic enough at this point.


loxs (approved):
LGTM

velzevur (approved):
This is compatible with the other currently existing consensus types.

PR #4149, opened 2023-06-10
Add ability to specify aeternity config file on command line

seanhinde:
Seems like this could be quite handy, and by popular request, add the ability to specify the aeternity config file in the start commands:

`bin/aeternity start --config /path/to/ae.yaml`


uwiger (changes requested):
I have no objection, but just wanted to point out for the record that a config file can be identified using
`AETERNITY_CONFIG=<config_file.[yaml | json]> bin/aeternity start`

uwiger (approved):

uwiger:
Sorry, must have hit the wrong checkbox.

PR #4153, opened 2023-06-14
Minor tweaks for HC in json config schema (stricter types).

rumyanar:
Specify parent network connection possible options into enum in schema. Disable additional properties.


dincho:
As it's an enum it does not makes much sense to keep it duplicated in the description I guess

loxs (approved):
Seems to work at least in the sense that the node no more crashes when obeying this schema

seanhinde (approved):
Looks great, thanks

Issue #4157, opened 2023-06-20
Enable mempool sync only when chain sync done


uwiger:
## Is your feature request related to a problem? Please describe.

When syncing against (esp.) mainnet, there are initially lots of info reports warning about failed signature verifications.
This mainly occurs until we reach the Lima hard-fork height where the new signing strategy is activated.

An argument for syncing the mempool during chain sync is that the node owner can inspect the mempool to see the status of given txs. However, it is advisable not to try to use the node before it's fully synced.

Alternatives:
1. Enable mempool sync once a configurable height has been reached
2. Enable mempool sync X number of blocks before what is perceived as the current top
3. Enable mempool sync once `sync_done` is achieved for a given peer.

Alternative no. 3 is the easiest one to implement.

PR #4160, opened 2023-06-27
Gh4157 control when mempool sync starts

uwiger:
See #4157. From the release notes:
> A config option, `mempool:sync_start` (integer) has been introduced to control when syncing of the mempool
  starts during chain sync. A positive number denotes the height at which to begin syncing; a negative number
  denotes how far from the network top height (best guess by the sync logic) to start. Default is `-500`, i.e.
  start syncing the mempool 500 blocks from the top. A mempool sync is always triggered when chain sync is done,
  and if a negative value greater than the top height is given, sync starts from `0`, i.e. from the beginning.

### DIFF
--- a/apps/aecore/src/aec_chain.erl
+++ b/apps/aecore/src/aec_chain.erl
@@ -96,6 +96,11 @@
                          micro_blocks := [aec_blocks:micro_block()],
                          dir := backward | forward }.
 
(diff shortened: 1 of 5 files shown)


zxq9 (approved):
An educational read.

velzevur (approved):

PR #4170, opened 2023-07-05
Gh 4138 rosetta mempool

mitchelli:
Added mempool/transaction endpoint for Rosetta

### DIFF
--- a/apps/aeapi/src/aeapi.erl
+++ b/apps/aeapi/src/aeapi.erl
@@ -38,6 +38,7 @@
         , balance_at_block/2
         , balance_change_events_in_tx/2
         , balance_change_events_in_block/1
+        , balance_change_events_in_mempool_tx/1
         , next_nonce/1
 
         , oracle_at_block/2
@@ -434,6 +435,21 @@ generation_at_height(Height) when is_integer(Height) ->
             end
     end.
 
+balance_change_events_in_mempool_tx(SignedTx) ->
+    Tx = aetx_sign:tx(SignedTx),
+    Nonce = aetx:nonce(Tx),
+    SenderAccount = aetx:origin(Tx),
+    Txs = case Nonce of
+        0 ->
+            [SignedTx];
+        _ ->
+            {ok, UnsortedTxs} = aec_tx_pool:peek(infinity, SenderAccount, Nonce - 1),
+            lists:sort(fun(Left, Right) -> aetx:nonce(aetx_sign:tx(Left)) < aetx:nonce(aetx_sign:tx(Right)) end, [SignedTx|UnsortedTxs])
+    end,
+    Ops = format_txs(Txs, undefined),
+    SpendOps = tx_spend_operations(Ops),
+    lists:last(SpendOps).            
+
 %% Format a single Tx in a block. We need to dry run all the Txs in the microblock
 %% up to and including the one we want to know about. dry run works on dummy signed
 %% Txs so the tx_hash included in the results is not the same as the one requested.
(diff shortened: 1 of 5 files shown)


velzevur (dismissed):
LGTM

apps/aecore/src/aec_tx_pool.erl
```diff
@@ -284,13 +285,20 @@ gproc_reg() ->
...
 %% Only return transactions for a specific account public key
 -spec peek(pos_integer() | infinity, aec_keys:pubkey()) ->
                                        {ok, [aetx_sign:signed_tx()]}.
 peek(MaxN, Account) when is_integer(MaxN), MaxN >= 0; MaxN =:= infinity ->
-    gen_server:call(?SERVER, {peek, MaxN, Account}).
+    gen_server:call(?SERVER, {peek, MaxN, Account, all}).
+
+
+%% Only return transactions for a specific account public key and until a maximum nonce
+-spec peek(pos_integer() | infinity, aec_keys:pubkey(), non_neg_integer() | all) ->
```
hanssv, line 298:
```suggestion
-spec peek(pos_integer() | infinity, aec_keys:pubkey() | all, non_neg_integer() | all) ->
```

apps/aeapi/src/aeapi.erl
```diff
@@ -434,6 +435,21 @@ generation_at_height(Height) when is_integer(Height) ->
...
+balance_change_events_in_mempool_tx(SignedTx) ->
+    Tx = aetx_sign:tx(SignedTx),
+    Nonce = aetx:nonce(Tx),
+    SenderAccount = aetx:origin(Tx),
+    Txs = case Nonce of
+        0 ->
+            [SignedTx];
+        _ ->
+            {ok, UnsortedTxs} = aec_tx_pool:peek(infinity, SenderAccount, Nonce - 1),
+            lists:sort(fun(Left, Right) -> aetx:nonce(aetx_sign:tx(Left)) < aetx:nonce(aetx_sign:tx(Right)) end, [SignedTx|UnsortedTxs])
```
hanssv, line 447: Nonces starts at 1 (0 is reserved for, or at least only used by transactions wrapped in a GAMetaTx) - I don't think this logic is correctly handling GAMetaTx?

hanssv (commented):
See the question about GAMetaTx - I think that might require a change?

apps/aeapi/src/aeapi.erl
mitchelli: The code is to avoid looking in the pool for transactions for an account if the nonce is 0. It should also check for 1. Otherwise it gets a list of transactions for a sender account for a dry run. There are issues with this, for example if there is a contract that has transactions from different accounts. The problem is that it is difficult to know what will happen until the transactions are mined. In the Rosetta documentation it says, "On this endpoint, it is ok that returned transactions are only estimates of what may actually be included in a block.", so I thought this should be good enough. I will do an example with a generalised account to see what happens.

hanssv: Ah, I didn't have the context... So you are trying to estimate how much assets an account will have once the transaction with nonce `Nonce` is included on chain? Well, there is no way to have a generic algorithm for ordering GA-transactions (they all have nonce 0), so skipping them is probably the right thing.

hanssv: This is only ever going to be an estimate, but as long as the user is aware of this I think this implementation is fine!

hanssv (approved):
This is only ever going to be an estimate, but as long as the user is aware of this I think this implementation is fine!

uwiger (approved):

uwiger (approved):

hanssv (approved):

PR #4179, opened 2023-08-04
Add meta information about the currency

mitchelli:


thepiwo (approved):
I don't think there has been a specific wallet integration discussed yet, but the information provided should be sufficient.

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -1562,6 +1562,61 @@
...
+                "display" : {
+                    "type" : "object",
+                    "additionalProperties" : false,
+                    "properties" : {
+                        "logo_file": {
+                            "description": "Logo file of the chain.",
+                            "type": "string"
+                        },
+                        "primary_colour": {
+                            "description": "Short name of currency.",
```
dincho, line 1604: wrong description :)

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -1562,6 +1562,61 @@
...
+                            "type": "string",
+                            "default": "#f5274e"
+                        },
+                        "secondary_colour" : {
+                            "description" : "Smallest denomination of currency.",
```
dincho, line 1609: wrong description

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -1562,6 +1562,61 @@
...
+                            "type" : "string",
+                            "default" : "#242424"
+                        },
+                        "network_id": {
```
dincho, line 1613: Shouldn't it be `network_name`, we already have the network_id in the status endpoint?
And the corresponding default "Mainnet"

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -1562,6 +1562,61 @@
                         }
                     }
                 },
+                "currency" : {
+                    "type" : "object",
+                    "additionalProperties" : false,
+                    "properties" : {
+                        "long_name": {
```
dincho, line 1569: perhaps `long_name` should be just `name` and `short_name` should be symbol 🤔  just a suggestion 💭 

I'm not sure if the properties schema has been agreed somewhere else already or this is first proposition

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -1562,6 +1562,61 @@
...
+                            "description": "Long name of currency.",
+                            "type": "string",
+                            "default": "aeternity"
+                        },
+                        "short_name": {
+                            "description": "Short name of currency.",
+                            "type": "string",
+                            "default": "AE"
+                        },
+                        "smallest_denomination" : {
```
dincho, line 1579: Alternative names would be `subunit` or `minor_units`

- https://en.wikipedia.org/wiki/Denomination_(currency)
- https://english.stackexchange.com/questions/223121/is-there-a-word-for-the-most-granular-parts-of-currency-eg-cents-pence-etc

dincho (commented):
just naming comments

apps/aeutils/priv/aeternity_config_schema.json
mitchelli: Yes, I think this makes more sense, this is just for display purposes.

hanssv:
I don't know the background - should there be defaults for Mainnet and Testnet perhaps?

mitchelli:
> I don't know the background - should there be defaults for Mainnet and Testnet perhaps?

The purpose of this PR was to allow the wallet to get information about a hyper chain but it should also work with the regular chain. I could change the default for network_name to network_id if network_name is not set, rather than "mainnet"?

dincho:
> I don't know the background - should there be defaults for Mainnet and Testnet perhaps?

good point

hanssv (approved):

dincho (approved):

Issue #4175, opened 2023-08-02
Announce capabilities in ping object


uwiger:
In preparation for snapshot and spine sync, we need to have a way to discover which capabilities are supported by a remote node.

For now, the most promising approach seems to be to have nodes announce this in the ping object.
This requires a new ping object version, and here's a suggestion for how to manage this without a hard-fork.

### Starting assumptions
* We need to expect nodes that don't have the latest software, and therefore recognize neither a version 2 ping object nor support any capabilities
* The `aec_peer_connection` code is pretty subtle, so we want to complicate the handshake as little as possible

### Suggestion
* The version 2 ping object includes a `capabilities` attribute, as well as a `max_ping_vsn` entry
* The first ping uses whatever is the configured lowest ping version for the network (default: 1)
* After a successful ping (connection established), iff the first ping was version 1, the connecting node tries sending a version 2 ping; otherwise, it checks the peer's `max_ping_vsn` and re-pings with the highest version supported by both parties.
* If we receive a timeout on the v2 ping request, we stick to v1; otherwise, we register the capabilities announced by the peer and optionally re-ping with the highest mutually supported version.

PR #4188, opened 2023-08-18
Ping v2 and automatic ping vsn upgrade

uwiger:
See issue #4175 

This PR implements the suggestion given in the issue description.

The v2 ping object adds a `versions` and a `capabilities` attribute.

```erlang
serialization_template(ping, ?PING_VSN_2) ->
    [ {versions, [#{items => [ {protocol, binary}
                             , {vsns, [int]}]
                   }]
      }
    , {port, int}
    , {share, int}
    , {genesis_hash, binary}
    , {height, int}                      % V2
    , {difficulty, int}
    , {best_hash, binary}
    , {sync_allowed, bool}
    , {capabilities, binary} % V2
    , {peers, [binary]} ];
```
The `versions` attribute relies on newly added encoding template support. Once serialized, the type is `[[binary, int]]`, but this notation wasn't supported. The `capabilities` payload needs context-sensitive serialization. An example is given with `chain_poi` (capability not yet implemented).

### DIFF
--- a/apps/aecore/src/aec_capabilities.erl
+++ b/apps/aecore/src/aec_capabilities.erl
@@ -36,8 +36,8 @@ delete_capability(Key) ->
 get_capabilities() ->
     call(get_capabilities).
 
-register_peer(Peer, Capabilities) ->
-    call({register_peer, Peer, Capabilities}).
+register_peer(Peer, Capabilities) when is_list(Capabilities) ->
+    call({register_peer, Peer, maps:from_list(Capabilities)}).
 
 remove_peer(Peer) ->
     call({remove_peer, Peer}).
--- a/apps/aecore/src/aec_capabilities.erl
+++ b/apps/aecore/src/aec_capabilities.erl
@@ -1,0 +1,90 @@
+-module(aec_capabilities).
+
+-behavior(gen_server).
+
+
+-export([ set_capability/2
+        , get_capability/1
+        , delete_capability/1
+        , get_capabilities/0
+        , register_peer/2
+        , remove_peer/1
+        , peers_with_capability/1 ]).
+
+%% Gen_server stuff
+-export([ start_link/0 ]).
+
+-export([ init/1
+        , handle_call/3
+        , handle_cast/2
+        , handle_info/2
+        , terminate/2
+        , code_change/3 ]).
+
+start_link() ->
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+
+get_capability(Key) ->
+    call({get_capability, Key}).
+
+set_capability(Key, Value) ->
+    call({set_capability, Key, Value}).
+
+delete_capability(Key) ->
+    call({delete_capability, Key}).
+
+get_capabilities() ->
+    call(get_capabilities).
+
+register_peer(Peer, Capabilities) ->
+    call({register_peer, Peer, Capabilities}).
+
+remove_peer(Peer) ->
+    call({remove_peer, Peer}).
+
+peers_with_capability(Key) ->
+    call({peers_with_capability, Key}).
+
+call(Req) ->
+    gen_server:call(?MODULE, Req).
+
+
+init([]) ->
+    {ok, #{ peers => #{}
+          , capabilities => #{} }}.
+
+handle_call({get_capability, Key}, _From, #{capabilities := Cs} = St) ->
+    {reply, maps:find(Key, Cs), St};
+handle_call({set_capability, Key, Value}, _From, #{capabilities := Cs} = St) ->
+    {reply, ok, St#{capabilities := Cs#{Key => Value}}};
+handle_call({delete_capability, Key}, _From, #{capabilities := Cs} = St) ->
+    {reply, ok, St#{capabilities := maps:remove(Key, Cs)}};
+handle_call(get_capabilities, _From, #{capabilities := Cs} = St) ->
+    {reply, maps:to_list(Cs), St};
+handle_call({register_peer, PeerId, Capabilities}, _From, #{peers := Peers} = St) ->
+    {reply, ok, St#{peers := Peers#{PeerId => Capabilities}}};
+handle_call({peers_with_capability, Key}, _From, #{peers := Peers} = St) ->
+    Result = maps:fold(fun(P, Cs, Acc) ->
+                               peers_with_cap_(P, Key, Cs, Acc)
+                       end, [], Peers),
+    {reply, Result, St}.
+
+handle_cast(_, St) ->
+    {noreply, St}.
+
+handle_info(_, St) ->
+    {noreply, St}.
+
+terminate(_Reason, _State) ->
+    ok.
+
+code_change(_FromVsn, St, _Extra) ->
+    {ok, St}.
+
+peers_with_cap_(PeerId, Key, PeerCaps, Acc) ->
+    case maps:find(Key, PeerCaps) of
+        {ok, Value} ->
+            [{PeerId, Value} | Acc];
+        error ->
+            Acc
+    end.
--- a/apps/aecore/src/aec_capabilities.erl
+++ b/apps/aecore/src/aec_capabilities.erl
@@ -64,8 +64,9 @@ handle_call(get_capabilities, _From, #{capabilities := Cs} = St) ->
 handle_call({register_peer, PeerId, Capabilities}, _From, #{peers := Peers} = St) ->
     {reply, ok, St#{peers := Peers#{PeerId => Capabilities}}};
 handle_call({peers_with_capability, Key}, _From, #{peers := Peers} = St) ->
-    Result = maps:fold(fun(P, Cs, Acc) ->
-                               peers_with_cap_(P, Key, Cs, Acc)
+    Result = maps:fold(fun(PeerId, #{Key := Caps}, Acc) ->
+                            [{PeerId, Caps}|Acc];
+                           (_,_, Acc) -> Acc
                        end, [], Peers),
     {reply, Result, St}.
 
@@ -80,11 +81,3 @@ terminate(_Reason, _State) ->
 
 code_change(_FromVsn, St, _Extra) ->
     {ok, St}.
-
-peers_with_cap_(PeerId, Key, PeerCaps, Acc) ->
-    case maps:find(Key, PeerCaps) of
-        {ok, Value} ->
-            [{PeerId, Value} | Acc];
-        error ->
-            Acc
-    end.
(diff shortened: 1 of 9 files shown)


zxq9 (approved):
LGTM

rebar.config
```diff
@@ -74,7 +74,7 @@
                      {tag, "v3.4.0"}}},
 
         {aeserialization, {git, "https://github.com/aeternity/aeserialization.git",
-                          {tag,"v1.0.0"}}},
```
hanssv, line 77: I think we should aim to tag the new version before merging this...

apps/aecore/src/aec_peer_messages.erl
```diff
@@ -199,7 +223,40 @@ latest_vsn(get_node_info)        -> ?GET_NODE_INFO_VSN;
...
+%% Default is to support only the latest version.
+msg_versions(ping) ->
+    {ok, Min} = aeu_env:find_config([<<"sync">>, <<"minimum_ping_version">>],
+				    [user_config, schema_default]),
+    [V || V <- [?PING_VSN_1, ?PING_VSN_2],
+	  V >= Min];
+msg_versions(_Msg) ->
+    undefined.
+
+all_msg_versions() ->
```
hanssv, line 237: Not sure I understand the purpose and/or the name of this function, it is not all messages?

uwiger: I'll revisit. May well be remnants of me trying to make up my mind on how to structure things.

apps/aecore/src/aec_peer_messages.erl
```diff
@@ -88,7 +109,7 @@ serialize(peers, Peers, Vsn = ?VSN_1) ->
     [ serialize(peer, Peer, Vsn) || Peer <- Peers ];
 serialize(peer, Peer, ?VSN_1) ->
     %% Don't pollute the encoding with lots of Vsn-fields...
-    Template = serialization_template(peer, ?PING_VSN),
+    Template = serialization_template(peer, ?VSN_1),
```
ThomasArts, line 112:
```suggestion
    Template = serialization_template(peer, ?PING_VSN_1),
```

mitchelli: Thanks, there is actually PEER_VSN already defined in aec_peer_messages.hrl. I'll change it to be more consistent. The ping serialisation already uses PING_VSN_2 and PING_VSN_1.

dincho (commented):
I'd like to see some documentation (in the protocol repo) in regards to this new functionality. I usually monitor the p2p network and also I've implemented a JS version of the protocol, so any serialization docs has to be updated as well.

mitchelli:
> I'd like to see some documentation (in the protocol repo) in regards to this new functionality. I usually monitor the p2p network and also I've implemented a JS version of the protocol, so any serialization docs has to be updated as well.

Thanks for the pointer, I'll add some documentation. For the moment, as long as your JS version can ignore messages it doesn't understand you won't have an issue. What will happen is that your monitor will receive a version 1 ping followed by a version 2 ping, if it doesn't respond to the version 2 ping the node will fallback to version 1.

Which reminds me, I will need to check what happens if the node receives a capability it doesn't understand.

apps/aecore/src/aec_peer_messages.erl
mitchelli: I've renamed it to supported_msg_versions is that clearer?

uwiger:
> Which reminds me, I will need to check what happens if the node receives a capability it doesn't understand.

The initial idea, at least, was that a node can advertise capabilities, and the peer decides whether to make use of any of them. For example, a node serving sync snapshots would advertise this capability, including the height range for which snapshots are served. Another future capability might be `archive`, where the node offers to provide legacy data from older generations.

In short, a node shouldn't worry about capabilities it doesn't understand, but may choose to draw upon capabilities that it has a use for.

mitchelli:
> > Which reminds me, I will need to check what happens if the node receives a capability it doesn't understand.
> 
> The initial idea, at least, was that a node can advertise capabilities, and the peer decides whether to make use of any of them. For example, a node serving sync snapshots would advertise this capability, including the height range for which snapshots are served. Another future capability might be `archive`, where the node offers to provide legacy data from older generations.
> 
> In short, a node shouldn't worry about capabilities it doesn't understand, but may choose to draw upon capabilities that it has a use for.

Thanks, I have changed the serialisation to ignore capabilities it doesn't understand.

uwiger:
> Thanks, I have changed the serialisation to ignore capabilities it doesn't understand.

I haven't had time to look at the latest changes, but capability metadata should be serialized in a way that all nodes (which support PING_VSN_2) understand, even if they don't understand, or care about, the specific semantics of a certain capability. It should also be understood that capability announcement in the ping object should be minimalistic, since ping objects are repeated. Thus, if the user of a capability needs more detail in order to actually use it, it would be best if that is pushed to the next level, with the user getting just enough info in the ping to know that it's worth the effort to request more detail.

mitchelli:
> I haven't had time to look at the latest changes, but capability metadata should be serialized in a way that all nodes (which support PING_VSN_2) understand, even if they don't understand, or care about, the specific semantics of a certain capability.

Sorry, I meant to write I changed the deserialisation to ignore capabilities that are not understood. The format hasn't changed.

dincho:
> can ignore messages it doesn't understand you won't have an issue.

The thing is that I don't to ignore them but to implement the new protocol features and monitor them, thus I need them fully documented :)

mitchelli:
> > can ignore messages it doesn't understand you won't have an issue.
> 
> The thing is that I don't to ignore them but to implement the new protocol features and monitor them, thus I need them fully documented :)

I am updating p2p_messages.md in the protocol repo at the moment.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -603,55 +639,76 @@ handle_msg(S, MsgType, false, Request, {ok, Vsn, Msg}) ->
...
         txps_finish   -> handle_tx_pool_sync_rsp(S, finish, Request, Msg);
-        node_info     -> handle_node_info_rsp(S, Request, Msg)
+        node_info     -> handle_node_info_rsp(S, Request, Vsn, Msg)
     end.
 
 handle_request(S, {Request, Args}, From) ->
     MappedRequest = map_request(Request),
-    case get_request(S, Request) of
+    Vsn = case Request of
+            tx_pool ->
```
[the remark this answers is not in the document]
mitchelli: tx_pool is a series of messages which doesn't have a version.

```
map_request(txps_init)            -> tx_pool;
map_request(txps_unfold)      -> tx_pool;
map_request(txs)                     -> tx_pool;
map_request(txps_finish)        -> tx_pool;

```
there is no latest_vsn in aec_peer_messages. I originally set a artificial version in msg_version function but I wasn't sure if it was clear that it wasn't a real message.

apps/aecore/src/aec_capabilities.erl
```diff
@@ -0,0 +1,90 @@
...
+set_capability(Key, Value) ->
+    call({set_capability, Key, Value}).
+
+delete_capability(Key) ->
+    call({delete_capability, Key}).
+
+get_capabilities() ->
+    call(get_capabilities).
+
+register_peer(Peer, Capabilities) when is_list(Capabilities) ->
```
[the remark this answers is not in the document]
mitchelli: It will be used as a map, currently there is a sample capability which could be used for syncing

```
capabilities_template(<<"chain_poi">>, ?PING_VSN_2) ->
    [{chain_poi, #{ items => [ {height, int}
                             , {root_hash, binary}
                             , {genesis, binary}
                             , {top, binary}
                             , {poi, binary} ]}
     }];
```

apps/aecore/src/aec_capabilities.erl
```diff
@@ -0,0 +1,90 @@
...
+handle_call({set_capability, Key, Value}, _From, #{capabilities := Cs} = St) ->
+    {reply, ok, St#{capabilities := Cs#{Key => Value}}};
+handle_call({delete_capability, Key}, _From, #{capabilities := Cs} = St) ->
+    {reply, ok, St#{capabilities := maps:remove(Key, Cs)}};
+handle_call(get_capabilities, _From, #{capabilities := Cs} = St) ->
+    {reply, maps:to_list(Cs), St};
+handle_call({register_peer, PeerId, Capabilities}, _From, #{peers := Peers} = St) ->
+    {reply, ok, St#{peers := Peers#{PeerId => Capabilities}}};
+handle_call({peers_with_capability, Key}, _From, #{peers := Peers} = St) ->
+    Result = maps:fold(fun(P, Cs, Acc) ->
```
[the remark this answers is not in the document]
mitchelli: Thanks, I've changed it to use the pattern match.

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -453,29 +475,41 @@ handle_general_error(S) ->
...
-    TRef = erlang:start_timer(Timeout, self(), {request, Kind}),
-    S#{ requests => Rs#{ Kind => {Kind, From, TRef} } }.
+    TRef = erlang:start_timer(Timeout, self(), {request, Kind, Vsn}),
+    S#{ requests => Rs#{ request_key(Kind, Vsn) => {Kind, From, TRef} } }.
 
-remove_request_fld(S, Kind) ->
+remove_request_fld(S, Kind, Vsn) ->
     Rs = maps:get(requests, S, #{}),
-    case get_request(S, Kind) of
+    case get_request(S, Kind, Vsn) of
```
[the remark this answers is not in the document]
mitchelli: Changed to use maps:take

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1196,13 +1324,14 @@ handle_light_micro_block(_S, Header, TxHashes, PoF) ->
 expand_micro_block(State, #{ header := Header,
                              tx_data := TxsAndTxHashes } = MicroBlockFragment) ->
     {ok, Hash} = aec_headers:hash_header(Header),
-    case get_request(State, {expand, Hash}) of
+    Vsn = aec_headers:version(Header),
```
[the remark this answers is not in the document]
mitchelli: Thanks, that looks wrong. I'll change it.

mitchelli:
> Looks good I think - it is tricky stuff and somewhat hard to wrap your head around.
> 
> Would it make sense to have a system test with a couple of old nodes and a couple of new nodes? 🤔

There is already a test for when a new node joins a network of old nodes, I could extend that or add a new test.

hanssv (approved):

PR #4191, opened 2023-08-23
Add endpoint to get recent gas prices

davidyuk:
needed to implement https://github.com/aeternity/aepp-sdk-js/issues/1786

At some point, we would have a such blockchain load that won't allow us to mine all submitted transactions with the minimum fee. Then the gas price used to build a transaction would matter. To be able to estimate a reasonable gas price I'm proposing to implement an endpoint on the node side exposing the minimum gas price among the last blocks.

Screenshot 2023-08-23 at 20 01 14

### DIFF
--- a/apps/aehttp/src/aehttp_logic.erl
+++ b/apps/aehttp/src/aehttp_logic.erl
@@ -14,6 +14,7 @@
         , get_block_genesis/0
         , get_genesis_hash/0
         , get_top_blocks_time_summary/1
+        , get_top_blocks_gas_price_summary/0
         ]).
 
 -export([ get_account/1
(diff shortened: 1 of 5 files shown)


apps/aehttp/src/aehttp_logic.erl
```diff
@@ -160,3 +161,61 @@ get_top_blocks_time_summary(Count) ->
...
+        fun(BlockAmount) ->
+            HeightBottom = Height - BlockAmount + 1,
+            GasPrice =
+                case HeightBottom < 1 of
+                    true -> aec_tx_pool:minimum_miner_gas_price();
+                    false -> get_blocks_min_gas_price(Height, HeightBottom)
+                end,
+            [BlockAmount, GasPrice]
+        end,
+        [1, 5, 20, 120, 480]).
```
davidyuk, line 177: It should be 3 minutes, 15 minutes, 1 hour, 6 hours, and 1 day respectively.

hanssv (commented):
Looks like a valuable addition, but I'm slightly worried about performance? Any numbers on how this would perform on a crowded chain? Maybe we should cache it? At least we should compute once and forall per generation?

davidyuk:
I had a plan to calculate it once for a generation, but it is too difficult. Maybe someone can take over this PR? 🙏  This is more POC and API needed for sdk.

hanssv:
Replaced by #4277.

PR #4196, opened 2023-08-28
Cache core schema as persistent_term

uwiger:
Profiling indicates that schema default lookup when checking config values is inordinately expensive.

This PR caches the parsed schema as a persistent term rather than in ets.
This brings down schema default lookup time from ca 4 ms to ca 4 us.

(Plugin schemas were already cached as persistent terms)

This affects calls like:
```erlang
aeu_env:find_config( [<<"sync">>, <<"peer_analytics">>],
                     [user_config, schema_default, {value,false}]).
```

If the key is not present in the user config, the schema default will be returned. This is actually quite a common pattern.

### DIFF
--- a/apps/aeutils/src/aeu_env.erl
+++ b/apps/aeutils/src/aeu_env.erl
@@ -218,14 +218,22 @@ lists_map_key_find(_, []) ->
     error.
 
 schema() ->
-    case setup:get_env(aeutils, '$schema', undefined) of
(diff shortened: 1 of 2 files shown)


hanssv (dismissed):

hanssv (approved):

mitchelli (approved):
LGTM

PR #4198, opened 2023-08-30
Cache plugin config after validation

uwiger:
Based on feedback from @nikita-fuchs, it was found that `aeu_plugins:find_config/3` didn't actually work.
The user config, when defined using OS environment variables, was validated against the schema, but then thrown away.
This PR caches the validated user config and uses that during lookups.


uwiger:
Closing. Caching added as part of PR #4053

PR #4200, opened 2023-08-31
Add ability to configure directories for hard forks

mitchelli:


zxq9 (approved):
LGTM

hanssv (approved):

uwiger (dismissed):

dincho (changes requested):
This needs documentation, schema and reference config update?

apps/aeutils/test/data/epoch_full.yaml
```diff
@@ -138,8 +138,12 @@ chain:
     db_path: .
     # Hard forks
     hard_forks:
-      "1": 0
-      "2": 1
+      "1":
+        height: 0
+        use_hardcoded_directory : true
```
dincho, line 143: Why this explicit option is needed? What happens if directory is not provided, couldn't be that used to default to the hardcoded dirs?

apps/aecore/test/aec_fork_block_settings_tests.erl
```diff
@@ -12,38 +13,70 @@
...
 lima_accounts_test_() ->
     release_based(?ROOT_DIR ++ "/.lima",
+                  undefined,
                   fun ?TEST_MODULE:lima_accounts/0,
                   fun ?TEST_MODULE:lima_contracts/0,
                   lima_contracts_file_missing,
                   lima_accounts_file_missing).
 
-release_based(Dir, ReadAccountsFun, ReadContractsFun, CMissingErr, AMissingErr) ->
+generic_accounts_test_() ->
```
dincho, line 46: What the term "generic" means?

apps/aeutils/test/data/epoch_full.yaml
```diff
@@ -138,8 +138,12 @@ chain:
...
+      "2":
+        height: 1
+        directory: "/tmp/1"
```
dincho, line 146: Does it also allow relative to the node root path?

dincho:
> This needs documentation, schema and reference config update?

Did I overlooked the schema and configs yesterday 🤔

dincho:
How the configuration with ENV variables would look like?

dincho:
Also I think it worth finally being documented in configuration.md or maybe special custom_network doc? 🤔

dincho:
Good idea in general, much better than what we currently have, but I'd like to decided on file names as well, magic names caused me a lot of troubles in the past couple of years. While one would think it's just names - no, once a file path is specified in the configuration the use will be pretty much sure that it will be used or the node should stop with error, in case of miss-typed, non-existing etc.

I believe the proposed implementation still keeps going if some of the files are missing, on the other side requiring empty files is also stupid.

I propose explicit file path configuration:

```
    hard_forks:
      "1":
        height: 0
        accounts_file: /path/to/accounts.json
        contracts_file: hc_contracts.json # note the relativeness (to the node root)
        other_file: "do we have other files" ?
```

apps/aecore/test/aec_fork_block_settings_tests.erl
mitchelli: It is meant to mean the opposite of hard coded. I could change it to configurable.

apps/aeutils/test/data/epoch_full.yaml
mitchelli: That's possible but if you have custom blockchain, for example a hyperchain, you would either need to avoid using the same protocol versions as the mainnet or provide an empty directory. mainnet hasn't checked for account or contract files since Lima.

apps/aeutils/test/data/epoch_full.yaml
mitchelli: No I can add that.

mitchelli:
> Good idea in general, much better than what we currently have, but I'd like to decided on file names as well, magic names caused me a lot of troubles in the past couple of years. While one would think it's just names - no, once a file path is specified in the configuration the use will be pretty much sure that it will be used or the node should stop with error, in case of miss-typed, non-existing etc.
> 
> I believe the proposed implementation still keeps going if some of the files are missing, on the other side requiring empty files is also stupid.
> 
> I propose explicit file path configuration:
> 
> ```
>     hard_forks:
>       "1":
>         height: 0
>         accounts_file: /path/to/accounts.json
>         contracts_file: hc_contracts.json # note the relativeness (to the node root)
>         other_file: "do we have other files" ?
> ```



> Good idea in general, much better than what we currently have, but I'd like to decided on file names as well, magic names caused me a lot of troubles in the past couple of years. While one would think it's just names - no, once a file path is specified in the configuration the use will be pretty much sure that it will be used or the node should stop with error, in case of miss-typed, non-existing etc.
> 
> I believe the proposed implementation still keeps going if some of the files are missing, on the other side requiring empty files is also stupid.
> 
> I propose explicit file path configuration:
> 
> ```
>     hard_forks:
>       "1":
>         height: 0
>         accounts_file: /path/to/accounts.json
>         contracts_file: hc_contracts.json # note the relativeness (to the node root)
>         other_file: "do we have other files" ?
> ```

I think that is reasonable but then should the changes also be made to the hard coded configuration to avoid the magic names?

dincho:
> I think that is reasonable but then should the changes also be made to the hard coded configuration to avoid the magic names?



> I think that is reasonable but then should the changes also be made to the hard coded configuration to avoid the magic names?

My opinion is that users shouldn't be able to change the hard_forks configuration of predefined networks `ae_uat` and `ae_mainnet` but this decision needs more clever ppl than me like @hanssv  :)

hanssv:
> > I think that is reasonable but then should the changes also be made to the hard coded configuration to avoid the magic names?
> 
> > I think that is reasonable but then should the changes also be made to the hard coded configuration to avoid the magic names?
> 
> My opinion is that users shouldn't be able to change the hard_forks configuration of predefined networks `ae_uat` and `ae_mainnet` but this decision needs more clever ppl than me like @hanssv :)

I agree, those two networks are all defined in the code, and those should not be overridable - for other networks the above looks good to me!

apps/aecore/src/aec_fork_block_settings.erl
```diff
@@ -81,50 +61,17 @@ lima_contracts() -> preset_contracts(?LIMA_PROTOCOL_VSN,
...
-        <<"use_hardcoded_dir">> -> false;
-        _ ->
-            Default
-    end.
+contracts(ProtocolVsn) -> preset_contracts(ProtocolVsn, contracts_file_missing).
 
 -spec is_custom_fork() -> boolean().
 is_custom_fork() ->
-    is_map(aeu_env:user_map_or_env([<<"chain">>, <<"hard_forks">>], aecore, hard_forks, undefined)).
+    is_map(aeu_env:config_value([<<"chain">>, <<"hard_forks">>], aecore, hard_forks, undefined)).
```
dincho, line 74: I'm not familiar with `aeu_env` implementation bug the function name change concerns me a bit, does it still considers ENV variables ?

dincho: Just checked, seems to be the case still

mitchelli: I didn't realise either. Ulf explains the difference in the pull request here: https://github.com/aeternity/aeternity/pull/4211

mitchelli: Maybe I should change it back. There won't be any schema defaults.

nikita-fuchs:
hey @mitchelli , could you please add a description of this new feature in the docs and provide information and examples on their usage?

dincho (commented):
Can't reason about the erlang code, but functionally I'm happy with the proposition.

docs/configuration.md
```diff
@@ -143,6 +143,39 @@ fork_management:
...
+
+For hyperchains a contact file can also be specified alongside the other files.
+```yaml
+chain:
+    persist: true
+    hard_forks:
+        "genesis":
+            "height": 0
+            "accounts_file": hf_dir/genesis/accounts.json
+            "hc_contracts_file": hf_dir/genesis/contracts.json
```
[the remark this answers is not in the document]
mitchelli: The contracts file is applied when a certain height is hit but Lima is the last time it was used. The hyperchain contracts are called from hyperchain specific parts of the code and the format returned is different. I thought it might be needed to apply both types of contracts.

dincho (changes requested):
Needs proper rebase and addressing double contracts option

docs/configuration.md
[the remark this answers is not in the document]
mitchelli: Option 1 sounds like a good idea. At the moment, the consensus contracts are only called when creating the genesis block but I remember it was supposed to be possible to change the consensus. This would make it possible to change the consensus but I am not sure if anymore work would be needed.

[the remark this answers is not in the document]
mitchelli: The contracts are applied differently. The lima hard fork contract states "Note that there are no call objects, and no transactions recorded on the chain, and no fees or gas compensation, is given for this execution." The functionality is "to enable distributing funds to accounts that missed the final migration deadline". At the moment, I check at every protocol change if there is a contracts file and if so apply the functionality above that was added with Lima. I could either not apply the contracts file after Lima or try to distinguish between the consensus and account migration contracts. The account migration contracts contains no owner, I could use that fact, or have the format of the file specify what functionality the contract supports.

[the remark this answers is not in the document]
dincho: I don’t think we need to overwrite that then - “the old contracts”

dincho (approved):
LGTM from user/configuration perspective. Please test sync with mainnet and testnet from 0

dincho (approved):

hanssv (approved):

PR #4201, opened 2023-09-04
[Ceres] Make NamePreclaim optional

### LINKED ISSUE #4174: Allow mining NameClaimTx immediately
> A claim transaction MUST NOT be in included in the same block as its pre-claim.

https://github.com/aeternity/protocol/blob/master/AENS.md#claim

I'm not sure about motivation of the above. Usually, we send pre-claim tx, and right after it gets a block number (added to a microblock) we send a name claim. So, the name gets exposed almost immediately. Does it make sense to wait for a keyblock on the protocol level in this case? Won't be the same secure to delay NameClaim exposing on app side as defining a 1 block gap on the protocol level?

If 1 block gap in the protocol is necessary then does it make sense to support it for other transactions, like contract call? For example, if I'm creating an nft I may want to be sure that nobody else can create the same before me. To have better microfork resistance, this gap should be adjustable, the same as optional (for contract calls, unsafe/fast name claim). I would remove the 1 block gap restriction and add a field like "reverse ttl" -- to allow transaction mining after a specific height.

hanssv:
Fixes #4174 

Make NamePreclaimTx optional. Since the introduction of auctions, the front-running protection offered by the 'PreClaim -> Claim' flow is no longer as important. To simplify name registrations (or the start of a name auction) we now allow NameClaimTx without a previous NamePreclaimTx. In this case we set the `NameSalt` to `0`.

### DIFF
--- a/apps/aens/test/aens_SUITE.erl
+++ b/apps/aens/test/aens_SUITE.erl
@@ -21,16 +21,16 @@
          prune_claim/1,
          prune_claim_auction/1,
          preclaim_negative/1,
-         claim/1,
-         claim_auction/1,
+         claim/1, immediate_claim/1,
          ongoing_auction/1,
          second_claim_auction/1,
          second_claim_extend_auction/1,
          claim_locked_coins_holder_gets_locked_fee/1,
-         claim_negative/1,
+         claim_negative/1, claim_negative2/1,
          claim_empty_name/1,
          claim_race_negative/1,
          claim_race_negative_auction/1,
+         immediate_claim_race/1, immediate_claim_race2/1,
          update/1,
          update_negative/1,
          update_expire_at_once/1,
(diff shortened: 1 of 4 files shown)


apps/aens/test/aens_SUITE.erl
```diff
@@ -577,6 +613,35 @@ claim_negative(Cfg) ->
...
+    Protocol = aec_hard_forks:protocol_effective_at_height(1),
+    if Protocol >= ?CERES_PROTOCOL_VSN -> {skip, preclaimless_claim_in_ceres};
+       true                            -> claim_negative2_(Cfg)
+    end.
+
+claim_negative2_(Cfg) ->
+    State = case proplists:get_value(state, Cfg) of
+                undefined -> aens_test_utils:new_state();
+                State0 -> State0
+            end,
```
uwiger, line 626: Seems like this could be factored out into a helper function.

uwiger (approved):
LGTM

zxq9 (approved):
LGTM

PR #4207, opened 2023-11-01
Add current hashrate to /status response

hanssv:
Inspired by a recent discussion with Dincho, let's just add this value to the status such that we won't have to remember the details 😅 


uwiger (dismissed):
LGTM

uwiger (approved):

dincho (approved):

Issue #4209, opened 2023-11-06
aecore_forking_SUITE fails intermittently testing orphaned txs


uwiger:
## Expected Behavior

Passing test suite

## Actual Behavior

Various different occasional failures, example:

```
%%% aecore_forking_SUITE ==> all_nodes.orphaned_txs_get_included.on_key_block.spend_on_dev1: FAILED
%%% aecore_forking_SUITE ==> {{badmatch,
     {ok,[{signed_tx,
              {aetx,spend_tx,aec_spend_tx,86,
                  {spend_tx,
                      {id,account,
                          <<217,202,108,173,192,99,13,10,129,124,71,86,232,
                            121,148,177,243,254,160,88,174,204,22,114,15,42,
                            51,71,75,19,135,16>>},
                      {id,account,
                          <<217,202,108,173,192,99,13,10,129,124,71,86,232,
                            121,148,177,243,254,160,88,174,204,22,114,15,42,
                            51,71,75,19,135,16>>},
                      1,20000000000000,0,1,<<"foo">>}},
              [<<103,144,211,190,11,94,97,192,238,148,8,216,125,226,20,
                 221,98,47,29,39,152,210,171,151,221,61,59,75,223,10,129,
                 181,117,253,92,69,21,69,196,151,210,50,102,116,80,20,193,
                 34,156,9,200,198,64,215,12,194,140,150,254,220,230,201,
                 104,9>>]}]}},
 [{aecore_forking_SUITE,spend,1,
      [{file,
           "/home/builder/aeternity/apps/aecore/test/aecore_forking_SUITE.erl"},
       {line,459}]},
```

## Steps to Reproduce the Problem

  1. Run aecore_forking_SUITE primarily in CI

uwiger:
CI stability issue (see #3475)

PR #4210, opened 2023-11-06
Ensure tx actually makes it on-chain in test

uwiger:
See issue #4209 

As a first measure, ensure that the tx actually makes it onto the chain before stopping the node.

### DIFF
--- a/apps/aecore/test/aecore_forking_SUITE.erl
+++ b/apps/aecore/test/aecore_forking_SUITE.erl
@@ -27,6 +27,7 @@
     spend_on_dev2/1,
     mine_a_micro_block_on_dev1/1,
     mine_a_micro_block_on_dev2/1,
+    mine_micro_blocks_on_dev1_until_empty/1,
     mine_micro_blocks_on_dev2_until_empty/1,
     start_nodes_and_wait_sync_dev1_chain_wins/1,
     start_nodes_and_wait_sync_dev2_chain_wins/1,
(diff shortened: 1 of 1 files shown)


apps/aecore/test/aecore_forking_SUITE.erl
```diff
@@ -437,6 +438,11 @@ mine_a_micro_block(Node) ->
 mine_a_micro_block_on_node(NName) ->
     aecore_suite_utils:mine_blocks(NName, 1, ?MINE_RATE, micro, #{}).
 
+mine_micro_blocks_on_dev1_until_empty(_Config) ->
```
hanssv, line 441: Maybe generalize with the function below? But the naming in this test module is very explicit...

hanssv (commented):
LGTM

apps/aecore/test/aecore_forking_SUITE.erl
uwiger: Yes, I tried to stay consistent with the existing convention

uwiger: I think the idea is that since the test flows are actually implemented at the group level, specifically named test cases make those flows more readable.

uwiger:
Lots of restarts of the CI yielded no failures in the forking_SUITE. Several failures in the staking_contract_SUITE, however.

hanssv (approved):

dincho (approved):

PR #4213, opened 2023-11-07
Fix intermittent test failure in aecore_sync_SUITE

hanssv:
Make sure to generate non-overlapping random ports. Also fix cleanup when tests fail (we sometimes have to start nodes to clean up properly).

Also fix some whitespace.

### DIFF
--- a/apps/aecore/test/aecore_sync_SUITE.erl
+++ b/apps/aecore/test/aecore_sync_SUITE.erl
@@ -83,7 +83,7 @@ groups() ->
                               %% the peers
                               %% TODO: investigate why
                               {group, persistence}
-                              
+
                              ]},
      {two_nodes, [sequence],
       [start_first_node,
@@ -166,8 +166,7 @@ groups() ->
            start_with_trusted_peers,
            restart_with_no_trusted_peers,
            add_and_delete_untrusted_peers_and_restart,
-           trusted_peer_is_untrusted_after_a_restart,
-           stop_devs %% delete the db to cleanup
+           trusted_peer_is_untrusted_after_a_restart
       ]},
      {node_info, [sequence],
       [%% all nodes respond:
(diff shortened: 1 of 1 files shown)


uwiger (approved):
LGTM

apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -1245,9 +1244,23 @@ encode_peer_for_config(Peer) ->
...
 
-random_peer() ->
+%% Make sure to generate non-overlapping ports...
+random_peers(N) ->
+    Ports = random_ports(N, []),
+    [ random_peer(Port) || Port <- Ports ].
+
+random_ports(0, Ports) -> Ports;
+random_ports(N, Ports) ->
+  Port = 4000 + rand:uniform(999),
```
dincho, line 1254: I'd go even further up to ~60000 for a pinch of more randomness

dincho (approved):
🤞🏻

apps/aecore/test/aecore_sync_SUITE.erl
```diff
@@ -1245,9 +1244,23 @@ encode_peer_for_config(Peer) ->
     Port = integer_to_binary(aec_peer:port(Peer)),
     <<"aenode://", PK/binary, "@", Host/binary, ":", Port/binary>>.
 
-random_peer() ->
+%% Make sure to generate non-overlapping ports...
+random_peers(N) ->
```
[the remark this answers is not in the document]
hanssv: You are correct, but each test case is separate as far as I could tell so it is enough if each invocation is unique

I tried to run this with `rand:uniform(4)` and it worked!

apps/aecore/test/aecore_sync_SUITE.erl
hanssv: Probably not neecessary, I don't even think the randomness brings _anything_ to the table. We could just as well have used 4001-4004 sequentially I think.

PR #4217, opened 2023-11-16
Make sure build and package works on OTP26

### LINKED ISSUE #4180: OTP26 build fails due to `argparse` dependency
The [new stdlib](https://www.erlang.org/doc/apps/stdlib/notes.html#improvements-and-new-features) includes a module called `argparse`, just like one of [our dependencies](https://github.com/max-au/argparse). If you try to build the node on OTP26, then you get a clash:

```
===> Error generating release: 
Duplicated modules: 
    argparse specified in stdlib and argparse
```

While we don't officially support OTP26 now, this will have to be addressed at some point.

hanssv:
Fixes #4180 


mitchelli (approved):

dincho (approved):

hanssv:
This version won't work, sadly. While the rebar3 sorcery is cool, the actual `argparse` library included in OTP26 is subtly different, so we'd still have to keep to variants around...

Let's do #4218 instead...

PR #4227, opened 2023-11-30
Fix /status for HC

hanssv:

### DIFF
--- a/apps/aehttp/src/aehttp_dispatch_ext.erl
+++ b/apps/aehttp/src/aehttp_dispatch_ext.erl
@@ -647,10 +647,11 @@ handle_request_('GetPeerPubkey', _Params, _Context) ->
 
 handle_request_('GetStatus', _Params, _Context) ->
     {ok, TopKeyBlock} = aec_chain:top_key_block(),
+    Consensus = aec_blocks:consensus_module(TopKeyBlock),
     GenesisBlockHash = aec_consensus:get_genesis_hash(),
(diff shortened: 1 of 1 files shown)


dincho (approved):

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -771,10 +772,20 @@ deserialize_transaction(Tx) ->
...
+target_to_hashrate(Target, aec_consensus_bitcoin_ng) ->
     case aeminer_pow:scientific_to_integer(Target) of
       0 -> 0;
       _ ->
           Difficulty = aeminer_pow:target_to_difficulty(Target) / (1 bsl 24),
           round(Difficulty * (aec_governance:expected_block_mine_rate() / 1000) * 42)
-    end.
+    end;
+target_to_hashrate(_Target, _Consensus) ->
+    0.
```
velzevur, line 783: I believe CT tests use consensus `aec_consensus_common_tests` so this could be a bit unexpected a test if it wants to pretend to mine something. No such test I am aware of, though

velzevur (approved):

apps/aehttp/src/aehttp_dispatch_ext.erl
hanssv: Good point... This value only makes sense on Mainnet/Testnet so I think we are fine with 0.

PR #4249, opened 2024-01-16
[Ceres] Adjust cost for reading stored value

### LINKED ISSUE #4197: Ceres (v6) gas consumption much higher than on Iris (v5)
## Expected Behavior
There shouldn't be any difference in gas consumption when executing the same entrypoints on the exact same contract which has previously been deployed on Iris (v5).

## Actual Behavior
Apparently the gas consumption on Ceres is ~3x higher than on Iris for the following contract:
- https://github.com/marc0olo/aens-nfts/blob/ceres-playground/development/smart-contracts/contracts/AENSWrapping.aes

### Iris (v5)
```sh
Gas used (unwrap_all for 3 names): 52494
        ✔ unwrap_all (1443ms)
```

### Ceres (v6)
```sh
Gas used (unwrap_all for 3 names): 172494
        ✔ unwrap_all (1467ms)
```

Execute tests in [ceres-playground](https://github.com/marc0olo/aens-nfts/blob/ceres-playground) and [iris-only-gas-consumption](https://github.com/marc0olo/aens-nfts/blob/iris-only-gas-consumption).

To execute the tests, following commands need to be executed in the folder `development/smart-contracts`:

```sh
docker-compose up -d
npm i
npm run test:iris
```

Note:
- Node needs to be installed
- For comparison it is sufficient to execute the tests with a smaller set of AENS names. Those can be outcommented in the array starting in line 25 of the iris test file.

hanssv:
Fixes #4197. Reading from the contract store is expensive, we introduced a read cache to avoid the worst cases - still the first time reading an element is expensive (accessing the MPT). We set an approximate cost of 20000 gas; here we adjust to 5000 gas - it should be enough to deflect misuse, but not too big making normal usage overly expensive.

### DIFF
--- a/apps/aefate/src/aefa_engine_state.erl
+++ b/apps/aefate/src/aefa_engine_state.erl
@@ -538,7 +538,7 @@ spend_gas_for_new_cells1(NewCells, #es{ created_cells = Cells } = ES) ->
 spend_gas_for_store_values(StoreValues, ES) ->
     case consensus_version(ES) < ?CERES_PROTOCOL_VSN of
         true  -> ES;
-        false -> spend_gas(StoreValues * 20000, ES)
+        false -> spend_gas(StoreValues * 5000, ES)
     end.
 
 -define(GAS_DENOMINATOR, 1000).


dincho (approved):

ThomasArts (approved):
I like the hat this is drawn from :-)

PR #4253, opened 2024-01-22
Make BRI configurable by protocol version

mitchelli:


apps/aecore/src/aec_dev_reward.erl
```diff
@@ -20,9 +21,12 @@
...
 %%% for: "ak_2KAcA2Pp1nrR8Wkt3FtCkReGzAi8vJ9Snxa4PcmrthVx8AhPe8:109"
--define(MAINNET_BENEFICIARIES, [{<<172,241,128,85,116,104,119,143,197,105,4,192,224,207,200,138,230,84,111,38,89,33,239,21,201,183,185,209,19,60,109,136>>, 109}]).
+%%%%%%%%% TODO update with real values
+-define(MAINNET_BENEFICIARIES, [{<<172,241,128,85,116,104,119,143,197,105,4,192,224,207,200,138,230,84,111,38,89,33,239,21,201,183,185,209,19,60,109,136>>, 109, undefined, ?IRIS_PROTOCOL_VSN},
+                                {<<172,241,128,85,116,104,119,143,197,105,4,192,224,207,200,138,230,84,111,38,89,33,239,21,201,183,185,209,19,60,109,136>>, 109, ?CERES_PROTOCOL_VSN, undefined}]).
 %%% for: "ak_2A3PZPfMC2X7ZVy4qGXz2xh2Lbh79Q4UvZ5fdH7QVFocEgcKzU:109"
--define(TESTNET_BENEFICIARIES, [{<<152,57,168,5,218,153,177,254,226,207,243,133,11,50,143,68,121,242,94,41,187,198,158,67,133,88,6,71,55,26,85,54>>, 109}]).
+-define(TESTNET_BENEFICIARIES, [{<<152,57,168,5,218,153,177,254,226,207,243,133,11,50,143,68,121,242,94,41,187,198,158,67,133,88,6,71,55,26,85,54>>, 109, undefined, ?IRIS_PROTOCOL_VSN},
```
hanssv, line 28: Do we want this "change" (from `<<152, 57, ...>>` to `<<152, 57, ...>>`) just for symmetry, or should we just leave the testnet BRI unchanged?

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -82,18 +102,47 @@ activated(Protocol) ->
     end.
 
 
-unparse_beneficiary({PK, Share}) ->
-    <<(aeser_api_encoder:encode(account_pubkey, PK))/binary, ":", (integer_to_binary(Share))/binary>>.
+unparse_beneficiary({PK, Share, FromProtocol, ToProtocol}) ->
```
hanssv, line 105: unparse... How about `encode`? 😅 But I know it is legacy.

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -1797,8 +1797,8 @@
...
                     "type" : "array",
                     "items" : {
-                        "description" : "Public keys belonging to protocol maintainers with reward shares (100 is 10%). If not set testnet or mainnet beneficiaries and shares will be used based on network_id configuration value. IMPORTANT: The value of this setting is under governance, thus it should not be changed without previous agreement within the configured network on doing so.",
-                        "pattern": "^ak_[1-9A-HJ-NP-Za-km-z]*:[0-9]+$"
+                        "description" : "Public keys belonging to protocol maintainers with reward shares (100 is 10%), optionally a protocol version can be specified from and/or to this account will be rewarded, e.g. ak_2KAcA2Pp1nrR8Wkt3FtCkReGzAi8vJ9Snxa4PcmrthVx8AhPe8:109:6: specifies from Ceres onwards, ak_2KAcA2Pp1nrR8Wkt3FtCkReGzAi8vJ9Snxa4PcmrthVx8AhPe8:109::5 specifies to Iris. If not set testnet or mainnet beneficiaries and shares will be used based on network_id configuration value. IMPORTANT: The value of this setting is under governance, thus it should not be changed without previous agreement within the configured network on doing so.",
```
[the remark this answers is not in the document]
mitchelli: The protocol_beneficiaries is a list of beneficiaries rather than a single beneficiary. If it was a single beneficiary then it would be implicit but with a list I don't think it is.

apps/aecore/src/aec_dev_reward.erl
mitchelli: Yes, I prefer that, it is clearer.

apps/aeutils/priv/aeternity_config_schema.json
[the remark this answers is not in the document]
mitchelli: Yes, Addr:BRIAmount:: is the common case but if you don't need specify the to and from in this case, they are optional. You could just specify Addr:BRIAmount

From the code, it looks like it is possible to specify multi BRI accounts, it adds them together and makes sure they don't go above a certain limit.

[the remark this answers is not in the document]
dincho: okay, if there is technical requirement that's fine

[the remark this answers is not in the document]
hanssv: Calling it a requirement is perhaps a bit strong, but I think this variant is slightly more fool-proof. Though I'm sure Nikita or Thomas would break it in a minute 😅

hanssv (approved):
Looks good to me! 

I've run (locally) my hacky system test - where I configured a new BRI account at Ceres. And accounts are awarded as expected 👍

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -20,34 +21,47 @@
 -define(TOTAL_SHARES, 1000). % 100 shares == 10% of the reward
 %%% weighted avg BRI voting result of 1% to 20% (yes) votes is 10.89869526640124746202%, 109 of 1000 shares will be the protocol reward
 %%% for: "ak_2KAcA2Pp1nrR8Wkt3FtCkReGzAi8vJ9Snxa4PcmrthVx8AhPe8:109"
--define(MAINNET_BENEFICIARIES, [{<<172,241,128,85,116,104,119,143,197,105,4,192,224,207,200,138,230,84,111,38,89,33,239,21,201,183,185,209,19,60,109,136>>, 109}]).
+%%%%%%%%% TODO update with real values
```
ThomasArts, line 24: Is this a TO DO for later or something that was planned for this PR?

mitchelli: This is for later, it should be before updated before Ceres goes live but we are waiting for a new BRI account from the foundation.

apps/aecore/src/aec_dev_reward.erl
```diff
@@ -103,27 +151,64 @@ parse_beneficiary(BeneficiaryShareStr) ->
...
+    try lists:filtermap(fun parse_beneficiary/1, BeneficiarySharesProtocolsStrs) of
+        BeneficiarySharesProtocols ->
+            {ok, [{ProtocolVsn, beneficiaries_at_protocol(ProtocolVsn, BeneficiarySharesProtocols)} ||
+                    ProtocolVsn <- protocols()]}
     catch
         error:Reason ->
             {error, Reason}
     end.
 
+beneficiaries_at_protocol(ProtocolVsn, Beneficiaries) ->
```
ThomasArts, line 164: feels overly careful to check for FromProtocol and TOProtocol to be undefined.
The places where you call this function already guarantee that these are integers.

I would add a type spec that they must be integers and possibly crash if not by a guard is_integer in the function clause

mitchelli: The problem is that it is also called from the test function: set_beneficiaries which could include undefined as to or from parameters.

apps/aeutils/priv/aeternity_config_schema.json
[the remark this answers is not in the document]
ThomasArts: This is how a fool would try to change the beneficiary
1. Add NewAddr:1000 to config 
That fails, because the old beneficiary is there and now the total is more than 100% ???
2.  remove Addr:1000 from config.... 
That fails, I guess, because synchronizing and validating old blocks would no longer work?
Possibly not something one would notice straight away, but as soon as one starts that config on a new node to sync??
3. Introduce a new protocol vsn 2 and then configure
Addr:1000::1
NewAddr:1000::
Because someone said that that was the default use-case.
But it is actually better to say
NewAddr:1000:2:

Did I understand the documentation well?

ThomasArts (approved):

apps/aecore/src/aec_dev_reward.erl
ThomasArts: Yes, but that is test code... you don't need to expose this function from outside (only for testing) and therefore you can adapt the test to obtain clearer code and only export for test

mitchelli: The set_beneficiaries is only exported if TEST is set. It is used externally by the chain tests.

PR #4266, opened 2024-02-05
Deprecate UAT seed node (eu-central-1)

dincho:
The node itself will be running for about 2-3 releases until we see there is no old UAT releases still in the wild.


hanssv (commented):
Not sure if it makes sense to patch _old_ release notes? 🤔

dincho:
> Not sure if it makes sense to patch _old_ release notes? 🤔

yeah, I didn't patched it at first but then decided to remove any traces, all the old release notes are already in the packages, but it also renders on docs.aeternity.io

hanssv (approved):

ThomasArts (approved):
Little unnecessary to remove this node from the release notes, because at the time of release this was what one should do.
Now it looks like retro-fitting the releases ,but fine.

PR #4268, opened 2024-02-06
Fix mainline packaging

### LINKED ISSUE #4267: Broken distros build/package jobs
- ubuntu deb
- nix

dincho:
- fix NixOS builds
- remove .deb and brew packages

Since we switched to OTP26 (and other versions) we can't use anymore Erlang ESL repository (lack of versions and architectures). The most recent Ubuntu LTS (22.04) includes erlang `24.2.1`, while our minimum required version is `24.3.4.15`. Also the current builder images doesn't support 22.04 because of build issue that has to be fixed first (https://github.com/aeternity/docker-builder/issues/52)

Once the builder supports 22.04 we can try lowering the erlang requirement to 24.2.1 and build a .deb, however that would mean we're stuck with that minimum version until next Ubuntu LTS.

An alternative approach would be trying to package the erlang runtime in the .deb package, although   that sounds very wrong and abusive (isn't it? 🤔 )

fixes #4267

### DIFF
--- a/.circleci/config/workflows/commit.yml
+++ b/.circleci/config/workflows/commit.yml
@@ -228,15 +228,6 @@ jobs:
             - /releases\/.*/
             - << pipeline.parameters.master_branch >>
 
-  - ubuntu_package:
-      requires: []
-      context: ae-slack
-      filters:
-        branches:
-          only:
-            - /releases\/.*/
-            - << pipeline.parameters.master_branch >>
-
   - upload-artifacts-s3:
       name: upload-artifacts-s3
       bucket: << pipeline.parameters.s3_builds_bucket >>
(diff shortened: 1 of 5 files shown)


hanssv:
What is the reason for requiring 24.3.4.15? Any particular bugfix we are depending on there? Otherwise if we say we support OTP-24 I think you normally would require 24.1 🤔

hanssv:
It is a bit like the difference between saying 22.04 and 22.04.3 for Ubuntu... Most people would use 22.04.3 (hopefully) but it would still work with 22.04, right?

dincho:
> It is a bit like the difference between saying 22.04 and 22.04.3 for Ubuntu... Most people would use 22.04.3 (hopefully) but it would still work with 22.04, right?

Yes, probably. That's why I suggested lowering our minimal OTP to 24.2.1 to allow support of 22.04 packing.
That would mean, we should also build and test (in our pipelines) for that version. If you suggest we can go lower to 24.1

The main question would be are we OK to keep supporting OTP 24 until April 2032 (EOL of Jammy) :)

dincho:
Regardless of whatever decision we take, we have to drop the CI pipelines for now to unlock the master

hanssv:
No, I don't think we want to support OTP-24 until 2032... 

At that point we should support OTP-33 to OTP-35 😅

hanssv (approved):

mitchelli (approved):

PR #4270, opened 2024-02-08
Make the BRI optional per protocol version

mitchelli:
The BRI account should be optional per protocol version and optional in general.


hanssv (approved):

dincho (approved):

dincho:
Please set correct PR label if you want good categorization in the release notes

PR #4275, opened 2024-02-20
Prepare for OTP-27

hanssv:
Apart from 'else' and 'maybe' it was mainly deps with warning/errors/deprecations that needed update.

Also included a new rebar3.


hanssv:
Made this a drafs awaiting final versions of some repo updates and full sync tests on testnet/mainnet

ThomasArts (dismissed):

mitchelli (dismissed):

hanssv:
Rebased - because of conflict...

ThomasArts (approved):

dincho (changes requested):
You can't simply remove rebar3 :)

That would break:
- Makefile
- Dockerfile
- CI 

I'll try to find more specifics and eventual push fixes for that

dincho (approved):
Ah, I got confused by the UI, it's not removed but updated.

hanssv:
Yes the old one was broken for both OTP-26 and OTP-27...

PR #4277, opened 2024-02-26
Add endpoint to get recent gas prices

hanssv:
Replaces #4191 - with a branch in this repo for easier collaboration.

There are a few things to consider here. 
* A minimum gas price is only relevant when chain is congested
* When a micro-block is full isn't entirely obvious
* Min gas fee can also be somewhat miner specific (miner min gas fee configuration).

This PR is supported by Æternity Foundation and Æternity Crypto Foundation


apps/aehttp/src/aehttp_logic.erl
```diff
@@ -160,3 +161,58 @@ get_top_blocks_time_summary(Count) ->
...
+
+get_top_blocks_gas_price_summary() ->
+    TopKeyHash = aec_chain:top_key_block_hash(),
+    case get_gens_min_gas_price(TopKeyHash, 480) of
+        Err = {error, _} ->
+            Err;
+        MinGasPrices ->
+            MinPriceN = fun(N) ->
+                            case lists:min(lists:sublist(MinGasPrices, N)) of
+                                undefined -> 0;
```
ThomasArts, line 173: I OTP25 lists:min([]) throws an exception, but never undefined?

ThomasArts: Moreover, minimum should then possibly be the consensus defined minimum?

ThomasArts (dismissed):
Not sure you should use 0 as minimum instead of consensus minimum, but that would not be a problem in practise

apps/aehttp/test/aehttp_integration_SUITE.erl
```diff
@@ -2005,6 +2009,25 @@ get_status_sut(IntAsString) ->
...
+get_recent_gas_prices(_Config) ->
+    Host = external_address(),
+    {ok, 200, [
+        #{ <<"blocks">> := 1, <<"min_gas_price">> := MinGasPrice1 },
+        #{ <<"blocks">> := 5, <<"min_gas_price">> := MinGasPrice5 },
+        #{ <<"blocks">> := 20, <<"min_gas_price">> := MinGasPrice20 },
+        #{ <<"blocks">> := 120, <<"min_gas_price">> := MinGasPrice120 },
+        #{ <<"blocks">> := 480, <<"min_gas_price">> := MinGasPrice480 }
+    ]} = http_request(Host, get, "recent-gas-prices", []),
+    MinGasPrice = 1000000000,
```
dincho, line 2023: Not very convincing test :)

dincho (approved):
Same concern as @ThomasArts about protocol minimums

dincho:
> but that would not be a problem in practise

Why not, if the client solely relies on this result and the last X blocks are empty? 🤔

hanssv:
> > but that would not be a problem in practise
> 
> Why not, if the client solely relies on this result and the last X blocks are empty? 🤔

I think 0 would signal quite well that we don't have any data - the alternative would be protocol minimum, but there is no miner accepting that so that would be equally broken.

apps/aehttp/test/aehttp_integration_SUITE.erl
hanssv: Hmm, no... Didn't look at it, just updated it to pass 😅

dincho (approved):

dincho (dismissed):

ThomasArts (dismissed):

apps/aehttp/src/aehttp_logic.erl
davidyuk: Returning 0 works for me as a way to distinguish
- is there empty space in blocks (then SDK should use minimum gas price)
- all blocks full, but transactions using minimum gas price (then SDK should increase the gas price a bit to ensure tx would be mined)
Otherwise I would ask for a flag indicating that.

davidyuk (commented):
@hanssv In the latest version, I can't distinguish if blocks are full and I should use min gas price + 1 or if there is space in blocks and I can use 1e9 (default min gas price). If sdk would always do min gas price + 1 then multiple sdk instances can increase the gas price without reason.

hanssv:
> Another issue:
> 
>     * I've build this branch, run it with configuration for sdk tests
> 
>     * `recent-gas-prices` shows "min_gas_price": 0 everywhere
> 
>     * changed min gas price in sdk to 1e9 + 1
> 
>     * executed some sdk tests
> 
>     * `recent-gas-prices` shows "min_gas_price": 1000000001 everywhere
> 
>     * I'd waited for a minute
> 
>     * changed min gas price in sdk to 1e9 + 2
> 
>     * executed some sdk tests
> 
>     * `recent-gas-prices` shows "min_gas_price": 1000000001 everywhere, but I expected to get 1000000002 in 1-minute because all transactions made have gas price 1e9 + 2

Microseconds and milliseconds were mixed-up, so you'd have to wait for 1000 minutes to see the effect 🙈 

Fixed now.

ThomasArts (dismissed):
Missed that in review before, but indeed, times 1000

hanssv:
So we decided on sticking to this approach (reporting minimum gas price that made it on chain in the last X, Y and Z minutes) - but we'll also add a congestions metric (basically how many percent of of the available gas that was used in the the given time).

I will also sort out the caching properly - probably by an actual caching lib 😅

apps/aehttp/src/aehttp_logic.erl
```diff
@@ -160,3 +163,75 @@ get_top_blocks_time_summary(Count) ->
...
+    {min(MinGasPrice, AccMinGasPrice), AccUsedGas + UsedGas, AccTotGas + TotGas}.
+
+get_mb_stats(Hash, Protocol) ->
+    case persistent_term:get({aehttp_app, kache}, undefined) of
+        undefined ->
+            get_mb_stats_(Hash, Protocol);
+        Cache ->
+            case kache:get(Cache, Hash) of
+                {ok, _} ->
+                    lager:info("ZZZ: ~p in cache", [Hash]);
```
ThomasArts, line 210: possibly real log message or debug message that the cache was hit?

apps/aehttp/src/aehttp_logic.erl
```diff
@@ -160,3 +163,75 @@ get_top_blocks_time_summary(Count) ->
...
+                notfound ->
+                    lager:info("ZZZ: ~p NOT in cache", [Hash])
```
ThomasArts, line 212: same here

ThomasArts (approved):

dincho (approved):

mitchelli (approved):

PR #4285, opened 2024-03-05
Extra key for HTTP error responses

### LINKED ISSUE #4284: Provide extra info along "Invalid tx" error
Related issue https://github.com/aeternity/aeternity/issues/4283.

Currently, if for some reason I posted (to `/transactions`) an invalid transaction I'm getting `{"reason":"Invalid tx"}`. Various mistakes end up with this message, like:
- invalid signature,
- low nonce,
- low ttl,
- low fee.

Would be good if the node provide some key to indicate exactly the first check failed, like in `debug/check-tx/pool`. This would allow tools (like js-sdk) to provide developers with more friendly error messages, to simplify debugging in some cases.

As far as these checks are already done by the node while accepting transaction, it shouldn't require extra computations to provide it in response.

ThomasArts:
Fixes #4284 

### DIFF
--- a/apps/aehttp/priv/oas3.yaml
+++ b/apps/aehttp/priv/oas3.yaml
@@ -3378,6 +3378,8 @@ components:
       properties:
         reason:
           type: string
+        error_code:
+          type: string
       required:
         - reason
     ContractObject:
--- a/apps/aehttp/priv/swagger.yaml
+++ b/apps/aehttp/priv/swagger.yaml
@@ -3281,6 +3281,8 @@ definitions:
     properties:
       reason:
         type: string
+      error_code:
+        type: string
     required:
       - reason
   ContractObject:
--- a/apps/aehttp/src/aehttp_dispatch_ext.erl
+++ b/apps/aehttp/src/aehttp_dispatch_ext.erl
@@ -212,10 +212,12 @@ handle_request_('GetPendingKeyBlock', _Req, _Context) ->
                 error ->
                     {404, [], #{reason => <<"Block not found">>}}
             end;
-        {error, beneficiary_not_configured} ->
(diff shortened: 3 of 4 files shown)


hanssv (approved):

apps/aehttp/priv/swagger.yaml
```diff
@@ -3281,6 +3281,8 @@ definitions:
     properties:
       reason:
         type: string
+      error_code:
```
davidyuk, line 3284: can it be just `code`? As far as it is already in context of `Error`
seems I have troubles with snake keys in node's exceptions

davidyuk (approved):
It is fine, the only proposal is to rename `error_code` to `code`

apps/aehttp/priv/oas3.yaml
```diff
@@ -3378,6 +3378,8 @@ components:
       properties:
         reason:
           type: string
+        error_code:
+          type: string
       required:
         - reason
```
davidyuk, line 3384: `error_code` was added in other places, but there are still cases when it is missed? so it can't be required?

hanssv: It was only added where there is anything interesting to report I think.

apps/aehttp/priv/swagger.yaml
ThomasArts: This can be confusing with contract code, so rather not.

apps/aehttp/priv/oas3.yaml
ThomasArts: And it is not required. You can get responses without any `error_code`. It's just a bit of extra information the SDK can look at when deciding what to do.

dincho (approved):

PR #4294, opened 2024-03-10
Use `oneOf` instead of `anyOf` in oas3.yaml

davidyuk:

> [anyOf](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#anyof) – validates the value against any (one or more) of the subschemas

https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/

As I understand, it is better to use `oneOf` instead of `anyOf` for both transaction parameters and key/micro blocks because there are no use case when the value would comply with multiple schemas:
- in the case of transactions each item has its own `type` value;
- keyblock and microblock have their own unique fields, so the node can't return both at once.

### DIFF
--- a/apps/aehttp/priv/oas3.yaml
+++ b/apps/aehttp/priv/oas3.yaml
@@ -3287,7 +3287,7 @@ components:
           required:
             - version
             - type
-        - anyOf:
+        - oneOf:
             - $ref: "#/components/schemas/SpendTx"
             - $ref: "#/components/schemas/ChannelCreateTx"
             - $ref: "#/components/schemas/ChannelDepositTx"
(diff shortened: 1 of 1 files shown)


ThomasArts:
The OneOf used in the yaml is correct.
A transaction is one of the three listed transactions and the same string should not validate against more than one... this would mean that a transaction is ambiguous, which it isn't.

So I would like to reject this PR.

ThomasArts (approved):
Indeed, only one should match

This may have consequences for test case generation. I can have a look at that if indeed failures are due to that.

mitchelli (approved):

ThomasArts:
Two test cases fail due to this change.
and indeed, manual inspection shows that `ChannelSettleTx` and `ChannelCloseMutualTx` seem to be indistiguishable on the purely syntactic level that JSON offers.

ThomasArts (changes requested):
Does not hold for transactions...
There seem to be transactions that on syntactic level are the same.

PR #4301, opened 2024-03-12
Deprecate mainnet us-east-2 nodes

dincho:
This is part of a general shrinking strategy


ThomasArts (approved):

hanssv (approved):

hanssv:
While doing the OTP-27 sync tests I did notice there were fewer nodes that could serve blocks - sync was partially network bound...

dincho:
Okay, good point. Let's delay this until we have better sync parallelism that is not limited by the number of connected nodes.

PR #4302, opened 2024-03-13
Update rocksdb

ThomasArts:
Replaces #4139


ThomasArts:
Tested have old code sync from start upto height 310618 and then continue sync with new code on that same database.
Sync continues without problems.

hanssv (approved):

dincho (approved):

PR #4303, opened 2024-03-14
Add missed timer initialization in aedevmode_emitter:init

davidyuk:

The problem I had: in tests of aepp-sdk I needed more accurate intervals between keyblocks, and I decided to use devmode instead of pow miner. The config I used:
```yaml
dev_mode:
    keyblock_interval: 5
    microblock_interval: 1
    auto_emit_microblocks: false
```
this way, node was generating keyblocks, but no microblocks, also I wasn't able to mine transactions.

As I understand, keyblock timer was working because something generating 1st block and this triggers the event that restarts a timer
https://github.com/aeternity/aeternity/blob/b0a7140abdb6b6df79ebbb2a8b75e6397cbb6eae/apps/aedevmode/src/aedevmode_emitter.erl#L144-L145
but nothing was creating a microblock, so that timer wasn't working.

### DIFF
--- a/apps/aedevmode/src/aedevmode_emitter.erl
+++ b/apps/aedevmode/src/aedevmode_emitter.erl
@@ -77,9 +77,12 @@ init([]) ->
         _ ->
             ok
     end,
-    {ok, #st{ keyblock_interval   = cfg(<<"keyblock_interval">>)
+    St = #st{ keyblock_interval   = cfg(<<"keyblock_interval">>)
             , microblock_interval = cfg(<<"microblock_interval">>)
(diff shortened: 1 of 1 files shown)


ThomasArts:
@davidyuk Could you explain your use-case in more detail. 
It might be that this "fix" solves your particular problem, but it also changes the behaviour in a way that might ot have been intended.

ThomasArts:
Please create an issue to describe the exact limitation you bumped into.

The solution as chosen here and is duplicated in PR #4325 seems not the right one. If there is an issue that explains the use of this emitter state machine in general and what is impossible to achieve in that setup, then we can try to find the right solution.

PR #4308, opened 2024-03-18
Add ability to dry-run endpoint to run transactions from the mempool 

mitchelli:
This is to solve the following issue: https://github.com/aeternity/aeternity/issues/4283

### DIFF
--- a/apps/aehttp/priv/swagger.yaml
+++ b/apps/aehttp/priv/swagger.yaml
@@ -3376,6 +3376,8 @@ definitions:
     properties:
       tx:
         $ref: '#/definitions/EncodedByteArray'
+        tx_hash:
+          $ref: "#/components/schemas/EncodedHash"        
       call_req:
         $ref: '#/definitions/DryRunCallReq'
   DryRunCallReq:
--- a/apps/aehttp/src/aehttp_helpers.erl
+++ b/apps/aehttp/src/aehttp_helpers.erl
@@ -736,7 +736,7 @@ dry_run_tx_hash(TxHash) ->
                 {mempool, SignedTx} ->
                     Tx = aetx_sign:tx(SignedTx),
                     {Type, _} = aetx:specialize_type(Tx),
-                    case not lists:member(Type, [paying_for_tx, offchain_tx, ga_meta_tx]) of
+                    case not lists:member(Type, [offchain_tx, ga_meta_tx]) of
                         true  -> {ok, Tx};
                         false -> {error, lists:concat(["Unsupported transaction type ", Type])}
                     end;
@@ -822,6 +822,7 @@ ok_err({error, _}) -> <<"error">>;
 ok_err(_)          -> <<"ok">>.
 
 type(spend_tx)                  -> <<"spend">>;
+type(paying_for_tx)             -> <<"paying_for_tx">>;
 type(oracle_register_tx)        -> <<"oracle_register">>;
 type(oracle_extend_tx)          -> <<"oracle_extend">>;
 type(oracle_query_tx)           -> <<"oracle_query">>;
--- a/apps/aehttp/src/aehttp_helpers.erl
+++ b/apps/aehttp/src/aehttp_helpers.erl
@@ -741,7 +741,6 @@ dry_run_tx_hash(TxHash) ->
                         false -> {error, lists:concat(["Unsupported transaction type ", Type])}
                     end;
                 _ ->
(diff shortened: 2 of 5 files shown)


apps/aehttp/priv/swagger.yaml
```diff
@@ -3376,6 +3376,8 @@ definitions:
     properties:
       tx:
         $ref: '#/definitions/EncodedByteArray'
+        tx_hash:
```
ThomasArts, line 3379: wrong indentation
```suggestion
      tx_hash:
```

apps/aehttp/priv/swagger.yaml
```diff
@@ -3376,6 +3376,8 @@ definitions:
...
+          $ref: "#/components/schemas/EncodedHash"
```
ThomasArts, line 3380:
```suggestion
        $ref: "#/components/schemas/EncodedHash"
```

apps/aehttp/src/aehttp_helpers.erl
```diff
@@ -720,6 +728,25 @@ dry_run_txs_([#{ <<"tx">> := ETx } | Txs], Acc) ->
 dry_run_txs_([#{ <<"call_req">> := CallReq } | Txs], Acc) ->
     dry_run_txs_(Txs, [{call_req, CallReq} | Acc]).
 
+
+dry_run_tx_hash(TxHash) ->
```
ThomasArts, line 732: why do we need this? Can one not use the present dry_run machinery after having found the Tx via the hash (and removed signature)?

hanssv: It should be clear from the linked issue and its discussion? This is just a convenience feature - but since there is an actual use-case from SDK side, it simplifies things to provide it.

apps/aehttp/src/aehttp_helpers.erl
```diff
@@ -720,6 +728,25 @@ dry_run_txs_([#{ <<"tx">> := ETx } | Txs], Acc) ->
...
+    case aeser_api_encoder:safe_decode(tx_hash, TxHash) of
+        {ok, TxHashInternal} ->
+            case aec_chain:find_tx_with_location(TxHashInternal) of
+                {mempool, SignedTx} ->
+                    Tx = aetx_sign:tx(SignedTx),
+                    {Type, _} = aetx:specialize_type(Tx),
+                    case not lists:member(Type, [paying_for_tx, offchain_tx, ga_meta_tx]) of
```
hanssv, line 739: At this stage dry-running `paying_for_tx` and `ga_meta_tx` isn't really an issue - but I suspect it might be stopped further down the chain 🤔

mitchelli: This is the same check as for a transaction. I should be put them in a function to make sure they are consistent for mempool transactions and contributed transactions.

hanssv:
It needs a rebase I think 😅

apps/aehttp/src/aehttp_helpers.erl
ThomasArts: What I meant here was... there is kind of code duplication here, can we do without that duplication?

hanssv (approved):

apps/aehttp/src/aehttp_helpers.erl
```diff
@@ -720,6 +728,25 @@ dry_run_txs_([#{ <<"tx">> := ETx } | Txs], Acc) ->
...
 
+
+dry_run_tx_hash(TxHash) ->
+    case aeser_api_encoder:safe_decode(tx_hash, TxHash) of
+        {ok, TxHashInternal} ->
+            case aec_chain:find_tx_with_location(TxHashInternal) of
+                {mempool, SignedTx} ->
+                    Tx = aetx_sign:tx(SignedTx),
+                    {Type, _} = aetx:specialize_type(Tx),
+                    case not lists:member(Type, [offchain_tx]) of
```
ThomasArts, line 739: Here you do not dry run `offchain_tx`, but when the tx is not from a hash you also exclude `paying_for_tx`, and `ga_meta_tx`.
There is a comment in the PR that you would combine the logic in one function doing the same?

mitchelli: I was going to change it to a single function but executing dry runs on paying_for_tx and ga_meta_tx from mempool transactions doesn't have the same risks as running the transaction directly, so we decided to allow it.

ThomasArts (approved):

PR #4322, opened 2024-04-05
Load contracts from default directories

mitchelli:
Fixes https://github.com/aeternity/aeternity/issues/4320

### DIFF
--- a/apps/aecore/src/aec_fork_block_settings.erl
+++ b/apps/aecore/src/aec_fork_block_settings.erl
@@ -327,36 +327,6 @@ contracts_file_name(Release) ->
 
 
 -ifdef(TEST).
-accounts_json_file() ->
-    NetworkId = aec_governance:get_network_id(),
-    ConsensusModule = aec_consensus:get_consensus_module_at_height(0),
-    case NetworkId of
-        <<"ae_mainnet">>                  -> "accounts.json";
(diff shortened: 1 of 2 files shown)


apps/aecore/src/aec_fork_block_settings.erl
```diff
@@ -344,7 +348,15 @@ extra_accounts_json_file() ->
     "extra_accounts_test.json".
 
 contracts_json_file() ->
```
hanssv, line 350: Isn't this just a subset of the function below? I can't remember why we wan't the TEST one?

hanssv (approved):

apps/aecore/src/aec_fork_block_settings.erl
dincho: This filename is a legacy one ... back then one could only set accounts (and later contracts) via this file. The suffix "_test" I believe comes from the original usage for tests only, but at some point all "custom" network ID's has been using this.

dincho: Ah, I see what your comment is about now .. nvm

mitchelli: I've moved the functions out of test and made a generic function for creating the file name.

hanssv (approved):

ThomasArts (approved):

PR #4324, opened 2024-04-10
Fix typo in "block not found" message

davidyuk:
This can be a breaking change 🙃 


hanssv (approved):
Let's assume no-one is stupid enough to match explicitly on a misspelt error message 🙈

dincho (approved):

PR #4325, opened 2024-04-10
Add missed timer initialization in aedevmode_emitter:init

ThomasArts:
replaces PR #4303 to get permission to run CI

### DIFF
--- a/apps/aedevmode/src/aedevmode_emitter.erl
+++ b/apps/aedevmode/src/aedevmode_emitter.erl
@@ -77,9 +77,12 @@ init([]) ->
         _ ->
             ok
     end,
-    {ok, #st{ keyblock_interval   = cfg(<<"keyblock_interval">>)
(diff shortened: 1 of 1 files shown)


uwiger:
Is this necessarily what one wants, though?

My assumption has been that one would often want to manually start/stop block production.
Perhaps an option to control auto-start?

ThomasArts:
> Is this necessarily what one wants, though?
> 
> My assumption has been that one would often want to manually start/stop block production. Perhaps an option to control auto-start?

Indeed. I have questions here:
1. to @uwiger where is the specification on how this should work? If you have auto-emit `true` and a microblock interval set, how then do you guarantee that you keep that interval? 
2. If you use auto-emit `false` then @davidyuk why do you expect micro blocks to be produced at all unless you call `mine_until_txs_on_chain`?

ThomasArts:
This is not the right solution to the problem, we need to revisit the emitter to support additional use cases and document better what it does today.

PR #4344, opened 2024-04-29
Logdir fixes

### LINKED ISSUE #3860: bin/aeternity foreground should not write to log files
This mode should output all logs to stdout/stderr and not write logs to any file in the log directory.

richcarl:
Ensures that log files and directories are not created when running in foreground mode.
Closes #3860


dincho (commented):
Couple of issues as we already discussed:
- foreground doesn't handle CRTL+C which is a must in interactive usage and for orchestrators
- the default log level should be INFO instead of DEBUG as it was for console output

richcarl:
Turns out that Erlang actually handles SIGTERM these days as equivalent to running init:stop(), so plain `docker stop` will do a clean shutdown. Good to know.

@dincho Do you have some reference you can point me to that states clearly what is expected with regard to SIGINT in Docker?

dincho:
I don't think that this is docker specific, this general UNIX/POSIX process expectations. When you hit CRTL+C you expect the process to stop "immediately" or you (at least me) never run that crap again.

dincho:
> Turns out that Erlang actually handles SIGTERM these days as equivalent to running init:stop(), so plain `docker stop` will do a clean shutdown. Good to know.

I guess the process doesn't stop on time then and get's killed: https://docs.docker.com/reference/cli/docker/container/stop/#description

richcarl:
> I don't think that this is docker specific, this is general UNIX/POSIX process expectations. When you hit CRTL+C you expect the process to stop "immediately" or you (at least me) never run that crap again.

To clarify things: Running `bin/aeternity foreground` (as in my change) passes `-noinput +Bd` to `erl`, and that makes it behave just as you want. (Without `+Bd`, the default behaviour of `erl` is that it will install a break handler, which brings up the Erlang BREAK menu on Ctrl-C - not what you want.) Perhaps you are thinking about `+Bi`, which would make it ignore Ctrl-C, but that is not what I'm doing here. So running as `foreground` should be correct.

In any case, Ctrl-C sends SIGINT, but tools like Docker don't send SIGINT - only the terminal uses that signal. Docker sends SIGTERM first to tell the program to stop, and then SIGKILL if nothing happens in about ten seconds.

dincho (changes requested):
I'm not going to continue arguing on the signal handling in this PR as it is logs related.

Please take out the proposed change in separate PR if you think that would fix a problem, however as we discussed and I wrote in this PR the proposed change actually BREAKS the expected behaviour, regardless of the reason.

richcarl:
This can't be fixed right now because of how the test suites work.

Issue #4317, opened 2024-03-26
Don't encode binary pointer key as string in name details


davidyuk:
Together with mdw team we find out that an aens pointer key can be arbitrary bytes. But in name details it is encoded as string that doesn't works well if we are have raw bytes in key.

Example name update transaction: tx_+M4LAfhCuEDC0qgighRqQfsb7+rBdj3YgPQrn6eJlFg/JeSUGaHT0Ib1llr/D/EChZmpwH+APTgGSMlKVnABGZNkOKCNo8gCuIb4hCIBoQFx4GF37I8Up2XTBXfJoSVMFRTdDlJMSpnDZpC7Qr0M4QahAgfcQ3oAn1mHbkr0vac5wk5YqlaW8mSDRFGoGMJaW6Z1gwK/IO3sid6tvu9teUtleaEBK4KUmZC+wTFE6qE33lHiP5x0TC8fqd/kWaeSpvwKVYaCDhCGEBRyHqAAHpZnprU=

Example of name details output:
```
$ curl http://localhost:3013/v3/names/pdy67dxykdsrETvH6YIcGkvCNPHFTY.chain | jq .
{
  "id": "nm_4TnSeXE5LJ25t4qsqjBkSrY2tTpHVQGdiVU9NPtK2NeFU8yj6",
  "owner": "ak_s9pgJcAAiWBfpuV6pF9QnxEZBVfe9537t3bTxphrDFaV3LVBq",
  "pointers": [
    {
      "id": "ak_LAQanhrS2ZzL5nGyh5UZuZb5RvWNuggmQNM7vWq2kCPxDNdvi",
      "key": "ޭ��myKey"
    }
  ],
  "ttl": 180027
}
```

Node details:
```
$ curl http://localhost:3013/v3/status | jq .
{
  "network_id": "ae_dev",
  "node_revision": "b0a7140abdb6b6df79ebbb2a8b75e6397cbb6eae",
  "node_version": "6.12.0+52.b0a7140ab",
  "protocols": [
    {
      "effective_at_height": 1,
      "version": 6
    },
    {
      "effective_at_height": 0,
      "version": 1
    }
  ]
}
```

davidyuk:
Consider encoding keys in the pointers object as base64 to be consistent with the mdw implementation https://github.com/aeternity/ae_mdw/issues/1665

hanssv:
As discussed elsewhere - changing this is trivial, and makes sense - question is how much of a breaking change is it?

[...]

PR #4349, opened 2024-05-08
Add (bytearray) encoded key to pointers in HTTP API

hanssv:
Implements #4317 


davidyuk:
Looks good to me. The thing is in practice we don't need it, I just want pointer keys implemented the same way in node and mdw. @sborrazas are you fine to do the same way in the next mdw api? cc: @thepiwo

ThomasArts (approved):

dincho (approved):

PR #4350, opened 2024-05-09
Update sdk in sdk_smoke_test to 13.3.2

davidyuk:

https://hub.docker.com/r/circleci/node

### DIFF
--- a/.circleci/config/jobs/@docker-tests.yml
+++ b/.circleci/config/jobs/@docker-tests.yml
@@ -30,7 +30,7 @@ docker-js-sdk-smoke-test:
         tag: ${CIRCLE_BRANCH/\//_}
     - run:
         name: Start NodeJS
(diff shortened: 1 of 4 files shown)


davidyuk:
Is there a way to make node produce some blocks? JS tests stuck on waiting the first block 🙁

hanssv:
@davidyuk I don't know the setup here? How is the node configured? Is it mining? What difficulty? 

Why is the node waiting for a block, the test only seems to check if a preset account is funded or not 🤔

davidyuk:
Ok, lets check only the account balance as before

hanssv:
> Ok, lets check only the account balance as before

Ah, my bad, I only saw the waitForHeight, not the spend... I'm sure you could get the node to mine if you want a more excessive test, but let's make that in a separate PR.

hanssv (approved):

ThomasArts (approved):

PR #4354, opened 2024-05-20
Remove deprecated swagger stuff

ThomasArts:
Some tests are still testing the old Swagger API. Instead we only need to test the new OpenAPI endpoints
and some legacy endpoints.


hanssv (approved):
Looks good, great job cleaning this up!

PR #4379, opened 2024-07-12
Initial version of epochs for checking concept

mitchelli:
Test implementation of https://github.com/aeternity/aeternity/issues/4374 to check the concepts from the white paper are implemented.

### DIFF
--- a/apps/aecore/src/aec_consensus_hc.erl
+++ b/apps/aecore/src/aec_consensus_hc.erl
@@ -179,7 +179,6 @@ start_ae(StakersEncoded, PCSpendAddress) ->
             end,
             StakersEncoded),
     StakersMap = maps:from_list(Stakers),
-    lager:warning("Stakers ~p", [StakersMap]),
     %% TODO: ditch this after we move beyond OTP24
     _Mod = aec_preset_keys,
     start_dependency(aec_preset_keys, [StakersMap]),
@@ -530,13 +529,14 @@ set_new_entropy(Hash, Height, PCHeight) ->
     end.
 
 next_entropy(Hash, Height, PCHeight) ->
-    {NewState, _Seed, _Height, NewPCHeight} = case aeu_ets_cache:lookup(?ETS_CACHE_TABLE, seed) of
-                                {ok, {_State, _OldSeed, _OldHeight, PCHeight} = Result} ->
-                                    Result;
-                                _ ->
-                                    lager:info("Trying to set entropy for parent height ~p", [PCHeight]),
-                                    set_new_entropy(Hash, Height, PCHeight)
-                              end,
+    {NewState, _Seed, _Height, NewPCHeight} =
+        case aeu_ets_cache:lookup(?ETS_CACHE_TABLE, seed) of
+            {ok, {_State, _OldSeed, _OldHeight, PCHeight} = Result} ->
+                Result;
+            _ ->
+                lager:info("Trying to set entropy for parent height ~p", [PCHeight]),
+                set_new_entropy(Hash, Height, PCHeight)
+        end,
     next_entropy_from_seed(NewState, Height, NewPCHeight).
 
 next_entropy_from_seed(State, Height, PCHeight) ->
@@ -686,7 +686,6 @@ next_beneficiary() ->
                                                     [aefa_fate_code:encode_arg({string, Entropy}),
                                                      aefa_fate_code:encode_arg({bytes, NetworkId})
                                                     ]),
-            lager:warning("Entropy ~p ~p", [NextHeight, Entropy]),
             CallData = aeser_api_encoder:encode(contract_bytearray, CD),
             try call_consensus_contract_(?ELECTION_CONTRACT, TxEnv, Trees, CallData, "elect_next", 0) of
                 {ok, _Trees1, Call} ->
@@ -768,7 +767,6 @@ is_leader_valid(Node, Trees, TxEnv, PrevNode) ->
                     Leader = aec_headers:miner(Header),
                     Target = aec_headers:target(Header),
                     IsDefaultT = Target =:= default_target(),
-                    lager:warning("is_leader_valid ~p ~p", [ExpectedLeader, Leader]),
                     case ExpectedLeader =:= Leader of
                         true when IsDefaultT -> true;
                         true ->
--- a/apps/aecore/src/aec_consensus_hc.erl
+++ b/apps/aecore/src/aec_consensus_hc.erl
@@ -128,7 +128,8 @@ start(Config, #{block_production := BlockProduction}) ->
 
     start_dependency(aec_parent_connector, [ParentConnMod, FetchInterval, ParentHosts, NetworkId,
                                             SignModule, HCPCPairs, PCSpendPubkey, Fee, Amount]),
-    start_dependency(aec_parent_chain_cache, [StartHeight + pc_finality(), CacheSize, Confirmations,
+    start_dependency(aec_parent_chain_cache, [StartHeight + pc_finality(), fun(ChildHeight) -> pc_height(ChildHeight) + pc_finality() end,
+                                              CacheSize, Confirmations,
                                               BlockProduction, ProducingCommitments]),
     ok.
 
@@ -870,7 +871,7 @@ seal_padding_size() ->
 pc_height(Height) ->
     ChildEpochLength = child_epoch_length(),
     EpochNum = Height div ChildEpochLength,
-    EpochNum * parent_generation() + pc_start_height().%% child starts pinning from height 1, not genesis
+    EpochNum * parent_generation() + pc_start_height().
 
 elect_lazy_leader(Beneficiary, TxEnv, Trees) ->
     {ok, CDLazy} = aeb_fate_abi:create_calldata("elect_after_lazy_leader",
--- a/apps/aecore/src/aec_consensus_hc.erl
+++ b/apps/aecore/src/aec_consensus_hc.erl
@@ -253,11 +253,9 @@ state_pre_transform_key_node(Node, PrevNode, Trees) ->
                     aec_conductor:throw_error(parent_chain_block_not_synced);
                 {ok, Block} ->
                     Entropy = entropy(Block, Height),
-                    CommitmentsSophia = encode_commitments(Block),
                     NetworkId = aec_parent_chain_block:encode_network_id(aec_governance:get_network_id()),
                     {ok, CD} = aeb_fate_abi:create_calldata("elect",
                                                             [aefa_fate_code:encode_arg({string, Entropy}),
-                                                             CommitmentsSophia,
                                                              aefa_fate_code:encode_arg({bytes, NetworkId})
                                                             ]),
                     CallData = aeser_api_encoder:encode(contract_bytearray, CD),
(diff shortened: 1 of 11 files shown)


happi (commented):
It looks good so far.
Some tests and an example config file would help to see how things come together.

mitchelli:
> It looks good so far. Some tests and an example config file would help to see how things come together.

The latest committed version implements all the functionality but still needs some cleanup work. The relevant config looks like this:

```
chain:
  consensus:
    '0':
      type: 'hyper_chain'
      config:
        contract_owner: 'ak_11111111111111111111111111111115rHyByZ'
        election_contract: 'ct_t7ESuRFvGdYHwM6KJhKbSnFUJQhnB9aCFed1RWR89ShNDgwWA'
        rewards_contract: 'ct_LRbi65kmLtE7YMkG6mvG5TxAXTsPJDZjAtsPuaXtRyPA7gnfJ'
        expected_key_block_rate: 30000
        lazy_leader_trigger_time: 5000
        child_epoch_length: 100
        child_block_time: 1000
        parent_chain:
          confirmations: 0

          start_height: 20
          finality: 5
          parent_generation : 10

          producing_commitments: false
          consensus:
            fee: 100000000000000
            amount: 1
            network_id: 'devnet'
            spend_address: 'ak_2MGLPW2CHTDXJhqFJezqSwYSNwbZokSKkG7wSbGtVmeyjGfHtm'
            type: 'AE2AE'
          polling:
            fetch_interval: 250
            nodes:
              ['http://localhost:3013']
        stakers:
          - hyper_chain_account:
              priv: 'f191dcb1fc96fadf6c08677e7a479288977db3feab9375dd39ff85db187d851ab8440e688d6737ae51741b668c97def0bf8eff7159f77b9fd1b5b4f2c614fce4'
              pub: 'ak_2Q9pmuxKdMmgdjcbQr3BonaucbYGjdeDkqMS2mewMKYgybJ2WJ'
            parent_chain_account:
              priv: '91450EFE0516C24476390086426008147CFDEECFE6935FADA1C056C3A5BA73FBB1B577BCD327CB39E55E6C026BD6A74A1B35DE6C0650C4AE51EFAB759E415B66'
              pub: 'ak_2MGLPW2CHTDXJhqFJezqSwYSNwbZokSKkG7wSbGtVmeyjGfHtm'
          - hyper_chain_account:
              priv: '38940358aefaeb3c1c5ffc9ec9ed7dd1408c1377ea43cf02b04d72697c96e3ccb72fa51be777030b9163feb913da04e0a444ae0300d4103fa7c5cdd24c76cf23'
              pub: 'ak_2PgF7npGXkiAfJjhdkwWj4gnTi9UDwiRhP41Gru5Y9VGwe6bJS'
            parent_chain_account:
              priv: '91450EFE0516C24476390086426008147CFDEECFE6935FADA1C056C3A5BA73FBB1B577BCD327CB39E55E6C026BD6A74A1B35DE6C0650C4AE51EFAB759E415B66'
              pub: 'ak_2MGLPW2CHTDXJhqFJezqSwYSNwbZokSKkG7wSbGtVmeyjGfHtm'
          - hyper_chain_account:
              priv: 'd9f8966dc27065551c85381e72516769fdfd4c0f3496ffa350d6d8ddea9fabc86a8cb05e195a50004e495495bb9960acdae2ab466239e9b3761b0b722d968297'
              pub: 'ak_ovfcHoShxHfyKhvpwr4PWVhGmYdyhfM3DwAFztwgL886rhVS7'
            parent_chain_account:
              priv: '91450EFE0516C24476390086426008147CFDEECFE6935FADA1C056C3A5BA73FBB1B577BCD327CB39E55E6C026BD6A74A1B35DE6C0650C4AE51EFAB759E415B66'
              pub: 'ak_2MGLPW2CHTDXJhqFJezqSwYSNwbZokSKkG7wSbGtVmeyjGfHtm'
  db_direct_access: true
  hard_forks:
    '5':
      height: 0
      contracts_file: contracts/contracts.json
  persist: true
  genesis_accounts:
    ak_2B2wt3V8R92NKXRVT9ovbrT7kdSzhBBCsfVQf3PSfsCuL4JcgK: 1000000000000000000000000000000000000000000000000
    # Hyperchain account
    ak_2b8YNe9XoFDJtwaCAc6A4LLLj2sSUNqoJth66xRa87EQPGwxe7: 1000000000000000000000000000000000000000000000000
    ak_6i9dyDyiXgJZ8CKvDWbPrdzBAHzjkCix26ZasbBaUJyK7yLmu: 3100000000000000000000000000
    ak_2KAcA2Pp1nrR8Wkt3FtCkReGzAi8vJ9Snxa4PcmrthVx8AhPe8: 3100000000000000000000000000
    ak_2Q9pmuxKdMmgdjcbQr3BonaucbYGjdeDkqMS2mewMKYgybJ2WJ: 3100000000000000000000000000
    ak_2PgF7npGXkiAfJjhdkwWj4gnTi9UDwiRhP41Gru5Y9VGwe6bJS: 3100000000000000000000000000
    ak_ovfcHoShxHfyKhvpwr4PWVhGmYdyhfM3DwAFztwgL886rhVS7: 3100000000000000000000000000

fork_management:
  network_id: 'hc_devnet'
sync:
  port: 4015
  external_port: 4015
http:
  external:
    port: 4013
  internal:
    debug_endpoints: true
    port: 4113
    listen_address: 0.0.0.0
  rosetta:
    port: 4213
  rosetta_offline:
    port: 4413
  endpoints:
    dry-run: true

include_default_peers: false
mining:
  autostart: true
  beneficiary: 'ak_2b8YNe9XoFDJtwaCAc6A4LLLj2sSUNqoJth66xRa87EQPGwxe7'
peers: []
logging:
  level: debug
```

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -380,14 +384,15 @@ seal_correct_signature(Header, Signature, _Padding) ->
...
-generate_key_header_seal(_, Candidate, _PCHeight, #{expected_key_block_rate := _Expected} = _Config, _) ->
+generate_key_header_seal(_, Candidate, _PCHeight, #{child_block_time := ChildBlockTime} = _Config, _) ->
     Leader = aec_headers:beneficiary(Candidate),
     SignModule = get_sign_module(),
     case SignModule:set_candidate(Leader) of
         {error, key_not_found} ->
             timer:sleep(1000),
             {continue_mining, {error, no_solution} };
         ok ->
+            timer:sleep(ChildBlockTime),
```
hanssv, line 395: What is the purpose of sleeping in the code for signing a key block header? (The sleep in the error branch also looks dubious, but that seems legacy...)

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -447,10 +452,92 @@ pc_start_height() ->
...
+            lager:info("New epoch at parent height ~p", [PCHeight]),
+            Seed = hash_to_int(Hash),
+            State = rand:seed_s(exsss, Seed),
+            Result = {State, Hash, Height, PCHeight},
+            aeu_ets_cache:put(?ETS_CACHE_TABLE, seed, Result),
+            Result
+    end.
+
+next_entropy(Hash, Height, PCHeight) ->
+    {NewState, _Seed, _Height, NewPCHeight} = case aeu_ets_cache:lookup(?ETS_CACHE_TABLE, seed) of
```
hanssv, line 525: Linebreak and less indentation for `case`?

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -590,30 +672,32 @@ next_beneficiary() ->
     PCHeight = pc_height(NextHeight),
     case aec_parent_chain_cache:get_block_by_height(PCHeight) of
```
hanssv, line 673: This logic will still ensure that PC and CC are run in lock-step, right?

hanssv (commented):
I am not convinced the logic in `entropy`/`set_new_entropy`/`next_entropy` is correct, it checks for finality (in four different cases according to comments) but eventually it still does `next_entropy_from_seed`? Perhaps some unit tests to convince the reader it works as intended?

I think we shouldn't compute EPOCH-length many seeds, rather one seed and then compute the whole EPOCH schedule in one go?!

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -166,6 +171,7 @@ start_ae(StakersEncoded, PCSpendAddress) ->
             end,
             StakersEncoded),
     StakersMap = maps:from_list(Stakers),
+    lager:warning("Stakers ~p", [StakersMap]),
```
ThomasArts, line 174: Is this really a warning? I would say it is more an info level log

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -447,10 +452,92 @@ pc_start_height() ->
...
+            aeu_ets_cache:put(?ETS_CACHE_TABLE, seed, Result),
+            Result
+    end.
+
+next_entropy(Hash, Height, PCHeight) ->
+    {NewState, _Seed, _Height, NewPCHeight} = case aeu_ets_cache:lookup(?ETS_CACHE_TABLE, seed) of
+                                {ok, {_State, _OldSeed, _OldHeight, PCHeight} = Result} ->
+                                    Result;
+                                _ ->
+                                    lager:warning("Trying to set entropy for parent height ~p", [PCHeight]),
```
ThomasArts, line 529: again... warning or info level here. I would say info, but probably I missed some design guidelines.

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -447,10 +452,92 @@ pc_start_height() ->
...
+                                    set_new_entropy(Hash, Height, PCHeight)
+                              end,
+    next_entropy_from_seed(NewState, Height, NewPCHeight).
+
+next_entropy_from_seed(State, Height, PCHeight) ->
```
ThomasArts, line 534: Connected to remarks of @Hans... would it not be better to prepare for taking the entropy from the first block of a new parent-chain epoch. Given that this PR is about adding epochs, we can use these epochs to pick the entropy from in the first block of such epochs.

(Clearly, if the code should still sync each block, then the epoch length is just 1)

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -590,30 +672,32 @@ next_beneficiary() ->
...
-            Entropy = aec_parent_chain_block:hash(Block),
-            CommitmentsSophia = encode_commitments(Block),
+            Entropy = entropy(Block, NextHeight),
             NetworkId = aec_parent_chain_block:encode_network_id(aec_governance:get_network_id()),
             {ok, CD} = aeb_fate_abi:create_calldata("elect_next",
                                                     [aefa_fate_code:encode_arg({string, Entropy}),
-                                                     CommitmentsSophia,
                                                      aefa_fate_code:encode_arg({bytes, NetworkId})
                                                     ]),
+            lager:warning("Entropy ~p ~p", [NextHeight, Entropy]),
```
ThomasArts, line 681: This should be an info, we don't want to be warned what the entropy is

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -676,6 +760,7 @@ is_leader_valid(Node, Trees, TxEnv, PrevNode) ->
                     Leader = aec_headers:miner(Header),
                     Target = aec_headers:target(Header),
                     IsDefaultT = Target =:= default_target(),
+                    lager:warning("is_leader_valid ~p ~p", [ExpectedLeader, Leader]),
```
ThomasArts, line 763: info, not wrning

apps/aecore/src/aec_parent_chain_cache.erl
```diff
@@ -312,59 +314,31 @@ state_to_map(#state{child_start_height = StartHeight,
...
+    MinHeight.
 
-post_block(Block, #state{max_size = MaxSize,
-                         top_height = TopHeight} = State0) ->
-    TargetHeight = target_parent_height(State0),
+max_cachable_parent_height(State) ->
+    [MaxHeight|_] = lists:reverse(lists:sort(target_parent_heights(State))),
+    MaxHeight.
+
+post_block(Block, #state{top_height = TopHeight} = State0) ->
```
ThomasArts, line 329: I am unsure about the working of the cache, but it seems that the privious version did consider a number of weird cases, potentiallly attacks or things spotted. These cases are now all ignored. Probably right but hard to see consequences of this, because tests for cache are disabled in this commit.

apps/aecore/src/aec_consensus_hc.erl
mitchelli: Sorry that was for my debugging. I'll remove it completely.

apps/aecore/src/aec_parent_chain_cache.erl
mitchelli: Only the tests related to commitments should be disabled. Which tests should I add back?

apps/aecore/src/aec_consensus_hc.erl
mitchelli: Currently it sets the entropy for the first block of the new child epoch by reading the parent hash from the cache.  The pc_height function decides which parent hash to choose. Is there something wrong with the logic?

mitchelli:
> I am not convinced the logic in `entropy`/`set_new_entropy`/`next_entropy` is correct, it checks for finality (in four different cases according to comments) but eventually it still does `next_entropy_from_seed`? Perhaps some unit tests to convince the reader it works as intended?
> 
> I think we shouldn't compute EPOCH-length many seeds, rather one seed and then compute the whole EPOCH schedule in one go?!

What I meant to say is that the external functions on the module can be called up to 4 times per block, which means the underlying entropy function will also be called four times. Since the entropy is cached it is only calculated once and the seed is only calculated the first time an epoch changes.

I will add some use cases, those warning messages that display the entropy are from when I was checking the behaviour manually.

apps/aecore/src/aec_consensus_hc.erl
hanssv: I assume this logic will move elsewhere once we start having proper time-slots etc, so perhaps this can stay for now.

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -1762,6 +1762,11 @@
                                                     "description": "The number of blocks in a parent generation.",
                                                     "type": "integer"
                                                 },
+                                                "retry_interval": {
```
hanssv, line 1765: Seems very similar to `polling > retry_interval` below?!

hanssv (dismissed):

ThomasArts (approved):

apps/aeutils/priv/aeternity_config_schema.json
mitchelli: Thanks, I'll move it into the polling section.

hanssv (approved):

ThomasArts (approved):

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -1746,6 +1754,14 @@
                                                     "default": 0,
                                                     "type": "integer"
                                                 },
+                                                "finality": {
```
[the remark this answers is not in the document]
mitchelli: No that is related to commitments. I think Måns will be revisiting that.

PR #4391, opened 2024-08-25
Define `pattern` for some schemas in OpenApi

davidyuk:

sdk uses the `pattern` property to generate more accurate TS types in https://github.com/aeternity/aepp-sdk-js/pull/2011

### DIFF
--- a/apps/aehttp/priv/oas3.yaml
+++ b/apps/aehttp/priv/oas3.yaml
@@ -3037,7 +3037,7 @@ components:
         tx:
           $ref: "#/components/schemas/EncodedTx"
         tx_hash:
-          $ref: "#/components/schemas/EncodedHash"
+          $ref: "#/components/schemas/EncodedTxHash"
         stateful:
           description: This call will have effects on the next call in this dry-run (or not)
           type: boolean
@@ -3090,7 +3090,7 @@ components:
         tx:
           $ref: "#/components/schemas/EncodedTx"
         tx_hash:
-          $ref: "#/components/schemas/EncodedHash"
+          $ref: "#/components/schemas/EncodedTxHash"
         call_req:
           $ref: "#/components/schemas/DryRunCallReq"
     DryRunResult:
@@ -3151,6 +3151,11 @@ components:
       type: string
       example: tx_+E8hAaEB4TK48d23oE5jt/qWR5pUu8UlpTGn8bwM5JISGQMGf7ABoQOvDVCf43V7alNbsUvTarXaCf7rjtWX36YLS4+JTa4jn4YPHaUyOAAAxRZ6Sg==
       pattern: ^tx_\w+$
+    EncodedTxHash:
+      description: Base58Check encoded transaction hash
+      type: string
+      example: th_2Zfo2ALfRkKQxEaUj3HhcUGyMrTqYpDMgr2u15fPanYD1d55kz
+      pattern: ^th_\w{38,50}$
     EncodedValue:
       description: Base58Check encoded tagged value
       type: string
@@ -3961,7 +3966,7 @@ components:
       type: object
       properties:
         tx_hash:
-          $ref: "#/components/schemas/EncodedHash"
+          $ref: "#/components/schemas/EncodedTxHash"
       required:
         - tx_hash
     Pow:
@@ -4031,7 +4036,7 @@ components:
         block_hash:
           $ref: "#/components/schemas/EncodedHash"
         hash:
-          $ref: "#/components/schemas/EncodedHash"
+          $ref: "#/components/schemas/EncodedTxHash"
         encoded_tx:
           $ref: "#/components/schemas/EncodedTx"
         signatures:
--- a/apps/aehttp/priv/oas3.yaml
+++ b/apps/aehttp/priv/oas3.yaml
@@ -1214,12 +1214,7 @@ paths:
         content:
           application/json:
             schema:
-              type: object
-              properties:
-                tx:
-                  $ref: "#/components/schemas/EncodedTx"
-              required:
-                - tx
+              $ref: "#/components/schemas/EncodedTx"
         description: The new transaction
         required: true
       responses:
@@ -2569,7 +2564,7 @@ components:
         from_id:
           $ref: "#/components/schemas/EncodedPubkey"
         payload:
-          $ref: "#/components/schemas/EncodedTx"
+          $ref: "#/components/schemas/EncodedTxByteArray"
         ttl:
           $ref: "#/components/schemas/UInt64"
         fee:
@@ -2653,7 +2648,7 @@ components:
         from_id:
           $ref: "#/components/schemas/EncodedPubkey"
         payload:
-          $ref: "#/components/schemas/EncodedTx"
+          $ref: "#/components/schemas/EncodedTxByteArray"
         round:
           $ref: "#/components/schemas/UInt64"
         update:
@@ -2696,7 +2691,7 @@ components:
         round:
           $ref: '#/components/schemas/UInt64'
         payload:
-          $ref: '#/components/schemas/EncodedTx'
+          $ref: '#/components/schemas/EncodedTxByteArray'
         ttl:
           $ref: '#/components/schemas/UInt64'
         fee:
@@ -2744,7 +2739,7 @@ components:
         from_id:
           $ref: "#/components/schemas/EncodedPubkey"
         payload:
-          $ref: "#/components/schemas/EncodedTx"
+          $ref: "#/components/schemas/EncodedTxByteArray"
         ttl:
           $ref: "#/components/schemas/UInt64"
         fee:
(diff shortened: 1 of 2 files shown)


apps/aehttp/priv/oas3.yaml
```diff
@@ -1080,7 +1078,12 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/EncodedTx"
+              type: object
```
ThomasArts, line 1081: Is this not "breaking" the specification in such a way that you would need a new API version number??
You re-use the same name, but give it a different meaning.

apps/aehttp/priv/oas3.yaml
```diff
@@ -2351,15 +2354,15 @@ components:
         from_id:
           $ref: "#/components/schemas/EncodedPubkey"
         payload:
-          $ref: "#/components/schemas/EncodedByteArray"
+          $ref: "#/components/schemas/EncodedTx"
```
ThomasArts, line 2357: This really breaks it, since this is not what the code does. In this stage, I think we can only accept OAS changes that leave the code as is.

ThomasArts (changes requested):

apps/aehttp/priv/oas3.yaml
davidyuk: Do you mean the payload is not only a transaction?

apps/aehttp/priv/oas3.yaml
```diff
@@ -2903,24 +2906,64 @@ components:
...
       description: Base58Check encoded tagged pubkey
       type: string
     EncodedTx:
-      type: object
-      properties:
-        tx:
-          $ref: "#/components/schemas/EncodedByteArray"
-      required:
-        - tx
-    EncodedValue:
```
davidyuk, line 2921: @ThomasArts is it okay to remove this type since it is unused anymore?

apps/aehttp/priv/oas3.yaml
```diff
@@ -2903,8 +2901,23 @@ components:
       required:
         - results
     EncodedByteArray:
-      description: Base64Check encoded tagged byte array
+      description: Base64Check encoded byte array
+      type: string
+      example: ba_AAAAAfy4hFE=
+      pattern: ^ba_\w+$
```
venimus: `\w` is `[A-Za-z0-9_]` — it excludes `+`, `/` and `=`, but this is **standard Base64** (not URL-safe), which uses exactly those characters. Proof: the example directly below (`EncodedTxByteArray`'s `tx_+E8hAaEB4TK48d23oE5jt/qWR5pUu8UlpTGn8bwM5JISGQMGf7ABoQOvDVCf43V7alNbsUvTarXaCf7rjtWX36YLS4+JTa4jn4YPHaUyOAAAxRZ6Sg==`) contains `+`, `/` and `==` and fails this exact style of pattern.

I verified this isn't just a TS-typegen nuance: `apps/aehttp/test/aehttp_validation_SUITE.erl` validates real HTTP responses against `oas3.yaml` via `jesse` (a JSON-Schema `pattern`/regex validator), so as written this would start rejecting real, valid byte arrays/tx encodings that happen to contain `+`, `/`, or `=` — which is common since it's essentially random binary data.

The historical (commented-out) `swagger.yaml` already had the correct character class for this: `^(cb|tx|...)_[0-9A-Za-z+/=]*$`.

```suggestion
      pattern: ^ba_[0-9A-Za-z+/=]+$
```

apps/aehttp/priv/oas3.yaml
```diff
@@ -2903,8 +2901,23 @@ components:
...
+    EncodedContractByteArray:
+      description: Base64Check encoded contract byte array
+      type: string
+      example: cb_AAAAAfy4hFE=
+      pattern: ^cb_\w+$
```
venimus: Same Base64Check alphabet issue as `EncodedByteArray` above — `\w+` excludes `+`/`/`/`=`.
```suggestion
      pattern: ^cb_[0-9A-Za-z+/=]+$
```

apps/aehttp/priv/oas3.yaml
```diff
@@ -2903,8 +2901,23 @@ components:
...
+    EncodedProofOfInclusion:
+      description: Base64Check encoded proof of inclusion
       type: string
+      pattern: ^pi_\w+$
```
venimus: Same Base64Check alphabet issue.
```suggestion
      pattern: ^pi_[0-9A-Za-z+/=]+$
```

apps/aehttp/priv/oas3.yaml
```diff
@@ -2903,8 +2901,23 @@ components:
...
+    EncodedStateTrees:
+      description: Base64Check encoded state trees
+      type: string
+      pattern: ^ss_\w+$
```
venimus: Same Base64Check alphabet issue.
```suggestion
      pattern: ^ss_[0-9A-Za-z+/=]+$
```

apps/aehttp/priv/oas3.yaml
```diff
@@ -2915,12 +2928,44 @@ components:
...
+          $ref: "#/components/schemas/EncodedTxByteArray"
       required:
         - tx
-    EncodedValue:
-      description: Base58Check encoded tagged value
+    EncodedTxByteArray:
+      description: Base64Check encoded transaction
+      type: string
+      example: tx_+E8hAaEB4TK48d23oE5jt/qWR5pUu8UlpTGn8bwM5JISGQMGf7ABoQOvDVCf43V7alNbsUvTarXaCf7rjtWX36YLS4+JTa4jn4YPHaUyOAAAxRZ6Sg==
+      pattern: ^tx_\w+$
```
venimus: Same Base64Check alphabet issue — and this one's example literally violates its own pattern (see above).
```suggestion
      pattern: ^tx_[0-9A-Za-z+/=]+$
```

apps/aehttp/priv/oas3.yaml
```diff
@@ -2915,12 +2928,44 @@ components:
...
+    EncodedSignature:
+      description: Base58Check encoded signature
+      type: string
+      example: sg_8hhU15cVMbukFj4FbdrFwwYnbzXYPnstu9PDnWZGbfQNpeHcB6tK1F3wvG1MPYySARgDRJYUh3YPJD3HctFwg6Y4rUGSR
+      pattern: ^sg_\w+$
+    Name:
+      description: AENS name
+      type: string
+      example: dimitar.chain
+      pattern: ^\w+\.chain$
```
venimus: This pattern is too restrictive in two ways:
1. Hardcodes `.chain` only, but `aec_governance:possible_name_registrars/0` returns `[<<"chain">>, <<"test">>]` — `.test` is a legacy but still-recognized registrar (see `aens_hash.erl`'s handling of ".test domain" names). A lookup for an old `.test` name would fail client-side validation before ever reaching the node.
2. `\w+` excludes `-`, but internationalized AENS names are IDNA/punycode-encoded with an `xn--` prefix containing hyphens — this repo's own test demonstrates it: `aens_utils_test:to_ascii(<<"詹姆斯詹姆斯.chain"/utf8>>) -> {ok, <<"xn--8wsa062gba0028bca.chain">>}`. This pattern would reject that name.

```suggestion
      pattern: ^[\w-]+\.(chain|test)$
```

apps/aehttp/priv/oas3.yaml
```diff
@@ -3146,7 +3191,10 @@ components:
         height:
           $ref: "#/components/schemas/UInt64"
         pof_hash:
-          $ref: "#/components/schemas/EncodedHash"
+          description: \"no_fraud\" or Base58Check encoded Proof of Fraud hash
+          type: string
+          pattern: ^no_fraud|bf_\w{38,50}$
```
venimus: Missing grouping parens: `^no_fraud|bf_\w{38,50}$` parses as `(^no_fraud)|(bf_\w{38,50}$)` — i.e. "starts with `no_fraud`" (unanchored at the end) OR "ends with `bf_...`" (unanchored at the start). That means e.g. `no_fraud_extra_garbage` or `garbage_bf_<38-50 chars>` would both incorrectly pass.

```suggestion
          pattern: ^(no_fraud|bf_\w{38,50})$
```

venimus (commented):
Went through the full diff and cross-checked patterns against the actual node encoders (`aeser_api_encoder`), governance config (`aec_governance:possible_name_registrars/0`), and the runtime `jesse`-based schema validation in `apps/aehttp/test/aehttp_validation_SUITE.erl` (so these aren't just cosmetic — they'd affect real request/response validation). Left inline suggestions for the fixes; all suggested patterns were verified to (a) match their own declared `example` values and (b) reject/accept the right edge cases (checked with a quick script, happy to share).

Also re: the older open threads on this PR (the `/transactions` POST double-wrapping concern, the byte-array naming-convention comment, and the `ChannelCloseSoloTx.payload` "this breaks the code" thread) — I checked the current diff against `master` and against the actual channel tx serialization code (`aesc_close_solo_tx.erl` et al., which does `aeser_api_encoder:encode(transaction, Payload)`, i.e. a plain `tx_`-prefixed string). All three look like they were already resolved by later commits in this PR (the problematic object-wrapping was reverted, and `payload` is now correctly typed as `EncodedTxByteArray`, a plain string) — they just weren't marked resolved on GitHub. Might be worth closing those out. The one remaining open question ("is it okay to remove `EncodedValue`?") — confirmed there are no remaining `$ref`s to it anywhere in the current spec, so yes, safe to remove.

PR #4397, opened 2024-09-02
Use better hyperchain configuration names 

mitchelli:

Use pinning_accounts instead of stakers and pinning_recipient instead of spend_address

See https://github.com/aeternity/aeternity/issues/4388

### DIFF
--- a/apps/aecore/src/aec_consensus_hc.erl
+++ b/apps/aecore/src/aec_consensus_hc.erl
@@ -94,29 +94,29 @@ can_be_turned_off() -> false.
 assert_config(_Config) -> ok.
 
 start(Config, #{block_production := BlockProduction}) ->
-    #{ <<"stakers">>      := StakersConfig,
-       <<"parent_chain">> := PCConfig      } = Config,
+    #{ <<"pinning_accounts">>    := PinningAccountsConfig,
+       <<"parent_chain">>        := PCConfig      } = Config,
 
-    StakersConfig = maps:get(<<"stakers">>, Config, []),
-    PCConfig      = maps:get(<<"parent_chain">>, Config),
+    PinningAccountsConfig = maps:get(<<"pinning_accounts">>, Config, []),
+    PCConfig              = maps:get(<<"parent_chain">>, Config),
 
-    Confirmations        = maps:get(<<"confirmations">>, PCConfig, 6),
-    StartHeight          = maps:get(<<"start_height">>, PCConfig, 0),
-    ProducingCommitments = maps:get(<<"producing_commitments">>, PCConfig, false),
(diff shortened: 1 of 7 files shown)


dincho:
Didn't we agreed to use the sender as recipient ?

klercker:
Name change in itself all cool, but this might get really confusing, as the action performed by the code is actually the same old commitment code, but now with pinning terms thrown into the mix. TBH I think perhaps it's better to change terminology and functionality at the same time? I'm just mucking about in this code, and would've been very confused with this. But perhaps this type of "in-flight" change is what is required if we feel the need to PR stuff into master daily?

mitchelli:
> Didn't we agreed to use the sender as recipient ?

Yes but since the commitment code is being replaced with pinning I wasn't sure if it would be needed. I'll checks with Måns.

mitchelli:
This pull request I think makes the code a little clearer because stakers was also used in another context. But I see it could make things difficult for you.

klercker:
Looking at it again, I think that we should skip this. The pinning process is quite different to commitment (although the actual transactions to the PC is not), and for example these lists of Stakers etc are not even needed for pinning. Renaming them at this point adds only confusion and nothing else IMHO. They should be removed as part of the code refac/rewrite instead (I'm doing this now but I'm by epoch time is much different from the rest of the team at this moment). I will essentially remove all the commitment code as part of the issues I'm resolving, and some of it will be refurbished as part of pinning. OK?

mitchelli:
Sounds good to me. I'll close the related issues.

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -94,65 +94,65 @@ can_be_turned_off() -> false.
...
-
-    FetchInterval  = maps:get(<<"fetch_interval">>, PollingConfig, 500),
-    RetryInterval  = maps:get(<<"retry_interval">>, PollingConfig, 1000),
-    CacheSize      = maps:get(<<"cache_size">>, PollingConfig, 200),
-    Nodes          = maps:get(<<"nodes">>, PollingConfig, []),
-    ParentHosts    = lists:map(fun aehttpc:parse_node_url/1, Nodes),
+    #{ <<"pinning_accounts">>    := PinningAccountsConfig,
+       <<"parent_chain">>        := PCConfig      } = Config,
+
+    PinningAccountsConfig = maps:get(<<"pinning_accounts">>, Config, []),
```
ThomasArts, line 100: What are pinning accounts? And what are pinning recipients?

dincho:
> lists of Stakers etc are not even needed for pinning.

@klercker I then wonder how the actual transaction will be signed if there are no accounts configured?

PR #4398, opened 2024-09-03
Added initial version of hacking guide

richcarl:
An overview of the codebase to help programmers.


docs/hacking.md
```diff
@@ -0,0 +1,240 @@
+Hacking the Aeternity Codebase
+
+- Building
+  - See [Build](build.md)
+    - Ubuntu dependencies: `sudo apt install autoconf build-essential cmake
+      erlang libsodium-dev libgmp-dev`
+    - Mac OS: `brew install erlang@24 openssl libsodium autoconf gmp cmake
```
hanssv, line 7: Picking the _oldest_ OTP-version we support is a sure way to make it outdated as quickly as possible 😅

richcarl:
Can I get a couple of approvals so I can merge this? Then we can all help adding and modifying.

dincho (approved):
The bits on the relx and rebar3 seems to be outdated, we have copied and modified(?) some of the relx binaries/scripts I think.

richcarl:
> The bits on the relx and rebar3 seems to be outdated, we have copied and modified(?) some of the relx binaries/scripts I think.

Let's investigate that and update the doc later in that case.

dincho:
> > The bits on the relx and rebar3 seems to be outdated, we have copied and modified(?) some of the relx binaries/scripts I think.
> 
> Let's investigate that and update the doc later in that case.

While I did approved this PR, it doesn't make sense in general to commit an outdated info into master

hanssv:
Wouldn't this info fit better in the wiki? Though that is probably even more outdated.

hanssv (approved):

PR #4400, opened 2024-09-04
Add hc specific test suite

ThomasArts:
First test suite for hyperchains where we can steer block production in detail

Splits aehttp_stake_contract SUITE into one PoS suite and one HC suite. This removes BTC/Doge connection testing for now. It introduces mining steering for HC, including distributing validators on different nodes.


klercker (approved):
Have worked off most of these commits the last day or two. works well.

klercker (approved):

klercker:
Will this PR include the `more_hc_spcific_tests` commits as well? Would be good of course...

hanssv (approved):

PR #4406, opened 2024-09-07
PC/CC speed sync

mitchelli:
Introduces Epochs in contract instead of code
The notion of length of an epoch is now part of the contract


apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -472,6 +472,13 @@ parent_generation() ->
     Fun = fun() -> get_consensus_config_key([<<"parent_chain">>, <<"parent_generation">>]) end,
     aeu_ets_cache:get(?ETS_CACHE_TABLE, parent_generation, Fun).
 
+acceptable_sync_offset() ->
+    Fun = fun() -> get_consensus_config_key([<<"parent_chain">>, <<"acceptable_sync_offset">>], 60000) end,
```
ThomasArts, line 476: Should this not be in percent of a parent epoch or so?

Feels strange that a note can on itself just configure a number. Surely at the end one wants to vote whether or not to increase/decrease child epoch length. Bit on what basis? If each node can have its own config setting and decides based on that, then "consensus" of change is a weird thing.

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -472,6 +472,13 @@ parent_generation() ->
...
+    aeu_ets_cache:get(?ETS_CACHE_TABLE, acceptable_sync_offset, Fun).
+
+child_epoch_length(Length) ->
```
ThomasArts, line 479: Should this be saved "on chain"?
If one later synchronizes with an existing HC, then one probably needs to know what child epoch length was at a given height? If so, we may want this information stored on chain.

A possible place to store this is having it in the election contract.
Another possibility, but more intrusive, is to have a special state tree for this.

mitchelli: I agree it makes sense to have the sync offset and the epoch length on chain.

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -1762,6 +1758,11 @@
                                                     "description": "The number of blocks in a parent generation.",
                                                     "type": "integer"
                                                 },
+                                                "acceptable_sync_offset": {
+                                                    "description": "The maximum amount of time the parent chain block generation can be off the expected time.",
```
happi, line 1762: Perhaps add the unit (ms?) to the description.

test/contracts/HCElection.aes
klercker, line 1: As we discussed, the "API"/information defined in this and other contracts should be available as HC chain information in ```aec_chain.erl``` (or something similar for HC chains) to make it easily available to other code.

klercker (approved):
Looks reasonable to me, good to see things starting to coming together.

hanssv (approved):

PR #4407, opened 2024-09-09
Use Config to store active nodes in

ThomasArts:
Make tests aware which nodes are running and close down only running nodes at end of tests.

Common test is restricted in passing on information from one test to another (for good reasons), but the context of what nodes are running is probably the kind of data one really wants to be passed on.

### DIFF
--- a/apps/aehttp/test/aehttp_hyperchains_SUITE.erl
+++ b/apps/aehttp/test/aehttp_hyperchains_SUITE.erl
@@ -25,8 +25,8 @@
          empty_parent_block/1,
          sync_third_node/1,
          verify_fees/1,
-         % elected_leader_did_not_show_up/1,
-         % block_difficulty/1,
+         elected_leader_did_not_show_up/1,
+         block_difficulty/1,
          epoch_with_slow_parent/1
         ]).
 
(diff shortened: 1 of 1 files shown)


hanssv (approved):

apps/aehttp/test/aehttp_hyperchains_SUITE.erl
```diff
@@ -1095,12 +1116,12 @@ mine_key_blocks(ParentNodeName, NumParentBlocks) ->
     ct:log("Parent block mined ~p ~p number: ~p", [KBs, ParentNodeName, NumParentBlocks]),
     {ok, KBs}.
 
-get_block_producer_name(Parties, Node, Height) ->
-    Producer = get_block_producer(Node, Height),
-    case lists:keyfind(Producer, 1, Parties) of
-        false -> Producer;
-        {_, _, Name} -> Name
-    end.
+%get_block_producer_name(Parties, Node, Height) ->
```
happi, line 1119: Why not deleted?

happi (approved):

apps/aehttp/test/aehttp_hyperchains_SUITE.erl
ThomasArts: I believe I'll need it if I restore the other tests

Issue #4394, opened 2024-09-02
HC: Remove Comittment code from everywhere


klercker:
sub to #4393 

Since comittment is not done anymore, remove all such code from repo. Will be replaced by somewhat similar function for pinning, in somewhat the same places.

PR #4409, opened 2024-09-09
#4394 Removed all HC Commitment code

klercker:
Removal of all HC Comittment code
- functionality
- method args
- tests
- dead code removal


apps/aecore/src/aec_parent_connector.erl
```diff
@@ -132,31 +112,20 @@ fetch_block_by_hash(Hash) ->
...
-        true -> ok
-    end,
-    CDetails =
-        #commitment_details{ parent_network_id  = NetworkId, %% TODO: assert all nodes are having the same network id
-                             sign_module = SignModule,
-                             recipient = Recipient,
-                             amount = Amount,
-                             fee = Fee
-                            },
+init([ParentConnMod, FetchInterval, ParentHosts, _NetworkId, _SignModule, HCPCPairs, _Recipient, _Fee, _Amount]) ->
```
hanssv, line 120: Are you expecting to use these arguments for pinning or could they be removed completely?

apps/aecore/src/aec_parent_connector.erl
```diff
@@ -132,31 +112,20 @@ fetch_block_by_hash(Hash) ->
...
+    % if is_integer(FetchInterval) ->
+    %     erlang:send_after(FetchInterval, self(), check_parent);
+    %     true -> ok
+    % end,
```
hanssv, line 124: Remove?

hanssv (commented):
Looks good to me, some minor comments.

Also I suspect there is a bit too much removed in BTC/Doge HTTP clients - pinning will be creating/signing/posting transactions 🤔 Since we are supposed to support at least BTC we probably want to keep that?

apps/aecore/src/aec_parent_connector.erl
```diff
@@ -8,15 +8,6 @@
 
 -module(aec_parent_connector).
 
-%% Functionality:
```
ThomasArts, line 11: You deleted all functionality, does that mean that the remaining code has no functionality at all?

klercker:
Yes, A bunch of code that is very likely usable in pinning was removed
there. Unused methods were causing fails in dialyzer tests (at least that
was what it looked like). I will dig in Git history and reinsert once we
get to that point.

klercker:
Exactly :P

apps/aecore/src/aec_parent_connector.erl
klercker: I must admit I was a bit scared to remove them completely, but as that was last week I might have gathered up some courage now. Will take a look again then...

apps/aecore/src/aec_parent_connector.erl
klercker: Hmm, yes perhaps.

apps/aehttp/src/aehttpc_aeternity.erl
```diff
@@ -7,9 +7,6 @@
 -export([get_latest_block/2,
```
ThomasArts, line 7: My feeling is that this whole module can be deleted if we don't care about commitments. What else did it offer that is not in the standard http client software??

hanssv:
> Yes, A bunch of code that is very likely usable in pinning was removed there. Unused methods were causing fails in dialyzer tests (at least that was what it looked like). I will dig in Git history and reinsert once we get to that point.

Either (temporarily) export the functions, or keep them commented out. I think the first is preferred.

apps/aehttp/src/aehttpc_aeternity.erl
ThomasArts: For similarity to bitcoin and dodge module, we could keep it, of course. We still need to be able to see what epoch the parent chain is in

apps/aehttp/src/aehttpc_doge.erl
```diff
@@ -193,67 +94,17 @@ getblock(NodeSpec, Seed, Hash, _Verbosity) ->
...
-getrawtransaction(NodeSpec, TxHash) ->
-    try
-        Seed = <<>>,
-        Body = jsx:encode(request_body(<<"getrawtransaction">>, [TxHash, true], seed_to_utf8(Seed))),
-        {ok, Res} = request(<<"/">>, Body, NodeSpec, 5000),
-        {ok, result(Res)}
-    catch E:R ->
-        {error, {E, R}}
-    end.
+% -spec getrawtransaction(aehttpc:node_spec(), binary()) -> {ok, binary()} | {error, term()}.
```
ThomasArts, line 98: Possibly useful as well when dealing with dodge as parent chain?

apps/aehttp/test/aehttp_parent_connector_tests.erl
```diff
@@ -10,7 +10,8 @@
 -define(CHILD_CHAIN_NETWORK_ID, <<"hc_network_id">>).
 -define(SIGN_MODULE, aec_preset_keys).
 
-ae_sim_test_() ->
+% removed as test, should/could be removed altogether, same goes for btc version below.
+ae_sim_test_rem() ->
```
ThomasArts, line 14: `test_` as ending of a function name has a specific meaning. Cannot just at `rem` after it without changing that meaning. See eunit documentation for detail

apps/aehttp/src/aehttpc_doge.erl
```diff
@@ -165,24 +66,24 @@ getblockhash(NodeSpec, Seed, Height) ->
...
-                    {ok, Tx} -> [Tx];
-                    _ -> []
-                end
-            end, TxHashes),
-        Block = block(maps:put(<<"tx">>, Txs, Obj)),
-        {ok, Block}
-    catch E:R:S ->
-        {error, {E, R, S}}
-    end;
+% getblock(NodeSpec, Seed, Hash, 2) ->
```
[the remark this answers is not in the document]
klercker: It's not used anywhere, and is only called with (_,_,_,1) internally, but I kept it because it felt, I dunno, prudent(?)

apps/aehttp/test/aehttp_parent_connector_tests.erl
klercker: I know, this leaves that particular test out of the suite. So here I could keep that code, in case it was actually testing something besides commitments. But perhaps it is not?

apps/aehttp/src/aehttpc_doge.erl
klercker: Again, I can reintroduce all the pruned code from these aehttpc modules (or whatever they are called) and export them. then these would make more sense. And not trigger build fails I suppose?

apps/aehttp/test/aehttp_parent_connector_tests.erl
klercker: So, my question is if the two "_rem" methods (that are left in the code but not invoked as tests...) actually do anything but commitment testing, or of they can be removed altogether?

apps/aehttp/src/aehttpc_aeternity.erl
klercker: It's just a wrapper but since it is a wrapper that gets dynamically selected (ae, btc, doge) there needs to be one for aeternity as well. I kind-of like this way, so I think it should be kept.

apps/aehttp/src/aehttpc_doge.erl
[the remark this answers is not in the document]
klercker: @ThomasArts the problem with that particular method (head) is that it fails dialyzer:

The pattern <NodeSpec, Seed, Hash, 2> can never match the type <_,_,_,1>

That is why I edited it out What to you suggest I do with it? Expose it directly, write additional wrappers for get_latest_blocks?

apps/aecore/src/aec_parent_connector.erl
```diff
@@ -8,23 +8,20 @@
...
-%%   available.
-%% - New top block from parent chain should include commitment transactions
-%%   on the parent chain.
-%% - Call consensus smart contract
-%% Parent chain commitment txs:
-%% - To AE parent - signed by a parent chain AE account hosted in the HC node via aec_keys
-%% - To BTC/Doge parent - signed by the external bitcoind/dogecoind hosted wallet
+%% Functionality
+%% Holds parent chain connection state and contains various utilities for 
+%% interacting withvthe PC. Wraps parent implementation specific stuff
```
ThomasArts, line 13:
```suggestion
%% interacting with the PC. Wraps parent implementation specific stuff
```

apps/aehttp/test/aehttp_parent_connector_tests.erl
hanssv: You want to write very similar tests as soon as you add pinning, right?

hanssv (approved):

klercker:
So, these worked on my build locally. Intermittent? Re-run?

ThomasArts (approved):

PR #4414, opened 2024-09-12
Add docs on testing and Rebar3

richcarl:
- hacking.md - reformatted to use headings instead of just bullets (lots of noise - sorry, but no real changes in this file).
- testing.md - brings in the text on Testing from the GitHub wiki pages ("where documentation goes to die").
- rebar.md - quick guide to Rebar3


klercker (approved):
good stuff

kvakvs:
I had some situations where `rebar {something}` would fail because its counterpart `make {something}` had extra dependencies and extra build steps. Namely, i remember running ct via rebar was a problem, but works with Makefile. I.e. rebar and make targets with same name are not equivalent.

ThomasArts:
Do we really advocate people to test via starting an Erlang shell and then manually starting applications... sounds to me like a very error prone and old fashion way of testing?

richcarl:
> Do we really advocate people to test via starting an Erlang shell and then manually starting applications... sounds to me like a very error prone and old fashion way of testing?

I just collected the Testing stuff from the wiki. That particular part was written by yourself in 2017 :-) - should we drop it?

hanssv:
Is it important for Hyperchains?

dincho (approved):
Finally the rebar magic documented in a single place :)

PR #4427, opened 2024-10-02
Add configuration files for devcontainers

richcarl:


richcarl:
> Cool, I didn't know you could do this without Dockerfiles and such. What exactly is in that container? Do you want me to test on macos?

That would be nice. It just uses the existing Aeternity build image, and also adds some plugin recommendations. If you use VSCode it should just pick it up automatically and ask if you want to run in a container and install the plugins.

klercker:
Started on Macos Sequoia and I could `make prod-build` and `make ct-latest`. Pretty slow though, but that could of course be bcz of limited resources inside of the container. Also got some errors I didn't recognize while doing the prod-build, although everything seems to work ok.

klercker:
don't know if you need to rebase, but re-run tests to make them pass end then Bob's yer' uncle.

klercker:
One interesting thing is that I got the complete(?) list of AE GH repos in the VSC repo view when the cloned aeternity repo was opened in the devcontainer. That was a bit disconcerting and somewhat confusing.

richcarl:
Rebased. Let's see if that helps.

klercker (approved):
Good stuff, tested on macos and worked excellent for (at least) build and ct. Useful.

hanssv:
I'm either old or out of fashion (or both 😅). What am I looking at? What is a devcontainer and why would I need one?

richcarl:
It just means that when VSCode or other editors that understand this see the config file, they will offer to launch a container as specified (you can either point to an image, as I did, or write a Dockerfile that it will build on the fly). The editor will map the local workspace directory into the container and open a connection to it so you can build, run shell commands etc., from within the container, ignoring whatever operating system you are actually running. (On Windows, all containers run via Docker-with-WSL-as-backend.)
You don't need to know much to set anything up - you just say "yes" when VSCode asks you if you want to run it containerized.  In this case, the image used will be the same as we use for building Aeternity releases.

hanssv:
No, _the editor_ will not ask about this, I've restarted Vim twice with this branch just to make sure... 😛 

I'll let everyone enjoying VSCode review this one.

richcarl:
> No, _the editor_ will not ask about this, I've restarted Vim twice with this branch just to make sure... 😛
> 
> I'll let everyone enjoying VSCode review this one.

Ask Google and it will provide: https://github.com/esensar/nvim-dev-container

happi (approved):

hanssv:
Let's not confuse neovim with vim 😅 

@klercker I used to use Emacs, but converted to the brighter side almost 10 years ago now!

richcarl:
> Let's not confuse neovim with vim 😅

Vim, schwimm, ...

PR #4430, opened 2024-10-03
Parent connector return chain type

klercker:
- aehttpc type define get_chain_type()- implemented by ae/btc/doge impls
- pinning agent updated to use get_chain_type
- disaptcher updated to handle atom() instead of binary() type
- test case updated to check for type aeternity


klercker:
I just realized, I could push it all the way up to the parent_connector..., now I get the module and call it. hmmm

klercker:
Addressed my own concern and added the parent_connector wrapper around the module call. So now we do what it says in the PR title.

ThomasArts (approved):

happi (approved):

PR #4431, opened 2024-10-04
Hc pinning from wallet

ThomasArts:
If external part is going to post pin then nice to know who last leader is in order to signal where the pin landed on parent chain.

### DIFF
--- a/apps/aecore/src/aec_pinning_agent.erl
+++ b/apps/aecore/src/aec_pinning_agent.erl
@@ -1,7 +1,7 @@
 %%%-------------------------------------------------------------------
 %%% @copyright (C) 2024, Aeternity
 %%% @doc
-%%% Gather child->parent chain pinning information for AE HC and make it 
+%%% Gather child->parent chain pinning information for AE HC and make it
 %%% available through API
 %%% @end
 %%%-------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


apps/aecore/src/aec_pinning_agent.erl
```diff
@@ -20,28 +20,46 @@
...
     {ok, BlockHash} = aec_chain_state:get_key_block_hash_at_height(First-1),
     {ok, ChainType} = aec_parent_connector:get_parent_chain_type(),
-    
-    #{epoch             => Epoch-1,
-      height            => First-1,
-      block_hash        => BlockHash,
-      parent_type       => ChainType,
-      parent_network_id => aec_parent_connector:get_network_id()}.
+    ConMod = aec_parent_connector:get_parent_conn_mod(),
+    {_,Type} = lists:split(8,atom_to_list(ConMod)), % split off "aehttpc_" from mod name to get type
```
hanssv, line 32:
```suggestion
   "aehttpc_" ++ Type = atom_to_list(ConMod),
```

klercker: or just use the the parent_connector:get_chain_type() implemented in #4430, it´s on `master`

ThomasArts: Removed this altogether, since now type is known by parent connector.

klercker (approved):

apps/aecore/src/aec_pinning_agent.erl
```diff
@@ -20,28 +20,44 @@
...
+    end.
 
 
 -spec encode_pin_payload(#{epoch => integer(), height => integer(), block_hash => binary()}) -> binary().
 encode_pin_payload(#{epoch := Epoch, height := Height, block_hash := BlockHash}) ->
-    <<Epoch:64/integer, Height:64/integer, BlockHash/binary>>.
+    EpochHex = list_to_binary(erlang:integer_to_list(Epoch, 16)),
+    HeightHex = list_to_binary(erlang:integer_to_list(Height, 16)),
+    EncBlockHash = aeser_api_encoder:encode(key_block_hash, BlockHash),
+    <<EpochHex/binary, ":", HeightHex/binary, " ", EncBlockHash/binary>>.
```
hanssv, line 53: Shouldn't this be a proper serialization? Wasn't there a size limitation (on Bitcoin), then a colon, a space and an API-encoded thing can't be correct?

hanssv (approved):

apps/aecore/src/aec_pinning_agent.erl
hanssv: Let's worry about that if it turns out a problem

ThomasArts: Point made: if parent chain can recognise what is posted, these transaction can be sensored. Possibly a valid point.
In this solution we made it even human readable, so certainly filterable.
Other point made: more streamline if we would serialize in aeser_api_encode for this and have a special type `hc_pin` for this. That is indeed more coherent. 

I made an issue: https://github.com/aeternity/aeternity/issues/4435

PR #4432, opened 2024-10-07
Mining helper for hc tests

kvakvs:
## Motivation

Current tests for hyperchains are highly experimental and wordy. A lot of code is used to express simple actions, and it would be great if the scene set up for a test would be simpler and easier to read.

## Proposal

Proposing an improvement for hc test flow:

A new module which runs a list of commands, mining-related actions. This can start/stop mining, mine a pc block, mine a cc block, key/micro block, wait, sync, and assert that no transactions are hanging uncommitted. And you also specify a node which will do the work.

This new module takes a bunch of helper functions from aehttp_hyperchains_suite, so eventually those can move here permanently and the original hyperchains suite will instead use shorthand commands to perform the mining actions.

An example of test flow looks like:
```erlang
aecore_suite_director:run([
  {cc, dev1, 3},
  {kb, dev2, 1},
  {wait, 1000},
  {sync, [dev1, dev2, dev3]},
  {assert_empty_mempool, dev3}
], CtConfig).
```

## TODO ideas
Currently the new module and aehttp_hyperchains_suite share a bunch of macros, account key pairs etc. This can be something you pass once at setup, so they don't share this data anymore.

CtConfig is necessary currently, because it carries list of all nodes, and some other stuff. This might be a sub-goal to eliminate CtConfig from the helper module.

## Credits

This work supported by a grant from Aeternity foundation


klercker:
Like this! Right now sitting with code to "fast forward" to next epoch. Perhaps an action for this framework?

klercker:
You want to add "supported by a grant from Aeternity foundation" to the issue as well.

Issue #4418, opened 2024-09-17
Micro block before key block


hanssv:
Implement the possibility to build a chain starting with micro blocks and finish with a key block

PR #4434, opened 2024-10-07
Reverse key-block/micro-block order

hanssv:
This is a complex operation - some of the existing tests for PoW should preferably be run also on this...

This change also taints the existing code (especially the rewrite of `get_generation_by_<height/hash` is invasive) but it is impossible to avoid as far as I can see. But let's see what the review can unravel.

This implements #4418.

### DIFF
--- a/apps/aecore/src/aec_chain.erl
+++ b/apps/aecore/src/aec_chain.erl
@@ -753,7 +753,13 @@ get_generation_by_height(Height, KeyBlock, forward) ->
             case get_key_block_by_height(Height + 1) of
                 {error, _Reason} -> error;
                 {ok, KeyBlock1}  ->
-                    get_generation_by_height(Height, KeyBlock, aec_blocks:prev_hash(KeyBlock1), forward)
+                    {ok, KeyBlockHash} = aec_blocks:hash_internal_representation(KeyBlock),
+                    case aec_blocks:prev_key_hash(KeyBlock1) of
+                        KeyBlockHash ->
+                            get_generation_by_height(Height, KeyBlock, aec_blocks:prev_hash(KeyBlock1), forward);
+                        _ ->
+                            error
+                    end
             end
     end.
 
@@ -766,6 +772,8 @@ get_generation_by_height(Height, KeyBlock, LastMBHash, forward) ->
     MBHeight = ConsensusModule:micro_block_height_relative_previous_block(key, Height),
     get_generation(forward, KeyBlock, MBHeight, LastMBHash).
 
+get_generation(Dir, KeyBlock, MBHeight, _LastMBHash) when MBHeight < 0 ->
+    {ok, #{ key_block => KeyBlock, micro_blocks => [], dir => Dir }};
 get_generation(Dir, KeyBlock, MBHeight, LastMBHash) ->
     {ok, MBs} = get_micro_blocks_at_height(MBHeight, LastMBHash),
     {ok, #{ key_block => KeyBlock, micro_blocks => MBs, dir => Dir }}.
--- a/apps/aecore/src/aec_chain.erl
+++ b/apps/aecore/src/aec_chain.erl
@@ -723,76 +723,67 @@ get_key_block_by_height(Height) when is_integer(Height), Height >= 0 ->
 
 -spec get_current_generation() -> 'error' | {'ok', generation()}.
 get_current_generation() ->
-    get_generation_(top_block_node()).
-
-get_generation_(#{hash := Hash, header := Header}) ->
-    case aec_headers:type(Header) of
-        key ->
-            {ok, #{ key_block => aec_blocks:new_key_from_header(Header), micro_blocks => [], dir => forward }};
-        micro ->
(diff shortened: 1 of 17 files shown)


hanssv:
The tests running on OTP-24 run slower than the others, hence the repeated test issues there. What happens is that the test runs behind. The nodes are driven by two parameters, the schedule, and the tests commanding it to start/stop mining. Thus, when the test falls behind it tries to stick to the schedule, despite not yet being told to go, and it computes the next block, but won't release it until the test says `start mining`. This results in the test missing its intended slot with the submitted Tx, and the nodes will produce a key-block when the test expects a micro-block.

apps/aecore/src/aec_chain.erl
```diff
@@ -723,76 +723,67 @@ get_key_block_by_height(Height) when is_integer(Height), Height >= 0 ->
...
     case aec_db:find_key_block(KeyBlockHash) of
         none -> error;
-        {value, KeyBlock} when Dir == backward ->
-            case get_generation_(aec_blocks:prev_hash(KeyBlock)) of
-                error         -> error;
-                {ok, G = #{}} -> {ok, G#{ key_block => KeyBlock, dir => Dir }}
-            end;
-        {value, KeyBlock} when Dir == forward ->
-            case get_generation_by_height(aec_blocks:height(KeyBlock), forward) of
-                {ok, G = #{ key_block := KeyBlock }} -> {ok, G};
```
ThomasArts, line 769: I have a hard time following where this logic has gone and why it was here at the first place.

I assume that somehow this would kick in if you are on a fork, get a hash of a key block, then start searching, but the generation you get returned does no longer contain the keyblock you started of from, due to the fork??

apps/aecore/src/aec_conductor.erl
```diff
@@ -1330,6 +1340,42 @@ handle_key_block_candidate_reply({{error, Reason}, _}, State) ->
...
+                    TopHash;
+                [_ | _ ] ->
+                    SignModule = ConsensusModule:get_sign_module(),
+                    {ok, SignedMBlock} = SignModule:sign_micro_block(MBlock, Leader),
+                    add_signed_block(SignedMBlock),
+                    {ok, MBHash} = aec_blocks:hash_internal_representation(SignedMBlock),
+                    epoch_mining:debug("New micro-block added, top is: ~p", [MBHash]),
+                    MBHash
+            end;
+        _ ->
```
ThomasArts, line 1374: Tracing the code a bit and it seems this can only happen in case where something is actually wrong. Possibly log an info here instead of debug.

apps/aecore/src/aec_conductor.erl
```diff
@@ -1330,6 +1340,42 @@ handle_key_block_candidate_reply({{error, Reason}, _}, State) ->
...
+            epoch_mining:debug("Failed micro-block, top is still: ~p", [TopHash]),
```
ThomasArts, line 1375:
```suggestion
            epoch_mining:info("Failed micro-block, top is still: ~p", [TopHash]),
```

apps/aecore/src/aec_peer_connection.erl
```diff
@@ -1692,12 +1696,16 @@ validate_prev_key_block(_MicroHeader, not_found, _PrevKeyHeader) ->
...
 validate_micro_signature(MicroHeader, _PrevHeader, PrevKeyHeader)
   when PrevKeyHeader =/= not_found ->
-    Bin = aec_headers:serialize_to_signature_binary(MicroHeader),
-    Sig = aec_headers:signature(MicroHeader),
-    Signer = aec_headers:miner(PrevKeyHeader),
-    case enacl:sign_verify_detached(Sig, Bin, Signer) of
-        true  -> ok;
-        false -> {error, signature_verification_failed}
+    case aec_headers:consensus_module(MicroHeader) of
+        aec_consensus_hc -> ok;
```
ThomasArts, line 1700: Here I would like a little comment on why we don't need to validate the signature here.
I guess this is because the leader of the epoch is known and all validators will reject it if the leader has not signed... but where is that done?

ThomasArts (approved):
Some smaller remarks.

apps/aecore/src/aec_chain.erl
hanssv: This logic has not gone anywhere - I think it is an artefact of the convoluted way this function was implemented before. I.e. to get the generation _forward_ from a key-block(hash) you basically grabbed the _next_ key-block and then fetched a generation backwards from that. 

In effect that makes this a check that the keyblock-hash is on the current chain. Not sure if that is intended or not, but it is a good observation, I think we can keep the old behavior - I'll add an extra check.

apps/aecore/src/aec_conductor.erl
hanssv: Makes sense!

hanssv:
Tests passed, after some tweaks, but I am not convinced this is 100% correct...

mitchelli (approved):

happi (approved):

PR #4437, opened 2024-10-11
Review and update uses of time functions in the system.

richcarl:
Use modern time API everywhere. Do not do time conversions manually when there are standard functions available. Don't use the deprecated time units. Prefer to use monotonic time for plain time delta measurements.

### DIFF
--- a/apps/aecore/src/aec_sync_stats.erl
+++ b/apps/aecore/src/aec_sync_stats.erl
@@ -161,7 +161,7 @@ set_sync_stat(Id, SyncStat, State) ->
     lists:keystore(Id, #sync_stat.id, State, touch(SyncStat)).
 
 touch(SyncStat) ->
-    SyncStat#sync_stat{ last_update = now_ms() }.
+    SyncStat#sync_stat{ last_update = erlang:system_time(millisecond) }.
 
 %% -- Stats and aggregation -----------------------------------------------
 empty_agg(Step, Limit) ->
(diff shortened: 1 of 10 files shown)


system_test/common/aest_sync_SUITE.erl
```diff
@@ -210,23 +210,23 @@ new_node_joins_network(Cfg) ->
     %% Starts a chain with two nodes
     start_node(old_node1, Cfg),
     start_node(old_node2, Cfg),
-    T0 = erlang:system_time(seconds),
```
hanssv, line 213: Ouch...

Ah, wait, no, `seconds` is just deprecated, not wrong! Then this is just a good update.

hanssv (approved):

apps/aecore/src/aec_sync_stats.erl
```diff
@@ -161,7 +161,7 @@ set_sync_stat(Id, SyncStat, State) ->
     lists:keystore(Id, #sync_stat.id, State, touch(SyncStat)).
 
 touch(SyncStat) ->
-    SyncStat#sync_stat{ last_update = now_ms() }.
+    SyncStat#sync_stat{ last_update = erlang:system_time(millisecond) }.
```
ThomasArts, line 164: Why not monotonic here?

ThomasArts (approved):

apps/aecore/src/aec_sync_stats.erl
richcarl: I just inlined a single use utility function here. Changing the api call could change the meaning of  the 'touch' function, and I don't want to do that unless I know exactly how it's used.

hanssv: I am pretty sure it is only used for (internal) statistics so monotonic or not makes little difference.

PR #4440, opened 2024-10-15
Add building with dynamic librocksdb to documentation

richcarl:


dincho (changes requested):
please move the addition to the ubuntu section as the proposition is ubuntu specific, even better if all the supported OS's are covered

dincho (approved):

hanssv (approved):

PR #4441, opened 2024-10-15
First (semi) complete pin validation

klercker:
First "complete" code of pin validation (currently mostly implemented in hyperchain test suite):
1. write pin data to PC -> get TX hash back -> "send to last leader in CC spend tx
2. Last leader collects pin spends (from 1) -> selects one
3. Last leader calls election contract endpoint with txhash + epoch
4. Once this is on chain, it is picked up, used to find pin info on PC, which is validated against blockhash of (by that time) last block of current epoch-2.

New pin test case in `aehttp_hyperchains_SUITE.erl` executes this complete thing (with help from a few other classes), and validates the validation.

- refactored from test case to library (at least some)
- refactored/moved into logical modules (`aec_parent_connector.erl` mostly, or `aec_chains_hc.erl`), the `aec_pinning_agent.erl` has been removed.


klercker:
@ThomasAerts @hanssv After rebasing against the key-micro reverse merge, the contract seem to fail ever picking up the stateful call tx I make to make note of the pin. Previously, It took a block or two for the state update to get on chain (i.e a dry-run call at top for contract state would reflect the call tx) but it eventually did. I was expecting it to actually be on-chain earlier (immediately after the micro->key block was agreed on, before the next microblock is started?), alas no.

klercker:
pushed rebase that has the above mentioned test failure (CCi will therefore fail miserably)

klercker:
It should be noted that the pinning rewards code is to be considered more of proof of concept at this point. It could even be argued that it should be dropped for this PR, but I figure we can keep it in and replace it in the next round of PRs for pinning (which seem to be what I get into next?). It's not doing any harm from what I can see?

apps/aecore/src/aec_parent_connector.erl
```diff
@@ -326,3 +363,103 @@ get_sign_module() ->
...
+            lists:seq(FromHeight, ToHeight)),
+    lists:reverse(ReversedBlocks).
+
+
+find_spends_to(Account) ->
+    %% TODO Not happy with this. are we in the correct epoch when this is supposed to be 
+    %%      called, and what if you want to go back through the chain and validate backwards?
+    {ok, #{last := Last, first := First}} = aec_chain_hc:epoch_info(), 
+    Blocks = get_generations(First, Last-1),
+    lists:flatten([ pick_pin_spends_to(Account, Txs) || {mic_block, _, Txs, _} <- Blocks ]).
```
hanssv, line 438: Don't use the internal structures like `mic_block` there are accessor functions in `aec_blocks`

apps/aecore/src/aec_parent_connector.erl
```diff
@@ -326,3 +363,103 @@ get_sign_module() ->
...
+
+pick_pin_spends_to(Account, Txs) ->
+    [ T || {signed_tx,{aetx,spend_tx,aec_spend_tx,_,{spend_tx,_,{id,account,Account2},_,_,_,_,T}},_} <- Txs, Account =:= Account2, is_pin(T)].
```
hanssv, line 441: Again internal data structure horribly used - please refer to `aetx` and `aec_spend_tx`

apps/aehttp/src/aehttpc_aeternity.erl
```diff
@@ -94,20 +103,79 @@ get_generation_by_height(NodeSpec, Height) ->
...
+        Epoch = erlang:list_to_integer(binary_to_list(HexEpoch), 16),
+        Height = erlang:list_to_integer(binary_to_list(HexHeight), 16),
+        {ok, BlockHash} = aeser_api_encoder:safe_decode(key_block_hash, EncBlockHash),
+        #{epoch => Epoch, height => Height, block_hash => BlockHash}
+    catch 
+        _ -> {error, {bad_parent_pin_payload, Binary}}
+    end.
+
+encode_child_pin_payload(TxHash) ->
+    <<$p,$i,$n, TxHash/binary>>.
```
hanssv, line 130:
```suggestion
    <<"pin", TxHash/binary>>.
```

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -833,6 +845,36 @@ is_leader_valid(Node, _Trees, TxEnv, _PrevNode) ->
...
+                end
+            catch
+                Type:Err -> 
+                    lager:debug("bad pin proof posted: ~p : ~p", [Type, Err]),
+                    pin_validation_fail               
+            end
+    end.
+    
+add_pin_reward(Trees, Leader) ->
+    lager:debug("PINNING: correct pin in current Epoch. Rewarding 4711"),
```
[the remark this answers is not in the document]
klercker: Exactly, that's the next thing after this PR.

apps/aecore/src/aec_parent_connector.erl
```diff
@@ -326,3 +363,103 @@ get_sign_module() ->
...
+                case aec_chain:get_generation_by_height(Height, backward) of
+                    {ok, #{micro_blocks := MBs}} ->
+                        ReversedGeneration = lists:reverse(MBs),
+                        ReversedGeneration ++ Accum;
+                    error -> error({failed_to_fetch_generation, Height})
+                end
+            end,
+            [],
+            lists:seq(FromHeight, ToHeight)),
+    lists:reverse(ReversedBlocks).
```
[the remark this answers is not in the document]
klercker: I stole it from the test case :) Suggestions are welcome.

apps/aecore/src/aec_parent_connector.erl
klercker: Agreed. Can you point me to some place where this kind of thing is done in a better way, and I can mimic that style? These APIs are (at least not for the casual viewer) beacons of clarity and docs...

apps/aecore/src/aec_parent_connector.erl
klercker: See above...

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -274,9 +274,21 @@ state_pre_transform_node(Type, Height, PrevNode, Trees) ->
         key ->
            case Height == maps:get(last, EpochInfo) of
                true ->
+                    Trees1 =
```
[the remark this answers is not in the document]
klercker: Again, I have nothing against creating well indented code, I just want to understand the rules of that game...

apps/aehttp/src/aehttpc_aeternity.erl
klercker: Fixed. In the commit queue.

apps/aecore/src/aec_consensus_hc.erl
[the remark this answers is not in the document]
hanssv: `Trees1` and the rest is indented 5 spaces further than `true` - it just hurts my eyes 🤷‍♂️

apps/aecore/src/aec_parent_connector.erl
```diff
@@ -326,3 +363,103 @@ get_sign_module() ->
...
+    gen_server:call(?SERVER, {encode_parent_pin_payload, Pin}).
+
+decode_parent_pin_payload(PinPayload) ->
+    gen_server:call(?SERVER, {decode_parent_pin_payload, PinPayload}).
+
+encode_child_pin_payload(TxHash) ->
+    gen_server:call(?SERVER, {encode_child_pin_payload, TxHash}).
+
+decode_child_pin_payload(TxHash) ->
+    gen_server:call(?SERVER, {decode_child_pin_payload, TxHash}).
```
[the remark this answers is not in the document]
hanssv: This module implements a `gen_server` there are templates for those, and there are huge headers like
```
%%% API
```

Code is normally organized accordingly.

apps/aecore/src/aec_parent_connector.erl
hanssv: Here you would need `aec_blocks:type/1` to do the filter and then `aec_blocks:txs/1` to access the transactions.

apps/aecore/src/aec_parent_connector.erl
hanssv: `aetx_sign:tx/1`, `aetx:specialize_type/1`, `aec_spend_tx:recipient_id/1` and `aeser_id:specialize/1` should be what you need.

apps/aehttp/src/aehttpc_aeternity.erl
hanssv: Possibly you need `"pin"/utf8` but otherwise it should work...

apps/aecore/src/aec_consensus_hc.erl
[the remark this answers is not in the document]
klercker: Should it be 2? 4? None? Happy to oblige :)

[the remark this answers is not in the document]
hanssv: 4
```
    true ->
        Trees1 =
            ...
```

apps/aecore/src/aec_parent_connector.erl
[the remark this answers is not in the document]
klercker: New bad-ass impl in incoming commit

apps/aecore/src/aec_consensus_hc.erl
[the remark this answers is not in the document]
klercker: Fixed. Although, I am bit disappointed that you didn't spot the ACTUAL indentation error: the lines immediately outside of the true-clause were incorrectly indented one space too much! :P

apps/aecore/src/aec_parent_connector.erl
```diff
@@ -415,30 +420,30 @@ get_pinning_data() ->
...
+    lists:flatten(
+        lists:map(
+        fun(Height) ->
+            case aec_chain:get_generation_by_height(Height, backward) of
+                {ok, #{micro_blocks := MBs}} -> MBs;
+                error -> error({failed_to_fetch_generation, Height})
+            end
+        end,
+        lists:seq(FromHeight, ToHeight)
+    )).
```
hanssv, line 429: Won't this make exactly as many calls to `get_generation_by_height` as the previous version?

apps/aecore/src/aec_parent_connector.erl
```diff
@@ -415,30 +420,30 @@ get_pinning_data() ->
     end.
 
 get_generations(FromHeight, ToHeight) ->
```
hanssv, line 419: I think this function should be called `get_microblocks_between_heights` - a generation contains both keyblocks and microblocks 🤔

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -289,11 +289,11 @@ state_pre_transform_node(Type, Height, PrevNode, Trees) ->
                    {ok, Seed} = get_entropy_hash(Epoch + 2),
                    cache_validators_for_epoch({TxEnv, Trees}, Seed, Epoch + 2),
                    step_eoe(TxEnv, Trees1, Leader, Seed, 0);
```
hanssv: Sorry, I think now these three lines are misaligned (they were correct relative to `true` _before_ that was fixed).

apps/aecore/src/aec_parent_connector.erl
```diff
@@ -415,30 +420,30 @@ get_pinning_data() ->
...
     Blocks = get_generations(First, Last-1),
-    lists:flatten([ pick_pin_spends_to(Account, Txs) || {mic_block, _, Txs, _} <- Blocks ]).
+    lists:flatten([ pick_pin_spends_to(Account, aec_blocks:txs(B)) || B <- Blocks ]).
+
 
 pick_pin_spends_to(Account, Txs) ->
-    [ T || {signed_tx,{aetx,spend_tx,aec_spend_tx,_,{spend_tx,_,{id,account,Account2},_,_,_,_,T}},_} <- Txs, Account =:= Account2, is_pin(T)].
+    BareTxs = [ aetx_sign:tx(T) || T <- Txs],
+    InnerTxs = [ aetx:specialize_type(Tx) || Tx <- BareTxs,  spend_tx == aetx:tx_type(Tx)],
+    [ aec_spend_tx:payload(T) || {_, T} <- InnerTxs, aeser_id:specialize(aec_spend_tx:recipient_id(T)) == {account,Account}, is_pin(aec_spend_tx:payload(T))].
```
hanssv, line 443: Looks much better! 👍  
Alternatively the first to could perhaps be combined into:
```
SpendTxs = [ Tx || STx <- Txs, {spend_tx, Tx} <- [aetx:specialize_type(STx)] ],
```

apps/aecore/src/aec_parent_connector.erl
hanssv: Yes it will...

You need something like (not tested):
```
get_micro_blocks_between(From, To) ->
    {ok, KeyHdr} = aec_chain:get_key_header_by_height(To),
    get_micro_blocks_between(From, KeyHdr, []).

get_micro_blocks_between(Stop, KeyHdr, MBs) ->
    case Stop > aec_headers:height(KeyHdr) of
        true -> lists:reverse(MBs);
        false ->
            {ok, PrevKeyHdr} = aec_chain:get_header(aec_headers:prev_key_hash(KeyHdr)),
            case aec_headers:prev_hash(KeyHdr) == aec_headers:prev_key_hash(KeyHdr) of
                true -> %% No Mb
                    get_micro_blocks_between(Stop, PrevKeyHdr, MBs);
                false ->
                    MB = aec_chain:get_block(aec_header:prev_hash(KeyHdr)),
                    get_micro_blocks_between(Stop, PrevKeyHdr, [MB | MBs])
            end
    end.
```

hanssv (commented):
Getting closer 👍

apps/aecore/src/aec_parent_connector.erl
klercker: TBH, this should go into `aec_chain` instead.

apps/aecore/src/aec_parent_connector.erl
klercker: Yea, but double generators make my eyes hurt.

test/contracts/HCElection.aes
```diff
@@ -101,6 +105,14 @@ main contract HCElection =
     // One extra hash operation to convert from bytes() to bytes(32)/hash
     validator_schedule_(Crypto.blake2b(seed), (s) => Bytes.to_int(s) mod total_stake, validators, length, [])
 
+  stateful entrypoint pin(s : string) =
+    // PINTODO: determine that caller is "correct"
```
ThomasArts, line 109: OK! This is solved when PR#4451 is merged

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -833,6 +845,36 @@ is_leader_valid(Node, _Trees, TxEnv, _PrevNode) ->
...
+    case aec_chain_hc:pin_info({TxEnv, Trees}) of
+        undefined -> pin_missing;
+        EncTxHash ->  
+            % TODO make this code much more robust - incorrect EncTxHash, bad value from PC, incorrect hash etc.etc
+            lager:debug("PINNING: EncHash: ~p", [EncTxHash]),
+            try 
+                #{epoch := _PinEpoch, height := PinHeight, block_hash := PinHash} = 
+                    aec_parent_connector:get_pin_by_tx_hash(EncTxHash),
+                case {ok, PinHash} =:= aec_chain_state:get_key_block_hash_at_height(PinHeight) of 
+                    true -> pin_correct;
```
ThomasArts, line 858: 1. You have a TxHash on parent chain
2. You get the transaction from parent chain (via connector and cache, hence finite)
3. The transaction on parent chain has the right format and contains a PinHeight and PinHash.
4. you check that that hash is for that height... BUT
**also to check that that is the last height of the previous epoch!**
   for example use  `epoch_info` on current height and verify that `epoch` in there is PinEpoch+1, that  `first` is `PinHeight`-1 or get the previous epoch.

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -833,6 +845,36 @@ is_leader_valid(Node, _Trees, TxEnv, _PrevNode) ->
...
+                    lager:debug("bad pin proof posted: ~p : ~p", [Type, Err]),
+                    pin_validation_fail               
+            end
+    end.
+    
+add_pin_reward(Trees, Leader) ->
+    lager:debug("PINNING: correct pin in current Epoch. Rewarding 4711"),
+    aec_events:publish(pin, {pin_accepted}),
+    LeaderAcc = aec_accounts_trees:get(Leader, aec_trees:accounts(Trees)),
+    Reward = 4711,
```
ThomasArts, line 872:
```suggestion
    Reward = 4711,   %% TODO Fix the rewards
```

apps/aecore/src/aec_parent_connector.erl
```diff
@@ -314,15 +386,82 @@ handle_fetch_block(Fun, Arg,
...
+                {ok, #{micro_blocks := MBs}} -> MBs;
+                error -> error({failed_to_fetch_generation, Height})
+            end
+        end,
+        lists:seq(FromHeight, ToHeight)
+    )).
+
+
+find_spends_to(Account) ->
+    %% TODO Not happy with this. are we in the correct epoch when this is supposed to be
```
ThomasArts, line 433: We never need to validate this. The transaction is just there for leader convenience has nothing to do with consensus

apps/aehttp/test/aehttp_hyperchains_SUITE.erl
```diff
@@ -1080,11 +1080,201 @@ wallet_post_pin_to_pc(Config) ->
...
+    AccPKEncEnc = aeser_api_encoder:encode(account_pubkey, AccountPK),
+    {ok, []} = rpc(?PARENT_CHAIN_NODE, aec_tx_pool, peek, [infinity]), % no pending transactions
+    PinTx = rpc(Node, aec_parent_connector, create_pin_tx, [AccPKEncEnc, AccountPK, 1, 30000 * ?DEFAULT_GAS_PRICE, PinningData]),
+    SignedPinTx = sign_tx(PinTx, privkey(?DWIGHT),?PARENT_CHAIN_NETWORK_ID),
+    rpc(Node, aec_parent_connector, post_pin_tx, [SignedPinTx]).
+
+% PINREFAC
+tx_hash_to_child(Node, EncTxHash, SendAccount, Leader, Config) ->
+    NodeName = aecore_suite_utils:node_name(Node),
+    NetworkId = ?config(network_id, Config), % TODO not 100% sure about this one...
```
ThomasArts, line 1258:
```suggestion
    NetworkId = ?config(network_id, Config),
```

apps/aehttp/test/aehttp_hyperchains_SUITE.erl
```diff
@@ -1080,11 +1080,201 @@ wallet_post_pin_to_pc(Config) ->
...
+    SignedTx = sign_tx(Tx, privkey(SendAccount), NetworkId),
+    ok = rpc:call(NodeName, aec_tx_pool, push, [SignedTx, tx_received]),
+    Hash = rpc:call(NodeName, aetx_sign, hash, [SignedTx]),
+    Hash.
+
+mine_to_next_epoch(Node, Config) ->
+    Height1 = rpc(Node, aec_chain, top_height, []),
+    {ok, #{last := Last1, length := _Len}} = rpc(Node, aec_chain_hc, epoch_info, []),
+    {ok, Bs} = produce_cc_blocks(Config, Last1 - Height1 + 1), 
+    ct:log("Block last epoch: ~p", [hd(lists:nthtail(length(Bs) - 2, Bs))]).
```
ThomasArts, line 1277: This does not work if you already are at last block of epoch. It will crash
Just do `ct:log` of all `Bs` to avoid that.

ThomasArts (commented):
some comments, most important to actuall validate the pin data

apps/aecore/src/aec_parent_connector.erl
klercker: Correct. If we feel the need to correct this later, we'll fix it then.

apps/aecore/src/aec_consensus_hc.erl
klercker: Testing against the current epoch (PinEpoch = CurEpoch - 1).

apps/aehttp/test/aehttp_hyperchains_SUITE.erl
klercker: Not sure I get the ct:log part of this?

klercker: You mean we dig too deep in a list which imght be only 1 long? Is `ct:log("Block last epoch: ~p", [BS]).`your suggestion?

ThomasArts: I mean that if Height1 == Last1 then you only get 1 bkock and then your nth tail crashes... Which you don't want

ThomasArts: So, yes, your line above is a better solution

klercker: Will go with that.

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -845,15 +845,17 @@ is_leader_valid(Node, _Trees, TxEnv, _PrevNode) ->
...
         undefined -> pin_missing;
         EncTxHash ->  
             % TODO make this code much more robust - incorrect EncTxHash, bad value from PC, incorrect hash etc.etc
             lager:debug("PINNING: EncHash: ~p", [EncTxHash]),
             try 
-                #{epoch := _PinEpoch, height := PinHeight, block_hash := PinHash} = 
+                #{epoch := CurEpoch} = CurEpochInfo,
+                #{epoch := PinEpoch, height := PinHeight, block_hash := PinHash} = 
                     aec_parent_connector:get_pin_by_tx_hash(EncTxHash),
+                PinEpoch = CurEpoch - 1, % validate it was actually last epoch
```
hanssv, line 858: This will crash if it isn't true - someone complained that crashes in consesus-module (called by worker in conductor) was a bit opaque, perhaps use `aec_conductor:throw_error/1`?

apps/aecore/src/aec_parent_connector.erl
```diff
@@ -416,24 +416,11 @@ get_pinning_data() ->
...
-        lists:seq(FromHeight, ToHeight)
-    )).
-
-
 find_spends_to(Account) ->
     %% TODO Not happy with this. are we in the correct epoch when this is supposed to be 
     %%      called, and what if you want to go back through the chain and validate backwards?
     {ok, #{last := Last, first := First}} = aec_chain_hc:epoch_info(), 
-    Blocks = get_generations(First, Last-1),
+    Blocks = aec_chain_hc:get_micro_blocks_between(First, Last-1), %get_generations(First, Last-1),
```
hanssv, line 423: superfluous comment

apps/aecore/src/aec_consensus_hc.erl
klercker: I believe this will be caught by the (very wide) `catch` on the outside, resulting in an `pin_validation_fail`?

apps/aecore/src/aec_parent_connector.erl
klercker: Will fix that when I get into rewards.

klercker: Since I had a test case to check in, this made it as well.

apps/aecore/src/aec_parent_connector.erl
klercker: Another convenience is that these spends enable the middleware to check for pins posted (and possibly validate them as well) to display in timelines etc.

ThomasArts (approved):

hanssv (approved):

PR #4443, opened 2024-10-21
- hyperchain/contract-pubkeys endpoint

klercker:
- endpoint implemented
- method to get contract pubkeys from HC consensus module (which also considers the staking contract - change from before
- updated HC test suite to configure all contracts (rewards == staking)

- ATM endpoint fails completely if any of the contracts are not configured. The consensus modeul probably shouldn't even start if this is the case. Right now this is actually due to the ETS cache failing if the contracts aren't there, so there needs to be some fixing around there to handle this in a good way if allowed.

This PR is supported by the Aeternity foundation

### DIFF
--- a/apps/aecore/src/aec_consensus_hc.erl
+++ b/apps/aecore/src/aec_consensus_hc.erl
@@ -483,7 +483,7 @@ staking_contract_pubkey() ->
                   _ -> erlang:error({contract_owner_not_valid_contract, EncContractId})
               end
           end,
-    aeu_ets_cache:get(?ETS_CACHE_TABLE, rewards_contract_pubkey, Fun).
+    aeu_ets_cache:get(?ETS_CACHE_TABLE, staking_contract_pubkey, Fun).
 
 pc_start_height() ->
     Fun = fun() -> get_consensus_config_key([<<"parent_chain">>, <<"start_height">>], 0) end,
--- a/apps/aecore/src/aec_consensus_hc.erl
+++ b/apps/aecore/src/aec_consensus_hc.erl
@@ -83,6 +83,7 @@
         , call_consensus_contract_result/5
         , entropy_height/1
         , get_entropy_hash/1
+        , get_contract_pubkey/1
         ]).
 
 -ifdef(TEST).
@@ -474,6 +475,16 @@ rewards_contract_pubkey() ->
           end,
     aeu_ets_cache:get(?ETS_CACHE_TABLE, rewards_contract_pubkey, Fun).
 
+staking_contract_pubkey() ->
+    Fun = fun() ->
+              EncContractId = get_consensus_config_key([<<"staking_contract">>]),
+              case aeser_api_encoder:safe_decode(contract_pubkey, EncContractId) of
+                  {ok, Pubkey} -> Pubkey;
+                  _ -> erlang:error({contract_owner_not_valid_contract, EncContractId})
+              end
+          end,
+    aeu_ets_cache:get(?ETS_CACHE_TABLE, rewards_contract_pubkey, Fun).
+
 pc_start_height() ->
     Fun = fun() -> get_consensus_config_key([<<"parent_chain">>, <<"start_height">>], 0) end,
     aeu_ets_cache:get(?ETS_CACHE_TABLE, pc_start_height, Fun).
@@ -697,14 +708,17 @@ call_consensus_contract_result(ContractType, TxEnv, Trees, ContractFun, Args) ->
     lager:debug("Call result ~p", [Result]),
     {ok, Result}.
 
+-spec get_contract_pubkey(election | rewards | staking) -> binary().
+get_contract_pubkey(ContractType) ->
+    case ContractType of
+        ?ELECTION_CONTRACT -> election_contract_pubkey();
+        ?REWARDS_CONTRACT -> rewards_contract_pubkey();
+        ?STAKING_CONTRACT -> staking_contract_pubkey()
+    end.
+
 call_consensus_contract_(ContractType, TxEnv, Trees, EncodedCallData, Keyword, Amount) ->
     log_consensus_call(TxEnv, Keyword, EncodedCallData, Amount),
-    ContractPubkey =
-        case ContractType of
-            ?ELECTION_CONTRACT -> election_contract_pubkey();
-            ?REWARDS_CONTRACT -> rewards_contract_pubkey();
-            ?STAKING_CONTRACT -> rewards_contract_pubkey()
-        end,
+    ContractPubkey = get_contract_pubkey(ContractType),
     OwnerPubkey = contract_owner(),
     Contract = aect_state_tree:get_contract(ContractPubkey,
                                             aec_trees:contracts(Trees)),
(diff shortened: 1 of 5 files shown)


apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -474,6 +475,16 @@ rewards_contract_pubkey() ->
...
 
+staking_contract_pubkey() ->
+    Fun = fun() ->
+              EncContractId = get_consensus_config_key([<<"staking_contract">>]),
+              case aeser_api_encoder:safe_decode(contract_pubkey, EncContractId) of
+                  {ok, Pubkey} -> Pubkey;
+                  _ -> erlang:error({contract_owner_not_valid_contract, EncContractId})
+              end
+          end,
+    aeu_ets_cache:get(?ETS_CACHE_TABLE, rewards_contract_pubkey, Fun).
```
ThomasArts, line 486: Why here the rewards contract pubkey as default???

apps/aehttp/src/aehttp_dispatch_ext.erl
```diff
@@ -909,6 +910,14 @@ handle_request_('GetPinningTx', _Params, _Context) ->
            {404, [], #{reason => <<"No pin data available">>}}
     end;
 
+handle_request_('GetHyperchainContractPubkeys', _Params, _Context) ->
+    % TODO handle not in HC at all or consensus not initialized?
+    lager:debug("HyperchainsGetContract"),
+    {200, [], #{<<"staking">> => aeser_api_encoder:encode(contract_pubkey, aec_consensus_hc:get_contract_pubkey(staking)),
```
ThomasArts, line 916: Here we spill over the logic on how many contracts there are and what they are named from consensus to the API. The names  `election`, `rewards` are macros in the consensus code, but now hard-coded in the API.

Not sure about this, but my feeling is that it would be cleaner to ask the consensus module for all contracts and have that module return this map. In that way, the logic of how many and what contracts we have can be adapted rather easily.... but the price is that we cannot say in the API specification what we expect back other than an object with keys and contract addresses.

ThomasArts (approved):
Might consider not hard-coding naming of contracts. On the other hand, if SDK is to use this, they need to know a bit on what kind of contracts they can expect.

ThomasArts:
Failure on otp26 is a state channel test. Seems to me unrelated to code delivered. Please have a look and run again.

ThomasArts (commented):
- [ ] Look at reward_contract as default!
- [ ] rebase on master

apps/aecore/src/aec_consensus_hc.erl
klercker: CopyPaste error, will fix. ran since rewards and staking are the same in the current setup.

apps/aehttp/src/aehttp_dispatch_ext.erl
klercker: Sure, can do that no prob. Not sure how the endpoint will feel about getting a map with unspeced properties though. What's the yaml construct for that?

klercker:
Fixed the cache key and rebased. I say we merge, and if we decide to go with the less hardcoded route (map of pubkeys) we can implement that then? But you guys decide.

apps/aehttp/test/aehttp_hyperchains_SUITE.erl
```diff
@@ -818,6 +820,33 @@ epochs_with_fast_parent(Config) ->
...
+get_contract_pubkeys(Config) ->
+    [{Node, _, _} | _] = ?config(nodes, Config),
+    %% Verify that endpoint is available
+    {ok, IsChildChain} = rpc(Node, aeu_env, find_config,
+                             [[<<"http">>, <<"endpoints">>, <<"hyperchain">>], [user_config, schema_default]]),
+    ?assert(IsChildChain),
+    StakingContractPK = rpc(Node, aec_consensus_hc, get_contract_pubkey, [staking]),
+    ElectionContractPK = rpc(Node, aec_consensus_hc, get_contract_pubkey, [election]),
+    RewardsContractPK = rpc(Node, aec_consensus_hc, get_contract_pubkey, [rewards]),
+    %% note: the pins are for the last block in previous epoch
```
hanssv, line 836: I'm not entirely sure what this has to do with the code...

klercker:
@hanssv are we having timing issues on OTP 24? A test for block type in `hc.verify_fees` fails on 24 but is ok on 25/26?

apps/aehttp/test/aehttp_hyperchains_SUITE.erl
klercker: I test what I get back from the call is identical to what was in the contract?

klercker: But good Q about that comment. C&P leftovers I presume. Gone now, will be in my next functional commit.

klercker:
@hanssv @ThomasArts any suggestions what we do to get pass this? Have re-run 3 times with no luck. same `hc.verify_fees` block type check (in `produce_cc_to_height`). The failure seems to be the first time blocks are produced in the first epoch. But it's beyond me why it would fail in 24. Is there a way to run ct-latest on OTP24 locally? Can't find it in Makefile at least, but perhaps rebar has options?

hanssv:
> @hanssv @ThomasArts any suggestions what we do to get pass this? Have re-run 3 times with no luck. same `hc.verify_fees` block type check (in `produce_cc_to_height`). The failure seems to be the first time blocks are produced in the first epoch. But it's beyond me why it would fail in 24. Is there a way to run ct-latest on OTP24 locally? Can't find it in Makefile at least, but perhaps rebar has options?

Running on OTP24 should be a breeze with devcontainers and VSCode, right? I've got all OTP versions 22-27 locally and switch between them when I need to. Switching OTP-version can't be done in the Makefile.

I think the test fails because it doesn't get a micro-block when expected, the reason is that the test falls behind, and when the Transaction is submitted the producing node is already past the micro-block producing point. Possible solutions are make the test more robust in this case, or, make block-time longer (thus making it less likely to fall behind, but also slower!).

klercker:
Tests run fully!! If everybody is reasonably happy with the state of things (We are at tip of master, I rebased this morning based on @ThomasArts merge from yesterday), I suggest approving and getting this into the tree now :)

apps/aehttp/test/aehttp_hyperchains_SUITE.erl
hanssv: The comment is still there, not resolved.

apps/aehttp/priv/oas3.yaml
```diff
@@ -535,6 +535,27 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  /hyperchain/contract-pubkeys:
```
hanssv, line 538: Why the obsession with the term 'pubkey' here? What else would the endpoint return (the bytecode?) I'd name the endpoint `/hyperchain/contracts`?

hanssv (approved):

apps/aehttp/priv/oas3.yaml
klercker: True. Will resolve these two then. Look forward to new CCI run and re-approval...

hanssv (approved):

ThomasArts (approved):

PR #4450, opened 2024-10-23
Cleanup and consistently spell/name HC config related things

hanssv:


dincho (approved):

happi (approved):
Not obvious that 
                                   "description": "The type of the consensus algorithm used at the given height (ex. pow_cuckoo, smart_contract or hyperchain)",

                                    "default": "hyperchain"
           was ok. (even before) 
        
       I guess the  preceding documentation:
        Ignored if 'fork_management > network_id' has value 'ae_mainnet' or 'ae_uat'."
        
        is the key here.

Issue #4448, opened 2024-10-22
Make HCElection contract aware of last leader in epoch


ThomasArts:
We plan to have a contract call on chain in which the PIN proof is stored.
This call should only update the contract state if it is the last leader that calls the contract.

(We don't want others to be able to manipulate the proof provided).

CanutoIX:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

PR #4451, opened 2024-10-23
only last leader in epoch can pin

ThomasArts:
Fixes Issue #4448 in a particular way.

Instead of carrying around who the last leader is in the epochs (which resulted in rather complex code and unsatisfactory solution for first two epochs), decided to keep the contract simple and enforce pin call to be added to the last micro block. In that way, it is easy to check who the leader is.

Up to debate when to clear the PIN. If done in `step_eoe` then inspecting the last key block will not show it. Hence one needs to look before that. If done in `step` then extra check needed that this is first block of epoch or reset to NONE in each call (since one cannot set it anyway).

This PR is supported by the Aeternity foundation

### DIFF
--- a/apps/aecore/src/aec_block_micro_candidate.erl
+++ b/apps/aecore/src/aec_block_micro_candidate.erl
@@ -98,11 +98,12 @@ int_create(PrevBlock, KeyBlock) ->
     MBEnv = #{prev_hash := PrevBlockHash} = create_micro_block_env(PrevBlock, KeyBlock),
     case aec_chain:get_block_state(PrevBlockHash) of
         {ok, Trees} ->
-            TxEnv = create_tx_env(MBEnv),
+            TxEnv0 = create_tx_env(MBEnv),
             MaxGas = aec_governance:block_gas_limit(),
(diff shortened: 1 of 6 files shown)


ThomasArts:
After some discussion with @hanssv we realized that this won't work, because in the micro block before a key block, the leader is not updated... Leaders are updated in `step_eoe` when the key block is produced.
This is a remainder of having micro blocks after key blocks. However, it makes sense to actually fix that, because other contracts may also want to do things relative to current leader.

Opened issue #4453 to get a solution on that.
Alternative is the the contract logic is changed and that it from start of epoch stores in its state the last leader. That is less elegant and does not solve any future issue with other than the last block in an epoch.

ThomasArts:
Still need to add a test

ThomasArts (commented):
Looked at it, but as initiator I cannot approve

happi (approved):

klercker (commented):
Sure, I can live with this. Once the pinning agent is up, these "artificial test will likely be removed, or at least changed quite a bit.

klercker (approved):

PR #4452, opened 2024-10-23
Make the call to `init_epochs` explicit in consensus

hanssv:


happi (dismissed):

hanssv:
> missing `child_epoch_length` in the config schema, it might sounds like a not bit deal, but the tool I'm using (hacking) is relying on the schema

Now it is there.

dincho (approved):
thanks

PR #4454, opened 2024-10-24
Hole production + a separate test app for hyperchains tests

kvakvs:
Branch contains hole production logic in conductor/consensus_hc modules. This logic changes height used in calculations to timeslots (from epoch 2) and timeslots are based on time only regardless of chain height.

Tests are organized as a separate app, which can be made test-only in Rebar and does not appear in the final release.

### Production
* Late producing — the late block must be ignored by the other node(s)
  * [x] Created
  * [ ] Passing
* Producing two sequential blocks — the block out of allowed height must be ignored by the other node(s)
  * [x] Created
  * [ ] Passing
### Slashable Offences
* Double voting
  * [x] Created
  * [ ] Passing
* [ ] Ignoring valid votes
* [ ] Ignoring finalize epoch block
* [ ] Ignoring correctly pinned block
### Forks
* [ ] Producing two valid blocks with the same root at the allowed height
* [ ] A network split before EoE
* [ ] Combinatoric explosion (each microblock is a fork, see how they consolidate)
### End of an Epoch
* [ ] A network split in EoE
* (duplicate from slashable) Double vote cast in EoE
* [ ] The leader in EoE is not producing (inactive producer)
* (duplicate from slashable) The vote results were ignored by the leader
* [ ] Ensure that finalizing transaction is created including the votes (what if it isn't?)


hanssv:
Looks like a good list. Note: this is _just_ the tests for the consensus logic, there is a lot of other things that also need further testing, like sync (see #4458 🙈 ), etc.

PR #4461, opened 2024-10-29
More robust caching of validator schedules

hanssv:
Main observation, at the point where we call `step_eoe` we _know_ the seed/hash (and thus in effect the schedule) for next-next Epoch. No point in not caching that schedule right away.

Second issue is bootstrapping after restart, that should be handled by the current code, but there are no tests for it.

This PR also adds a test that puts microblocks in all positions in an epoch.

Finally it aligns `entropy_height` with the Whitepaper.

### DIFF
--- a/apps/aecore/src/aec_consensus_hc.erl
+++ b/apps/aecore/src/aec_consensus_hc.erl
@@ -705,16 +705,25 @@ leader_for_height(Height, RunEnv) ->
     end.
 
 cache_validators_for_epoch(RunEnv, Epoch) ->
-    case aec_chain_hc:epoch_info_for_epoch(RunEnv, Epoch) of
+    case epoch_info_for_epoch(RunEnv, Epoch) of
         {ok, EpochInfo} -> cache_validators_for_epoch_info(RunEnv, EpochInfo);
         _Err            -> error
     end.
 
 cache_validators_for_epoch(RunEnv, Hash, Epoch) ->
-    case aec_chain_hc:epoch_info_for_epoch(RunEnv, Epoch) of
+    case epoch_info_for_epoch(RunEnv, Epoch) of
         {ok, EpochInfo} -> cache_validators_for_epoch_info_(RunEnv, Hash, EpochInfo);
         _Err            -> error
     end.
+
+epoch_info_for_epoch(RunEnv, Epoch) ->
+    try
+        aec_chain_hc:epoch_info_for_epoch(RunEnv, Epoch)
+    catch _:_ ->
+        %% Not sure this is possible...
+        error
+    end.
+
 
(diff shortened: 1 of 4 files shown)


apps/aehttp/test/aehttp_hyperchains_SUITE.erl
```diff
@@ -362,13 +360,6 @@ contract_call(ContractPubkey, Src, Fun, Args, Amount, From, Nonce) ->
     {ok, Tx} = aect_call_tx:new(TxSpec),
     Tx.
 
-%% This test is trivially true, since we already check in
-%% sync when producting the blocks.
-mine_and_sync(Config) ->
```
ThomasArts, line 367: Perfectly fine to remove obsolete test

apps/aehttp/test/aehttp_hyperchains_SUITE.erl
```diff
@@ -473,9 +480,8 @@ respect_schedule(Node, EpochStart, Epoch, TopHeight) ->
 
     ct:log("Checking epoch ~p info: ~p at height ~p", [Epoch, EI, EpochStart]),
 
-    %% We buffer the seed two epochs, entropy height already looks at previous Epoch,
-    %% hence Epoch - 1
-    ParentHeight = rpc(?NODE1, aec_consensus_hc, entropy_height, [Epoch - 1]),
+    %% We buffer the seed two epochs
+    ParentHeight = rpc(?NODE1, aec_consensus_hc, entropy_height, [Epoch]),
```
ThomasArts, line 484: Much nicer now with aligned logic

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -261,10 +262,11 @@ state_pre_transform_key_node(Node, PrevNode, Trees) ->
             {TxEnv0, _} = aetx_env:tx_env_and_trees_from_hash(aetx_transaction, PrevHash),
             TxEnv = aetx_env:set_height(TxEnv0, Height),
             {ok, #{epoch := Epoch} = EpochInfo} = aec_chain_hc:epoch_info({TxEnv, Trees}),
-            {ok, Leader} = safe_leader_for_height(TxEnv, Trees, Height),
+            {ok, Leader} = leader_for_height(Height, {TxEnv, Trees}),
             case Height == maps:get(last, EpochInfo) of
                 true ->
-                    {ok, Seed} = get_entropy_hash(Epoch + 1),
+                    {ok, Seed} = get_entropy_hash(Epoch + 2),
```
happi, line 268: Possibly document the "random" +2 used here ;).

happi (approved):

PR #4465, opened 2024-10-29
Cleanup get state hc election

ThomasArts:

### DIFF
--- a/apps/aecontract/test/aecontract_staking_contract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_staking_contract_SUITE.erl
@@ -66,7 +66,9 @@
 
 -define(OWNER_PUBKEY, <<42:32/unit:8>>).
 
--define(GENESIS_HEIGHT, 5).
+-define(GENESIS_HEIGHT, 0).
+%% Height to start tests after contracts have been initialized
+-define(DEFAULT_HEIGHT, 1).
 
 -define(NETWORK_ID, <<"hc_stake_test">>).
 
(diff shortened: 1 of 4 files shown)


klercker (approved):

hanssv:
@dincho is not relying on this endpoint?

hanssv (commented):
If it is always `?GENESIS_HEIGHT + 1` then maybe _both_ the name _and_ the value of the macro are wrong?

ThomasArts:
The name is then wrong.
Long story short, GENESIS_HEIGHT was 5 before. But contract HCElection is now written in a way that it can only kick in on Height 0 when computing the initial epochs.

So, I use GENESIS_HEIGHT to start the chain and then can only use it from height 1.

happi:
We should devise a (good) name for the "start" block. We have Genesis and then the first "live" block, which is perhaps just block one or the first block (following block 0 ;) ).

ThomasArts:
But these tests do not really depend on running in the start block or first block, they just need a height larger than 0.
I am happy to fiddle with a new name, but I would go for `HEIGHT_ABOVE_GENESIS` or so and define that to be 1.

happi:
It was a bit more philosophiocal idea. For these test I am fine with Genesis + 1. I was thinking of the whitepaper and the rest of the code.

hanssv:
I'm thinking that `GENESIS_HEIGHT` and `HEIGHT` might be good constants in this case.

happi (dismissed):
I approve;  you can merge at your discretion.

hanssv (dismissed):

klercker (approved):
No merge until #4441 is in. PLEEEEEEAASE? 🥹

hanssv (approved):

happi (approved):

PR #4471, opened 2024-11-04
WIP: Adding a way to create a 'Hole' keyblock.

happi:
Playing with the idea of adding a bit-flag to the info field in a key block header for signaling an empty 'Hole' block.

No tests.
Not fully developed.
No safeguards for other parts of bit field spilling over.
No real way of creating holes yet.

I would appreciate any comments on the approach, though.


apps/aecore/src/aec_block_key_candidate.erl
```diff
@@ -7,71 +7,104 @@
...
-    InfoField = aec_chain_state:get_info_field(Height, Fork),
+    InfoField =
+        case maps:get(hc_hole, Flags, false) of
+            true ->
+                %% Keep backwards comatibility for ae
+                %% WIP Note: Not sure if this should live here.
+                case aec_chain_state:get_info_field(Height, Fork) of
+                    default when Protocol  >= ?MINERVA_PROTOCOL_VSN ->
+                    PointReleaseInfo = aeu_info:block_info(),
+                    Info = (1 bsl 31) bor PointReleaseInfo,
```
happi, line 94: I should create a hc_hole_flag function (or create_hc_hole_flag, and is_hc_hole functions) somewhere. (In aec_headers, perhaps).

hanssv:
This and my current work with adding a vote transaction has me starting to really wonder if we don't want to be able to know/detect at compile time if we are building for HC or not...

apps/aecore/src/aec_block_key_candidate.erl
```diff
@@ -7,71 +7,87 @@
 -module(aec_block_key_candidate).
 
 -export([ create/3
+        , hc_create_hole/3
```
hanssv, line 10: We should avoid, if possible, to change the API of this module with HC specific things (unless we do it at compile time).

Look at the solution for `Height` below, make a callback in the consensus-module and take it that route?

apps/aecore/src/aec_conductor.erl
```diff
@@ -1311,17 +1311,36 @@ handle_key_block_candidate_reply({{error, Reason}, _}, State) ->
...
+                    time ->
+                        {Err, TopHash}
+                end;
             {error, _} = Err ->
                 {Err, TopHash}
         end
     end.
 
+hc_create_hole(_ConsensusModule, TopHash0, Leader) ->
+    %% WIP: TODO: Empty microblock also?
```
hanssv, line 1338: No empty microblocks! That is an abomination.

apps/aecore/src/aec_block_key_candidate.erl
happi: Yes, perhaps compile-time ifdefs are the way to go here. 
It would be a step towards an HC fork/branch of the repo.
Having a callback to determine if you are producing a hole feels convoluted.
It would require some extra make setup but perhaps not much more complicated than the config setup we need to build the complete docker.

hanssv: > Having a callback to determine if you are producing a hole feels convoluted.

That is what you get when your fitting the round peg in the square hole...

apps/aecore/include/blocks.hrl
```diff
@@ -5,10 +5,21 @@
...
 -define(KEY_HEADER_TAG, 1).
 -define(MICRO_HEADER_TAG, 0).
 
+%% The flag field is a 32 bit (4 bytes) field.
+%% With the following meaning:
+%% BIT        MEANING
+%% 31         Block Type: 1 - key, 0 - micro (KEY_HEADER_TAG/MICRO_HEADER_TAG)
+%% 30         For key block: ContainsInfo 1 - true, 0 - false
+%%            For micro blocks: PoFFlag: 1 - pof_hash, 0 no pof_hash
+%% 29         HC Hole block
```
hanssv, line 20: Only for key-block, right?

happi: Yes, and only for POS. But it is probably a good idea not to reuse flags if not strictly necessary. We have 29 free flags, but having the flags mean different things depending on other flags in the same field (as flag 30) might lead to confusion.

happi:
Concepts incorporated in other PRs

PR #4472, opened 2024-11-04
Pin Reward Config and Carry-over

klercker:
Implements
https://github.com/aeternity/aeternity/issues/4469

This PR is supported by the Aeternity Foundation

### DIFF
--- a/test/contracts/HCElection.aes
+++ b/test/contracts/HCElection.aes
@@ -9,10 +9,11 @@ contract interface MainStaking =
 
 main contract HCElection =
   record epoch_info =
-    { start                : int,
-      length               : int,
-      seed                 : option(bytes()),
-      staking_distribution : option(list(address * int))
+    { start                 : int,
+      length                : int,
+      seed                  : option(bytes()),
+      staking_distribution  : option(list(address * int)),
+      carry_over_pin_reward : int
     }
 
   record state =
@@ -31,7 +32,7 @@ main contract HCElection =
       epoch                 = 0,
       epochs                = {},
       pin                   = None }   
-
+ 
   stateful entrypoint init_epochs(epoch_length : int) =
     assert_protocol_call()
     require(Chain.block_height == 0, "Only in genesis")
@@ -48,7 +49,7 @@ main contract HCElection =
                          length : int,
                          seed : option(bytes),
                          staking_distribution : option(list(address * int))) =
-    {start = start, length = length, seed = seed, staking_distribution = staking_distribution}
+    {start = start, length = length, seed = seed, staking_distribution = staking_distribution, carry_over_pin_reward = 0}
 
   stateful entrypoint step(leader : address) =
     assert_protocol_call()
@@ -58,7 +59,7 @@ main contract HCElection =
     assert_protocol_call()
     put(state{ leader = leader })
 
-  stateful entrypoint step_eoe(leader : address, seed : bytes(), epoch_adjust : int) =
+  stateful entrypoint step_eoe(leader : address, seed : bytes(), epoch_adjust : int, carry_over : int) =
     assert_protocol_call()
     let epoch = state.epoch
     let ei = state.epochs[epoch]
@@ -66,7 +67,7 @@ main contract HCElection =
     let ei2 = state.epochs[epoch + 2]
     let ei_adjust = state.epochs[epoch + 3]{ length = ei2.length + epoch_adjust }
     let new_epochs = { [epoch] = state.epochs[epoch],
-                       [epoch + 1] = state.epochs[epoch + 1],
+                       [epoch + 1] = state.epochs[epoch + 1]{ carry_over_pin_reward = carry_over },
                        [epoch + 2] = state.epochs[epoch + 2]{ seed = Some(seed) },
                        [epoch + 3] = ei_adjust,
                        [epoch + 4] = mk_epoch_info(ei_adjust.start + ei_adjust.length, ei_adjust.length, None,
--- a/test/contracts/HCElection.aes
+++ b/test/contracts/HCElection.aes
@@ -13,6 +13,8 @@ main contract HCElection =
       length                : int,
       seed                  : option(bytes()),
       staking_distribution  : option(list(address * int)),
+      base_pin_reward       : int,
+      cur_pin_reward        : int,
       carry_over_pin_reward : int
     }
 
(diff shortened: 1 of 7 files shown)


test/contracts/HCElection.aes
```diff
@@ -9,10 +9,13 @@ contract interface MainStaking =
...
-      length               : int,
-      seed                 : option(bytes()),
-      staking_distribution : option(list(address * int))
+    { start                 : int,
+      length                : int,
+      seed                  : option(bytes()),
+      staking_distribution  : option(list(address * int)),
+      base_pin_reward       : int,
+      cur_pin_reward        : int,
+      carry_over_pin_reward : int
```
hanssv, line 18: Do we really need to keep this in epoch info? We don't expect it to change often, right? Won't it be enough to keep a `pin_reward_per_epoch` and a `current_pin_reward` and at the end of epoch we either reset or accumulate 🤔 This looks a little bit like overkill...

klercker: `carry_over` seems to be a given for the epoch, but I can definitely see the case for the other two being in the state. What this way _does_ do is that it keeps the number of calls needed to the contract down: everything is fetched once per validation (or even epoch, although I don't think `epoch_info` is cached). I'm open for suggestions. There is also future voting procedures to take into account. And I guess just because it's called init_epochs doesn't mean we can init other things in the state as well?

klercker: As I said in the comment, my preferred way of doing it would be for it to be dynamically included in the epoch_info call (as it is relevant for the epoch _and_ to keep contract calls down), so pointers to solve that will be muchos appreciar.

hanssv: My point was that you don't seem to have different values in different epochs!? `carry_over` isn't known for next-next epoch, and surely not for next-next-next epoch... At the end of each epoch these numbers needs to be adjusted, but you don't know them several epochs in advance 🤔

klercker: Correct. Again, the main argument for keeping them with the epoch is to simplify getting to epoch related state (which they truly are, it’s just that they can’t be calculated in advance like some of the other ones as you point out). This is useful for both the end-of-epoch validation and also for MDW, who don’t need to assemble this from multiple end-points. 

I don’t have particular love for either, I just want something that is efficient, usable and reasonably elegant. Perhaps @ThomasArts or @happi  can chime in?

klercker: Is _resource_ efficiency (i.e. cost of calling the contract) actually an issue. My main pro about current solution is developer efficiency?

klercker: My suggestion is we leave this current (working, but slightly contentious) implementation in, and if we either see issues with it or just can't stand it given a couple of days I will happily rewrite it?

klercker: Consensus on how to do this:
1. Move pin fields into it's own (record?) fields on the state
2. Create accessors for the pin state (`pin_info`) on the contract and in `aec_chain_ch.erl`
3. change `aec_consensus_ch.erl`code to use the new `pin_info`.

klercker:
pin reward now moved from epoch to state in HCElection.aes (and other relevant places). Also fixed staking contract test.

test/contracts/HCElection.aes
```diff
@@ -125,7 +145,20 @@ main contract HCElection =
     pick_validator(n, (_, stake) :: validators)            = pick_validator(n - stake, validators)
 
   function assert_protocol_call() =
-      require(Call.caller == Contract.creator, "Must be called by the protocol")
+    require(Call.caller == Contract.creator, "Must be called by the protocol")
+
+  function calc_carry_over(carry_over_flag : bool) =
+    let ei = state.pin_reward
```
hanssv, line 151: `ei` is a very weird variable name?

test/contracts/HCElection.aes
```diff
@@ -58,10 +67,17 @@ main contract HCElection =
     assert_protocol_call()
     put(state{ leader = leader })
 
-  stateful entrypoint step_eoe(leader : address, seed : bytes(), epoch_adjust : int) =
+  stateful entrypoint step_eoe(leader : address, seed : bytes(), epoch_adjust : int,
+                               next_base_pin_reward : int, carry_over_flag : bool) =
```
hanssv, line 71: Probably we should use `None/Some(x)` for both adjusting epoch length and base pin reward. But feel free to ignore that in this PR.

test/contracts/HCElection.aes
```diff
@@ -125,7 +145,20 @@ main contract HCElection =
...
+    require(Call.caller == Contract.creator, "Must be called by the protocol")
+
+  function calc_carry_over(carry_over_flag : bool) =
+    let ei = state.pin_reward
+    if (carry_over_flag)
+      ei.base + ei.carry_over
+    else
+      0
+
+  function calc_next_base_reward(next_base_pin_reward : int) =
```
hanssv, line 157: Yes, `-1` isn't super-elegant here...

test/contracts/HCElection.aes
```diff
@@ -125,7 +145,20 @@ main contract HCElection =
     pick_validator(n, (_, stake) :: validators)            = pick_validator(n - stake, validators)
 
   function assert_protocol_call() =
-      require(Call.caller == Contract.creator, "Must be called by the protocol")
+    require(Call.caller == Contract.creator, "Must be called by the protocol")
+
+  function calc_carry_over(carry_over_flag : bool) =
```
hanssv, line 150: You can match in the function header, something like this is possible:
```
function 
  calc_carry_over : bool => int
  calc_carry_over(false) = 0
  calc_carry_over(true) = state.pin_reward.base + state.pin_reward.carry_over
```

hanssv (approved):

test/contracts/HCElection.aes
klercker: No, but people seem to be sensitive about adding extra complexity etc to the code, so now this parameter has two different meanings depending on it's value. I can't sem to figure out how to pass null or similar as a parameter (which is definitely a good thing that you don't).

klercker: ...or perhaps more specifically to the API, as this was actually complexity _added_ to the code.

test/contracts/HCElection.aes
klercker: Will do that. If we feel that this is too egregious later, I'll just redo.

test/contracts/HCElection.aes
klercker: Oh, that is nice. Did not realize that. Will change to that, mostly because it's cool. Don't _really_ know if it adds to the readability of the code (anywhere in any language really).

klercker:
@hanssv thanks for the approval, for some reason I did not see comments and updates on the PR. I will keep things as they are and address your comments in a future commit/PR (as they all are good points).

happi (approved):

PR #4475, opened 2024-11-08
Create HCstakeDelegationLogic.aes

nikita-fuchs:
As per request, here is again the draft PR of the staking logic. A few gas optimisations might be possible here and there. 

- Everything that says "CRITICAL TODO" is signaling something that needs to be taken into account of in the main staking logic.

- Everything that has stub in its name (functions, vars) are supposed to represent their equivalent gathered from the main staking contract and/or the protocol in HC. 

Reasoning: As we do not get changes to FATE before HC launch that would allow for a claim-and-verify pattern (people claim to know how high the rewards they are owed and a smart contract verifies this information), we need to go the way of book-keeping the rewards for each delegator, so they can withdraw them at some point. This is not as scalable as the first approach, so therefore, the number of delegates per delegatee are limited - the exact number is TBD. So as the minimum amount a delegator needs to stake to prevent DoS, which is set to be 3% of the delegatee's staked amount at the time of delegation.


dincho:
Obsolete

PR #4477, opened 2024-11-11
Add conditonal POS compilation target.

### LINKED ISSUE #4476: Add conditional compilation for HC/POW
We want to be able to conditionally compile a POW and a POS version to avoid changing the old API when adding HC functionality.

happi:
Fixes #4476 

Create some pos make targets, where POS is defined in the Erlang compilation for conditional compilation of HyperChains/POS versions of aeternity. 

When using this, consider how we should run tests and build docker images for HC in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ REBAR ?= ./rebar3
 
 PROTOCOLS = roma minerva fortuna lima iris ceres arcus
 CT_TARGETS = $(patsubst %,ct-%,$(PROTOCOLS))
+CT_POS_TARGETS = $(patsubst %,ct-pos-%,$(PROTOCOLS))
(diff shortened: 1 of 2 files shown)


hanssv:
What is the general idea w.r.t. testing PoS? A lot of tests are not meaningful to repeat (mining, HTTP API, etc.) and the HC tests require both a PoW node and a PoS node 🤔...

ThomasArts (commented):
The thinking is right, we need a way to re-use old test suites for PoS. At the same time, even when only working on PoS we need to run all tests for PoW to make sure things there don't break.
Not sure that having a POS macro and a lot of `ifdef` in test code is the best way, but it could work

hanssv (approved):

hanssv (approved):

ThomasArts (approved):

PR #4480, opened 2024-11-11
Add flag field to header records.

### LINKED ISSUE #4479: Add flags field to header record
The current flags in a block header can be derived from the content of the header. For Hyperchains, we want to add a flag independent of content. Therefore, we need to keep the flag field in the header and the database.

happi:
Fixes #4479 

Include the flags for both key- and micro-blocks in the header record.

Handle old versions of the record when read from the database.
Handles the old and new versions in binary and client serialization/deserialization.

Fixes the db function *get_top_block_node, which did not do a header upgrade.

### DIFF
--- a/apps/aecore/src/aec_db.erl
+++ b/apps/aecore/src/aec_db.erl
@@ -1121,7 +1121,7 @@ dirty_get_top_block_node() ->
 %% Some migration code: Ideally, top_block_node is there, and we're done.
 %% If not, we should find top_block_hash. Fetch the corresponding
 %% header, delete the obsolete top_block_hash and put in the new top_block_node.
-%% We do not update the top blok in the DB if it is there, the next top block
+%% We do not update the top block in the DB if it is there, the next top block
 %% will be of the latest version and the get_top_block_node function converts
 %% on the fly.
 convert_top_block_entry() ->
--- a/apps/aecore/src/aec_db.erl
+++ b/apps/aecore/src/aec_db.erl
@@ -1103,14 +1103,27 @@ get_top_block_hash() ->
     end.
 
 get_top_block_node() ->
-    get_chain_state_value(top_block_node).
+    case get_chain_state_value(top_block_node) of
+        #{ header := Header, hash := Hash} ->
+            NewHeader = aec_headers:from_db_header(Header),
+            #{ header => NewHeader, hash => Hash};
+        undefined -> undefined
+    end.
 
 dirty_get_top_block_node() ->
-    dirty_get_chain_state_value(top_block_node).
+    case dirty_get_chain_state_value(top_block_node) of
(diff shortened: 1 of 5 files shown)


apps/aecore/src/aec_db.erl
```diff
@@ -1103,14 +1103,27 @@ get_top_block_hash() ->
...
+        #{ header := Header, hash := Hash} ->
+            NewHeader = aec_headers:from_db_header(Header),
+            #{ header => NewHeader, hash => Hash};
+        undefined -> undefined
+    end.
 
 %% Some migration code: Ideally, top_block_node is there, and we're done.
 %% If not, we should find top_block_hash. Fetch the corresponding
 %% header, delete the obsolete top_block_hash and put in the new top_block_node.
+%% We do not update the top blok in the DB if it is there, the next top block
```
hanssv, line 1124:
```suggestion
%% We do not update the top block in the DB if it is there, the next top block
```

apps/aecore/src/aec_headers.erl
```diff
@@ -753,8 +837,11 @@ deserialize_micro_from_binary(<<Version:32,
                             signature = Signature,
                             txs_hash = TxsHash,
                             time = Time,
-                            version = Version},
-            {ok, populate_extra(H)};
+                            version = Version,
+                            flags = << ?MICRO_HEADER_TAG:1,
+                                       PoFTag:1,
```
hanssv, line 842: Why bother with `PoFTag`, it will be overwritten by `fix_flags` anyway?

apps/aecore/src/aec_headers.erl
```diff
@@ -190,7 +192,13 @@ type(#mic_header{}) -> micro.
 
 -spec from_db_header(tuple() | header()) -> header().
 %% We might have a legacy tuple in the db
```
ThomasArts, line 194: As far as I understand the PR, it would be fine to upgrade to this version, get a database with fixed flags and then downgrade. Because the flags do not mean anything for the old version?

ThomasArts (dismissed):

apps/aecore/src/aec_headers.erl
happi: Good catch.
I started by trying to keep the flags correct when changing other values but added fix_flags to everything when the tests kept failing. The culprit was in aec_db. I will remove the redundant code.

ThomasArts (dismissed):

hanssv (dismissed):

apps/aecore/src/aec_headers.erl
ThomasArts: That's a pity...

ThomasArts (approved):

mitchelli (approved):

PR #4482, opened 2024-11-12
Update installation.md

ploetzly:
was missing Install from package manager instructions, which is included in the readthedocs doc site


klercker:
@ploetzly any reasons why this could not be brought up to date with `master` and merged? @dincho could probably help with merge if it's a problem. (I'm in PR cleanup mode today)

klercker (approved):

dincho:
We don't support packages anymore, see https://github.com/aeternity/aeternity/pull/4268 for the reasons.

ploetzly:
Ah ok. Is this an instance where the docs site (or some ways to access some of the pages) is not updated to the latest version of the GH docs?

klercker:
Cool, just as long as it gets handled 😃

dincho:
> Ah ok. Is this an instance where the docs site (or some ways to access some of the pages) is not updated to the latest version of the GH docs?

Not sure what do you mean, the node docs are using just a single source of truth, this repository. The ways to access it are via source code/git/github and Readthedocs.

dincho:
> but I made this PR because the ReadTheDocs version had information in it that the GH repo didn't. Why was that the case? Is it because there is a delay in the updates sometimes?

Can you provide a link to that info please?

PR #4484, opened 2024-11-13
Less strict expectations about MB production in HC-tests

hanssv:
This removes the mempool-check and the explicit MB check if there are TXs in the pool. Not sure if we need this kind of mechanism or if the test itself should just check for MB content. I'm leaning towards adding such a mechanism...

### DIFF
--- a/apps/aehttp/test/aehttp_hyperchains_SUITE.erl
+++ b/apps/aehttp/test/aehttp_hyperchains_SUITE.erl
@@ -1735,31 +1735,35 @@ produce_cc_blocks(Config, BlocksCnt, ParentProduce) ->
 produce_to_cc_height(Config, TopHeight, GoalHeight, ParentProduce) ->
     NodeNames = [ Name || {_, Name, _} <- ?config(nodes, Config) ],
     BlocksNeeded = GoalHeight - TopHeight,
-    case  BlocksNeeded > 0 of
+    case BlocksNeeded > 0 of
         false ->
             TopHeight;
         true ->
             NewParentProduce =
                 case ParentProduce of
-                    [{CH, PBs} | PRest ]  when CH == TopHeight+1 ->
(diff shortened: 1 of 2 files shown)


klercker (approved):
👍🏻

ThomasArts (approved):

apps/aehttp/test/aehttp_hyperchains_SUITE.erl
```diff
@@ -1735,31 +1735,35 @@ produce_cc_blocks(Config, BlocksCnt, ParentProduce) ->
...
+            %% This will mine 1 key-block (and 0 or 1 micro-blocks)
+            {ok, Blocks} = mine_cc_blocks(NodeNames, 1),
+
+            {Node, KeyBlock} = lists:last(Blocks),
+            case Blocks of
+                [{Node, MB}, _] ->
+                    ?assertEqual(micro, aec_blocks:type(MB)),
+                    ct:log("CC ~p produced micro-block: ~p", [Node, MB]);
+                [_] ->
+                    ok
```
klercker, line 1762: What is the expectation to see here? In the HC suite, `verify_fees`, I'm seeing two micro block (candidates?), and FAIL with `case_clause`. I'm guessing the first one (single spend to/from a non-validator adress i.e. not ALICE, LISA or BOB) has something to do with the actual test case, and the second one is the pinning "notarisation" spend to/from a validator. It looks like this should match the case, but who am I to say? BEAM obviously don't think so...

hanssv: There should never be two MBs for one "generation"/block. The check is correct.

klercker: Houston, we have a problem then.

PR #4486, opened 2024-11-14
Add child block production time configuration key

richcarl:
Needed for synchronizing hyperchains.

This PR is supported by Æternity Foundation.

PR #4485, opened 2024-11-14
Save epoch info in cache

richcarl:
Put the necessary epoch info in the cache for later use.

This PR is supported by Æternity Foundation.

PR #4487, opened 2024-11-14
Compute correct production time for the child chain block

richcarl:
Requires #4485 and  #4486.

### DIFF
--- a/apps/aecore/src/aec_consensus_hc.erl
+++ b/apps/aecore/src/aec_consensus_hc.erl
@@ -84,7 +84,6 @@
         , entropy_height/1
         , get_entropy_hash/1
         , get_contract_pubkey/1
-        , get_child_epoch_info/1
         ]).
 
 -ifdef(TEST).
@@ -627,6 +626,12 @@ child_block_time() ->
     Fun = fun() -> get_consensus_config_key([<<"child_block_time">>]) end,
     aeu_ets_cache:get(?ETS_CACHE_TABLE, child_block_time, Fun).
 
+child_block_production_time() ->
+    Fun = fun() -> get_consensus_config_key([<<"child_block_production_time">>]) end,
+    aeu_ets_cache:get(?ETS_CACHE_TABLE, child_block_production_time, Fun).
+
+
+
 %% This is the contract owner, calls shall be only available via protocol
 contract_owner() ->
     Fun = fun() ->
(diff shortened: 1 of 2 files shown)


apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -656,15 +662,44 @@ beneficiary() ->
...
+    next_beneficiary(TxEnv, Trees);
+next_beneficiary(TxEnv, Trees, CurrHeight) ->
+    %% note that this gets the epoch of the next block, not of the latest block
+    {ok, #{epoch := BlockEpoch, first := BlockEpochFirst}} = aec_chain_hc:epoch_info({TxEnv, Trees}),
+    CurrEpoch = if CurrHeight < BlockEpochFirst ->
+                        BlockEpoch-1;
+                   true ->
+                        BlockEpoch
+                end,
+    {Epoch, EpochFirst, EpochStartTime} = get_child_epoch_info(CurrEpoch),
```
kvakvs, line 688: `Epoch` is unused

richcarl:
I'm working on resolving the conflict with master.

happi:
Now I added some more conflict ;)

richcarl:
If you want to see what happens, run `rm -r _build/test/logs && make ct-latest SUITE=apps/aehttp/test/aehttp_hyperchains`, and then `less _build/test/logs/latest.hyperchains/dev1/log/aeternity.log*`. Searching for `EPOCH OF NEXT BLOCK` shows you each new timing calculation. (Currently not stable until epoch 2.) Search for `NO SLEEP` to find places where it has lagged way behind.

richcarl:
I have fixed the timing in the first epoch, dropped the extra logging, and disabled the old blocktime test since it is based on the old timing behaviour. If everyone's ok with the state of this branch now, I can commit it so we have it in master.

apps/aehttp/test/aehttp_hyperchains_SUITE.erl
```diff
@@ -430,6 +430,7 @@ check_blocktime(_Config) ->
     {ok, TopBlock} = rpc(?NODE1, aec_chain, top_key_block, []),
     check_blocktime_(TopBlock).
 
+%% TODO: rework this test - currently assumes naive block timing
```
hanssv, line 433: It only assumes that the chain never produces two blocks with less than block-time distance between them. That is a rather reasonable assumption, I guess. But maybe you want this on average or something... Some checks is possibly needed.

hanssv (approved):
I think this might work, there isn't really any test?

But the observation from looking at the code is that it might make many things simpler if T0 is the time of the last block of the previous epoch?! 

Not sure if it is worth reconsidering this assumption?! 🤔

hanssv:
Yes, I guess the computatation should rather be when is the next time it is my turn to produce something. (For some definition of "my" - given that we might act on behalf of multiple validators.) Not sure if you need to adapt any other parts of the logic for this 🤔

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -745,38 +755,86 @@ beneficiary() ->
...
+    ChildHeight = aec_blocks:height(KeyBlock),
+    RunEnv = aetx_env:tx_env_and_trees_from_top(aetx_transaction),
+    case leader_for_height(ChildHeight + 1, RunEnv) of
         {ok, Leader} ->
             SignModule = get_sign_module(),
             case SignModule:set_candidate(Leader) of
                  {error, key_not_found} ->
                      lager:debug("node shall not produce (leader ~p)", [Leader]),
-                     timer:sleep(200),
+                     timer:sleep(child_block_production_time()),
```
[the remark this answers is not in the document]
richcarl: Sorry, that's a leftover from some attempts at debugging. Good that you noted it. I just tried to put in something that would make more sense than a hardcoded "200" but didn't want to sleep too long either.

richcarl:
> But the observation from looking at the code is that it might make many things simpler if T0 is the time of the last block of the previous epoch?!

Could be. My reasoning was that if we want to make the block time a matter of voting, the current block time has to apply to the epoch we're in.

hanssv:
I don't think we will make block time adjustable by voting any time soon, we have mainly talked about changing the epoch length...

klercker:
Any reasons this hasn't been merged yet?

richcarl:
> Any reasons this hasn't been merged yet?

The code got worked into Dima's branch instead. Will close this.

PR #4491, opened 2024-11-15
Make logdir configuration work as intended

richcarl:
The logdir configuration is defined in `setup` part of the sys.config, just like the data dir, but it did not work as intended because lager wants to create the log directory on disk as soon as it starts, and at that point it will have its own default config settings. This fixes the problem and switches the start order of setup and lager.
Depends on #4496.

### DIFF
--- a/apps/aecore/src/aecore.app.src
+++ b/apps/aecore/src/aecore.app.src
@@ -22,6 +22,7 @@
     exometer_core,
     redbug,
     yamerl,
+    setup,
     lager,
     aeutils,
     base58,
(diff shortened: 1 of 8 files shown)


klercker:
build and run setup is beyond my expertise in so many ways, but are there any reasons we could not get this one out of the way?

richcarl:
There were still some complications as I recall, and it wasn't a high priority. But it would be nice to do because this blocked other things we wanted to fix with logging when running under Docker.

klercker (approved):

happi (approved):

PR #4492, opened 2024-11-18
Faster dev container builds

richcarl:
To speed up building from scratch in a dev container, we need to set the environment variable ERLANG_ROCKSDB_OPTS like we do in the main Aeternity Dockerfile, so that we link with the pre-built RocksDb library in the container, instead of building the entire RocksDb C++ codebase which takes quite a long time.

Also includes an example of how to easily mount your ~/.aeternity directory into the container.


kvakvs (approved):
najs

klercker (approved):

PR #4495, opened 2024-11-20
Add proper event handling for HC key-blocks

### LINKED ISSUE #4494: Create events on CC related to epochs
Aeternity has some event generation solution.

For a CC there are specific events that may be useful to emit:
- block generated with new stakeholder distribution
- last block of an epoch generated (or/and first block)
- pinned evidence appeared on chain
 

NOTE: Will affect MW

hanssv:
- Let the key-block pre transformation return events and emit these _after_ the block is successfully inserted. 
- Use this mechanism for `pin` events
- Add `new_epoch` event
- Use the `new_epoch` event to implement GC in the HC vote pool

Closes #4494 (though an event for staking pool is not explicitly added)


klercker (approved):
NIce. My next refac will then be to build out the pinning agent spawning to be triggered by the new post-block-insert instead.

ThomasArts (approved):

PR #4499, opened 2024-11-21
Pinning Agent refactor to `new_epoch` event

klercker:
Move pinning agent invocation code from inner consensus callbacks to separate gen_server+worker process.

### DIFF
--- a/apps/aecore/src/aec_pinning_agent.erl
+++ b/apps/aecore/src/aec_pinning_agent.erl
@@ -88,10 +88,6 @@ terminate(_Reason, _LoopState) ->
 code_change(_OldVsn, LoopState, _Extra) ->
     {ok, LoopState}.
 
-    % case aec_parent_connector:has_parent_account(LastLeader) of
-    %     true -> aec_pinning_agent:spawn_for_epoch(NextEpochInfo, get_contract_pubkey(?ELECTION_CONTRACT), LastLeader);
-    %     false -> lager:debug("AGENT: No parent chain account found for ~p", [LastLeader])
-    % end
 stop() ->
     lager:debug("STOPPED"),
     gen_server:stop(?SERVER).
@@ -113,7 +109,7 @@ spawn_for_epoch(#{last := Last} = EpochInfo, Contract) ->
 
 spawn_for_epoch(EpochInfo, Contract, LastLeader) ->
     try
-    case whereis(my_unique_process) of
+    case whereis(?WORKER) of
         undefined ->
             Pid = spawn(aec_pinning_agent, start, [EpochInfo, Contract, LastLeader]),
             register(?WORKER, Pid),
@@ -129,10 +125,12 @@ spawn_for_epoch(EpochInfo, Contract, LastLeader) ->
 start(EpochInfo, Contract, LastLeader) ->
     #{ epoch      := _Epoch
      , last       := Last
+     , first      := First
      , validators := _Validators} = EpochInfo,
     subscribe(),
     lager:debug("pinning worker started ~p", [self()]),
-    wait_for_top_changed(Last - 1, none, false, Contract, LastLeader).
+    PCPinTx = post_pin_to_pc(LastLeader, First),
+    wait_for_top_changed(Last - 1, PCPinTx, false, Contract, LastLeader).
 
 
 %%%=============================================================================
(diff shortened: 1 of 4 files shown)


apps/aecore/src/aec_pinning_agent.erl
```diff
@@ -8,120 +8,192 @@
...
+handle_info(_Info, LoopState) ->
+    {noreply, LoopState}.
+
+terminate(_Reason, _LoopState) ->
+    ok.
 
+code_change(_OldVsn, LoopState, _Extra) ->
+    {ok, LoopState}.
+
+    % case aec_parent_connector:has_parent_account(LastLeader) of
```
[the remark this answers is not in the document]
klercker: Oops, was that still in there?! I'll hold on with comitting in case some other comments drop in.

apps/aecore/src/aec_pinning_agent.erl
```diff
@@ -8,120 +8,192 @@
...
+
 spawn_for_epoch(EpochInfo, Contract, LastLeader) ->
-    lager:debug("AGENT: Trying to spawn pinning agent...", []),
     try
     case whereis(my_unique_process) of
         undefined ->
             Pid = spawn(aec_pinning_agent, start, [EpochInfo, Contract, LastLeader]),
-            register(?MODULE, Pid),
-            Pid;
+            register(?WORKER, Pid),
```
hanssv, line 119: This is not the unique name `my_unique_process` that is tested for above?  

But bigger question, why on earth do we spawn yet another process in here, and also wait for another top change.

hanssv (changes requested):
The gen-server process should be enough to run this, it is even aptly named something_worker. Spawning yet another process is overkill.

klercker:
The `gen_server` process is actually named `?SERVER` as per usual, the sub process is `?MODULE_worker`. Again, the state handling becomes a lot cleaner with the spawned process, and that single process is pretty cheap as we all know.

apps/aecore/src/aec_pinning_agent.erl
```diff
@@ -8,120 +8,184 @@
...
+
+terminate(_Reason, _LoopState) ->
+    ok.
+
+code_change(_OldVsn, LoopState, _Extra) ->
+    {ok, LoopState}.
+
+stop() ->
+    lager:debug("STOPPED"),
+    gen_server:stop(?SERVER).
```
hanssv, line 93: This is an API function, not a callback

apps/aecore/src/aec_pinning_agent.erl
```diff
@@ -8,120 +8,184 @@
...
+    {noreply, LoopState}.
+
+terminate(_Reason, _LoopState) ->
+    ok.
+
+code_change(_OldVsn, LoopState, _Extra) ->
+    {ok, LoopState}.
+
+stop() ->
+    lager:debug("STOPPED"),
```
hanssv, line 92: Is that logging really necessary 🤔

hanssv:
> The `gen_server` process is actually named `?SERVER` as per usual, the sub process is `?MODULE_worker`. Again, the state handling becomes a lot cleaner with the spawned process, and that single process is pretty cheap as we all know.

Yes, naming looks updated and correct, the drawback of spawning yet another process is that you need to register for events every time, etc. It is much cleaner and not a lot of work to just put that logic directly in the gen_server.

klercker:
As per "advice" from @richcarl and @happi and others I put it in a separate worker process and I like the look of it (and I really don't like the look of `gen_server`; it's a great pattern/whatever but the code looks like the llamas arse). I think the number of LOC managing event subscription will actually be higher with `gen_server` but my .erl incompetence might be fooling me here. Since you might or might not want to listen to `top_change` all the time (when you are not pinning), you need to either match for it somewhere (unpacking `#state{}` etc) or turn on and off subscription as you go along. But please prove me wrong.

If it's very very important to you @hanssv, I will refac again, but I really do like the current impl and it's not particularly non-standard.

klercker:
Not superior perhaps, but valid 😅

apps/aehttp/src/aehttpc_aeternity.erl
```diff
@@ -204,17 +204,17 @@ post_pin_tx(SignedSpendTx, NodeSpec) ->
...
-          , nonce       => Nonce
-          , contract_id => aeser_id:create(contract, ContractPubkey)
-          , abi_version => ABI
-          , fee         => 1000000 * min_gas_price()
-          , amount      => Amount
-          , gas         => 1000000
-          , gas_price   => min_gas_price()
-          , call_data   => CallData},
+        #{caller_id   => aeser_id:create(account, Who)
+        , nonce       => Nonce
```
hanssv, line 210: Indentation?

hanssv (approved):
Looks good, this one is much nicer!

mitchelli (approved):

happi (approved):

PR #4503, opened 2024-11-26
static, configurable coinbase for HC

klercker:
Static, configurable Coinbase for Hyperchain.

Set value with `hyperchain_static_coinbase`config parameter in consensus config.

This PR is supported by Æternity foundation.


klercker:
**BEWARE**: renaming of the ~900+ guard method (to be internal) makes the `aec_coinbase.erl`patch pretty extensive. It is really only adding a "_" to the end of the name...

apps/aehttp/src/aehttpc_aeternity.erl
happi, line 1: Is there an actual change here?

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -1748,6 +1748,11 @@
                                             "default": 0,
                                             "type": "integer"
                                         },
+                                        "hyperchain_static_coinbase": {
+                                            "type": "integer",
+                                            "default": 0,
+                                            "description": "HC coinbase value for HyperChains consensus"
```
happi, line 1754: Could you explain more, such as that this is the block production reward?

happi: And that this many new tokens will be minted for each block resulting in some inflation.

apps/aehttp/src/aehttpc_aeternity.erl
```diff
@@ -206,15 +206,15 @@ pin_contract_call(ContractPubkey, PinTx, Who, Amount, _Fee, SignModule) ->
...
-        , call_data   => CallData},
+        #{ caller_id   => aeser_id:create(account, Who),
+           nonce       => Nonce,
+           contract_id => aeser_id:create(contract, ContractPubkey),
+           abi_version => ABI,
+           fee         => 1000000 * min_gas_price(),
+           amount      => Amount,
+           gas         => 1000000,
+           gas_price   => min_gas_price(),
+           call_data   => CallData},
```
hanssv, line 217: With this style you're missing a space between `CallData` and `}` 😅

apps/aeutils/priv/aeternity_config_schema.json
happi: We should probably not go too deep into it here and explain it better in the external documentation, but I think a nice message to the person configuring this would be good.

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -555,6 +556,10 @@ default_pinning_behavior() ->
     Fun = fun() -> get_consensus_config_key([<<"default_pinning_behavior">>], false) end,
     aeu_ets_cache:get(?ETS_CACHE_TABLE, default_pinning_behavior, Fun).
 
+hyperchain_static_coinbase() ->
+    Fun = fun() -> get_consensus_config_key([<<"hyperchain_static_coinbase">>], false) end,
```
hanssv, line 560:
```suggestion
    Fun = fun() -> get_consensus_config_key([<<"hyperchain_static_coinbase">>], 0) end,
```

apps/aeutils/priv/aeternity_config_schema.json
klercker: Since I really have no exact idea what it's for, I hereby solicit suggestions of what this should read.

apps/aecore/src/aec_consensus_hc.erl
hanssv: I don't remember how it works but `false` is definitely wrong, better with `0`

klercker: But, I'll make that change, if my fix for default values from schema _is_ in, then that will sort itself out.

apps/aeutils/priv/aeternity_config_schema.json
klercker: @happi ???

dincho:
The "hyperchain_" prefix makes no sense to me as it's already under the consensus configuration subtree

hanssv (approved):

hanssv (dismissed):

happi (dismissed):

hanssv (approved):

happi (approved):

PR #4504, opened 2024-11-27
Staking contracts overhaul

hanssv:
New staking contracts... 

Hyperchains suite passes, and a new rather sparse staking contract suite also works.

Rewards are accumulated in the election contract and then distributed to stakers at the end of epoch (with one epoch delay). Each staker has a `StakingValidator` contract, created through a factory in `MainStaking`.

### DIFF
--- a/test/contracts/HCElection.aes
+++ b/test/contracts/HCElection.aes
@@ -105,6 +105,7 @@ main contract HCElection =
     put(state{pin = Some(proof)})
 
   payable stateful entrypoint add_reward(height : int, to : address) =
+    assert_protocol_call()
     put(state{rewards[height = []] @ rs = (to, Call.value) :: rs})
 
   entrypoint leader() =
--- a/test/contracts/HCElection.aes
+++ b/test/contracts/HCElection.aes
@@ -3,6 +3,7 @@ include "Pair.aes"
 
 contract interface MainStaking =
   entrypoint sorted_validators : () => list((address * int))
+  entrypoint add_rewards : (int, list(address * int)) => unit
 
 main contract HCElection =
   record epoch_info =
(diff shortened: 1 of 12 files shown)


test/contracts/HCElection.aes
```diff
@@ -102,12 +104,12 @@ main contract HCElection =
     require(Call.caller == state.leader, "Must be called by the last leader of epoch")
     put(state{pin = Some(proof)})
 
+  payable stateful entrypoint add_reward(height : int, to : address) =
```
ThomasArts, line 107: is there a reason to assert that the height is not in the past? Would not be catastrophic in itself if consensus puts an award in validator pocket, but can someone else fiddle with a small reward and destroy the logic of the contract?

hanssv: Since the reward delay (how many blocks in the past we distribute rewards for) is configurable we would need to carry that value in here to make a correct check. Perhaps we shouldn't let it be configurable (keep it at 1?), for HC, then we can tighten this.

There is another assertion missing here though, this should only be allowed to be called by the consensus logic... Adding that in a later commit.

happi (approved):
Hard to read the diffs, had to checkout out locally to read the contracts.

hanssv:
> Hard to read the diffs, had to checkout out locally to read the contracts.

For the contracts it is almost a complete rewrite, so I am not surprised diff's aren't super-helpful.

test/contracts/StakingValidator.aes
```diff
@@ -1,130 +1,70 @@
...
+    assert_owner_caller()
+    put(state{reward_callback = Some(cb_ct)})
 
-  function assert_caller() =
-    require(Call.caller == state.main_staking_ct, "Only allowed through MainStaking")
+  entrypoint rewards(epoch : int, amount : int, restaked : bool) =
+    assert_main_staking_caller()
+    switch(state.reward_callback)
+      None => ()
+      Some(cb_ct) => cb_ct.reward_cb(epoch, amount, restaked)
```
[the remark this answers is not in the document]
hanssv: That is probably a good idea 👍 - I haven't had time to test in that detail yet.

PR #4509, opened 2024-12-04
Improve staking contracts

hanssv:
- Pinning reward should be part of the staking reward flow
- Stake should be recorded (and enforced/locked) on a per-epoch basis
- To facilitate for the StakingValidator owner being a contract, we need a separate field for the block signing key
- The callback function should be restricted, both with protected = true and a gas limit (the exact limit could be discussed…)
- Add the ability to change the auto re-stake setting
- Cache epoch info in MainStaking + getter
- Function to check if callback is defined in `StakingValidator`
- Rudimentary tests

Implements #4506

### DIFF
--- a/test/contracts/HCElection.aes
+++ b/test/contracts/HCElection.aes
@@ -3,6 +3,7 @@ include "Pair.aes"
 
 contract interface MainStaking =
   entrypoint sorted_validators : () => list((address * int))
+  entrypoint lock_stake  : (int) => list((address * int))
   entrypoint add_rewards : (int, list(address * int)) => unit
 
 main contract HCElection =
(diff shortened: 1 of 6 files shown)


test/contracts/HCElection.aes
```diff
@@ -3,6 +3,7 @@ include "Pair.aes"
 
 contract interface MainStaking =
   entrypoint sorted_validators : () => list((address * int))
+  entrypoint lock_stake  : (int) => list((address * int))
```
ThomasArts, line 6: Not important, but would `locked_stake` not be a better name for the stake locked at that epoch?
Whereas `lock_state` suggests that you do lock state for that epoch?

ThomasArts (approved):
looks good to me

happi (approved):

test/contracts/HCElection.aes
hanssv: We do lock stake for that epoch, but for convenience we also return the stake that was just locked?!

PR #4513, opened 2024-12-08
[deprecated branch] Hole block production (no fork resolution)

kvakvs:
This is rewrite of `aec_consensus_hc` timing and production logic. It brings the new logic implementing a stable normal (happyu path only) production algorithm. Not included: Fork resolution/best fork selection.

## Notable Changes

- New function added to `aec_blocks` to create a block with a new flag (29th bit for a hole block), a new function added to `aec_headers` to check whether one is a hole.
- For epoch 1 the production works from chain height, the chain will be slow to start, so timeslots are not going to cut it
- From epoch 2 production is based on timeslots, which are calculated from each epoch's start and based on the time passed
- In hyperchains conductor implementation and with the lack of PoW to slow the things down, `aec_consensus_hc` module is called in a tight loop (very quick repeated invocations) so there were some extra changes added like caching "do not produce" decisions and reducing the debug logging density (does not affect release builds).
- The test for this code is moved to a separate application `hc_test` and more tests will be added here later. The normal production test is active in CT `groups` function, and the other isn't.

## Note to Reviewers

* Sophia file changes *.aes do not contain my changes but the test wants to read them, might have to revert/take a specific version of those.
* Code in hctest_* modules (non-CT suites) is a lightly modified copy of setup code from `aehttp_hyperchains_SUITE`, most importantly it allows skipping some nodes in RPC calls and checks, when your test brings them offline.

## Things To Solve Next

* EOE halt, when production of EOE block fails for example due to missing the time slot, the chain is soft-locked.
* Fork resolution when one or more hole blocks appear


apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -275,15 +275,17 @@ micro_block_height_relative_previous_block(micro, MicroHeight) ->
...
     Height = aec_block_insertion:node_height(Node),
     state_pre_transform_node(key, Height, PrevNode, Trees).
 
 state_pre_transform_micro_node(Height, PrevNode, Trees) ->
-    state_pre_transform_node(micro, Height, PrevNode, Trees).
+    {T, _} = state_pre_transform_node(micro, Height, PrevNode, Trees),
+    T.
 
+-spec state_pre_transform_node(_Type :: atom(), Height :: non_neg_integer(), _PrevNode :: any(), Trees :: aec_trees:trees())
+        -> {aec_trees:trees(), list()}.
```
hanssv, line 288: What is in the list?

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -275,15 +275,17 @@ micro_block_height_relative_previous_block(micro, MicroHeight) ->
...
 state_pre_transform_node(_Type, Height, _PrevNode, Trees) when Height < 1 ->
     %% No leader for genesis
     Trees;
```
hanssv, line 291: Shouldn't it return a 2-tuple?

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -292,40 +294,84 @@ state_pre_transform_node(Type, Height, PrevNode, Trees) ->
...
+            try
+            case aec_parent_connector:has_parent_account(LastLeader) of
+                true -> aec_pinning_agent:spawn_for_epoch(NextEpochInfo, get_contract_pubkey(?ELECTION_CONTRACT), LastLeader);
+                false -> lager:debug("AGENT: No parent chain account found for ~p", [LastLeader])
+            end
+            catch
+                T:E -> lager:debug("AGENT throws: ~p:~p", [T,E])
+            end;
+        _ -> ok
+    end.
```
hanssv, line 373: Old stuff, right?

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -347,19 +393,64 @@ cache_child_epoch_info(Epoch, Height, StartTime) ->
             ok
     end.
 
-get_child_epoch_info(Epoch) ->
+%% Cache stores a tuple {MaxEpoch, [Info...]} where Info :: {Epoch, Height, StartTime}
+%% If the epoch cache is empty and start time is unknown, i will
+%% * Try read epoch first block, if it can, and its time will be epoch start
+%% * Else, will try read last block of previous epoch, then its time + block production time
+%%   will be the epoch start
```
hanssv, line 400: Isn't it block_time that decides the time between blocks, not block_production_time?

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -347,19 +393,64 @@ cache_child_epoch_info(Epoch, Height, StartTime) ->
...
+            #{epoch => Epoch,
+              height => EpochFirst,
+              start_time => aec_headers:time_in_msecs(aec_blocks:to_header(FirstBlock))
+            };
+        _Error1 ->
+            %% Not found, try read the last block of the previous epoch
+            case aec_chain:get_key_block_by_height(EpochFirst - 1) of
+                {ok, PrevBlock} ->
+                    T = aec_headers:time_in_msecs(aec_blocks:to_header(PrevBlock))
+                        + child_block_time() - child_block_production_time(),
```
hanssv, line 443: Ok, that is not what the comment said... But this is what I expected.

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -292,40 +294,84 @@ state_pre_transform_node(Type, Height, PrevNode, Trees) ->
     {ok, PrevHash} = aec_headers:hash_header(PrevHeader),
     {TxEnv0, _} = aetx_env:tx_env_and_trees_from_hash(aetx_transaction, PrevHash),
     TxEnv = aetx_env:set_height(TxEnv0, Height),
+
     {ok, #{epoch := Epoch, first := EpochFirst, last := EpochLast} = EpochInfo} =
         aec_chain_hc:epoch_info({TxEnv, Trees}),
-    {ok, Leader} = leader_for_height(Height, {TxEnv, Trees}),
+
+    #{timeslot := Timeslot} = get_timeslot_and_epoch({TxEnv, Trees}),
```
hanssv, line 301: Wait, what? Timeslot is computed using wallclock time (Now)? How can this work when syncing? 🤔

test/contracts/HCElection.aes
```diff
@@ -3,8 +3,9 @@ include "Pair.aes"
 
 contract interface MainStaking =
   entrypoint sorted_validators : () => list((address * int))
-  entrypoint lock_stake  : (int) => list((address * int))
-  entrypoint add_rewards : (int, list(address * int)) => unit
+  entrypoint is_validator : (address) => bool
+  entrypoint total_stake : () => int
+  stateful entrypoint post_elect : () => unit
```
hanssv, line 8: Of course you can't just revert the consensus contracts, but the tests will explain that 😅

hanssv (commented):
Only a first look, I am a bit worried about the use of wallclock time in state_pre_transform...

Also the rebase looks weird, some things are new, some are old!?

apps/aecore/src/aec_block_hole_candidate.erl
```diff
@@ -0,0 +1,110 @@
...
+                                BlockHash,
+                                Block,
+                                Beneficiary,
+                                Miner,
+                                Headers,
+                                Protocol
+                            );
+                        error ->
+                            {error, headers_for_target_adjustment_not_found}
+                    end
```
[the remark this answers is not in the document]
kvakvs: This code was provided by Erik, and i have experience that the code was creating what it should.

apps/aecore/src/aec_consensus_hc.erl
kvakvs: Dialyzer has also pointed that out. Nice catch!

apps/aecore/src/aec_consensus_hc.erl
kvakvs: Will try and roll it back, these changes do not belong here but got them from Richard's branch which is somewhat old.

apps/aecore/src/aec_consensus_hc.erl
kvakvs: Makes sense, changed. Thanks for catching!
Not sure how to test this, and the problem of not closing the previous epoch is an unsolved problem, the chain will halt if this happens. Otherwise the value is always cached in ETS and this code normally does not run, only for an emergency situation.

apps/aecore/src/aec_consensus_hc.erl
kvakvs: Okay, sounds like we need a different source of time, which will correctly point to the chain time when syncing, and which will be wall clock when producing, i.e. discriminate between two modes of operation. I might need more reading on how sync mechanism works.

kvakvs: For syncing i use height from the epoch which is equivalent of timeslots if no slot was skipped, so that should work.

kvakvs:
The work has moved to a new branch at #4526

PR #4514, opened 2024-12-15
Implement BFT voting

mitchelli:
Implement basics of BFT voting as described here: [BFT Voting Process](https://github.com/aeternity/hyperchains-whitepaper/blob/master/Periodically-Syncing-HyperChains.md#bft-voting-process)

### DIFF
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -1339,26 +1339,29 @@ hc_create_block_fun(ConsensusModule, TopHash) ->
     end.
 
 hc_create_block(ConsensusModule, TopHash0, Leader) ->
-    TopHash = hc_create_microblock(ConsensusModule, TopHash0, Leader),
+    VoteResult = ConsensusModule:vote_result(),
+    TopHash = hc_create_microblock(ConsensusModule, TopHash0, Leader, VoteResult),
     Res = aec_block_key_candidate:create(TopHash, Leader, Leader),
     Res.
 
-hc_create_microblock(ConsensusModule, TopHash, Leader) ->
-    case aec_block_micro_candidate:create(TopHash) of
+hc_create_microblock(ConsensusModule, TopHash, Leader, VoteResult) ->
+    CreateResult = case VoteResult of
+                        {ok, VoteTx} ->
+                            Tx = aetx_sign:tx(VoteTx),
+                                case aec_chain:get_block(TopHash) of
+                                    {ok, Block} ->
+                                        Gas = aetx:gas_limit(Tx, aec_blocks:height(Block), aec_blocks:version(Block)),
+                                        aec_block_micro_candidate:create(TopHash, Gas);
+                                    Error ->
+                                        lager:warning("Couldn't calculate gas for vote transaction ~p", [Error]),
+                                        aec_block_micro_candidate:create(TopHash)
+                                end;
+                        _ ->
+                            aec_block_micro_candidate:create(TopHash)
+                    end,
+    case CreateResult of
         {ok, MBlock, MBlockInfo} ->
-            {ok, Trees} = aec_block_micro_candidate:trees(MBlockInfo),
-            MBlock1 = case ConsensusModule:vote_result(Trees) of
-                                {ok, VoteTransaction} ->
-                                    case aec_block_micro_candidate:update(MBlock, [VoteTransaction], MBlockInfo) of
-                                        {ok, UpdatedMBlock, _MBlockInfo} ->
-                                            UpdatedMBlock;
-                                        Error ->
-                                            lager:warning("Error adding vote transaction ~p", [Error]),
-                                            MBlock
-                                    end;
-                                _ ->
-                                    MBlock
-                            end,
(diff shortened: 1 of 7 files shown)


apps/aehttp/test/aehttp_hyperchains_SUITE.erl
```diff
@@ -1317,6 +1325,29 @@ check_default_pin(Config) ->
...
+    mine_to_last_block_in_epoch(Node, Config),
+    {ok, _} = produce_cc_blocks(Config, 2),
+    #{producer := Producer, epoch := FEpoch, votes := Votes} = rpc(Node, aec_chain_hc , finalize_info, []),
+    FVoters = lists:map(fun(#{producer := Voter}) -> Voter end, Votes),
+    TotalStake = lists:foldl(fun({_, Stake}, Accum) -> Stake + Accum end, 0, Validators),
+    VotersStake = lists:foldl(fun(Voter, Accum) -> proplists:get_value(Voter, Validators) + Accum end, 0, FVoters),
+    TotalVotersStake = proplists:get_value(LastLeader, Validators) + VotersStake,
+    ?assertEqual(Producer, LastLeader),
+    ?assertEqual(Epoch, FEpoch),
+    ?assert(TotalVotersStake >= TotalStake * 0.6).
```
hanssv, line 1349: 0.66?

test/contracts/HCElection.aes
```diff
@@ -109,6 +127,14 @@ main contract HCElection =
     assert_protocol_call()
     put(state{rewards[height = []] @ rs = (to, Call.value) :: rs})
 
+  stateful entrypoint finalize_epoch(epoch_number : int, fork : bytes(), epoch_length : int, pc_hash : bytes(), producer : address, votes : list(vote)) =
+    let epoch = state.epoch
+    let last = state.epochs[epoch].start + state.epochs[epoch].length - 1
+    require(Chain.block_height == last, "Only in last block")
+    require(Call.caller == state.leader, "Must be called by the last leader of epoch")
+    require(epoch_number == state.epoch, "Not correct epoch")
```
hanssv, line 135: nitpick, but let's do the epoch check first?

apps/aecore/src/aec_eoe_vote.erl
```diff
@@ -0,0 +1,512 @@
...
+
+-spec get_finalize_transaction(aec_trees:trees()) -> {ok, aetx_sign:signed_tx()} | {error, not_ready} | {error, term()}.
+get_finalize_transaction(Trees) ->
+    gen_statem:call(?MODULE, {get_finalize_transaction, Trees}).
+
+%%% gen_statem callbacks
+
+%% Initialization: Start in the await_eoe state
+init([Stakers, BlockTime]) ->
+    aec_events:subscribe(tx_received),
```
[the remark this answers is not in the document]
mitchelli: Yes, I think that would be more efficient. I have checked with other types of transactions. I noticed if the leader had a spend transaction in the last micro block the creation of the finalisation transaction failed because the nonce was incorrect. Now that I use the trees from aec_block_micro_candidate this is no longer an issue.

happi (dismissed):
I think this is good for now, we should sit down and figure out what we want voting to look like in 1.0 and beyond that when we have time to clean up blocks and perhaps add some fields for the voting there.

apps/aecore/src/aec_eoe_vote.erl
```diff
@@ -0,0 +1,512 @@
...
+
+handle_proposal(SignedTx, #data{leader=Leader, epoch=Epoch, fork_hash=Hash} = Data) ->
+    %% Handle the proposal phase
+    %% Check the transaction contains a proposal
+    case convert_transaction(SignedTx) of
+        {ok, ?PROPOSAL_TYPE, Epoch, Leader, ProposalFields} ->
+            Data1 = Data#data{proposal=ProposalFields},
+            case Hash of
+                %% Negotation hasn't begun yet
+                undefined ->
```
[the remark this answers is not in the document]
mitchelli: The next line is supposed to store the proposal for later but I notice it will fail because the leader, etc. have not been set.

apps/aecore/src/aec_conductor.erl
```diff
@@ -1339,26 +1339,29 @@ hc_create_block_fun(ConsensusModule, TopHash) ->
...
 
-hc_create_microblock(ConsensusModule, TopHash, Leader) ->
-    case aec_block_micro_candidate:create(TopHash) of
+hc_create_microblock(ConsensusModule, TopHash, Leader, VoteResult) ->
+    CreateResult = case VoteResult of
+                        {ok, VoteTx} ->
+                            Tx = aetx_sign:tx(VoteTx),
+                                case aec_chain:get_block(TopHash) of
+                                    {ok, Block} ->
+                                        Gas = aetx:gas_limit(Tx, aec_blocks:height(Block), aec_blocks:version(Block)),
```
hanssv, line 1353: Just make sure that `Gas` is reasonably large here, there was a TODO w.r.t. gas/fee?

hanssv (dismissed):
Yes, just like that!

hanssv (approved):

PR #4515, opened 2024-12-17
Parent connector pinning refac

klercker:
Refactoring pinning-related methods, mostly out of `parent_connector` into the `pinning_agent` and some other places. Generally, the `parent_connector` now only handles the two calls to the parent chain (posting and fetching the pin data). This is in prep for pinning to Doge and BTC.

### DIFF
--- a/apps/aecore/src/aec_parent_connector.erl
+++ b/apps/aecore/src/aec_parent_connector.erl
@@ -44,17 +44,17 @@
         get_parent_chain_type/0,
         %% Pinning
         pin_to_pc/3,
-        pin_tx_to_cc/4,
+        %pin_tx_to_cc/4,
         pin_contract_call/5,
         get_pinning_data/0,
         create_pin_tx/5,
         post_pin_tx/1,
         get_pin_by_tx_hash/1,
-        encode_parent_pin_payload/1,
-        decode_parent_pin_payload/1,
-        encode_child_pin_payload/1,
-        decode_child_pin_payload/1,
-        is_pin/1,
+        % encode_parent_pin_payload/1,
+        % decode_parent_pin_payload/1,
+        % encode_child_pin_payload/1,
+        % decode_child_pin_payload/1,
+        % is_pin/1,
         find_spends_to/1,
         has_parent_account/1
         ]).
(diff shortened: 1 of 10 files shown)


apps/aecore/src/aec_parent_connector.erl
```diff
@@ -157,38 +147,18 @@ get_parent_chain_type() ->
 get_pinning_data() ->
     gen_server:call(?SERVER, get_pinning_data).
 
+%%=============================================================================
+%% Pin the last block of the previous epoch to the parent chain, using the
+%% parent chain account associated with the provided staker pubkey. Returns the
+%% parent chain TX hash (if TX is finalized on the parent chain, it will be
+%% featchable using pet_pin_by_tx_hash/1).
```
hanssv, line 154:
```suggestion
%% featchable using get_pin_by_tx_hash/1).
```

hanssv (dismissed):

apps/aecore/src/aec_parent_connector.erl
```diff
@@ -453,7 +384,7 @@ get_pinning_data_(Mod, NetworkID) ->
         {ok, #{epoch             => PrevEpoch,
                height            => Height,
                block_hash        => BlockHash,
-               parent_payload    => Mod:encode_parent_pin_payload(#{epoch => PrevEpoch, height => Height, block_hash => BlockHash}),
+               parent_payload    => aeser_hc:encode_parent_pin_payload(#{epoch => PrevEpoch, height => Height, block_hash => BlockHash}),
```
ThomasArts, line 387: This assumes the same binary fits on each parent chain (since parent chain is not passed as an argument). I guess you checked that indeed this works for Doge and BTC

klercker: I'm about to. Let's assume that. This refactoring is AE certified compatible only :)

klercker: From early tests it looks pretty good tho

ThomasArts (approved):
You remove quite a bit of PIN testing, but keep the pin by wallet test. Is there anything untested now by this cleanup?

Nice cleanup in general.

klercker:
The way I'm thinking is that the removed test cases esentially tested the code that is now in the `pinning agent` module, so the `default_pin` test group for that excercises that code now. What tests were removed that we might want to add back in the future are the various failing pinning "methods", that are now not possible to do using `parent_connector`, and only through various kinds of third party pinning. The wallet test could/should probably be added to in future to test those things. But the fail-over code in the various modules is of course still intact.

klercker:
@ThomasArts I now have verified a parent pin roundtrip on Doge, so the pindata encode does seem to be compatible with at least Doge, and probably with BTC as well, since they both have the same 80 byte `OP_RETURN` limit.

hanssv (approved):

ThomasArts (approved):

PR #4516, opened 2024-12-17
Correctly adjust current_stake upon withdrawal (if necessary)

hanssv:
Fix a bug in withdrawal where `current_stake` is not properly adjusted + add `get_current_stake` accessor function in StakeValidator contract (and MainStaking contract).

### DIFF
--- a/test/contracts/MainStaking.aes
+++ b/test/contracts/MainStaking.aes
@@ -80,6 +80,13 @@ main contract MainStaking =
   entrypoint get_staked_amount(epoch : int) =
     get_staked_amount_(Call.caller, epoch)
 
+  entrypoint get_current_stake() =
+    get_current_stake_(Call.caller)
+
+  entrypoint get_current_stake_(validator : address) =
+    let v = state.validators[validator]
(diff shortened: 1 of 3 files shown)


test/contracts/MainStaking.aes
```diff
@@ -80,6 +80,13 @@ main contract MainStaking =
   entrypoint get_staked_amount(epoch : int) =
     get_staked_amount_(Call.caller, epoch)
 
+  entrypoint get_current_stake() =
+    get_current_stake_(Call.caller)
+
+  entrypoint get_current_stake_(validator : address) =
```
ThomasArts, line 86: function or entrypoint?

ThomasArts (approved):

test/contracts/MainStaking.aes
hanssv: I think you might want to use it directly in some cases, so it doesn't hurt to have it as an entrypoint, though the name is questionable in that case...

happi (approved):

PR #4519, opened 2024-12-20
Doge parent_connector 

klercker:
Parent connector for Dogecoin that supports default pinning. Requires a dogecoin core node running (preferrably local) with a primed wallet. See `aehttp_hyperchains_btc_doge_SUITE.erl` test suite for configuration options.

Test suite can only be run manually, comment out first line of `init_per_suite` if you have a proper (local) docgecoin core node with a primed wallet available.

Epoch length and timing is crucial when running a slower chain like Doge or BTC. Remember to set properly. This will also differ between for example `regtest` and `testnet` so adjust accordingly when running different kinds of tests.

The code should also support BTC without change, but this is largely untested.

### DIFF
--- a/apps/aehttp/src/aehttpc_doge.erl
+++ b/apps/aehttp/src/aehttpc_doge.erl
@@ -27,6 +27,7 @@ createrawtransaction/3,
 signrawtransaction/2,
 sendrawtransaction/2,
 getrawtransaction/2,
+decoderawtransaction/2,
 to_hex/1]).
 
 -behavior(aehttpc).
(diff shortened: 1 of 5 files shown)


richcarl (approved):

apps/aehttp/src/aehttpc_doge.erl
```diff
@@ -57,59 +63,56 @@ hash_to_integer(Hash) ->
...
+    {ok, Vout} = decoderawtransaction(NodeSpec, RawTx),
+    {ok, Hex} = find_pin(Vout),
+    {ok, PinMap} = aeser_hc:decode_parent_pin_payload(aeu_hex:hex_to_bin(Hex)),
+    UpdMap =
+        case Conf of
+            0 -> maps:put(pc_height, -1, PinMap);
+            _ ->
+                Hash = maps:get(<<"blockhash">>, Tx),
+                {ok, {Height, _, _, _, _}} = getblock(NodeSpec, <<>>, Hash, 2),
+                maps:put(pc_height, Height, PinMap) % TODO fetch block using BlockHash from Tx
```
mitchelli, line 85: Is this TODO needed? Don't you already fetch the block by hash?

mitchelli (approved):

apps/aehttp/src/aehttpc_doge.erl
klercker: Yes, fixed that and forgot to take out. Will fix and stash for next PR :)

PR #4521, opened 2024-12-20
4428 adjust child epoch length based on vote

mitchelli:
This is to adjust the child epoch length based on the speed of the parent chain and the result of validators voting on the change to the epoch length. Each validator has a vote based on their stake.

### DIFF
--- a/apps/aecore/src/aec_consensus_hc.erl
+++ b/apps/aecore/src/aec_consensus_hc.erl
@@ -319,7 +319,7 @@ state_pre_transform_node(Type, Height, PrevNode, Trees) ->
     {ok, PrevHash} = aec_headers:hash_header(PrevHeader),
     {TxEnv0, _} = aetx_env:tx_env_and_trees_from_hash(aetx_transaction, PrevHash),
     TxEnv = aetx_env:set_height(TxEnv0, Height),
-    {ok, #{epoch := Epoch, first := EpochFirst, last := EpochLast} = EpochInfo} =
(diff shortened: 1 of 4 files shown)


richcarl (dismissed):

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -337,7 +337,16 @@ state_pre_transform_node(Type, Height, PrevNode, Trees) ->
...
+                            FinalizeHeight = EpochFirst - 1,
+                            FinalizeEpoch = Epoch - 1,
+                            NewEpochLength = case aec_chain_hc:finalize_info(FinalizeHeight) of
+                                                #{epoch_length := FinalizeEpochLength, epoch := FinalizeEpoch} ->
+                                                    FinalizeEpochLength;
+                                                _ ->
+                                                    lager:warning("Finalize info not found for epoch ~p height ~p", [FinalizeEpoch, FinalizeHeight]),
+                                                    EpochLength
+                                              end,
+                            Trees2 = step_eoe(TxEnv, Trees1, Leader, Seed, NewEpochLength, -1, CarryOverFlag),
```
hanssv, line 349: Isn't this code basically asking the contract for a value, just so that the same value can be posted back to the same (consensus) contract?
Wouldn't it be more elegant to handle this in the contract?

mitchelli: Good point, I'll try to move it into the contract.

hanssv (approved):

richcarl (approved):

PR #4526, opened 2025-01-13
Hole block production (no fork resolution)

kvakvs:
This is rewrite of `aec_consensus_hc` timing and production logic. It brings the new logic implementing a stable normal (happyu path only) production algorithm. Not included: Fork resolution/best fork selection.

## Notable Changes

- New function added to `aec_blocks` to create a block with a new flag (29th bit for a hole block), a new function added to `aec_headers` to check whether one is a hole.
- For epoch 1 the production works from chain height, the chain will be slow to start, so timeslots are not going to cut it
- From epoch 2 production is based on timeslots, which are calculated from each epoch's start and based on the time passed
- In hyperchains conductor implementation and with the lack of PoW to slow the things down, `aec_consensus_hc` module is called in a tight loop (very quick repeated invocations) so there were some extra changes added like caching "do not produce" decisions and reducing the debug logging density (does not affect release builds).
- The test for this code is moved to a separate application `hc_test` and more tests will be added here later. The normal production test is active in CT `groups` function, and the other isn't.

## Note to Reviewers

* Sophia file changes *.aes do not contain my changes but the test wants to read them, might have to revert/take a specific version of those.
* Code in hctest_* modules (non-CT suites) is a lightly modified copy of setup code from `aehttp_hyperchains_SUITE`, most importantly it allows skipping some nodes in RPC calls and checks, when your test brings them offline.

## Things To Solve Next

* EOE halt, when production of EOE block fails for example due to missing the time slot, the chain is soft-locked.
* Fork resolution when one or more hole blocks appear


apps/aecore/src/aec_conductor.erl
```diff
@@ -1338,11 +1359,37 @@ hc_create_block_fun(ConsensusModule, TopHash) ->
...
+        undefined ->
+            gproc:reg(GprocKey),
+            Result = aec_block_hole_candidate:create(TopHash, Producer, Producer),
+            gproc:unreg(GprocKey),
+            Result;
+        _Pid ->
+            %% Production is already running in another worker, we cannot start a duplicate
+            lager:debug("Busy, try again: Production by ~w already running at ~w", [Producer, _Pid]),
+            {error, busy_try_again}
+    end.
```
hanssv, line 1376: This looks very hacky, and also like the wrong solution to a problem we shouldn't have? We shouldn't try to create holes/blocks in parallel?

Also, is this actually a problem? I couldn't find `Busy, try again` in the logs after running the tests...

apps/aecore/src/aec_conductor.erl
```diff
@@ -1478,14 +1525,37 @@ handle_add_block(Header, Block, State, Origin) ->
...
+%% Used in PoS hyperchains to allow invalid_leader error and not crash the conductor.
+%% This error is still fatal in other chain types.
+safe_insert_block(aec_consensus_hc, Block, Origin) ->
+    try
+        aec_chain_state:insert_block_conductor(Block, Origin)
+    catch exit:{aborted, {{leader_validation_failed, invalid_leader} = Error, _Stack1}}:_Stack2 ->
+        %% This error is not fatal in PoS hyperchains, the block which failed validation should be ignored
+        %% Convert exception to an error value
+        {error, Error}
+    end;
```
hanssv, line 1537: Similar feeling about this one, why would we produce blocks with the wrong leader? And if we do, why should it fail in a different way for PoS?

Removing the `aec_consensus_hc` clause does not make the HC-suite fail 🤔

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -851,6 +1141,26 @@ get_entropy_hash(ChildEpoch) ->
             {error, not_in_cache}
     end.
 
+%% Final timeslot decisions, are cached so that we do not attempt to produce again when the decision
+%% has already been made. Example fact: We're not a leader and its not going to change in the same timeslot.
+-spec get_timeslot_skip_decision(Timeslot :: pos_integer()) -> false | atom().
+get_timeslot_skip_decision(Timeslot) ->
+    Decisions = aeu_ets_cache:lookup(?ETS_CACHE_TABLE, timeslot_skip_decision, #{}),
+    maps:get(Timeslot, Decisions, false).
```
hanssv, line 1149: Generic question, do we really need all this caching of decisions, shouldn't/couldn't we just let the conductor loop sleep until a point in time where the decision is no longer relevant? There is a lot of complexity here that I'm not convinced is necessary...

apps/aecore/src/aec_block_hole_candidate.erl
```diff
@@ -0,0 +1,116 @@
...
+                                BlockHash,
+                                Block,
+                                Beneficiary,
+                                Miner,
+                                Headers,
+                                Protocol
+                            );
+                        error ->
+                            {error, headers_for_target_adjustment_not_found}
+                    end
```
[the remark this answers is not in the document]
kvakvs: I will see to remove it

apps/aecore/src/aec_consensus_hc.erl
kvakvs: The conductor calls the consensus repeatedly with very small intervals, as there is no PoW. And it logs (debug logs) all the decision steps every time. Returning a status requesting a sleep (or sleeping in the consensus module) might work just as good. I will see to remove this and see what happens.

apps/aecore/src/aec_conductor.erl
kvakvs: The conductor calls the consensus repeatedly with very small intervals, as there is no PoW.
I think at some point from logs we've (me and Richard) observed that it does something weird and one possible explanation was this (sealing more than one block in parallel). I can roll back/remove this one and see if it still works, and that's great if it does.

hanssv:
This functionality got merged through #4546

PR #4529, opened 2025-01-20
Add comments to aec_block_generator

richcarl:


klercker (approved):

mitchelli (approved):

apps/aecore/src/aec_block_generator.erl
```diff
@@ -91,13 +107,13 @@ handle_cast(start_generation, State) ->
     {noreply, do_start_generation(State)};
 handle_cast({worker_done, Pid, {candidate, Candidate, CandidateState}},
             State = #state{ worker = {Pid, _} }) ->
-    %% Only publish non-empty micro-blocks
+    %% Only publish non-empty microblocks
     case aec_blocks:txs(Candidate) of
         [] ->
-            lager:debug("New empty microblock candidate generated, and discarded", []),
+            lager:debug("Empty microblock candidate prepared", []),
```
hanssv, line 113: `discarded` was the important bit here, I don't like it being removed from the message.

apps/aecore/src/aec_block_generator.erl
```diff
@@ -91,13 +107,13 @@ handle_cast(start_generation, State) ->
...
             ok;
         _  ->
-            epoch_mining:info("New microblock candidate generated", []),
+            epoch_mining:info("New microblock candidate ready", []),
```
hanssv, line 116: Not an improvement?

hanssv (changes requested):
The effort is noble...

apps/aecore/src/aec_block_generator.erl
richcarl: I was trying to find a wording that made sense. It's not discarded, just not published, and stays in the state ready to be filled when there are incoming tx events. Any better suggestion?

apps/aecore/src/aec_block_generator.erl
richcarl: Well, it's getting published as being ready to fetch. Tried to make it contrast with the alternative message above

hanssv (approved):
I give up if this is going to make anyone work faster please go ahead and merge it 🤯

PR #4532, opened 2025-01-22
HC: More better fee distribution

hanssv:
We want to distribute fees between current producer, and previous AND following producer. Here is the logic and configuration for this, not that master doesn't have any hole production yet, so this is a draft.

### DIFF
--- a/apps/aecore/src/aec_consensus_hc.erl
+++ b/apps/aecore/src/aec_consensus_hc.erl
@@ -500,9 +500,8 @@ state_grant_rewards(Node, Delay, Trees, {Fees, BlockReward}, [_Node1, _Node2, _N
 
     state_grant_reward_(TxEnv, Trees, Height, lists:zip(Nodes, Rewards), 0).
 
-%% TODO: placeholder
-is_hole(_Node) ->
-    false.
+is_hole(Node) ->
+    aec_headers:is_hole(aec_block_insertion:node_header(Node)).
 
 state_grant_reward_(_TxEnv, Trees, Height, [], HoleRewards) ->
     lager:info("Burning ~p tokens, because of holes, while distributing rewards for height ~p",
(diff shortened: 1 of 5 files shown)


ThomasArts (approved):
Seems to do the job. Later some more tests should be added to validate "after" block can get part of the fees as well, but obvius from code.

Documentation in schema, not in markdown yet, but that can be fixed when the actual distribution is known

dincho:
How this work together with `fixed_coinbase` ?

hanssv:
> How this work together with `fixed_coinbase` ?

It only affects the distribution of fees not coinbase.

ThomasArts (approved):
The placeholder for holes is not yet filled, so a little TODO for someone that wants to extend upon this in a later PR.

Another TODO is to document and test with the new default values that tokenomics people come up with.

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -450,13 +451,43 @@ cache_child_epoch_info(Epoch, Height, StartTime) ->
...
     TxEnv = aetx_env:tx_env_from_key_header(
               Header, aec_block_insertion:node_hash(Node),
               aec_block_insertion:node_time(Node), aec_block_insertion:node_prev_hash(Node)),
     Height = aec_block_insertion:node_height(Node) - Delay,
-    add_reward(TxEnv, Trees, Height, Beneficiary, Amount).
+
+    state_grant_reward_(TxEnv, Trees, Height, lists:zip(Nodes, Rewards), 0).
+
+%% TODO: placeholder
+is_hole(_Node) ->
```
happi, line 474: There is  aec_headers:is_hole now.

hanssv:
Fixed placeholder and rebased

happi (approved):

ThomasArts (dismissed):

hanssv:
Test updates are needed, but the config is going to be like this...

@kenodressel

happi (approved):

ThomasArts (approved):

PR #4533, opened 2025-01-24
Separate `hyperchains` application with hyperchains tests

kvakvs:
* Based on the existing `aehttp_hyperchains_suite`
* The generic code is split into `hctest.erl` and `hctest_ct_shared.erl`, unused code in both suites is eliminated.
* The parent and child chain parameters are passed from the `init_per_suite` into shared `init_per_suite`, and into shared `child_node_config` instead of hardcoding them as macros.
* The existing test can be deleted (currently not deleted)
* Potentially other hyperchains code might want to move here one day. Currently it is not included into the release and runs on the CT host only. NOTE: Is it automatically included into `ct-latest`?

## Trying Locally

* `make ct-latest SUITE=apps/hyperchains/test/hyperchains`
* `make ct-latest SUITE=apps/hyperchains/test/hyperchains_pinning`

This merge request is sponsored by the Æternity Foundation.


richcarl:
We will move these tests into the hole production PR instead.

PR #4537, opened 2025-01-28
Mark `started_at` as required in AuctionEntry

davidyuk:

I don't know for sure, but the missing `started_at` looks like a typo. How can an auction have an end but no beginning?

initially was implemented in https://github.com/aeternity/aeternity/commit/01dd226e9042e31117b43b1d9f59a43e040663c6


hanssv (approved):
Yes, looks like a mistake

PR #4541, opened 2025-02-02
Mark `encoded_key` as required in NamePointer

davidyuk:


hanssv:
I am pretty sure that value should not be `required` and CI seems to agree https://app.circleci.com/pipelines/github/aeternity/aeternity/15054/workflows/9fefdcbc-423b-4801-b818-c7c3f5de937b

davidyuk:
How do you like this? I'm not sure about the naming.
My goal is: I'm getting name details in typescript and I want it to don't warn me about probably missed `encoded_key`

hanssv:
Yes, I guess you need two different names/types then since that stupid language for describing the types is broken... I have no strong preferences on the exact naming, if this works I'm fine.

hanssv (approved):

apps/aehttp/priv/oas3.yaml
```diff
@@ -3285,6 +3285,14 @@ components:
       required:
         - key
         - id
+    NamePointerComplete:
```
davidyuk, line 3288: as alternative naming I can suggest
```suggestion
    NamePointerOut:
```
and `NamePointerIn` above

davidyuk:
> if this works I'm fine

I've removed $ref inside properties because I found that it is not working in autorest

> A common misconception is that $ref is allowed anywhere in an OpenAPI specification file

`parameters` can be an array of refs, not a ref by itself
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#fixed-fields-8

hanssv:
I've pushed the branch for tests...

hanssv (approved):

richcarl (approved):

PR #4543, opened 2025-02-07
4534 v10 voting follow branch fork

mitchelli:
This code follows the fork that has been elected by the vote.


apps/aecore/include/blocks.hrl
```diff
@@ -31,7 +32,8 @@
 %% Key blocks (bit 31:1)
 %% 30         ContainsInfo 1 - true, 0 - false
 %% 29         HC Hole block: 1 - is a hole block, 0 - ordinary block
-%% 0 - 28     Unused, should be 0.
+%% 28         EOE : 1 - end of epoch, 0 - not end of epoch)
```
hanssv, line 35: Only EoE with successful/quorum voting, right?

apps/aecore/src/aec_chain_state.erl
```diff
@@ -873,11 +900,27 @@ handle_top_block_change(OldTopNode, NewTopDifficulty, Node, Events, State) ->
...
+                    case aec_headers:is_eoe(NewHeader) of
+                        true ->
+                            false;
+                        false ->
+                            OldHeight = node_height(OldNode),
+                            NewHeight = node_height(NewNode),
+                            OldDifficulty > NewDifficulty orelse
+                                (OldDifficulty == NewDifficulty andalso OldHeight > NewHeight)
+                    end
+            end
```
hanssv, line 923: I'm confused, the idea with "resetting" target to height at Eoe was that the original logic for fork-resolution would cover also this case? I.e. I wasn't expecting _any_ change to this part of the logic 🤔

apps/aecore/src/aec_eoe_vote.erl
```diff
@@ -54,6 +54,29 @@
...
+                            true;
+                        _ ->
+                            false
+                    end;
+                false ->
+                    false
+            end;
+        _ ->
+            false
+    end.
```
hanssv, line 78: Shouldn't this also check that the call is successful? 

Would it be possible to instead directly inspect the consensus contract state _after_ the microblock 🤔

hanssv (approved):
I'm sorry I am late to the party, too many other obligations :sadpanda:

I'm not convinced the check is correct (shouldn't it check that the ContracCallTx is successful?) - but we have other issues that is perhaps more urgent, @happi your call I guess. I'll tentatively approve not to block progress.

apps/aecore/include/blocks.hrl
mitchelli: Yes, I only add if it is successful, I will update the comment.

apps/aecore/src/aec_chain_state.erl
mitchelli: Yes, I wasn't sure about this either but it what we decided in Stockholm, I think it can be removed.

apps/aecore/src/aec_eoe_vote.erl
mitchelli: That sounds like a better idea. I guess I can check when validating the key block.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -733,6 +737,29 @@ assert_height_delta(sync, Height, TopHeight) ->
 assert_height_delta(undefined, Height, TopHeight) ->
     Height >= ( TopHeight - gossip_allowed_height_from_top() ).
 
+assert_eoe(Node, Top) ->
+    Header = node_header(Node),
+    case aec_headers:is_eoe(Header) of
+        true ->
+            case aec_headers:type(Header) of
```
happi, line 744: Can a microblock be eoe?

apps/aecore/include/blocks.hrl
happi: Waiting for this update to finish the review.

apps/aecore/src/aec_chain_state.erl
happi: This test should not be needed if the target is set to height in an eoe block.

apps/aecore/src/aec_chain_state.erl
mitchelli: I set the eoe flag in both. I tried removing it from the micro block but for some reason in the finalize info test I am consistently getting a hole in the key block after the micro block with the vote contract call. It doesn't make sense to me, I am investigating.

happi (approved):

apps/aecore/src/aec_chain_state.erl
```diff
@@ -707,7 +707,11 @@ assert_not_illegal_fork_or_orphan(Node, Origin, State) ->
     TopHeight = node_height(Top),
     Height    = node_height(Node),
     case assert_height_delta(Origin, Height, TopHeight) of
-        true -> ok;
+        true ->
+            case assert_eoe(Node) of
+                true -> ok;
+                false -> aec_block_insertion:abort_state_transition(invalid_eoe)
```
hanssv, line 713: Is this really the right place for this check? I had guessed somewhere in aec_consensus_hc?

hanssv (commented):
I think the logic is right, but I don't like the HC-specific check right in the middle of the insertion logic for _all_ consensus types in `aec_chain_state`?

apps/aecore/src/aec_chain_state.erl
happi: There is also the call to this function:
`assert_not_illegal_fork_or_orphan(Node, Origin, State2), %% TODO: move this check to aec_conductor... it doesn't really belong here`
Just moving this to the conductor will not solve the problem that pow will also do this  check.

apps/aecore/src/aec_chain_state.erl
```diff
@@ -707,7 +707,8 @@ assert_not_illegal_fork_or_orphan(Node, Origin, State) ->
     TopHeight = node_height(Top),
     Height    = node_height(Node),
     case assert_height_delta(Origin, Height, TopHeight) of
-        true -> ok;
+        true ->
```
happi, line 710: Is this change necessary ;)

apps/aecore/src/aec_consensus_hc.erl
```diff
@@ -266,7 +266,38 @@ ctx_validate_key_target(Node, _Block, Ctx) ->
     case aec_headers:is_hole(NodeHeader) of
         true when NodeTarget == PrevNodeTarget -> ok;
         false when NodeTarget == PrevNodeTarget + 1 -> ok;
-        _ -> {error, wrong_number_of_non_holes}
+        _ ->
+            case aec_headers:is_eoe(NodeHeader) of
```
happi, line 270: Are we sure that  NodeTarget =/= PrevNodeTarget + 1 for an eoe block?

mitchelli: I know in some circumstances it is not because I made a mistake in the check_eoe function and it crashed the node but I think you're right it is safer to move the logic.

hanssv: 😅 No, rather the opposite, it will be true most of the time?!

happi (approved):

happi (approved):

hanssv (approved):

PR #4547, opened 2025-02-17
Make logging diskless when in foreground mode (-noinput)

### LINKED ISSUE #3860: bin/aeternity foreground should not write to log files

richcarl:
When the system is started using foreground mode, i.e., with the `-noinput` flag, as when running in a Docker container, we want all logging to go to the console only, and no log files or directories should be created.

Fixes #3860


richcarl:
Rebased on master

apps/aeutils/src/aeu_logging_env.erl
```diff
@@ -3,12 +3,65 @@
 -export([adjust_log_levels/0]).
 
 adjust_log_levels() ->
+    %% redirect all logs to stdout if running in foreground mode (Docker)
+    case init:get_argument(noinput) of
+        {ok, _} ->
+            case os:getenv("SYSTEM_TEST") of
+                false ->
+                        exit(system_test_not_set),
+                        set_lager_console_only();
```
hanssv, line 12: This looks weird?

richcarl: Just a temporary debugging thing

richcarl:
This is all working now, except that the system tests run in Docker, which (as intented) then redirects logging to stdout, which makes those tests not find any log files, which they expect to be able to grep through. I'm not sure what to do about that.

hanssv:
No clue, I assume @dincho had a plan for this since he wanted the change?

dincho:
> No clue, I assume @dincho had a plan for this since he wanted the change?

Absolutely, it has to be configurable and to default to console in foreground mode

dincho:
Or as an alternative, the tests should grep stdout

PR #4548, opened 2025-02-18
Staking Contract Penalty Support

klercker:
Contract level support of penalties.

- penalties (implemented as negative rewards...)
- pooling of penalties in MainStaking (slash fund)
- reported penalties (reporter get half of penalty)
- HCElection handles "rules" around penalties, Staking+Validator contracts do bookkeeping

Missing:

- what to do with pooled tokens
- actual use of the penalties


test/contracts/HCElection.aes
```diff
@@ -135,6 +137,21 @@ main contract HCElection =
...
+
+  stateful entrypoint add_penalty(height : int, penalty : int, to : address) =
+      //assert_protocol_call()
+      put(state{rewards[height = []] @ rs = (to, (-penalty)) :: rs})
+      put(state{penalty_reward_diff @ prd = prd + (penalty)})
+
+  stateful entrypoint add_reported_penalty(height : int, penalty : int, to : address, reporter : address) =
+      //assert_protocol_call()
+      put(state{rewards[height = []] @ rs = (to, (-penalty)) :: rs})
+      let reporter_reward = (penalty / 2)
```
hanssv, line 150: I assume this will become configurable 😅

hanssv (dismissed):

test/contracts/HCElection.aes
klercker: Would it be preferred to send it in as a percentage argument in the call? Then that value could be configured wherever it's used in the Erlang code.

klercker: Did exactly that.

test/contracts/HCElection.aes
```diff
@@ -135,6 +137,24 @@ main contract HCElection =
     assert_protocol_call()
     put(state{rewards[height = []] @ rs = (to, Call.value) :: rs})
 
+
+
+  stateful entrypoint add_penalty(height : int, penalty : int, to : address) =
+      //assert_protocol_call()
```
hanssv, line 143: This must change or else anyone can hand out penalties 😅

hanssv (changes requested):

test/contracts/HCElection.aes
klercker: But wouldn't that be great though?

happi (approved):

apps/aehttp/test/aehttp_hyperchains_SUITE.erl
```diff
@@ -1317,8 +1322,65 @@ check_finalize_info(Config) ->
...
+
+    mine_to_next_epoch(Node, Config),
+    {ok, AliceTot2} = inspect_validator(AliceCt, ?ALICE, get_total_balance, Config),
+    {ok, LisaTot2} = inspect_validator(LisaCt, ?LISA, get_total_balance, Config),
+    ct:log("Alice, Lisa bal: ~p ~p", [AliceTot2, LisaTot2]),
+
+    mine_to_next_epoch(Node, Config),
+    {ok, AliceTot3} = inspect_validator(AliceCt, ?ALICE, get_total_balance, Config),
+    {ok, LisaTot3} = inspect_validator(LisaCt, ?LISA, get_total_balance, Config),
+    ct:log("Alice, Lisa bal: ~p ~p", [AliceTot3, LisaTot3]),
```
hanssv, line 1353: Is there an actual test somewhere I only see reads and prints no comparison?

klercker: True, I wasn't really sure these we're going to end up existing in the end (esp since they shouldn't really run - not protocol), so I left them as they were FTM. 
Also, it's hard to get proper numbers to test against which *really* are constant sure tests. In the end - as we discusssed - the HC tests should be more about triggering true penalties. When that is implemented

PR #4552, opened 2025-02-20
Penalty contract and test fixes

klercker:
Fixes brainfart error in `add_reported_penalty()`
Moved tests from HC suite to staking ct suite
Added test access method to HCElection to enable more granualr "unit" testing of the functionality.

NOTE: This PR (and it's predecessor) only implements and tests the support for penalties - mostly in the core HC contracts. How to register them, how distribution of the fees are done etc. No actual staker/validator penalty offences are implemented here.

This PR is supported by the Aeternity Foundation

### DIFF
--- a/apps/aecontract/test/aecontract_staking_contract_SUITE.erl
+++ b/apps/aecontract/test/aecontract_staking_contract_SUITE.erl
@@ -252,25 +252,27 @@ rewards(_Config) ->
 
     ok.
 
-    sct_penalties(_Config) ->
-        Trees0 = genesis_trees(),
+    sct_penalties(_Config) -> % test MainStaking contract penalty functions
+        TreesGen = genesis_trees(),
         TxEnv = aetx_env:tx_env(?DEFAULT_HEIGHT),
         {ok, Trees1, #{res := {contract, AliceCt}}} =
-            new_validator_(pubkey(?ALICE), ?VALIDATOR_MIN, TxEnv, Trees0),
+            new_validator_(pubkey(?ALICE), ?VALIDATOR_MIN, TxEnv, TreesGen),
+        {ok, Trees11, #{res := {contract, CarolCt}}} =
+            new_validator_(pubkey(?CAROL), ?VALIDATOR_MIN, TxEnv, Trees1),
         {ok, Trees2, #{res := {contract, BobCt}}} =
-            new_validator_(pubkey(?BOB), ?VALIDATOR_MIN, false, TxEnv, Trees1, undefined),
+            new_validator_(pubkey(?BOB), ?VALIDATOR_MIN, false, TxEnv, Trees11, undefined),
 
         {ok, _, #{res := BaseEmptyPool}} = get_penalty_pool_(TxEnv, Trees2),
         ?assertEqual(BaseEmptyPool, 0),
 
-        Rewards = [{{address, pubkey(?ALICE)}, -4 * ?AE}, {{address, pubkey(?BOB)}, 2 * ?AE}],
+        Penalties1 = [{{address, pubkey(?ALICE)}, 4 * ?AE}, {{address, pubkey(?BOB)}, -2 * ?AE}], % essentially penalize Alice and reward Bob 50% of the penalty
         {ok, Trees3, #{res := ?UNIT}} =
-            add_penalties_(1, Rewards, 0, 2 * ?AE, TxEnv, Trees2, 2),
+            add_penalties_(1, Penalties1, TxEnv, Trees2, 2),
 
         {ok, _, #{res := ABalance}}   = get_total_balance_(AliceCt, ?ALICE, TxEnv, Trees3),
-        {ok, _, #{res := AAvailable}} = get_available_balance_(AliceCt, ?ALICE, TxEnv, Trees3),
+        {ok, _, #{res := AStake}} = get_current_stake_(AliceCt, ?ALICE, TxEnv, Trees3),
         {ok, _, #{res := BBalance}}   = get_total_balance_(BobCt, ?BOB, TxEnv, Trees3),
-        {ok, _, #{res := BAvailable}} = get_available_balance_(BobCt, ?BOB, TxEnv, Trees3),
+        {ok, _, #{res := BStake}} = get_current_stake_(BobCt, ?BOB, TxEnv, Trees3),
 
         {ok, _, #{res := AfterRewardsPool}} = get_penalty_pool_(TxEnv, Trees3),
         ?assertEqual(AfterRewardsPool, 2 *? AE),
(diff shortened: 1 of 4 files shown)


test/contracts/HCElection.aes
```diff
@@ -262,3 +262,10 @@ main contract HCElection =
       else
         let Some((_, haystack_)) = Bytes.split_any(haystack, 1)
         contains_bytes(needle, haystack_)
+
+  // Testing endpoints
```
dincho, line 266: 🙄

klercker: Well there's a reason the PR is in `wip` status. In order to properly unit test the contract functionality, you need to access some internal functions from test code, so this could be one of the ways to do this. If you have better suggestion how to build composable testing, ie not having to simulate the complete internet (or at least a properly running multi-node HC setup), I'm all ears.

apps/aecontract/test/aecontract_staking_contract_SUITE.erl
```diff
@@ -102,7 +103,8 @@ groups() ->
          adjust_stake,
          withdraw,
          rewards,
-         penalties,
+         sct_penalties,
+         elct_penalties,
```
hanssv, line 107: These acronyms are not obvious to me? sct? elct?

test/contracts/HCElection.aes
hanssv: The actual HCElection should live outside our tests? This is a long standing issue 😅 - the version in tests could have this function. But we need to make sure the "real" one and the test one stay in sync...

test/contracts/HCElection.aes
```diff
@@ -150,7 +150,7 @@ main contract HCElection =
       require(penalty > 0, "Penalty is expressed as a positive integer")
       require(reporter_percentage >= 0 && reporter_percentage =< 100, "reporter reward percentage must be between 0 and 100")
       put(state{rewards[height = []] @ rs = (to, (-penalty)) :: rs})
-      let reporter_reward = penalty * (100/reporter_percentage)
+      let reporter_reward = (penalty * reporter_percentage) / 100
```
ThomasArts, line 153: if anything, then (100*p) / r, right? Or did math change?
I mean, the change in code changes the computation, it's not just a rewrite?

test/contracts/HCElection.aes
ThomasArts: +1, try to compute this outside the contract and have tests that provoke payment of penalties and rewards

apps/aecontract/test/aecontract_staking_contract_SUITE.erl
ThomasArts: if they stay for staker contract and election contract, then I find it weird that you would associate penalties with contracts, they are associated to actions you perform or not perform?

apps/aecontract/test/aecontract_staking_contract_SUITE.erl
```diff
@@ -281,6 +283,39 @@ rewards(_Config) ->
...
+        {ok, Trees1, #{res := {contract, AliceCt}}} =
+            new_validator_(pubkey(?ALICE), ?VALIDATOR_MIN, TxEnv, Trees0),
+        {ok, Trees2, #{res := {contract, BobCt}}} =
+            new_validator_(pubkey(?BOB), ?VALIDATOR_MIN, false, TxEnv, Trees1, undefined),
+
+        {ok, _, #{res := BaseEmptyPool}} = get_penalty_pool_(TxEnv, Trees2),
+        ?assertEqual(BaseEmptyPool, 0),
+
+        Rewards = [{{address, pubkey(?ALICE)}, -4 * ?AE}, {{address, pubkey(?BOB)}, 2 * ?AE}],
+        {ok, Trees25, #{res := ?UNIT}} =
```
ThomasArts, line 298: I expect a test in which Alice does something bad and therefore, the result of that is that she gets a penalty.
The test must make her do something bad and the system should respond with the penalty, that then the test should observe.

apps/aecontract/test/aecontract_staking_contract_SUITE.erl
```diff
@@ -250,34 +252,114 @@ rewards(_Config) ->
...
+            new_validator_(pubkey(?CAROL), ?VALIDATOR_MIN, TxEnv, Trees1),
         {ok, Trees2, #{res := {contract, BobCt}}} =
-            new_validator_(pubkey(?BOB), ?VALIDATOR_MIN, false, TxEnv, Trees1, undefined),
+            new_validator_(pubkey(?BOB), ?VALIDATOR_MIN, false, TxEnv, Trees11, undefined),
 
         {ok, _, #{res := BaseEmptyPool}} = get_penalty_pool_(TxEnv, Trees2),
         ?assertEqual(BaseEmptyPool, 0),
 
-        Rewards = [{{address, pubkey(?ALICE)}, -4 * ?AE}, {{address, pubkey(?BOB)}, 2 * ?AE}],
+        Penalties1 = [{{address, pubkey(?ALICE)}, 4 * ?AE}, {{address, pubkey(?BOB)}, -2 * ?AE}], % essentially penalize Alice and reward Bob 50% of the penalty
```
hanssv, line 268: If we didn't like the idea of having penalites as negative rewards, I think we should avoid having rewards as negative penalties 😅

apps/aecontract/test/aecontract_staking_contract_SUITE.erl
```diff
@@ -250,34 +252,114 @@ rewards(_Config) ->
...
 
         ?assertEqual(ABalance, ?VALIDATOR_MIN - 4 * ?AE),
         ?assertEqual(BBalance, ?VALIDATOR_MIN + 2 * ?AE),
 
-        ?assertEqual(AAvailable, 6 * ?AE),
-        ?assertEqual(BAvailable, 2 * ?AE),
+        ?assertEqual(AStake, ?VALIDATOR_MIN - 4 * ?AE),
+        ?assertEqual(BStake, ?VALIDATOR_MIN),
+
+        % resditirbuting the penalty pool, i.e rewards which are taken from the penalty pool
```
hanssv, line 286:
```suggestion
        % redistributing the penalty pool, i.e rewards which are taken from the penalty pool
```

test/contracts/HCElection.aes
```diff
@@ -98,11 +100,12 @@ main contract HCElection =
     let next_reward = next_base + next_carry_over
     let pr = {current = next_reward, carry_over = next_carry_over, base = next_base}
 
+    collect_penalties(state.epoch - 1)
     // pay rewards to sunset epoch
     pay_rewards(state.epoch - 1)
 
     // update epochs
-    require(ei.start + ei.length - 1 == Chain.block_height, "This is not the end")
+    require(ei.start + ei.length - 1 == Chain.block_height, String.concats(["This is not the end: ", Int.to_str(ei.start + ei.length - 1), " : ", Int.to_str(Chain.block_height)]))
```
hanssv, line 108: This is only relevant for testing, right?

test/contracts/HCElection.aes
```diff
@@ -245,6 +240,28 @@ main contract HCElection =
...
+
+  stateful function
+    collect_penalties_(_, 0, acc) = Map.to_list(acc)
+    collect_penalties_(h, n, acc) =
+      switch(Map.lookup(h, state.penalties))
+        None      => collect_penalties_(h + 1, n - 1, acc)
+        Some(aas) =>
+          let acc1 = List.foldl(collect_penalty_, acc, aas)
+          put(state{penalties @ r = Map.delete(h, r)})
+          collect_penalties_(h + 1, n - 1, acc1)
```
hanssv, line 256: This is not super-nice, this is almost the same function again (`pay_rewards`) - but let's ignore that.

hanssv (dismissed):
I think these primitives should be enough to implement what we want in terms of penalties, some software engineering could be applied later to repeat less code.

apps/aecontract/test/aecontract_staking_contract_SUITE.erl
klercker: The actual "end-user" API is in HCElection, where the reward is expressed as percentage of the penalty. We should also note that ATM there is no way to distribute funds from the pool. On the MainStaking/internal side this is how it's done but there should be a strategy etc in HCElection I believe.

apps/aecontract/test/aecontract_staking_contract_SUITE.erl
klercker: Will do in next PR.

test/contracts/HCElection.aes
klercker: It was there before, I just added some additional info that might actually be of interest?

happi (approved):

hanssv (approved):

PR #4557, opened 2025-02-27
Mark `nonce` as required in SpendTx

davidyuk:

partially implements the same as #4539

This time I've tested locally!

### DIFF
--- a/apps/aehttp/priv/oas3.yaml
+++ b/apps/aehttp/priv/oas3.yaml
@@ -2175,7 +2175,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/SpendTx"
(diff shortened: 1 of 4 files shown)


hanssv:
Sorry, I should have looked more carefully the first time... It is a _feature_ that you can leave the `nonce` out for this endpoint... If you change this, and if the removal of this feature is intentional - please also remove the code that relies on it.

davidyuk:
> if the removal of this feature is intentional

I didn't remove anything. The idea is to use SpendTxInput (with nonce optional) in `/debug/transactions/spend` and SpendTx (with nonce required) indirectly in `/transactions/` and in mdw definitions.
Currently, both depend on SpendTx with nonce optional.

hanssv (approved):

klercker (approved):

PR #4560, opened 2025-03-14
Split voting

mitchelli:
Split the end of epoch vote, so that the fork and epoch length votes are voted on separately.

This PR is supported by the AEternity Foundation


hanssv (approved):
Looks good! This should work until there is time to refactor and restructure.

klercker (approved):

happi (approved):

PR #4587, opened 2025-04-25
fix: avoid starting multiple copies of dependencies for plugins

adas0x:
when they share deps it fails with the following error when loading multiple plugins like dev mode plugin + another

> aeu_plugins:load_plugins()-> ERROR: {badmatch,{error,same_version}}

i.e. [epbkdf2 vsn "1.0.0"] fails when loaded a second time as it's iterating through the libs it's collected.


venimus (approved):

venimus:
Thanks for finding and fixing this, @adas0x! The `{badmatch,{error,same_version}}` root cause and the dedup-by-app-name fix are correct.

This PR is stale relative to current `master` (opened against an older base, now blocked), so rather than rebase this fork PR I've opened a rebased replacement carrying your original commit/authorship intact:

- #4668 → `master`
- #4669 → `releases/stable`

Closing this one in favor of those, with credit preserved. Thanks again for the contribution!

PR #4594, opened 2025-08-17
Sort paths and components in OpenApi

davidyuk:
previous PR https://github.com/aeternity/aeternity/pull/4330


venimus:
That may probably be redone some day, for now it just seems noise and conflicts

PR #4598, opened 2026-03-17
fix: potential zero block issue patch

venimus:
This pull request introduces improvements to the block generation and mining restart logic in the conductor and block generator modules, and updates related configuration and tests. The main focus is to ensure key-block mining resumes only after the first microblock attempt completes, especially when there are no transactions, and to handle late-arriving transactions more robustly. Additionally, it updates the default for `skip_num_blocks` in the stratum configuration to minimize stale work.

Block Generation & Mining Restart Logic:

* Added `restart_mining_after_micro` and `first_micro_pending` fields to the conductor state record to track mining restart conditions and microblock status.
* Enhanced conductor event handling to delay key-block mining restart until after the first microblock attempt, including logic for empty microblocks and late-arriving transactions. [[1]](diffhunk://#diff-a40481895519f5b63a736d599c5e5a4d79df3521c356c99ca66a7cf7d3930e48R450-R498) [[2]](diffhunk://#diff-a40481895519f5b63a736d599c5e5a4d79df3521c356c99ca66a7cf7d3930e48L933-R987) [[3]](diffhunk://#diff-a40481895519f5b63a736d599c5e5a4d79df3521c356c99ca66a7cf7d3930e48L1492-R1592)
* Updated block generator to publish explicit candidate info (`empty_candidate` or `new_candidate`) and improved candidate publishing logic. [[1]](diffhunk://#diff-79b2db1916e5d0dc24a5fe6792b03e6c7e2e1f3f680426e88a07dfd630a80c68L94-R100) [[2]](diffhunk://#diff-79b2db1916e5d0dc24a5fe6792b03e6c7e2e1f3f680426e88a07dfd630a80c68L159-R158) [[3]](diffhunk://#diff-79b2db1916e5d0dc24a5fe6792b03e6c7e2e1f3f680426e88a07dfd630a80c68L240-R240)

Configuration & Documentation:

* Changed the default for `skip_num_blocks` from 10 to 0 in stratum config, schema, and documentation, recommending this for production pools to minimize stale work. [[1]](diffhunk://#diff-d3cd2b6592da49b74af4e7f80ce52bb6a82ce3cb50570f25543d6c449e04fe6aL19-R19) [[2]](diffhunk://#diff-690f49ae9dfba4976ae5844eca53cdb772079fc2a517d9eab92a1dab0f1d08c7L2559-R2559) [[3]](diffhunk://#diff-e4d4c900cf7c99c91739f4369506ce5572cbdc03361d3a3742e5b31f25427bb0L106-R107)

Testing Improvements:

* Added and updated tests to verify mining restart delay after keyblock creation and handling of late-arriving transactions, including new test cases and utility functions. [[1]](diffhunk://#diff-1c8f3f8a96dae39921f1ae950f1edc676d1dbe4ad848c42ffa185f2d614ef173L436-R440) [[2]](diffhunk://#diff-1c8f3f8a96dae39921f1ae950f1edc676d1dbe4ad848c42ffa185f2d614ef173R463-R540) [[3]](diffhunk://#diff-1c8f3f8a96dae39921f1ae950f1edc676d1dbe4ad848c42ffa185f2d614ef173L952-R1055) [[4]](diffhunk://#diff-64e44ee52f38f4dc46db4ad52ffa588e75da769a8bf5359e7d06b425347923d5R179) [[5]](diffhunk://#diff-64e44ee52f38f4dc46db4ad52ffa588e75da769a8bf5359e7d06b425347923d5R824-R850)

### DIFF
--- a/apps/aecore/src/aec_block_generator.erl
+++ b/apps/aecore/src/aec_block_generator.erl
@@ -91,14 +91,13 @@ handle_cast(start_generation, State) ->
     {noreply, do_start_generation(State)};
 handle_cast({worker_done, Pid, {candidate, Candidate, CandidateState}},
             State = #state{ worker = {Pid, _} }) ->
-    %% Only publish non-empty micro-blocks
     case aec_blocks:txs(Candidate) of
         [] ->
-            lager:debug("New empty microblock candidate generated, and discarded", []),
-            ok;
+            lager:debug("Empty microblock candidate prepared", []),
+            publish_candidate(empty_candidate);
         _  ->
-            epoch_mining:info("New microblock candidate generated", []),
-            publish_candidate(Candidate)
+            epoch_mining:info("New microblock candidate ready", []),
+            publish_candidate(new_candidate)
     end,
     State1 = finish_worker(State),
     State2 = State1#state{ candidate = Candidate
@@ -156,7 +155,7 @@ do_start_generation(S = #state{ generating = false }) ->
     S1#state{ generating = true };
 do_start_generation(S = #state{ candidate = Candidate }) ->
     %% If we are asked to start generation and already have a block, signal this.
-    [ publish_candidate(Candidate)
+    [ publish_candidate(new_candidate)
       || Candidate /= undefined andalso aec_blocks:txs(Candidate) /= [] ],
     S.
 
@@ -237,5 +236,5 @@ failed_attempt(Reason) ->
 new_candidate(NewBlock, NewBlockInfo) ->
     gen_server:cast(?MODULE, {worker_done, self(), {candidate, NewBlock, NewBlockInfo}}).
 
-publish_candidate(_Block) ->
-    aec_events:publish(candidate_block, new_candidate).
+publish_candidate(Info) ->
+    aec_events:publish(candidate_block, Info).
--- a/apps/aecore/src/aec_conductor.erl
+++ b/apps/aecore/src/aec_conductor.erl
@@ -50,6 +50,8 @@
 
 -module(aec_conductor).
 
+-define(FIRST_MICRO_RESTART_GRACE_MS, 100).
+
 -behaviour(gen_server).
 
 %% Mining API
@@ -257,6 +259,8 @@ init(Options) ->
                            aec_blocks:height(aec_chain:top_block())),
     epoch_mining:info("Miner process initialized ~p", [State5]),
     aec_events:subscribe(candidate_block),
+    aec_events:subscribe(tx_created),
+    aec_events:subscribe(tx_received),
     %% NOTE: The init continues at handle_info(init_continue, State).
     self() ! init_continue,
     {ok, State5}.
@@ -443,14 +447,55 @@ handle_cast(Other, State) ->
 handle_info({gproc_ps_event, candidate_block, _}, State = #state{consensus = #consensus{leader = false}}) ->
     %% ignore new candidates if we are not a leader any more.
     {noreply, State};
+handle_info({gproc_ps_event, candidate_block, #{info := empty_candidate}},
+            State = #state{first_micro_pending = true}) ->
+    %% The first microblock attempt for a freshly won generation completed
+    %% without any transactions, so key-block mining can resume immediately.
+    State1 = State#state{micro_block_candidate = undefined,
+                         restart_mining_after_micro = false,
+                         first_micro_pending = false},
+    {noreply, start_block_production_(State1)};
+handle_info({gproc_ps_event, candidate_block, #{info := empty_candidate}},
+            State = #state{restart_mining_after_micro = true}) ->
+    %% The first microblock attempt completed without any transactions to sign,
+    %% so key-block mining can resume.
+    State1 = State#state{micro_block_candidate = undefined,
+                         restart_mining_after_micro = false,
+                         first_micro_pending = false},
+    {noreply, start_block_production_(State1)};
+handle_info({gproc_ps_event, candidate_block, #{info := empty_candidate}}, State) ->
+    {noreply, State#state{micro_block_candidate = undefined,
+                          first_micro_pending = false}};
 handle_info({gproc_ps_event, candidate_block, #{info := new_candidate}}, State) ->
     case try_fetch_and_make_candidate() of
         {ok, Candidate} ->
-            State1 = State#state{ micro_block_candidate = Candidate },
+            State1 = State#state{ micro_block_candidate = Candidate,
+                                  first_micro_pending = false },
             {noreply, start_micro_signing(State1)};
         {error, no_candidate} ->
-            {noreply, State#state{ micro_block_candidate = undefined }}
+            {noreply, State#state{ micro_block_candidate = undefined,
+                                   first_micro_pending = false }}
     end;
+handle_info({gproc_ps_event, Event, _},
+            State = #state{consensus = #consensus{leader = true},
+                           first_micro_pending = true,
+                           restart_mining_after_micro = false})
+  when Event =:= tx_created; Event =:= tx_received ->
+    %% A transaction arrived just after we won a keyblock while keyblock mining
+    %% had already been allowed to resume. Pause that mining and finish the
+    %% first microblock attempt first.
+    State1 = kill_all_workers_with_tag(mining, State),
+    {noreply, State1#state{restart_mining_after_micro = true}};
+handle_info({gproc_ps_event, Event, _}, State)
+  when Event =:= tx_created; Event =:= tx_received ->
+    {noreply, State};
+handle_info({maybe_restart_after_first_micro, TopHash},
+            State = #state{top_block_hash = TopHash,
+                           first_micro_pending = true,
+                           restart_mining_after_micro = false}) ->
+    {noreply, start_block_production_(State)};
+handle_info({maybe_restart_after_first_micro, _TopHash}, State) ->
+    {noreply, State};
 handle_info(init_continue, State) ->
     {noreply, start_block_production_(State)};
 handle_info({worker_reply, Pid, Res}, State) ->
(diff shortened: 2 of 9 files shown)


davidyuk:
Would it break the assumption that every block takes 3 minutes? Before, if we needed to wait for N blocks for some reason, we could estimate when it would happen by adding N * 3 minutes to the current time. 
@CedrikNikita, please provide the collected stats proving the effect of this improvement

venimus:
> Would it break the assumption that every block takes 3 minutes? Before, if we needed to wait for N blocks for some reason, we could estimate when it would happen by adding N * 3 minutes to the current time. @CedrikNikita, please provide the collected stats proving the effect of this improvement

AFAIK that is not a strict rule in the first place, it is more like an average over some period and a proper ammount of mining power, that comes from the complexity (which is not changed).

apps/aecore/src/aec_block_generator.erl
```diff
@@ -91,14 +91,13 @@ handle_cast(start_generation, State) ->
     {noreply, do_start_generation(State)};
 handle_cast({worker_done, Pid, {candidate, Candidate, CandidateState}},
             State = #state{ worker = {Pid, _} }) ->
-    %% Only publish non-empty micro-blocks
     case aec_blocks:txs(Candidate) of
         [] ->
-            lager:debug("New empty microblock candidate generated, and discarded", []),
-            ok;
+            lager:debug("Empty microblock candidate prepared", []),
+            publish_candidate(empty_candidate);
```
davidyuk, line 97: so, conductor gets extra information on empty micro blocks

scripts/prove_empty_generation_fix.escript
```diff
@@ -0,0 +1,325 @@
+#!/usr/bin/env escript
```
davidyuk, line 1: Other scripts have a utility purpose, but this one is a reproduction. "/scripts/" looks to be a wrong place for it, this is more like a test.

apps/aecore/src/aec_conductor.erl
```diff
@@ -1489,27 +1543,53 @@ is_leader(NewTopBlock, PrevKeyHeader, ConsensusModule) ->
...
 
+schedule_restart_after_first_micro(State = #state{top_block_hash = TopHash}) ->
+    erlang:send_after(?FIRST_MICRO_RESTART_GRACE_MS, self(),
+                      {maybe_restart_after_first_micro, TopHash}),
+    State.
+
+should_delay_mining_restart(block_created, true) ->
+    aec_tx_pool:size() > 0;
+should_delay_mining_restart(_Origin, _RestartMining) ->
+    false.
```
davidyuk, line 1591:
```suggestion
should_delay_mining_restart(true) ->
    aec_tx_pool:size() > 0;
should_delay_mining_restart(_RestartMining) ->
    false.
```
since the first argument is always `block_created`

apps/aecore/src/aec_conductor.erl
```diff
@@ -257,6 +259,8 @@ init(Options) ->
                            aec_blocks:height(aec_chain:top_block())),
     epoch_mining:info("Miner process initialized ~p", [State5]),
     aec_events:subscribe(candidate_block),
+    aec_events:subscribe(tx_created),
+    aec_events:subscribe(tx_received),
```
davidyuk, line 263: conductor subscribes to transaction events that it doesn't strictly need

davidyuk (approved):
I like the `skip_num_blocks` change, and I would try this separately, but I'm not sure about the rest.

In this PR conductor gets extra information (empty blocks, tx created/received events) that it doesn't strictly needed. That info is used to postpone the mining until the first micro block is ready. As we found recently, blocks with 0 transactions are normal for the protocol we are using. This change will reduce the amount of them, but I don't see it as related to the sequences of key blocks with 0 transactions we are facing at the moment.

The probability of 0 blocks in our protocol depends on how frequently the mining job is updated.

I expect that miners which not applied this update will produce a bit more blocks.

davidyuk:
> Would it break the assumption that every block takes 3 minutes?

Nope, because I assume the difficulty is calculated by the time difference of mined blocks. That time will already include the introduced delay.

docs/stratum.md
```diff
@@ -103,7 +103,8 @@ For example - if you set `extra_nonce_bytes` to 1, `max_connections` shouldn't e
 (default 4)
 
 `skip_num_blocks` - (optional) - a new keyblock candidate is generated roughly every 3 seconds. `skip_num_blocks` determines how many of these candidates are skipped before a new job is sent to miner.
-(default 10)
```
ghallak, line 106: How is this change related to the issue?

davidyuk: I expected to see it in a separate commit. As I understand it will make stratum to broadcast jobs to miners more frequently. So it should reduce the number of 0 blocks and microforks' length in the right way.

hanssv:
> > Would it break the assumption that every block takes 3 minutes?
> 
> Nope, because I assume the difficulty is calculated by the time difference of mined blocks. That time will already include the introduced delay.

Initially there would be a slight increase in block-time since the mining effort is slightly delayed (seconds?). (But the adjustment should be pretty quick - a couple of dozen blocks or so??)

apps/aecore/src/aec_conductor.hrl
```diff
@@ -41,6 +41,8 @@
 -type mode() :: local_pow | stratum | pos.
 -record(state, {key_block_candidates                :: list({candidate_hash(), #candidate{}}) | 'undefined',
                 micro_block_candidate               :: #candidate{} | 'undefined',
+                restart_mining_after_micro = false  :: boolean(),
+                first_micro_pending = false         :: boolean(),
```
ghallak, line 45: I don't get what this fields do by only looking at their names. Would be nice to see something more descriptive, just like the names of other fields here.

apps/aecore/src/aec_conductor.erl
```diff
@@ -443,14 +447,55 @@ handle_cast(Other, State) ->
...
+                         first_micro_pending = false},
+    {noreply, start_block_production_(State1)};
+handle_info({gproc_ps_event, candidate_block, #{info := empty_candidate}},
+            State = #state{restart_mining_after_micro = true}) ->
+    %% The first microblock attempt completed without any transactions to sign,
+    %% so key-block mining can resume.
+    State1 = State#state{micro_block_candidate = undefined,
+                         restart_mining_after_micro = false,
+                         first_micro_pending = false},
+    {noreply, start_block_production_(State1)};
```
ghallak, line 465: Aren't the function bodies the same in both cases?

hanssv:
I'm still curious as to when and how the problems started. I wrote a quick script looking back in history, and the pattern with multiple consecutive empty generations seems to have started in early 2025 and then went away in early April 2025 (during this time only one of the major mining pools affected) and then it started again in late May 2025, now with a much more random spread over all miners. There are no related changes in the code-base as I can see, but something must have changed.

I haven't had time to look at this PR in detail, but I guess the hypothesis is that the miner gets stuck not producing micro-blocks; I would have prefered to see a test that showcase this but I'm often told I'm old-fashioned.

venimus:
> I'm still curious as to when and how the problems started. I wrote a quick script looking back in history, and the pattern with multiple consecutive empty generations seems to have started in early 2025 and then went away in early April 2025 (during this time only one of the major mining pools affected) and then it started again in late May 2025, now with a much more random spread over all miners. There are no related changes in the code-base as I can see, but something must have changed.
> 
> I haven't had time to look at this PR in detail, but I guess the hypothesis is that the miner gets stuck not producing micro-blocks; I would have prefered to see a test that showcase this but I'm often told I'm old-fashioned.

I'm far not an expert but I can say the original cause is that the last microfork is discarded from the new leader and all transactions are rewritten as a new microblock, and if leader is elected too soon it rewrites all transactions leaving a 0 block. If that repeats, all transactions are actually pushed back for some time and lower client TTL could expire (i.e. wallet or sdk) resulting to a "transaction lost complains". Yet this PR is actually not handling this particular behavior. I think Nikita could explain it better.

However I think the main reason for most 0 blocks is already described here and is visible in both testnet and mainnet, so the miner configs or misconfigurations are not to blame.
source: https://medium.com/aeternity-crypto-foundation/aeternity-bitcoin-ng-the-way-it-was-meant-to-be-df7bb1d65a4b

image

So those 33% that are moved to the next keyblock could cause the previous block to get 0 if the new one is mined under a minute
because it is not 33% of the tx, but 33% of the time between blocks (which is average 3min, so 33% is 1m) and the "moved tx" are in fact the last 33% (of the block time)

image

On the picture (Txs X and Txs Y) are the last ~30 seconds microblocks, all their tx are "stolen" and moved to Keyblock 2 as the first microblock. This would mean if the time between Keyblock 1 and Keyblock 2 is short enough, all these tx will be moved to Keyblock 2 leaving Keyblock 1 with 0

Example from TESTNET:
image

You see a 0 block 1231389, and very quickly it is the next block 1231390 (on the screenshot), time is ~40sec. And you see that the first microblock (highlighterd) includes 152 tx, which apparently are what is collected during the previous miner for the last 30s, technically all were rewritten by the next leader (as it is described by the article)

As a side-effect, if that repeats in a series of 3+ blocks, we could say that some transaction was delayed more than 3min as each new keyblock lasts 1min for example. We also discovered that low TTL causes more troubles as the transaction could get discarded and users get frustrated. Furthermore if it is a contract call that could lead to some other unexpected results (or dApp bugs).

davidyuk:
> As a side-effect, if that repeats in a series of 3+ blocks

The original issue that we are trying to solve is different. For some reason, we have really long chains of empty keyblocks. In most of them, the time difference is more than 3 minutes, so normally they shouldn't be empty.

For example, from 1268958 (16:43) to 1268997 (18:24):
39 empty blocks in 101 minutes, 2.58 min per block, mostly by 2miners, but also WoolyPooly

hanssv:
> For example, from 1268958 (16:43) to 1268997 (18:24):
> 39 empty blocks in 101 minutes, 2.58 min per block, mostly by 2miners, but also WoolyPooly

Were there micro-blocks produced and discarded in that time? Or, were there just no micro-blocks being produced? (I guess hard to see afterwards, unless you have a node and can inspect the logs?!)

venimus:
I will close this PR in favor of #4613 (though it does handle different case)

PR #4609, opened 2026-03-25
Add timeout to get_key_block_candidate gen_server call

davidyuk:
We found out that in some cases, `/v2/key-blocks/pending` takes longer than expected. The default timeout is 5 seconds. As a workaround, I propose increasing it to 60 seconds. It is intended to solve the issue of sequences of zero blocks that we see in mainnet at the moment.

Originally found by @ifaouibadi


davidyuk:
superseded by #4611

PR #4611, opened 2026-03-25
fix: increase timeout for get_key_block_candidate call in aec_conduct…

ifaouibadi:
…or module


apps/aecore/src/aec_conductor.erl
davidyuk, line 184:
```suggestion
            gen_server:call(?SERVER, {add_synced_block, Block}, 180000);
```
Should this also be increased? Since a syncing node can be slower than the one that generated a block
(30000 * (30000 / 5000))

PR #4613, opened 2026-03-26
Optimize FATE store map finalization with optimistic reuse fixpoint

ifaouibadi:
The `compute_reuse_fixpoint` algorithm in `aefa_stores.erl` performs full MPT subtree reads for store maps not yet proven reusable. For contracts with large store maps (thousands of entries), a single subtree read can take 14+ seconds — only to discover the map was safe to reuse all along.

This PR adds an optimistic fast path that skips the expensive subtree reads entirely. When all store maps end up reused in-place (the common case), the optimistic result is exact. Otherwise, it falls back to the original algorithm with a subtree cache.

**Tested against a production mainnet factory-pattern contract** with 4 large store maps, 7 contracts touched per call, and thousands of accumulated entries:

| Metric | Before | After |
|---|---|---|
| Store finalize (target contract) | 15.1s | 6ms |
| End-to-end dry-run | 14.6s | 57–83ms |

No consensus impact. Bit-identical store writes. No hard fork required.

### DIFF
--- a/apps/aefate/src/aefa_stores.erl
+++ b/apps/aefate/src/aefa_stores.erl
@@ -396,24 +396,71 @@ compute_store_updates(Metadata, #cache_entry{terms = TermCache, store = Store}) 
     {Metadata2, lists:sort(Compare, Updates)}.
 
 compute_reuse_fixpoint(Maps, Metadata, Store) ->
-    compute_reuse_fixpoint(unused_maps(Metadata), Maps, Metadata, Store, 100).
+    Unused0 = unused_maps(Metadata),
+    %% Optimistic fast path: run fixpoint using only reuse-branch refcount
+    %% adjustments (no expensive MPT subtree reads). If all store maps end up
+    %% reused in-place, the copy branch would have contributed zero adjustments
+    %% so the optimistic result is exact.
+    {UnusedOpt, ReuseOpt, _MetaOpt} =
+        optimistic_reuse_fixpoint(Unused0, Maps, Metadata, Store, 100),
+    NumStoreMaps = length([1 || {_, ?FATE_STORE_MAP(_, _)} <- maps:to_list(Maps)]),
+    case maps:size(ReuseOpt) =:= NumStoreMaps of
+        true ->
+            {RefCounts1, _} = compute_copy_refcounts(Metadata, ReuseOpt, Maps, Store, #{}),
+            Metadata1 = update_refcounts(RefCounts1, Metadata),
+            {UnusedOpt, ReuseOpt, Metadata1};
+        false ->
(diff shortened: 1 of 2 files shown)


hanssv (commented):
I don't have enough bandwidth to look at it in detail, but the idea looks sound! 

Really nice that someone finally had a chance to work on these bits! 🎉

apps/aefate/src/aefa_stores.erl
```diff
@@ -396,24 +396,71 @@ compute_store_updates(Metadata, #cache_entry{terms = TermCache, store = Store})
...
+    end.
 
-compute_reuse_fixpoint(Unused, Maps, Metadata, Store, Fuel) ->
-    Reuse      = compute_inplace_updates(Unused, Maps),
-    RefCounts1 = compute_copy_refcounts(Metadata, Reuse, Maps, Store),
-    Metadata1  = update_refcounts(RefCounts1, Metadata),
-    Unused1    = unused_maps(Metadata1),
+%% Only compute refcount adjustments for reused maps (lightweight individual lookups).
+%% Copied maps contribute zero adjustments — their subtrees are NOT read.
+compute_reuse_only_refcounts(Meta, Reuse, Maps, Store) ->
```
davidyuk: I suggest to move it below, near `compute_copy_refcounts`. Since it does similar thing

apps/aefate/src/aefa_stores.erl
```diff
@@ -396,24 +396,71 @@ compute_store_updates(Metadata, #cache_entry{terms = TermCache, store = Store})
...
+                              Count
+                      end;
+                 (_, _, Count) -> Count
+              end, #{}, Maps).
+
+%% Full fixpoint (original algorithm with subtree cache) — used as fallback.
+%% NOTE: We pass Metadata (not Metadata1) to recursive calls. Reason:
+%% compute_copy_refcounts updates refcounts assuming no inplace updates have
+%% been taken into account. So, each iteration updates the original metadata.
+full_reuse_fixpoint(Unused, Maps, Metadata, Store, Fuel, SubtreeCache) when Fuel =< 0 ->
```
davidyuk: it looks similar to `optimistic_reuse_fixpoint`, maybe a common code can be extracted

davidyuk (commented):
This should be merged separately from #4611

CedrikNikita (approved):
Looks good to me from code perspective. Test that were failing are now passing.

davidyuk (approved):
if possible, make commits' titles shorter, probably moving the rest to the commit description
also Nikita's test is unrelated, but not very important

venimus (approved):

PR #4621, opened 2026-05-07
Bound /v3/contracts/{id}/poi store-key set with keys/prefix/cursor/limit

ifaouibadi:
## Problem

`GET /v3/contracts/{id}/poi` walks the full contract store and issues a fresh root-to-leaf MPT descent for every key, so a single request against a contract with a large store (e.g. `ct_2rtXsV55jftV36BMeR5gtakN2VjcPtZa3PBURvzShSYWEht3Z7`, ~113 k keys on mainnet) ties up a Cowboy worker for 90 s and returns 504. Any unauthenticated caller can repeatedly stall the public listener.

## Resolution

Bound the per-request work by letting the caller select which store keys to include. Hard cap of 1024 keys per request, default of 10.

New optional query params:

| Param    | Default | Notes                                                              |
| -------- | ------- | ------------------------------------------------------------------ |
| `keys`   | —       | Comma-separated `ba_<base64check>` raw store keys. Max 1024.       |
| `prefix` | —       | `ba_<base64check>` raw store-key prefix (filter).                  |
| `cursor` | —       | `ba_<base64check>` raw store key to resume *after* (exclusive).    |
| `limit`  | 10      | Range `[1, 1024]`. Used with `prefix`/`cursor` and the default.    |

`keys` is mutually exclusive with `prefix` and `cursor`. No params → first 10 store keys in MPT iteration order.

Response gains three fields alongside the existing `poi`:

```json
{
  "poi":       "pi_...",
  "keys":      ["ba_...", "ba_..."],
  "truncated": true,
  "next_key":  "ba_..."
}
```

`next_key` is the last included key when truncated; pass it back as `cursor` to fetch the next page.

Existing callers that pass no params still get HTTP 200 with the same `poi` field shape — just bounded. State channels (`aesc_offchain_state`) keep the unbounded `aect_state_tree:add_poi/3` path; their state trees are tiny so the cap is not a concern.

## Performance

Measured against `ct_2rtXsV55jftV36BMeR5gtakN2VjcPtZa3PBURvzShSYWEht3Z7` (mainnet, ~113 k store keys):

| Request                              | Status | p50 latency |
| ------------------------------------ | -----: | ----------: |
| `/poi` (was 90 s timeout)            |    200 |    ~110 ms |
| `/poi?limit=3`                       |    200 |     ~30 ms |
| `/poi?cursor=<prev next_key>`        |    200 |    ~100 ms |
| `/poi?limit=1024`                    |    200 |    ~7.5 s |

## Testing

`aec_trees_tests:bounded_contract_poi_test_/0` (eunit) drives `add_contract_poi_keys/4` and `add_contract_poi_prefix/6` against a synthetic 15-key store: limit truncation, `cursor` round-trip pagination, prefix filtering, and per-key proof presence/absence.

`aehttp_integration_SUITE:get_contract_poi_query_params/1` covers the default path, explicit `prefix`/`limit`/`cursor`, mutual exc

### DIFF
--- a/apps/aecontract/src/aect_state_tree.erl
+++ b/apps/aecontract/src/aect_state_tree.erl
@@ -33,6 +33,8 @@
 
 %% API - Proof of inclusion
 -export([ add_poi/3
+        , add_poi_for_keys/4
+        , add_poi_for_prefix/6
         , verify_poi/3
         , lookup_poi/2
         ]).
@@ -62,6 +64,7 @@
 -opaque tree() :: #contract_tree{}.
 
 -define(VSN, 1).
+-define(POI_KEY_LIMIT, 1024).
 
 %% ==================================================================
 %% Tracing support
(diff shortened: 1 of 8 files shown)


venimus (dismissed):

venimus (approved):

CedrikNikita (commented):
There are small potential regressions now coming with params. I have added test to prove my point here.
https://github.com/aeternity/aeternity/tree/feature/add-contract-poi-bounded-keys

venimus:
@ifaouibadi may be worths considering this for progressing too

PR #4623, opened 2026-05-11
Add Merkle Patricia Tree read cache functionality

ifaouibadi:
## Why

`aeu_mp_trees_db` only caches **writes** (the `cache` field of `#db{}` is a
write‑through `gb_trees` that is flushed on commit). Every state lookup
during block execution therefore walks the MPT through Mnesia / RocksDB,
even for nodes the same generator has just read or just written. This PR
adds the missing read‑side cache.

## What

A new content‑addressed, shared ETS read cache that the persistent backends
in `aec_db_backends` attach to their MPT DB handles. Reads in
`aeu_mp_trees_db:get/2,3` consult the cache before falling through to
Mnesia; writes in `unsafe_write_to_backend/3` populate it.

Eviction is a **two‑segment generational LRU**: new entries land in the
young set, reads on cold entries promote them back, and when the young
counter reaches `max_size` the cold segment is dropped and the young
segment demoted — all in two `ets:select_*` calls. O(1) eviction, no
per‑read writes on the hot path.

| File | Change |
| --- | --- |
| `apps/aecore/src/aec_mpt_cache.erl` (new, 186 lines) | ETS table, generational LRU, `start/0`, `get/2`, `put/3`, `clear/0`, `enabled/0`, `size/0`, `memory/0` |
| `apps/aecore/src/aec_mpt_cache_metrics_probe.erl` (new, 93 lines) | Exometer probe for `size` / `memory` |
| `apps/aeutils/src/aeu_mp_trees_db.erl` | One optional `read_cache` field on `#db{}` + two optional callbacks. When `read_cache = undefined` every path is byte‑for‑byte unchanged. |
| `apps/aecore/src/aec_db_backends.erl` | Attaches `read_cache => tree_table_name(...)` to each persistent tree's `db_spec` **only if** `aec_mpt_cache:enabled()` is true. In‑memory / proof / map / test backends never opt in. |
| `apps/aecore/src/aecore_sup.erl` | Starts the ETS table once at boot. |
| `apps/aeutils/priv/aeternity_config_schema.json` | New `chain.mpt_read_cache.{enabled, max_size}` block. |
| `apps/aecore/priv/exometer_predefined.script` | New counters `[ae,epoch,aecore,mpt,read_cache,{hit,miss,evict}]`. |
| `apps/aecore/src/aecore.app.src` | Registers the metrics probe. |
| `apps/aecore/test/aec_mpt_cache_tests.erl` (new, 56 lines) | EUnit: round‑trip, `clear/0`, rotation. |
| `apps/aeutils/test/aeu_mp_trees_db_tests.erl` (new, 151 lines) | EUnit: hits bypass the backend; writes populate the cache; `read_cache = undefined` matches today's behaviour. |

Total: 10 files, +580 / −10.

## Config

- `chain.mpt_read_cache.enabled` — boolean, **default `true`**. Set to
  `false` (or `AE__CHAIN__MPT_READ_CACHE__ENABLED=false`) at any restart
  to fully bypass the cache code path — that is the kill switch.
- `chain.mpt_read_cache.max_size` — integer, `1000 ≤ N ≤ 10000000`,
  default `100000` entries per generation. Resident memory is bounded
  by `2 × max_size` entries.

Both can also be set through `AE__CHAIN__MPT_READ_CACHE__ENABLED` /
`AE__CHAIN__MPT_READ_CACHE__MAX_SIZE` env vars.

## Metrics

- counters: `ae.epoch.aecore.mpt.read_cache.{hit,miss,evict}`
- probe: `ae.epoch.aecore.mpt.read_cache.{size,memory}`

## Safety (no breaking change)

- **Content‑addressed keys.** MPT nodes are stored under
  `Hash = aec_hash:hash(header, aeser_rlp:encode(Value))`, so a given key
  can never map to a different value. No stale reads, no invalidation on
  write.
- **Backend deletes are not asynchronous.** `aec_db_gc` only compacts
  tables during a controlled restart, so cached entries cannot lag behind
  disk during normal operation.
- **Opt‑in at the spec layer.** Backends that don't set `read_cache` take
  the existing code path unchanged; in‑memory / proof / map / test
  backends never opt in.
- **Block‑level correctness gate.** Block insertion already verifies the
  resulting state root (`root_hash_mismatch` in `aec_chain_state`). If
  the cache ever returned wrong data, every existing test and every block
  in production would catch it.
- **Bounded memory.** ETS cap is `2 × max_size` rows; the schema enforces
  `max_size ≤ 10 000 000` (~500 MiB worst case). Empirical resident memory
  in the perf section below is ~8 MiB.
- **Kill switch.** Setting `chain.mpt_read_cache.enabled: false` (or the
  matching env var) at the next restart fully bypasses the cache code
  path — `aec_db_backends:db_spec/1` does not attach `read_cache` and
  every `aeu_mp_trees_db` call goes through the existing path
  byte‑for‑byte.

## Performance impact

Isolated MPT read microbench: 400 000 `aeu_mtrees:lookup/2` calls against
a real Mnesia / RocksDB‑backed accounts tree of 200 000 entries, 200 hot
keys, same random seed on both branches. No HTTP, no signing, no consensus
— just the read path.

| Metric | Before (`releases/stable`, cache off) | After (cache on) | Speed‑up |
| --- | ---: | ---: | ---: |
| mean per‑lookup | 137.37 µs | **12.09 µs** | **11.4×** |
| p50 | 104 µs | 11 µs | 9.5× |
| p90 | 193 µs | 14 µs | 13.8× |
| p99 | 593 µs | **20 µs** | **29.7×** |
| max | 15.70 ms | 7.02 ms | 2.2× |
| reads wall‑clock | 54.95 s | **4.83 s** | 11.4× |

Cache state at end of run: 2 487 704 hits / 296 misses / 100 000 evictions
during the prepare phase; 175 437 entries / ~8.3 MiB resident.

Real chain TPS is gated by the per‑block gas limit and microblock cadence
(protocol ceiling ~149 TPS), so wall‑clock throughput will not move by
11× from this change alone. The cache buys headroom in three concrete
places:

1. **Hot contract paths.** Workloads that revisit the same MPT subtree
   inside a microblock (DEX pools, registries, social graphs) get most
   `FATE Map.lookup` / `Map.get` reads served in ~12 µs instead of ~140 µs.
2. **Sync / fork resolution.** Re‑executing blocks against an already
   warm cache reduces catch‑up time on a freshly restarted node.
3. **Disk pressure.** Fewer RocksDB block‑cache misses and iterator
   constructions per lookup — relevant on nodes whose disk is slower than
   the protocol ceiling.

## Tests

- `aec_mpt_cache_tests` — idempotent `start`, `none` on missing key, basic
  round‑trip, `clear/0` empties the table, **rotation drops the cold
  generation**.
- `aeu_mp_trees_db_tests` — hits bypass the backend, writes populate the
  cache, `read_cache = undefined` is byte‑identical to today.
- Existing root‑hash CT suites (`aec_db_SUITE`, `aec_chain_SUITE`, …)
  remain the hard correctness gate.

## Rollout

Default‑on out of the gate. The cache is opt‑in *at the spec layer*
(only persistent tree backends attach it; in‑memory / proof / map / test
backends never do), and the only behavioural difference for operators
is the new `ae.epoch.aecore.mpt.read_cache.*` metrics in their
Prometheus / Grafana stack. If anything goes wrong post‑deploy,
`chain.mpt_read_cache.enabled: false` at the next restart restores the
previous behaviour exactly.

## Follow‑ups (intentionally out of scope)

- Sampled `lookup_time_us` and `backend_lookup_time_us` histograms.
- `bypass` counter for the `read_cache = undefined` path.
- CT integration suite (determinism, fork / reorg, GC interaction, hot
  toggle).
- PropER properties: `prop_cache_equivalence`, `prop_capacity_bound`.
- Perf CT bench with a ≥ 2× speed‑up gate in CI.
- Review the new 4‑tuple DB record format from `to_new_db/1` and the
  `source => read_cache` tag returned by `get/3`.


venimus:
## Review — bugs, gotchas & impact

The design (content-addressed, opt-in at the spec layer, generational LRU, kill switch) is sound and `aec_mpt_cache` itself is clean. There is, however, one **critical correctness bug** in how the cache interacts with online garbage collection.

### 🔴 Critical — read cache defeats GC promotion → live state can be deleted

The cache short-circuits the exact read that GC relies on to copy live nodes forward, so with `garbage_collection.enabled = true` (and the cache on by default) live trie nodes can be dropped from disk.

Mechanism:

1. GC does **online** table switching, not restart-only. On a new key block, `aec_db_gc:perform_switch/3` clears the secondary and makes it the new (empty) primary; the old primary becomes secondary. Live nodes are then lazily copied to the new primary by `promote_tree_node` inside `aec_db:lookup_tree_node` as they are read.
2. The scanner reads through a dirty backend built by `find_block_state(Hash, true)` → `deserialize_from_db` → `new_with_dirty_backend` → `aec_db_backends:dirty_accounts_backend`, whose `db_spec` **attaches `read_cache`** when the cache is enabled.
3. In `aeu_mp_trees_db:read_through/2`, a read-cache **hit returns immediately and skips `int_db_get`** — i.e. skips `mpt_db_get → aec_db:lookup_tree_node`, which is the only place `promote_tree_node` runs.
4. The cache key is `{tree_table_name, Hash}` — stable across the physical primary/secondary swap — so a node warmed before the switch is still a hit after it. It never gets promoted into the new primary, and the next `clear_secondary` wipes it. Result: `{hash_not_present_in_db, Hash}` / state corruption.

Nothing clears the cache on `perform_switch` (the only `aec_mpt_cache:clear/0` call is in the test), and the real GC scan does **not** set `db_safe_access` (only `db_safe_access_scan/1` does), so the bypass in `mpt_db_read_cache_get` does not protect it.

> The PR's Safety section is explicitly wrong on this point: it states *"aec_db_gc only compacts tables during a controlled restart, so cached entries cannot lag behind disk"* — GC compaction is online (`handle_call(#maybe_garbage_collect{top_change=true})` → `perform_switch`). That false assumption is likely how this slipped through.

Fix options: clear (or invalidate the affected tree) in `perform_switch`; or force a cache bypass while a scanner is active; or drive GC promotion through a path that ignores the read cache.

### 🟠 Medium

- **Metrics on the hot path.** `bump/2` calls `aec_metrics:try_update_or_create` on *every* hit and miss, which resolves to `exometer:update_or_create` (lookup + maybe-create), not a plain `update`. At millions of reads/block this erodes the ~12 µs the cache is buying. Consider the sampling already noted in the follow-ups before default-on.
- **No benefit (only cost) under `db_safe_access`.** With `chain.db_safe_access = true` globally, every `mpt_db_read_cache_get/put` bypasses, so the cache does zero work but still pays the extra check and holds the table. Worth an ops-docs note.

### 🟡 Minor / nits

- **`source => read_cache` in `get/3`** omits the `promoted` key that the GC map path sets. Safe today only because the GC scanner uses the arity-3 `visit_count/3` → `db_get/2` path; any future arity-4 visitor consuming `promoted` would `badkey`. (Already in the follow-up list.)
- **Rotation under concurrency** — multiple schedulers can both cross the threshold and double-rotate, and gen-2 promotions bump the young counter without adding an entry. All benign (cache-efficiency only, content-addressed so never wrong), but the "bounded by 2×max_size" claim has small concurrency slack.
- **`to_new_db({db, M, H, C})`** upgrade clause — field mapping is correct; fine as defensive hot-upgrade handling.

### Impact

- **Correctness:** critical risk **only when GC is enabled** (`garbage_collection.enabled`, default `false`, but commonly turned on to reclaim disk). Cache defaults to `true`, so upgrading such a node silently arms the bug. Non-GC nodes are safe — content-addressing keeps cached reads correct, and block-level root-hash checks guard the rest.
- **Performance:** strong win on repeated MPT reads (per the microbench); real TPS still gated by the protocol ceiling. Per-hit metrics add measurable hot-path cost.
- **Memory:** ~2×max_size rows (≈ up to ~500 MiB at schema max; ~8 MiB typical).
- **Ops:** new `ae.epoch.aecore.mpt.read_cache.*` metrics; clean kill switch via `chain.mpt_read_cache.enabled: false` at restart.

**Recommendation:** block on the GC-promotion issue before merge (or ship default-off until it's resolved), and address the per-hit metrics cost.

venimus (commented):
## Re-review of `401575f` "fix(mpt-cache): stop the cache breaking state GC"

Re-reviewed against the critical GC-promotion bug flagged in my previous review.

### ✅ The critical bug is correctly fixed, not just patched around
Traced the full path:
- `aec_db:new_tree_context/2` selects `#tree_gc{}` for any GC-enabled tree in **both** dirty and transaction mode — so the GC scanner's dirty-backend reads (the exact trigger of the original defect) get `#tree_gc{}` too, not just regular transactional reads.
- New `aec_db:tree_context_cache_key/1` returns `none` for `#tree_gc{}`, `{ok, Table}` for plain `#tree{}`.
- `aec_db_backends:mpt_db_get/2` checks `read_cache_tab/1` first; on `none` it falls through to the **same unconditional** `aec_db:lookup_tree_node/2` call that existed pre-cache — so `promote_tree_node/4` always fires for GC trees.
- The new `aec_db_gc_SUITE:mpt_cache_gc_promotion_test/1` is the right regression test for this: drives real read traffic through a live GC switch with the cache enabled, asserts cache stats stay at `{entries:0, hits:0, misses:0}` under GC (i.e. structurally unreachable, not just "happens to be empty"), then runs `db_safe_access_scan` across all 8 state trees to confirm nothing reachable was lost.

Verdict: this closes the defect at the right layer (the backend callback boundary), not via a narrower workaround.

### Bonus: two of my earlier "minor" findings turned out to be worse than estimated, and both got fixed here
- **Concurrent rotation**: commit message states two concurrent triggers previously emptied the whole table — "200k of 200k entries gone." I'd filed this as minor concurrency slack; it was a full-wipe bug. Now single-flighted through the owning gen_server, with `concurrent_load_survives` covering it.
- **Memory gauge**: now counted in bytes including tracked refc binary payload, not ETS words. Matches my "don't use for capacity planning" note — now actually fixed instead of just documented as a caveat.

### New fix not in the original PR — worth calling out
`aec_mpt_cache` is now a `temporary` supervisor child (was `permanent`). `aecore_sup` runs with `MaxRestarts=0`, so a permanent cache child would have taken the whole node down on any cache-internal fault. Good catch, this was a real availability risk in the original version.

### Default flipped to `enabled: false`
Confirmed in the config schema ("Experimental, disabled by default"). This reverses my earlier deployment-risk concern about the old `enabled: true` default silently combining with GC — but it also means enabling the cache anywhere now requires an explicit `enabled: true`, not the other way around.

### ⚠️ One thing worth weighing before merging/rolling out
The commit message includes a real-world measurement:

> Measured on 24 mainnet micro-blocks: 1.02x. Cheap blocks get 4.6x, but the contracts that dominate rewrite their whole subtree per call and invalidate their own entries, so there is nothing to reuse where it would matter.

That's a large downgrade from the ~11x microbench number in the original PR description. The dominant real-world workload (heavy contract calls rewriting their own subtree per call) self-invalidates the cache on every call — which is exactly the traffic pattern behind the CPU-saturation issue this cache was hoped to help with on our `api_main` nodes. Worth setting expectations accordingly: this is very unlikely to move the needle on that specific problem. The durable fix there remains gas repricing for state access, not caching.

No blocking concerns from my side on correctness/safety of this fix. The effectiveness caveat above is a product/priority question for the team, not a defect.

PR #4629, opened 2026-07-07
Feat/protocol v7 salus

ifaouibadi:


venimus:
Flagging a possible protocol-version-number collision before this goes further: this branch defines `?SALUS_PROTOCOL_VSN, 7` in `apps/aecontract/include/hard_forks.hrl` (based on `releases/stable`), but `master` (7.3.x) has **already claimed protocol version 7 for a different hardfork**: `?ARCUS_PROTOCOL_VSN, 7` in `aec_hard_forks.erl`/`hard_forks.hrl` on `master` (used for the not-yet-public hyperchains work).

Since `master` is the eventual merge target for everything in `releases/stable` (mdw itself runs the master/7.3.x line), landing Salus as protocol 7 on `releases/stable` and Arcus as protocol 7 on `master` means one of them will have to be renumbered (almost certainly Salus → 8, since Arcus/hyperchains looks further along and already merged on master) before these two lines converge. If both ship independently as "protocol 7" on their respective branches without reconciling the number, we'd end up with two incompatible hard forks claiming the same consensus version — any node syncing across both lines, or any tooling (SDK, explorers) hardcoding `protocol == 7`, would misidentify which ruleset is actually active.

Worth resolving explicitly (either renumber Salus to 8 now, or confirm there's a reconciliation plan) rather than discovering it at merge time.

venimus:
@ifaouibadi ok Since there is Arcus v7 already defined as a placeholder protocol, I suggest to rebase this PR on master, then use the existing Arcus protocol for the hard-fork, then bump Salus as potential next (v8) protocol placeholder.

venimus:
since this branch is going to be deprecated in favor of master, should be closed, all changes already applied on master

PR #4630, opened 2026-07-08
Add a wall-clock timeout to the dry-run endpoint

ifaouibadi:
The dry-run endpoints simulate a transaction against a state snapshot without committing it. A pathological transaction — for example a call against a very large contract store — could run for a long time and tie up a request worker.

Bound each dry-run to a wall-clock limit: 50% of the microblock cycle (1.5s by default), configurable via http.dry_run.timeout_ms. On timeout the run is killed and the caller gets a clean error instead of hanging; the worker also stops if the client disconnects, so nothing is orphaned.

This only affects the off-chain dry-run path (aec_dry_run, reached from the HTTP and Rosetta API). It is never on the block-production or validation path, so it cannot affect consensus.

### DIFF
--- a/apps/aeapi/src/aeapi.erl
+++ b/apps/aeapi/src/aeapi.erl
@@ -513,7 +513,7 @@ format_txs(Txs, MBHash) ->
             _ ->
                 {in, MBHash}
             end,
-    case aec_dry_run:dry_run(Top, [], DryTxs, [tx_events]) of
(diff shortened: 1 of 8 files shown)


venimus (approved):

davidyuk:
The idea was to limit the external api dry-runs by gas, it is more reliable than limiting them by time in ms
https://github.com/aeternity/aeternity/blob/36d12bff9f614d813bc28f05e43a33acba8054da/apps/aeutils/priv/aeternity_config_schema.json#L1304-L1308

PR #4632, opened 2026-07-20
Avoid redundant prev-node DB reads during block insertion

CedrikNikita:


venimus (approved):
LGTM — reuses the already-fetched `PrevNode` (via `ctx_prev(Ctx)`) instead of a second `db_get_node(node_prev_hash(Node))` lookup for `prev_version/1`. Semantics-preserving, saves a redundant DB read per block insertion.

PR #4633, opened 2026-07-20
Avoid redundant tx hash recomputation in aec_tx_pool

CedrikNikita:


venimus (approved):
LGTM — mechanical change, threads the already-computed tx hash into `pool_db_key/2` instead of recomputing it via `aetx_sign:hash/1` at each call site. Same key shape, no behavior change.

PR #4634, opened 2026-07-20
Avoid unnecessary peer lookups in gossip event handler

CedrikNikita:


venimus (approved):
LGTM — `gossip_peer_ids()` is now only computed for the branches that actually use it, avoiding a wasted `aec_peers:get_random_connected/1` call on the "just produced our own block" path that previously computed it unconditionally. Same result on all paths that use it.

PR #4635, opened 2026-07-20
Thread precomputed header hash into conductor block insertion

CedrikNikita:


venimus (commented):
Nice reduction in redundant hash computation. One thing worth double-checking before merge, given how central block identity is to chain-state correctness: `wrap_block/2` now trusts the `Hash` threaded in from the conductor rather than recomputing it via `aec_headers:hash_header/1`. If that precomputed hash were ever stale/mismatched (e.g. a future refactor upstream of `handle_add_block` computes it from slightly different bytes), the block would get silently indexed under the wrong key with no error raised.

Would it be worth adding either (a) a debug/test-mode assertion that the threaded-in `Hash` equals a freshly computed `aec_headers:hash_header(Header)`, or (b) at least a unit test asserting `insert_block_conductor/3` with a deliberately-wrong hash is caught? Not blocking, just flagging since this is the one change in the recent optimization batch that touches block-identity trust rather than pure redundant-computation removal.

venimus:
Checked how this would apply to `master` (7.3.x — the line mdw runs, per its `node_version: 7.3.0-rc7`). Not a clean cherry-pick: master's `handle_add_block` is consensus-module-aware for hyperchains (`aec_consensus_hc`, "HOLE" block logging) and still calls the old 2-arg `insert_block_conductor(Block, Origin)`. This PR's 3-arg `insert_block_conductor(Block, Hash, Origin)` would need to be threaded through that branch too — small, mechanical, but not a blind port.

Didn't open a master PR for this one yet since it touches block-identity trust (see my earlier comment) and I'd rather see that resolved on `releases/stable` first. Happy to do the master port once this settles, just flagging the conflict now so it's not a surprise later.

venimus (approved):

PR #4636, opened 2026-07-20
fix: delay key-block mining restart until first microblock completes

venimus:
## Summary

Reintroduces the conductor/block-generator changes from #4598 ("fix: potential zero block issue patch", closed 2026-03-27 without merging), rebased on current `releases/stable` and validated against fresh live-network evidence.

## What changed vs. the original #4598

The original PR bundled several changes together. This PR carries forward **only** the core fix:
- `aec_conductor.erl` / `aec_conductor.hrl`: delays a leader's restart of its own key-block PoW search until the first microblock attempt for a freshly-won generation has completed (bounded to a single microblock-candidate cycle via `first_micro_pending`/`restart_mining_after_micro`, gated on `aec_tx_pool:size() > 0`).
- `aec_block_generator.erl`: publishes an explicit `empty_candidate` vs `new_candidate` event so the conductor can distinguish "no pending work" from "a candidate is ready."
- `aec_conductor_tests.erl`: the two targeted unit tests from the original PR (`test_delay_restart_until_first_micro_block`, `test_delay_restart_for_late_tx`).
- `scripts/prove_empty_generation_fix.escript`: the reproduction/validation script from the original PR.

**Intentionally excluded** from this PR (unrelated to this fix, and one includes a reduction in test coverage we don't want to reintroduce):
- Changes to `aec_dry_run.erl` and removal of `aec_dry_run_tests.erl` / `aec_dry_run_heavy_tests.erl`.
- Changes to `aefa_stores.erl`.
- The `stratum.skip_num_blocks` default change (10→0) — this was a separate, already-approved concern in the original review thread and should be proposed on its own.

## Why resubmit now

In the original review, @davidyuk asked for supporting data showing this fix actually addresses the observed "long chains of empty keyblocks," and noted that zero-transaction keyblocks are expected/normal in this protocol in isolation — his concern was whether this change addresses the *specific pattern* of long, repeated streaks.

Since then we've gathered direct live evidence on mainnet (2026-07-20) of exactly that pattern: extended sequences of consecutive keyblocks with zero microblocks/transactions, followed by a single keyblock absorbing an unusually large backlog once one finally completes — consistent with transactions being repeatedly rebuilt into new microblock candidates that never get signed before being superseded.

Worth noting: this doesn't require many independent competing miners to occur. A **single pool that repeatedly wins consecutive rounds** can exhibit this same pattern purely through its own leader-restart timing racing against its own microblock-signing — i.e., it's a self-preemption effect, not necessarily a cross-pool race. This can happen with any number of active pools on the network, including a small number of them, whenever consecutive winning solutions land with the same pool. We verified the pending-tx-pool condition (`aec_tx_pool:size() > 0`) that gates this fix's delay logic was genuinely satisfied throughout the observed streak, so the fix's trigger condition matches the real-world scenario we saw.

## Credits
Original implementation: @CedrikNikita. Original PR: @venimus. Original review: @davidyuk, @hanssv.

### DIFF
--- a/apps/aecore/src/aec_block_generator.erl
+++ b/apps/aecore/src/aec_block_generator.erl
@@ -91,14 +91,13 @@ handle_cast(start_generation, State) ->
     {noreply, do_start_generation(State)};
 handle_cast({worker_done, Pid, {candidate, Candidate, CandidateState}},
             State = #state{ worker = {Pid, _} }) ->
-    %% Only publish non-empty micro-blocks
     case aec_blocks:txs(Candidate) of
         [] ->
-            lager:debug("New empty microblock candidate generated, and discarded", []),
-            ok;
+            lager:debug("Empty microblock candidate prepared", []),
+            publish_candidate(empty_candidate);
         _  ->
-            epoch_mining:info("New microblock candidate generated", []),
-            publish_candidate(Candidate)
+            epoch_mining:info("New microblock candidate ready", []),
+            publish_candidate(new_candidate)
     end,
     State1 = finish_worker(State),
     State2 = State1#state{ candidate = Candidate
@@ -156,7 +155,7 @@ do_start_generation(S = #state{ generating = false }) ->
     S1#state{ generating = true };
 do_start_generation(S = #state{ candidate = Candidate }) ->
     %% If we are asked to start generation and already have a block, signal this.
-    [ publish_candidate(Candidate)
+    [ publish_candidate(new_candidate)
       || Candidate /= undefined andalso aec_blocks:txs(Candidate) /= [] ],
     S.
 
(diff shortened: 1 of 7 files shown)


venimus:
Added a `mining.delay_restart_after_micro` config flag (default `true`) so miners who prefer to keep hunting for the next key block continuously — accepting the self-preemption risk — can opt out and get byte-identical pre-patch behavior (`false`). Restart delay is otherwise not free for the miner: it's a real pause in that node's own PoW search (min ~100ms, unbounded up to the first microblock's build time when the mempool is non-empty), so this shouldn't be forced on operators who'd rather not take that trade-off.

venimus:
Closing this draft in favor of #4639, which reimplements the same self-preemption delay-restart mechanism and is further along. Leaving it to @CedrikNikita to carry forward from here rather than maintaining two parallel copies.

One thing worth folding in from this version if useful: an opt-out config flag (`mining.delay_restart_after_micro`, default `true`) so operators who'd rather keep mining continuously (accepting self-preemption risk over pausing PoW search) can restore byte-identical old behavior with `false` — happy to help wire that into #4639 if wanted.

venimus:
Resurrected against `master` (not a cherry-pick — `aec_conductor.erl` diverged significantly since, HC/PoS support) in #4675.

PR #4637, opened 2026-07-20
fix: bound wall-clock time spent building a microblock candidate

venimus:
## Summary

Bounds the wall-clock time spent building a microblock candidate. Currently, a leader pulls pending transactions into a candidate and applies them with no time limit — a slow/expensive transaction (or several) can consume the whole microblock cycle, risking the candidate being preempted before it's ever signed. Complementary to #4636 (self-preemption restart delay): that addresses *when* a leader restarts its own key-block search, this addresses unbounded candidate *construction* itself.

## How it works

- `int_pack_block/8` in `aec_block_micro_candidate.erl` now takes a `Deadline` (65% of `micro_block_cycle()`). Once passed, it stops pulling in more pending transactions and signs whatever has already been assembled; anything left waits for the next microblock cycle.
- Per-round gas requested from `aec_tx_pool:get_candidate/3` is also capped (`CANDIDATE_ROUND_GAS_CAP`), since a single call can otherwise return every eligible transaction up to the full `block_gas_limit()` in one shot on the first round, before the deadline check ever gets a chance to fire. Capping the requested gas forces smaller batches and more frequent deadline checks.
- A capped round returning empty falls back to a single uncapped attempt (still bounded by the same deadline check), so a transaction whose own declared gas exceeds the cap isn't starved forever.
- `get_candidate/3`'s own selection already walks pending transactions in per-account nonce order; only the requested gas budget is shrunk, the returned list is never sliced, so nonce ordering is preserved automatically.
- Mempool transaction selection order is left unchanged (fee-priority) — no reordering/prioritization logic was added, to avoid altering fee-market semantics.

This is opened as a draft pending running the eunit suite (no full rebar3 build available in the environment used to prepare this) and adding dedicated test coverage.

### DIFF
--- a/apps/aecore/src/aec_block_micro_candidate.erl
+++ b/apps/aecore/src/aec_block_micro_candidate.erl
@@ -98,10 +98,36 @@ int_create(PrevBlock, KeyBlock) ->
         {ok, Trees} ->
             TxEnv = create_tx_env(MBEnv),
             MaxGas = aec_governance:block_gas_limit(),
-            int_pack_block(MaxGas, [], [], MBEnv, TxEnv, Trees, []);
+            Deadline = candidate_deadline(),
+            int_pack_block(MaxGas, [], [], MBEnv, TxEnv, Trees, [], Deadline);
         error ->
             {error, block_state_not_found}
     end.
+
+%% Bound the wall-clock time spent pulling in more pending transactions while
+%% building a microblock candidate, so a slow/expensive transaction (or a long
+%% tail of them) cannot indefinitely delay signing a microblock and risk total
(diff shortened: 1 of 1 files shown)


venimus:
Closing in favor of #4628, which addresses the same problem (a slow/expensive transaction stalling microblock candidate generation) with a more targeted and better-tested approach: it identifies and quarantines the specific offending transaction (via in-flight tracking + preempt-kill detection) with a TTL-bounded re-eligibility window, rather than this PR's blanket per-round gas cap. #4628 has significantly more thorough test coverage for the relevant edge cases.

PR #4639, opened 2026-07-20
fix: delay key-block mining restart until first microblock attempt

CedrikNikita:
After winning a key-block the conductor restarted key-block mining immediately. If the next key-block was then found quickly, the freshly won generation ended up without any microblocks (a "zero block") and the pending transactions only appeared in the next leader's first microblock.

Now, when the winning node is the leader:
- the block generator publishes an explicit empty_candidate / new_candidate event instead of silently discarding empty candidates, so the conductor knows when the first microblock attempt completed;
- key-block mining resumes only once the first microblock attempt of the new generation has completed, or after a short grace period when the tx pool is empty;
- a transaction arriving during that grace period pauses key-block mining again until the first microblock attempt completes.

Cleaner version of #4598

### DIFF
--- a/apps/aecore/src/aec_block_generator.erl
+++ b/apps/aecore/src/aec_block_generator.erl
@@ -91,14 +91,13 @@ handle_cast(start_generation, State) ->
     {noreply, do_start_generation(State)};
 handle_cast({worker_done, Pid, {candidate, Candidate, CandidateState}},
             State = #state{ worker = {Pid, _} }) ->
-    %% Only publish non-empty micro-blocks
     case aec_blocks:txs(Candidate) of
         [] ->
-            lager:debug("New empty microblock candidate generated, and discarded", []),
-            ok;
+            lager:debug("Empty microblock candidate prepared", []),
+            publish_candidate(empty_candidate);
         _  ->
-            epoch_mining:info("New microblock candidate generated", []),
-            publish_candidate(Candidate)
+            epoch_mining:info("New microblock candidate ready", []),
+            publish_candidate(new_candidate)
     end,
     State1 = finish_worker(State),
     State2 = State1#state{ candidate = Candidate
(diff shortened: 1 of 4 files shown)


venimus:
This looks like the same fix we ported as #4636 (itself a cleanup of #4598) — delaying key-block mining restart until the first microblock attempt completes.

One difference: #4636 added an opt-out config flag (`mining.delay_restart_after_micro`, default `true`) so operators who'd rather keep mining continuously (accepting self-preemption risk over pausing PoW search) can restore byte-identical old behavior with `false` — the delay isn't free for a miner (min ~100ms grace, unbounded up to full microblock-build time when the mempool is non-empty).

Might be worth reconciling the two so we don't duplicate review/merge effort — happy to close whichever one loses out, or fold the flag into this version if you'd prefer this implementation. Let us know which way you'd like to go.

venimus:
Checked how this would apply to `master` (7.3.x — the line mdw runs). Not a clean cherry-pick, and more involved than #4635's conflict:

- `aec_block_generator.erl`: master already has slightly different wording/behavior for discarding empty candidates (doesn't yet publish `empty_candidate`/`new_candidate` events at all), so the `worker_done` handling here needs real reconciliation, not a mechanical merge.
- `aec_conductor.erl`: master's `preempt_on_new_top`/`ResetWorkers` logic has been restructured for hyperchains — it now branches on `Mode` (`pos` vs `_pow`) with a dedicated PoS path, replacing the PoW/stratum-only branching this PR (and our own #4636, same mechanism) assumes. The delay-restart logic would need to be integrated into that `pos`/`_pow` split deliberately.

Same applies to our own #4636 — we're holding off on porting it to master until this stabilizes on `releases/stable`, specifically because of this hyperchains-branching conflict. Wanted to flag it here too since whichever of #4639/#4636 lands will hit the same wall going to master.

PR #4640, opened 2026-07-21
Encode gossip wire message once per block instead of once per peer

CedrikNikita:


venimus (approved):
LGTM — the wire message is content-identical per peer for a fixed protocol version, so encoding it once and reusing the bytes via `do_send`'s existing fragmentation logic is safe. Good win for peer-count/large-block scaling (serialization cost drops from O(peers) to O(1) per gossip).

PR #4641, opened 2026-07-21
Use account-ordered nonce index for account-scoped mempool queries

CedrikNikita:


venimus (commented):
Great perf idea and nice test coverage on the new nonce-index path. One thing to flag before merge: this changes result *ordering*, not just performance — `pool_db_peek/3` for an account used to return in fee order (same as the general mempool order), and now returns in **nonce order** (explicitly called out in your own comment: "Results come out in nonce order rather than the fee order of peek/1").

Could you confirm nothing external relies on the old fee-ordering for account-scoped peeks (e.g. any HTTP/GA-facing endpoint that surfaces this to wallets/aepps/explorers, or docs promising an order)? If it's purely internal/undocumented ordering, this is a non-issue — just want to make sure it's a deliberate, reviewed behavior change rather than an incidental side effect of the index swap.

CedrikNikita:
> Could you confirm nothing external relies on the old fee-ordering for account-scoped peeks (e.g. any HTTP/GA-facing endpoint that surfaces this to wallets/aepps/explorers, or docs promising an order)? If it's purely internal/undocumented ordering, this is a non-issue — just want to make sure it's a deliberate, reviewed behavior change rather than an incidental side effect of the index swap.

I have checked it myself that no aepps are depending on the fact that previously it was fee-ordered.

PR #4642, opened 2026-07-21
Read only the header of the previous block for type lookups

CedrikNikita:


venimus (approved):
LGTM — `aec_headers:type/1` on the header is equivalent to `aec_blocks:type/1` for this purpose, and avoids deserializing a full micro-block (all its txs) just to check key/micro type on hot HTTP paths (`GetTopBlock`, generation endpoints, etc.). Nice, low-risk win.

PR #4643, opened 2026-07-21
Cache several variables in the fate engine state

CedrikNikita:
Isolated microbenchmark (20M iterations each, real records via the actual constructors, in the CI builder container):

| Accessor | Old path | New path | Saved | Speedup |
|---|---|---|---|---|
| `consensus_version/1` | 7.64 ns/call | 6.15 ns/call | **1.49 ns** | 1.2x |
| `in_auth_context/1` | 8.36 ns/call | 5.89 ns/call | **2.47 ns** | 1.4x |

Call frequency, traced with `dbg` across the real `aefate_engine_test` + `aefa_sophia_test` suites (216 tests spanning arithmetic, tuples, patterns, records, variants, maps, higher-order calls, remote calls — 1,390 FATE instructions executed total):

- `consensus_version/1`: 443 calls → **0.32 calls per instruction**
- `in_auth_context/1`: 22 calls → **0.016 calls per instruction**

**Combined estimate:** weighted average savings ≈ `0.32 × 1.49ns + 0.016 × 2.47ns` ≈ **0.51 nanoseconds per FATE instruction executed**.

### DIFF
--- a/apps/aefate/src/aefa_engine_state.erl
+++ b/apps/aefate/src/aefa_engine_state.erl
@@ -138,7 +138,11 @@
             , stores            :: aefa_stores:store()
             , trace             :: list()
             , vm_version        :: non_neg_integer()
+              %% Read from the tx_env once, in new/7. Only values that are fixed
+              %% for the lifetime of the engine state belong here - the tx_env is
+              %% not immutable as a whole, primops keep appending events to it.
             , consensus_version :: non_neg_integer()
+            , in_auth_context   :: boolean()
             , debug_info        :: debug_info()
             }).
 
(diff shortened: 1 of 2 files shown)


venimus (approved):
LGTM — caching `consensus_version`/`in_auth_context` in the engine state at `new/7` relies on both being fixed for that engine state's lifetime. The new `simple_contract_call_spend` test is good evidence this holds (inner meta-tx execution gets its own fresh engine state rather than mutating a shared one), so the cache shouldn't go stale mid-execution. Thanks for adding that regression test alongside the optimization.

PR #4644, opened 2026-07-21
Serialize signed tx once in aetx_sign:serialize_for_client

CedrikNikita:
## Benchmark: `aetx_sign:serialize_for_client` double-serialization fix

Measured in the `ae-builder` container (OTP 26), N=20000 iterations per case, warmed up, against the **old** code (2× serialize) to isolate the redundant work.

| Case | 1× `serialize_to_binary` | Full `serialize_for_client_pending` (old, 2×) | Estimated new (1×) | Gain |
|---|---|---|---|---|
| spend tx, empty payload, 1 sig | 4.0 µs | 28.7 µs | 24.7 µs | 14.0% (1.16×) |
| spend tx, 2KB payload, 1 sig | 5.3 µs | 49.5 µs | 44.3 µs | 10.6% (1.12×) |
| spend tx, empty payload, 3 sigs | 3.2 µs | 39.1 µs | 35.9 µs | 8.2% (1.09×) |

**Verdict: real but modest — roughly a 1.1× speedup on this function, not 2×.**

The description "halves tx serialization" is literally true (2 calls → 1 call to `serialize_to_binary`), but that redundant call is only ~10-14% of what `serialize_for_client`/`serialize_for_client_pending` actually costs. The remaining ~85-90% goes into `aetx:serialize_for_client(Tx)` (per-field base58/api encoding for the human-readable client map) plus `aeser_api_encoder:encode` on the hash and tx binary — raw binary serialization is cheap next to that.

**Absolute impact:** ~3-5 µs saved per transaction rendered. For a block-transactions endpoint, that's roughly `N_txs × 4µs` — tens of µs for a typical block, up to ~1 ms for a large block near the gas limit. Not something that moves a p99 latency dashboard, but a correct, zero-risk elimination of wasted work with no downside.

### DIFF
--- a/apps/aetx/src/aetx_sign.erl
+++ b/apps/aetx/src/aetx_sign.erl
@@ -71,7 +71,11 @@ new(Tx, Signatures) ->
 
 -spec hash(signed_tx()) -> binary().
 hash(#signed_tx{} = Tx) ->
-    aec_hash:hash(signed_tx, serialize_to_binary(Tx)).
+    hash_from_binary(serialize_to_binary(Tx)).
+
+-spec hash_from_binary(binary_signed_tx()) -> binary().
+hash_from_binary(TxBin) ->
+    aec_hash:hash(signed_tx, TxBin).
 
 -spec add_signatures(signed_tx(), list(binary())) -> signed_tx().
 add_signatures(#signed_tx{signatures = OldSigs} = Tx, NewSigs)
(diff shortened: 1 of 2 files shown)


venimus (approved):
LGTM — good, thorough test coverage (`aetx_sign_tests.erl`) explicitly asserting the returned hash/bytes match a from-scratch serialization. Appreciate the honest benchmark numbers (~10-14%, not overclaiming the "halves serialization" framing in the description).

PR #4651, opened 2026-07-22
[master port] Read only the header of the previous block for type lookups (cherry-pick of #4642)

venimus:
Master (7.3.x) port of #4642, clean cherry-pick (no conflicts against current `master`).

mdw runs the `master`/7.3.x line (not `releases/stable`), so this perf fix doesn't reach mdw until it's ported here too. Note: once this merges, mdw itself will need a corresponding node update/rebuild to actually pick up the change — porting the code alone isn't enough.

Original PR: #4642


venimus:
Closing in favor of the composite master port #4655, which bundles this and the other 7 clean cherry-pick PRs into a single PR to review/merge instead of 8 separate ones.

PR #4653, opened 2026-07-22
[master port] Use account-ordered nonce index for account-scoped mempool queries (cherry-pick of #4641)

venimus:
Master (7.3.x) port of #4641, clean cherry-pick (no conflicts against current `master`).

mdw runs the `master`/7.3.x line (not `releases/stable`), so this perf fix doesn't reach mdw until it's ported here too. Note: once this merges, mdw itself will need a corresponding node update/rebuild to actually pick up the change — porting the code alone isn't enough.

Also flagging (see comment on #4641): this change returns account-scoped mempool `peek/2,3` results in nonce order instead of the previous fee order — worth confirming that's fine before merging either copy.

Original PR: #4641

### DIFF
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -26,8 +26,9 @@
         { %% Tuple of arity 1 where the single element is the mempool key tuple.
          ?KEY(NegFee, NegGasPrice, Origin, Nonce, TxHash)
         }).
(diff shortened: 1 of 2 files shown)


venimus:
Closing in favor of the composite master port #4655, which bundles this and the other 7 clean cherry-pick PRs into a single PR to review/merge instead of 8 separate ones.

PR #4657, opened 2026-07-22
Deserialize gossiped txs independently instead of all-or-nothing

### LINKED ISSUE #4601: Gossip Batch Transaction Poisoning
### Description

A malicious peer can cause an entire batch of gossiped transactions to be silently dropped by including a single malformed serialized transaction in the batch. The `handle_new_txs` function in `aec_peer_connection.erl` uses an all-or-nothing list comprehension for deserialization. One failure in `aetx_sign:deserialize_from_binary` aborts the entire batch, and the broad `catch _:_ -> ok` silently swallows the error.

### Attack Path

1. Attacker establishes a Noise P2P connection to the target node
2. Constructs a `txs` message with N-1 valid serialized transactions + 1 malformed binary
3. Target node receives in `handle_new_txs` (line 1417)
4. List comprehension at line 1420 fails on the malformed tx — `aetx_sign:deserialize_from_binary` throws
5. `catch _:_ -> ok` at line 1424 catches the exception
6. All N-1 valid transactions are silently dropped — never reach `tx_push` or the tx pool

### Impact

Per-batch transaction censorship. Under normal network conditions, transactions arrive from multiple peers, limiting impact. Under an eclipse attack (attacker controls all peer connections), this enables persistent selective transaction censorship.

### Proof of Concept

```erlang
%% aec_peer_connection.erl:1417-1427
handle_new_txs(S, Msg) ->
    try
        #{ txs := EncSignedTxs } = Msg,
        %% ALL-OR-NOTHING: one bad tx kills the entire batch
        SignedTxs = [ aetx_sign:deserialize_from_binary(Tx) || Tx <- EncSignedTxs ],
        spawn(fun() -> [ tx_push(SignedTx) || SignedTx <- SignedTxs ] end)
    catch _:_ ->
        ok  %% SILENT — no logging, no metrics, no peer penalty
    end,
    S.
```

### Mitigation

Replace the list comprehension with per-element error handling:

```erlang
handle_new_txs(S, Msg) ->
    try
        #{ txs := EncSignedTxs } = Msg,
        SignedTxs = lists:filtermap(
            fun(Tx) ->
                try {true, aetx_sign:deserialize_from_binary(Tx)}
                catch _:_ ->
                    epoch_sync:info("Failed to deserialize tx from ~p", [peer_id(S)]),
                    false
                end
            end, EncSignedTxs),
        spawn(fun() -> [ tx_push(SignedTx) || SignedTx <- SignedTxs ] end)
    catch _:_ -> ok
    end,
    S.
```

---

venimus:
Fixes #4601.

`handle_new_txs/2` deserialized a gossiped `txs` batch via a single list comprehension over `aetx_sign:deserialize_from_binary/1`, wrapped in a blanket `catch _:_ -> ok`. `deserialize_from_binary/1` has no internal try/catch and throws (`badmatch`) on a malformed binary, so one bad/malformed entry aborted the whole comprehension — silently dropping every other (valid) tx in that batch, with no logging or peer penalty. Unchanged since 2018.

Fix: deserialize each tx independently via `lists:filtermap/2`, dropping only the malformed entry (with a debug log) instead of the entire batch.

Low risk / minimal diff, no behavior change for the all-valid case.

### DIFF
--- a/apps/aecore/src/aec_peer_connection.erl
+++ b/apps/aecore/src/aec_peer_connection.erl
@@ -1413,7 +1413,9 @@ send_send_tx(S = #{ status := {connected, _ESock} }, SerTx) ->
 handle_new_txs(S, Msg) ->
     try
         #{ txs := EncSignedTxs } = Msg,
-        SignedTxs = [ aetx_sign:deserialize_from_binary(Tx) || Tx <- EncSignedTxs ],
+        %% Deserialize each tx independently - a single malformed/malicious
+        %% entry must not cause the whole batch to be discarded.
+        SignedTxs = lists:filtermap(fun deserialize_tx/1, EncSignedTxs),
         %% Offload the handling to a temporary worker process - it might block
         %% on the tx_pool_push queue.
         spawn(fun() -> [ tx_push(SignedTx) || SignedTx <- SignedTxs ] end)
@@ -1421,6 +1423,13 @@ handle_new_txs(S, Msg) ->
         ok
     end,
     S.
+
+deserialize_tx(Tx) ->
+    try {true, aetx_sign:deserialize_from_binary(Tx)}
+    catch _:_ ->
+        lager:debug("Failed to deserialize gossiped tx, dropping just this entry", []),
+        false
+    end.
 
 tx_push(STx) ->
     try aec_tx_pool:push(STx, tx_received)
--- a/apps/aecore/src/aec_peer_connection.erl
+++ b/apps/aecore/src/aec_peer_connection.erl
@@ -46,6 +46,7 @@
 
 -ifdef(TEST).
 -export([is_node_info_sharing_enabled/0]).
+-export([deserialize_tx/1]).
 -endif.
 
 -include("aec_peer_messages.hrl").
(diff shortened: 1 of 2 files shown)


venimus:
**Performance impact note**

Switching from a single list-comprehension + one `try/catch` around the whole batch to `lists:filtermap/2` + a per-element `try/catch` (`deserialize_tx/1`) has negligible overhead on the happy path (all txs well-formed):
- Same O(n) single pass over the batch either way.
- Entering a `try/catch` that doesn't throw is essentially free on the BEAM (just pushes a catch frame — cost isn't proportional to work done). The extra per-element catch only has a real cost when a tx actually fails to deserialize, which is exactly the malicious/malformed case this fix targets.
- `lists:filtermap/2` calling a named fun per element vs. an inlined comprehension body adds a fixed, tiny per-call overhead, immaterial next to the actual binary-parsing cost of `aetx_sign:deserialize_from_binary/1`.
- The new `lager:debug/2` call only fires on the malformed-entry path, and is a no-op when debug logging isn't enabled (the production default).

Net: no measurable regression for the normal case; for the previously-broken case (one bad tx in a batch) it now correctly processes the other N-1 valid txs instead of silently discarding all of them, at the cost of one caught exception + one debug log per bad entry. This is a static/code-level assessment, not a benchmarked measurement.

PR #4675, opened 2026-07-28
fix: delay key-block mining restart until first microblock completes (master port)

venimus:
Resurrects #4636 ("fix: delay key-block mining restart until first microblock completes", closed 2026-07-22 as it targeted `releases/stable` and had drifted from `master`), reapplied here directly against current `master` — not a cherry-pick, since `aec_conductor.erl`/`aec_block_generator.erl` have diverged substantially since (HC/PoS consensus support). Full technical detail in the first comment below.

Original implementation: @CedrikNikita. Original PR: @venimus. Original review: @davidyuk, @hanssv.

### DIFF
--- a/apps/aecore/src/aec_block_generator.erl
+++ b/apps/aecore/src/aec_block_generator.erl
@@ -111,10 +111,10 @@ handle_cast({worker_done, Pid, {candidate, Candidate, CandidateState}},
     case aec_blocks:txs(Candidate) of
         [] ->
             lager:debug("Empty microblock candidate prepared", []),
-            ok;
+            publish_candidate(empty_candidate);
         _  ->
             epoch_mining:info("New microblock candidate ready", []),
-            publish_candidate(Candidate)
+            publish_candidate(new_candidate)
     end,
     State1 = finish_worker(State),
     State2 = State1#state{ candidate = Candidate
@@ -171,7 +171,7 @@ do_start_generation(S = #state{ generating = false }) ->
     S1#state{ generating = true };
 do_start_generation(S = #state{ candidate = Candidate }) ->
     %% If we are asked to start generation and already have a block, signal this.
-    [ publish_candidate(Candidate)
+    [ publish_candidate(new_candidate)
       || Candidate /= undefined andalso aec_blocks:txs(Candidate) /= [] ],
     S.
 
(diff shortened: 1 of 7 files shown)


venimus:
## Why not a cherry-pick

`#4636` targeted `releases/stable`. Since then, `aec_conductor.erl`/`aec_block_generator.erl` gained substantial HC/PoS-aware restructuring on `master` (module doc rewrite, `preempt_on_new_top` now short-circuits `mode =:= pos` into its own `{pos, State1}` branch before ever reaching the delay logic, `setup_loop`'s clause structure, etc.). A cherry-pick would have conflicted badly and risked reintroducing PoW-only logic into the now-PoS-aware code paths, so this reapplies the same fix by hand against master's current structure.

## What the fix does

Under PoW, `aec_conductor` used to eagerly restart key-block mining as soon as a new key block was won, racing against `aec_block_generator`'s first microblock-candidate attempt for that generation. If a transaction was pending, this could self-preempt the in-flight microblock before it ever got signed — a **single pool repeatedly winning consecutive rounds** can reproduce this purely through its own restart timing, no competing miners required. Confirmed on mainnet (2026-07-20): long streaks of empty key blocks followed by one absorbing an unusually large backlog.

The fix delays a leader's own key-block restart until the first microblock attempt for a freshly-won generation completes (bounded to a single cycle via `first_micro_pending`/`restart_mining_after_micro`, gated on `aec_tx_pool:size() > 0`), and only applies to the PoW (`_pow`) branch of `preempt_on_new_top` — PoS is unaffected, since it already bypasses this logic entirely (time-slot-based candidate creation, no busy-loop restart to race against).

New `mining.delay_restart_after_micro` config (default `true`) allows opting back into the old, continuous-restart behavior.

## Validation
- `aec_conductor_tests`: 23/23 passed (incl. the 2 new tests from the original PR: `test_delay_restart_until_first_micro_block`, `test_delay_restart_for_late_tx`), run against the `aeternity/builder:focal-otp26` container.
- `scripts/prove_empty_generation_fix.escript initial_tx`: `PROOF_STATUS=passed` against current master.

Credits unchanged from #4636: original implementation @CedrikNikita, original PR @venimus, original review @davidyuk @hanssv.

PR #4679, opened 2026-07-30
Bound and speed up micro block candidate selection

CedrikNikita:


venimus (changes requested):
the tests seems green, but this pr should be rebuild/rebased on master after salus gets branched on master

venimus:
Reviewed the selection rewrite in depth. The core premise is right: a transaction queued ahead of the chain is *early, not wrong*, and offering it only to have the apply fail — then charging it a failure for that — was the wrong answer. Following each sender's nonce sequence is the correct fix, and the rewritten `skipped_nonce_tx_is_not_charged_for_being_early` proves the good case end to end, including that both transactions land once the gap is filled.

One substantive concern, then some smaller points.

## Removing the failure charge also removes the only bound on far-future-nonce mempool growth

The failure charging that this PR stops was, in practice, the only mechanism evicting far-future gossiped transactions. Three things combine here:

* `nonce_check_by_event(tx_received) -> false`, so `mempool.nonce_offset` is not applied to gossiped transactions at all — there is no ingress bound on how far ahead of the chain a sender may queue.
* `?DEFAULT_TX_TTL` is `20 * 24 * 2 * 7` key blocks (~2 weeks), against the old eviction after a handful of micro-block attempts.
* There is no mempool size cap — neither in `aec_tx_pool` nor in the config schema.

So the retention window for transactions that cannot be mined goes from seconds to roughly two weeks, with no ceiling on how many are held. Since they are never mined, no fee is ever paid for them, so there is no cost to the sender either.

The release note is honest about this ("a mempool given such traffic will hold more than it did before"), and `mempool.tx_ttl` is correctly named as the remaining bound. But that is disclosure rather than mitigation, and it sits awkwardly against the PR's own goal: the timeout work bounds one DoS vector, and this widens another. Given the PR is otherwise a hardening change, I would rather settle it here than inherit it.

Options that would close it, roughly in order of how invasive they are:

1. Apply a nonce-distance bound at ingress for `tx_received` as well — possibly a larger, separate limit from `nonce_offset`, since that one is deliberately tight for locally created transactions.
2. Keep the shorter TTL for transactions that have never once been selectable, so an early transaction still ages out faster than a valid one waiting on a gap.
3. A mempool size cap with eviction ordered by fee and distance ahead of the chain.

(3) is clearly out of scope for this PR. (1) or (2) look proportionate. If the preference is to ship as-is and track the bound separately, that works too, provided it is an explicit decision with an issue behind it rather than something the release note absorbs.

## Verified and correct

Noting these so the next reader does not have to re-derive them — several looked like plausible failure modes and are not:

* `account_nonce_entries/3`'s match spec matches the `insert_nonce/2` shape (`{{Account, Nonce, TxHash}, {NegFee, NegGasPrice}}`), which this PR leaves unchanged, and sorting on `NegFee` does give best-paying-first. A mismatch here would have silently `mark_done`'d whole accounts.
* `get_account/3` memoises on pubkey only, not on `Source`. That is safe: the accounts tree is fixed for a whole `get_candidate/4` call, selection never applies anything, and the cache is rebuilt per call from `Acc0`.
* `walk_reaches/3` correctly serves entries the mempool-proper walk has already passed instead of deferring them to a walk that will not return.
* `mark_done/2` persisting across both table passes is safe, because `serve_nonces/11` already resolves each nonce through both tables via `resolve_pool_entry/2`.
* The "never empty a block, run on to the block time instead" rule is implemented consistently in both `int_pack_block/9` and `deadline_expired/1`, and `create(TopHash, infinity)` is the right call for the on-demand and common-tests consensus paths.

## Smaller points

**`micro_block_candidate_timeout` schema description.** It is around 150 words covering clamping, the overrun-by-one-batch caveat, and Hyperchains block time. Schema descriptions should be one sentence; the rest already exists in the release note and belongs in `docs/configuration.md`. This has come up on other schema entries recently, so worth trimming for consistency.

**`build_deadlines/2` writes application env.** Repairing an invalid value with `application:set_env/3` from inside a candidate build is a global side effect from a worker process. The intent — log once per node rather than once per build — is reasonable, but a `persistent_term` flag or a one-shot guard would express that without the global write, and would not leave the running config differing from what the operator set.

**Release note.** The filename `PR-mining-micro-block-candidate-timeout.md` does not match the existing convention in `docs/release-notes/next/` (`GH-<issue>-...` or a plain descriptive name). The note itself is also essay-length; the content is accurate and the behavioural changes are all disclosed, but it is hard to extract "what changed" from it on a first read. Worth tightening to a few bullets with the operator-facing consequences up front.

venimus (approved):

PR #4586, opened 2025-04-25
Block order fixes and PoF disabling

klercker:
- Fix bug coming from block order swap in PoS
- Disable PoF and PogF in PoS

PR #4680, opened 2026-07-31
Reject Hyperchains HOLE/EOE header flags under PoW consensus

CedrikNikita:
#4676 restored zero-enforcement on the reserved key-header flag bits but deliberately preserved the Hyperchains HOLE (bit 29) and EOE (bit 28) bits, ungated. On a PoW chain those two bits are still reserved: 7.2.2 rejects any key header with them set as malformed_header, while master accepts it. A miner who grinds valid PoW with either bit set therefore produces a block that master nodes accept and 7.2.2 nodes reject - an unguarded chain split on live Ceres. Nothing downstream closes the gap; validate_key_block_header/2 checks only protocol, seal and max time.

Gate acceptance on consensus type, matching how #4586 gated PoF/PogF: the bits are honoured under pos and rejected under pow, in both the wire binary deserializer and the client-JSON flags path. Gating on protocol version would be wrong - Hyperchains deployments run on Ceres as well as Arcus, so a
>= ARCUS gate would break them. ae_mainnet and ae_uat hardcode
aec_consensus_bitcoin_ng and cannot be configured into pos, so mainnet can never opt in.

The common case (both bits clear) short-circuits before the consensus lookup, which is a cached persistent_term read.

Converts the #4676 round-trip test into a meck fixture, since eunit runs under local_<protocol>_testnet, which resolves to pow for every protocol, so the unmocked test would now correctly expect rejection. Pins both directions: rejection under pow for each bit and for the client path, acceptance under pos for the binary path (HOLE has no setter, so the bit is tampered in) and for the client path. The client pos case is also the control for the client pow case - deserialize_from_client catches every exception and reports invalid_header regardless of cause, so without it the rejection test would pass vacuously on a malformed fixture.

### DIFF
--- a/apps/aecore/src/aec_headers.erl
+++ b/apps/aecore/src/aec_headers.erl
@@ -683,12 +683,28 @@ deserialize_key_flags(KeyBlock) ->
             Decoded = decode(bytearray, Flags),
             %% Client-supplied flags must not set any unused/reserved bit.
             <<F:?FLAG_BITS>> = Decoded,
-            AllowedMask = ?KEY_HEADER_FLAG bor ?CONTAINS_INFO_FLAG bor ?HOLE_FLAG bor ?EOE_FLAG,
+            Base = ?KEY_HEADER_FLAG bor ?CONTAINS_INFO_FLAG,
+            AllowedMask = case hc_flags_allowed() of
+                              true  -> Base bor ?HOLE_FLAG bor ?EOE_FLAG;
+                              false -> Base
+                          end,
             case F band (bnot AllowedMask) of
                 0 -> Decoded;
                 _ -> error(illegal_flags)
             end
     end.
+
+%% The HOLE and EOE key-header flag bits are Hyperchains constructs, only
+%% meaningful under PoS consensus. On a PoW chain they are reserved bits that
+%% must be zero, exactly as 7.2.2 requires: a PoW-sealed key block with either
+%% bit set would otherwise be accepted here and rejected as malformed_header by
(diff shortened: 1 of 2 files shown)


venimus (commented):
This looks like the right fix for the Hole/EOE-under-PoW chain-split issue. Compared against #4681, which targets the same problem:

- This PR's `hc_flags_allowed/0` gates purely on `aec_consensus:get_consensus_type/0`, matching the existing precedent set by #4586 (PoF/PoGF gating), which uses the identical pattern (`case aec_consensus:get_consensus_type() of pow -> ...; pos -> ... end`) with no protocol-version fallback.
- #4681 additionally OR's in `Version >= ?ARCUS_PROTOCOL_VSN` as a forward-compatibility path. Checked this against `aec_hard_forks.erl`: Arcus is commented out of every real network's protocol table (mainnet, uat, ...) - the only place it's wired up is a synthetic `local_arcus_testnet` id used for internal testing, never reachable on any live network today. The actual next-protocol work is on a branch named `feat/protocol-v7-salus`, so the real next hard fork's version number isn't even settled yet, which makes gating on `ARCUS_PROTOCOL_VSN` specifically fragile/premature rather than a genuine forward path.
- Test coverage here (mocking `get_consensus_type` via meck, covering pos/pow x hole/eoe x binary/client paths) looks complete and targeted at exactly the boundary that matters.

Recommendation: go with this PR. I don't see anything in #4681 worth porting over here - its point of difference from this PR is exactly the part we're recommending against (see comment posted there).

PR #4681, opened 2026-07-31
Gate Hole/EOE header flags on consensus type, not protocol version

ifaouibadi:
#4676 (merged) restored strict zero-enforcement on the reserved header bits and closed the micro-header path completely, but still accepts bits 29/28 (Hole/EOE) on every protocol. On a PoW chain those bits carry no meaning and a v7.2.x node rejects any header that sets them, so accepting one makes this node follow a chain such a node can never follow, at any depth, however much work accumulates -- a split for the price of one mined block.

Measured in a 3-node lab before the gate: an unmodified master node followed a Hole-flagged producer onto a separate chain while v7.2.2 stayed on its own, both still mining -- the fleet silently partitions rather than halting. Measured after: the node sides with v7.2.2 and orphans the flagged blocks.

The discriminator is the CONSENSUS TYPE, not the protocol version. Arcus is defined but activated on no network (commented out of every aec_hard_forks table), so latest_protocol_version() is Ceres and Hyperchains legitimately produces Hole/EOE blocks AT CERES. A 'protocol >= Arcus' gate alone would therefore have broken Hyperchains. The protocol check is kept as the forward path for when Arcus activates.

#4676's hole_and_eoe_flags_roundtrip_test asserted a successful round-trip of EOE on a pre-Arcus header -- which a PoW node must now refuse -- so it is re-pointed at an Arcus header, with refusal asserted across all six earlier protocols and on the client-JSON path.

### DIFF
--- a/apps/aecore/src/aec_db.erl
+++ b/apps/aecore/src/aec_db.erl
@@ -408,6 +408,8 @@ encoding(aec_chain_state, _, _) -> {sext, {object, term}};
 encoding(_Tab, set, 2) -> {raw, {value, term}};
 encoding(_Tab, set, _) -> {raw, {object, term}}.
 
+%% Do not bump until check_table_vsn/4 is fixed: its mismatch branch badmatches, so a bump blocks
+%% the upgrade direction rather than the downgrade one.
 tab_vsn(_) -> 1.
 
 tab_type(_) -> set.
--- a/apps/aecore/src/aec_headers.erl
+++ b/apps/aecore/src/aec_headers.erl
@@ -683,7 +683,11 @@ deserialize_key_flags(KeyBlock) ->
             Decoded = decode(bytearray, Flags),
             %% Client-supplied flags must not set any unused/reserved bit.
             <<F:?FLAG_BITS>> = Decoded,
-            AllowedMask = ?KEY_HEADER_FLAG bor ?CONTAINS_INFO_FLAG bor ?HOLE_FLAG bor ?EOE_FLAG,
+            AllowedMask =
+                case hc_flags_permitted(maps:get(<<"version">>, KeyBlock)) of
+                    true  -> ?KEY_HEADER_FLAG bor ?CONTAINS_INFO_FLAG bor ?HOLE_FLAG bor ?EOE_FLAG;
+                    false -> ?KEY_HEADER_FLAG bor ?CONTAINS_INFO_FLAG
+                end,
             case F band (bnot AllowedMask) of
                 0 -> Decoded;
                 _ -> error(illegal_flags)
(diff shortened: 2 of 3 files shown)


venimus (commented):
#4680 targets the same chain-split issue (Hole/EOE header flags accepted under PoW consensus) with a simpler fix: it gates purely on `aec_consensus:get_consensus_type/0`, matching the existing precedent set by #4586 (PoF/PoGF gating), which uses the identical pattern with no protocol-version fallback.

This PR additionally OR's in `Version >= ?ARCUS_PROTOCOL_VSN` as a forward path. Checked this against `aec_hard_forks.erl`: Arcus is commented out of every real network's protocol table (mainnet, uat, ...) and only wired up under a synthetic `local_arcus_testnet` id used for internal testing - it isn't reachable on any live network today. The actual next-protocol work is on a branch named `feat/protocol-v7-salus`, so the real next hard fork's version number isn't settled either, which makes gating on `ARCUS_PROTOCOL_VSN` specifically feel premature/fragile rather than a solid forward-compatibility path.

Also flagging: this diff includes an unrelated `aec_db.erl` (`tab_vsn`) comment change that looks out of scope for this fix.

We're going to go with #4680 for this one - happy to revisit an Arcus/Salus forward path as its own separate follow-up once that protocol's real semantics and version number are actually decided, rather than baking in an assumption about it now.

venimus:
Follow-up on the Arcus-gate concern above, worth being explicit about since it's more than a style preference:

`aec_hard_forks:protocols_from_network_id/1` falls through to `default_protocols_from_network_id/0` (reading `chain.hard_forks` from user config) for any network ID other than the hardcoded `ae_mainnet`/`ae_uat`/`local_*_testnet` ones. Any custom network can therefore declare its genesis directly at any protocol version, completely independent of consensus type - `aehc_demo` itself does exactly this today, starting straight at Ceres (v6) with `height: 0`, skipping all prior history.

That means `hc_flags_permitted(Version) -> Version >= ?ARCUS_PROTOCOL_VSN orelse get_consensus_type() =/= pow` would incorrectly permit Hole/EOE flags on a **custom PoW network** configured to start at protocol >= 7 for reasons unrelated to Hyperchains entirely (e.g. someone wanting newer FATE/VM features on a private PoW testnet). That reopens the exact chain-split condition this PR sets out to close, for any such custom deployment - not just a hypothetical future mainnet scenario. #4680's pure consensus-type check doesn't have this gap, since it never looks at protocol version at all, so it's correct regardless of which version a custom network chooses to start at.

This reinforces going with #4680 as-is.

venimus:
closed in favor of #4680

PR #4682, opened 2026-07-31
Fix HC producer deadlock after long production stalls

venimus:
## Problem

Found while recovering a real hyperchain (`aehc_demo`) that had stopped producing blocks for
~5 months after both validators went down. After fixing the underlying db-corruption issue that
originally took the validators down, block production still did not resume, with no crash and no
error logs (default `warning` log level only surfaces the relevant path at `lager:debug`).

Root cause is in the interaction between `aec_consensus_hc:get_slot_info/2` and `next_producer/0`:

```erlang
ElapsedBlocks = (Now - T0) div BlockTime,
Delta = if Epoch == 1 -> ElapsedBlocks + 1; true -> ElapsedBlocks end,
Slot = First + Delta,
```

`Delta` is derived from **real wall-clock time** elapsed since the current epoch's first block,
uncapped. After a production gap longer than one epoch's real-time span (`child_epoch_length *
child_block_time`), the projected `Slot` exceeds the epoch's `last` slot.

`next_producer/0` then calls `leader_for_timeslot(Slot, RunEnv)`, which (correctly, since the
validator schedule for a not-yet-reached epoch can't exist) returns `{error, epoch_did_not_end}`
once `Timeslot > EpochLast`. `next_producer/0` turns that into `{error, not_in_cache, BlockTime div
5}`, and `aec_conductor:hc_next_producer/1` just retries every ~600ms forever.

This is a genuine deadlock: production needs the epoch to close to advance to a resolvable slot,
but the epoch can only close by producing its last block — which never happens because the
projected slot is unresolvable. The existing hole-block catch-up mechanism (`leader_hole` /
`hc_create_hole`) never even engages, because it's gated behind the same failing lookup.

This is a known, previously-documented limitation, not a new discovery: PR #4556 ("Additional
tests for eoe-holes and stuck chain", merged Feb 2025) states explicitly *"the CC can recover if
we are less than 1 child epoch behind"* — i.e. exactly this boundary. `state_pre_transform_node/4`
even has a matching `%% TODO: fix timings, prevent busy loop` for the symmetric case of a block
somehow arriving past `EpochLast`.

**Note**: there is a second, separate, still-open failure mode covered by the existing
`epochs_with_slow_parent` test/comments — a stuck/slow *parent* chain can prevent an epoch's last
block from finalizing at all (`get_entropy_hash` unavailable → `parent_chain_not_synced`). This PR
does **not** address that case; it's a different code path (`state_pre_transform_node`'s
`Height =:= EpochLast` branch) and remains a known limitation.

## Fix

The wall-clock-projected `Slot` stays **uncapped** in `get_slot_info/2`. Instead, `next_producer/0`
now handles `{error, epoch_did_not_end}` explicitly: `own_slot_after_stall/2` scans the
passed-but-unproduced slots — bounded by `Last + Length`, i.e. the current epoch plus the
already-resolvable next-epoch spill-over, stopping at the first uncached slot — for the earliest
slot **this node** holds the leader key for, and produces there; the existing hole-block mechanism
fills the gap. Nodes without a leader key in range return `{wait, not_producer}` instead of
hot-looping `not_in_cache`.

Recovery therefore does not depend on any *specific* leader being alive: whichever producer comes
back first takes its own earliest slot and unsticks the chain, epoch by epoch. Once the epoch's
`last` slot is produced, the normal end-of-epoch transition (`step_key/4`) advances to the next
epoch, and the process repeats until the chain catches up to real time.

An earlier iteration of this PR instead clamped the projected slot to the epoch's `last`
(`Slot = min(First + Delta0, Last)`). That trades one deadlock for another: if the last slot's
leader is itself offline (or excluded, as in the `hole_production` test), every other producer
waits forever on a slot only that leader can fill. The `hc_hole` CT group reproduced this
deterministically, which is why the clamp was replaced by the per-node own-slot fallback. More
generally, any fallback to a single fixed slot has this failure mode — the fallback must be
keyed to what *this* node can actually produce.

## Why this doesn't need a fork/protocol bump

- `get_slot_info/2` / `next_producer/0` are **pure producer-side liveness logic** — they only
  decide *when/whether this node attempts to create a candidate*. They are not part of block
  validation.
- The blocks a fixed node produces while catching up are ordinary hole blocks — already a
  supported, validated block type (`aec_block_hole_candidate`), accepted from any on-schedule
  producer (`is_leader_valid_for_hole`) by unpatched peers the same way as any other hole.
  Validation logic (`leader_for_timeslot/2` at the validation call sites in
  `dirty_validate_micro_node_with_ctx`/`state_pre_transform_node`) is untouched by this change.
- A mixed network of patched/unpatched nodes stays consensus-compatible: unpatched nodes simply
  keep retrying `not_in_cache` (their pre-existing behavior) until a patched node produces the
  catch-up blocks, at which point they sync normally. Freshly syncing nodes replay the recovery
  region deterministically from chain state (the wall-clock path is never consulted during sync).

## Impact / gotchas (please review carefully)

- **Catch-up volume for very long stalls is large and is now unconditional.** For `aehc_demo`
  specifically: `child_epoch_length: 600`, `child_block_time: 3000ms` → 30 minutes/epoch. A ~5
  month stall requires ~4.7M hole blocks across ~7800 epochs to fully catch up to real time. Each
  hole still runs `aec_chain_state:calculate_state_for_new_keyblock` (real state-tree work, not a
  no-op), so catching up is not instantaneous and will measurably grow chain size/db. This fix
  makes the chain *able* to recover instead of being deadlocked forever, but recovery from a
  months-long stall could still take a long time and produce a large volume of empty blocks. For
  a real incident like this, operators may still prefer a state-reset/manual epoch fast-forward
  over waiting for millions of holes — this PR only removes the hard deadlock, it doesn't make
  catch-up instantaneous.
- **Does not address the slow-parent-chain deadlock** (see `epochs_with_slow_parent`) — that's a
  separate mechanism/gate and is out of scope here.
- During recovery, competing producers may briefly race on different own-slots and produce
  short-lived forks; these resolve through normal fork choice (observed and handled in the CT
  runs — see the reorg-safe retry accumulation below).
- No config schema or contract changes — this is a pure Erlang logic fix in the node's own
  producer path.

## Testing

The full `hc_hole` CT group passes (`--suite=apps/aehttp/test/aehttp_hyperchains_SUITE
--group=hc_hole`), including:

- `hole_production` / `hole_production_eoe` — previously deadlocked deterministically under the
  clamp approach (the excluded node owned the epoch's last slot); now pass.
- `production_recovers_after_long_stall` (added by this PR) — stops mining on all nodes, sleeps
  past 2 full child epochs of real wall-clock time, resumes mining, asserts the chain makes
  progress, then runs a further full epoch to confirm the chain stays healthy. This test had
  previously always been auto-skipped (it runs last in the sequence), which hid three latent
  bugs in it that are also fixed here: full node names passed to `rpc/4` (double-expanded by
  `node_name/1` → `{badrpc, nodedown}`), an `{ok, _}` match on `produce_n_epochs/2` (returns
  bare `ok`), and mining restarted before `hc_mine_blocks/4` (which asserts it stopped on entry).

Test-helper hardening in the same commit: `produce_cc_blocks_with_retry` accumulates blocks
across timed-out attempts by re-fetching the whole range from the settled top (recovery forks can
reorg away heights an earlier attempt reached, so stitching per-attempt prefixes raced), and it
mines a few parent blocks before each retry to break the circular wait between next-epoch entropy
and child progress.

This PR is supported by findings from a live-incident recovery on `aehc_demo`.

### DIFF
--- a/apps/aecore/src/aec_consensus_hc.erl
+++ b/apps/aecore/src/aec_consensus_hc.erl
@@ -997,16 +997,24 @@ get_slot_info(0, _RunEnv) ->
     Now = aeu_time:now_in_msecs(),
     #{epoch => 1, slot => 1, t0 => Now, delta => 0, now => Now};
 get_slot_info(_TopHeight, RunEnv) ->
-    {ok, EpochInfo = #{epoch := Epoch, first := First}} = aec_chain_hc:epoch_info(RunEnv),
+    {ok, EpochInfo = #{epoch := Epoch, first := First, last := Last}} = aec_chain_hc:epoch_info(RunEnv),
     BlockTime     = child_block_time(),
     Now = aeu_time:now_in_msecs(),
     T0  = get_t0(Epoch, First),
 
     ElapsedBlocks = (Now - T0) div BlockTime,
-    Delta = if Epoch == 1 -> ElapsedBlocks + 1;
-               true       -> ElapsedBlocks
-            end,
-    Slot = First + Delta,
+    Delta0 = if Epoch == 1 -> ElapsedBlocks + 1;
+                true       -> ElapsedBlocks
+             end,
+    %% Never project a slot beyond the current on-chain epoch's last slot. The
+    %% validator schedule for a future epoch can't be resolved until this one
+    %% is actually closed on-chain (leader_for_timeslot/2 returns
+    %% `epoch_did_not_end` past `Last`), so an uncapped wall-clock Delta after
+    %% any production gap longer than one epoch's real-time span would make
+    %% next_producer/0 retry `not_in_cache` forever instead of catching up via
+    %% the existing hole-block mechanism one epoch at a time.
+    Slot = min(First + Delta0, Last),
+    Delta = Slot - First,
     #{epoch => Epoch, slot => Slot, t0 => T0, delta => Delta,
       now => Now, epoch_info => EpochInfo}.
 
--- a/apps/aecore/src/aec_consensus_hc.erl
+++ b/apps/aecore/src/aec_consensus_hc.erl
@@ -935,6 +935,22 @@ next_producer() ->
             case leader_for_timeslot(Slot, RunEnv) of
                 {ok, Leader} ->
                     activate_next_leader(Leader, TopHeight, SlotInfo, RunEnv);
+                {error, epoch_did_not_end} ->
+                    %% The wall-clock slot drifted past the resolvable schedule
+                    %% (leader_for_timeslot/2 has already cached the current and,
+                    %% when resolvable, the next epoch). Waiting for the drifted
+                    %% slot would stall forever, and any single fallback slot
+                    %% deadlocks if its leader is offline. Instead produce at the
+                    %% earliest passed slot this node leads (holes fill the gap),
+                    %% so any live producer can unstick a stalled chain.
+                    {ok, #{last := Last, length := Length}} = aec_chain_hc:epoch_info(RunEnv),
+                    MaxSlot = min(Slot, Last + Length),
+                    case own_slot_after_stall(TopHeight + 1, MaxSlot) of
+                        {ok, OwnSlot, Leader} ->
+                            activate_next_leader(Leader, TopHeight, SlotInfo#{slot := OwnSlot}, RunEnv);
+                        none ->
+                            {wait, not_producer, BlockTime}
+                    end;
                 {error, _} ->
                     {error, not_in_cache, BlockTime div 5}
             end
@@ -997,24 +1013,16 @@ get_slot_info(0, _RunEnv) ->
     Now = aeu_time:now_in_msecs(),
     #{epoch => 1, slot => 1, t0 => Now, delta => 0, now => Now};
 get_slot_info(_TopHeight, RunEnv) ->
-    {ok, EpochInfo = #{epoch := Epoch, first := First, last := Last}} = aec_chain_hc:epoch_info(RunEnv),
+    {ok, EpochInfo = #{epoch := Epoch, first := First}} = aec_chain_hc:epoch_info(RunEnv),
     BlockTime     = child_block_time(),
     Now = aeu_time:now_in_msecs(),
     T0  = get_t0(Epoch, First),
 
     ElapsedBlocks = (Now - T0) div BlockTime,
-    Delta0 = if Epoch == 1 -> ElapsedBlocks + 1;
-                true       -> ElapsedBlocks
-             end,
-    %% Never project a slot beyond the current on-chain epoch's last slot. The
-    %% validator schedule for a future epoch can't be resolved until this one
-    %% is actually closed on-chain (leader_for_timeslot/2 returns
-    %% `epoch_did_not_end` past `Last`), so an uncapped wall-clock Delta after
-    %% any production gap longer than one epoch's real-time span would make
-    %% next_producer/0 retry `not_in_cache` forever instead of catching up via
-    %% the existing hole-block mechanism one epoch at a time.
-    Slot = min(First + Delta0, Last),
-    Delta = Slot - First,
+    Delta = if Epoch == 1 -> ElapsedBlocks + 1;
+               true       -> ElapsedBlocks
+            end,
+    Slot = First + Delta,
     #{epoch => Epoch, slot => Slot, t0 => T0, delta => Delta,
       now => Now, epoch_info => EpochInfo}.
 
@@ -1108,6 +1116,23 @@ cache_validators_for_epoch(RunEnv, Epoch) ->
         {ok, EpochInfo} -> cache_validators_for_epoch_info(RunEnv, EpochInfo);
         _Err            -> error
     end.
+
+%% After a stall, scan the passed-but-unproduced slots (already cached by
+%% leader_for_timeslot/2, incl. next-epoch spill-over) for the earliest one
+%% this node holds the leader key for. Stops at the first uncached slot.
+own_slot_after_stall(FromSlot, MaxSlot) when FromSlot =< MaxSlot ->
+    case leader_for_height(FromSlot) of
+        {ok, Leader} ->
+            SignModule = get_sign_module(),
+            case SignModule:is_key_present(Leader) of
+                true  -> {ok, FromSlot, Leader};
+                false -> own_slot_after_stall(FromSlot + 1, MaxSlot)
+            end;
+        {error, _} ->
+            none
+    end;
+own_slot_after_stall(_FromSlot, _MaxSlot) ->
+    none.
 
 cache_validators_for_epoch(RunEnv, Hash, Epoch) ->
     case epoch_info_for_epoch(RunEnv, Epoch) of
--- a/apps/aehttp/test/aehttp_hyperchains_SUITE.erl
+++ b/apps/aehttp/test/aehttp_hyperchains_SUITE.erl
@@ -197,9 +197,9 @@ groups() ->
     , {hc_hole, [sequence],
           [ start_two_child_nodes
           , produce_first_epoch
-          , production_recovers_after_long_stall
           , hole_production
           , hole_production_eoe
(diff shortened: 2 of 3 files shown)


venimus:
Follow-up note on an interaction worth reviewers double-checking:

Since holes are produced back-to-back during catch-up (not paced to wall-clock), a long stall can
close many child epochs far faster than the normal 30min/epoch cadence. Each epoch closed
increases the *required parent height* for `get_entropy_hash(Epoch+2)` by `parent_epoch_length`
(independent of real elapsed time). So during a fast catch-up the child can outrun how many
parent blocks actually exist yet.

This is already handled by existing, unmodified code in `state_pre_transform_node` (the
`Height =:= EpochLast` branch): if `get_entropy_hash(Epoch+2)` returns `{error, _}`, the block
production/application attempt is aborted via `aec_conductor:throw_error(parent_chain_not_synced)`
and retried later - no invalid state is ever committed, since `entropy_height` is a deterministic
function of a real/immutable parent block, so this resolves identically for every node whether hit
live or during replay.

Net effect: this PR's fix removes the *earlier* deadlock (wall-clock-projected Slot exceeding the
already-known epoch's `last`), but catch-up speed is still naturally rate-limited by how far the
parent chain has actually progressed once the two chains are back in sync at "the edge" of already
-mined parent history. That's expected/pre-existing behavior, not something this PR needs to
handle further, but worth having in mind when reviewing/testing recovery from very long stalls.

venimus:
Rollout note, in case it affects how this gets deployed to `aehc_demo` or other HC networks:

Traced what happens with a partial/rolling upgrade (only some validators running this fix). Two
different validation rules matter here:

- `is_leader_valid_for_block/2`: a real (non-hole) block must be produced by the *exact* scheduled
  leader for that height - strict.
- `is_leader_valid_for_hole/2`: a hole block only needs the producer to be *a member of the cached
  schedule*, not the originally-assigned leader for that specific height (see the existing comment
  there: "A hole block can be mined by any node on the schedule, which is NOT the current leader.
  But we also accept other nodes' hole blocks" - also empirically exercised by the existing
  `hole_production_eoe` test, which excludes the scheduled leader and asserts other validators
  still close the epoch via holes).

Consequence: a single patched validator does **not** need to wait for the others to be upgraded
too. Once it's its own turn per the (now-fixed) wall-clock projection, it can flush the *entire*
outstanding backlog as holes in one sweep - including heights nominally assigned to a still-
unpatched/still-down validator - because hole validation only checks schedule membership, not
per-height ownership. The unpatched validator just keeps retrying `not_in_cache` harmlessly in the
background until it's also patched/restarted, at which point it's just a normal behind-node doing
regular sync.

Only side effect: reward/pinning credit for slots nominally belonging to the still-unpatched
validator goes to whichever patched validator filled them as holes instead - an economic/fairness
consideration, not a safety/consensus issue (holes carry no transactions).

Practical takeaway for recovering a real stalled network like `aehc_demo`: patching and restarting
just one validator should be sufficient to fully unstick production, rather than requiring a
coordinated simultaneous upgrade of all validators.

venimus:
Confirming this has zero effect on PoW chains (mainnet/testnet), verified via the actual dispatch
code rather than assumed:

- `aec_consensus_bitcoin_ng:get_type() -> pow.` vs `aec_consensus_hc:get_type() -> pos.`
- `aec_conductor:set_mode/1` sets `State#state.mode` from that: `pow` -> `local_pow`/stratum,
  `pos` -> `pos`.
- The whole HC production path this PR touches -
  `create_key_block_candidate(#state{mode = pos})` -> `hc_create_block_fun` ->
  `aec_conductor:hc_next_producer` -> `aec_consensus_hc:next_producer/0` -> the modified
  `get_slot_info/2` - is only reachable when `mode =:= pos`.
- Which consensus module is active per height is itself config-driven
  (`consensus_module_from_type(<<"pow_cuckoo">>) -> aec_consensus_bitcoin_ng`), so a PoW chain
  never even loads `aec_consensus_hc` as its active module in the first place.

This change only touches `aec_consensus_hc.erl` and is structurally unreachable from a PoW chain's
conductor state - no blast radius on mainnet/testnet.

venimus:
Recovery-speed note, since this will matter for anyone actually deploying this fix to a long-
stalled network:

Catch-up is NOT paced to the nominal 3s block time. In `aec_conductor:hc_next_producer/1`, the
"produce a hole now" branch (`WaitT =< 0`) has no `maybe_sleep` before creating the next hole - it
goes straight into the next iteration. So the entire backlog (which, for a stall this long, is
100% holes) gets produced back-to-back, bound only by:

1. Real per-hole cost: `aec_chain_state:calculate_state_for_new_keyblock` (state-tree recompute +
   reward accounting + signing) + a RocksDB/mnesia write + the conductor's async worker
   round-trip. This is CPU/disk-bound, not protocol-bound.
2. Marginal chain-length growth effects on the db over millions of extra records.
3. Parent-chain (testnet) pace, but only near the very tail of the catch-up - per the earlier
   entropy-height math, most of the missed epochs should resolve against parent blocks already
   mined during the same stall window; only the last epoch or so should need to wait for genuinely
   new parent blocks.

For `aehc_demo` specifically (~4.7M missed blocks over ~163 days), that puts full catch-up
somewhere in the range of hours to a few days depending on actual per-hole throughput on the
validator hardware - I have not benchmarked the hole-production path directly, so this is a
reasoned order-of-magnitude estimate, not a measured figure. Recommend timing actual
blocks/sec from a live catch-up (e.g. watch `top_block_height` over the first minute after
restart) before committing to any recovery-time estimate operationally.

venimus:
## Final resolution

The initial fix (clamping the wall-clock slot to the epoch's `last` in `get_slot_info/2`) turned out to introduce its own deadlock, caught deterministically by the `hc_hole` group: if the last slot's leader is itself offline/excluded, every other producer waits on a slot only that leader can fill (`Node is not producer for <Last>` forever), and the chain freezes. `hole_production` reproduced this every run.

### Final design (a6a1054c3)

- `get_slot_info/2` keeps the wall-clock slot **uncapped** (as on master).
- `next_producer/0` handles `{error, epoch_did_not_end}` explicitly: `own_slot_after_stall/2` scans the passed-but-unproduced slots (bounded by `Last + Length`, i.e. current epoch plus already-resolvable next-epoch spill-over; stops at the first uncached slot) for the earliest slot **this node** holds the leader key for, and produces there — holes fill the gap. Nodes without a leader key in range return `{wait, not_producer}` instead of hot-looping `not_in_cache`.

Recovery therefore no longer depends on any *specific* leader being alive: whichever producer comes back first takes its own earliest slot and unsticks the chain, epoch by epoch, via the existing hole mechanism.

Consensus/sync safety is unchanged: the fallback lives purely in the producer-side liveness path. Block validation stays chain-state-deterministic (`state_pre_transform_node` → `leader_for_timeslot(Height, {TxEnv, Trees})`), and hole blocks from any on-schedule producer were already valid — so unpatched peers and freshly syncing nodes accept the recovery blocks exactly as before.

### Test fixes in the same commit

- `production_recovers_after_long_stall` had three latent bugs, never surfaced because the test was auto-skipped whenever `hole_production` failed earlier in the sequence:
  - full node names passed to `rpc/4`, which expands them again via `node_name/1` → `{badrpc, nodedown}`;
  - `{ok, _}` match on `produce_n_epochs/2`, which returns bare `ok`;
  - mining restarted before `produce_cc_blocks`, while `hc_mine_blocks/4` asserts it is stopped on entry.
- `produce_cc_blocks_with_retry` now accumulates blocks across timed-out attempts by re-fetching the whole range from the settled top after success — recovery forks can reorg away heights an earlier attempt had reached, so stitching per-attempt prefixes raced (`{failed_to_fetch_generation, H}`). It also mines a few parent blocks before each retry to break the circular wait between next-epoch entropy and child progress.

With these, the full `hc_hole` group passes, including the previously-skipped stall-recovery test.

PR #4384, opened 2024-08-13
Next protocol name: Arcus

ThomasArts:
Prepare code and tests for a future protocol change, named Arcus.

closes #4357 
This PR is supported by the Æternity Foundation

PR #4685, opened 2026-08-01
Size-proportional FATE store-read gas at Arcus (v7), plus Salus (v8) placeholder

venimus:
Reworked continuation of #4629 (credit @ifaouibadi), rebased on `master` with the fork gate on **Arcus (v7)**, the next scheduled consensus protocol. **Salus (v8)** is added as a bare placeholder reserved for Hyperchains.

## What
- From **Arcus (v7)**, FATE contract-storage reads are charged in proportion to the size of the value read (`STORE_READ_BASE_GAS = 100` + `STORE_READ_BYTE_GAS = 10`/byte), so gas accurately reflects the resources a read consumes.
- Adds protocol version 8 (**Salus**) as a placeholder with no behaviour of its own — protocol constant, name mappings, fork dir, `local_salus_testnet` and test lane only. Reserved for Hyperchains.

## How
- Pre-Arcus store behaviour is pinned: Iris..Ceres dispatch to a frozen copy (`aefa_stores_ceres`) whose source is locked to a golden sha256 by `aefa_stores_ceres_freeze_test` / `check_stores_ceres_freeze.sh`, so the repricing is provably inert on all existing networks until an Arcus height is scheduled.
- Protocol dispatch is centralized in `aefa_engine_state:aefa_stores_for_protocol/1` (Arcus+ → live `aefa_stores`, Iris..Ceres → frozen, pre-Iris → lima). The Arcus clause is a `>=` guard, so Salus and every later protocol inherit the repricing without a code change.
- Repricing proof and regression tests added.

## Behaviour changes at Arcus (fork-gated)
- Store reads cost gas proportional to value size; reads of large values are priced accordingly instead of flat.
- Updating an existing store key includes a priced read of the old value, so fresh-key writes become cheaper than same-size updates.
- Failed remote calls roll back slightly cheaper for small stores.

## Dry-run metering
Dry-run estimate profiles (`public`/`internal`) meter store reads at the Arcus cost by default (`http.dry_run.store_read_gas_metering`, default `true`), independent of the activated protocol — gas amounts only, never tx validity/ABI/VM version, never committed. Historical replay (Rosetta/indexer) and mempool includability stay pinned to the real protocol.

## Drive-by fixes to pre-existing Arcus scaffolding
Arcus was added in #4384 as a behaviourless placeholder; two spots assumed it would stay that way and are now wrong:
- `aecontract_SUITE:init_tests/2` mapped the `arcus` release label to `?CERES_PROTOCOL_VSN` — harmless while Arcus was inert, but it would silently run Ceres semantics now. Fixed, and the missing `salus` row added.
- `aefa_engine_state_test`'s dispatch-boundary table pinned Arcus to the frozen module; it now asserts the real boundary (Ceres → frozen, Arcus → live, Salus → live).

## Validation
- `make eunit-arcus`: 1033 tests, 0 failures.
- Targeted repricing/dispatch/dry-run suites green under both the `arcus` and `salus` lanes (Salus correctly inherits the repricing).
- `aecontract_SUITE` under arcus: state-gas and protected-call gas expectations made protocol-aware for the repricing.
- Freeze golden-hash check passes; mining test nonces computed for both new protocols.

## Ecosystem notes (before any v7 activation)
- **js-sdk**: `ConsensusProtocolVersion` only knows Ceres=6; `getNodeInfo()` throws `UnsupportedVersionError` for unknown protocols. Arcus=7 (+ `ProtocolToVmAbi`) must land in the SDK before any network activates v7.
- **ae-mdw**: hardfork handling is generic via `:aec_hard_forks`; no change needed unless `.arcus` presets ship (`origin.ex` `hc_contracts/0` hardcodes `in [5, 6]`). Reported dry-run `gas_used` for store-heavy contracts will increase post-fork; no schema break.
- **Dry-run**: with size-proportional pricing, the per-request gas cap on dry-run endpoints (node default 6M) becomes an accurate bound on the work a read-only call can consume.

Supersedes #4629.


venimus:
Pushed a batch of fixes addressing review feedback (both the in-repo review notes and the æGency review deck).

**Real bugs fixed (verified against git history, not just the review text):**
- `finalize/1`'s `terms_to_finalize` bypassed the store-module dispatcher even for Iris..Ceres, where `aefa_stores_ceres` has its own copy - a future edit to the live module could've silently changed already-forked replay. Now only falls back to the live module for the Lima case (which genuinely has no such function).
- `force_arcus_for_profile/1` failed open: a *missing* dry-run profile silently forced forward Arcus gas estimates, same as explicit `internal`. Now defaults safe (no forced metering) unless a caller opts in explicitly.
- Gas-accounting asymmetry in `compute_reuse_fixpoint`'s fallback paths - gas already spent in the optimistic pass was discarded instead of carried forward, letting the same store reads be redone for free.
- `store_map_size`/`store_map_get_clean` bypassed the new byte-proportional Arcus metering entirely; now gated consistently with the other store-map ops.
- Internal (trusted) dry-run listener was hardcoded to the `public` profile, defeating the schema's documented "internal calls are unbounded."

**Reverted two of my own "fixes"** after tracing `git show` on the introducing commits proved the original code was correct (pre-Iris store routing, and the Lima `terms_to_finalize` hardcode) - noting this since it's relevant to how much to trust either review's line-level claims.

**Housekeeping:** removed a stray release note describing an already-merged ancestor change, fixed a couple of stale "Arcus (v8)"/"Arcus/Arcus" comments, pinned `aebytecode` to its exact commit instead of a mutable tag (matching every other pinned dep), added the missing `.arcus` release-overlay dir, removed dead test-only exports, de-flaked a test that compared VM-wide process count across a ~2.3s window, added Lima-protocol store-map test coverage (lookup/member/size/to_list/GC - previous tests only touched a plain register), and pinned a literal gas constant in eunit so a future accidental repricing change wouldn't go undetected.

**Deferred, not fixed (flagged for follow-up):** subtree bytes are still charged only after full materialization (bounded by the tx gas limit, but a min-gas tx still forces one full read before aborting); and a pre-existing (pre-2019, unrelated to this PR) uncharged per-key read in `size_delta/3` during store-map write-commit. Both would need a deeper, riskier refactor of the write/read primitives than is appropriate to bundle in here.

All 60 eunit tests across the touched modules pass; full project compiles clean.

venimus:
Follow-up on the two items I deferred rather than fixing in this PR — filed as tracked issues instead of leaving them only in local notes: #4693 (subtree bytes charged only after full materialization) and #4694 (`size_delta/3`'s uncharged per-key read during write-commit, pre-dates this PR to 2019).

Revising my earlier "bounded, low severity" framing: bounded-per-block is true, but aeternity's gas price floor is tiny in fiat terms, so a sustained, continuous, automated attacker doesn't need a single big exploit — they can just repeat either of these every block indefinitely at negligible real dollar cost. "Can't get more than X per call" isn't much reassurance against "can get X for free, forever, cheaply." Neither is a correctness/consensus bug (gas charged on a *successful* call is still accurate), and neither is introduced by this PR, but both are real hardening gaps worth tracking and fixing on their own timeline rather than bundled into the store-read repricing here.

PR #4686, opened 2026-08-02
Cache the network id instead of resolving it per signature

CedrikNikita:
aec_governance:get_network_id/0 was resolved from scratch on every call. Resolving asks the node-global 'init' process for the command line arguments and, failing that, copies the whole validated user config out of the application environment and runs setup's value expansion over it. aetx_sign did that once per candidate signature, and FATE does it per NETWORK_ID instruction.

The value cannot change while the node runs - the only runtime config update path, aecli, only touches min_miner_gas_price - so resolve it once from a setup hook and keep it in a persistent term. The hook runs at phase 105, after the user config has been read and expanded (aeutils hooks 100-103) and before the hooks that derive their env from the network id. A cache miss deliberately does not populate the cache, so environments that never run the hook keep following the configuration.

The cache is erased again in aecore_app:stop/1. Setup hooks only run once per VM, and the aecore application is a dependency of setup's own start, so a test node that stops aecore, reconfigures it and starts it again - aec_eunit_SUITE, and the devmode suites that stuff a different network id into the aecore env - would otherwise keep being served the value resolved for the previous incarnation.

In aetx_sign, the payloads a signature is checked against depend on the transaction alone, so build them once per transaction instead of once per signature. The hashed payload stays lazy - it is only needed once a signature fails to match the plain one - but is now reused across the remaining signatures and signers.

  network id lookup      ns/call
  resolve                 1038.8
  cached get                31.6      30x cheaper

  verify_half_signed/3   before us   after us
  plain, 1 signer            41.5       39.8
  plain, 2 signers          158.5      156.6
  plain, 8 signers          898.8      864.6
  hash,  1 signer            80.6       78.5
  hash,  2 signers          240.5      233.4
  hash,  8 signers         1215.4     1185.4

End to end that is a 2-4% saving: Ed25519 verification is ~40us per check and dominates. The lookup cost is what the change actually removes, which matters more for the callers where it is the whole operation - the FATE NETWORK_ID instruction, the peer connection prologue, the HTTP and rosetta handlers - and for removing a serialisation point on the node-global 'init' process from parallel transaction validation, which a single threaded benchmark cannot show.

### DIFF
--- a/apps/aecore/src/aec_governance.erl
+++ b/apps/aecore/src/aec_governance.erl
@@ -41,6 +41,10 @@
          vm_gas_table/0,
          min_tx_gas/0]).
 
+%% aecore_app only - between them they repoint every signature check in the node
+-export([ensure_env/0,
+         clear_network_id_cache/0]).
+
 -include("blocks.hrl").
 -include_lib("aebytecode/include/aeb_opcodes.hrl").
 -include_lib("aecontract/include/hard_forks.hrl").
@@ -49,9 +53,17 @@
 
 -ifdef(TEST).
 -define(NETWORK_ID, <<"local_testnet">>).
+%% Exported and called fully qualified only here, so that the tests can mock the
+%% resolution: the eunit VM is started with '-network_id local_<protocol>_testnet'
+%% (see the Makefile), which wins over anything a test could set from the inside.
+-export([resolve_network_id/0]).
+-define(RESOLVE_NETWORK_ID(), ?MODULE:resolve_network_id()).
 -else.
 -define(NETWORK_ID, <<"ae_mainnet">>).
+-define(RESOLVE_NETWORK_ID(), resolve_network_id()).
 -endif.
+
+-define(NETWORK_ID_CACHE_KEY, {?MODULE, network_id}).
 
 -define(BLOCKS_TO_CHECK_DIFFICULTY_COUNT, 17).
 -define(TIMESTAMP_MEDIAN_BLOCKS, 11).
@@ -415,7 +427,32 @@ add_network_id_last(Payload) ->
     NetworkId = ?MODULE:get_network_id(),
     <<Payload/binary, NetworkId/binary>>.
 
+%% Resolving walks 'init' args and the user config tree - too expensive per
+%% signature verified and per FATE NETWORK_ID instruction. Pinned in a persistent
+%% term by the aecore setup hook, and cleared by aecore_app:stop/1, since setup
+%% hooks run only once per VM.
+-spec ensure_env() -> ok.
+ensure_env() ->
+    persistent_term:put(?NETWORK_ID_CACHE_KEY, ?RESOLVE_NETWORK_ID()).
+
+-spec clear_network_id_cache() -> ok.
+clear_network_id_cache() ->
+    _ = persistent_term:erase(?NETWORK_ID_CACHE_KEY),
+    ok.
+
+-spec get_network_id() -> binary().
 get_network_id() ->
+    %% A miss must not populate: CT nodes sign without running the setup hook and
+    %% swap the id under themselves (aehttp_devmode_SUITE).
+    case persistent_term:get(?NETWORK_ID_CACHE_KEY, undefined) of
+        undefined -> ?RESOLVE_NETWORK_ID();
+        NetworkId -> NetworkId
+    end.
+
+%% Not exported outside test builds: a caller that resolves instead of asking
+%% get_network_id/0 would silently disagree with every signature check.
+-spec resolve_network_id() -> binary().
+resolve_network_id() ->
     %% Easy switch networks with console argument, it overwrites config
     case init:get_argument(network_id) of
         {ok, [[NetworkId]]} -> list_to_binary(NetworkId);
--- a/apps/aecore/src/aec_governance.erl
+++ b/apps/aecore/src/aec_governance.erl
@@ -429,11 +429,21 @@ add_network_id_last(Payload) ->
 
 %% Resolving walks 'init' args and the user config tree - too expensive per
 %% signature verified and per FATE NETWORK_ID instruction. Pinned in a persistent
-%% term by the aecore setup hook, and cleared by aecore_app:stop/1, since setup
+%% term by the aecore setup hook, and again by aecore_app:start/2, since setup
 %% hooks run only once per VM.
 -spec ensure_env() -> ok.
 ensure_env() ->
-    persistent_term:put(?NETWORK_ID_CACHE_KEY, ?RESOLVE_NETWORK_ID()).
+    %% Resolve before the put: clearing first opens a window for readers, and
+    %% re-putting an unchanged value is free where a clear is a global sweep.
+    try ?RESOLVE_NETWORK_ID() of
+        NetworkId -> persistent_term:put(?NETWORK_ID_CACHE_KEY, NetworkId)
+    catch
+        Class:Reason:Stacktrace ->
+            %% Fail closed: an id resolved under a configuration that is already
+            %% gone must not outlive it. Resolving raises the same error again.
+            ok = clear_network_id_cache(),
+            erlang:raise(Class, Reason, Stacktrace)
+    end.
 
 -spec clear_network_id_cache() -> ok.
 clear_network_id_cache() ->
--- a/apps/aecore/test/aec_governance_tests.erl
+++ b/apps/aecore/test/aec_governance_tests.erl
@@ -135,18 +135,24 @@ set_resolved_network_id(NetworkId) ->
 %% Both failures are silent: a missing hook only costs the lookup again, a hook
 %% renumbered ahead of aeutils pins the schema default for the node's lifetime.
 setup_hook_test_() ->
-    {setup,
-     fun ensure_apps_loaded/0,
-     [{"The network id is pinned by a setup hook",
-       fun() ->
-               ?assertMatch([_], network_id_hook_phases())
-       end},
-      {"The hook runs after the config is read and before what derives from it",
-       fun() ->
-               [Phase] = network_id_hook_phases(),
-               ?assert(Phase > lists:max(normal_hook_phases(aeutils))),
-               ?assert(Phase < lists:min(normal_hook_phases(aecore) -- [Phase]))
-       end}]}.
+    [{"The network id is pinned by a setup hook",
+      fun() ->
+              ?assertMatch([_], network_id_hook_phases())
+      end},
+     {"The hook runs after the config is read and before what derives from it",
+      fun() ->
+              [Phase] = network_id_hook_phases(),
+              ?assert(Phase > lists:max(normal_hook_phases(aeutils))),
+              ?assert(Phase < lists:min(normal_hook_phases(aecore) -- [Phase]))
+      end},
+     {"Reading the hooks does not disturb the applications of the eunit VM",
+      fun() ->
+              Loaded = fun() -> lists:sort(application:loaded_applications()) end,
+              Before = Loaded(),
+              _ = normal_setup_hooks(aecore),
+              _ = normal_setup_hooks(aeutils),
+              ?assertEqual(Before, Loaded())
+      end}].
 
 network_id_hook_phases() ->
     [Phase || {Phase, {?TEST_MODULE, ensure_env, []}}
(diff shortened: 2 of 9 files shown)


venimus (dismissed):
Nice find and clean fix. Confirmed the core assumption directly: grepped the whole tree for any production write path to the `network_id` app env besides tests/system_test tooling — there is none (`aecli` only reads it via `get_network_id/0`), so pinning it at startup via the `ensure_env/0` hook is safe.

The lifecycle handling (cache populated by a phase-105 setup hook, cleared in `aecore_app:stop/1`, deliberate no-op-populate on a cache miss) is well thought through for the test-suite pattern of stopping/reconfiguring/restarting `aecore` in the same VM (`aec_eunit_SUITE`, devmode suites).

The `aetx_sign.erl` restructuring (building the plain + lazy-hash payload once per transaction and threading it through `verify_signatures_`/`verify_one_pubkey_`, plus short-circuiting invalid pubkeys before serializing the tx) is a correct, careful refactor — traced through the recursion and it preserves the original semantics (same plain-then-hash fallback order, same result for the fully-GA-authenticated `[]`/`[]` case, same `{error, signature_check_failed}` for bad pubkeys/leftover signatures).

Ran the new/modified eunit suites locally (`aec_governance_tests`, `aetx_sign_tests`): 36 tests, 0 failures.

Nothing blocking. 👍

venimus:
Re-reviewed the new commit (`0cab42436`) — this closes a real gap my first pass missed: `aec_consensus_bitcoin_ng` stops and restarts `aecore` in-process on a community-fork rollback (confirmed at `apps/aecore/src/aec_consensus_bitcoin_ng.erl:126,128`), and since setup hooks only fire once per VM (during `application:start(setup)`, well before `aecore` even starts), that restart alone would never re-populate the cache - correct behavior (falls back to resolving), but at exactly the wrong time performance-wise (mid-recovery). Calling `ensure_env/0` from `aecore_app:start/2` directly fixes that, and re-pins to the same value on an ordinary fresh boot since the hook already ran.

The resolve-before-clear ordering in the hardened `ensure_env/0` is the right call too: a failing resolution now leaves the cache empty (falls back to config) rather than either leaving a stale pinned value in place, or paying a persistent-term GC sweep by clearing preemptively on the common, non-failing path.

Ran the updated eunit suite locally: `aec_governance_tests`, `aetx_sign_tests` - 38 tests, 0 failures (including the new `failed_ensure_env_leaves_cache_empty` and `app_lifecycle_test_` cases).

Still approved - nothing further blocking.

dincho:
Do you guys just relaying your slop generator outputs via github? 🤣

apps/aeutils/priv/aeternity_config_schema.json
```diff
@@ -2591,7 +2591,7 @@
             "additionalProperties" : false,
             "properties" : {
                 "network_id" : {
-                    "description" : "Identification of the network in case of hard forks.",
+                    "description" : "Identification of the network in case of hard forks. Read once when the node boots; a change takes effect only after a node restart.",
```
velzevur: The change took effect after a restart in the previous version as well but it looks like AI was really proud of its work. Maybe a short description how changing this value would break backwards compatibility and render the DB useless would be more helpful for an end user. What do you think @dincho?

docs/release-notes/next/network_id_read_once_at_boot.md
```diff
@@ -0,0 +1,8 @@
+* `fork_management: network_id` is now read once, while the node boots, and kept for as
+  long as the node runs, instead of being resolved from the configuration on every use.
+  The network id is prepended to every transaction before it is signed and verified, is
+  returned by the FATE `NETWORK_ID` instruction, and selects the fork schedule and the
+  genesis accounts, so holding it constant keeps all of those consistent with the id the
+  node started under. Changing it therefore requires a node restart - as it did in
```
velzevur: Since the network id is a component of signature validation, changing it would instantly render all signatures on previous blocks, headers and transactions invalid so maybe here focus more on the chaos that will rain upon your users if they change it?

If you really want your users to change network ids on existing chains, then the network id should depend **on height** and switch it automatically.

docs/configuration.md
```diff
@@ -143,6 +143,8 @@ fork_management:
     network_id: ae_uat
 ```
 
+The network id is read once, when the node boots, and kept for as long as the node runs. Changing it in the configuration file of a running node takes effect only after a restart.
```
velzevur: You might want to revisit that statement as well, I strongly suggest you don't incentivise users to change it for an existing chain.

apps/aeutils/priv/aeternity_config_schema.json
venimus: I'm pretty sure it will fail long before any writing on db

docs/release-notes/next/network_id_read_once_at_boot.md
venimus: The idea of the PR is to avoid useless resloving all the time, when it makes no sense to change it once it is set initially.

docs/configuration.md
venimus: I guess adding a note that it makes no sense to change the id it once blockhain synced the first block.

venimus:
> Do you guys just relaying your slop generator outputs via github? 🤣

All changes are actually tested multiple times in a real network testing setup for few days under heavy load (including A/B testing).

apps/aecore/src/aec_governance.erl
```diff
@@ -415,7 +420,54 @@ add_network_id_last(Payload) ->
...
+    end.
+
+-spec clear_network_id_cache() -> ok.
+clear_network_id_cache() ->
+    _ = persistent_term:erase(?NETWORK_ID_CACHE_KEY),
+    ok.
+
+-spec get_network_id() -> binary().
 get_network_id() ->
+    %% Note that a cache miss deliberately does *not* populate the cache: in
```
velzevur: This makes no sense whatsoever. What environments do not run the setup hook? How is network id changed on the go? I'd populate the cache...

apps/aecore/src/aec_governance.erl
```diff
@@ -37,6 +37,9 @@
          add_network_id_last/1,
          add_custom_network_id/2,
          get_network_id/0,
+         resolve_network_id/0,
```
velzevur: So if a user changes their config and does not restart their machine, resolve_network_id/0 and get_network_id/0 would return different values? Is this by design? Sounds a bit dangerous. May I suggest you do NOT export that function?

apps/aecore/src/aec_governance.erl
venimus: https://github.com/aeternity/aeternity/pull/4686/changes#diff-38046bd54184f0c8f88234bbdc78c5dae186d43c85cdc9e858ca8947a25d7aecR471

apps/aecore/test/aec_governance_tests.erl
```diff
@@ -0,0 +1,222 @@
...
+    ok = ?TEST_MODULE:ensure_env(),
+    ?assertEqual(?NW_ID_A, ?TEST_MODULE:get_network_id()),
+    %% The failure resolve_network_id/0 can actually produce: its last clause
+    %% guards on is_binary/1, so a non-binary left in the user config or in the
+    %% aecore env - which test code does write directly - is a case_clause.
+    %% Declared through meck:raise/2 rather than raised from inside a fun, which
+    %% would leave the mock marked invalid for anything calling meck:validate/1.
+    meck:expect(?TEST_MODULE, resolve_network_id, 0,
+                meck:raise(error, {case_clause, "ae_uat"})),
+    ?assertError({case_clause, "ae_uat"}, ?TEST_MODULE:ensure_env()),
```
velzevur, line 84: Uhm, you test that `meck` will rise the error you set it to rise? Are you testing that `meck` works?

apps/aecore/test/aec_governance_tests.erl
```diff
@@ -0,0 +1,222 @@
...
+     end,
+     fun(_) ->
+             meck:unload(aec_jobs_queues),
+             meck:unload(aec_db),
+             meck:unload(aec_db_gc),
+             meck:unload(?TEST_MODULE),
+             ok = ?TEST_MODULE:clear_network_id_cache()
+     end,
+     fun(_) ->
+             [{"Stopping aecore drops the pinned network id",
```
velzevur: Yeah, another useful test. Who needs `aecore` anyway...

apps/aecore/test/aec_governance_tests.erl
```diff
@@ -0,0 +1,222 @@
...
+               fun() ->
+                       set_resolved_network_id(?NW_ID_A),
+                       ok = ?TEST_MODULE:ensure_env(),
+                       ?assertEqual(?NW_ID_A, ?TEST_MODULE:get_network_id()),
+                       set_resolved_network_id(?NW_ID_B),
+                       ?assertEqual(?NW_ID_A, ?TEST_MODULE:get_network_id()),
+                       ok = aecore_app:stop(undefined),
+                       ?assertEqual(?NW_ID_B, ?TEST_MODULE:get_network_id())
```
velzevur: So this tests that if `aecore` dies, the network id... changes?

apps/aecore/test/aec_governance_tests.erl
```diff
@@ -0,0 +1,222 @@
...
+                       %% it keeps the test honest: start/2 cannot have returned
+                       %% before reaching the statement after the re-pin.
+                       ?assertError({badmatch, {error, not_in_a_unit_test}},
+                                    aecore_app:start(normal, [])),
+                       set_resolved_network_id(?NW_ID_A),
+                       ?assertEqual(?NW_ID_B, ?TEST_MODULE:get_network_id())
+               end}]
+     end}.
+
+%% Nothing else would notice the hook going missing: a cache miss falls back to
```
velzevur: Yeah, who would notice....

apps/aecore/src/aec_governance.erl
venimus: but get_network_id calls reslove, how it could be different? Though not sure about the exporting, probably your suggestion makes sense

velzevur: Nope, `get_network_id/0` checks persistent term and if found (which is always, since the setup hook had passed) it returns whatever was put there at startup. `resolve_network_id/0` checks the env so if it had changed - it will return whatever is the new value and your node will happily split into a different chain.

docs/configuration.md
velzevur: Well the genesis block is not synced but generated. It makes no sense changing it at all, ever. That is unless you start supporting dynamic network id depending on the height.

docs/release-notes/next/network_id_read_once_at_boot.md
velzevur: There might be useful cases for changing it: ex ETH and ETC. You might want to split in a separate chain and if you want to take advantage of the replay attack protection - you do want a different network id.

apps/aecore/src/aec_governance.erl
venimus: The concern seems legitimate, so definitely it should not export it, although it doesn't really makes sense to me to use it somewhere either way.

apps/aecore/test/aec_governance_tests.erl
venimus: nice catch tho

apps/aeutils/priv/aeternity_config_schema.json
venimus: Btw It seems the original description is confusing. The core problem with "in case of hard forks" is that it reads like a conditional trigger ("use this setting in the event that you want to hard fork"), looks like the added stuff is further confusing

docs/release-notes/next/network_id_read_once_at_boot.md
venimus: I thought so too, but to be able to do that network_id should be height bind IMHO, similar to protocol switch. Need to implement get_network_id_for_height, not a config that only makes sense on genesis block.

docs/configuration.md
venimus: yeah i meant genesis+1

venimus:
Worths mentioning that #4691 might make the speed gain of this PR marginal, should probably be revaluated

apps/aecore/test/aec_governance_tests.erl
dincho: Typical slop I’m also fighting with my junk generators everyday 😆

apps/aecore/test/aec_governance_tests.erl
dincho: You don’t want to die with unfinished business, otherwise your soul won’t pass into the afterlife

apps/aecore/src/aec_governance.erl
velzevur: I'd rather not have 2 similar function doing almost the same thing and then sometimes returning different results. If it is exported, someone will use it.

PR #4687, opened 2026-08-02
Avoid deserializing the top block when validating pushed txs

CedrikNikita:


venimus (dismissed):
Good catch, and a nice bonus fix bundled in: `aec_chain:top_header_hash_and_state/0` already existed on master but its spec claimed `'undefined' | {...}` while the implementation actually returned the bare atom `error` for the empty-chain case (and wasn't wrapped in `activity/1` like its siblings, so the top-node lookup and state fetch weren't a consistent dirty-read snapshot). This PR fixes both while wiring up its first real callers in `aec_hc_vote_pool`/`aec_tx_pool`.

The `aec_headers:version/height` swap for `aec_blocks:version/height` in the pool validation checks is correct — confirmed `top_block_with_state_/0` was going through `aec_db:get_block/1` (full block, including tx list, deserialized) purely to read protocol version and height, whereas `top_block_node/0` returns the already-resident header map, avoiding that.

The empty-chain-db fallback path (restart chain db without inserting genesis, assert `get_onchain_env/0` still falls back correctly) is exactly the right edge case to cover given the sentinel-value fix, and it's tested in both `aec_hc_vote_pool_tests` and `aec_tx_pool_tests`.

Ran the affected eunit suites locally (`aec_chain_tests`, `aec_hc_vote_pool_tests`, `aec_tx_pool_tests`): 71 tests, 0 failures.

Nothing blocking. 👍

venimus (approved):

PR #4689, opened 2026-08-02
Stop re-reading mempool config on push replay

CedrikNikita:
Moves the mempool config reads (`tx_ttl`, `nonce_offset`, `invalid_tx_ttl`) out of the serialized `aec_tx_pool` gen_server loop. Each read walks the whole user config map, so doing it per-transaction inside the pool process serialized every push and made fork replay / startup restore / candidate selection pay the cost once per tx.

Now the value is read once per operation in the calling process (once per push, once for a whole fork replay, once for a whole mempool restore, once per candidate selection) and threaded through as an explicit parameter. No caching, no persisted state — each operation still reads fresh config at its start, so a runtime config change is picked up on the next operation exactly as before.

`allow_reentry` is also made lazy: only a tx that was already deleted can reenter, so the ordinary push path no longer pays for that config read at all.

## Also fixes a latent dead-code bug

`pool_db_raw_put`'s no-TTL branch matched integer `0`, but `aetx:ttl/1` returns the atom `max_ttl` (not `0`) for a tx with no TTL — so that clause never actually matched. The code produced correct results only by accident of Erlang term ordering (`min(Int, max_ttl)` always yields the integer). It now matches `max_ttl` explicitly; behavior is unchanged for every input, the reliance on term-ordering is removed.

## Behavior / compatibility

- No client-visible change: a tx with no explicit TTL (the common SDK/wallet case, `ttl = 0` → `max_ttl`) still gets the configured `tx_ttl` horizon; a tx with its own TTL can shorten its stay but never extend it — same as before.
- Fully self-contained: the changed `{push, ...}` gen_server message and the changed-arity internal helpers have no callers outside `aec_tx_pool`. No wire/DB/serialization change.

## Tests

New `aec_tx_pool_tests` coverage for the TTL combination on both the push path and the fork-replay / persisted-mempool-restore paths, deliberately run at a non-zero GC height so the two `min/2` operands don't commute (which would otherwise hide a swap of the configured TTL and the tx's own TTL).

### DIFF
--- a/apps/aecore/src/aec_tx_pool.erl
+++ b/apps/aecore/src/aec_tx_pool.erl
@@ -212,7 +212,11 @@ push_(Tx, TxHash, Event, Timeout, Forced) ->
     Push =
         fun() ->
             incr([push]),
-            Res = gen_server:call(?SERVER, {push, Tx, TxHash, Event}, Timeout),
+            %% Read the configured TTL here rather than in the pool process,
+            %% which serialises every push - each read walks the whole user
+            %% config map.
+            Res = gen_server:call(?SERVER, {push, Tx, TxHash, Event, tx_ttl()},
+                                  Timeout),
             instant_tx_confirm_hook(TxHash),
             Res
         end,
@@ -227,10 +231,7 @@ push_(Tx, TxHash, Event, Timeout, Forced) ->
                     incr([push, error]),
                     E;
                 ok ->
-                    incr([push]),
-                    Res = gen_server:call(?SERVER, {push, Tx, TxHash, Event}, Timeout),
-                    instant_tx_confirm_hook(TxHash),
-                    Res
+                    Push()
             end
     end.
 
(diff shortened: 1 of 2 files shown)


venimus (approved):
Clean, well-scoped change — moves configured `tx_ttl`/`nonce_offset`/`invalid_tx_ttl` reads out of the per-transaction hot paths (push, candidate selection, fork replay, mempool restore) into a single read per batch/call, threaded through as explicit state instead of re-reading a walked user-config map per tx.

The `pool_db_raw_put/6` match on `max_ttl` (replacing the old literal `0`) is correct, not a regression — `aetx:ttl/1` already converts `0 -> max_ttl` internally on master, so the old `0 ->` clause never actually matched; it only "worked" because `min(Integer, max_ttl)` happens to sort the integer first in standard term order. This makes that explicit rather than relying on the coincidence.

The added GC-height test explicitly guards the case that would have hidden a TTL-argument swap (non-zero GC height, since at height 0 the two `min/2` operands commute) — good attention to a real testing trap.

Ran the affected eunit suite locally: `aec_tx_pool_tests`, 18 tests, 0 failures.

Nothing blocking. 👍

PR #4690, opened 2026-08-02
Encode tx gossip wire message once per gossip, not once per peer

CedrikNikita:


venimus (approved):
Straightforward mirror of the existing block-gossip path: `gossip_serialize_tx/1` now wraps the signed tx in the same `aec_peer_messages:serialize(txs, ...)` + tag envelope that `gossip_serialize_block/1` already produces for blocks, so the encode happens once per gossip instead of once per peer, and `send_tx`/`send_gossip` collapse to the one cast. Confirmed the receive path (`handle_info({noise, _, <<Type:16, Payload/binary>>}, ...)`) is already generic over message type/tag, so no receiver-side change is needed — the test explicitly covers the round-trip through that shape plus the byte-identical-across-calls property the whole optimization depends on.

Ran the affected eunit suite locally: `aec_peer_connection_tests`, 7 tests, 0 failures.

Nothing blocking. 👍

PR #4691, opened 2026-08-03
Stop expanding boot fixed config on every read

CedrikNikita:


venimus (commented):
Really nice general fix — caching the *expanded* config only when it's provably non-expandable (`is_expandable/1` scanning for any `$` anywhere in the term) is the right way to do this safely, correctly sidestepping the `$LOG_DIR`-style rewrite-at-a-later-hook trap that a naive "cache on first read" would have hit. The follow-on commits swapping `aeu_env:get_env/2,3` for plain `application:get_env/2,3` on internally-written keys (`aecore > fork`, dev-reward keys) are well justified, and the `aec_hard_forks:ensure_env/0` fix (unsetting a stale `fork` env left by a previous run instead of a no-op) is a real, separate bug fix bundled in cleanly. The mid-PR test-isolation fix (commit 854026384) shows good rigor — caught a real cross-test VM-sharing leak.

One thing that needs reconciling before merge, though: **this duplicates #4686's network-id caching almost exactly** (`ensure_env/0`/cache-clear hook wired into `aecore.app.src`/`aecore_app.erl`, persistent-term-backed `get_network_id/0`) — same author, same mechanism, different names. Landing both as-is will conflict directly in `aec_governance.erl` and `aecore.app.src`.

Beyond the conflict, it looks strictly redundant here too: `network_id_from_config/0` (this PR's `resolve_network_id/0` equivalent) already goes through `aeu_env:user_config_or_env/4` → the now-cached `user_config()`, so the general config cache in this same PR (commit 7ea0f3ff7) should already make network-id resolution cheap on its own — the dedicated `?PT_NETWORK_ID` persistent-term cache and its own `ensure_env`/`cleanup` hook (commit 0df5e24d9) look like they can just be dropped in favor of #4686 (or vice versa, whichever lands first), rather than shipping two caches for the same value.

Ran the affected eunit suites locally: `aec_governance_tests`, `aeu_env_tests`, `aec_hard_forks_tests`, `aec_dev_reward_tests`, `aec_tx_pool_tests` — 57 tests, 0 failures.

Everything else here looks solid — just flagging the overlap for you to reconcile with #4686 before either merges.

venimus (approved):

PR #4692, opened 2026-08-04
Add protocol parameters endpoint exposing consensus constants

CedrikNikita:
Exposes the consensus parameters and node policy settings a client needs to build a valid transaction without hardcoding governance values.

Aettos amounts are encoded as decimal strings rather than JSON numbers: the name fee table reaches 5702887 * 10^14 (~2^69), far past the 2^53 an IEEE-754 client can hold, and a rounded fee produces a rejected claim. Gas, heights, counts and versions stay numbers.

The per-protocol block is memoised in persistent_term, keyed on the protocol version and the mining > name_claim_bid_timeout override so a runtime config change cannot be masked by a stale entry. Without it every request re-ran ~165 UTS46 IDNA conversions through aens_utils:to_ascii/1, measured at 1.43 ms of CPU per request on mainnet. aehttp_app fills the memo at startup: get-then-put is not atomic, and a racing pair of cold requests would each write, where the second write replaces a key and so schedules a scan of every process in the VM.

The response is the largest fixed body on the external API - ~17 kB and ~1100 leaves on mainnet - and jesse resolves every $ref by re-walking the whole decoded spec, which is a proplist. The numeric leaves of the new schemas therefore inline the UInt/UInt32/UInt64 shape instead of $ref-ing it, which cuts validation of a six-protocol body by ~4.2x (measured against both specs with the real validator). Named object schemas stay $ref'd so SDK generators still emit types. Inlining is also what makes the field descriptions reach clients at all: OpenAPI 3.0 ignores a $ref's siblings, and so does jesse, which was silently dropping ~23 of them - including "absent when the protocol imposes no limit" and the rule that name lengths are measured after IDNA conversion.

aehttp_cache_etag gains a strong validator. Its inputs are cheap only relative to the memoised expansion - they are ten config-map expansions and a dirty chain read - but a 304 skips schema validation and encoding of the whole body, so it pays for itself many times over. The int-as-string parameter is normalised to the boolean the request pipeline will make of it, so that an absent parameter and an explicit false share one validator instead of fragmenting a shared cache across identical bodies.

Supporting changes outside the endpoint:

  - aec_tx_pool:tx_ttl/0 and nonce_offset/0 were exported only under -ifdef(TEST). CT nodes are built from the dev1 profile, which defines TEST, so the endpoint worked under test and returned 500 on every production build. They are plain config readers and now sit in the main export list next to minimum_miner_gas_price/0.

  - aec_governance specs for name_pointers_max_count/1 and name_pointer_max_key_size/1 said 'inifinity'. The typo made dialyzer treat the infinity branch of the new add_unless_unlimited/3 as unreachable, failing the static-analysis build.

  - aetx:tx_types/0 enumerates the aetx:tx_type() union at runtime, next to the union itself. The endpoint's fee tables must cover every transaction type, and two hand-written copies of the list - one in the handler, one in the test - cannot detect a type neither knows about. The handler still partitions the list explicitly, so that a new type cannot reach aec_governance:tx_base_gas/2 (which has no catch-all) and turn every request into a 500; the test derives its expected key set from aetx:tx_types/0 and fails when the partition stops covering it.

block_interval_setting/0 gains a catch-all clause: aec_consensus: get_consensus_type/0 returns only pow | pos today, but a third consensus type should omit the block interval rather than 500 the request. The Hyperchains suite now exercises the pos branch, which the integration suite - always a proof-of-work node - never reaches. Both node mutations in the integration test are restored from end_per_testcase as well as from the case body, since a CT timetrap runs neither `after` nor the rest of the case, and the suite shares one node with the later naming group.

### DIFF
--- a/apps/aecore/src/aec_governance.erl
+++ b/apps/aecore/src/aec_governance.erl
@@ -16,8 +16,11 @@
          name_preclaim_expiration/0,
          name_claim_locked_fee/0,
          name_claim_fee/2,
+         name_claim_fee_for_size/2,
          name_claim_bid_timeout/2,
+         name_claim_bid_timeout_for_size/3,
          name_claim_bid_extension/2,
+         name_claim_bid_extension_for_size/3,
          name_claim_bid_increment/0,
          name_claim_max_expiration/1,
          name_protection_period/0,
@@ -265,13 +268,13 @@ name_claim_max_expiration(Protocol) when Protocol < ?IRIS_PROTOCOL_VSN ->
 name_claim_max_expiration(_Protocol) ->
     180000. %% At 480 generations per day this is 375 days, i.e. ~1 year.
 
--spec name_pointers_max_count(aec_hard_forks:protocol_vsn()) -> non_neg_integer() | inifinity.
+-spec name_pointers_max_count(aec_hard_forks:protocol_vsn()) -> non_neg_integer() | infinity.
 name_pointers_max_count(Protocol) when Protocol < ?IRIS_PROTOCOL_VSN ->
     infinity;
 name_pointers_max_count(_Protocol) ->
     32.
 
--spec name_pointer_max_key_size(aec_hard_forks:protocol_vsn()) -> non_neg_integer() | inifinity.
+-spec name_pointer_max_key_size(aec_hard_forks:protocol_vsn()) -> non_neg_integer() | infinity.
 name_pointer_max_key_size(Protocol) when Protocol < ?IRIS_PROTOCOL_VSN ->
     infinity;
 name_pointer_max_key_size(_Protocol) ->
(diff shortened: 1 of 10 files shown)


venimus:
Reviewed the diff and compiled the branch locally. Solid PR overall - two things worth addressing before merge, plus two nice bug catches along the way.

**1. `current_protocol_version/0` calls the function it argues against**

```erlang
current_protocol_version() ->
    case aec_chain:dirty_top_header() of
        undefined -> aec_hard_forks:protocol_effective_at_height(0);
        Header    -> aec_headers:version(Header)
    end.
```

The comment above this function explains why `aec_hard_forks:protocol_effective_at_height/1` is deliberately avoided here (recomputes the fork table, and can disagree with the block just accepted during community-fork signalling). `aec_hard_forks.erl` itself documents: *"This function is supposed to be used only when [inserting a block / preparing a candidate]... shouldn't be used elsewhere (apart from tests)."* The `undefined` fallback branch calls exactly that function in production code. Low practical risk (only reachable before any header exists), but it contradicts the rationale given right above it - worth either removing the branch (can a running node ever have no top header?) or acknowledging the exception explicitly.

**2. Always-string aettos fields are a silent deviation from the existing API convention**

`minimum_gas_price`, `name_claim_fees`, `name_claim_locked_fee`, and `min_miner_gas_price` are hard-coded via `aettos/1` to always serialize as decimal strings, ignoring `int-as-string` - justified by the name-fee table reaching ~2^69, past the `2^53` an IEEE-754 client can hold. But every other aettos amount on the external API (`Account.balance`, `SpendTx.amount`/`fee`, channel amounts, etc.) uses the shared `UInt`/`UInt64` schema and returns a raw JSON number by default, with exactly the same precision hazard (balances can exceed 2^53 too). This PR introduces a stricter, one-off rule for a handful of fields without reconciling it with the identical issue elsewhere in the API.

Smaller nit from the same issue: the OAS3 schema for these 4 fields still declares `oneOf: [integer, string]`, which is misleading since they're never actually integers - `type: string` would document actual behavior and avoid surprising SDK codegen.

**Nice catches confirmed while reviewing:**
- `aec_tx_pool:nonce_offset/0` and `tx_ttl/0` were exported only under `-ifdef(TEST)` while the function bodies are unconditional - confirmed any non-test build's export list omitted them, so this (and anything else calling them) would 500 in production only, never under CT. Good fix.
- The `-spec` typo `inifinity` -> `infinity` in `aec_governance.erl` was making dialyzer treat the `infinity` branch as unreachable, confirmed real.

Tests are unusually thorough (tx-type coverage swept from `aetx:tx_types/0`, VM/ABI sweep wider than what exists today, ETag/cache-fragmentation assertions, bid-timeout and block-gas-limit live-override checks with proper `end_per_testcase` restores against CT timetraps) - nothing there looks like a bad test or an untested code path. Compiles clean with only pre-existing unrelated warnings.

venimus (approved):

